PR #5, opened 2019-03-31
Fix neqo-client build

quininer:
:)


martinthomson:
Thanks, but the .drain(..) thing is unnecessary.  I fixed this with vec![] instead.

PR #7, opened 2019-05-02
New data helpers

martinthomson:
This introduces new data helpers.

The old `Data` has been split into two:

`Decoder` is for decoding.  It's designed to borrow a buffer and not copy anything, but it shouldn't be unpleasant to use for that purpose.  `neqo-transport` should find that this is enough.

`Encoder` is for building messages.  It owns the buffer it uses.

`IncrementalDecoder` is for reading incomplete pieces.  This is similar in function to the stuff that was in `readbuf.rs`, but the API is fairly different.  The intent is that this is enough for the HTTP pieces where we have to deal with partial input.  The comments explain how this works and tests show more detail.

There are more tests for these than the old code.  I think that there is fairly thorough coverage, but I haven't used a tool for that yet.

This only changes `neqo-common`.  The other crates will fail to compile with these changes.  Before I start on that, I'd like feedback on the shape of the API.


quininer:
Old Data looks similar to [`bytes`](https://docs.rs/bytes/), but `bytes` use reference counting and have less copy overhead.

agrover:
### Decoder

Decoder looks fine but has some capabilities I don't understand. The way I would think one would use this is to decode values from a Decoder-wrapped buf, and in doing so the "remaining" part would get smaller until there was none left. But, is there a need to ever backtrack? Methods like `len()` and `seek()` seem to indicate there are cases where users of Decoder need to care how big the original backing allocation was, and I'm not sure if that's true.

API-wise, maybe suggest having `decode_` methods return `Option` instead of `Res`, which aligns better with "not there" semantics, and then leave it up to the user of Decoder to turn option into a Res if appropriate.

### Encoder

:+1: 

### IncrementalDecoder

:100:  I didn't care for ReadBuf at all tbh, this looks good.

### Other stuff 

Just to touch on what quininer mentioned, I do think we should look at the bytes crate and consider using it, to minimize bulk copies. We have one unavoidable one in decrypt_packet (or...any way to decrypt in place? Maybe it's not worth it), but we are copying entire stream frames from the packet in decode_frame, that seems like it could be a good opportunity to leave the data from multiple streams' frames in a single refcounted packet buffer to avoid that copy.

We're also pretty free with copies on the send side, if you have any thoughts on how we can structure things to minimize those.

martinthomson:
Good feedback.

> Decoder looks fine but has some capabilities I don't understand. The way I would think one would use this is to decode values from a Decoder-wrapped buf, and in doing so the "remaining" part would get smaller until there was none left. But, is there a need to ever backtrack? Methods like len() and seek() seem to indicate there are cases where users of Decoder need to care how big the original backing allocation was, and I'm not sure if that's true.

The `seek()` and `len()` functions were speculative.  I did use `len()` in a test, but it was the `seek()` test.  I have removed them.  

Was there anything else that looked bad?  Hopefully the `decode_xxx()` functions are OK.

> API-wise, maybe suggest having decode_ methods return Option instead of Res, which aligns better with "not there" semantics, and then leave it up to the user of Decoder to turn option into a Res if appropriate.

I considered that, but the short-term appeal of the `?` operator won out over the architectural purity argument.  I have switched to `Option<T>` though, it's a tiny bit less ergonomic, but better overall (and mostly API-compatible).

> Just to touch on what quininer mentioned, I do think we should look at the bytes crate and consider using it, to minimize bulk copies. 

I don't think that we need bytes.  It's good, but more fully-featured than what we need.  And we would need to add a wrapper to get the varint encoding and that sort of thing.  I don't think that we need to worry about refcounting, because we have fairly good control over lifetime of buffers generally.

> We have one unavoidable one in decrypt_packet (or...any way to decrypt in place? Maybe it's not worth it), 

We can do encryption and decryption in place.  NSS supports it.  I'd need to add another API, because the borrow checker would have an aneurysm if we tried to use the existing API.  Let's start using issues to track that sort of thing.

> we are copying entire stream frames from the packet in decode_frame, that seems like it could be a good opportunity to leave the data from multiple streams' frames in a single refcounted packet buffer to avoid that copy.

One reason I don't think we need to use bytes is that I don't think we need refcounting for these buffers.  We should be able to lift everything we need from the buffer inline.  We'll be taking a copy for stream data, but that is all.  There is an optimization we might consider that uses the packet buffer as the backing for stream buffers, but we can do that when the time comes.

> We're also pretty free with copies on the send side, if you have any thoughts on how we can structure things to minimize those.

I don't think that we have to actively avoid copies, but we could certainly get better.  Things like the `encode_vec_with()` function I added should help with the more egregious instances.  (There's an instance in the ACK frame that really sucks, but we should just fix the ACK frame encoding in the way Intel would prefer we fix it and the problem will go away.)

agrover:
> Was there anything else that looked bad? Hopefully the decode_xxx() functions are OK. 

hmm, well..... all uses of decode_uint currently in the repo take length 1,2, or 4 so maybe having decode_u{8,16,32}() would be all that's needed, rather than allowing all values 1..8 inclusive.

But anyway, I'm guessing the next step is to convert over existing code?

martinthomson:
Yeah, I will start on that soon.

I didn't change the decode_uint() like that, though it makes a degree of sense.  NSS uses 1, 2, 3, and 4 for different reasons, so I thought that this was probably OK for now.  The optimization for the general case isn't as good as the specific case, so maybe we can look into whether there is any gain to splitting these later.

martinthomson:
I've converted.  And now I think that it's time that I look more closely at this business with random CI failures.  It's gotten to the point where it is annoying.

PR #30, opened 2019-05-27
Fix clippy errors

### LINKED ISSUE #29: Fix Clippy issues in neqo-crypto
Disabled running clippy on some files in neqo-crypto, see lib.rs. It would be great to re-enable it, but I don't have enough knowledge of the code to determine if Clippy is finding things that should be fixed or not.

martinthomson:
Closes #29.

This adds clippy to the CI build as well.


agrover:
Wow that was fast.

Can we use stable instead of a nightly snapshot?

PR #45, opened 2019-06-13
0_RTT

### LINKED ISSUE #42: Establish common time base for tests using Once
https://github.com/mozilla/neqo/pull/39#discussion_r292239695

> This changes the tests significantly. It exposes the tests to timing variance that the fixed value didn't. If we can fix the value here, that will help a lot.

> I like that this now uses now() instead of 0 in a lot of places, but I think that we want a constant function. In doing the 0-RTT code, I need a lot more control over the way that we manage timers and tying this to the system clock is unworkable.

> I know that it's hard to get a concrete Instant instance, but maybe we can use Once for picking that point in time.

martinthomson:
With #39 done, I have most of what I need for 0-RTT.  This does a lot unfortunately.  Something for @agrover to spend some time looking at.

There are a bunch of NSS patches that need to be applied to get this to work, so this will block on those landing.

Bug 1558681 provides anti-replay contexts, which allows us to run 0-RTT tests concurrently
Bug 1556591 removes some racey stuff
Bug 1555207 adds an option to just reject 0-RTT in the HRR callback

All of these are still awaiting review as of now.

Closes #42, #13. (Not automatically, because I edited this comment to add those.)


neqo-common/src/once.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub const fn new() -> Self {
+        OnceResult {
+            once: Once::new(),
+            v: None,
+        }
+    }
+
+    pub fn call_once<F: FnOnce() -> T>(&mut self, f: F) -> &T {
+        // Using a pointer to access `v` allows us up dupe the borrow checker.
+        let ptr: *mut Option<T> = &mut self.v;
```
martinthomson: Is this something that would better with `RefCell` for `self.v`?

martinthomson: Or `UnsafeCell`?

neqo-transport/src/connection.rs
```diff
@@ -921,11 +966,26 @@ impl Connection {
...
 
         if self.state == State::Init {
-            let res = self.client_start();
+            let res = self.client_start(cur_time);
             self.absorb_error(cur_time, res);
         }
     }
 
+    /// Get the time that we next need to be called back, relative to `now`.
+    fn next_delay(&self, now: Instant) -> Option<Duration> {
```
agrover: you wanna call it `now` or you wanna call it `cur_time`? Let's pick one.

neqo-transport/src/connection.rs
```diff
@@ -1208,26 +1286,36 @@ impl Connection {
...
+                };
+                if cs.tx.is_none() {
+                    continue;
+                }
+            }
+
+            // Always send an ACK if there is one to send.
+            if self.acks[space].ack_now(cur_time) {
+                let mut recvd = mem::replace(&mut self.acks[space], RecvdPackets::new(space));
+                if let Some((frame, token)) = recvd.generate(
```
agrover: This looks like it wants to go on the list of generators, but I'm guessing something prevents this? If so, could you open an issue and describe what's blocking it?

agrover:
afaik this is now only blocked by the needed NSS changes (I'm assuming that's what's causing the compilation error) and a couple review comments.

neqo-transport/src/connection.rs
```diff
@@ -993,20 +1049,24 @@ impl Connection {
...
-                    if self.retry_token.is_some() {
-                        qwarn!("received another Retry, dropping it");
-                        return Ok(());
+                    {
+                        let path = self.paths.iter_mut().find(|p| p.remote_cid == *odcid);
+                        if path.is_none() {
+                            qwarn!("received Retry, but not for us, dropping it");
+                            return Ok(());
+                        }
+                        path.unwrap().remote_cid =
```
quininer: avoiding `unwrap` with `if let Some()` will be fine.

agrover: :+1:

neqo-common/src/once.rs
```diff
@@ -0,0 +1,42 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::sync::Once;
+
+pub struct OnceResult<T> {
```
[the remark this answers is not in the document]
agrover: opened #61

PR #53, opened 2019-06-14
Add pre-commit hook for format

martinthomson:
I encourage everyone to enable this.  It's fast enough to run on every commit, which saves a bunch of mucking around with formatting.  It's also accurate - it only checks staged changes, so it won't be fooled by stuff that you don't plan to commit.


agrover:
useful, thanks!

Issue #20, opened 2019-05-20
Server class


martinthomson:
We probably need a server class at the transport layer that manages routing of packets to different connections, based on connection ID.

agrover:
Should this be called Endpoint??

martinthomson:
This shouldn't be called Endpoint, because that refers to one end of a connection.  If anything, Connection could be called Endpoint, but I think that Connection is the right name.

Issue #37, opened 2019-06-07
Refactor generators


agrover:
Right now, generators are just a little dynamic -- there are three up until the connection enters Closing state, and then the three get replaced by the one CloseGenerator.

This could be improved by being a little more dynamic and fine-grained. For example, right now StreamGenerator iterates through streams and each stream presents available data in lowest to highest offset so retransmit ranges are first, but between streams, new data for one stream could go out before retransmitted bytes for another, which is unfortunate.

This just all needs a rethink and refactor.

agrover:
I'm currently thinking we don't actually need a list of generators. If we're in a Connected state, there are a series of packet sources we should check when attempting to build a datagram -- acks, flow control, re-sent stream data, new stream data, etc. This list is always ordered the same, isn't it? The only other state is Closing, where we have a list with just one thing on it -- send CloseConnection frame.

We could consider a much less flexible solution that does not require the generator abstraction.

PR #82, opened 2019-07-05
Operate on one datagram at a time

martinthomson:
When trying to add Retry for a server, I found that the API that handles
multiple packets was inconvenient.  There is no real need to have the
loop inside Connection anyway, so I externalized it.  This simplifies
the API and implementation a decent amount and should help avoid
problems.

This isn't complete at the moment because we don't set any timeout when
there is more data to send.  As a result, if we have more to say than
fits in a single packet, we won't say it until given the opportunity.
For the moment, that means either hitting the loss recovery timer or
getting a new packet inbound.  The way that the generators are
structured makes it very difficult to do that right now because we have
a single generator squatting on the queue always.

As part of fixing process_input and process_output, I removed the idle
timeout code that was there.  It wasn't in any way right, so I am not
concerned about its loss; I just moved the TODO.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -123,8 +123,8 @@ trait Handler {
     fn handle(&mut self, client: &mut Http3Connection) -> bool;
 }
 
-fn emit_packets(socket: &UdpSocket, out_dgrams: &[Datagram]) {
-    for d in out_dgrams {
+fn emit_datagram(socket: &UdpSocket, d: Option<Datagram>) {
+    if let Some(d) = d {
         let sent = socket.send(&d[..]).expect("Error sending datagram");
         if sent != d.len() {
             eprintln!("Unable to send all {} bytes of datagram", d.len());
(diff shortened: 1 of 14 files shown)


agrover:
> When trying to add Retry for a server, I found that the API that handles
multiple packets was inconvenient. 

Retry made this inconvenient how?

> This isn't complete at the moment because we don't set any timeout when
there is more data to send. As a result, if we have more to say than
fits in a single packet, we won't say it until given the opportunity.
For the moment, that means either hitting the loss recovery timer or
getting a new packet inbound.

Should we be defining the API such that clients are required to call process_output repeatedly until they get `None`, or something?

> The way that the generators are
structured makes it very difficult to do that right now because we have
a single generator squatting on the queue always.

You're talking about fairness within the stream generator when multiple streams have data ready to send? If it's just within the stream generator, maybe we can come up with a straightforward solution without depending on #37.

martinthomson:
> Retry made this inconvenient how?

Sorry for being terse.  My ideal design for retry has a Server class that manages the server as a whole.  Retry (and version negotiation) isn't generated by individual connections, but it is something the Server does.  That allows the server to generate Retry without instantiating a Connection.

In doing that, I'm trying to fix #20.  And having multiple datagrams handled there was quite tricky.  You have to iterate over each and route to different connections.  Then you have to worry about the order of joining any datagrams that each connection sends.

> Should we be defining the API such that clients are required to call process_output repeatedly until they get None, or something?

I'm thinking that if the Duration that comes out is zero, then we would ask them to run the function again.  Ideally, that doesn't happen.  More likely, what happens is that the time that comes back is smaller than their timer granularity, at which point they might want to call in again anyway just to get the next packet to send ahead of time.

> You're talking about fairness within the stream generator when multiple streams have data ready to send? 

I'm thinking that it would be rather nice if frame generators were to only sit in the queue if they had something to say.  Then we could do `queue.is_empty()` to determine whether there is more data waiting.  Then we could generate a Duration from `Connection::process_output` that includes the pacing timer in the calculation, in addition to the loss recovery and delayed acknowledgment timers.

agrover:
OK I see where this is going and I'm :+1: . I know there's a lot of to-be-done work around generators and Server, but I'd really like it if this PR could make its changes without giving up the current behavior when there are multiple datagrams to send. We can address queue fairness in a later PR but that (while important) is less important than not blocking a second out packet on an incoming packet arriving or a timer expiring.

agrover:
rebased.

PR #106, opened 2019-07-30
Expose the info needed for an external certificate verification.

ddragana:
I am not really familiar with the certificate verification in psm. This works. Please give me any suggestion if there is a better way.

I will discuss the changes to gecko/psm code with Dana.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -5,14 +5,13 @@
 // except according to those terms.
 
 #![deny(warnings)]
-use neqo_common::Datagram;
-use neqo_crypto::init_db;
+use neqo_common::{matches, Datagram};
+use neqo_crypto::init;
 use neqo_http3::{Http3Connection, Http3Event, Http3State};
 use neqo_transport::Connection;
 use std::collections::HashSet;
 use std::io::{self, ErrorKind};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, ToSocketAddrs, UdpSocket};
-use std::path::PathBuf;
 use std::process::exit;
 use std::str::FromStr;
 use std::string::ParseError;
@@ -61,9 +60,6 @@ impl FromStr for Headers {
     about = "A basic QUIC HTTP/0.9 and HTTP3 client."
 )]
 pub struct Args {
-    #[structopt(short = "d", long, default_value = "./db", parse(from_os_str))]
-    db: PathBuf,
-
     #[structopt(short = "a", long, default_value = "h3-22")]
     /// ALPN labels to negotiate.
     ///
@@ -180,6 +176,10 @@ fn process_loop(
 struct PreConnectHandler {}
 impl Handler for PreConnectHandler {
     fn handle(&mut self, _args: &Args, client: &mut Http3Connection) -> bool {
+        let authentication_needed = |e| matches!(e, Http3Event::AuthenticationNeeded);
+        if client.events().into_iter().any(authentication_needed) {
+            client.authenticated(0, Instant::now());
+        }
         Http3State::Connected != client.state()
     }
 }
@@ -244,7 +244,6 @@ fn client(args: Args, socket: UdpSocket, local_addr: SocketAddr, remote_addr: So
             args.alpn.clone(),
             local_addr,
             remote_addr,
-            false,
         )
         .expect("must succeed"),
         args.max_table_size,
@@ -285,8 +284,8 @@ fn client(args: Args, socket: UdpSocket, local_addr: SocketAddr, remote_addr: So
 }
 
 fn main() {
+    init();
     let args = Args::from_args();
-    init_db(args.db.clone());
 
     let remote_addr = match args.remote_addr() {
         Err(e) => {
@@ -442,7 +441,6 @@ mod old {
             vec!["http/0.9"],
             local_addr,
             remote_addr,
-            false,
         )
         .expect("must succeed");
         // Temporary here to help out the type inference engine
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -244,6 +244,7 @@ fn client(args: Args, socket: UdpSocket, local_addr: SocketAddr, remote_addr: So
             args.alpn.clone(),
             local_addr,
             remote_addr,
+            false,
         )
         .expect("must succeed"),
         args.max_table_size,
@@ -441,6 +442,7 @@ mod old {
             vec!["http/0.9"],
             local_addr,
             remote_addr,
+            false,
         )
         .expect("must succeed");
         // Temporary here to help out the type inference engine
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -10,7 +10,7 @@ use crate::assert_initialized;
 pub use crate::cert::CertificateChain;
 use crate::constants::*;
 use crate::convert::to_c_uint;
-use crate::err::{Error, Res};
+use crate::err::{Error, PRErrorCode, Res};
 use crate::ext::{ExtensionHandler, ExtensionTracker};
 use crate::p11;
 use crate::prio;
@@ -37,7 +37,7 @@ pub enum HandshakeState {
     New,
     InProgress,
     AuthenticationPending,
-    Authenticated,
+    Authenticated(PRErrorCode),
     Complete(SecretAgentInfo),
     Failed(Error),
 }
@@ -146,6 +146,7 @@ pub struct SecretAgentInfo {
     resumed: bool,
     early_data: bool,
     alpn: Option<String>,
+    signature_scheme: SignatureScheme,
 }
 
 impl SecretAgentInfo {
@@ -166,6 +167,7 @@ impl SecretAgentInfo {
             resumed: info.resumed != 0,
             early_data: info.earlyDataAccepted != 0,
             alpn: get_alpn(fd, false)?,
+            signature_scheme: info.signatureScheme as SignatureScheme,
         })
     }
 
@@ -186,6 +188,9 @@ impl SecretAgentInfo {
     }
     pub fn alpn(&self) -> Option<&String> {
         self.alpn.as_ref()
+    }
+    pub fn signature_scheme(&self) -> SignatureScheme {
+        self.signature_scheme
     }
 }
 
@@ -213,7 +218,7 @@ pub struct SecretAgent {
 }
 
 impl SecretAgent {
-    fn new() -> Res<SecretAgent> {
+    fn new(external_cert_verification: bool) -> Res<SecretAgent> {
         let mut agent = SecretAgent {
             fd: null_mut(),
             secrets: Default::default(),
@@ -230,7 +235,7 @@ impl SecretAgent {
 
             no_eoed: false,
         };
-        agent.create_fd()?;
+        agent.create_fd(external_cert_verification)?;
         Ok(agent)
     }
 
@@ -241,8 +246,10 @@ impl SecretAgent {
     // minimal, but it means that the two forms need casts to translate
     // between them.  ssl::PRFileDesc is left as an opaque type, as the
     // ssl::SSL_* APIs only need an opaque type.
-    fn create_fd(&mut self) -> Res<()> {
-        assert_initialized();
+    fn create_fd(&mut self, external_cert_verification: bool) -> Res<()> {
+        if !external_cert_verification {
+            assert_initialized();
+        }
 
         let label = CString::new("sslwrapper").expect("cstring failed");
         let id = unsafe { prio::PR_GetUniqueIdentity(label.as_ptr()) };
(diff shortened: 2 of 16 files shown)


neqo-transport/src/connection.rs
```diff
@@ -192,11 +195,12 @@ impl Connection {
         protocols: PI,
         local_addr: SocketAddr,
         remote_addr: SocketAddr,
+        external_cert_verification: bool,
     ) -> Res<Connection> {
         let dcid = ConnectionId::generate(CID_LENGTH);
         let mut c = Connection::new(
             Role::Client,
-            Client::new(server_name)?.into(),
+            Client::new(server_name, external_cert_verification)?.into(),
```
martinthomson: Any change for the client needs to be mirrored for the server.  A server also gets a peer certificate that might need authentication, even if the certificate is optional (and currently not supported).

neqo-crypto/src/agent.rs
```diff
@@ -492,11 +499,31 @@ impl SecretAgent {
...
     }
 
+    /// Get the peer's certificate.
+    pub fn peer_certificate(&self) -> *mut ssl::CERTCertificate {
+        unsafe { ssl::SSL_PeerCertificate(self.fd) }
+    }
+
     /// Get the peer's certificate chain.
-    pub fn peer_certificate(&self) -> Option<CertificateChain> {
+    pub fn peer_certificate_chain(&self) -> Option<CertificateChain> {
```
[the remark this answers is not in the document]
ddragana: I wanted this to return ssl::CERTCertList so that I do not need to convert it from CertificatChain to ssl::CERTCertList again.

neqo-crypto/src/agent.rs
```diff
@@ -555,8 +582,8 @@ impl SecretAgent {
             // Within this scope, _h maintains a mutable reference to self.io.
             let _h = self.io.wrap(input);
             match self.state {
-                HandshakeState::Authenticated => unsafe {
-                    ssl::SSL_AuthCertificateComplete(self.fd, 0)
+                HandshakeState::Authenticated(ref err) => unsafe {
```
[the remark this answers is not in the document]
ddragana: We can. Here we are only being nice and sending the allert back to the peer.

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
ddragana: This should work, I do not remember why I made the other one.

neqo-transport/src/connection.rs
```diff
@@ -746,11 +785,28 @@ impl Connection {
         out
     }
 
+    fn get_last_epoch(&mut self) -> u16 {
```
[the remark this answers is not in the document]
ddragana: When certificate validation fails, we are sending CONNECTION_CLOSE frame with a cert error. Without this change we would send it in a Initial packet and server would drop it. All in all we need this so that CONNECTION_CLOSE is sent in right packet.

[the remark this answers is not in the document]
martinthomson: That seems like a separable piece from this.  For this specific PR, all we need to do is walk away from a bad connection (closing the port is fine).

If we want to properly signal CONNECTION_CLOSE, then we should do that, but I think that the way you do that is the frame builder just builds frames in both epochs.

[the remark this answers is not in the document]
ddragana: the frame builder is not different than before. It does exactly the same thing as before. We ere checking if we have keys for each epoch. The only thing that has change here is to send CONNECTION_CLOSE in the newst epoch not in the first one (initial epoch).

I will open a separate issue for this and remove this change from this PR. We can discuss if we want to do this in the new issue.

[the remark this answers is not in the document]
martinthomson: A new issue would be good.  From what I can see, the frame builder should be adding the packet to every epoch on its on recognizance.  Right now, we track whether CONNECTION_CLOSE has been sent, but we really need something more fine-grained.  I see now that we only send CONNECTION_CLOSE in epoch 3, which is a mistake, so let's open an issue.

To add to that, we won't get loss signals for packets containing CONNECTION_CLOSE, we have to drive sending of the frame based on receipt of packets.

test-fixture/src/lib.rs
```diff
@@ -75,9 +75,13 @@ pub fn handshake(client: &mut Connection, server: &mut Connection) {
     let mut b = server;
     let mut datagram = None;
     let is_done = |c: &Connection| matches!(c.state(), State::Connected | State::Closing { .. } | State::Closed(..));
-    while !is_done(a) {
+    let authentication_needed = |e| matches!(e, ConnectionEvent::AuthenticationNeeded);
+    while !is_done(a) || !is_done(b) {
         let d = a.process(datagram, now());
         datagram = d.dgram();
+        if a.events().any(authentication_needed) {
+            a.authenticated(0, now());
```
martinthomson: This is great, but can we add a function to the fixture to conditionally authenticate?  I see a lot of code in the above tests that effectively duplicate this.  It might need to return a boolean so that callers can assert that authentication happened as expected.

One thing to note (perhaps in a comment) is that this is destructive.  It consumes any outstanding events on the connection.  That's not a problem for this function, but it is worth noting because it's a potentially surprising side-effect.

neqo-transport/src/connection.rs
```diff
@@ -1751,6 +1784,18 @@ mod tests {
...
+        qdebug!("---- client: cert verification");
+        let out = client.process(out.dgram(), now());
+        assert!(out.as_dgram_ref().is_some());
+        qdebug!("Output={:0x?}", out.as_dgram_ref());
+
+        let out = server.process(out.dgram(), now());
+        assert!(out.as_dgram_ref().is_none());
+
+        let authentication_needed = |e| matches!(e, ConnectionEvent::AuthenticationNeeded);
+        assert!(client.events().into_iter().any(authentication_needed));
```
martinthomson: Can you run `cargo clippy --tests` and fix anything that comes up?  Assuming that #122 is merged ahead of this, that is.  Otherwise, I think that you need to do this (and elsewhere as well):

```suggestion
        assert!(client.events().any(authentication_needed));
```

neqo-transport/src/connection.rs
```diff
@@ -946,20 +979,14 @@ impl Connection {
             });
         }
 
-        let m = {
-            let m = self.crypto.tls.handshake_raw(now, rec);
+        let m = self.crypto.tls.handshake_raw(now, rec);
+        if (*self.crypto.tls.state() == HandshakeState::AuthenticationPending)
+            && !self.authentication_needed_notified
```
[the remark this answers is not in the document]
ddragana: Actually we can drop this. an app using neqo should take care of this.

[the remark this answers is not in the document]
martinthomson: How so?  Completing authentication is a prerequisite for completing the handshake and you need to know when to start validating the cert.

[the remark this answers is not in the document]
ddragana: I meant only authentication_needed_notified. That is what I understood from your first comment comment (remove authentication_needed_notified and dedup authentication_needed events).
the event is going to stay.

neqo-client/src/main.rs
```diff
@@ -284,8 +284,8 @@ fn client(args: Args, socket: UdpSocket, local_addr: SocketAddr, remote_addr: So
 }
 
 fn main() {
+    init();
```
martinthomson: So you aren't reading the database now?  That's OK until we need a client certificate.

test-fixture/src/lib.rs
```diff
@@ -70,14 +70,26 @@ pub fn default_server() -> Connection {
...
+    false
+}
+
 pub fn handshake(client: &mut Connection, server: &mut Connection) {
     let mut a = client;
     let mut b = server;
     let mut datagram = None;
     let is_done = |c: &Connection| matches!(c.state(), State::Connected | State::Closing { .. } | State::Closed(..));
-    while !is_done(a) {
+    while !is_done(a) || !is_done(b) {
```
martinthomson: This change isn't necessary if you change the statement below to:

```rust
if !maybe_authenticate(a) {
  // If we authenticated a, keep a in place.
  mem::swap(&mut a, &mut b);
}
```

Or you could do as above: https://github.com/mozilla/neqo/pull/106/files#diff-46de255ccea3f53c2144ec0036f9698bR1963

neqo-transport/src/connection.rs
```diff
@@ -1770,6 +1804,49 @@ mod tests {
...
+        qdebug!("Output={:0x?}", out.as_dgram_ref());
+
+        let out = server.process(out.dgram(), now());
+        assert!(out.as_dgram_ref().is_none());
+        qdebug!("Output={:0x?}", out.as_dgram_ref());
+
+        let authentication_needed = |e| matches!(e, ConnectionEvent::AuthenticationNeeded);
+        assert!(client.events().any(authentication_needed));
+        qdebug!("---- client: Alert(certificate_revoked)");
+        client.authenticated(-(0x2000) + 12, now());
```
martinthomson: This is where I think that this API is a little troublesome.  Getting the right `PRErrorCode` value seems to break encapsulation badly.  Maybe don't fix that now, but open an issue to track passing an enum in here instead.

neqo-transport/src/connection.rs
```diff
@@ -1679,6 +1688,17 @@ mod tests {
         .expect("create a default server")
     }
 
+    /// If state is AuthenticationNeeded call authenticated(). This funstion will
```
```suggestion
    /// If state is AuthenticationNeeded call authenticated(). This function will
```

neqo-transport/src/connection.rs
```diff
@@ -374,6 +376,22 @@ impl Connection {
...
+        self.crypto.tls.info()
+    }
+
+    /// Get the peer's certificate chain and other info.
+    pub fn peer_certificate(&self) -> Option<CertificateInfo> {
+        self.crypto.tls.peer_certificate()
+    }
+
+    /// Call by application when the peer cert has been verified
+    pub fn authenticated(&mut self, error: err::PRErrorCode, now: Instant) {
```
martinthomson: See below about making `neqo_crypto::err::PRErrorCode` public.

neqo-crypto/src/lib.rs
```diff
@@ -17,15 +17,15 @@ mod agentio;
 mod cert;
 pub mod constants;
 mod convert;
-mod err;
+pub mod err;
```
martinthomson: I missed these previously, but I think that these are not good changes.  These were internal for a reason and making them public makes a lot more public than I would like.  It exposes all of the SSL FFI for instance.

Can you `pub use` just the types that you specifically want to expose here instead?

neqo-crypto/src/cert.rs
```diff
@@ -7,25 +7,80 @@
...
+impl CertificateInfo {
     pub(crate) fn new(fd: *mut PRFileDesc) -> Option<Self> {
-        let chain = unsafe { SSL_PeerCertificateChain(fd) };
-        let certs = match NonNull::new(chain as *mut CERTCertList) {
-            Some(certs_ptr) => CertList::new(certs_ptr),
-            None => return None,
-        };
-        let cursor = CertificateChain::head(&certs);
-        Some(CertificateChain { certs, cursor })
+        match peer_ccertificate_chain(fd) {
```
martinthomson:
```suggestion
        match peer_certificate_chain(fd) {
```

neqo-crypto/src/cert.rs
```diff
@@ -7,25 +7,80 @@
...
     certs: CertList,
     cursor: *const CERTCertListNode,
+    /// stapled_ocsp_responses and signed_cert_timestamp are properties
+    /// associated with each of the certificates. Right now, NSS only
+    /// reports the value for the end-entity certificate (the first).
+    stapled_ocsp_responses: Option<Vec<Vec<u8>>>,
+    signed_cert_timestamp: Option<Vec<u8>>,
+}
+
+fn peer_ccertificate_chain(fd: *mut PRFileDesc) -> Option<(CertList, *const CERTCertListNode)> {
```
martinthomson:
```suggestion
fn peer_certificate_chain(fd: *mut PRFileDesc) -> Option<(CertList, *const CERTCertListNode)> {
```

neqo-transport/src/connection.rs
ddragana: I will open another issue.

test-fixture/src/lib.rs
ddragana: fix as above.

neqo-transport/src/connection.rs
ddragana: https://github.com/mozilla/neqo/issues/142

neqo-crypto/src/lib.rs
ddragana: done.

neqo-http3/src/connection.rs
```diff
@@ -1037,11 +1034,11 @@ impl Http3Events {
...
 
     pub fn authentication_needed(&mut self) {
-        self.events.insert(Http3Event::AuthenticationNeeded);
+        self.insert(Http3Event::AuthenticationNeeded);
     }
 
-    pub fn connected(&mut self) {
-        self.events.insert(Http3Event::ConnectionConnected);
+    pub fn goaway_received(&self) {
+        self.insert(Http3Event::GoawayReceived);
```
martinthomson: We don't also block new requests after receiving GOAWAY?

PR #122, opened 2019-08-07
Clippy tests

martinthomson:
VS Code runs clippy on tests.  I don't know how to do that from the command line (or else I would change our CI setup as well), but these are the problems I found and fixed.

The only thing I didn't fix were a couple of functions that were past the cognitive complexity limit.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -2095,7 +2095,7 @@ mod tests {
         hconn.process(out.dgram(), now());
 
         // Read first frame
-        match hconn.events().into_iter().skip(1).next().unwrap() {
+        match hconn.events().into_iter().nth(1).unwrap() {
             Http3Event::DataReadable { stream_id } => {
                 assert_eq!(stream_id, request_stream_id);
                 let mut buf = [0u8; 100];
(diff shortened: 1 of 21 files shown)


neqo-http3/src/connection.rs
```diff
@@ -1047,22 +1047,25 @@ mod tests {
...
         // connection on the other side so that we can check what the http3
         // side sends and also to simulate an incorrectly behaving http3
         // server/client.
 
-        let mut hconn;
-        let mut neqo_trans_conn;
-        if client {
-            hconn = Http3Connection::new(default_client(), 100, 100, None);
-            neqo_trans_conn = default_server();
+        let (mut hconn, mut neqo_trans_conn) = if client {
```
martinthomson: I don't like how clippy warns about this one.  But I'm not in a mood to suppress the change, even if it isn't great.

neqo-http3/src/connection.rs
```diff
@@ -1332,16 +1333,14 @@ mod tests {
         let events = neqo_trans_conn.events();
         let mut stop_sending_event_found = false;
         for e in events {
```
martinthomson: This is a common pattern.  I was uncertain about whether to change this to use Iterator::any, which would be better for looking for the event.  But that would be lots of typing.

o0Ignition0o:
> I don't know how to do that from the command line (or else I would change our CI setup as well)

[it seems](https://github.com/rust-lang/rust-clippy/issues/1436#issuecomment-462059561) `cargo clippy --tests` does the trick :)

martinthomson:
> [it seems](https://github.com/rust-lang/rust-clippy/issues/1436#issuecomment-462059561) `cargo clippy --tests` does the trick :)

I went with something a little broader than that. `cargo clippy --all-features --all-targets` gets more than just the one set of configs.  Gee it would be nice if `--help` actually listed these things.

martinthomson:
> Gee it would be nice if `--help` actually listed these things.

Oh, it does:

```
Other options are the same as `cargo check`.
```

PR #120, opened 2019-08-07
Add a comma after the optional bracketed expression in qlog!()

o0Ignition0o:
Fixes #91

PR #123, opened 2019-08-08
optional comma after the bracketed item in qlog!()

### LINKED ISSUE #91: Add a comma after the optional bracketed expression in qlog!()
The current arrangement doesn't get formatted because it doesn't look like real code. Adding a comma will allow rustfmt to format this as code.

```
qinfo!([self] "blah");
```
becomes
```
qinfo!([self], "blah");
```

martinthomson:
@o0Ignition0o, this seems to work.  I don't know what you were doing differently.

Closes #91, #120.

### DIFF
--- a/neqo-common/src/log.rs
+++ b/neqo-common/src/log.rs
@@ -46,26 +46,26 @@ macro_rules! qlog {
 }
 #[macro_export]
 macro_rules! qerror {
-    ([$ctx:expr] $($arg:tt)*) => ( ::neqo_common::qlog!(::log::Level::Error, $ctx, $($arg)*); );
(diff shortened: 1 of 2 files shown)


martinthomson:
Oh, I just added tests for this and something wonderful happens.  This change is OK...until you use the version with the comma.  It's like the rust version of [unique particle attribution](https://en.wikipedia.org/wiki/Unique_Particle_Attribution).  Everything old is new again.

o0Ignition0o:
Yeah that's really sad indeed.
I prefer your PR over mine because it has tests in it btw. so I'll close it :)

PR #125, opened 2019-08-08
Ensure that states don't regress

martinthomson:
There were cases when testing retry where a connection would go from
Closed to Closing.  That isn't good, so let's safeguard against that.


martinthomson:
Oops, this isn't complete...  More coming.

Issue #55, opened 2019-06-19
Retry -- Server


agrover:
like #15 but for server role.

martinthomson:
This landed a while back.

Issue #15, opened 2019-05-17
Retry -- client role


agrover:

martinthomson:
I have some code in there, but this needs:

1. tests
2. sending the original destination connection ID transport parameter from the server.
3. checking the original destination connection ID transport parameter at the client.

agrover:
@martinthomson are you still hacking on this or were you looking for someone to pick up the work for the things mentioned?

martinthomson:
Still hacking on this.  I just needed to write down what needs to happen.

agrover:
@martinthomson let me know if you can work on this soon, or if I should pick it up. This is currenly blocking us from interoperating with F5, Cloudflare, and probably also other server implementations too.

PR #127, opened 2019-08-08
Retry

martinthomson:
This is a big set of changes (sorry), but once it is rebased, the core set are hard to disentangle.  This is, until it is rebased, just a draft for the purposes of discussion and preview.

The main change here is to create a proper, functioning `Server` class that manages multiple server connections.  It does so with the aid of a new indirection for connections: a `ConnectionIdManager`.  That trait embodies both the ability to generate new connection IDs and the ability to parse them out of short header packets.

The new `Server` has some added responsibilities:

* It manages which connection has the next timer.
* It adds connection IDs to a map as the connection asks for them (by wrapping the `ConnectionIdManager` and capturing requests to generate connection IDs)
* It generates Version Negotiation and Retry packets as needed.

I haven't done:

* the self-encryption piece for Retry
* conversion of the HTTP server part to use `Server`
* conversion of command-line tools to use this

All of those will have to be follow-on issues and pull requests.

I started on the HTTP server conversion and ran into too many problems trying to disentangle client and server parts.  That code will need some serious rearchitecting to fit into this model.  There are a few small things that the `Server` class needs to add in order to support HTTP, but those are minor and can be added as part of any integration.


neqo-transport/src/connection.rs
```diff
@@ -245,11 +301,20 @@ impl Connection {
         tps.set_empty(tp_const::DISABLE_MIGRATION);
     }
 
-    fn new<A: ToString, I: IntoIterator<Item = A>>(
+    pub(crate) fn original_connection_id(&mut self, odcid: &ConnectionId) {
```
martinthomson: Depending on the outcome of Retry discussions, it might make sense to move this into the constructor.  A private constructor that included the ODCID would be a potential path forward.  We need to ensure that the server creates epoch 0 keys at the right time and that it doesn't end up creating them twice if it doesn't need to.

neqo-transport/src/connection.rs
```diff
@@ -831,13 +910,14 @@ impl Connection {
             }
 
             qdebug!([self] "Need to send a packet");
-            initial_only = epoch == 0;
+            // Packets without Handshake or 1-RTT need to be padded.
+            needs_padding = epoch <= 1;
```
martinthomson: Note that this is wrong, but the other PR is correct.

agrover:
fixes #55, looks like?

martinthomson:
This should fix a few different issues.  It does Retry for both server (#55) and client (#15), as well as providing a basic server at the transport layer (#20).

In reviewing this, I'm happy to take input on how this might be split into smaller pieces.

neqo-transport/src/send_stream.rs
```diff
@@ -542,8 +542,8 @@ impl SendStream {
 
     pub fn mark_as_lost(&mut self, offset: u64, len: usize, fin: bool) {
         if let Some(buf) = self.state.tx_buf_mut() {
-            buf.mark_as_lost(offset, len)
-        };
+            buf.mark_as_lost(offset, len);
+        }
```
agrover: this looks weird... clearly we don't need the ; after the closing brace, but do we really instead need it after the statement, or not?

neqo-transport/src/send_stream.rs
```diff
@@ -740,6 +747,12 @@ impl SendStreams {
         self.0.retain(|_, stream| !stream.is_terminal())
     }
 
+    pub(crate) fn retry(&mut self) {
+        for (_, stream) in self {
```
agrover: yes SendStreams implements &mut IntoIterator  but it might be clearer to use tuple notation inside SendStreams. e.g. `for stream in self.0.values()` instead here. That's what other methods are doing.

neqo-transport/src/server.rs
```diff
@@ -4,88 +4,423 @@
...
+const TIMER_GRANULARITY: Duration = Duration::from_millis(10);
+const TIMER_CAPACITY: usize = 1024;
 const FIXED_TOKEN: &[u8] = &[1, 2, 3];
 
+type StateRef = Rc<RefCell<ServerConnectionState>>;
+type CidMgr = Rc<RefCell<dyn ConnectionIdManager>>;
+type ConnectionTableRef = Rc<RefCell<HashMap<ConnectionId, StateRef>>>;
+
+#[derive(Debug)]
+struct ServerConnectionState {
```
agrover: if there is agreement on previous comment then this struct+impls is not needed.

neqo-transport/src/server.rs
```diff
@@ -4,88 +4,423 @@
...
+    /// as a result of calling `process()`.
+    pub fn active_connections(&mut self) -> Vec<ActiveConnectionRef> {
+        mem::replace(&mut self.active, Default::default())
+            .into_iter()
+            .collect()
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ActiveConnectionRef {
```
agrover: ok now can *this* go away and just use ConnectionRef?

neqo-transport/src/server.rs
```diff
@@ -4,88 +4,423 @@
...
+}
+
+impl RetryToken {
+    pub fn generate_token(&mut self, dcid: &ConnectionId) -> Vec<u8> {
+        let mut token = Vec::from(FIXED_TOKEN);
+        token.extend_from_slice(dcid);
+        token
+    }
+
+    pub fn require_retry(&mut self, retry: bool) {
```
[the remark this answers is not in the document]
agrover: `set_retry_required`?

neqo-common/src/lib.rs
```diff
@@ -11,6 +11,7 @@ mod datagram;
 mod incrdecoder;
 pub mod log;
 pub mod once;
+pub mod timer;
```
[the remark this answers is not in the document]
agrover: ok I've opened #190 and can come back to this someday.

neqo-client/src/main.rs
```diff
@@ -241,6 +241,7 @@ fn client(args: Args, socket: UdpSocket, local_addr: SocketAddr, remote_addr: So
         Connection::new_client(
             args.url.host_str().unwrap(),
             &args.alpn,
+            FixedConnectionIdManager::make(0),
```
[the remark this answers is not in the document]
agrover: I think the ownership semantics are better defined by the creator, rather than the createe (or at least, this is more idiomatic.) So we should rename `make` to `new` and then have the caller wrap with Rc::new(RefCell::new()) if that's what's called for.

PR #131, opened 2019-08-09
Splitted the request handler function into a RequestHandler type.

### LINKED ISSUE #41: Simplify type in http3::Connection
https://github.com/mozilla/neqo/pull/39#discussion_r292239230

o0Ignition0o:
WIP to fix #41 .
There's a couple of uses of &[Header] and Vec<Header>.

@martinthomson would it be worth considering switching every use to a Vec and having a Headers type? 

I'm not really sure of the performance implications, but it might make the code easier to read ?

### DIFF
--- a/neqo-http3-server/src/main.rs
+++ b/neqo-http3-server/src/main.rs
@@ -8,6 +8,7 @@
 
 use neqo_common::{qdebug, qinfo, Datagram};
 use neqo_crypto::{init_db, AntiReplay};
+use neqo_http3::request_stream_server::{Header, Response};
 use neqo_http3::{Http3Connection, Http3State};
 use neqo_transport::{Connection, Output};
 use std::collections::HashMap;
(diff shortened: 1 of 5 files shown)


agrover:
I know this is a WIP but looks fine to me so far.

o0Ignition0o:
Marking it as ready for review so we can keep the momentum.

There's still a couple of occurrences of headers marked as a Vec<(String,String)> and &[(String,String)] but they're in other directories, and I'm not quite sure moving the Header type to neqo-common would be a good idea.

Let me know if it would be worth it (that and having a Headers type as well) and I'll do it in a subsequent PR :)

image

agrover:
Cool. Does this let us remove the clippy(type_complexity) directive?

o0Ignition0o:
I think it does, I have fixed the conflicts, and I'll remove the directive as soon as I get home :)

martinthomson:
The only reason I can see to have a `HeaderList` or similar type is to enable additional functions.  The current regime of using `Vec<Header>` or `&[Header]` seems pretty workable.

PR #132, opened 2019-08-09
ConnectionEvents for the connection state changes.

ddragana:
lets make Connected, Closing and Close events. In current http3 code we need to check whether state has change every time we call process_*


agrover:
How about going a little further and adding a `StateChange(Connection::State)` event?

martinthomson:
I like Andy's suggestion.

ddragana:
`StateChange(Connection::State)` event is nicer.

PR #133, opened 2019-08-09
Rename request stream

ddragana:
- rename request_stream into transaction
- split RequestStreamClient state into send_state and recv_state, needed for making sending request data work (this will be my next PR)


ddragana:
it was late when I made this. I will fix it.

PR #134, opened 2019-08-09
Http3 nits

ddragana:
Http3 nits:

- move API functions to single place,

- make some functions easier to read,

-  rename read_data and get_headers into read_response_data, get_reasponse_headers (the names are more descriptive)


ddragana:
it was late when I made this. I will fix it.

PR #136, opened 2019-08-09
Http3 nits

ddragana:
Http3 nits:
- move API functions to single place,
- make some functions easier to read,
- rename read_data and get_headers into read_response_data, get_reasponse_headers (the names are more descriptive)

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -202,7 +202,7 @@ impl Handler for PostConnectHandler {
                         return false;
                     }
 
-                    let headers = client.get_headers(stream_id);
+                    let headers = client.get_response_headers(stream_id);
                     println!("READ HEADERS[{}]: {:?}", stream_id, headers);
                 }
                 Http3Event::DataReadable { stream_id } => {
@@ -212,7 +212,7 @@ impl Handler for PostConnectHandler {
                     }
 
                     let (sz, fin) = client
-                        .read_data(Instant::now(), stream_id, &mut data)
+                        .read_response_data(Instant::now(), stream_id, &mut data)
                         .expect("Read should succeed");
                     if args.omit_read_data {
                         println!("READ[{}]: {} bytes", stream_id, sz);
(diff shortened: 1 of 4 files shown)


neqo-client/src/main.rs
```diff
@@ -202,7 +202,7 @@ impl Handler for PostConnectHandler {
                         return false;
                     }
 
-                    let headers = client.get_headers(stream_id);
+                    let headers = client.get_response_headers(stream_id);
```
martinthomson: is this `read_response_headers()` or an accessor?

neqo-http3/src/connection.rs
```diff
@@ -187,6 +187,19 @@ impl ::std::fmt::Display for Http3Connection {
     }
 }
 
+macro_rules! transaction {
+    ($t:expr) => {
+        match $t {
+            Some(t) => t,
+            _ => {
+                return Err(Error::TransportError(
+                    neqo_transport::Error::InvalidStreamId,
```
martinthomson: If this code is generating the error, does it really make sense to use a TransportError?

neqo-client/src/main.rs
ddragana: it should be read here. But I could actually make it an accessor. I will do that in a separate issue.

neqo-http3/src/connection.rs
ddragana: having a transaction should match having a quic stream, therefore I use this error. I could make a new one.

neqo-client/src/main.rs
ddragana: I will do this in a separate PR, because I think I will refactor transaction_client/server) a bit more.

agrover:
@ddragana is this PR still active or has it been superseded?

ddragana:
I will close this PR. I needed to move forward with other that touch the same code, so this one now needs a lot of rebasing. I will just make a new one

PR #140, opened 2019-08-12
Move header type to neqo-qpack, reexport it through neqo-http3.

### LINKED ISSUE #137: Consider having a Headers type ?
As explained in https://github.com/mozilla/neqo/pull/131#issuecomment-520134820 It might be useful to:

Move the Header type to the common crate.
Have only one Headers type, which would be `Vec<Header>`.

Replace uses of `&[Header]` (or `&[(String,String)]`) to use the new Headers type.

Not sure using a Vec instead of a slice ref for fix sized arrays has a performance impact.

What's your opinion ? :)

o0Ignition0o:
fix #137

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -7,7 +7,7 @@
 #![deny(warnings)]
 use neqo_common::{matches, Datagram};
 use neqo_crypto::init;
-use neqo_http3::{Http3Connection, Http3Event, Http3State};
+use neqo_http3::{Header, Http3Connection, Http3Event, Http3State};
 use neqo_transport::Connection;
 use std::collections::HashSet;
 use std::io::{self, ErrorKind};
@@ -21,7 +21,7 @@ use url::Url;
 
 #[derive(Debug)]
 struct Headers {
-    pub h: Vec<(String, String)>,
+    pub h: Vec<Header>,
 }
 
 // dragana: this is a very stupid parser.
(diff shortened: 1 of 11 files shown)


agrover:
This suggests a `Headers` type is a `Vec<Header>` would also be handy.

neqo-client/src/main.rs
```diff
@@ -7,6 +7,7 @@
 #![deny(warnings)]
 use neqo_common::Datagram;
 use neqo_crypto::init_db;
+use neqo_http3::request_stream_server::Header;
```
agrover: re-export `Header` at top of http3 so apps can just do `use neqo_http3::Header`?

o0Ignition0o:
Oh I think I just enraged Chronos, the god of monotonic time ^^

o0Ignition0o:
Not really sure what's going on here :(

neqo-crypto/src/time.rs
```diff
@@ -180,7 +180,7 @@ mod test {
 
     #[test]
     fn convert_stable() {
-        let now = Time::from(Instant::now());
+        let now = Time::now();
```
agrover: this looks like an unrelated change that may be the source of the CI trouble, so just omit from this PR please.

o0Ignition0o:
I have actually just added this change to retrigger the CI.

I think there's a weird bug with timestamps which has been unfortunately
caught during my try.

I'll revert the commit, can you rerun the CI then ? ^^

o0Ignition0o:
There you go @agrover 
Btw should I open an issue referencing this build fail ? It would be nice to figure it out, given the number of Firefox users I'm pretty sure it will happen again someday.

agrover:
@o0Ignition0o yes please open an issue. FWIW I cannot reproduce locally.

As far as I'm concerned this PR can be merged once conflicts are resolved.

o0Ignition0o:
MB, forgot to update a test :(

PR #141, opened 2019-08-12
Split part of input() into separate functions.

### LINKED ISSUE #94: Reduce complexity of connection::input
see 3026668 and #82

o0Ignition0o:
Fixes #94

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -625,8 +625,8 @@ impl Connection {
 
     /// Process input and generate output.
     pub fn process(&mut self, dgram: Option<Datagram>, now: Instant) -> Output {
-        if let Some(dgram) = dgram {
-            self.process_input(dgram, now);
+        if let Some(d) = dgram {
+            self.process_input(d, now);
         }
         self.process_output(now)
     }
@@ -635,7 +635,6 @@ impl Connection {
         self.valid_cids.contains(cid) || self.paths.iter().any(|p| p.local_cids.contains(cid))
     }
 
-    #[allow(clippy::cognitive_complexity)]
     fn input(&mut self, d: Datagram, now: Instant) -> Res<()> {
         let mut slc = &d[..];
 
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection.rs
```diff
@@ -634,88 +633,106 @@ impl Connection {
...
+            .largest_acknowledged(PNSpace::from(hdr.epoch));
+        let res = match self.crypto.obtain_crypto_state(self.role, hdr.epoch) {
+            Ok(cs) => match cs.rx.as_ref() {
+                Some(rx) => {
+                    let pn_decoder = PacketNumberDecoder::new(largest_acknowledged);
+                    decrypt_packet(rx, pn_decoder, &mut hdr, slc)
                 }
-                self.set_state(State::Handshaking);
-            }
+                _ => Err(Error::KeysNotFound),
```
martinthomson: Can't you just use `None` here and below rather than be forced to map back to `Option` using `Result::ok`?

neqo-transport/src/connection.rs
```diff
@@ -634,88 +633,106 @@ impl Connection {
...
+        res.ok()
+    }
 
-            if !self.paths.iter().any(|p| p.received_on(&d)) {
-                // Right now, we don't support any form of migration.
-                // So generate an error if a packet is received on a new path.
-                return Err(Error::InvalidMigration);
-            }
+    /// Ok(true) if the packet is a duplicate
+    fn process_ack(&mut self, hdr: &PacketHdr, body: Vec<u8>, now: Instant) -> Res<bool> {
```
martinthomson: Based on the commentary, `is_duplicate()` might be a better name for this one, but then I see that we actually process the packet in here.  I think that you might want to find a new name for `process_input` and then rename this to `process_input`.

neqo-transport/src/connection.rs
```diff
@@ -634,88 +633,106 @@ impl Connection {
...
+            qdebug!([self] "Received duplicate packet epoch={} pn={}", hdr.epoch, hdr.pn);
+            self.stats.dups_rx += 1;
+            Ok(true)
+        } else {
+            self.acks[space].set_received(now, hdr.pn, ack_eliciting);
+            Ok(false)
         }
+    }
 
+    fn process_handshake(&mut self, hdr: PacketHdr, d: &Datagram) -> Res<()> {
```
martinthomson: The purpose of this function is to update connections IDs.  It also sets zero_rtt_state, which seems a little different.  Maybe name it `start_handshake()` or something like that.

o0Ignition0o:
Oh I'm sorry I get so many notifications I didn't notice you reviewed it :/

neqo-transport/src/connection.rs
o0Ignition0o: the input/output names seem pretty widely used, I wouldn't want to break it too much (unless it really makes sense).

neqo-transport/src/connection.rs
```diff
@@ -736,89 +735,110 @@ impl Connection {
...
-                        .expect("should have a path for sending Initial");
-                    p.remote_cid = hdr.scid.unwrap();
+    fn decrypt_body(&mut self, mut hdr: &mut PacketHdr, slc: &[u8]) -> Option<Vec<u8>> {
+        // Decryption failure, or not having keys is not fatal.
+        // If the state isn't available, or we can't decrypt the packet, drop
+        // the rest of the datagram on the floor, but don't generate an error.
+        let largest_acknowledged = self
+            .loss_recovery
+            .largest_acknowledged(PNSpace::from(hdr.epoch));
+        match self.crypto.obtain_crypto_state(self.role, hdr.epoch) {
```
martinthomson: Is this a case where we can use `if let (...) = ...`  or even `.map`?  In other words, will clippy::pedantic immediately object to this?

agrover: I guess we'll find out.

o0Ignition0o: Oh I'm sorry I havent seen your comment :( (my notification setup must be messed up)
We can (and should) use a map.
I tried to setup clippy::pedantic just for this function but it's complaining about a lot of other issues occuring inside it.
I hope I'll have more time to keep on refactoring towards clippy::pedantic soon!

PR #143, opened 2019-08-12
Rename request_stream_client/server into transaction_client/server

ddragana:
This is the same as PR #135. For some reason that one did not got merged.

### DIFF
--- a/neqo-http3/src/transaction_client.rs
+++ b/neqo-http3/src/transaction_client.rs
@@ -7,7 +7,7 @@
 use crate::hframe::{HFrame, HFrameReader, H3_FRAME_TYPE_DATA, H3_FRAME_TYPE_HEADERS};
 
 use crate::connection::Http3Events;
-use crate::request_stream_server::Header;
+use crate::transaction_server::Header;
 
 use neqo_common::{qdebug, qinfo, Encoder};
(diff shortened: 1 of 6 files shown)


neqo-http3/src/transaction_client.rs
```diff
@@ -88,22 +88,41 @@ impl Response {
...
+ *                     state we do no read from the stream.
+ *    WaitingForData : we got HEADERS, we are waiting for one or more data
+ *                     frames. In this state we can receive one or more
+ *                     PUSH_PROMIS frames or a HEADERS frame carrying trailers.
+ *    ReadingData : we got a DATA frame, now we letting the app read payload.
+ *                  From here we will go back to WaitingForData state to wait
+ *                  for more data frames or to CLosed state
+ *    ReadingTrailers : reading trailers.
+ *    ClosePending : waiting for app to pick up data, after that we can delete
+ * the TransactionClient.
```
martinthomson:
```suggestion
 *                   the TransactionClient.
```

ddragana:
> This looks like a good change, but why are there no new tests?

this is only renaming and splitting states. I am working on 2 more PR that are going to change behavior and add tests. This PR is only preparation for the next 2.

PR #146, opened 2019-08-13
Fixes for different build variants

martinthomson:
Building on Windows wasn't working again.  And I wasn't getting good
results from --release either.

I will open an issue to investigate enabling --release and windows
builds on CI.  Circle supports windows now, but that might take a little
while to get working.

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -1,3 +1,5 @@
+#![deny(warnings)]
+
 use bindgen::Builder;
 use serde_derive::Deserialize;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


neqo-crypto/build.rs
```diff
@@ -255,9 +267,12 @@ fn main() {
     if is_debug() {
         static_link(&nsstarget);
     } else {
-        println!("cargo:rustc-link-lib=nspr4");
-        println!("cargo:rustc-link-lib=nss3");
-        println!("cargo:rustc-link-lib=ssl3");
+        let libs = if env::consts::OS == "windows" {
+            &["nssutil3.dll", "nss3.dll", "ssl3.dll"]
```
martinthomson: The best way to put this is that NSS builds are *quirky*.

PR #148, opened 2019-08-13
Crypto error handling

### LINKED ISSUE #145: Potentially unsafe uses of unsafe
(I'm super excited to see QUIC for Firefox built in rust!)

I did a quick review of the uses of `unsafe` and a few potential issues jumped out at me.

https://github.com/mozilla/neqo/blob/5e5249a056ba397919135b83f43d3763a5c8e00d/neqo-crypto/src/aead.rs#L79-L86

If `p.len()` larger than can fit in `u32`, you'll wrap around and produce an incorrect result. I don't think this can happen in reality, but using something like `try_into()` to propagate an error in that case would be safer.

https://github.com/mozilla/neqo/blob/5e5249a056ba397919135b83f43d3763a5c8e00d/neqo-crypto/src/agent.rs#L680-L684

I believe this produces a use-after-free issue. `CString::new` returns a `Result<CString, SomeError>`. Calling `unwrap()` gives you a `CString`, and then `as_ptr()` gets you the raw `* const c_char`. However, because `as_ptr()` returns a raw pointer, it doesn't participate in the lifetime, and thus nothing is keeping the `CString` alive. This issue is described here: https://doc.rust-lang.org/stable/std/ffi/struct.CString.html#method.as_ptr

martinthomson:
Alex pointed out a few potential bugs in the code.  This fixes the integer overflow and the UAF, but in doing the latter I decided that it would be better to centralize the error handling into the experimental API macro.  That allows the wrapper to report errors with CString::new() rather than just calling panic!.

Fixes #145.


neqo-crypto/src/result.rs
```diff
@@ -34,16 +25,12 @@ where
     }
 }
 
-fn result_helper(rv: ssl::SECStatus, allow_blocked: bool) -> Res<bool> {
+pub fn result(rv: ssl::SECStatus) -> Res<()> {
```
agrover: Is there a way to make places where we're calling `result::result` a little more clear? Can it be named `res_from_status` or `res_from_rv` or something similar?

neqo-crypto/src/aead.rs
```diff
@@ -76,15 +76,14 @@ impl Aead {
...
         let p = prefix_str.as_bytes();
         let mut ctx: *mut ssl::SSLAeadContext = null_mut();
-        let rv = SSL_MakeAead(
+        SSL_MakeAead(
             version,
             cipher,
             secret,
             p.as_ptr() as *const i8,
-            p.len() as u32,
+            p.len().try_into().unwrap(),
```
agrover: prefer tryfrom? e.g.  `u32::try_from(p.len()).unwrap()`. More explicit about target type.

neqo-crypto/src/result.rs
agrover: or can an `impl TryFrom<ssl::SECStatus> for Res` be of help?

neqo-crypto/src/aead.rs
martinthomson: I think that I might prefer try_from where possible.  That doesn't seem to be idiomatic and I don't want to just hit this one instance when we do that.

neqo-crypto/src/result.rs
martinthomson: Given that `SECStatus` is always convertable, that would be `From`, wouldn't it?  Except that we don't define any of the involved types:
* `SECStatus` is a type alias for `i32`.
* `Res` is a type alias for `Result`.
* `From` is similarly in the standard library.

This seems like a great idea, but we can't realistically do that.  Renaming the function is the best we can do.  `res_from_secstatus` is the best I can come up with offhand.  I've opened #166.

martinthomson:
I've rebased this and done some more serious refactoring of the error handling pieces.  I've removed a couple of files in the process, which were created before I understood the whole conversion piece better.

PR #151, opened 2019-08-13
 Send fin if we close the send side of a remote bidi stream. 

ddragana:
I think this is not completely correct, but I need some help.
@agrover

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -2315,4 +2315,26 @@ mod tests {
             _ => false,
         });
     }
+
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection.rs
```diff
@@ -2315,4 +2315,26 @@ mod tests {
             _ => false,
         });
     }
+
+    #[test]
+    // Send fin even if a peer closes a reomte bidi send stream before sending any data.
```
martinthomson: typo in comment: reomte

PR #152, opened 2019-08-13
Fix hframe if no data has been send on a stream.

ddragana:


martinthomson:
@ddragana, did mergify do something wrong, or was this superseded?

ddragana:
mergify did something wrong here. 
Anyway I will update my other PR that fixes more things in hframe so that it covers this problem as well.

PR #153, opened 2019-08-13
Prepare to use clippy::pedantic

### LINKED ISSUE #150: Enable clippy::pedantic
There are a bunch of problems that might be revealed by clippy::pedantic.  While false positives are a risk, we can suppress individual warnings as they are analyzed.

o0Ignition0o:
Part of #150

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -53,7 +53,7 @@ impl HandshakeState {
 
 fn get_alpn(fd: *mut ssl::PRFileDesc, pre: bool) -> Res<Option<String>> {
     let mut alpn_state = ssl::SSLNextProtoState::SSL_NEXT_PROTO_NO_SUPPORT;
-    let mut chosen = vec![0u8; 255];
+    let mut chosen = vec![0_u8; 255];
     let mut chosen_len: c_uint = 0;
     let rv = unsafe {
         ssl::SSL_GetNextProto(
@@ -99,18 +99,18 @@ macro_rules! preinfo_arg {
 }
 
 impl SecretAgentPreInfo {
-    fn new(fd: *mut ssl::PRFileDesc) -> Res<SecretAgentPreInfo> {
+    fn new(fd: *mut ssl::PRFileDesc) -> Res<Self> {
         let mut info: ssl::SSLPreliminaryChannelInfo = unsafe { mem::uninitialized() };
         let rv = unsafe {
             ssl::SSL_GetPreliminaryChannelInfo(
                 fd,
                 &mut info,
-                mem::size_of::<ssl::SSLPreliminaryChannelInfo>() as ssl::PRUint32,
+                mem::size_of::<ssl::SSLPreliminaryChannelInfo>().try_into()?,
             )
         };
         result::result(rv)?;
 
-        Ok(SecretAgentPreInfo {
+        Ok(Self {
             info,
             alpn: get_alpn(fd, true)?,
         })
@@ -150,24 +150,24 @@ pub struct SecretAgentInfo {
 }
 
 impl SecretAgentInfo {
-    fn new(fd: *mut ssl::PRFileDesc) -> Res<SecretAgentInfo> {
+    fn new(fd: *mut ssl::PRFileDesc) -> Res<Self> {
         let mut info: ssl::SSLChannelInfo = unsafe { mem::uninitialized() };
         let rv = unsafe {
             ssl::SSL_GetChannelInfo(
                 fd,
                 &mut info,
-                mem::size_of::<ssl::SSLChannelInfo>() as ssl::PRUint32,
+                mem::size_of::<ssl::SSLChannelInfo>().try_into()?,
             )
         };
         result::result(rv)?;
-        Ok(SecretAgentInfo {
+        Ok(Self {
             version: info.protocolVersion as Version,
             cipher: info.cipherSuite as Cipher,
-            group: info.keaGroup as Group,
+            group: info.keaGroup.try_into()?,
             resumed: info.resumed != 0,
             early_data: info.earlyDataAccepted != 0,
             alpn: get_alpn(fd, false)?,
-            signature_scheme: info.signatureScheme as SignatureScheme,
+            signature_scheme: info.signatureScheme.try_into()?,
         })
     }
 
@@ -194,8 +194,9 @@ impl SecretAgentInfo {
     }
 }
 
-/// SecretAgent holds the common parts of client and server.
+/// `SecretAgent` holds the common parts of client and server.
 #[derive(Debug)]
+#[allow(clippy::module_name_repetitions)]
 pub struct SecretAgent {
     fd: *mut ssl::PRFileDesc,
     secrets: SecretHolder,
@@ -218,10 +219,10 @@ pub struct SecretAgent {
 }
 
 impl SecretAgent {
-    fn new() -> Res<SecretAgent> {
-        let mut agent = SecretAgent {
+    fn new() -> Res<Self> {
+        let mut agent = Self {
             fd: null_mut(),
-            secrets: Default::default(),
+            secrets: SecretHolder::default(),
             raw: None,
             io: Box::new(AgentIo::new()),
             state: HandshakeState::New,
@@ -230,8 +231,8 @@ impl SecretAgent {
             alert: Box::new(None),
             now: Box::new(0),
 
-            extension_handlers: Default::default(),
-            inf: Default::default(),
+            extension_handlers: Vec::new(),
+            inf: None,
 
             no_eoed: false,
         };
@@ -290,13 +291,10 @@ impl SecretAgent {
             // Fatal alerts demand attention.
             let p = arg as *mut Option<Alert>;
             let st = p.as_mut().unwrap();
-            match st {
-                None => {
-                    *st = Some(alert.description);
-                }
-                _ => {
-                    qwarn!([format!("{:p}", fd)] "duplicate alert {}", alert.description);
-                }
+            if st.is_none() {
+                *st = Some(alert.description);
+            } else {
+                qwarn!([format!("{:p}", fd)] "duplicate alert {}", alert.description);
             }
         }
     }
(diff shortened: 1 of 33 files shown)


neqo-crypto/src/agent.rs
```diff
@@ -344,9 +342,9 @@ impl SecretAgent {
     /// Default configuration.
     fn configure(&mut self) -> Res<()> {
         self.set_version_range(TLS_VERSION_1_3, TLS_VERSION_1_3)?;
-        self.set_option(ssl::Opt::Locking, false)?;
-        self.set_option(ssl::Opt::Tickets, false)?;
-        self.set_option(ssl::Opt::OcspStapling, true)?;
+        self.set_option(&ssl::Opt::Locking, false)?;
```
martinthomson: We can probably use `#[derive(Copy)]` on that enum to avoid this change.

neqo-crypto/src/agent.rs
```diff
@@ -309,29 +307,29 @@ impl SecretAgent {
 
     // Ready this for connecting.
     fn ready(&mut self, is_server: bool) -> Res<()> {
-        let rv = unsafe {
+        let mut rv = unsafe {
```
martinthomson: This will rot with #148.

neqo-crypto/src/agent.rs
```diff
@@ -768,11 +766,8 @@ struct ZeroRttCheckState {
 }
 
 impl ZeroRttCheckState {
-    pub fn new(
-        fd: *mut ssl::PRFileDesc,
-        checker: Box<dyn ZeroRttChecker>,
-    ) -> Box<ZeroRttCheckState> {
-        Box::new(ZeroRttCheckState { fd, checker })
+    pub fn new(fd: *mut ssl::PRFileDesc, checker: Box<dyn ZeroRttChecker>) -> Box<Self> {
```
martinthomson: I seem to remember clippy getting snippy about not returning `Self` from functions called `new`.

neqo-crypto/src/agentio.rs
```diff
@@ -321,8 +321,10 @@ unsafe extern "C" fn agent_available64(mut fd: PrFd) -> prio::PRInt64 {
     io.input.available as prio::PRInt64
 }
 
+#[allow(clippy::cast_possible_truncation)]
 unsafe extern "C" fn agent_getname(_fd: PrFd, addr: *mut prio::PRNetAddr) -> PrStatus {
     let a = addr.as_mut().unwrap();
+    // Cast is safe because prio::PR_AF_INET is 2
     a.inet.family = prio::PR_AF_INET as prio::PRUint16;
```
martinthomson: Can the annotation be moved to the specific line?

neqo-crypto/src/cert.rs
```diff
@@ -38,6 +38,9 @@ fn peer_certificate_chain(fd: *mut PRFileDesc) -> Option<(CertList, *const CERTC
     Some((certs, cursor))
 }
 
+// As explained in rfc6961, an OCSPResponseList can have at most
+// 2^24 items. Casting its length is therefore safe even on 32 bits targets.
+#[allow(clippy::cast_possible_wrap)]
```
martinthomson: This can probably use use try_into, even if this is guaranteed to be safe.

neqo-crypto/src/ext.rs
```diff
@@ -99,7 +102,8 @@ impl ExtensionTracker {
         arg: *mut c_void,
     ) -> SECStatus {
         let d = std::slice::from_raw_parts(data, len as usize);
```
martinthomson: Are you sure that this 'as' isn't being suppressed here also?

neqo-crypto/src/ext.rs
```diff
@@ -99,7 +102,8 @@ impl ExtensionTracker {
...
-        ExtensionTracker::wrap_handler_call(arg, |handler| {
+        Self::wrap_handler_call(arg, |handler| {
+            // Cast is safe here because the message type is always part of the enum
             match handler.handle(message as HandshakeMessage, d) {
```
martinthomson: Hmm, this is one of those cases where the C bindings produce a u32 where a u8 would suffice.  I wonder if we can't fix the bindings somehow.

neqo-crypto/src/agent.rs
o0Ignition0o: Hmm It didnt trigger _yet_, maybe it's in clippy::Restriction ?

neqo-crypto/src/agentio.rs
o0Ignition0o: I think annotations inside blocks are experimental for now.

Edit: They are https://github.com/rust-lang/rust/issues/15701

neqo-crypto/src/ext.rs
o0Ignition0o: I'll read some bindgen and co documentation and try to figure out, that would be much cleaner indeed !

neqo-transport/src/send_stream.rs
```diff
@@ -325,10 +334,12 @@ impl TxBuffer {
...
         // We can drop contig acked range from the buffer
         let new_retirable = self.ranges.acked_from_zero() - self.retired;
-        if new_retirable > 0 {
-            let keep_len = self.buffered() - new_retirable as usize;
+        if let Ok(nr) = usize::try_from(new_retirable) {
+            let keep_len: usize = self.buffered() - nr;
             self.send_buf.truncate_front(keep_len);
             self.retired += new_retirable;
+        } else {
+            // FIXME: What do we want to do here ?
```
o0Ignition0o: Raising awareness wrt this path. It's never supposed to happen but well...

neqo-transport/src/send_stream.rs
```diff
@@ -705,9 +716,13 @@ impl SendStreams {
...
     }
 
-    pub fn acked(&mut self, token: StreamRecoveryToken) {
+    pub fn acked(&mut self, token: &StreamRecoveryToken) {
         if let Some(ss) = self.0.get_mut(&token.id) {
-            ss.mark_as_acked(token.offset, token.length as usize, token.fin);
+            if let Ok(len) = token.length.try_into() {
+                ss.mark_as_acked(token.offset, len, token.fin);
+            } else {
+                //FIXME: Do we want to log something?
```
o0Ignition0o: Raising awareness wrt this path. It's never supposed to happen but well...

neqo-transport/src/send_stream.rs
```diff
@@ -717,9 +732,13 @@ impl SendStreams {
...
     }
 
-    pub fn lost(&mut self, token: StreamRecoveryToken) {
+    pub fn lost(&mut self, token: &StreamRecoveryToken) {
         if let Some(ss) = self.0.get_mut(&token.id) {
-            ss.mark_as_lost(token.offset, token.length as usize, token.fin);
+            if let Ok(len) = token.length.try_into() {
+                ss.mark_as_lost(token.offset, len, token.fin);
+            } else {
+                //FIXME: Do we want to log something?
```
o0Ignition0o: Raising awareness wrt this path. It's never supposed to happen but well...

neqo-transport/src/send_stream.rs
```diff
@@ -302,15 +301,25 @@ impl TxBuffer {
...
-
-        if start == self.retired + self.buffered() as u64 {
-            return None;
-        }
-
-        let buff_off = (start - self.retired) as usize;
-        match maybe_len {
-            Some(len) => Some((start, &self.send_buf[buff_off..buff_off + len as usize])),
-            None => Some((start, &self.send_buf[buff_off..])),
+        //TODO ME: ITS UGLY LETS REFACTOR IT
```
o0Ignition0o: Will try to figure out a less ugly way to do it.

neqo-transport/src/frame.rs
```diff
@@ -495,18 +495,18 @@ pub fn decode_frame(dec: &mut Decoder) -> Res<Frame> {
         }
         FRAME_TYPE_STREAM...FRAME_TYPE_STREAM_MAX => {
             let s = dv!(dec);
-            let o = if t & STREAM_FRAME_BIT_OFF != 0 {
-                dv!(dec)
-            } else {
+            let o = if t & STREAM_FRAME_BIT_OFF == 0 {
```
martinthomson: This one is a little annoying.  I think of non-zero testing the other way around.  Oh well.

neqo-transport/src/send_stream.rs
```diff
@@ -302,15 +301,25 @@ impl TxBuffer {
...
+        match (usize::try_from(start), usize::try_from(self.retired)) {
+            (Ok(s), Ok(retired)) => {
+                if s == retired + self.buffered() {
+                    return None;
+                }
+                let buff_off = s - retired;
```
```suggestion
                debug_assert!(s > retired);
                let buff_off = s - retired;
```

neqo-transport/src/send_stream.rs
martinthomson: Don't log anything, just use `unwrap()` and we can panic instead.  The number of layers this has to pass with both u64 and usize variables carrying the value is insane.  If it doesn't fit in both, we're in a very bad place.

The alternative is to change token.length to a usize.  That ends up requiring far fewer conversions overall.

neqo-transport/src/tracking.rs
```diff
@@ -329,26 +330,29 @@ impl AckTracker {
...
+            };
+            Some((
+                ack,
+                Some(RecoveryToken::Ack(AckToken {
+                    space: PNSpace::from(epoch),
+                    ranges,
+                })),
+            ))
+        } else {
+            None
```
martinthomson: Might be worth putting a qwarn!() here.

neqo-crypto/src/agentio.rs
```diff
@@ -291,8 +300,12 @@ unsafe extern "C" fn agent_write(
     if amount <= 0 {
         return PR_FAILURE;
     }
-    io.save_output(buf as *const u8, amount as usize);
-    amount as prio::PRInt32
+    if let Ok(a) = amount.try_into() {
+        io.save_output(buf as *const u8, a);
+        amount as prio::PRInt32
```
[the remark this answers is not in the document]
o0Ignition0o: Yeah I think I got lost in the depth of changes I was doing ^^

neqo-crypto/src/agentio.rs
```diff
@@ -307,22 +320,28 @@ unsafe extern "C" fn agent_send(
...
 
 unsafe extern "C" fn agent_available(mut fd: PrFd) -> prio::PRInt32 {
     let io = AgentIo::borrow(&mut fd);
-    io.input.available as prio::PRInt32
+    io.input.available.try_into().unwrap_or(PR_FAILURE)
 }
 
 unsafe extern "C" fn agent_available64(mut fd: PrFd) -> prio::PRInt64 {
     let io = AgentIo::borrow(&mut fd);
     io.input.available as prio::PRInt64
```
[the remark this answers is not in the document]
o0Ignition0o: Oh I missed that one, good catch !

neqo-transport/src/frame.rs
o0Ignition0o: I can disable it if you would like, since it is more explicit for our use case.

martinthomson: Nah.  Better to minimize the warnings, even if they sometimes go against preconceptions.  Warnings reduce variation in coding styles, which is better than pandering to those preconceptions.

o0Ignition0o: :ok_hand:

neqo-crypto/src/ext.rs
o0Ignition0o: It is not, because len is a c_uint (u32) so it's always lossless (even on 32 bits platform).
However since the wrap_handler_call is an other block, I have been able to move the allow(clippy) statement to wrap it only, which is always better \o/

neqo-crypto/src/ext.rs
o0Ignition0o: opened #168

o0Ignition0o:
I think I've adressed almost all the comments, is there anything I'm missing ?
Once this one is merged I'll dig a bit more into bindgen and figure out how to have it produce the relevant types :)

PR #157, opened 2019-08-14
Fix hframe.

### LINKED ISSUE #169: Issues with HFrameReader
There are a coupe of issues with HFrameReader:

- if a sender closes a stream before sending anything the HFrameReader will be in state GetType instead of BeforeFrame and neqo will fail reading frame with MalformatedFrame error.

- This code in fn HFrameReader::receive() is not correct:

```
            let mut input = match conn.stream_recv(stream_id, &mut buf[..]) {
                Ok((_, true)) => {
```

If stream_recv returns some data and fin we will exit the function before actually consuming the data. Therefore the HFrameREader will be in wrong state and we will report MalformatedFrame.

- If the length of a frame is 0 for some frames we change to state HFrameReaderState::GetData, but there is not more data to receive.

- Also I will move breaking out of the loop to the end of the loop so that it is consistent and at one place.

- Currently we return `break Err(Error::NoMoreData)`, if we are in state that is not the one already handled. I think we can never be in any other state than the states we already handle. I will put a panic instead of returning error.

ddragana:
Fixes #169.

### DIFF
--- a/neqo-http3/src/hframe.rs
+++ b/neqo-http3/src/hframe.rs
@@ -283,7 +283,7 @@ impl HFrameReader {
             let fin;
             let mut input = match conn.stream_recv(stream_id, &mut buf[..]) {
                 Ok((0, true)) => {
-                    qtrace!([conn] "HFrameReader::receive: stream has been closed.");
+                    qtrace!([conn] "HFrameReader::receive: stream has been closed");
                     break match self.state {
                         HFrameReaderState::BeforeFrame => Ok(true),
                         _ => Err(Error::MalformedFrame(0xff)),
@@ -291,7 +291,7 @@ impl HFrameReader {
                 }
                 Ok((0, false)) => break Ok(false),
                 Ok((amount, f)) => {
-                    qtrace!([conn] "HFrameReader::receive: reading {} byte, fin={}.", amount, f);
+                    qtrace!([conn] "HFrameReader::receive: reading {} byte, fin={}", amount, f);
                     fin = f;
                     Decoder::from(&buf[..amount])
                 }
@@ -316,7 +316,7 @@ impl HFrameReader {
                     IncrementalDecoderResult::InProgress => {
                         self.state = HFrameReaderState::GetType;
                     }
-                    _ => panic!("We most be in one of the states above."),
+                    _ => panic!("We must be in one of the states above"),
                 },
 
                 HFrameReaderState::GetLength => {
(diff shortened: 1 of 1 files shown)


neqo-http3/src/hframe.rs
```diff
@@ -908,4 +935,213 @@ mod tests {
             panic!("wrong frame type");
         }
     }
+
+    fn test_reading_frame(
+        buf: &[u8],
+        send_fin_with_data: bool,
+        send_fin_after_data: bool,
+        expect_fin: bool,
+        succeed: bool,
```
martinthomson:
```suggestion
        expect_success: bool,
```

neqo-http3/src/hframe.rs
```diff
@@ -908,4 +935,213 @@ mod tests {
...
+        frame_done: bool,
+    ) {
+        let (mut conn_c, mut conn_s) = connect();
+
+        // create a stream
+        let stream_id = conn_s.stream_create(StreamType::BiDi).unwrap();
+
+        let mut fr: HFrameReader = HFrameReader::new();
```
martinthomson:
```suggestion
        let mut fr = HFrameReader::new();
```

Also, you can move this down closer to where it is needed.

neqo-http3/src/hframe.rs
```diff
@@ -908,4 +935,213 @@ mod tests {
...
+        conn_c.process(out.dgram(), now());
+
+        if send_fin_after_data {
+            conn_s.stream_close_send(stream_id).unwrap();
+            let out = conn_s.process(None, now());
+            conn_c.process(out.dgram(), now());
+        }
+
+        if succeed {
+            assert_eq!(Ok(expect_fin), fr.receive(&mut conn_c, stream_id));
```
martinthomson: Move `fr.receive` out of the if statement.

neqo-http3/src/hframe.rs
```diff
@@ -908,4 +935,213 @@ mod tests {
             panic!("wrong frame type");
         }
     }
+
+    fn test_reading_frame(
```
martinthomson: This looks really good from a distance, but the call sites are very hard to read.  I think that you want to turn that list of 5 booleans into an options struct (or an enum for the sending options) so that the call sites can be clearer:

```rust
enum FrameReadingTestSend {
  DataOnly,
  DataWithFin,
  DataThenFin,
}

enum FrameReadingTestExpect {
  Error,
  Incomplete,
  FrameComplete,
  FrameAndStreamComplete,
}

// Then test_reading_frame(&buf[..i], false, false, false, true, false) becomes
             test_reading_frame(&buf[..i], FrameReadingTestSend::OnlyData, FrameReadingTestExpect::Incomplete); 
```

ddragana:
> Love the added tests, but one thing I'd like to make sure is recorded somewhere is the bug that this PR is fixing. You could either open an issue and then mark this PR as fixing it, or just amend/repush the "fix hframe" commit with an explanation in the commitlog.

I have open issue #169

ddragana:
The ci error will be fix by #170

PR #158, opened 2019-08-14
Refactor the receiv part of transaction_client

ddragana:


neqo-http3/src/connection.rs
```diff
@@ -192,6 +192,15 @@ impl ::std::fmt::Display for Http3Connection {
     }
 }
 
+macro_rules! transaction {
+    ($t:expr) => {
+        match $t {
+            Some(t) => t,
+            _ => return Err(Error::InvalidStreamId),
```
agrover: don't think you need this macro. Instead add `.ok_or_else(|| Error::InvalidStreamId)?`

martinthomson: Or `.ok_or(Error::InvalidStreamId)`.

ddragana:
I will make a new commit with more info in the commit message.

PR #167, opened 2019-08-19
Upgrade rustup

martinthomson:
Once this merges, we should push to the `docker` branch.  But we shouldn't do that until we can verify that the image builds.


martinthomson:
Pushed to docker.  I'll report if there are any issues.

martinthomson:
https://cloud.docker.com/repository/registry-1.docker.io/martinthomson/neqo/builds/54f56272-a6d6-4afe-b22d-f865d01729f2

PR #181, opened 2019-08-26
Refactor receiving part of the transaction_client:

ddragana:
- change get_headers/read_data names into something more descriptive,
- make read_response_headers return fin flag as well,
- make read_response_headers and read_response_data a bit nicer,
- add new error InvalidStreamId so that we do not use a transport error,
- add ResponseHeadersState to be easier to decide if we need to add HeadersREady or Data Ready event, also we do not need struct Response,
- the TransactionClient::receive code was not handling fin bit correctly in all cases, I moved setting state to TransactionRecvState::ClosePending in functions actually handling new data and new frames
- make separate functions for some parts of TransactionClient::receive,
- add test for cases wher fin was handled wrong.


ddragana:
Sorry, this is the same PR as before, I have merged it into a branch instead of master.

PR #192, opened 2019-09-02
Refactor request sending part and reset handling:

ddragana:
- make sending a request body possible,
 - add events DataWritable and StopSending,
 - StopSending will be only posted for the EarlyResponse error code,
 - fix reset/stop_sending handling (reset will reset sending side as well, stop_sending with EarlyResponse will only close sending side, stop_sending with another error will close both sides of a stream)

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1863,8 +1863,9 @@ mod tests {
                         .read_response_data(now(), stream_id, &mut buf)
                         .unwrap();
                     assert_eq!(fin, true);
-                    assert_eq!(amount, 3);
-                    assert_eq!(buf[..3], [0x61, 0x62, 0x63]);
+                    const EXPECTED_RESPONSE_BODY: &[u8] = &[0x61, 0x62, 0x63];
+                    assert_eq!(amount, EXPECTED_RESPONSE_BODY.len());
+                    assert_eq!(&buf[..3], EXPECTED_RESPONSE_BODY);
                 }
                 _ => {}
             }
@@ -1918,14 +1919,13 @@ mod tests {
                     let mut buf = [0u8; 100];
                     let (amount, fin) = neqo_trans_conn.stream_recv(stream_id, &mut buf).unwrap();
                     assert_eq!(fin, true);
-                    assert_eq!(amount, 23);
-                    assert_eq!(
-                        buf[..23],
-                        [
-                            0x01, 0x10, 0x00, 0x00, 0xd1, 0xd7, 0x50, 0x89, 0x41, 0xe9, 0x2a, 0x67,
-                            0x35, 0x53, 0x2e, 0x43, 0xd3, 0xc1, 0x0, 0x3, 0x64, 0x65, 0x66,
-                        ]
-                    );
+                    const EXPECTED_RESPONSE_BODY: &[u8] = &[
+                        0x01, 0x10, 0x00, 0x00, 0xd1, 0xd7, 0x50, 0x89, 0x41, 0xe9, 0x2a, 0x67,
+                        0x35, 0x53, 0x2e, 0x43, 0xd3, 0xc1, 0x0, 0x3, 0x64, 0x65, 0x66,
+                    ];
+                    assert_eq!(amount, EXPECTED_RESPONSE_BODY.len());
+                    assert_eq!(&buf[..23], EXPECTED_RESPONSE_BODY);
+
                     // send response - 200  Content-Length: 3
                     // with content: 'abc'.
                     let _ = neqo_trans_conn.stream_send(
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -363,8 +363,8 @@ impl Http3Connection {
         if self.role() == Role::Client {
             for stream_id in to_send {
                 if let Some(cs) = &mut self.transactions_client.get_mut(&stream_id) {
-                    cs.send(&mut self.conn, &mut self.qpack_encoder)?;
-                    if cs.has_data_to_send() {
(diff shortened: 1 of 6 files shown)


neqo-http3/src/connection.rs
```diff
@@ -1040,6 +1125,28 @@ impl Http3Events {
...
+                | Http3Event::NewPushStream { stream_id }
+                | Http3Event::Reset { stream_id, .. }
+                | Http3Event::StopSending { stream_id, .. } => *stream_id != remove_stream_id,
+                Http3Event::AuthenticationNeeded
+                | Http3Event::GoawayReceived
+                | Http3Event::StateChange { .. }
+                | Http3Event::RequestsCreatable => true,
+            })
+            .cloned()
+            .collect::<BTreeSet<_>>();
```
martinthomson: This seems like a lot of work.  If we use a HashSet, we could use `.retain()`.  BTrees don't respond well to that sort of treatment.

neqo-http3/src/connection.rs
```diff
@@ -1722,12 +1830,896 @@ mod tests {
...
+                }
+                Http3Event::DataReadable { stream_id } => {
+                    assert_eq!(stream_id, request_stream_id);
+                    let mut buf = [0u8; 100];
+                    let (amount, fin) = hconn
+                        .read_response_data(now(), stream_id, &mut buf)
+                        .unwrap();
+                    assert_eq!(fin, true);
+                    assert_eq!(amount, 3);
+                    assert_eq!(buf[..3], [0x61, 0x62, 0x63]);
```
martinthomson: You should put the body in a constant so that you can replace all these constants with references.

neqo-http3/src/connection.rs
```diff
@@ -1722,12 +1830,896 @@ mod tests {
...
+                    assert_eq!(stream_id, request_stream_id);
+                    let mut buf = [0u8; 100];
+                    let (amount, fin) = neqo_trans_conn.stream_recv(stream_id, &mut buf).unwrap();
+                    assert_eq!(fin, true);
+                    assert_eq!(amount, 23);
+                    assert_eq!(
+                        buf[..23],
+                        [
+                            0x01, 0x10, 0x00, 0x00, 0xd1, 0xd7, 0x50, 0x89, 0x41, 0xe9, 0x2a, 0x67,
+                            0x35, 0x53, 0x2e, 0x43, 0xd3, 0xc1, 0x0, 0x3, 0x64, 0x65, 0x66,
```
martinthomson: See previous comment.

neqo-http3/src/transaction_client.rs
```diff
@@ -14,9 +14,17 @@ use neqo_qpack::encoder::QPackEncoder;
...
+use std::cmp::min;
 use std::mem;
 
-#[derive(Debug)]
+const MAX_DATA_HEADER_SIZE_2: usize = 63; // Maximal amount of data with DATA frame header size 2
+const MAX_DATA_HEADER_SIZE_2_LIMIT: usize = 66; // 63 + 3 (size of the next biffer data frame header)
+const MAX_DATA_HEADER_SIZE_3: usize = 16383; // Maximal amount of data with DATA frame header size 3
+const MAX_DATA_HEADER_SIZE_3_LIMIT: usize = 16388; // 16383 + 5 (size of the next biffer data frame header)
+const MAX_DATA_HEADER_SIZE_5: usize = 1_073_741_823; // Maximal amount of data with DATA frame header size 3
+const MAX_DATA_HEADER_SIZE_5_LIMIT: usize = 1_073_741_832; // 1073741823 + 9 (size of the next biffer data frame header)
```
martinthomson:
```suggestion
const MAX_DATA_HEADER_SIZE_5_LIMIT: usize = 1_073_741_832; // 1073741823 + 9 (size of the next buffer data frame header)
```

neqo-http3/src/transaction_client.rs
```diff
@@ -14,9 +14,17 @@ use neqo_qpack::encoder::QPackEncoder;
 use neqo_transport::Connection;
 
 use crate::{Error, Res};
+use std::cmp::min;
 use std::mem;
 
-#[derive(Debug)]
+const MAX_DATA_HEADER_SIZE_2: usize = 63; // Maximal amount of data with DATA frame header size 2
+const MAX_DATA_HEADER_SIZE_2_LIMIT: usize = 66; // 63 + 3 (size of the next biffer data frame header)
```
martinthomson:
```suggestion
const MAX_DATA_HEADER_SIZE_2_LIMIT: usize = 66; // 63 + 3 (size of the next buffer data frame header)
```

But if this is really a sum, then you want to do this:
```suggestion
const MAX_DATA_HEADER_SIZE_2_LIMIT: usize = MAX_DATA_HEADER_SIZE_2 + 3; // 63 + 3 (size of the next buffer data frame header)
```

But then I wonder if this is right.

neqo-http3/src/transaction_client.rs
```diff
@@ -14,9 +14,17 @@ use neqo_qpack::encoder::QPackEncoder;
 use neqo_transport::Connection;
 
 use crate::{Error, Res};
+use std::cmp::min;
 use std::mem;
 
-#[derive(Debug)]
+const MAX_DATA_HEADER_SIZE_2: usize = 63; // Maximal amount of data with DATA frame header size 2
```
martinthomson: I think that you are looking for (1<<6) -1

neqo-http3/src/connection.rs
ddragana: Currently we rely on events being in order, actually only HeaderReady and DataReadable. This could be refactored, but I would suggest to open a new issue.

agrover: opened #195

PR #205, opened 2019-09-23
Read data frames as well after a stream has been unblocked.

### LINKED ISSUE #204: Read data frames as well after a stream has been unblocked
Currently we only read header frame after a stream has been unblock and data frame will be only read when a new packet arrives.  This is not ideal because a new packet may not arrive. We should start reading data frame after the stream has been unblocked.

ddragana:
Fixes #204


neqo-http3/src/connection.rs
```diff
@@ -1304,7 +1298,7 @@ mod tests {
 
         // assert no error occured.
         assert_eq!(hconn.state(), Http3State::Connected);
-        (hconn, neqo_trans_conn, control_stream)
+        (hconn, neqo_trans_conn, control_stream, encoder)
```
martinthomson: This function is now a little hard to read.  Could we perhaps make a test peer connection object that holds the neqo-transport::Connection, control stream ID, and QPackEncoder?

neqo-http3/src/connection.rs
```diff
@@ -3483,4 +3477,108 @@ mod tests {
...
+        // Send encoder instructions to unblock the stream.
+        encoder.send(&mut neqo_trans_conn).unwrap();
+
+        let out = neqo_trans_conn.process(None, now());
+        let _out = hconn.process(out.dgram(), now());
+        let _out = hconn.process(None, now());
+
+        let mut recv_header = false;
+        let mut recv_data = false;
+        // Now the stream is unblocked and both headers abnd data will be consumed.
```
```suggestion
        // Now the stream is unblocked and both headers and data will be consumed.
```

neqo-http3/src/connection.rs
```diff
@@ -3483,4 +3477,108 @@ mod tests {
...
+            match e {
+                ConnectionEvent::NewStream {
+                    stream_id,
+                    stream_type,
+                } => {
+                    assert_eq!(stream_id, request_stream_id);
+                    assert_eq!(stream_type, StreamType::BiDi);
+                }
+                ConnectionEvent::RecvStreamReadable { stream_id } => {
+                    assert_eq!(stream_id, request_stream_id);
```
martinthomson: You can do this entire piece outside of the event loop.  Only the assertion needs to be here.

I say that because this whole blob could benefit from being extracted into a function that is called `send_response()` or similar.

ddragana: I will merge this patch and fix the test in issue #211

PR #206, opened 2019-09-23
Fix error: enum variants on type aliases are experimental.

ddragana:
mozilla-central is complaining about this.


martinthomson:
What version of the compiler are you using?  I'm on 1.37, which doesn't complain about these.

ddragana:
> What version of the compiler are you using? I'm on 1.37, which doesn't complain about these.

For me, It is not happening locally neither. It is try that is complaining:
https://treeherder.mozilla.org/logviewer.html#/jobs?job_id=267920510&repo=try&lineNumber=19145

ddragana:
from log:
[task 2019-09-23T08:59:41.060Z] 08:59:41     INFO -  checking rustc version... 1.37.0-nightly
[task 2019-09-23T08:59:41.180Z] 08:59:41     INFO -  checking cargo version... 1.37.0

martinthomson:
Yeah, that -nightly suffix is a little bit of a concern.

How about we merge this and then look into reverting it later once gecko sorts this out.

ddragana:
> Yeah, that -nightly suffix is a little bit of a concern.
> 
> How about we merge this and then look into reverting it later once gecko sorts this out.

yes, let's merge it.

PR #214, opened 2019-09-25
Clean up http3 tests.

ddragana:
This change also use connect_and_send_request more often.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1138,16 +1138,23 @@ mod tests {
         };
     }
 
+    #[derive(PartialEq)]
+    enum Role {
+        Client,
+        Server,
+    }
+    use crate::connection::tests::Role::{Client, Server};
+
     // Start a client/server and check setting frame.
     #[allow(clippy::cognitive_complexity)]
-    fn connect(client: bool) -> (Http3Connection, Connection) {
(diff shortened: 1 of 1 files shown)


neqo-http3/src/connection.rs
```diff
@@ -1256,9 +1258,13 @@ mod tests {
...
-    fn connect_and_receive_control_stream(
-        client: bool,
-    ) -> (Http3Connection, Connection, u64, QPackEncoder) {
+    struct PeerConnection {
+        conn: Connection,
+        control_stream_id: u64,
+        encoder: QPackEncoder,
+    }
+
+    fn connect_and_receive_control_stream(client: bool) -> (Http3Connection, PeerConnection) {
```
martinthomson: I wonder if this argument could be changed to an enum so that call sites can see `connect_and_receive_control_stream(Client)`.  Also, this could probably just be renamed to `connect()`.

neqo-http3/src/connection.rs
```diff
@@ -1133,16 +1133,23 @@ mod tests {
         };
     }
 
+    #[derive(PartialEq)]
+    enum Role {
+        Client,
+        Server,
```
agrover: Can we use neqo_transport::Role here? It is pretty simple and might also serve here.

ddragana: removed.

PR #221, opened 2019-09-27
Fix gecko build

ddragana:
So we need this changes to make firefox compile.

The current version does not use  extra-bindgen-flags, but I found out that such a file is the newest way and encouraged approach we are doing this.

Also we need mozilla-config.h.

@martinthomson : should we add this change to this repository or strip all gecko build parts and add them in m.-c. by applying a patch on each update?

Probably I will need to update this patch once again when I figure out how to make a http3 test server be built by gecko. Note: the server builds with this patch on all platforms except windows and asan builds are broken as well.

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -223,7 +223,7 @@ fn get_includes(nsstarget: &Path, nssdist: &Path) -> Vec<PathBuf> {
     includes
 }
 
-fn build_bindings(base: &str, bindings: &Bindings, includes: &[PathBuf]) {
+fn build_bindings(base: &str, bindings: &Bindings, flags: &[String], gecko: bool) {
     let suffix = if bindings.cplusplus { ".hpp" } else { ".h" };
     let header_path = PathBuf::from(BINDINGS_DIR).join(String::from(base) + suffix);
     let header = header_path.to_str().unwrap();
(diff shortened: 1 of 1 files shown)


neqo-crypto/build.rs
```diff
@@ -305,58 +306,103 @@ fn setup_standalone() -> Vec<PathBuf> {
...
+    let libs = match env::var("CARGO_CFG_TARGET_OS").as_ref().map(|x| x.as_str()) {
+        Ok("android") | Ok("macos") => vec!["nss3"],
+        _ => vec!["nssutil3", "nss3", "ssl3", "plds4", "plc4", "nspr4"],
+    };
+
+    for lib in libs.iter() {
+        println!("cargo:rustc-link-lib=dylib={}", lib);
+    }
 
     match env::var_os("MOZ_TOPOBJDIR").map(PathBuf::from) {
```
martinthomson: Use `if let Some(path) = ...` for this.

neqo-crypto/build.rs
```diff
@@ -305,58 +306,103 @@ fn setup_standalone() -> Vec<PathBuf> {
...
+            println!(
+                "cargo:rustc-link-search=native={}",
+                path.join("dist").join("bin").to_str().unwrap()
+            );
+            println!(
+                "cargo:rustc-link-search=native={}",
+                path.join("security")
+                    .join("nss")
+                    .join("lib")
+                    .join("nss")
```
martinthomson: Can you factor out the common elements of this path builder?

neqo-crypto/build.rs
```diff
@@ -305,58 +306,103 @@ fn setup_standalone() -> Vec<PathBuf> {
...
 }
 
 fn main() {
-    setup_clang();
-
-    let includes = if cfg!(feature = "gecko") {
+    let flags = if cfg!(feature = "gecko") {
         setup_for_gecko()
     } else {
+        setup_clang();
```
martinthomson: Move the setup_clang() call to the standalone setup.

PR #234, opened 2019-10-07
Http3 rework part1

### LINKED ISSUE #20: Server class

ddragana:
The first change moves some structures form connection.rs int their own files:
- ControlStreamLocal
- ControlStreamRemote
- NewStreamTypeReader
- Http3Event (this are now called Http3ClientEvent because they are only for the client side.)

The second change makes Http3Connection base class that deals with common behaviors. There are Http3ClientHandler and Http3ServerHandler that handles pieces that are client/server specific. Http3Client/Http3Server are client/server APIs. This change does not change server and client API or any behavior of the client nor server.
There will be a follow up patch that changes Server behavior. There will be similar event as for client. That will make possible, for example, to cancel stream in  any stage, send early-response, etc. This will make testing easier. Also there should be a goaway function, etc.
This change adds Http3ServerEvent that will be use when the interface has been extended and it is only added to this patch to make more common code.
This will make possible to resolve #20.

@agrover I would like your feedback before I continue, thanks.


neqo-http3/src/control_stream_local.rs
```diff
@@ -0,0 +1,57 @@
...
+                    self.buf = b;
+                }
+            }
+            return Ok(());
+        }
+        Ok(())
+    }
+
+    pub fn create_control_stream(&mut self, conn: &mut Connection) -> Res<()> {
+        qdebug!([self] "create_control_stream.");
```
agrover: qtrace?

neqo-http3/src/control_stream_local.rs
```diff
@@ -0,0 +1,57 @@
...
+                    let b = self.buf.split_off(sent);
+                    self.buf = b;
+                }
+            }
+            return Ok(());
+        }
+        Ok(())
+    }
+
+    pub fn create_control_stream(&mut self, conn: &mut Connection) -> Res<()> {
```
agrover: Maybe just name `create()` or even `new()`? Also, consider making this a constructor (return `Res<ControlStreamLocal>`). stream_id would no longer be an Option), but instead make ControlStreamLocal at the Http3Connection level an `Option<ControlStreamLocal>`?

EDIT: I see this is just carried over from existing code so this could possibly be addressed separately if desired.

neqo-http3/src/control_stream_local.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl ::std::fmt::Display for ControlStreamLocal {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        write!(f, "Local control stream {:?}", self.stream_id)
+    }
+}
+
+impl ControlStreamLocal {
+    pub fn send_frame(&mut self, f: HFrame) {
```
agrover: doesn't actually send, it enqueues.

neqo-http3/src/control_stream_remote.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::hframe::{HFrame, HFrameReader};
+use crate::{Error, Res};
+use neqo_common::{qdebug, qinfo};
+use neqo_transport::Connection;
+
+// The remote control stream is responsible only for reading frames. The frames are handled by Http3Connection
+#[derive(Debug)]
+pub struct ControlStreamRemote {
+    stream_id: Option<u64>,
```
agrover: same suggestion as ControlStreamLocal could apply here.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+    pub fn get_type(&mut self, conn: &mut Connection, stream_id: u64) -> Option<u64> {
+        // On any error we will only close this stream!
+        loop {
+            let to_read = self.reader.min_remaining();
+            let mut buf = vec![0; to_read];
+            match conn.stream_recv(stream_id, &mut buf[..]) {
+                Ok((_, true)) => {
+                    self.fin = true;
+                    break None;
```
agrover: Maybe use `return` instead of `break` in this method? break says to me the value will be used later in the method, when in this case since the whole method is a loop, is not the case.

neqo-qpack/src/encoder.rs
```diff
@@ -495,15 +495,14 @@ impl QPackEncoder {
             .write_byte(QPACK_UNI_STREAM_TYPE_ENCODER as u8);
     }
 
-    pub fn add_recv_stream(&mut self, stream_id: u64) {
-        if self.remote_stream_id.is_some() {
-            panic!("Adding multiple remote streams");
+    pub fn add_recv_stream(&mut self, stream_id: u64) -> Res<()> {
```
agrover: If `add_recv_stream` now returns Res for Error::WrongStreamCount, then callers (e.g. Http3Connection::decode_new_stream) do not also need to convert the Err to the same WrongStreamCount error.

neqo-http3/src/connection.rs
```diff
@@ -4,155 +4,81 @@
...
-            if id == stream_id {
-                self.fin = self.frame_reader.receive(conn, stream_id)?;
-                return Ok(true);
-            }
-        }
-        Ok(false)
-    }
+pub trait Http3Events {
+    fn new() -> Self;
+    fn data_writable(&self, stream_id: u64);
```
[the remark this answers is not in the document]
ddragana: I was thinking about this, but til now we did not really need it. Let's see when I do the rest of refactor.

neqo-http3/src/connection.rs
```diff
@@ -901,2519 +703,178 @@ impl Http3Connection {
...
-    pub fn goaway_received(&self) {
-        self.insert(Http3Event::GoawayReceived);
-    }
-
-    pub fn connection_state_change(&self, state: Http3State) {
-        self.insert(Http3Event::StateChange(state));
+    fn handle_new_bidi_stream(
+        &mut self,
+        transactions: &mut HashMap<u64, TransactionServer>,
+        _conn: &mut Connection,
```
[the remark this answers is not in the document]
ddragana: this is because server and client behaves differently. The server adds a new transaction and server closes the stream, one need 'transactions' the other 'conn'.

neqo-http3/src/control_stream_local.rs
ddragana: The very first version of http3 code has used your suggestion but the code in Http3Connection gets complicated. Also for the control_stream_remote even more complicated because of not being able to borrow 'self' mutable twice, for example here:
  
```
        } else if self
            .control_stream_remote
            .receive_if_this_stream(&mut self.conn, stream_id)?
        {
            qdebug!(
                [self]
                "The remote control stream ({}) is readable.",
                stream_id
            );
            while self.control_stream_remote.frame_reader_done()
                || self.control_stream_remote.recvd_fin()
            {
                self.handle_control_frame()?;
                self.control_stream_remote
                    .receive_if_this_stream(&mut self.conn, stream_id)?;
            }
        }
```

neqo-qpack/src/encoder.rs
ddragana: This is because they are not the same type neqo_qpack::Error::WrongStreamCount anf neqo_http3::Error::WrongStreamCount. I can fix this in a different way as well, I will do it in a follow up.

neqo-http3/src/connection.rs
```diff
@@ -280,9 +230,18 @@ impl Http3Connection {
         });
     }
 
+    fn create_qpack_streams(&mut self) -> Res<()> {
+        qdebug!([self] "create_qpack_streams.");
```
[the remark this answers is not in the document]
agrover: Fixed, but there are other spots :smiley: can you do a quick grep for qdebugs and just convert to qtrace if they just print the function name? (or maybe consider removing them, since maybe the value they provide isn't high enough to justify if they only print the function name...)

neqo-http3/src/control_stream_local.rs
agrover: Would this be more correct as `queue_frame`? Not that this deserves a lot of back-and-forth, but let's just spend a minute to see if a better name is out there.

ddragana: I called it enqueue_frame, but queue_frame is fine too.

agrover:
:+1:, please merge once conflict has been resolved.

agrover:
There are some trivial issues that I am going to just go ahead and fix, if that's ok.

agrover:
oops, that went a little beyond what I was intending, but the end result is good so let's just move on :smiley:

PR #239, opened 2019-10-09
update to draft 23 constants

agrover:


agrover:
@martinthomson, `mitm_retry` test is failing and I don't understand why that might be. Could you take a look and maybe push a fix to this branch/PR, or give some guidance please?

martinthomson:
I had a copy of the initial salt in that test so that I could decrypt it.  Easy fix.

agrover:
ok cool. If this is good then can some kind person approve this PR?

PR #248, opened 2019-10-15
Interop resumption

ekr:
This fixes resumption and also makes it work with the interop harness.


agrover:
Just needs some trivial clippy cleanups @ekr

agrover:
I updated this a little but it needs more work to not require changes to Connection. Maybe looking at connection statechanged events can achieve the same goal?

PR #253, opened 2019-10-16
Calling stream_stop_sending should do something

ddragana:
After calling stream_stop_sending on a stream, we should not accept anymore data from the peer on that stream.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -625,7 +625,7 @@ impl SendStream {
         }
 
         if !matches!(self.state, SendStreamState::Send{..}) {
-            return Err(Error::FinalSizeError)
+            return Err(Error::FinalSizeError);
         }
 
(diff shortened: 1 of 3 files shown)


ddragana:
We may need more test wit stop_sending and reset.

ddragana:
Some test will fail and they will be fix when #252 lands.

ddragana:
It looks like there is more problems with stop_sending, e.g. receiving side ignores it as well.

neqo-transport/src/send_stream.rs
```diff
@@ -624,6 +624,11 @@ impl SendStream {
             });
         }
 
+        match self.state {
+            SendStreamState::Send { .. } => {}
+            _ => return Err(Error::FinalSizeError),
+        }
+
```
agrover: suggest `if matches!(self.state, SendStreamState::Send{..}) { return Err(Error::FinalSizeError)}`

PR #260, opened 2019-10-18
Http3 rework part2

### LINKED ISSUE #259: Extend the http3 server API
The http3 server should be able to handle requests with a body.
We will need to change how the response is sent as well. Currently the server holds a handler that is called when a complete request(currently this are only headers) is received.
If a request  contains body, the server may respond just after the headers are received or after the complete body is received.
Also the server should be able to cancel a request or only cancel the receiving side, i.e. calling STOP_SENDING.

ddragana:
Fixes #259

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -35,9 +35,7 @@ const HTTP3_UNI_STREAM_TYPE_PUSH: u64 = 0x1;
 const MAX_HEADER_LIST_SIZE_DEFAULT: u64 = u64::max_value();
 
 pub trait Http3Events: Default + Debug {
-    fn data_writable(&self, stream_id: u64);
     fn reset(&self, stream_id: u64, error: AppError);
-    fn new_requests_creatable(&self, stream_type: StreamType);
     fn connection_state_change(&self, state: Http3State);
     fn remove_events_for_stream_id(&self, remove_stream_id: u64);
 }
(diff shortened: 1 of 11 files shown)


neqo-http3/src/transaction_server.rs
```diff
@@ -75,35 +70,115 @@ impl TransactionServer {
...
+            ref mut buf,
+            ref mut offset,
+        } = self.recv_state
+        {
+            let (amount, fin) = conn.stream_recv(self.stream_id, &mut buf[*offset..])?;
+            qtrace!([label] "read_headers: read {} bytes fin={}.", amount, fin);
+            *offset += amount as usize;
+            if *offset < buf.len() {
+                if fin {
+                    // Malformated frame
```
agrover: typo, malformed

neqo-http3/src/server_events.rs
```diff
@@ -14,12 +15,20 @@ use std::rc::Rc;
 
 #[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone)]
 pub enum Http3ServerEvent {
-    /// A stream can accept new data.
-    DataWritable { stream_id: u64 },
+    /// Headers are ready.
+    Headers {
+        stream_id: u64,
+        headers: Option<Vec<Header>>,
```
agrover: Does this need to be an option or would the empty vec convey the same thing?

neqo-http3/src/connection.rs
```diff
@@ -694,15 +699,47 @@ impl Http3Handler<Http3ClientEvents, TransactionClient> for Http3ClientHandler {
...
+        events: &mut Http3ClientEvents,
+        stream_type: StreamType,
+    ) -> Res<()> {
+        events.new_requests_creatable(stream_type);
+        Ok(())
+    }
+
+    fn handle_new_push_stream(&mut self) -> Res<()> {
+        // TODO implement PUSH
+        qtrace!([self] "PUSH is not implemented!");
```
agrover: I think this is more important than trace. Maybe info? Or even warn?

neqo-http3/src/connection.rs
```diff
@@ -801,31 +833,38 @@ impl Http3Handler<Http3ServerEvents, TransactionServer> for Http3ServerHandler {
...
+    fn handle_stream_creatable(
+        &mut self,
+        _events: &mut Http3ServerEvents,
+        _stream_type: StreamType,
+    ) -> Res<()> {
+        Ok(())
+    }
+
+    fn handle_new_push_stream(&mut self) -> Res<()> {
+        qtrace!([self] "Error: server receives a push stream!");
```
agrover: qerror here?

PR #272, opened 2019-11-03
std::mem::MaybeUninit, not std::mem::uninitialized

hawkinsw:
Per https://doc.rust-lang.org/beta/std/mem/fn.uninitialized.html,
std::mem::uninitialized is deprecated as of Rust 1.39.0.
Getting ahead of this deprecation with a change does not affect
compilation under stable.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -23,7 +23,7 @@ use neqo_common::{qdebug, qinfo, qwarn};
 use std::cell::RefCell;
 use std::convert::{TryFrom, TryInto};
 use std::ffi::CString;
-use std::mem;
+use std::mem::{self, MaybeUninit};
 use std::ops::{Deref, DerefMut};
 use std::os::raw::{c_uint, c_void};
 use std::ptr::{null, null_mut, NonNull};
@@ -97,17 +97,17 @@ macro_rules! preinfo_arg {
 
 impl SecretAgentPreInfo {
     fn new(fd: *mut ssl::PRFileDesc) -> Res<Self> {
-        let mut info: ssl::SSLPreliminaryChannelInfo = unsafe { mem::uninitialized() };
+        let mut info: MaybeUninit<ssl::SSLPreliminaryChannelInfo> = MaybeUninit::uninit();
         secstatus_to_res(unsafe {
             ssl::SSL_GetPreliminaryChannelInfo(
                 fd,
-                &mut info,
+                info.as_mut_ptr(),
                 c_uint::try_from(mem::size_of::<ssl::SSLPreliminaryChannelInfo>())?,
             )
         })?;
 
         Ok(Self {
-            info,
+            info: unsafe { info.assume_init() },
             alpn: get_alpn(fd, true)?,
         })
     }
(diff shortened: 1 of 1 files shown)


neqo-crypto/src/agent.rs
```diff
@@ -97,7 +97,8 @@ macro_rules! preinfo_arg {
 
 impl SecretAgentPreInfo {
     fn new(fd: *mut ssl::PRFileDesc) -> Res<Self> {
-        let mut info: ssl::SSLPreliminaryChannelInfo = unsafe { mem::uninitialized() };
+        let mut info: ssl::SSLPreliminaryChannelInfo =
```
quininer: This is not the correct usage of `MaybeUninit`. You should write

```rust
let mut info: MaybeUninit<ssl::SSLPreliminaryChannelInfo> =
    MaybeUninit::uninit();

secstatus_to_res(unsafe {
    ssl::SSL_GetPreliminaryChannelInfo(
        fd,
        info.as_mut_ptr(),
        c_uint::try_from(mem::size_of::<ssl::SSLPreliminaryChannelInfo>())?,
    )
})?;

Ok(Self {
    info: info.assume_init(),
    alpn: get_alpn(fd, true)?,
})

Comment (hawkinsw) [reply neqo-crypto/src/agent.rs]: I *thought* that was the way to use it, but the Rust documentation "told" me otherwise. See here: https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html. That said, I understand that the documentation is making a point about instantaneously triggering the undefined behavior. So, of course, you are correct! Sending in a fix.

Comment (hawkinsw) [reply neqo-crypto/src/agent.rs]: Looks like we also need an unsafe{} around anywhere that we use assume_init(). Does this make sense to you?

Comment (quininer) [reply neqo-crypto/src/agent.rs]: Yes, `assume_init `requires unsafe. I have neglected this.

File: neqo-crypto/src/agent.rs [hunk 2]
```diff
@@ -147,22 +150,24 @@ pub struct SecretAgentInfo {
...
-            version: info.protocolVersion as Version,
-            cipher: info.cipherSuite as Cipher,
-            group: Group::try_from(info.keaGroup)?,
-            resumed: info.resumed != 0,
-            early_data: info.earlyDataAccepted != 0,
-            alpn: get_alpn(fd, false)?,
-            signature_scheme: SignatureScheme::try_from(info.signatureScheme)?,
+        Ok(unsafe {
+            Self {
+                version: info.assume_init().protocolVersion as Version,
```
quininer: It should only be `assumed_init` once.

hawkinsw: So, I call it once and assign the result to that T and then use that in the resulting Self{}? I don't see that noted anywhere in the docs for `assume_init` which is the only reason I am asking!

Thanks!

hawkinsw: I see it now. It's moving operation on self.

neqo-crypto/src/agent.rs
```diff
@@ -97,18 +97,21 @@ macro_rules! preinfo_arg {
...
                 c_uint::try_from(mem::size_of::<ssl::SSLPreliminaryChannelInfo>())?,
             )
         })?;
 
-        Ok(Self {
-            info,
-            alpn: get_alpn(fd, true)?,
+        Ok(unsafe {
+            Self {
+                info: info.assume_init(),
```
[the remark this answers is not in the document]
hawkinsw: Thanks!

PR #277, opened 2019-11-06
Fix last part of issue 26

ddragana:


agrover:
Sorry, that silly comma PR caused conflicts that will need to be rebased on top of.

Also while you're at it, can you please add commitlog coments for the first two commits?

Issue #278, opened 2019-11-06
blog.cloudflare.com assertion fails after connection close


agrover:
cmdline: `RUST_BACKTRACE=1 RUST_LOG=neqo_transport::dump ./target/debug/neqo-client https://blog.cloudflare.com/`
result:

```
</html>
<FIN[0]>
0s576ms DEBUG [Client 0x7fff66571b50] pn=27 type=Short
  TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: [107, 116, 104, 120, 98, 121, 101, 33] }
thread 'main' panicked at 'assertion failed: self.state_active()', neqo-http3/src/connection.rs:523:9
stack backtrace:
   0: std::panicking::default_hook::{{closure}}
   1: std::panicking::default_hook
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic
             at /builddir/build/BUILD/rustc-1.38.0-src/src/libstd/panicking.rs:411
   4: neqo_http3::connection::Http3Connection<E,T,H>::handle_connection_closing
             at neqo-http3/src/connection.rs:523
   5: neqo_http3::connection::Http3Connection<E,T,H>::check_connection_events
             at neqo-http3/src/connection.rs:366
   6: neqo_http3::connection::Http3Connection<E,T,H>::process_http3
             at neqo-http3/src/connection.rs:281
   7: neqo_http3::connection_client::Http3Client::process_http3
             at neqo-http3/src/connection_client.rs:243
   8: neqo_client::process_loop
             at neqo-client/src/main.rs:133
   9: neqo_client::client
             at neqo-client/src/main.rs:281
  10: neqo_client::main
             at neqo-client/src/main.rs:318
  11: std::rt::lang_start::{{closure}}
             at /builddir/build/BUILD/rustc-1.38.0-src/src/libstd/rt.rs:64
```

hawkinsw:
Patch incoming.

hawkinsw:
This happens because the `Http3Connection` sets its state to `Closing` in the `Http3Connection::close` function before it calls `Connection::close`. `Connection::close()` sets /its/ state to `Closing` which adds a `ConnectionEvent::StateChange` which, when handled, calls `Http3Connection::handle_connection_closing` which asserts whether its state is connected!

Got that? :-)

I think that there are two ways to handle this:
[...]

PR #281, opened 2019-11-09
Fix Issue #278

hawkinsw:
Closing is an acceptable state of the connection when `Http3Connection::handle_connection_closing` is invoked.

Add an `Http3Connection::state_closing` method to support this change.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -12,7 +12,7 @@ use crate::server_events::Http3ServerEvents;
 use crate::stream_type_reader::NewStreamTypeReader;
 use crate::transaction_client::TransactionClient;
 use crate::transaction_server::TransactionServer;
-use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn, Datagram};
+use neqo_common::{matches, qdebug, qerror, qinfo, qtrace, qwarn, Datagram};
 use neqo_crypto::AntiReplay;
 use neqo_qpack::decoder::{QPackDecoder, QPACK_UNI_STREAM_TYPE_DECODER};
 use neqo_qpack::encoder::{QPackEncoder, QPACK_UNI_STREAM_TYPE_ENCODER};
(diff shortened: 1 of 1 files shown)


hawkinsw:
Yes, I now realize that taking approach (1), as described in #278, causes multiple test failures. :-( Taking approach (2) does not cause any test failures.  I am pushing a new version of the fix that takes that approach. Please let me know what you think! 

I could not find a standards reference that defines an active connection, so it seems like it would reasonable to redefine /active/, as argued in the aforementioned comments on #278. 

Thanks again for all the great work!

neqo-http3/src/connection.rs
```diff
@@ -696,6 +696,13 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
         }
     }
 
+    fn state_closing(&self) -> bool {
+        match self.state {
```
martinthomson: I wonder if we can start using `matches!(self.state, Http3State::Closing(_))` for these more.

hawkinsw: Seems like this is a feature that is in nightly (https://doc.rust-lang.org/nightly/std/macro.matches.html) unless we want to take it from an external crate. Either way, it would be so cool to be able to write

`!matches!(self.state, Http3State::Closing(_))`

rather than what's there now.

Do you want to include the https://docs.rs/matches/0.1.8/matches/macro.matches.html crate to do this?

martinthomson: https://github.com/mozilla/neqo/blob/fa4b17afd26be8bf2c78cbc7247a31223b9cd2ca/neqo-common/src/lib.rs#L25 should do until we can get the one from `std`

hawkinsw: Thanks for pointing me to that. Sorry I didn't see it myself :-) Resolving this review -- see the new commit! Thanks again for the review!

hawkinsw:
Let me know what you think about this one! Thanks again for the feedback!

PR #292, opened 2019-11-11
Bump the version of rustup in the CI image

martinthomson:
If we're going to start using rust 1.39, then we probably should make sure that the CI image is up to date.


martinthomson:
Hold approval.  I'm getting failures building the image.

martinthomson:
OK, the time-travel problem seems to have passed.

PR #297, opened 2019-11-12
The server will have epoch 3 keys before it is in Connected state.

ddragana:


neqo-transport/src/connection.rs
```diff
@@ -867,6 +867,11 @@ impl Connection {
         let largest_acknowledged = self
             .loss_recovery
             .largest_acknowledged(PNSpace::from(hdr.epoch));
+        if (self.state == State::Handshaking) && (hdr.epoch == 3)  {
+            // Server has keys for epoch 3 but it is still in state Handshaking -> discharge packet.
+            assert_eq!(self.role(), Role::Server);
```
```suggestion
            debug_assert_eq!(self.role(), Role::Server);
```

neqo-transport/src/connection.rs
```diff
@@ -2849,4 +2854,42 @@ mod tests {
...
+        assert!(client_fin.as_dgram_ref().is_some());
+
+        let client_stream_id = client.stream_create(StreamType::UniDi).unwrap();
+        client.stream_send(client_stream_id, &[1, 2, 3]).unwrap();
+        let client_stream_data = client.process(None, now());
+        assert!(client_stream_data.as_dgram_ref().is_some());
+
+        // Now stream data gets before client_fin
+        let server_out = server.process(client_stream_data.dgram(), now());
+        assert!(server_out.as_dgram_ref().is_none()); // the packet will be discharged...
```
```suggestion
        assert!(server_out.as_dgram_ref().is_none()); // the packet will be discarded
```

Issue #27, opened 2019-05-23
http3: server-initiated bidi stream should be connection error


agrover:
I think I just made it just do stop sending, but no it should be a connection error.

ddragana:
fixed by #252

PR #299, opened 2019-11-12
Implement 0-Rtt in the Http3 client.

ddragana:


ddragana:
The test in this PR depend on #296, some will fail.

neqo-http3/src/connection.rs
```diff
@@ -298,7 +316,28 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
         self.streams_have_data_to_send.insert(stream_id);
     }
 
+    pub fn set_resumption_settings(&mut self, settings: Vec<(HSettingType, u64)>) -> Res<()> {
```
martinthomson: The interface to NSS and transport doesn't work like this.  It has a token that pickles all the necessary state.  It requires serialization/parsing, but I think that it is easier to conceptualize.

ddragana: This is the second time we have this confusion. Http3Connection does not implement an interface to the application. Http3Client does. Https3Client does the serialization, similar to the transport. PR #277 will pull out conn(neqo_transport::Connection) from Http3Connection into Http3Client (this is needed because of the Http3Server). To prepare for PR #27 I implemented the serialization/parsing in Http3Client.

neqo-http3/src/connection.rs
```diff
@@ -613,23 +670,83 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
...
+                            "The old value of {:?} was {} and not it is not defined.",
+                            st,
+                            zero_rtt_value.unwrap().1
+                        );
+                        return Err(Error::HttpSettingsError);
+                    } else {
+                        differ = true;
+                    }
+                }
+                if !differ {
```
martinthomson: You need to invert the value of `differ` for this to make sense.

neqo-http3/src/connection.rs
```diff
@@ -613,23 +670,83 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
...
+            Http3RemoteSettingsState::NotReceived => {
+                self.set_qpack_settings(s)?;
+                self.settings_state = Http3RemoteSettingsState::Received(s.to_vec());
+                Ok(())
+            }
+            Http3RemoteSettingsState::ZeroRtt(settings) => {
+                let mut differ = false;
+                for st in &[
+                    HSettingType::MaxHeaderListSize,
+                    HSettingType::MaxTableSize,
```
martinthomson: This should be called MaxTableCapacity.  The spec uses SETTINGS_QPACK_MAX_TABLE_CAPACITY.  Note also that this is a value that [cannot change](https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html#section-3.2.3) if the value is non-zero.

neqo-http3/src/connection.rs
```diff
@@ -613,23 +670,83 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
...
 
+    fn handle_settings(&mut self, s: &[(HSettingType, u64)]) -> Res<()> {
+        qinfo!([self], "Handle SETTINGS frame.");
+        match &self.settings_state {
+            Http3RemoteSettingsState::NotReceived => {
+                self.set_qpack_settings(s)?;
+                self.settings_state = Http3RemoteSettingsState::Received(s.to_vec());
+                Ok(())
+            }
+            Http3RemoteSettingsState::ZeroRtt(settings) => {
```
martinthomson: This function a little too much knowledge about settings.  Some of that could be moved to a wrapper.  I think that this would be much easier if you had a class that wrapped the finding of values with defaults.   

Before you ask, maximum header list size can default to 2^62 if you are concerned about "unlimited"; the effect is the same.

Then the calculation is easier: you find the value, check if it is bigger.  Then this code is:

```rust
Http3RemoteSettingsState::ZeroRtt(old_settings) => { 
  let mut qpack_changed = false;
  for st in &[  ... ] {
    let old_value = old_settings.get(s);
    let new_value = new_settings.get(s);
    let diff = i64::try_from(new_settings.get(s)).unwrap() - i64::try_from(old_value).unwrap();
    if old_value == new_value { continue; }
    if old_value > new_value { error!(); }
    match s {
      HSettingsType::MaxTableCapacity => {
        if old_value != 0 { error!(); }
        qpack_changed = true;
      },
      HSettingsType::BlockedStreams => qpack_changed = true,
      _ => (),
    }
  }
  if qpack_changed { self.qpack.update_settings(); }
  Ok(())
```

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
         }
         assert!(recv_header && recv_data);
     }
+
+    #[allow(clippy::cognitive_complexity)]
```
martinthomson: I'm resigned to having old tests grow a little so that they end up with this annotation, but not having new functions start with it.  Please try to find some way to break this down into more manageable chunks.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                }
+                ConnectionEvent::RecvStreamReadable { stream_id } => {
+                    if stream_id == 2 {
+                        // the control stream
+                        // Only stream type will be received no Settings frame will be received.
+                        let mut buf = [0u8; 100];
+                        let (amount, fin) = server.stream_recv(stream_id, &mut buf).unwrap();
+                        assert_eq!(fin, false);
+                        assert_eq!(amount, CONTROL_STREAM_DATA.len());
+                        assert_eq!(&buf[..9], CONTROL_STREAM_DATA);
```
martinthomson: This seems brittle.  When you define this, please define it alongside the default settings and explain their relationship.  That way, there is a better chance that they will get updated together.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                    } else if stream_id == 6 {
+                        let mut buf = [0u8; 100];
+                        let (amount, fin) = server.stream_recv(stream_id, &mut buf).unwrap();
+                        assert_eq!(fin, false);
+                        // the remote encoder stream may send a change capacity instruction
+                        // (0x3f, 0x45 = change capacity to 100)
+                        if expect_cap_instruction {
+                            // with a instruction
+                            assert_eq!(amount, 3);
+                            assert_eq!(buf[..3], [0x02, 0x3f, 0x45]);
```
martinthomson: Define this alongside the other control-related stuff.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                        assert_eq!(amount, CONTROL_STREAM_DATA.len());
+                        assert_eq!(&buf[..9], CONTROL_STREAM_DATA);
+                        control_stream = true;
+                    } else if stream_id == 6 {
+                        let mut buf = [0u8; 100];
+                        let (amount, fin) = server.stream_recv(stream_id, &mut buf).unwrap();
+                        assert_eq!(fin, false);
+                        // the remote encoder stream may send a change capacity instruction
+                        // (0x3f, 0x45 = change capacity to 100)
+                        if expect_cap_instruction {
```
martinthomson: Under what conditions do we expect the capacity instruction to be sent?

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                            assert_eq!(amount, 1);
+                            assert_eq!(buf[..1], [0x02]);
+                        }
+                        qpack_encoder_stream = true;
+                    } else if stream_id == 10 {
+                        let mut buf = [0u8; 100];
+                        let (amount, fin) = server.stream_recv(stream_id, &mut buf).unwrap();
+                        assert_eq!(fin, false);
+                        assert_eq!(amount, 1);
+                        assert_eq!(buf[..1], [0x3]);
```
martinthomson: These types need constants.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                    } else if stream_id == 0 {
+                        assert!(expect_request);
+                        check_header_frame(&mut server, stream_id, true);
+                        let _ = server.stream_send(
+                            stream_id,
+                            &[
+                                // headers
+                                0x01, 0x06, 0x00, 0x00, 0xd9, 0x54, 0x01, 0x33,
+                                // the data frame
+                                0x0, 0x3, 0x61, 0x62, 0x63,
```
martinthomson: As before, if you are using a standard request, then you should define the shape of the request and its expected serialization together so that people who have to maintain these tests have an easier time if they want to change something.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                            ],
+                        );
+                        server.stream_close_send(stream_id).unwrap();
+                        request = true;
+                    } else {
+                        panic!("unexpected event");
+                    }
+                }
+                ConnectionEvent::SendStreamWritable { stream_id } => {
+                    assert!((stream_id == 2) || (stream_id == 6) || (stream_id == 10));
```
martinthomson: Perhaps:

```rust
assert!(matches!(stream_id, 2 | 6 | 10));
```

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+                }
+                ConnectionEvent::StateChange(_) => (),
```
martinthomson: Are there any state changes that are not allowed?

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+        client.resumption_token().expect("should have token")
+    }
+
+    #[test]
+    fn zero_rtt_negotiated() {
+        let (mut client, mut server) = connect();
+        let token = exchange_token(&mut client, &mut server.conn);
+
+        let mut client = default_http3_client();
+
```
martinthomson:
```suggestion
```

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn zero_rtt_send_request() {
+        let (mut client, mut server) = connect();
+        let token = exchange_token(&mut client, &mut server.conn);
+
+        let mut client = default_http3_client();
+
```
martinthomson:
```suggestion
```

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+        assert!(client_out.as_dgram_ref().is_some());
+        let recvd_0rtt_reject = |e| e == Http3ClientEvent::ZeroRttRejected;
+        assert!(client.events().any(recvd_0rtt_reject));
+
+        // ...and the client stream should be gone.
+        let res = client.stream_close_send(request_stream_id);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), Error::InvalidStreamId);
+
+        // Client will send Setting frame and open new qpack streams.
```
martinthomson: This might make the test too complex, but it would be good to test now that the client can still make a request.  You might like to check that it sits on stream 0 too.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+            .set_resumption_token(now(), &token)
+            .expect("Set resumption token.");
+        assert_eq!(client.state(), Http3State::ZeroRtt);
+        let out = client.process(None, now());
+
+        assert_eq!(client.state(), Http3State::ZeroRtt);
+        assert_eq!(*server.state(), State::WaitInitial);
+        let out = server.process(out.dgram(), now());
+
+        // Check that control and qpack streams anda SETTINGS frame are received.
```
martinthomson:
```suggestion
        // Check that control and qpack streams and a SETTINGS frame are received.
```

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+    }
+
+    #[test]
+    fn zero_rtt_new_server_setting_are_the_same() {
+        // Connect and get a token
+        let (mut client, mut server) = connect();
+        let token = exchange_token(&mut client, &mut server.conn);
+
+        // Reconnect
+        let (mut client, mut server) = zero_rtt_server_setting(token, true);
```
martinthomson: Why not also include `connect()` and `exchange_token()` calls in `zero_rtt_server_setting`?

I think that what you really want is a range of functions:

* make_server(settings) -- make a server with specific settings
* default_server() -- make a server with default settings, which just calls `make_server()`
* connect_with(client, server, token) -- connects a pre-configured client and server and 0-RTT token
* connect() -- default connection

Or something like that.

neqo-http3/src/connection_client.rs
```diff
@@ -2364,4 +2430,676 @@ mod tests {
...
+
+        assert_eq!(
+            client.state(),
+            Http3State::Closing(CloseError::Application(265))
+        );
+        assert_eq!(*server.state(), State::Connected);
+    }
+
+    #[test]
+    fn zero_rtt_new_server_setting_blocked_streams_bigger() {
```
martinthomson: There is a lot of repetition in these tests.  They should each reduce to "server starts with config X, resumes with config Y, result is Z".  These repeat a lot of boilerplate.

neqo-http3/src/hframe.rs
```diff
@@ -426,6 +390,50 @@ impl HFrameReader {
     }
 }
 
+pub fn encode_settings(enc: &mut Encoder, settings: &[(HSettingType, u64)]) {
```
martinthomson: These functions should be attached to a real type.

ddragana:
The patch is rebased. Some parts needed to move, sorry for that.
My apologies for the big change in tests. I started with small changes, but to do it properly I needed to change connect function and add some constants, so this ended up being a bit of cleaning up of http3 client test. If you want I can try to separate it into a different patch.

neqo-http3/src/hframe.rs
```diff
@@ -37,14 +38,116 @@ pub enum HStreamType {
...
+    }
+}
+
+impl From<Vec<HSetting>> for HSettings {
+    fn from(settings: Vec<HSetting>) -> Self {
+        HSettings { settings }
+    }
+}
+
+impl From<&HSettings> for Vec<HSetting> {
```
martinthomson: You might instead implement `IntoIterator` here.  Or `Deref<[HSetting]>` so that people can just get at the values directly via the slice.

neqo-http3/src/hframe.rs
```diff
@@ -37,14 +38,116 @@ pub enum HStreamType {
...
+            };
+            if setting_type != HSettingType::UnknownType {
+                self.settings.push(HSetting::new(setting_type, value));
+            }
+        }
+        Ok(())
+    }
+}
+
+impl From<Vec<HSetting>> for HSettings {
```
martinthomson: You should implement conversion from a slice reference (`From<&[HSetting]>`) instead.  But how often do you really need this?

neqo-http3/src/hframe.rs
```diff
@@ -104,23 +207,7 @@ impl HFrame {
...
-                                enc_inner.encode_varint(iter.1);
-                            }
-                            HSettingType::BlockedStreams => {
-                                enc_inner.encode_varint(SETTINGS_QPACK_BLOCKED_STREAMS as u64);
-                                enc_inner.encode_varint(iter.1);
-                            }
-                            HSettingType::UnknownType => {}
-                        }
-                    }
+                    settings.encode_settings(enc_inner);
```
martinthomson: Maybe move the outer `encode_vvec_with` and rename the function to `encode_frame` or `encode_frame_contents` (and change `decode_settings` to `decode_frame` to match)

neqo-http3/src/hframe.rs
```diff
@@ -37,14 +38,116 @@ pub enum HStreamType {
...
+    }
+}
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct HSettings {
+    settings: Vec<HSetting>,
+}
+
+impl HSettings {
+    pub fn new(settings: Vec<HSetting>) -> Self {
```
martinthomson: This should take a slice and not a Vec.  That will make some of the code below in tests slightly easier to work with.

neqo-http3/src/hframe.rs
```diff
@@ -37,14 +38,116 @@ pub enum HStreamType {
...
+#[derive(Clone, PartialEq, Debug, Copy)]
 pub enum HSettingType {
     MaxHeaderListSize,
-    MaxTableSize,
+    MaxTableCapacity,
     BlockedStreams,
     UnknownType,
 }
 
+fn hsetting_default(setting_type: HSettingType) -> u64 {
```
martinthomson: I think that there is enough setting-specific code here that settings can be moved to a new file.

neqo-http3/src/connection_client.rs
```diff
@@ -2535,4 +2360,569 @@ mod tests {
...
+    }
+
+    #[test]
+    fn zero_rtt_send_request() {
+        let (mut client, mut server) = start_with_0rtt();
+
+        let request_stream_id = client
+            .fetch("GET", "https", "something.com", "/", &[])
+            .unwrap();
+        assert_eq!(request_stream_id, 0);
```
martinthomson: It might pay to factor this into a function as well.  It's the standard request you make, you have its encoding up above, but you rely on copy-paste to get it right.

PR #303, opened 2019-11-15
Ack parsing bug.

ddragana:
This is a n old bug from fixing clippy.


ddragana:
It looks like we need more test here :)

ddragana:
> Did you want to add the tests now?

Sorry, I was busy with other stuff. I will add a test. The line may panic because we try to do 0-1 where 0 is a u64

PR #310, opened 2019-11-16
Draft 24

martinthomson:


agrover:
`RUST_BACKTRACE=1 RUST_LOG=debug ./target/debug/neqo-client https://quic.rocks:4433/`

`0s663ms INFO [Client 0x7fff3a116758] ConnectionClose received. Error code: Transport(28) frame type 6 reason 28:Server did not receive a known ALPN`

martinthomson:
So I fixed the ALPN, but we're getting `Output::Callback(0)`, which is not working very well.

agrover:
> So I fixed the ALPN, but we're getting Output::Callback(0), which is not working very well.

I'm not seeing a Callback(0). You were seeing it against quic.rocks yes? Can you still repro? If so let's deal with that separately in a new issue. I'm going to approve this PR.

Issue #316, opened 2019-11-17
Cannot publicly create a `neqo_transport::Connection::TransportParameter`


hawkinsw:
Right now, `neqo_transport::Connection::set_local_tparam` is public but cannot actually be used because there is no way to publicly create a `neqo_transport::Connection::TransportParameter`. Refactor to correct for this.

This seems like an oversight. Please let me know if I am misunderstanding. Just trying to help! 

PR to follow.

agrover:
@hawkinsw I think you fixed this, closing. Plz reopen if I'm mistaken.

PR #317, opened 2019-11-17
Issue #316: Make neqo_transport::Connection::set_local_tparam publicly usable

hawkinsw:
Right now, `neqo_transport::Connection::set_local_tparam` is public but cannot actually be used  because there is no way to publicly `create a neqo_transport::Connection::TransportParameter`. Refactor to correct for this.


hawkinsw:
This is my first substantial PR -- I tried to do everything right. I hope that it is okay. Please let me know how I can make it better or if I have misunderstood anything. Thank you for all the work that you are doing on this library!

hawkinsw:
FYI: I chose `tp_constants` because I thought that `transportparameters_constants` was too verbose. I don't know whether `tp_constants`, `transportparameters_constants`, `transportparameters_consts`, `tparams_constants` or something else matches the style the best.

PR #318, opened 2019-11-18
transport: Remove derive_more dependency.

emilio:
I need to update this dependency to get rid of old syn and quote in
mozilla-central.

But it seems like in this case it's easier to get rid of it, it saves like five
lines of code, and brings in a ton of dependencies.

Let me know if you want to keep it and update the dependency instead, but I
think it's not worth it in this case.

### DIFF
--- a/neqo-transport/Cargo.toml
+++ b/neqo-transport/Cargo.toml
@@ -8,7 +8,6 @@ license = "MIT/Apache-2.0"
 [dependencies]
 neqo-crypto = { path = "../neqo-crypto" }
 neqo-common = { path = "../neqo-common" }
-derive_more = "0.13"
 lazy_static = "1.0"
 rand = "0.6"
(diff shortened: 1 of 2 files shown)


emilio:
r? @ddragana 

(Also, information on how to get this patch into Gecko would be appreciated. Thanks!)

emilio:
Ah, guess it's just a matter of updating the git revision in `netwerk/socket/neqo_glue/Cargo.toml`, that wfm :)

ddragana:
> Ah, guess it's just a matter of updating the git revision in `netwerk/socket/neqo_glue/Cargo.toml`, that wfm :)

in bug 1596069, I have switch to use `tag`, so we will need to make a new release. When do you want this to be in m.-c.?

emilio:
No particular rush, whenever you get to it :)

Issue #319, opened 2019-11-18
thread 'main' panicked at 'assertion failed: delta < self.items.len()', neqo-common/src/timer.rs:85:9


hawkinsw:
To repro, 

`RUST_BACKTRACE=1 cargo run --bin neqo-http3-server -- -d test-fixture/db/`

and wait longer than the duration of the server's underlying timer wheel (3 minutes will do). Then, 

`cargo run --bin neqo-client https://localhost:4433`

There are other ways to trigger this, but this is the easiest. I have found the reason we have this problem, and would like to get your advice on the solution. 

It happens in the scenario where:
1. The server's timer wheel is not getting tick'd because there are no connections to service for the entirety of the timer wheel's span, and
2. a newly-accepted client connection, `a`, is `process()`'d resulting in a callback. 

At that point, the server will attempt to remove the timer at `a.last_time` (which matches the time the connection was accepted). Unfortunately, the assertion is triggered because that is outside the span of the timer wheel. 

The solution seems to be rather simple:
gate the logic in `Timer::remove(t)` so that it immediately returns `None` if `t` is somewhere in the future beyond `Timer::now + Timer::span()`. That seems reasonable since there is no way that `t` could somehow be in the timer wheel. 

I have a patch that uses that logic and I will put in a PR. I tested with `cargo test` and everything passes. If that's bad logic, please feel free to tell me so. I'm glad to work with you on a fix that follows best practices, etc. 

As ever, thanks!
Will

agrover:
fixed by #320

PR #320, opened 2019-11-18
Issue #319: Do not try to remove a timer outside a timer wheel's span

hawkinsw:
If the timer-to-be-removed is more than `Timer::span()` in the future,
there's no way that it can be in the timer wheel. With this patch,
the function will immediately return `None` and won't even search for it.

### DIFF
--- a/neqo-common/src/timer.rs
+++ b/neqo-common/src/timer.rs
@@ -139,6 +139,9 @@ impl<T> Timer<T> {
         if time < self.now {
             return None;
         }
+        if time > self.now + self.span() {
+            return None;
+        }
         let bucket = self.time_bucket(time);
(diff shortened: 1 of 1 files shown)


hawkinsw:
> I don't completely understand the issue here, but this is unquestionably good. I would appreciate if you added a test for this though.

@martinthomson Thanks for the feedback! I will absolutely create a test. The updated patch might take until the end of week (a little backed up this week with "real" work). Hope that's okay. Thanks again!!

hawkinsw:
@martinthomson Thanks for reviewing my changes last weekend. I hope that the tests live up to your expectations. It's the first time that I have written tests for this project so I was mostly copying-and-pasting from existing tests.

Issue #326, opened 2019-11-25
Possible bug in setting flow control connection values from transport parameters 


hawkinsw:
Please be gentle ...

I am going way outside of my comfort zone by suggesting that our (I apologize for the presumption that I have any ownership in this!) implementation has used the wrong transport parameter to set flow-control connection parameters in a particular instance. The description of `INITIAL_MAX_STREAM_DATA_BIDI_REMOTE`/`LOCAL` in the spec are hard for me to understand. I have contacted several people in the QUIC world to verify my understanding but still I am not 100% confident.

All that said, it seems like neqo uses the remote's tp value at the index of `tp_constants::INITIAL_MAX_STREAM_DATA_BIDI_REMOTE` when it should use the remote's tp value at the index of `tp_constants::INITIAL_MAX_STREAM_DATA_BIDI_LOCAL` value to set a new stream's sending limit for a remote-originated BiDi stream in `neqo_transport::Connection::obtain_stream`. 

I am putting a PR that includes documentation on how each of flow control limit values are chosen for BiDi streams that I hope will make (what I think is) the flipped case more obvious than the textual description above. 

As I said, I am taking a pretty big leap in noting this and suggesting a fix, so please be kind if I am wrong :-)

agrover:
let's assign unique values to params on both sides of a connection.

host A
* initial_max_stream_data_bidi_local = 2
* initial_max_stream_data_bidi_remote = 3
* initial_max_stream_data_uni = 4

Host B
* initial_max_stream_data_bidi_local = 5
* initial_max_stream_data_bidi_remote = 6
* initial_max_stream_data_uni = 7

A (client) opens uni to B: id = 2 A->B msd = 7
A opens bidi to B: id = 0 A->B msd = 6 B->A msd = 2

B (server) opens uni to A: id = 3 B->A msd = 4
B opens bidi to A: id = 1 B->A msd = 3 A->B msd = 5

There's also a difference in the number of times BIDI_LOCAL and BIDI_REMOTE are used in the code, which seems weird.

Ok I'm convinced you are correct. Great catch!

PR #327, opened 2019-11-25
Issue #326: Use initial_max_stream_data_bidi_local to set this peer's sending limit on remote-originated bidi streams.

hawkinsw:


hawkinsw:
> Great find, thanks again.

Thank you for your review and for thinking through this with me! 

> 
> Could you please convert block quotes to use `//`? I think this plus extending comment lines to column 100 could get each comment block down to three lines perhaps. Maybe even two.

Gladly! I :heart: `rustfmt` but as someone who developed a commenting style writing C, old habits die hard. In this revision I used [`wrap_comments`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#wrap_comments) (and friends) to implement your formatting suggestions. I hope that I got it right. 

> 
> Not a critique of this PR, but there was a change recently to allow tests to change tp vals. Would that allow a test to be added for this area, perhaps?

That *seems* doable. Would you like me to look into this? I'm always trying to improve my skills writing tests, so I'd be happy to do it. 

Thanks again for the review!

PR #329, opened 2019-11-27
Build NSS into docker image

martinthomson:
This should make CI builds a LOT faster.

Unfortunately, I have been unable to test that this works because dockerhub is acting strangely and I'm unwilling to give it the permissions that it thinks it needs.

The resulting image is 1.7G, which is hugely overweight.  For reference, NSS builds use an image that is half that size.  I wanted to keep the ability to track NSS and NSPR versions, and about 300M of that extra is in the NSS checkout, a debug build of NSS and NSPR, and other intermediate files.  

I thought that NSS was the main contributor to the bloat here, but it seems like installing rust is the main contributor.  Installing the minimal set of things we rely on bumps the image from 561M to 1.42G.  That probably bears further investigation.

Originally thinking that NSS was the reason for the bloat, I learned something interesting about mercurial.  If you remove a directory, it will happily update even if there are changes to that directory.  That allowed me to remove some of the larger directories from NSS.  That's worth a couple of hundred meg.

As a bonus, I've disabled building of tests for NSS, which - in addition to saving a lot of space - drives the compilation time down a lot.  That applies always, so you should see that benefit elsewhere.


martinthomson:
So it looks like there is no easy way to trim the image down in size.  The rust toolchain is enormous in ways that seem to be unavoidable.  Most of the space is taken with .rlib files that look somewhat important.  They seem to compress reasonably well, so it's not so bad.

Our existing image is only just shy of the 1.71G of this new image at 1.66G.

martinthomson:
Hey Andy, I'm looking at further improving the CI setup, but this is in the way.  After #393 lands (that contains a workaround for the clippy problems), I think that this should be good to get in.

A verbal approval is probably what I'll need, as mergify will screw things up as this needs to land on the docker branch first.

agrover:
Please proceed!

martinthomson:
Hmm, I realized after pushing this that I can't delete some of the extra cruft from NSS.  That's unfortunate, but it doesn't really change the image size much.  I'll now take a look at caching in circle.  I've learned a few things lately that should save us a ton of time.

martinthomson:
So I've learned a few things now that I've updated the docker image.  I've tried to capture that here.

Firstly, circleci caching doesn't seem to help much.  I've kept it for the git repository and NSS checkout, because that is cheap enough and I can't say for certain if there are benefits long term as we add more code here and NSS continues to add more code on top of what we capture in the docker image (currently, the docker image contains NSS head-of-tree, so it's hard to see any benefit from the cache).

Caching ~/.cargo and the target directory costs a lot.  To the point that saving the cache causes the build to take longer than before.  It also didn't drive build times down noticeably.  I think that is because the network on these boxes is super fast, so even large clones are cheap (which is what we are likely seeing with NSS too).  Also, it appears as though cargo doesn't use the cache that much, if at all.

With these changes, the build takes 1 minute less, which seems like a reasonable outcome.  If I could drop some of the bigger directories from NSS easily, I would, but newer versions of mercurial prompt on update if files are removed.  When I first tested this, mercurial didn't update files that were removed, which was sketchy, but exactly what I wanted.  Now it throws out prompts.

martinthomson:
So right now it looks like caching NSS isn't helping.  But the builds in my sample aren't great:

No caching: https://app.circleci.com/jobs/github/mozilla/neqo/1121
Repeat build: https://app.circleci.com/jobs/github/mozilla/neqo/1122
Another repeat: https://app.circleci.com/jobs/github/mozilla/neqo/1123

Note that everything takes longer in the latter builds, so it's not clear that the 32s to save and restore the NSS cache is helping, but the same build on my fast machine (with many more cores) takes 38s, and that doesn't include cloning the repo, so caching is still probably a win overall.

agrover:
Do you want to just leave as-is? Shorter CI times are only nice-to-have, not critical.

martinthomson:
On balance, I think that this is the right approach.  As NSS diverges from the baseline we have in the image, this should start to pay dividends.

martinthomson:
Took me a while, but this is now green.

PR #334, opened 2019-12-04
Http3conn rework

ddragana:
Http3Connection only handles common elements:
- holds control and qpack streams,
- holds a HashMap of active transactions,
- holds the Http3State of a http3 connection,
- performs sending data for control, qpack and transaction streams,
- handles decoding of new unidi streams(the client/server handlers differ in how they handle a  Push streams, therefore handle_new_unidi_stream returns if a stream was a push stream and the client/server handler consume them appropriately),
- handles streams reading (this function may return that a stream is a push stream or a list of control frames to be handle by the client/server handler)

The client/server handlers:
- holds a base_handler(Http3Connection)
- http3 events
- neqo_transport::Connection (this is only for the client)
- process_htttp3 has been moved from Http3Connection to the client/server handlers,
- also check_connection_events which iterates through ConnectionEvents and calls handlers for the events (handlers are implemented in Http3Connection or in the client/server handlers)

Move server_events::Http3Handler into a separate file.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -22,6 +22,12 @@ const HTTP3_UNI_STREAM_TYPE_PUSH: u64 = 0x1;
 
 const MAX_HEADER_LIST_SIZE_DEFAULT: u64 = u64::max_value();
 
+pub(crate) enum HandleReadableOutput {
+    NoOutput,
+    PushStream,
+    ControlFrames(Vec<HFrame>),
+}
+
 pub trait Http3Transaction: Debug {
     fn send(&mut self, conn: &mut Connection, encoder: &mut QPackEncoder) -> Res<()>;
     fn receive(&mut self, conn: &mut Connection, decoder: &mut QPackDecoder) -> Res<()>;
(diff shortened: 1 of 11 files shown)


neqo-http3/src/transaction_server.rs
```diff
@@ -333,12 +333,10 @@ impl Http3Transaction for TransactionServer {
         matches!(self.send_state, TransactionSendState::SendingResponse { .. })
     }
 
-    fn is_state_sending_data(&self) -> bool {
-        self.has_data_to_send()
+    fn reset_receiving_side(&mut self) {
+        self.recv_state = TransactionRecvState::Closed;
```
martinthomson: Did you add a test case for this?

neqo-http3/src/connection.rs
```diff
@@ -421,8 +262,6 @@ impl<E: Http3Events + Default, T: Http3Transaction, H: Http3Handler<E, T>>
         assert!(self.state_active());
 
         if let Some(t) = self.transactions.get_mut(&stream_id) {
-            // Post the reset event.
-            self.events.reset(stream_id, app_err);
             // Close both sides of the transaction_client.
             t.reset_receiving_side();
             t.stop_sending();
```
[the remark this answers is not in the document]
ddragana: I think it is fine to make this decision in the http3 code. We can open a separate issue to discuss this, let's leave this only for refactoring.

neqo-http3/src/transaction_server.rs
ddragana: I added this to be completely correct, but anytime rest_receiving_side is called on the server side we also remove transaction, so we cannot actually test this, it is also fine not to set the state.

Issue #261, opened 2019-10-18
Optimise sending data in http3 code


ddragana:
If a stream is block but still have data we will try to write on every iteration. We should handle this better and wait for stream to be unblocked.

hawkinsw:
I would like to tackle this and wanted to make sure that I have context correct. You are referring to the condition where a stream has data to send but is blocked because it has exhausted its tokens?

```
   If a sender runs out of flow control credit, it will be unable to
   send new data and is considered blocked.  A sender SHOULD send a
   STREAM_DATA_BLOCKED or DATA_BLOCKED frame to indicate it has data to
   write but is blocked by flow control limits.  These frames are
   expected to be sent infrequently in common cases, but they are
   considered useful for debugging and monitoring purposes.
```

from https://tools.ietf.org/html/draft-ietf-quic-transport-22#section-4.1

Thanks for all the great work you're doing on neqo. I hope that I'll be able to help.

martinthomson:
My understanding is that `Http3Transaction::send` is invoked every time the stack is invoked.  That object doesn't track whether the stream is blocked or not, but it could track that and return `false` from `Http3Transaction::has_data_to_send` so that it wouldn't be called every time.

I'm not sure how much of an advantage this is unless we see a lot of blocking.

ddragana:
I think I wanted to change behavior so that a stream is not add to streams_have_data_to_send if it is blocked and only added if it gets unblocked. This will not make a big difference. Also I wanted to change how we consume streams_have_data_to_send. Currently we do
`let to_send = mem::replace(&mut self.streams_have_data_to_send, BTreeSet::new());`
we should actually take single element until it is empty, because a stream can be added again to let self.streams_have_data_to_send during iteration.

hawkinsw:
I completely understand that distinction and that discussion. But, I still wanted to confirm that I am looking in the right part of the spec to understand the context of how a stream would get into the blocked state.

martinthomson:
Yep, you got it right.  A peer provides MAX_STREAM_DATA and MAX_DATA.  Streams are blocked as long as data sent runs to that limit and those values haven't increased.

hawkinsw:
Thanks for confirming. I am really excited about working on this! It's a free-time thing for me, but I will do my best to make progress as quickly as possible!

hawkinsw:
[...]

PR #345, opened 2019-12-08
Create events for when send streams become blocked

hawkinsw:
In light of use cases like #261, it seems like a good idea
to create/expose a DataBlocked event (a parallel of a DataWritable
event) when a send stream becomes blocked because of a lack
of credits.

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -18,6 +18,8 @@ pub enum Http3ClientEvent {
     HeaderReady { stream_id: u64 },
     /// A stream can accept new data.
     DataWritable { stream_id: u64 },
+    /// A stream has no more send credits available.
+    DataBlocked { stream_id: u64 },
     /// New bytes available for reading.
     DataReadable { stream_id: u64 },
     /// Peer reset the stream.
@@ -48,6 +50,10 @@ impl Http3ClientEvents {
 
     pub fn data_writable(&self, stream_id: u64) {
         self.insert(Http3ClientEvent::DataWritable { stream_id });
+    }
+
+    pub fn data_blocked(&self, stream_id: u64) {
+        self.insert(Http3ClientEvent::DataBlocked { stream_id });
     }
 
     pub fn data_readable(&self, stream_id: u64) {
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -320,6 +320,9 @@ impl Http3Client {
                     stream_id,
                     app_error,
                 } => self.handle_stream_stop_sending(stream_id, app_error)?,
+                ConnectionEvent::SendStreamBlocked { stream_id } => {
+                    self.events.data_blocked(stream_id)
+                }
                 ConnectionEvent::SendStreamComplete { .. } => {}
                 ConnectionEvent::SendStreamCreatable { stream_type } => {
                     self.events.new_requests_creatable(stream_type)
--- a/neqo-http3/src/connection_server.rs
+++ b/neqo-http3/src/connection_server.rs
@@ -117,7 +117,9 @@ impl Http3ServerHandler {
                         }
                     }
                 },
-                ConnectionEvent::SendStreamWritable { .. } => {}
+                ConnectionEvent::SendStreamWritable { stream_id } => {
+                    self.events.data_writable(stream_id)
+                }
                 ConnectionEvent::RecvStreamReadable { stream_id } => {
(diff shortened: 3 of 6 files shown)


hawkinsw:
This is my first substantive patch and I am submitting tentatively. I hope that it is at least worthy of your review. I am eager to see your suggestions for improving it and to make the requested changes. Please let me know if there are others who I should tag for review instead -- I just took github's advice on who to request :-)

Thanks for all the support you've given me to date.
Will

hawkinsw:
> This seems like a good change, but it would be easier to tell if there were tests that showed the value of the change.

Funny you mentioned tests! As soon as I submitted this, I thought, "Wow, I should add a test.". Sorry that I didn't include one initially. 

The updated PR I just submitted include a test that, _I think_, addresses your concern. As usual, thank you!

agrover:
I'm not clear why we need a new event here. The key async piece seems to be when the stream is unblocked, and there already is an event for that.

Maybe we could tackle this from the other direction -- fixing the issue first mentioned in #261 and seeing what events are needed to work efficiently? Dragana's [comment](https://github.com/mozilla/neqo/issues/261#issuecomment-549727369) could be a good place to start -- either don't add a stream to `have_data_to_send` until unblocked, or maybe add a `blocked` bool to the item in the collection, and only try sending if blocked=false. I'm guessing `Http3Transaction::send()` will likely need to return either a bool (blocked true/false) or usize (bytes sent) instead of `()` so that you can know if lower layers could accept everything or not, and may mean you have to wait to receive a StreamWritable event before trying again.

hawkinsw:
> I'm not clear why we need a new event here. The key async piece seems to be when the stream is unblocked, and there already is an event for that.
> 
> Maybe we could tackle this from the other direction -- fixing the issue first mentioned in #261 and 

Thanks for the feedback, @agrover ! A great suggestion to use that approach and one that I was already taking. I simply broke things in two because I thought this was a conceptually different issue and was trying to do the right thing.  

> seeing what events are needed to work efficiently? Dragana's [comment](https://github.com/mozilla/neqo/issues/261#issuecomment-549727369) could be a good place to start -- either don't add a stream to `have_data_to_send` until unblocked, or maybe add a `blocked` bool to the item in 

I agree that this is definitely one part of the optimization. I, too, agree that to do this we do *not* need the event introduced here. As you rightly say, the key async piece is that the stream is writable and we already have an event for that. 

I introduced this event to solve what I thought was the other part of the optimization. Here's the case I was considering:
1. The Http3 server generates a response and inserts it into the underlying transport (with `set_response`). 
1. Only part of that response can be sent before the stream runs out of credit.

In this case, any subsequent invocations of `set_response` can easily tell whether the underlying stream is blocked (as you rightly pointed out) and we could optimize without recourse to this event. However, any part of the Http3 server that continues to attempt to send the outstanding bytes would have no indication that the stream is blocked and, in an act of futility, attempt to resend those as it services the open streams. 

It would be possible for the Http3 server  to set a flag on its reference to the stream if/when `set_response` said that the stream is blocked (if we modified that function in the way you suggested below). However, that would assume that there would be subsequent calls to `set_response`. The Http3 server could also poll the stream during its event loop to check for the status and set the proper flag then. That is certainly an option and would not require this event.

I am sure that I am not thinking through the issue the way you are, as someone who has great experience with the codebase and the protocol. However, I wanted to write this so you didn't think that I just slapped something together and submitted it for review. :-)

Thank again you for the response and I hope this made some sort of sense!

Will

> the collection, and only try sending if blocked=false. I'm guessing `Http3Transaction::send()` will likely need to return either a bool (blocked true/false) or usize (bytes sent) instead of `()` so that you can know if lower layers could accept everything or not, and may mean you have to wait to receive a StreamWritable event before trying again.

hawkinsw:
Rebased the PR to be compatible with master after @ddragana's refactor of Http3Connection.

hawkinsw:
@agrover and @dragana, I was wondering if you could take another look at this and provide me some guidance? Thanks!

agrover:
It's hard for me to assess fully because this PR doesn't include the changes that would make use of the new event, only the addition of the event itself. Would it be possible to also see another commit that uses the events to implement the http3 processing improvement you describe?

hawkinsw:
Absolutely! As soon as I resolve the documentation issue I'll return to
this and post a follow up!

Thanks for the note!

agrover:
If this PR is going to be reviewed and move forward I think it's probably better to open a fresh one. Closing.

Issue #346, opened 2019-12-08
StreamId should be in the public interface


martinthomson:
Everywhere that we use `u64` for a stream ID, we should be using `StreamId`.  Then that type can be moved to the public interface.

ddragana:
The reason why I use u64 and not streamId is because http3 code do not care of stream types, streamId is just a identifire. it just trust what neqo-transport gives it. 
Also I did not want to expose StreamId  structure to c++ code.

agrover:
`StreamId` is basically helper methods on a u64. The value of exposing `StreamId` would be that clients could call these methods -- but if it's just used as an opaque token/identifier then the value is lost. I haven't looked but it sounds like that's how neqo_glue & Necko code treat stream id, so there wouldn't actually be a benefit, and actually might require some conversions in neqo_glue if this change is made.

That said, I'm still in favor of this change. Typing things is good. We just need to be aware that this may be nice on the Neqo side of things but actually adds a little work on the Gecko/Necko side of things to adapt to the change.

mahakbansal2019:
Hello!!
I want to work on this so could you please tell more details about this ?

PR #347, opened 2019-12-09
change u64 to streamId

michiboo:
Hi I changed the type for stream_id from u64 to StreamId, as describe in #346

Where can I add the type to public interface?

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -10,6 +10,7 @@
 use neqo_common::{matches, Datagram};
 use neqo_crypto::{init, AuthenticationStatus};
 use neqo_http3::{Header, Http3Client, Http3ClientEvent, Http3State, Output};
+use neqo_transport::stream_id::StreamId;
 use neqo_transport::FixedConnectionIdManager;
 
 use std::cell::RefCell;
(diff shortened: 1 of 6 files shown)


martinthomson:
in lib.rs add `pub use stream_id::StreamId`. But first, you need to get this passing tests. `cargo fmt` is the first step.

michiboo:
Hi for qdebug, I try to use stream_id.0 but it has private restriction, how should I go around this?

martinthomson:
`#[derive(Debug)]` if you can use `{:?}`, or you can implement `Display`.

Or you can use the `Into` implementation.  Or you can make the `u64` "pub".

agrover:
poke, this looks like it needs some rebasing but has received a favorable overall response.

agrover:
Thanks so much for updating!!!

So... I know this is contrary to what @martinthomson said above but I would really prefer that StreamId implement `Display` rather than making the u64 pub. This should be something like:
```
impl Display for StreamId {
        write!(f,"{}", self.as_u64())
    }
}
```
but then all the `.0`s in all the qdebugprints go away.

Also, running `git grep "stream_id: u64"` shows a bunch of hits still. Those spots should all be changed over too. Let's go all the way.

agrover:
Finally, let's re-export StreamId at the top level of neqo_transport so `use neqo_transport::stream_id::StreamId` becomes `use neqo_transport::StreamId`

martinthomson:
What @agrover said would be great, I agree.

agrover:
@michiboo ping?

PR #355, opened 2019-12-15
Fix #354: QPACK: Reset increment count after sending a header acknowledgment

### LINKED ISSUE #354: QPACK: Sending a header ack does not reset the incremented count
I am still very new to reading the quic specs, so please go easy on me if I have erred:

According to https://tools.ietf.org/html/draft-ietf-quic-qpack-11#section-2.1.4, sending a header acknowledgement is an implicit recognition that the entries processed in the header were inserted into the table. Therefore, sending increment instructions for those entries would be erroneous. Neqo does not reset the decoder's increment value upon sending a header acknowledgment. 

I discovered this while interoperating `h2load`. After adding the (incoming) patch, things worked again!

Patch to follow.

hawkinsw:
Per https://tools.ietf.org/html/draft-ietf-quic-qpack-11#section-2.1.4,
it seems that a header acknowledgment obviates the requirement of
sending increment instructions, in certain cases. This patch makes sure
that the decoder does not send increment instructions for only the
potentially-blocking references from the header block which are
implictly acknowledged by the header acknowledgment.

### DIFF
--- a/neqo-qpack/src/decoder.rs
+++ b/neqo-qpack/src/decoder.rs
@@ -74,7 +74,6 @@ enum QPackDecoderState {
 pub struct QPackDecoder {
     state: QPackDecoderState,
     table: HeaderTable,
-    increment: u64,
     total_num_of_inserts: u64,
     max_entries: u64,
     send_buf: QPData,
@@ -91,7 +90,6 @@ impl QPackDecoder {
         Self {
             state: QPackDecoderState::ReadInstruction,
             table: HeaderTable::new(false),
-            increment: 0,
             total_num_of_inserts: 0,
             max_entries: (f64::from(max_table_size) / 32.0).floor() as u64,
             send_buf: QPData::default(),
@@ -212,7 +210,6 @@ impl QPackDecoder {
                             qdebug!([label], "received instruction - duplicate index={}", v);
                             self.table.duplicate(v)?;
                             self.total_num_of_inserts += 1;
-                            self.increment += 1;
                             self.state = QPackDecoderState::ReadInstruction;
                         } else {
                             self.state = QPackDecoderState::Duplicate { index: v, cnt };
@@ -317,7 +314,6 @@ impl QPackDecoder {
                                     value_to_insert,
                                 )?;
                                 self.total_num_of_inserts += 1;
-                                self.increment += 1;
                                 self.state = QPackDecoderState::ReadInstruction;
                             } else {
                                 // waiting for more data
@@ -433,7 +429,6 @@ impl QPackDecoder {
                                 qdebug!([label], "received instruction - insert with name literal name={:x?} value={:x?}", name_to_insert, value_to_insert);
                                 self.table.insert(name_to_insert, value_to_insert)?;
                                 self.total_num_of_inserts += 1;
-                                self.increment += 1;
                                 self.state = QPackDecoderState::ReadInstruction;
                             } else {
                                 // waiting for more data
@@ -453,7 +448,6 @@ impl QPackDecoder {
                         qdebug!([label], "received instruction - duplicate index={}", index);
                         self.table.duplicate(*index)?;
                         self.total_num_of_inserts += 1;
-                        self.increment += 1;
                         self.state = QPackDecoderState::ReadInstruction;
                     } else {
                         // waiting for more data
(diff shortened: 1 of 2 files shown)


hawkinsw:
Thank you for your response! I really appreciate it!

I thought about the patch all day and came to the same conclusion. Thank
you for confirming!

I will rework and resend!

Thanks again!

Will

hawkinsw:
I *think* that this (at least more closely) adheres to the spec and addresses the issues that @martinthomson raised. Thanks again for the earlier review!

neqo-qpack/src/decoder.rs
```diff
@@ -553,7 +554,8 @@ impl QPackDecoder {
             if reader.done() {
                 // Send header_ack
                 if req_inserts != 0 {
-                    self.header_ack(stream_id);
+                    let acknowledged_inserts_count = req_inserts - base;
+                    self.header_ack(stream_id, acknowledged_inserts_count);
```
martinthomson: It would be good if you could also check that the highest index mentioned in an instruction matches the header~/code/interdependency-trust.

hawkinsw: Confirmed with @martinthomson that this is about making sure that a header that says it needs access to the Nth entry in a table actually references that entry. 

@martinthomson is going to investigate whether or not this is required by the spec or not.

ddragana:
I would like to have tests for this as well before landing.
I want all new code landing in http3/qpack to have tests as well.

ddragana:
@hawkinsw if you need help writing tests, just tell me.

hawkinsw:
Thanks for all the help @martinthomson and @ddragana on getting this whipped into shape. The updated PR has tests and I *think* that they test the functionality. Please let me know what you think! Thanks again!

ddragana:
instead of increment we will need to keep track index of acked inserts.

hawkinsw:
> instead of increment we will need to keep track index of acked inserts.

I do not understand why. Can you help?

The header acknowledgement, no matter when it happens, brings the Known Insert Count up to the Required Insert Count (when RIC is non-zero). Even though this is not the case now, assume that decoders (and not just encoders) used the `acked_inserts_cnt` of the table to keep track of KIC. That value could be compared with the number of entries in `self.base` to determine the number of unacknowledged but already processed insertions (call that U which is roughly equivalent to `self.insertions` at this point). When we call `header_ack`, we could adjust U such that len(decoder.table) = RIC + U. (decoder.table is updated immediately upon receiving an encoding instruction that inserts into the dynamic table, I think). Then, next time we emit an increment instruction, we use the updated value U as the increment value. Again, U is a mental experiment at this point and we would keep track of it differently in the code. This is far more complicated written out than it would be in code, I think. Would you mind if I took a stab at implementing this as part of an updated PR? It will hopefully make more sense in code form? 

As usual, thanks for thinking through this and helping me!
Will

ddragana:
> > instead of increment we will need to keep track index of acked inserts.
> 
> I do not understand why. Can you help?
> 
> The header acknowledgement, no matter when it happens, brings the Known Insert Count up to the Required Insert Count (when RIC is non-zero). Even though this is not the case now, assume that decoders (and not just encoders) used the `acked_inserts_cnt` of the table to keep track of KIC. That value could be compared with the number of entries in `self.base` to determine the number of unacknowledged but already processed insertions (call that U which is roughly equivalent to `self.insertions` at this point). When we call `header_ack`, we could adjust U such that len(decoder.table) = RIC + U. (decoder.table is updated immediately upon receiving an encoding instruction that inserts into the dynamic table, I think). Then, next time we emit an increment instruction, we use the updated value U as the increment value. Again, U is a mental experiment at this point and we would keep track of it differently in the code. This is far more complicated written out than it would be in code, I think. Would you mind if I took a stab at implementing this as part of an updated PR? It will hopefully make more sense in code form?
> 

I will look at the PR. Thanks.

hawkinsw:
@ddragana / @martinthomson I believe that this is ready for another round of review! I hope that we are getting closer. Sorry for the delay!

hawkinsw:
> The decoder side looks good, but we will need to fix the encoder side as well.
> Thanks for the tests.

Thank you for the quick review. I'm sorry to say that I don't quite understand what you mean by, "we will need to fix the encoder side as well." Do you foresee a similar set of issues on the encoding side as we saw on the decoding side? We already track the ack'd insert count on the encoder side which led me to assume that this would only be a problem on the receiver side. If you could be a little more specific, that would be great. Once I know what to look for, I can get started. 

Thank you!

ddragana:
> > The decoder side looks good, but we will need to fix the encoder side as well.
> > Thanks for the tests.
> 
> Thank you for the quick review. I'm sorry to say that I don't quite understand what you mean by, "we will need to fix the encoder side as well." Do you foresee a similar set of issues on the encoding side as we saw on the decoding side? We already track the ack'd insert count on the encoder side which led me to assume that this would only be a problem on the receiver side. If you could be a little more specific, that would be great. Once I know what to look for, I can get started.
> 
> Thank you!

we will need to increase acked inserts for header acks. we only increase it for an increment instruction but not for header acks.

hawkinsw:
> Let's land this.
> I will work on fixing the encoder side.

Thanks, @ddragana ! I am absolutely willing to do this work but I was just asking for clarification :-) I'll dig back in now that I have some additional context. Thank you!

PR #356, opened 2019-12-15
Toggle clippy::pedantic on neqo-crypto.

o0Ignition0o:
Part of #150, I was struggling trying to disable the default features (which doesn't do too well with build.rs) but given the number of warnings, we might as well clear them all instead :)

I don't know enough about the internals to be 100% confident Aead::expansion() will remain static, so please let me know if I should revert the change and allow unused self instead :)

### DIFF
--- a/neqo-crypto/src/aead.rs
+++ b/neqo-crypto/src/aead.rs
@@ -59,6 +59,7 @@ impl Aead {
         unsafe { Self::from_raw(version, cipher, s, prefix) }
     }
 
+    #[must_use]
     pub fn expansion(&self) -> usize {
         16
     }
--- a/neqo-crypto/src/time.rs
+++ b/neqo-crypto/src/time.rs
@@ -200,6 +200,10 @@ mod test {
     }
 
     #[test]
+    // We allow replace_consts here because
+    // std::u64::max_value() isn't available
+    // in all of our targets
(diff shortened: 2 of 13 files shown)


o0Ignition0o:
cannot find function `max_value` in module `std::u64`
It seems target related, I'm going to find out which one, and fix it :)

neqo-crypto/src/aead.rs
```diff
@@ -59,7 +59,8 @@ impl Aead {
         unsafe { Self::from_raw(version, cipher, s, prefix) }
     }
 
-    pub fn expansion(&self) -> usize {
+    #[must_use]
+    pub fn expansion() -> usize {
```
martinthomson: Removing `&self` here might be something that clippy wants, but I'm not sure that I do.

neqo-crypto/src/time.rs
```diff
@@ -200,6 +200,10 @@ mod test {
     }
 
     #[test]
+    // We allow replace_consts here beceuse
```
martinthomson:
```suggestion
    // We allow replace_consts here because
```

neqo-crypto/src/aead.rs
o0Ignition0o: Yeah I'm not really sure about that either 😁

PR #358, opened 2019-12-17
Implement server push, mostly the client side.

ddragana:


ddragana:
I will open a follow up bug about cancel states. To be able to detect that a push id was used twice we need to keep track of cancel states. The spec also defines a lot of SHOULD, e.g. server should not send CANCEL_PUSH if push stream has been open already, etc. Therefore we have a lot of cancel states. In the current patch some of them will stay in the push_stream list forever. In a follow up PR we should remove them after some time that will only cause use not to catch some duplicate uses of push ids.

neqo-client/src/main.rs
```diff
@@ -40,12 +40,20 @@ pub struct Args {
...
-    #[structopt(name = "max-table-size", short = "t", long, default_value = "128")]
-    max_table_size: u32,
+    #[structopt(name = "max-table-capacity", short = "t", long, default_value = "128")]
+    max_table_capacity: u32,
 
     #[structopt(name = "max-blocked-streams", short = "b", long, default_value = "128")]
     max_blocked_streams: u16,
 
+    #[structopt(
+        name = "max-concurent_push-streams",
```
martinthomson: Is the mingling of "-" and "_" intentional or a typo?

concurrent has two "r"s.

ddragana:
Closing this PR. There is a new PR #677

PR #366, opened 2019-12-19
Remove all events when a session is closed and fix some assertions.

ddragana:


ddragana:
This should fix bug 1601997.

Issue #372, opened 2019-12-22
Incorrect decoder capacity in decoder::tests::test_duplicate 


hawkinsw:
In the process of getting all the tests to pass for a patch for Issue #354, I ran into what _I think_ is a problem with the `test_duplicate` function in the QPACK decoder code. 

The encoded entry is 50 bytes. Duplicating that will insert another 50-byte entry into the table. That will put it over the 60-byte capacity set by the test. I think that the test passes currently because the decoder's header `table`'s `acked_insert_cnt` is `std::u64::MAX`. 

I am going to submit a PR in the hope that this is a real issue, but I know that you are all smarter than me (I? See, I don't even know proper English grammar) and I expect that you will tell me where I am wrong.

martinthomson:
Closed in 99bcef4.

PR #373, opened 2019-12-22
Issue #372: Properly set decoder capacity in decoder::test_duplicate

hawkinsw:
The duplicated entry is 50 bytes. This means that the table's capacity
has to be 100 bytes.


hawkinsw:
Thanks for confirming! It popped up when implementing #355.

PR #374, opened 2019-12-29
Toggle pedantic on neqo common

o0Ignition0o:
- Fixed new clippy (default) lints
- Fixed new pedantic lints in neqo-crypto
- Toggled pedantic in neqo-common

Please let me know if you would like me to squash the commits, and if there's anything I can do to improve the PR :)

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -2514,7 +2514,7 @@ mod tests {
         assert_eq!(0, client.stats().dups_rx);
 
         qdebug!("---- Dup, ignored");
-        let out = client.process(out_to_rep.dgram().clone(), now());
+        let out = client.process(out_to_rep.dgram(), now());
         assert!(out.as_dgram_ref().is_none());
         qdebug!("Output={:0x?}", out.as_dgram_ref());
 
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -523,7 +523,7 @@ mod tests {
         let flow_mgr = Rc::new(RefCell::new(FlowMgr::default()));
         let conn_events = ConnectionEvents::default();
 
-        let mut s = RecvStream::new(567.into(), 1024, flow_mgr.clone(), conn_events.clone());
+        let mut s = RecvStream::new(567.into(), 1024, Rc::clone(&flow_mgr), conn_events);
 
         // test receiving a contig frame and reading it works
         s.inbound_stream_frame(false, 0, vec![1; 10]).unwrap();
@@ -576,7 +576,7 @@ mod tests {
         let flow_mgr = Rc::new(RefCell::new(FlowMgr::default()));
         let conn_events = ConnectionEvents::default();
 
-        let mut s = RecvStream::new(3.into(), 1024, flow_mgr.clone(), conn_events.clone());
+        let mut s = RecvStream::new(3.into(), 1024, Rc::clone(&flow_mgr), conn_events);
 
         let mut buf = vec![0u8; 100];
 
@@ -670,8 +670,8 @@ mod tests {
         let mut s = RecvStream::new(
             4.into(),
             RX_STREAM_DATA_WINDOW,
-            flow_mgr.clone(),
-            conn_events.clone(),
+            Rc::clone(&flow_mgr),
+            conn_events,
         );
 
         let mut buf = vec![0u8; RX_STREAM_DATA_WINDOW as usize * 4]; // Make it overlarge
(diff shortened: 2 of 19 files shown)


neqo-crypto/src/aead.rs
```diff
@@ -53,6 +53,11 @@ pub struct Aead {
     ctx: AeadContext,
 }
 
+// This allows fn expansion() to take &self.
+// The directive is placed here because
+// we cannot stack directives.
+// Placing it on top of #[must_use] doesn't work.
+#[allow(clippy::unused_self)]
```
o0Ignition0o: It seems like this lint did not exist for the toolchain our CI uses.
Not really sure if/when we want to bump versions, so I'll remove this directive for now.
Let's keep in mind we will have to add it again once we update our clippy/rustc versions in the CI

neqo-transport/src/frame.rs
```diff
@@ -267,23 +267,26 @@ impl Frame {
         if offset > 0 {
             remaining = remaining.saturating_sub(Encoder::varint_len(offset));
         }
-        let (fin, fill) = if data.len() > remaining {
-            if remaining == 0 {
-                return None;
+        let (fin, fill) = match data.len().cmp(&remaining) {
```
martinthomson: This is going to conflict with #370.

neqo-transport/src/recv_stream.rs
```diff
@@ -523,7 +523,7 @@ mod tests {
         let flow_mgr = Rc::new(RefCell::new(FlowMgr::default()));
         let conn_events = ConnectionEvents::default();
 
-        let mut s = RecvStream::new(567.into(), 1024, flow_mgr.clone(), conn_events.clone());
+        let mut s = RecvStream::new(567.into(), 1024, flow_mgr, conn_events);
```
martinthomson: I'm not sure that I follow this change.  These are `Rc<>` instances that presumably we want to clone for passing into multiple invocations of this function.  Has `Rc` picked up an implementation of `Copy` in a recent version of rust?

I would have thought that moving to `Rc::clone(&flow_mgr)` would be wise, but I don't see that here either.

neqo-transport/src/connection.rs
```diff
@@ -2514,7 +2514,7 @@ mod tests {
         assert_eq!(0, client.stats().dups_rx);
 
         qdebug!("---- Dup, ignored");
-        let out = client.process(out_to_rep.dgram().clone(), now());
```
martinthomson: I totally get this one: the dgram() function returns an object directly, so cloning it is wasteful.  But see below.

neqo-transport/src/recv_stream.rs
o0Ignition0o: Oh nice catch, I'll move it to Rc::clone() to make it more explicit.
I think I got fooled by the linter, which advises to remove .clone(), which doesn't make sense in an Rc context.

It might be worth opening an issue to clippy.

neqo-transport/src/frame.rs
o0Ignition0o: Let's wait until #370 lands then so I can rebase :)

o0Ignition0o:
Rebased against master, now that #370 has merged

neqo-transport/src/send_stream.rs
```diff
@@ -912,7 +912,7 @@ mod tests {
         flow_mgr.borrow_mut().conn_increase_max_credit(4096);
         let conn_events = ConnectionEvents::default();
 
-        let mut s = SendStream::new(4.into(), 1024, flow_mgr.clone(), conn_events.clone());
+        let mut s = SendStream::new(4.into(), 1024, flow_mgr.clone(), conn_events);
```
```suggestion
        let mut s = SendStream::new(4.into(), 1024, Rc::clone(&flow_mgr), conn_events);
```

neqo-transport/src/send_stream.rs
```diff
@@ -1030,7 +1030,7 @@ mod tests {
         flow_mgr.borrow_mut().conn_increase_max_credit(2);
         let conn_events = ConnectionEvents::default();
 
-        let _s = SendStream::new(4.into(), 100, flow_mgr.clone(), conn_events.clone());
+        let _s = SendStream::new(4.into(), 100, flow_mgr, conn_events.clone());
```
```suggestion
        let _s = SendStream::new(4.into(), 100, flow_mgr, Rc::clone(&conn_events));
```

o0Ignition0o: I think we need to keep conn_events.clone() since ConnectionEvents is Clone, but it's not an Rc.

martinthomson: Of course, I misremembered.

PR #381, opened 2020-01-13
No Issue: Pad datagrams with Initial packets only if necessary

hawkinsw:
Although the packet-coalescing loop attempts to keep datagrams
under _path mtu_ bytes, it is possible that datagrams end up
slightly longer. When a client connection is sending a datagram in
that case and that datagram contains an `Initial` packet,
`resize()`ing the packet would actually truncate meaningful
content.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -1279,8 +1279,8 @@ impl Connection {
             self.path = Some(path);
             Ok(None)
         } else {
-            // Pad Initial packets sent by the client to mtu bytes.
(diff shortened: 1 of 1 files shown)


hawkinsw:
I somehow failed to see `remaining` as a parameter to all the successive calls to the various `get_frame` functions which those functions obviously use to make sure that only frames that will fit in the remaining space are returned. Therefore, what I thought was an issue is not, actually, an issue. 

Sorry for the spam. I was too eager to want to find a way to contribute code :-)

agrover:
Code review/audit always appreciated.

PR #383, opened 2020-01-13
Send ping only in the last epoch.

### LINKED ISSUE #382: Multiple ping frames in the same packet
If we need to send a ping because pto has expired and we do not have any other frames to send we will send a ping in each epoch.
A server will not ack packets(pings) for epoch 0 and 2 if we already have keys for epoch 3, therefore we will always have outstanding unacked packets even if a connection is idle.

ddragana:
fixes #382

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -1075,6 +1075,25 @@ impl Connection {
         out
     }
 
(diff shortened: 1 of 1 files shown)


ddragana:
This will be fix in a better way with #388

PR #388, opened 2020-01-15
Fix PTO and implement a separate PTO for each PN

### LINKED ISSUE #308: PTO per PN Space
We should track a PTO timer for each space and then make sure to send a packet in that space when the matching timer fires.  Otherwise, tickling the PTO timer in 0-RTT can cause the Initial to never fire and similar sorts of problems.

### LINKED ISSUE #385: a packet with ack frames after a pto ping packet
I have not debug it yet, but what to document:
pto expires we send a packet that contains a ping,  self.tx_mode is set to TxMode::Normal. if we call process_output one more time in some cases a packet with acks will be sent. this does not look right.

ddragana:
This patch fix a couple of problems with PTO (fixes #308 and #385):
This patch fix a couple of problems with PTO:
- Implement a separate PTO timer for each packet number space,
- Move loss_recovery_state: LossRecoveryState into loss_recovery,
- Keep track of a number of ack_eliciting_outstanding packets,
- If a PTO expires send packets in its pn space and above. For example, if the PTO timer in the Handshake pn space expires we will try to send a datagram with a Handshake packet and a packet in the app pn space if we have keys for app space (this will increase chances that the server can decrypt them and send an ack),
- Fix #385 by adding LossRecoveryState::PtoExpired.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -34,7 +34,7 @@ use crate::packet::{
     decode_packet_hdr, decrypt_packet_body, decrypt_packet_hdr, encode_packet, ConnectionId,
     ConnectionIdDecoder, HeaderProtectionMask, PacketHdr, PacketType,
 };
-use crate::recovery::{LossRecovery, LossRecoveryMode, LossRecoveryState, RecoveryToken};
+use crate::recovery::{LossRecovery, RecoveryToken};
 use crate::recv_stream::{RecvStream, RecvStreams, RX_STREAM_DATA_WINDOW};
 use crate::send_stream::{SendStream, SendStreams};
 use crate::stats::Stats;
@@ -44,7 +44,7 @@ use crate::tparams::{
 };
 use crate::tracking::{AckTracker, PNSpace, SentPacket};
 use crate::QUIC_VERSION;
-use crate::{AppError, ConnectionError, Error, Res};
+use crate::{AppError, ConnectionError, Error, Res, LOCAL_IDLE_TIMEOUT};
 
 #[derive(Debug, Default)]
 struct Packet(Vec<u8>);
@@ -53,8 +53,6 @@ pub const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
 pub const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
 
 const LOCAL_MAX_DATA: u64 = 0x3FFF_FFFF_FFFF_FFFF; // 2^62-1
-
-const LOCAL_IDLE_TIMEOUT: Duration = Duration::from_secs(60); // 1 minute
 
 #[derive(Debug, PartialEq, Copy, Clone)]
 /// Client or Server.
@@ -374,11 +372,9 @@ pub struct Connection {
     pub(crate) flow_mgr: Rc<RefCell<FlowMgr>>,
     state_signaling: StateSignaling,
     loss_recovery: LossRecovery,
-    loss_recovery_state: LossRecoveryState,
     events: ConnectionEvents,
     token: Option<Vec<u8>>,
     stats: Stats,
-    tx_mode: TxMode,
 }
 
 impl Debug for Connection {
@@ -501,11 +497,9 @@ impl Connection {
             flow_mgr: Rc::new(RefCell::new(FlowMgr::default())),
             state_signaling: StateSignaling::Idle,
             loss_recovery: LossRecovery::new(),
-            loss_recovery_state: LossRecoveryState::default(),
             events: ConnectionEvents::default(),
             token: None,
             stats: Stats::default(),
-            tx_mode: TxMode::Normal,
         }
     }
 
@@ -693,8 +687,8 @@ impl Connection {
             self.set_state(State::Closed(ConnectionError::Transport(
                 Error::IdleTimeout,
             )));
-        } else {
-            self.check_loss_detection_timeout(now);
+        } else if let Some(packets) = self.loss_recovery.check_loss_detection_timeout(now) {
+            self.handle_lost_packets(&packets);
         }
     }
 
@@ -718,11 +712,9 @@ impl Connection {
     /// Get the time that we next need to be called back, relative to `now`.
     fn next_delay(&mut self, now: Instant) -> Duration {
         qtrace!([self], "Get callback delay");
-        self.loss_recovery_state = self.loss_recovery.get_timer();
-
         let mut delays = SmallVec::<[_; 4]>::new();
 
-        if let Some(lr_time) = self.loss_recovery_state.callback_time() {
+        if let Some(lr_time) = self.loss_recovery.calculate_timer() {
             qtrace!([self], "Loss recovery timer {:?}", lr_time);
             delays.push(lr_time);
         }
@@ -1128,11 +1120,30 @@ impl Connection {
         let mut needs_padding = false;
         let mut close_sent = false;
 
+        // Check whether we are sending packets in PTO mode.
+        let (tx_mode, cong_avail, min_pn_space) =
+            if let Some((min_pto_pn_space, can_send)) = self.loss_recovery.get_pto_state() {
+                if !can_send {
+                    return Ok(None);
+                }
+                (TxMode::Pto, path.mtu(), min_pto_pn_space)
+            } else {
+                (
+                    TxMode::Normal,
+                    usize::try_from(self.loss_recovery.cwnd_avail()).unwrap(),
+                    PNSpace::Initial,
+                )
+            };
+
         // Frames for different epochs must go in different packets, but then these
         // packets can go in a single datagram
         for space in PNSpace::iter() {
             let mut encoder = Encoder::default();
             let mut tokens = Vec::new();
+
+            if *space < min_pn_space {
+                continue;
+            }
 
             // Ensure we have tx crypto state for this epoch, or skip it.
             let tx = if let Some(tx_state) = self.crypto.states.tx(*space) {
@@ -1168,11 +1179,6 @@ impl Connection {
 
             let mut ack_eliciting = false;
             let mut has_padding = false;
-            let cong_avail = match self.tx_mode {
-                TxMode::Normal => usize::try_from(self.loss_recovery.cwnd_avail()).unwrap(),
-                TxMode::Pto => path.mtu(), // send one packet
-            };
-            let tx_mode = self.tx_mode;
 
             match &self.state {
                 State::Init
@@ -1195,7 +1201,7 @@ impl Connection {
 
                         // Try to get a frame from frame sources
                         let mut frame = None;
-                        if self.tx_mode == TxMode::Normal {
+                        if tx_mode == TxMode::Normal {
                             frame = self.acks.get_frame(now, *space);
                         }
                         if frame.is_none()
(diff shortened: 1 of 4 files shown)


ddragana:
I need to change something.

neqo-transport/src/connection.rs
```diff
@@ -1173,7 +1191,11 @@ impl Connection {
                         if frame.is_none() {
                             frame = self.send_streams.get_frame(epoch, tx_mode, remaining)
                         }
-                        if frame.is_none() && self.tx_mode == TxMode::Pto {
+
+                        if frame.is_none()
+                            && matches!(self.tx_mode, TxMode::Pto{..})
+                            && self.should_send_ping(epoch)
```
agrover: Now that we have a whole method to determine whether to send a ping, suggest moving the matches inside it as well.

neqo-transport/src/connection.rs
```diff
@@ -1275,25 +1300,49 @@ impl Connection {
...
+            // we will limit it on the client side. We will not send a PING if we have received a 1RTT
+            // ack. TODO: fix this when we implement HANDSHAKE_DONE
+            2 => {
+                if self.role == Role::Client {
+                    if let Some(l) = self
+                        .loss_recovery
+                        .largest_acknowledged_pn(PNSpace::ApplicationData)
+                    {
+                        if let Some(f) = self.first_1_rtt_pn {
+                            if l > f {
```
agrover: little more verbose on the variable names for l and f?

neqo-transport/src/recovery.rs
```diff
@@ -494,9 +505,18 @@ impl LossRecovery {
...
+                .values()
+                .any(|sp| sp.ack_eliciting),
+            self.spaces[PNSpace::Handshake]
+                .sent_packets
+                .values()
+                .any(|sp| sp.ack_eliciting),
+            self.spaces[PNSpace::ApplicationData]
+                .sent_packets
+                .values()
+                .any(|sp| sp.ack_eliciting)
```
agrover: how about outputting both number of packets AND if any are ack-eliciting in this qinfo?

neqo-transport/src/connection.rs
```diff
@@ -1261,6 +1279,10 @@ impl Connection {
                 SentPacket::new(now, ack_eliciting, tokens, packet.len(), in_flight),
             );
 
+            if self.first_1_rtt_pn.is_none() && epoch == 3 {
+                self.first_1_rtt_pn = Some(hdr.pn);
+            }
```
martinthomson: In the key update changes, packet numbers are already tracked by space.  Maybe you could use that information instead.

neqo-transport/src/recovery.rs
```diff
@@ -543,6 +560,53 @@ impl LossRecovery {
...
+            .min_by_key(|&time| time)
+    }
+
+    pub fn get_min_pto_pn_space(&self, now: Instant) -> Option<PNSpace> {
+        PNSpace::iter()
+            .filter_map(|spc| {
+                if self.spaces[*spc]
+                    .sent_packets
+                    .values()
+                    .any(|sp| sp.ack_eliciting)
```
martinthomson: Can we avoid doing this iteration on every PTO by recording this information on the object when we send a packet?  We could clear it when we process acknowledgments.

Which do you think would be easier to manage?

I you keep it this way around, I would prefer to add a function on LossRecoverySpace that performs this check.  `ack_eliciting_outstanding() -> bool` would help a lot.

neqo-transport/src/recovery.rs
```diff
@@ -543,6 +560,53 @@ impl LossRecovery {
...
+                            None
+                        }
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .min_by_key(|&spc| spc)
```
martinthomson: This might read better if you factor out the function that produces the iterator, so that these functions only do `min_by_key()` with different keys.

neqo-transport/src/recovery.rs
```diff
@@ -493,10 +504,22 @@ impl LossRecovery {
...
-            self.spaces[PNSpace::ApplicationData].sent_packets.len()
+            self.spaces[PNSpace::Handshake]
+                .sent_packets
+                .values()
+                .any(|sp| sp.ack_eliciting),
+            self.spaces[PNSpace::ApplicationData].sent_packets.len(),
+            self.spaces[PNSpace::ApplicationData]
+                .sent_packets
+                .values()
+                .any(|sp| sp.ack_eliciting)
```
martinthomson: See below about factoring this out into an accessor (or maybe a property).

neqo-transport/src/recovery.rs
```diff
@@ -288,7 +293,8 @@ impl LossRecovery {
     ) {
         qdebug!([self], "packet {:?}-{} sent.", pn_space, packet_number);
         if sent_packet.ack_eliciting {
-            self.time_of_last_sent_ack_eliciting_packet = Some(sent_packet.time_sent);
+            self.spaces[pn_space].time_of_last_sent_ack_eliciting_packet =
```
martinthomson: Rather than accessing the field directly, a setter seems appropriate.  That might help with my suggestion below too.

neqo-transport/src/connection.rs
```diff
@@ -1275,25 +1297,53 @@ impl Connection {
...
             Ok(Some(Datagram::new(path.local, path.remote, out_bytes)))
         }
     }
 
+    fn should_send_ping(&self, epoch: Epoch) -> bool {
+        if !matches!(self.tx_mode, TxMode::Pto{..}) {
+            return false;
+        }
+
+        match epoch {
```
martinthomson: This should match by PNSpace.

neqo-transport/src/connection.rs
```diff
@@ -1275,25 +1297,53 @@ impl Connection {
...
+            // If epoch is initial and we have keys we should send a PING.
+            0 => true,
+            // Currently we do not drop handshake keys, but to avoid sending PINGs for a long time
+            // we will limit it on the client side. We will not send a PING if we have received a 1RTT
+            // ack. TODO: fix this when we implement HANDSHAKE_DONE
```
martinthomson: That's coming; do you want to drop this now?

neqo-transport/src/connection.rs
```diff
@@ -1137,7 +1149,13 @@ impl Connection {
             let mut has_padding = false;
             let cong_avail = match self.tx_mode {
                 TxMode::Normal => usize::try_from(self.loss_recovery.cwnd_avail()).unwrap(),
-                TxMode::Pto => path.mtu(), // send one packet
+                TxMode::Pto { dgram_avail, .. } => {
```
martinthomson: Why are we saving this information on the connection?  Can't the loss recovery spaces hold their PTO state?

Let's say that we send PN 3 in Handshake, can't we ask the loss recovery whether it is time to PTO in that space when it comes time to send a packet in that space?

neqo-transport/src/connection.rs
```diff
@@ -1275,25 +1297,53 @@ impl Connection {
...
             Ok(None)
         } else {
             // Pad Initial packets sent by the client to mtu bytes.
             if self.role == Role::Client && needs_padding {
                 qdebug!([self], "pad Initial to max_datagram_size");
                 out_bytes.resize(path.mtu(), 0);
             }
+            if let TxMode::Pto { dgram_avail, .. } = &mut self.tx_mode {
+                assert!(*dgram_avail > 0);
+                *dgram_avail -= 1;
```
martinthomson: I'm not sure that I understand the use of dgram_avail.  Can you comment this with some explanations.

neqo-transport/src/connection.rs
ddragana: I will drop this and use HANDSHAKE_DONE and keys discarding. 

I wanted to land this code before we switch version to 25 and turn it on on Nightly (this problem is a blocker because it make use keep connection alive forever). The version switch PR landed so that is late not.... I will update mozilla-central when version 25 is good enough for interoperability (HANDSHAKE_DONE is blocking one)

neqo-transport/src/recovery.rs
ddragana: this is call only if PTO timer expires (or if app calls before PTO timer expires and pto time is set - if app is doing it right it should not happe). I do not think wee will do PTO a lot so this function is not a concern, but I think there is one other place that is problematic (in the code that this PR have not touched). I will look into fixing it.

neqo-transport/src/connection.rs
ddragana: I am not sure either whether we need tx_mode. I think this can be done in a different way.
Also our way of calling 'process' multiple times does not fit nicely with PTO for example. 
dgram_avail is number of packet available to be sent. it should be 1 or 2, but setting it to 2 doesn't work with the way we build frames and packet. I will write a long comment in the e-mail thread you have started.

neqo-transport/src/connection.rs
ddragana: I will refactor this and remove tx_mode from connection.

neqo-transport/src/connection.rs
```diff
@@ -718,11 +712,10 @@ impl Connection {
     /// Get the time that we next need to be called back, relative to `now`.
     fn next_delay(&mut self, now: Instant) -> Duration {
         qtrace!([self], "Get callback delay");
-        self.loss_recovery_state = self.loss_recovery.get_timer();
-
         let mut delays = SmallVec::<[_; 4]>::new();
+        self.loss_recovery.calculate_timer();
```
martinthomson: Why not return the time from this function?

neqo-transport/src/connection.rs
```diff
@@ -1128,12 +1121,31 @@ impl Connection {
         let mut needs_padding = false;
         let mut close_sent = false;
 
+        // Check whether we are sending packets in PTO mode.
+        let (tx_mode, cong_avail, min_pn_space) =
+            if let Some((min_pn, can_send)) = self.loss_recovery.get_pto_state() {
+                if !can_send {
+                    return Ok(None);
+                }
+                (TxMode::Pto, path.mtu(), min_pn)
```
martinthomson: Calling this min_pn is misleading.

neqo-transport/src/recovery.rs
```diff
@@ -356,6 +400,17 @@ impl LossRecovery {
         max(rtt * 9 / 8, GRANULARITY)
     }
 
+    // Calculate PTO duration
+    fn pto_timeout(&self, pn_space: PNSpace) -> Duration {
+        if let Some(d) = self.rtt_vals.pto(pn_space).checked_mul(1 << self.pto_count) {
```
martinthomson:
```suggestion
        self.rtt_vals.pto(pn_space).checked_mul(1 << self.pto_count).unwrap_or(LOCAL_IDLE_TIMEOUT*2)
```

neqo-transport/src/recovery.rs
```diff
@@ -533,6 +588,96 @@ impl LossRecovery {
...
+            })
+            .min_by_key(|&time| time)
+    }
+
+    pub fn get_min_pto_pn_space(&self, now: Instant) -> Option<PNSpace> {
+        PNSpace::iter()
+            .filter_map(|spc| {
+                if self.spaces[*spc].ack_eliciting_outstanding() {
+                    if let Some(time) = self.spaces[*spc].time_of_last_sent_ack_eliciting_packet {
+                        if time <= (now - self.pto_timeout(*spc)) {
```
martinthomson: I would use `time + self.pto_timeout(*spc) <= now` to avoid weird underflow bugs.  Instant is weird.

neqo-transport/src/recovery.rs
martinthomson: I can see the difference now. This is good, but I would prefer a little cleaner factoring.

neqo-transport/src/recovery.rs
```diff
@@ -533,6 +588,96 @@ impl LossRecovery {
...
+            return None;
+        }
+
+        if self.callback_time() > Some(now) {
+            // LR timer, but hasn't expired.
+            return None;
+        }
+
+        // Timer expired and LR was active timer.
+        match &mut self.loss_recovery_state.mode() {
```
martinthomson: You don't need a mutable reference.

neqo-transport/src/recovery.rs
```diff
@@ -533,6 +588,96 @@ impl LossRecovery {
...
+                    None
+                }
+            })
+            .min_by_key(|&spc| spc)
+    }
+
+    pub fn check_loss_detection_timeout(&mut self, now: Instant) -> Option<Vec<SentPacket>> {
+        qdebug!([self], "check_loss_timeouts");
+
+        if matches!(self.loss_recovery_state.mode(), LossRecoveryMode::None) {
```
martinthomson: A simple comparison should work.

PR #390, opened 2020-01-16
No Issue: Pad datagrams containing Initial packets from a client with Padding Frames.

hawkinsw:
NB: This is WIP. There are tests which will fail to pass as a result of this change. I will fix the tests if you decide that this is something worthwhile. 

As usual, please be gentle. I have made every effort to supply something that I think is worthwhile. 

The goal is to pad datagrams containing an Initial packet from a client with a subsequent, coalesced packet full of nothing but Padding frames.

With this change, our streams are decodable in Wireshark and we are set to comply more fully with the conclusion to the discussion in https://github.com/quicwg/base-drafts/issues/3333.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -1075,6 +1075,46 @@ impl Connection {
         out
     }
 
+    /// Build a long-header packet with nothing but PADDING frames to fill 
+    /// _size_ bytes and use the "Initial secret" encryption level to protect.
+    fn get_padding_packet(
+        &mut self,
+        size: usize,
+        path: &Path,
+    ) -> Vec<u8> {
+        let initial_epoch = 0;
+        let mut encoder = Encoder::default();
+        let padding_frame = Frame::Padding;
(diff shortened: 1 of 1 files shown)


hawkinsw:
> The problem with this approach is that it doesn't allow us to pad if we come too close to the MTU. We can over-pad, but that risks getting eaten by the path. If we are going to pad inside packets, then let's just add PADDING frames to the Initial.
> 
> The best way I have come up with is to encode the frames for the initial packet, take the space away based on calculating how big that would be, but then don't encrypt it until at the end. Then you can stuff the plaintext of the packet with zeros as you see fit.

Thank you for the review. The best part about your comment: I was already thinking about doing this! In fact, I completely implemented it and everything looked great. I ultimately submitted *this* other approach (that you reviewed) because, well, I don't know. I will go back to the first idea (the one that matches your suggestion here) and then I will submit another PR. 

Thanks again for the review!
Will

hawkinsw:
Closing in favor of a new PR.

PR #392, opened 2020-01-16
Update bindgen

martinthomson:
This updates to the latest, which doesn't have a bug that had to be
worked around.

As part of this, I discovered that the docker image needs to include
libclang-dev due to new dependencies.

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -169,8 +169,9 @@ fn dynamic_link_both(extra_libs: &[&str]) {
 
 fn static_link(nsstarget: &PathBuf) {
     let lib_dir = nsstarget.join("lib");
+    // Include a RPATH directive to make debugging easier.
     println!(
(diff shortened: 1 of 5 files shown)


martinthomson:
This is failing because it requires an updated docker image.  But I can't enable that because of the nasty `clippy::unused_self` bug.  Gah.

neqo-crypto/build.rs
```diff
@@ -169,8 +169,9 @@ fn dynamic_link_both(extra_libs: &[&str]) {
 
 fn static_link(nsstarget: &PathBuf) {
     let lib_dir = nsstarget.join("lib");
+    // Include a RPATH directive to make debugging easier.
     println!(
-        "cargo:rustc-link-search=native={}",
+        "cargo:rustc-cdylib-link-arg=-Wl,-rpath,{}",
```
martinthomson: Note that I don't think that this works.  But the previous line was a duplicate.  So I should just remove this.

PR #395, opened 2020-01-16
Implement dropping of Initial keys

ddragana:
CryptoState has 3 states: NotInit, Active and Discarded
Crypto streams are moved into CryptoState

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -784,6 +784,14 @@ impl Connection {
         Ok(())
     }
 
+    fn discard_keys(&mut self, pn_space: PNSpace) {
+        if self.crypto.is_discarded(pn_space) {
+            return;
+        }
+        self.loss_recovery.remove_state_for_pn_space(pn_space);
+        self.crypto.discard(pn_space);
+    }
+
     fn input(&mut self, d: Datagram, now: Instant) -> Res<Vec<(Frame, Epoch)>> {
         let mut slc = &d[..];
         let mut frames = Vec::new();
@@ -802,6 +810,7 @@ impl Connection {
                         hex(slc),
                         e
                     );
+                    self.stats.dropped_rx += 1;
                     return Ok(frames); // Drop the remainder of the datagram.
                 }
             };
@@ -819,6 +828,7 @@ impl Connection {
                 }
                 (PacketType::VN(_), ..) | (PacketType::Retry { .. }, ..) => {
                     qwarn!("dropping {:?}", hdr.tipe);
+                    self.stats.dropped_rx += 1;
                     return Ok(frames);
                 }
                 _ => {}
@@ -831,6 +841,7 @@ impl Connection {
                         hdr.version.unwrap(),
                         self.version,
                     );
+                    self.stats.dropped_rx += 1;
                     return Ok(frames);
                 }
             }
@@ -838,12 +849,14 @@ impl Connection {
             match self.state {
                 State::Init => {
                     qinfo!([self], "Received message while in Init state");
+                    self.stats.dropped_rx += 1;
                     return Ok(frames);
                 }
                 State::WaitInitial => {
                     qinfo!([self], "Received packet in WaitInitial");
                     if self.role == Role::Server {
                         if !self.is_valid_initial(&hdr) {
+                            self.stats.dropped_rx += 1;
                             return Ok(frames);
                         }
                         self.crypto.create_initial_state(self.role, &hdr.dcid);
@@ -852,7 +865,12 @@ impl Connection {
                 State::Handshaking | State::Connected => {
                     if !self.is_valid_cid(&hdr.dcid) {
                         qinfo!([self], "Ignoring packet with CID {:?}", hdr.dcid);
+                        self.stats.dropped_rx += 1;
                         return Ok(frames);
+                    }
+                    if self.role == Role::Server && matches!(hdr.tipe, PacketType::Handshake {..}) {
+                        // Server has received a Handshake packet -> discard Initial keys and states
+                        self.discard_keys(PNSpace::Initial);
                     }
                 }
                 State::Closing { .. } => {
@@ -863,6 +881,7 @@ impl Connection {
                 }
                 State::Closed(..) => {
                     // Do nothing.
+                    self.stats.dropped_rx += 1;
                     return Ok(frames);
                 }
             }
@@ -883,6 +902,8 @@ impl Connection {
                     self.start_handshake(hdr, &d)?;
                 }
                 self.process_migrations(&d)?;
+            } else {
+                self.stats.dropped_rx += 1;
             }
         }
         Ok(frames)
@@ -896,12 +917,8 @@ impl Connection {
             // but we still need to discharge the packet.
             None
         } else if epoch != 1 {
-            match self
-                .crypto
-                .states
-                .obtain(self.role, epoch, &self.crypto.tls)
-            {
-                Ok(CryptoState { rx, .. }) => rx.as_mut(),
+            match self.crypto.obtain(self.role, epoch) {
+                Ok(CryptoState::Active { rx, .. }) => rx.as_mut(),
                 _ => None,
             }
         } else if self.role == Role::Server {
@@ -1075,6 +1092,22 @@ impl Connection {
         out
     }
 
+    fn obtain_epoch_tx_crypto_state(&mut self, epoch: Epoch) -> Res<&mut CryptoDxState> {
+        if epoch == 1 && self.role == Role::Server {
+            Err(Error::KeysNotFound)
+        } else if epoch == 1 {
+            match &mut self.zero_rtt_state {
+                ZeroRttState::Sending(tx) => Ok(tx),
+                _ => Err(Error::KeysNotFound),
+            }
+        } else {
+            match self.crypto.obtain(self.role, epoch) {
+                Ok(CryptoState::Active { tx: Some(tx), .. }) => Ok(tx),
+                _ => Err(Error::KeysNotFound),
+            }
+        }
+    }
+
     #[allow(clippy::cognitive_complexity)]
     #[allow(clippy::useless_let_if_seq)]
     /// Build a datagram, possibly from multiple packets (for different PN
@@ -1096,23 +1129,9 @@ impl Connection {
             let mut tokens = Vec::new();
 
             // Ensure we have tx crypto state for this epoch, or skip it.
-            let tx = if epoch == 1 && self.role == Role::Server {
+            if self.obtain_epoch_tx_crypto_state(epoch).is_err() {
                 continue;
-            } else if epoch == 1 {
-                match &mut self.zero_rtt_state {
-                    ZeroRttState::Sending(tx) => tx,
-                    _ => continue,
-                }
-            } else {
-                match self
-                    .crypto
-                    .states
-                    .obtain(self.role, epoch, &self.crypto.tls)
-                {
-                    Ok(CryptoState { tx: Some(tx), .. }) => tx,
-                    _ => continue,
-                }
-            };
+            }
 
             let hdr = PacketHdr::new(
                 0,
@@ -1147,8 +1166,12 @@ impl Connection {
             match &self.state {
                 State::Init | State::WaitInitial | State::Handshaking | State::Connected => {
                     loop {
-                        let used =
-                            out_bytes.len() + encoder.len() + hdr.overhead(&tx.aead, path.mtu());
+                        let used = out_bytes.len()
+                            + encoder.len()
+                            + hdr.overhead(
+                                &self.obtain_epoch_tx_crypto_state(epoch)?.aead,
+                                path.mtu(),
+                            );
                         let remaining = min(
                             path.mtu().saturating_sub(used),
                             cong_avail.saturating_sub(used),
@@ -1242,7 +1265,8 @@ impl Connection {
             self.stats.packets_tx += 1;
             self.loss_recovery.inc_pn(space);
 
-            let mut packet = encode_packet(tx, &hdr, &encoder);
+            let mut packet =
+                encode_packet(self.obtain_epoch_tx_crypto_state(epoch)?, &hdr, &encoder);
 
             if self.tx_mode != TxMode::Pto && ack_eliciting {
                 self.idle_timeout.on_packet_sent(now);
@@ -1262,6 +1286,11 @@ impl Connection {
             dump_packet(self, "TX ->", &hdr, &encoder);
 
             out_bytes.append(&mut packet);
+
+            if self.role == Role::Client && space == PNSpace::Handshake {
+                // Client can send Handshake packets -> discard Initial keys and states
+                self.discard_keys(PNSpace::Initial);
+            }
         }
 
         if close_sent {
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection.rs
```diff
@@ -784,6 +784,17 @@ impl Connection {
...
     }
 
+    fn discarde_keys(&mut self, pn_space: PNSpace) {
+        debug_assert!((pn_space == PNSpace::Initial) || (pn_space == PNSpace::Handshake));
+        let epoch = if pn_space == PNSpace::Initial { 0 } else { 2 };
+        if self.crypto.is_discarded(epoch) {
+            return;
+        }
+        let lost_packets = self.loss_recovery.remove_state_for_pn_space(pn_space);
+        self.handle_lost_packets(&lost_packets);
```
martinthomson: When things are dropped, we don't consider the packets to be lost (because that drives congestion windows down).  All we have to do is remove them from bytes in flight.  For that, I think that the loss_recovery code can deal with that.

neqo-transport/src/connection.rs
```diff
@@ -784,6 +784,17 @@ impl Connection {
         Ok(())
     }
 
+    fn discarde_keys(&mut self, pn_space: PNSpace) {
```
martinthomson: "discard"

neqo-transport/src/connection.rs
```diff
@@ -1075,6 +1090,26 @@ impl Connection {
         out
     }
 
+    fn obtain_epoch_tx_crypto_state(&mut self, epoch: Epoch) -> Res<&mut CryptoDxState> {
```
martinthomson: I have a replacement for this in my patch, so as long as this works, I won't review it.

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3737,70 @@ mod tests {
...
+        assert!(init_pkt_s.as_dgram_ref().is_some());
+        assert_eq!(1, server.stats().packets_rx);
+        assert_eq!(0, server.stats().dups_rx);
+
+        qdebug!("---- client: cert verification");
+        let out = client.process(Some(init_pkt_s.as_dgram_ref().unwrap().clone()), now());
+        assert!(out.as_dgram_ref().is_some());
+
+        assert_eq!(2, client.stats().packets_rx);
+        assert_eq!(0, client.stats().dups_rx);
```
martinthomson: Should we be recording drops as well as dups?  That might make this easier to test.

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3737,70 @@ mod tests {
...
+
+        // Server will process the first Handshake packet. After this the Initial keys will be dropped.
+        let out = server.process(out.dgram(), now());
+        assert!(out.as_dgram_ref().is_some());
+
+        assert_eq!(3, server.stats().packets_rx);
+        assert_eq!(1, server.stats().dups_rx);
+
+        // Check that the Initial keys are dropped at server
+        // We will check this by processing init_pkt_c the third time.
```
```suggestion
        // We will check this by processing init_pkt_c a third time.
```

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3737,70 @@ mod tests {
...
+        assert!(out.as_dgram_ref().is_none());
+
+        assert_eq!(4, client.stats().packets_rx);
+        assert_eq!(1, client.stats().dups_rx);
+
+        assert!(maybe_authenticate(&mut client));
+
+        // Server has not removed Initial keys yet, because it has not received a Handshake packet
+        // from client.
+        // We will check this by processing init_pkt_c the second time.
```
```suggestion
        // We will check this by processing init_pkt_c a second time.
```

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3737,70 @@ mod tests {
...
+        assert_eq!(0, server.stats().dups_rx);
+
+        qdebug!("---- client: cert verification");
+        let out = client.process(Some(init_pkt_s.as_dgram_ref().unwrap().clone()), now());
+        assert!(out.as_dgram_ref().is_some());
+
+        assert_eq!(2, client.stats().packets_rx);
+        assert_eq!(0, client.stats().dups_rx);
+
+        // Client has received handskae packet it will remove Initial keys.
```
```suggestion
        // Client has received handhake packet, so it will remove Initial keys.
```

neqo-transport/src/connection.rs
```diff
@@ -1262,6 +1288,11 @@ impl Connection {
             dump_packet(self, "TX ->", &hdr, &encoder);
 
             out_bytes.append(&mut packet);
+
+            if self.role == Role::Client && epoch == 2 {
```
martinthomson: You can test the packet number space here.

ddragana:
> The management of the crypto states is something that duplicates what I've been working on, but I think that I can work with what you have here. For me, the important changes are the changes in recovery, which seem broadly OK.
> 
> I think that segmenting by epoch isn't a good fit here and my changes segment by PNSpace instead. As much as possible, I would ask that you test PNSpace rather than epoch where possible. I've noted a couple of places.

I was expecting that you will change epoch to PNSpace, therefore I dropped my changes, but I can add them again.
 
> 
> I also think that moving the streams into the crypto state is not the way to go here. Instead, I would prefer that the top-level Crypto object get the instruction to dispense with PNSpace X and then it can decide how to route that request to both the streams piece and the packet protection piece.

Ok, I will add streams back to Crypto object.

neqo-transport/src/connection.rs
ddragana: This one is only responsible for marking frames(data in frames) lost and mark them to be retransmitted. This does not change congestion control. While writing this I just realize that I do not need to call this function at all, this data will not be retransmitted.

Looking a bit, I just realize that retry also do not handle bytes in flight properly (my code just use the same function as the retry code and they both are wrong). I will fix both in this PR.

ddragana:
I did not make a lot of changes to Crypto, You will rebase your patch. I had a quick look at your patch, that was what I was planning to do next, something along these lines.

neqo-transport/src/connection.rs
```diff
@@ -852,8 +877,13 @@ impl Connection {
                 State::Handshaking | State::Connected => {
                     if !self.is_valid_cid(&hdr.dcid) {
                         qinfo!([self], "Ignoring packet with CID {:?}", hdr.dcid);
+                        self.stats.dropped_rx += 1;
                         return Ok(frames);
                     }
+                    if self.role == Role::Server && matches!(hdr.tipe, PacketType::Handshake {}) {
```
```suggestion
                    if self.role == Role::Server && matches!(hdr.tipe, PacketType::Handshake {..}) {
```

Don't you need `{..}` to match a complex enum arm?

neqo-transport/src/connection.rs
```diff
@@ -1259,9 +1295,17 @@ impl Connection {
                 SentPacket::new(now, ack_eliciting, tokens, packet.len(), in_flight),
             );
 
+            if self.first_1_rtt_pn == None && epoch == 3 {
```
martinthomson: I know that we aren't running clippy here in pedantic mode, but...

```suggestion
            if self.first_1_rtt_pn.is_none() && epoch == 3 {
```

neqo-transport/src/connection.rs
```diff
@@ -1374,7 +1418,16 @@ impl Connection {
                     _ => Error::CryptoError(e),
                 });
             }
-            Ok(msgs) => self.crypto.buffer_records(msgs),
+            Ok(msgs) => {
+                // Make sure we have build handshake state.
```
```suggestion
                // Make sure we have built handshake keys.
```

neqo-transport/src/connection.rs
```diff
@@ -1374,7 +1418,16 @@ impl Connection {
...
+                if matches!(self.crypto.states[2], CryptoState::NoInit) {
+                    match self.crypto.try_to_get_crypto_state(self.role, 2) {
```
```suggestion
                    if self.crypto.try_to_get_crypto_state(self.role, 2).is_err() {
```

neqo-transport/src/connection.rs
```diff
@@ -1641,6 +1694,9 @@ impl Connection {
             }
         }
         self.handle_lost_packets(&lost_packets);
+        if epoch == 3 {
+            self.maybe_discard_handshake_keys(largest_acknowledged);
```
martinthomson: I don't think that we should be doing handshake drop until we have HANDSHAKE_DONE.  Otherwise we're just going to have to redo this.

neqo-transport/src/recovery.rs
```diff
@@ -210,7 +210,6 @@ impl LossRecoverySpace {
     fn remove_ignored(&mut self) -> impl Iterator<Item = SentPacket> {
         // The largest acknowledged or loss_time should still be unset.
         // The client should not have received any ACK frames when it drops 0-RTT.
-        assert!(self.largest_acked.is_none());
```
martinthomson: I think that we can assert that either this is true or the PNSpace != PNSpace::ApplicationData

neqo-transport/src/recovery.rs
```diff
@@ -357,6 +356,14 @@ impl CongestionControl {
         }
     }
 
+    fn remove_ignore(&mut self, ignore_packets: &[SentPacket]) {
```
martinthomson:
```suggestion
    fn remove_ignored(&mut self, ignore_packets: &[SentPacket]) {
```

But I think that we should name this "discard_all()" instead, consistent with your other changes.

```suggestion
    fn discard_all(&mut self, ignore_packets: &[SentPacket]) {
```

neqo-transport/src/recovery.rs
```diff
@@ -464,8 +471,12 @@ impl LossRecovery {
         self.rtt_vals.pto()
     }
 
-    pub fn drop_0rtt(&mut self) -> impl Iterator<Item = SentPacket> {
-        self.spaces[PNSpace::ApplicationData].remove_ignored()
+    pub fn drop_0rtt(&mut self) -> Vec<SentPacket> {
+        let pkts: Vec<SentPacket> = self.spaces[PNSpace::ApplicationData]
```
martinthomson: This demonstrates a problem with the signature of the `cc.remove_ignore()` function.  That should instead take `T: IntoIter<SentPacket>` so that you don't have to create a temporary `Vec` to hold the discarded packets.

neqo-transport/src/crypto.rs
```diff
@@ -304,44 +346,76 @@ pub struct CryptoStream {
 
 #[derive(Debug, Default)]
 pub struct CryptoStreams {
-    streams: [CryptoStream; 4],
+    streams: [Option<CryptoStream>; 4],
```
martinthomson: How about this instead:

```rust
pub enum CryptoStreams {
  Init{
   initial: CryptoStream,
   handshake: CryptoStream,
   application: CryptoStream,
  },
  Handshake {
     handshake: CryptoStream,
     application: CryptoStream,
  },
  Application { application: CryptoStream },
}
```

That said, I'm OK with this for the moment.

neqo-transport/src/crypto.rs
```diff
@@ -230,69 +322,19 @@ impl std::fmt::Display for CryptoDxState {
...
-        write!(f, "CryptoStates")
-    }
+#[derive(Debug)]
+pub enum CryptoState {
+    NoInit,
+    Active {
+        tx: Option<CryptoDxState>,
+        rx: Option<CryptoDxState>,
+    },
+    Discarged,
```
martinthomson: typo: Discarded
```suggestion
    Discarded,
```

neqo-transport/src/connection.rs
```diff
@@ -3150,9 +3193,8 @@ mod tests {
 
         let frames = server.test_process_input(out.dgram().unwrap(), now + Duration::from_secs(11));
 
-        assert_eq!(frames[0], (Frame::Ping, 0));
```
martinthomson: 👍

neqo-transport/src/connection.rs
```diff
@@ -2518,9 +2560,10 @@ mod tests {
         assert!(out.as_dgram_ref().is_none());
         qdebug!("Output={:0x?}", out.as_dgram_ref());
 
-        // Four packets total received, two of them are dups
+        // Four packets total received, 1 of them are dups and one was dropped because Initial keys
+        // are dropped.
         assert_eq!(4, client.stats().packets_rx);
-        assert_eq!(2, client.stats().dups_rx);
+        assert_eq!(1, client.stats().dups_rx);
```
martinthomson: You need to count the drops here.

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3734,181 @@ mod tests {
         let c_tx_dgrams = send_bytes(&mut client, 0, now);
         assert_eq!(c_tx_dgrams.len(), 4);
     }
+
+    fn check_is_packet_drop(peer: &mut Connection, pkt: Option<Datagram>, dropped: u64, dups: u64) {
```
martinthomson: It looks like this is used for both duplicates or drops, so call this `check_discarded()`.

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3734,181 @@ mod tests {
...
+        let dropped_before = peer.stats.dropped_rx;
+        let dups_before = peer.stats.dups_rx;
+        let out = peer.process(pkt, now());
+        assert!(out.as_dgram_ref().is_none());
+        assert_eq!(dropped, peer.stats.dropped_rx - dropped_before);
+        assert_eq!(dups, peer.stats.dups_rx - dups_before);
+    }
+
+    #[test]
+    fn test_discarded_initial_keys() {
```
martinthomson: test_ is redundant in these names.

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3734,181 @@ mod tests {
...
+        qdebug!("---- client: generate CH");
+        let mut client = default_client();
+        let init_pkt_c = client.process(None, now());
+        assert!(init_pkt_c.as_dgram_ref().is_some());
```
martinthomson: Try this instead:

```rust
let init_pkt_c = client.process(None, now()).dgram();
assert!(init_pkt_c.is_some());
```
It's much easier to work with in general.  Less use of `as_dgram_ref()` and `unwrap`.

The line below would be regular rust:
```rust
assert_eq!(init_pkt_c.as_ref().unwrap().len(), 1232);
```

And if you wanted to clone the datagram, it is the much simpler:

```rust
let init_pkt_s = server.process(init_pkt_c.clone(), now()); 
```

neqo-transport/src/connection.rs
```diff
@@ -3694,4 +3734,181 @@ mod tests {
...
+
+        // Check that the Initial keys are dropped at the server
+        // We will check this by processing init_pkt_c a third time.
+        // The Initial packet has been dropped and padding that follows it.
+        // There is no dups, everything has been dropped.
+        check_is_packet_drop(&mut server, init_pkt_c.dgram(), 1, 0);
+    }
+
+    #[test]
+    fn test_discarded_handshake_keys() {
```
martinthomson: As I said before, we should wait for HANDSHAKE_DONE before writing tests for this.  Otherwise, we're in undefined territory.

neqo-transport/src/connection.rs
martinthomson: To be clear, I think that this is likely not correct now, or in future when HANDSHAKE_DONE is added, so I don't want to discard keys here and risk problems.  We can safely keep them indefinitely until we get HANDSHAKE_DONE.

neqo-transport/src/connection.rs
ddragana: it is funny that cargo is not complaining about this.

neqo-transport/src/connection.rs
ddragana: ok I will remove dropping of handshake keys.

neqo-transport/src/recovery.rs
```diff
@@ -555,9 +576,23 @@ impl LossRecovery {
...
+            .collect()
+    }
+
+    pub fn remove_state_for_pn_space(&mut self, pn_space: PNSpace) {
+        self.spaces[pn_space]
+            .remove_ignored()
+            .map(|p| {
+                self.cc.discard(&p);
+            })
             .collect()
```
martinthomson: I'm going to accept this, but this pattern is better replaced with a for loop.

martinthomson:
I didn't want to wait another day for this, so I rebased it, fixed a clippy warning and landed it manually.

PR #397, opened 2020-01-17
Key updates

martinthomson:
This changes the code to operate on packet number spaces instead of
epochs.  That makes it slightly easier to manage in some places.

The major shift here is in how keys are made.  The current code just
tries to make every key when it is sending packets, iterating through
the epochs as it goes.  This code attempts to only create keys at the
right points in the state machine.  This is only really perfect for
application read keys, which are generated when the handshake is
complete.  Other keys it creates after processing CRYPTO frames.

A later change that might help is to propagate events from the TLS layer
about key availability, which would make this more precise, but this was
already getting big.

This doesn't do key discard, but it should be much easier to implement
now that keys won't be re-created every time.

This version is a draft that includes a bunch of changes in packages
other than transport, but I'll drop that once I can rebase it (and those
other PRs land).  I wanted to put this out there to get a little bit of
feedback.  Only the stuff in `neqo-transport` is relevant here.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -655,18 +655,20 @@ impl SecretAgent {
         self.fd = null_mut();
     }
 
-    // State returns the status of the handshake.
+    /// State returns the status of the handshake.
     #[must_use]
     pub fn state(&self) -> &HandshakeState {
         &self.state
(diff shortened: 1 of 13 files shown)


agrover:
:heart:

PR #399, opened 2020-01-18
No Issue: Pad Initial packets from a client with Padding Frames.

hawkinsw:
Instead of padding a datagram containing an Initial packet from a client with null bytes, pad the datagram's Initial packet with with Padding frames.

Again, NB: This is a WIP. If @martinthomson and @agrover are okay with this approach, I will fix the tests to accommodate the new padding.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -1080,13 +1080,30 @@ impl Connection {
     /// Build a datagram, possibly from multiple packets (for different PN
     /// spaces) and each containing 1+ frames.
     fn output_pkt_for_path(&mut self, now: Instant) -> Res<Option<Datagram>> {
-        let mut out_bytes = Vec::new();
+        // A data structure used solely within this function to hold a representation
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection.rs
```diff
@@ -1080,13 +1080,30 @@ impl Connection {
...
     /// spaces) and each containing 1+ frames.
     fn output_pkt_for_path(&mut self, now: Instant) -> Res<Option<Datagram>> {
-        let mut out_bytes = Vec::new();
+        // A data structure used solely within this function to hold a representation
+        // of a packet that is pending transmission in the next datagram. We cannot
+        // encode the constructed packet at each epoch immediately because, depending
+        // on the ultimate size of the datagram and the current state of the connection,
+        // we may need to pad an Initial packet.
+        struct PendingPacket {
+            epoch: Epoch,
```
martinthomson: It might pay to wait a little here. I have removed epoch in my latest change.

PR #401, opened 2020-01-20
Better Pin handling

martinthomson:
Our current code was taking pinned objects and just casting them to the
requisite types.  This works, but I think that it was by accident more
than design.  I believe that what was happening was that `Pin<Box<T>`,
`Box<T>`, and `*mut T` are all effectively the same value once all the
rust compiler decorations are erased. So a cast to a pointer from any of
these does work.  However, this is not something the API guarantees.
Though `Pin` and `Box` are both annotated with `#[repr(transparent)]`,
this is not a promise of a particular layout, but a hint.

The correct way to handle these cases is to use `Pin::as_mut()` to get a
`Pin<&mut T>` and then use the unsafe method `Pin::as_unchecked_mut()`
on that to get a mutable reference to the memory.

This pinning stuff is a little hard to grok still, but my understanding
is that using `Box` consistently, as we do, ensures that we don't have a
problem unless we try to move the object that the `Box` points at.
Using `Pin` is supposed to help with that, but this code is so littered
with `unsafe` blocks, we will probably just have to be vigilant.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -228,7 +228,7 @@ pub struct SecretAgent {
 impl SecretAgent {
     fn new() -> Res<Self> {
         let mut io = Box::pin(AgentIo::new());
-        let fd = Self::create_fd(io.as_mut())?;
+        let fd = Self::create_fd(&mut io)?;
         Ok(Self {
             fd,
             secrets: SecretHolder::default(),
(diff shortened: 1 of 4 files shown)


neqo-crypto/src/p11.rs
```diff
@@ -99,6 +99,7 @@ impl std::fmt::Debug for SymKey {
 }
 
 /// Generate a randomized buffer.
+#[must_use]
```
martinthomson: I've landed this bit on master.  I have no idea how the other changes landed without hitting this.

martinthomson:
I just realized that the whole use of `Pin::as_mut` and whatnot is just wrong.  That the compiler lets me do this is a bit of a concern actually.  I'm going to produce another iteration on this soon.

martinthomson:
OK, I'm happy with this one now.

martinthomson:
Thanks!.  Merged manually with those calls inlined.  As you say, we don't need to have those calls outside the unsafe blocks just because they are safe.  (I think that I was more concerned about line length, but it isn't generally an issue.)

PR #407, opened 2020-01-22
HANDSHAKE_DONE

martinthomson:
This is incomplete in the sense that it doesn't discard handshake state (yet).  @ddragana, do you think that you could use this in combination with your previous code (and tests)?

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -85,6 +85,7 @@ pub enum State {
     WaitInitial,
     Handshaking,
     Connected,
+    Confirmed,
     Closing {
         error: ConnectionError,
         frame_type: FrameType,
@@ -92,6 +93,13 @@ pub enum State {
         timeout: Instant,
     },
     Closed(ConnectionError),
+}
+
+impl State {
+    #[must_use]
+    pub fn connected(&self) -> bool {
+        matches!(self, Self::Connected | Self::Confirmed)
+    }
 }
 
 // Implement Ord so that we can enforce monotonic state progression.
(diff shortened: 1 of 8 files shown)


neqo-transport/src/connection.rs
```diff
@@ -2781,11 +2861,36 @@ mod tests {
         assert!(client.events().any(stream_readable));
     }
 
+    /// Getting the client and server to reach an idle state is surprisingly hard.
+    /// The server sends HANDSHAKE_DONE at the end of the handshake, and the client
+    /// doesn't immediately acknowledge it.
+
+    /// Force the client to send an ACK by having the server send two packets out
```
ddragana: you can force it to send ack by adding some time to now() ACK_DALAY=200ms.

ddragana: but you will mix times a bit. reordering packets is better.

ddragana:
> This is incomplete in the sense that it doesn't discard handshake state (yet). @ddragana, do you think that you could use this in combination with your previous code (and tests)?

yes, this is fine. thanks!

martinthomson:
This should be ready for you to take over then.  Or you can merge it and create a follow-up.  I want to do the packet changes I talked about for Retry.

martinthomson:
I'd like to land this, so that I can use it as the basis of other changes.  Is there any reason why I couldn't?

neqo-transport/src/connection.rs
```diff
@@ -271,6 +291,48 @@ impl IdleTimeout {
...
+enum StateSignaling {
+    Idle,
+    HandshakeDone,
+    ConnectionClose,
+    CloseSent,
+}
+
+impl StateSignaling {
+    pub fn handshake_done(&mut self) {
+        *self = Self::HandshakeDone
```
ddragana: Can you add a debug assertion that we are in the right state? Also handle not transitioning from ConnectionClose and CloseSent into HandshakeDone for non-debug build.

neqo-transport/src/connection.rs
```diff
@@ -271,6 +291,48 @@ impl IdleTimeout {
...
+    pub fn closing(&self) -> bool {
+        *self == Self::ConnectionClose
+    }
+
+    pub fn close(&mut self) {
+        *self = Self::ConnectionClose
+    }
+
+    pub fn close_sent(&mut self) {
+        *self = Self::CloseSent
```
ddragana: add an assertion here as well.

ddragana:
We can also land this and I will add assertion in my PR for a keys discarding.

PR #410, opened 2020-01-22
No Issue: Add method to indicate server(s) is/are listening

hawkinsw:
As a UDP-based protocol with an event-driven implementation,
there's no real reason for a having a special method to tell a
server to start started listening for connections. However,
the qlog specification has a "server listening" event. Provide
a method for the server user to indicate that the server is
listening. Once qlog implementation progresses, this is the
method where neqo will log the server listening event.

### DIFF
--- a/neqo-http3-server/src/main.rs
+++ b/neqo-http3-server/src/main.rs
@@ -205,23 +205,20 @@ fn main() -> Result<(), io::Error> {
             PollOpt::edge(),
         )?;
         sockets.push(socket);
-        servers.insert(
-            local_addr,
-            (
-                Http3Server::new(
-                    Instant::now(),
-                    &[args.key.clone()],
-                    &[args.alpn.clone()],
-                    AntiReplay::new(Instant::now(), Duration::from_secs(10), 7, 14)
-                        .expect("unable to setup anti-replay"),
-                    Rc::new(RefCell::new(FixedConnectionIdManager::new(10))),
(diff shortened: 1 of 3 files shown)


martinthomson:
So I don't think that we should be doing listening.  The model for neqo is that calling code is responsible for managing sockets, sending, and receiving datagrams.

It's OK if we never log the "server listening" event.

hawkinsw:
Completely understood. Just thought I'd give it a try. Sounds like we'll just have to expose something to a neqo end user for creating such an event if they would like. Does that sound right to you?

martinthomson:
That seems right.  My view is that we should probably provide a qlog stream somehow; they can splice other events into that stream.  They can be responsible for serializing too.

hawkinsw:
That is the way that the awesome @agrover has constructed things. I'm just filling in the gaps now. Thanks for the review and discussion. Closing this now. Thanks again!

agrover:
This is for qlog, right? I think this just means that neqo-http3-server should be the entity emitting the qlog server listening event, not neqo-http3 or neqo-transport?

hawkinsw:
That's right, @agrover. Would you be okay with adding a `listen()` method to the neqo-http3-server that the user could call to note that the server started listening? With these two proposed `listen()` methods I was not trying define /where/ we would put the event emission, just offer a spot where it /could/ happen.

PR #412, opened 2020-01-24
Packet Builders

martinthomson:
This implements a new code path for constructing packets.  It's a lot simpler.

I have duplicated a bunch of code from the packet tests, which ended up with me disabling the tests in that file.  So right now, this is in a bit of a messy state.

In adding Retry and VN, I also broke Retry because of the new integrity check (yay).  I am currently truncating Retry on read so that those tests can continue to function.  I had to disable one of the Retry tests, but I should have that up and running soon.

(Based on #407).


martinthomson:
So, I think that I'd like to get this in.  This should be the last piece we need for draft -25.

(In order to get to -25 properly, a few constants in the tests need to change, but I have those ready and commented out.)

neqo-transport/src/dump.rs
```diff
@@ -9,11 +9,11 @@
 
 use crate::connection::Connection;
 use crate::frame::Frame;
-use crate::packet::PacketHdr;
+use crate::packet::{PacketNumber, PacketType};
 use neqo_common::{qdebug, Decoder};
 
 #[allow(clippy::module_name_repetitions)]
-pub fn dump_packet(conn: &Connection, dir: &str, hdr: &PacketHdr, payload: &[u8]) {
+pub fn dump_packet(conn: &Connection, dir: &str, pn: PacketNumber, pt: PacketType, payload: &[u8]) {
```
agrover: BuildPacketHeader outputs pt, pn. Suggest dump_packet also take them in this order, to salve my OCD.

neqo-transport/src/lib.rs
```diff
@@ -27,16 +28,16 @@ mod stream_id;
...
+pub use self::connection::{Connection, FixedConnectionIdManager, Output, Role, State};
 pub use self::events::{ConnectionEvent, ConnectionEvents};
 pub use self::frame::CloseError;
 pub use self::frame::StreamType;
 pub use self::tparams::{tp_constants, TransportParameter};
 
 /// The supported version of the QUIC protocol.
-pub const QUIC_VERSION: u32 = 0xff00_0000 + 25;
+pub type Version = u32;
+pub const QUIC_VERSION: Version = 0xff00_0000 + 25;
```
agrover: When did this happen? I like it.

neqo-transport/src/packet.rs
```diff
@@ -24,745 +27,931 @@ const PACKET_TYPE_RETRY: u8 = 0x03;
...
+        TLS_AES_128_GCM_SHA256,
+        &[
+            0x65, 0x6e, 0x61, 0xe3, 0x36, 0xae, 0x94, 0x17, 0xf7, 0xf0, 0xed, 0xd8, 0xd7, 0x8d,
+            0x46, 0x1e, 0x2a, 0xa7, 0x08, 0x4a, 0xba, 0x7a, 0x14, 0xc1, 0xe9, 0xf7, 0x26, 0xd5,
+            0x57, 0x09, 0x16, 0x9a,
+        ],
+    )
+    .unwrap();
+    Aead::new(TLS_VERSION_1_3, TLS_AES_128_GCM_SHA256, &secret, "quic ").unwrap()
+}
```
agrover: Would crypto.rs be a better home for this?

Also, just from looking at quic-tls  "5.8. Retry Packet Integrity " as a non-crypto expert, I see where the 0x65.. is coming from, but the label, why is it "quic " here?

martinthomson: This is what we do for all of the AEAD construction, it's the extra piece that distinguishes the QUIC keys from regular TLS 1.3 keys.

neqo-transport/src/connection.rs
```diff
@@ -1157,163 +1245,62 @@ impl Connection {
...
-            let mut packet = encode_packet(tx, &hdr, &encoder);
+            encoder = builder.build(self.crypto.states.tx(*space).unwrap())?;
+            assert!(encoder.len() <= path.mtu());
 
             if tx_mode != TxMode::Pto && ack_eliciting {
                 self.idle_timeout.on_packet_sent(now);
             }
 
+            // Normal packets are in flight if they include PADDING frames,
+            // but we don't send those.
```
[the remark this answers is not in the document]
martinthomson: It is there in the part where we add frames.

PR #414, opened 2020-01-24
Discard Handshake keys

ddragana:


ddragana:
This is based on #388

neqo-transport/src/connection.rs
```diff
@@ -306,6 +306,10 @@ enum StateSignaling {
 
 impl StateSignaling {
     pub fn handshake_done(&mut self) {
+        if !matches!(*self, Self::Idle) {
```
martinthomson: A simple self != Self::Idle works.

neqo-transport/src/connection.rs
```diff
@@ -1311,10 +1316,18 @@ impl Connection {
...
                 // Client can send Handshake packets -> discard Initial keys and states
                 self.discard_keys(PNSpace::Initial);
             }
+
+            if *space == PNSpace::Handshake
+                && self.role == Role::Server
+                && self.state == State::Confirmed
+            {
+                // We could discard handshake keys in set_state, but we are waiting to send an ack.
+                self.discard_keys(PNSpace::Handshake);
```
martinthomson: Hmm, it would be really nice to do this in set_state, but I understand the value of having the ACK sent, even if it is only once.  There aren't a lot of places where we have an opportunity to do this discard otherwise.  This is fine for the moment, but I want to think about this a little more.

neqo-transport/src/connection.rs
```diff
@@ -3320,7 +3333,7 @@ mod tests {
     }
 
     #[test]
-    fn pto_handshake() {
+    fn dd_pto_handshake() {
```
martinthomson: ??

neqo-transport/src/connection.rs
```diff
@@ -3364,9 +3377,19 @@ mod tests {
...
         // Server receives the first packet.
+        // The output will be a Handshake packet with an ack and a app pn space packet with
+        // HANDSHAKE_DONE.
         let pkt = server.process(pkt1, now).dgram();
         assert!(pkt.is_some());
 
+        // Check that the second packet(pkt2) has a Handshake and an app pn space packet.
+        // The server has discarded the Handshake keys already, therefore the handshake packet
+        // will be dropped.
+        let droppes_before = server.stats().dropped_rx;
```
martinthomson: spelling

PR #416, opened 2020-01-26
No Issue: Refactor QPACK instruction types

hawkinsw:
Refactor QPACK instruction types to the crate for use later by qlog.

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -10,19 +10,13 @@ use crate::huffman::encode_huffman;
 use crate::qpack_helper::read_prefixed_encoded_int_with_connection;
 use crate::qpack_send_buf::QPData;
 use crate::table::HeaderTable;
+use crate::DecoderInstructions;
 use crate::Header;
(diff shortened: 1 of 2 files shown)


hawkinsw:
It turns out that this is not a level of abstraction that I will need for qlog. If you would like to include it regardless, I will leave open the PR :-) Otherwise, feel free to say, "No!". 

Sorry for the false request.

ddragana:
It looks like that the change is not finish. It (almostt only) adds deadcode that you do not need anymore for qlog. Do you want to add more changes? I f not I would say no to this PR.

hawkinsw:
Yup, that was intentionally dead code because those values had no uses
yet. I expected that they would as we added the qlog implementation to
qpack. But, like I said, I don't think that we will need these values
after all. So, unless you think that there are other components that
might need to use these values, feel free to disregard! Again, sorry
for the noise!

Will

ddragana:
thanks. I will close this one, because we do not need the values at the moment. If we need them in the future we can add them then.

PR #427, opened 2020-02-03
WIP: Fix #426: Implement/enforce active_connection_id_limit TP.

### LINKED ISSUE #426: Implement/enforce active_connection_id_limit transport parameter
Neqo does not currently parse the `active_connection_id_limit` TP and does not enforce its semantics upon receiving a `NEW_CONNECTION_ID` frame.

hawkinsw:


neqo-transport/src/connection.rs
```diff
@@ -1620,8 +1652,21 @@ impl Connection {
...
-                    .insert(sequence_number, (connection_id, stateless_reset_token));
+                // If adding an additional active connection id sends us over the limit,
+                // return a ConnectionIdLimitError.
+                if self.connection_ids.len() + 1
+                    > self
+                        .tps
+                        .borrow()
+                        .remote()
+                        .get_integer(tp_constants::ACTIVE_CONNECTION_ID_LIMIT)
+                        as usize
```
martinthomson: We should be avoiding the cast in favour of `try_from()`.

Also, this would likely benefit from a temporary variable.

neqo-transport/src/connection.rs
```diff
@@ -1420,6 +1420,38 @@ impl Connection {
         }
     }
 
+    fn validate_active_connection_id_limit(&self) -> Res<()> {
+        if let Some(path) = &self.path {
+            let tph = self.tps.borrow();
+            let remote = tph.remote();
+            if path.remote_cid.len() == 0 {
```
martinthomson: `is_empty()` should work here.  I'm surprised that clippy didn't object.

hawkinsw:
Thanks for the feedback! I didn't expect it but really appreciate it. I was
just posting it so you knew I was making progress. It was a WIP because of
testing, specifically, so I'm glad you mentioned that. I will follow up
directly on your other feedback when I dig back in. Thanks again for the
comments -- very helpful!

neqo-transport/src/connection.rs
hawkinsw: I definitely agree with both. In fact, I had it as a temporary variable and then switched it back. Thanks for confirming my initial gut call on that. 

As for the `try_from`, I think that's a great idea. However, what I realized is what _should_ be done is to upcast from `usize` into `u64` and do the comparison on that. There is no possible data loss in that case. Does that sound okay?

neqo-transport/src/connection.rs
hawkinsw: It doesn't surprise me -- clippy and I are in a fight right now. For some reason I can't get it to consistently run in a way that I am confident it is actually doing any checking. Thanks for the note on this. I will definitely fix it.

hawkinsw:
@martinthomson See also #447

hawkinsw:
> So the good news here, if you like, is that the transport parameter is being simplified. The value must be >= 2. And that's all. The weird restriction on the server when the client chooses a zero-length connection ID will soon be gone.

Did this come out of the interim meeting? This is awesome. It was always odd to me!

martinthomson:
Yep.  Less weirdness = good :)

neqo-transport/src/connection.rs
agrover: Doesn't that mean you would just be `try_from`ing the other side of the comparison? Either is fine, this is an extreme corner case and all this is only to please the compiler.

agrover:
Let's try to get this one in soon.

hawkinsw:
I am going to close this and fold it into the larger work that I am doing on managing connection ids, if that is okay. Just let me know if anyone objects! @martinthomson @agrover

PR #440, opened 2020-02-04
Fix some part of qpack code

ddragana:
- fix unused_variables and dead_code
- fix use of Vec<u8>
- header ack should also ack the instructions on the encoder side
- properly use max_block_streams and improve how the references to the dynamic entries are stored. Also make sure this works if a stream has multiple header blocks.
- some tests refactoring
- add test for max_block_streams and reference countting for entries.

### DIFF
--- a/neqo-qpack/src/decoder.rs
+++ b/neqo-qpack/src/decoder.rs
@@ -4,7 +4,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![allow(unused_variables, dead_code)]
 use crate::huffman::Huffman;
 use crate::qpack_helper::{
     read_prefixed_encoded_int_slice, read_prefixed_encoded_int_with_connection, BufWrapper,
@@ -311,7 +310,7 @@ impl QPackDecoder {
                                 self.table.insert_with_name_ref(
                                     *name_static_table,
                                     *name_index,
-                                    value_to_insert,
+                                    &value_to_insert,
                                 )?;
                                 self.total_num_of_inserts += 1;
                                 self.state = QPackDecoderState::ReadInstruction;
@@ -427,7 +426,7 @@ impl QPackDecoder {
                                     mem::swap(&mut value_to_insert, value);
                                 }
                                 qdebug!([label], "received instruction - insert with name literal name={:x?} value={:x?}", name_to_insert, value_to_insert);
-                                self.table.insert(name_to_insert, value_to_insert)?;
+                                self.table.insert(&name_to_insert, &value_to_insert)?;
                                 self.total_num_of_inserts += 1;
                                 self.state = QPackDecoderState::ReadInstruction;
                             } else {
@@ -479,15 +478,18 @@ impl QPackDecoder {
         if cap > u64::from(self.max_table_size) {
             return Err(Error::EncoderStreamError);
         }
-        self.table.set_capacity(cap);
-        Ok(())
+        self.table
+            .set_capacity(cap)
+            .map_err(|_| Error::EncoderStreamError)
     }
 
     fn header_ack(&mut self, stream_id: u64, required_inserts: u64) {
         let ack_increment_delta = required_inserts - self.table.get_acked_inserts_cnt();
         self.send_buf
             .encode_prefixed_encoded_int(0x80, 1, stream_id);
-        self.table.increment_acked(ack_increment_delta);
+        self.table
+            .increment_acked(ack_increment_delta)
+            .expect("This should never happen");
     }
 
     pub fn cancel_stream(&mut self, stream_id: u64) {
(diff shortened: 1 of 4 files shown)


neqo-qpack/src/decoder.rs
```diff
@@ -479,15 +478,19 @@ impl QPackDecoder {
         if cap > u64::from(self.max_table_size) {
             return Err(Error::EncoderStreamError);
         }
-        self.table.set_capacity(cap);
+        if self.table.set_capacity(cap).is_err() {
+            return Err(Error::EncoderStreamError);
+        }
```
agrover: Suggest `map_err()?` here, if this is just converting one err type/value to another err type/value.

neqo-qpack/src/decoder.rs
```diff
@@ -479,15 +478,19 @@ impl QPackDecoder {
...
     }
 
     fn header_ack(&mut self, stream_id: u64, required_inserts: u64) {
         let ack_increment_delta = required_inserts - self.table.get_acked_inserts_cnt();
         self.send_buf
             .encode_prefixed_encoded_int(0x80, 1, stream_id);
-        self.table.increment_acked(ack_increment_delta);
+        if self.table.increment_acked(ack_increment_delta).is_err() {
+            panic!("This should never happen");
+        }
```
agrover: Use `x.unwrap("This should never happen")` instead of `if x.is.err() {panic!()}`

neqo-qpack/src/decoder.rs
```diff
@@ -501,7 +504,9 @@ impl QPackDecoder {
         if ack_increment_delta > 0 {
             self.send_buf
                 .encode_prefixed_encoded_int(0x00, 2, ack_increment_delta);
-            self.table.increment_acked(ack_increment_delta);
+            if self.table.increment_acked(ack_increment_delta).is_err() {
+                panic!("This should never happen");
+            }
```
agrover: Use `x.unwrap("This should never happen")` instead of `if x.is.err() {panic!()}`

neqo-qpack/src/encoder.rs
```diff
@@ -172,35 +174,99 @@ impl QPackEncoder {
...
+                self.unacked_header_blocks.remove(&stream_id);
+            }
+        }
+        if new_acked > self.table.get_acked_inserts_cnt() {
+            if self
+                .insert_count_instrunction(new_acked - self.table.get_acked_inserts_cnt())
+                .is_err()
+            {
+                panic!("This should neve happen");
+            }
```
agrover: same as previous comment: avoid `panic!()` by using `expect()`

neqo-qpack/src/table.rs
```diff
@@ -192,51 +264,66 @@ impl HeaderTable {
...
         Ok(())
     }
 
-    pub fn increment_acked(&mut self, increment: u64) {
+    pub fn increment_acked(&mut self, increment: u64) -> Res<()> {
+        qtrace!([self], "increment acked by {}", increment);
         self.acked_inserts_cnt += increment;
+        if self.base < self.acked_inserts_cnt {
+            return Err(Error::InternalError);
+        }
```
agrover: could this be an assert or debug_assert? This would simplify signature of this fn as well as caling fn.

neqo-qpack/src/table.rs
```diff
@@ -6,31 +6,46 @@
...
-    pub fn remove_ref(&mut self, stream_id: u64, _block: u8) {
-        self.refs.remove(&stream_id);
+    pub fn remove_ref(&mut self) -> Res<()> {
+        if self.refs == 0 {
+            // This should never happen, this is a bug in our code.
+            debug_assert!(false);
+            return Err(Error::InternalError);
+        }
+        self.refs -= 1;
+        Ok(())
```
agrover: I think long term `remove_ref` should not return Res. If we want to while this code is in development then I guess I'm ok but please add a qerror, and at some point I hope we can feel confident enough to stick an assert in here and not return Res.

neqo-qpack/src/encoder.rs
```diff
@@ -172,35 +174,99 @@ impl QPackEncoder {
...
+                .iter()
+                .flat_map(|hb| hb.iter())
+                .any(|e| *e >= acked_inserts_cnt)
+            {
+                self.blocked_stream_cnt += 1;
+            }
+        }
+    }
+
+    fn insert_count_instrunction(&mut self, increment: u64) -> Res<()> {
```
agrover: typo, "instruction"

neqo-qpack/src/encoder.rs
```diff
@@ -172,35 +174,99 @@ impl QPackEncoder {
...
+                for iter in ref_list {
+                    self.table.remove_ref(iter)?;
+                    if iter >= new_acked {
+                        new_acked = iter + 1;
+                    }
+                }
+            } else {
+                debug_assert!(false, "We should have at least one header block.");
+            }
+            if hb_list.len() == 0 {
```
agrover: use `is_empty()`

neqo-qpack/src/table.rs
ddragana: no this function is called when an Insert Count Increment instruction is received. Server implementation can have bugs, therefore we should not crash.

neqo-qpack/src/table.rs
ddragana: I will make the code to assert, let's find all the bus now, it is easier.

neqo-qpack/src/table.rs
agrover: ah ok!

agrover:
Feel free to merge once Clippy is made happy.

Issue #264, opened 2019-10-21
panic against nghttp2.org


agrover:
`RUST_BACKTRACE=1 RUST_LOG=debug,neqo_transport::dump ./target/debug/neqo-client https://nghttp2.org:4433/`

agrover:
http frame parsing issue. @ddragana maybe your refactor could also squash this one?

```
0s364ms DEBUG [TransactionClient 0] A new frame has been received.
thread 'main' panicked at 'Read should succeed: HttpFrameUnexpected',
```

Should be very reproducible.

ddragana:
I know about this one. It will be fixed soon.
The server send trailers and for time been, until I implement trailers, I interpreted it as a HttpFrameUnexpected. It is easy to fix. I may as well add trailer now. it is not a lot of work.

ddragana:
Fixed by #444

PR #444, opened 2020-02-05
HFrame::Header should contain header_block and make the receive side of transaction simpler

### LINKED ISSUE #344: Changing how we read headers
The current implementation reads HEADER frame only up to its length and leas the rest on the stream. This was implemented in this way thinking about hot necko reads data. I think this can be make simpler by reading the complete header.
Also we should supply headers in the HeaderReady event instead of letting app pick it up, again this is make this way because the way necko streams work, but I think I can make it work in necko even if HeaderReady event supplies the header block and I think the code will be not too terrible.

### LINKED ISSUE #437: Trailers
ngtcp2 sends them.  We should at least drop them.

ddragana:
Fixes #344     
Furthermore:
    - Move response part of the client transaction into a separate file
    - do not return an error if trailers are received (Fixes #437 #264 ).

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -709,6 +709,18 @@ mod tests {
         0x43, 0xd3, 0xc1,
     ];
 
+    const HTTP_HEADER_FRAME_0: &[u8] = &[0x01, 0x06, 0x00, 0x00, 0xd9, 0x54, 0x01, 0x30];
+
+    // The response header from HTTP_HEADER_FRAME (0x01, 0x06, 0x00, 0x00, 0xd9, 0x54, 0x01, 0x30) are
+    // decoded into:
+    fn check_response_header_0(header: Vec<Header>) {
+        let expected_response_header_1 = vec![
+            (String::from(":status"), String::from("200")),
+            (String::from("content-length"), String::from("0")),
+        ];
+        assert_eq!(header, expected_response_header_1);
+    }
+
     const HTTP_RESPONSE_1: &[u8] = &[
         // headers
         0x01, 0x06, 0x00, 0x00, 0xd9, 0x54, 0x01, 0x37, // the first data frame
@@ -2317,11 +2329,10 @@ mod tests {
             .encoder
             .encode_header_block(&headers, request_stream_id);
         let hframe = HFrame::Headers {
-            len: encoded_headers.len() as u64,
+            header_block: encoded_headers.to_vec(),
         };
         let mut d = Encoder::default();
         hframe.encode(&mut d);
-        d.encode(&encoded_headers);
         let d_frame = HFrame::Data { len: 3 };
         d_frame.encode(&mut d);
         d.encode(&[0x61, 0x62, 0x63]);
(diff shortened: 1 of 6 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -2907,4 +2985,173 @@ mod tests {
...
+            .conn
+            .stream_send(request_stream_id, HTTP_HEADER_FRAME_0);
+        server.conn.stream_close_send(request_stream_id).unwrap();
+
+        let out = server.conn.process(None, now());
+        client.process(out.dgram(), now());
+
+        let header_ready: fn(&Http3ClientEvent) -> _ =
+            |e| matches!(*e, Http3ClientEvent::HeaderReady { .. });
+        let data_readable: fn(&Http3ClientEvent) -> _ =
```
martinthomson: I would move the closures to just above the line they are used.

neqo-http3/src/response_stream.rs
```diff
@@ -0,0 +1,299 @@
...
+        match self.state {
+            ResponseStreamState::WaitingForResponseHeaders => {
+                if header_block.is_empty() {
+                    self.add_headers(None)?;
+                } else {
+                    self.state = ResponseStreamState::DecodingHeaders { header_block, fin };
+                }
+             }
+            ResponseStreamState::WaitingForData => {
+                // TODO implement trailers, for now just ignore them.
```
martinthomson: We should tolerate trailers in the middle of a sequence of DATA frames.

neqo-http3/src/response_stream.rs
```diff
@@ -0,0 +1,299 @@
...
+        } else {
+            qdebug!([self], "A new frame has been received.");
+            Ok((Some(self.frame_reader.get_frame()?), fin))
+        }
+    }
+
+    pub fn read_response_headers(&mut self) -> Res<(Vec<Header>, bool)> {
+        if let ResponseHeadersState::Ready(ref mut headers) = self.response_headers_state {
+            let mut tmp = Vec::new();
+            if let Some(ref mut hdrs) = headers {
```
martinthomson: You shouldn't use this pattern.  Try:

```rust
let tmp = headers.unwrap_or_else(|| Vec::new());
```

Try to think of a better name than `tmp` too.

neqo-http3/src/response_stream.rs
```diff
@@ -0,0 +1,299 @@
...
+                mem::swap(&mut tmp, hdrs);
+            }
+            self.response_headers_state = ResponseHeadersState::Read;
+            let mut fin = false;
+            if self.state == ResponseStreamState::ClosePending {
```
martinthomson: Same here:

```rust
let fin = if ... {
  // steps
   true
} else {
  false
};
```

neqo-http3/src/response_stream.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub fn read_response_data(
+        &mut self,
+        conn: &mut Connection,
+        buf: &mut [u8],
+    ) -> Res<(usize, bool)> {
+        match self.state {
+            ResponseStreamState::ReadingData {
+                ref mut remaining_data_len,
+            } => {
+                let to_read = if *remaining_data_len > buf.len() {
```
martinthomson: std::cmp::min?

ddragana:
> I think that this is a good change, but I have a few more things for you to think more about (though I wouldn't block this on answers).
> 
> Why do we treat response streams specially?

The push streams only need response stream. Separating response stream was done in the server push patch. I have extracted it from that patch  (I will need to rebase the server push patch as well).

> Why do we not allow trailers to appear in multiple places? (Ongoing discussion in HTTP seems to be moving toward allowing multiple sets of trailers, or "midders")

I know about that discussion. I wanted to wait for it to come to a conclusion. It is easy to change, updating test is more work :). I will do it in a follow up PR, because I want to land this as soon as possible.

> How might we expose trailers (and midders)?

There will be one more PR that will change HeaderReady events to hold headers, currently app needs to call read_response_header to actually get headers. In the similar way we will have Midders/TrailersReady event that will hold midders/trailers.
Adding this will be a bit work because decompressing midders (this is only relevant for midders) should not block reading data from a stream. (In case of Headers, if Headers are block on qpack instruction the whole stream is block). The state transition will not be as nice as it is now, e.g. midders may be blocked but we are in WaitingForData state or Closing state.

Note: Firefox expose only Server timing header that can be receive in a trailer. 

> 
> This is too big for me to review fully given time constraints, unless you are willing to wait a week or more. But what I have seen is good and should be landed.

PR #448, opened 2020-02-07
Send 2 packets on PTO

martinthomson:
This causes packets to be marked lost when the PTO timer fires.  As a result, the TxMode can be removed entirely.  More packets are marked lost each time the timer fires.  This means that older data is available (probably, some frames won't be retransmitted, such as DATAGRAM).

The only bit I'm uncertain about is the interaction between packet number spaces.  I found that sending Handshake before Application Data results in the PTO timer for Application Data doesn't fire in tests (it should eventually, but the timing is weird as a result of having a global `pto_count` value).  By marking Application Data as lost, fewer tests failed.

It seems like in general it is OK to mark stuff as lost, but I'm concerned that this logic is incorrect and that I should instead just accept the fact that Application Data packets will just include a PING frame on PTO.

That said, what I think happens if I don't include that hack is that we get strangely staggered PTOs.  I need to investigate more.


martinthomson:
So there is a weird effect if you send Handshake and Application packets at the same time.  The PTO timer for the Handshake packets is shorter.  So when the PTO fires for Handshake two things happen:

1. The timer for Application hasn't fired yet.  So we don't do anything about marking Application packets as lost.

2. The PTO count is increased, which further bumps the Application timer out.

To give an example, if we have a PTO that is 100ms base (rtt + 4rttvar) with ack delay of 25ms, we will PTO as follows:

* At 100ms, Handshake.
* At 250ms, Application (2x125).
* At 500m, Handshake (100 + 4x100).
* At 1250ms, Application (250 + 8x125).

This is just weird because the actual rate of PTO is skewed.

What I have done is to mark packets at *higher* levels lost when each PTO fires.  That helps a little, because Application data will be sent when the Handshake PTO fires.  But this doesn't help in the other direction because the counter is incremented by either PTO firing.  The worst here is that Application PTOs further delay Handshake PTOs AND we don't send Handshake packets when the Application PTO fires.

If there was only Application packets on PTO, timers would fire at 125, 375, 875, 1875.  Rather than 100 (sorta), 250, 500 (sorta), 1250.  Only Handshake would fire at 100, 300, 700, 1500 rather than 100, 500, 2100, 8500.  The Application packets are sent faster, the Handshake packets are sent much slower.

@ddragana, does this match your understanding of what should happen?  Is this enough of a problem to warrant changing the logic, perhaps by making the PTO count specific to a packet number space?

martinthomson:
Aside from the above weirdness, I think that this is ready for review.

martinthomson:
Thinking about this some more, I think that the biggest problem is that we don't send Handshake packets as often as we could.  I see a few options:

1. When there is a PTO, send a PING in spaces that are active but below the "min_pn_space".  This is probably most easily doable, we do cause the timers for individual spaces to move to a 4x exponent, but we end up sending packets in those spaces on the smallest PTO timer.

2. Remove the concept of minimum packet number space for sending on PTO.  The potential problem here is that we might not send packets from the space that needs a PTO if earlier spaces fill the packet.

3. Change the minimum packet number space to "first packet number space".  That is, we send in packet number space X, then iterate through all the others.  This might be closest to ideal in the sense that it addresses the primary PTO, but it requires some new iteration capabilities and that PTO space can still fill the packet and deny other spaces the option to send.

4. Don't worry so much about a corner case.

ddragana:
Let's look at the client side first:
1) if we only have initial key, there is not really question what to do :)
2) we have handshake keys but no 1RTT keys: the initial keys are discarded as soon as handshake keys are present, so we actually do not have any option
3) we have both handshake and 1RTT keys: It is important that the other side, in this case server, can respond. The server may have both keys, but it may have discarded the handshake keys, therefore it important to send 1RTT data -> always send at least a PING with 1RTT keys, it is the best to send packets in both encryption levels. The server may have not completed the handshake and therefore will not ack 1RTT data.

Server side:
1) server still has initial keys, it has not receive a handshake packet from the client yet. The client side may have handshake keys, but it may happen that it does not have them yet. Also the client may have discarded initial key already -> The best would be to send both, except all initial data has been acked than just send a handshake packet.
2) server has already received a handshake packet from a client. The initial key are already discarded. In this case it is important to send handshake packets.
3) server has received a 1RTT packet from a client, it has not send HANDSHAKE_DONE. That means the client still has handshake keys, but it has 1rtt keys as well. It is not important which one we are sending, it will be able to decrypt them both -> just send any data in-flight.
4) server has sent HANDSHAKE_DONE already. In this case the server has discarded the handshake keys and it does not have any choice.

PR #466, opened 2020-02-22
Convert transport parameters to varint encoding

martinthomson:
#464 won't be much good without this.


agrover:
@martinthomson can we get this one resolved and merged asap? We are in kind of a weird situation atm with this pending but #464 already merged.

martinthomson:
OK, that was trivial.  Sorry for not noticing the conflict.

PR #468, opened 2020-02-23
Make packet number decryption closer to constant time

martinthomson:
This likely isn't perfect, but this should reduce the side channels we
have from packet number decryption.

(This came up in the QUIPS workshop, which prompted me to check and fix.)


agrover:
:+1:

PR #471, opened 2020-02-25
Don't read extra bindgen flags from setup_for_gecko

glandium:
As of bug 1617769, the Firefox build system does due diligence.


glandium:
Bug 1617769 turns out to be a massive can of worms, and I may not go forward with it. Let's close this while I figure things out.

Issue #459, opened 2020-02-18
qpack: Fix Header Block Prefix


ddragana:
This will not be fix by issue #458. 
The current implementation assumes that  Required Insert Count (8+)  and Delta Base (7+)  are only 1 byte. Maybe that is fine but, we should make it work correctly at least limit how many elements cab added into a table.

PR #477, opened 2020-03-02
Refactor qpack

ddragana:


neqo-qpack/src/qpack_helper.rs
```diff
@@ -43,7 +82,41 @@ impl<'a> BufWrapper<'a> {
         }
     }
 
-    pub fn slice(&mut self, len: usize) -> Res<&[u8]> {
+    pub fn done(&self) -> bool {
+        self.offset == self.buf.len()
+    }
+
+    pub fn read_prefixed_int(&mut self, prefix_len: u8) -> Res<u64> {
+        assert!(prefix_len < 8);
```
martinthomson:
```suggestion
        debug_assert!(prefix_len < 8);
```

neqo-qpack/src/qpack_helper.rs
```diff
@@ -43,7 +82,41 @@ impl<'a> BufWrapper<'a> {
         }
     }
 
-    pub fn slice(&mut self, len: usize) -> Res<&[u8]> {
+    pub fn done(&self) -> bool {
+        self.offset == self.buf.len()
+    }
+
+    pub fn read_prefixed_int(&mut self, prefix_len: u8) -> Res<u64> {
```
martinthomson: It might pay to document these, including the assumptions you are making.  That is, that the underlying buffer is complete.

neqo-qpack/src/lib.rs
```diff
@@ -5,10 +5,13 @@
 // except according to those terms.
 
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
-#![warn(clippy::use_self)]
+//#![warn(clippy::use_self)]
```
martinthomson: Consider spending some more time with clippy on this crate.

neqo-qpack/src/header_block.rs
```diff
@@ -0,0 +1,971 @@
...
+use crate::qpack_helper::{to_string, ReceiverBufferWrapper};
+use crate::qpack_send_buf::QPData;
+use crate::table::HeaderTable;
+use crate::Header;
+use crate::{Error, Res};
+use crate::{Prefix, NO_PREFIX};
+use neqo_common::qtrace;
+use std::ops::Deref;
+
+const BASE_PREFIX_POSITIVE: Prefix = Prefix {
```
martinthomson: Given the way that prefixes are fragile, I wonder if a macro might not be a bad idea here.  You can pack a static assertion about the relationship between prefix and length into that using the `static_assertions` crate.

neqo-qpack/src/decoder.rs
```diff
@@ -526,206 +160,26 @@ impl QPackDecoder {
...
+        let mut decoder = HeaderDecoder::new(buf);
+
+        match decoder.decode_header_block(
+            &self.table,
+            self.max_entries,
+            self.total_num_of_inserts,
+        )? {
+            HeaderDecoderResult::Blocked(req_insert_cnt) => {
+                self.blocked_streams.push((stream_id, req_insert_cnt));
+                if self.blocked_streams.len() > self.max_blocked_streams as usize {
```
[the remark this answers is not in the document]
ddragana: I know that this code needs a bit more refactoring, I just needed to stop somewhere for now otherwise it would take me long to have this ready for review.
Thanks for all comments, I will fix them.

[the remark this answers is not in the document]
martinthomson: That's fine.  Feel free to set some of these comments aside as future work.

neqo-qpack/src/qpack_send_buf.rs
```diff
@@ -56,25 +59,24 @@ impl QPData {
     pub fn encode_prefixed_encoded_int_fix(
```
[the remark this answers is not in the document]
ddragana: the *_fix is writing on a certain place in a buffer and the one above appends to the end. I know this one is strange. I wanted to fix it in #459 that is the only usage of this code.
When I am already on it here, I will fix the function.

ddragana:
This patch concentrate on refactoring:
- encoding/decoding decoder instructions
- encoding/decoding encoder instructions
- encoding/decoding header blocks
- prefix.rs
- qpack_send_buf.rs(writing this line, I realize that this may need a better name, I will make that a follow up)
- reader.rs

The following patches will look into clippy, encoder.rs, decoder.rs, table.rs and huffman(make huffman do not use a table)
I want to look at encoder.rs, decoder.rs and table.rs after this refactor. A lot of code was extracted it is easier now to look at this file and think about if the structure(split between files) is correct.

neqo-qpack/src/decoder.rs
```diff
@@ -526,206 +161,26 @@ impl QPackDecoder {
...
+
+        match decoder.decode_header_block(
+            &self.table,
+            self.max_entries,
+            self.total_num_of_inserts,
+        )? {
+            HeaderDecoderResult::Blocked(req_insert_cnt) => {
+                self.blocked_streams.push((stream_id, req_insert_cnt));
+                if self.blocked_streams.len() > self.max_blocked_streams {
+                    return Err(Error::DecompressionFailed);
```
martinthomson: Use `if ... { Err(...) } else { Ok(None) }`

neqo-qpack/src/decoder_instructions.rs
```diff
@@ -0,0 +1,231 @@
...
+impl DecoderInstruction {
+    fn get_instruction(b: u8) -> Self {
+        if DECODER_HEADER_ACK.cmp_prefix(b) {
+            Self::HeaderAck { stream_id: 0 }
+        } else if DECODER_STREAM_CANCELLATION.cmp_prefix(b) {
+            Self::StreamCancellation { stream_id: 0 }
+        } else if DECODER_INSERT_COUNT_INCREMENT.cmp_prefix(b) {
+            Self::InsertCountIncrement { increment: 0 }
+        } else {
+            unreachable!("The above patterns match everything.");
```
martinthomson:
```suggestion
            unreachable!();
```

neqo-qpack/src/decoder_instructions.rs
```diff
@@ -0,0 +1,231 @@
...
+                    Ok(b) => {
+                        self.instruction = DecoderInstruction::get_instruction(b);
+                        self.state = DecoderInstructionReaderState::ReadInt {
+                            reader: IntReader::new(
+                                b,
+                                if DECODER_HEADER_ACK.cmp_prefix(b) {
+                                    DECODER_HEADER_ACK.len()
+                                } else {
+                                    DECODER_STREAM_CANCELLATION.len()
+                                },
```
martinthomson: It seems like this might be an opportunity to put a method on Prefix that handles this.  Maybe `DECODER_HEADER_ACK.int_reader(b).or_else(|| DECODER_STREAM_CANCELLATION.int_reader(b)).unwrap()`

Or you could do that on IntReader: `IntReader::make(&[DECODER_HEADER_ACK, DECODER_STREAM_CANCELLATION])` where the list is iterated over until the prefix matches.

PR #491, opened 2020-03-10
fix the bug

surender786:
watch this my mentor

Co-Authored-By: ddragana <dd.mozilla@gmail.com>


surender786:
Hi my mentor of outreachy  @ddragana ,
This is the right way to contribute?

martinthomson:
We do prefer pull requests, but usually we expect them to contain a useful title and a description of what the problem was, and how the change addresses that problem.  It also helps if the change is valid code that compiles and passes tests.  

This has none of those properties.  If you have a contribution to make, please try again.

PR #492, opened 2020-03-10
Use consistent spacing in toml files

martinthomson:
I have no idea how to enforce this on an ongoing basis, but having consistency now might help ensure that we keep consistency.


agrover:
Nope, the way it was was how they are formatted. Pick a random toml file in mozilla-central to compare.

If there are inconsistent spots, they should be changed to have this format.

agrover:
reverting.

PR #508, opened 2020-03-18
Force connection to be processed when Http3ServerHandler::stream_reset() was called.

vonasek:


neqo-http3/src/connection_server.rs
```diff
@@ -17,6 +17,7 @@ use std::time::Instant;
 pub struct Http3ServerHandler {
     base_handler: Http3Connection<TransactionServer>,
     events: Http3ServerConnEvents,
+    stream_was_reset: bool,
```
ddragana: can you call it needs_processing? it is more abstract.

PR #510, opened 2020-03-24
Replace OnceResult with lazy_static in neqo-transport and test-fixture

### LINKED ISSUE #61: Use lazy_static instead of OnceResult
See #45 review comments. 

btw lazy_static is already used in gecko so we should not hesitate to use it in Neqo if it's useful.

undef1nd:
Partially solves #61


agrover:
Looks good so far!

undef1nd:
@agrover as discussed, moved `once.rs` to `crypo` package

undef1nd:
YAY, happy to help)

PR #512, opened 2020-03-25
We should still send ackes when cc window is full.

### LINKED ISSUE #47: Timers that end up being set in the past
The recovery spec (A.8) does say something about  "This algorithm may result in the timer being set in the past, particularly if timers wake up late. Timers set in the past SHOULD fire immediately."

Should we just be ensuring the caller re-calls us, or should we go ahead and assert this never happens?

_Originally posted by @agrover in https://github.com/mozilla/neqo/pull/39_

Well, for our API, we can't pass out a negative `Duration` for the delay time, so if loss recovery sets a timer in the past, we will probably clamp to 0 (I hope we don't underflow).  In that case, I think that it is our responsibility to drive the state forward before we return.

ddragana:
Fixes #47


neqo-transport/src/connection.rs
```diff
@@ -1178,17 +1182,21 @@ impl Connection {
...
             } else {
+                // If avail == 0 we do not have available congestion window, we may send only
+                // non-congestion controlled frames
+                let avail = usize::try_from(self.loss_recovery.cwnd_avail()).unwrap();
                 (
                     false,
-                    usize::try_from(self.loss_recovery.cwnd_avail()).unwrap(),
+                    if avail == 0 { path.mtu() } else { avail },
                     PNSpace::Initial,
+                    avail == 0,
```
agrover: totally a style thing but you might see if a structure like `if avail == 0 { // return a tuple} else { // return a different tuple }` might be clearer than one tuple whose elements are conditional on avail.

neqo-transport/src/connection.rs
```diff
@@ -1178,17 +1182,21 @@ impl Connection {
         let mut needs_padding = false;
 
         // Check whether we are sending packets in PTO mode.
-        let (pto, cong_avail, min_pn_space) =
+        let (pto, cong_avail, min_pn_space, only_not_cc) =
```
martinthomson: I'm having real trouble with the name of this variable.  How about `cc_limited` if the goal is to not send anything but ACK when you are CC limited.  Comment to that effect on the check above.

neqo-transport/src/connection.rs
```diff
@@ -1219,7 +1227,7 @@ impl Connection {
 
             // Add frames to the packet.
             let (tokens, ack_eliciting) = if *space >= min_pn_space {
-                let r = self.add_frames(&mut builder, *space, limit, now);
+                let r = self.add_frames(&mut builder, *space, limit, only_not_cc, now);
                 if builder.is_empty() {
                     if pto {
                         // Add a PING if there is a PTO and nothing to send.
```
martinthomson: Somewhere down there, it might pay to have some debug_assert!() calls to cross-verify cc_limited and pto and cc_limited and the available space.

neqo-transport/src/connection.rs
```diff
@@ -1178,17 +1182,21 @@ impl Connection {
...
-                (true, path.mtu(), min_pto_pn_space)
+                (true, path.mtu(), min_pto_pn_space, false)
             } else {
+                // If avail == 0 we do not have available congestion window, we may send only
+                // non-congestion controlled frames
+                let avail = usize::try_from(self.loss_recovery.cwnd_avail()).unwrap();
                 (
                     false,
-                    usize::try_from(self.loss_recovery.cwnd_avail()).unwrap(),
+                    if avail == 0 { path.mtu() } else { avail },
```
martinthomson: I think that we want to avoid sending silly packets, so I would encourage you to find a constant greater than 0 to treat as the threshold for sending ACK-only frames.  Then you don't need to jump up to the MTU.

I note that we don't limit the size of ACK frames, so we won't respect this limit anyway...

neqo-transport/src/connection.rs
ddragana: cc_limited won against bypass_cc.

neqo-transport/src/connection.rs
ddragana: I put the limit to be  512 bytes.

PR #519, opened 2020-04-03
Remember smoothed_rtt from previous connection

### LINKED ISSUE #479: Remember RTT from previous connection
We can use this to seed the RTT.

undef1nd:
Closes #479

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -540,7 +540,7 @@ impl Connection {
         let mut dec = Decoder::from(token);
 
         let smoothed_rtt = match dec.decode_varint() {
-            Some(v) => v,
+            Some(v) => Duration::from_millis(v),
             _ => return Err(Error::InvalidResumptionToken),
         };
 
@@ -561,9 +561,8 @@ impl Connection {
 
         self.tps.borrow_mut().remote_0rtt = Some(tp);
 
-        if smoothed_rtt > GRANULARITY.as_millis() as u64 {
-            self.loss_recovery
-                .set_initial_rtt(Duration::from_millis(smoothed_rtt));
+        if smoothed_rtt > GRANULARITY {
+            self.loss_recovery.set_initial_rtt(smoothed_rtt);
         }
         self.set_initial_limits();
         // Start up TLS, which has the effect of setting up all the necessary
@@ -2610,22 +2609,32 @@ mod tests {
 
     #[test]
     fn remember_smoothed_rtt() {
-        const RTT: Duration = Duration::from_millis(600);
         let mut client = default_client();
         let mut server = default_server();
-
         let mut now_time = now();
-        connect_with_rtt(&mut client, &mut server, &mut now_time, RTT);
 
-        // In exchange_ticket we only send a packet from a server to a client
-        // and client won't calculate rtt from that packet, so no need to increase time here.
+        const RTT1: Duration = Duration::from_millis(130);
+        connect_with_rtt(&mut client, &mut server, &mut now_time, RTT1);
+        assert_eq!(client.loss_recovery.rtt(), RTT1);
+
         let token = exchange_ticket(&mut client, &mut server);
         let mut client = default_client();
-
         client
             .set_resumption_token(now(), &token[..])
             .expect("should set token");
-        assert_eq!(client.loss_recovery.rtt(), RTT);
+        assert_eq!(
+            client.loss_recovery.rtt(),
+            RTT1,
+            "client should remember previous RTT"
+        );
+
+        //        const RTT2: Duration = Duration::from_millis(70);
+        //        connect_with_rtt(&mut client, &mut server, &mut now_time, RTT2);
+        //        assert_eq!(
+        //            client.loss_recovery.rtt(),
+        //            RTT2,
+        //            "previous RTT should be completely erased"
+        //        );
     }
 
     #[test]
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -2432,27 +2432,50 @@ mod tests {
     }
 
     /// Drive the handshake between the client and server.
-    fn handshake(client: &mut Connection, server: &mut Connection) {
+    fn handshake(
+        client: &mut Connection,
+        server: &mut Connection,
+        now: Instant,
+        rtt: Duration,
+    ) -> Instant {
         let mut a = client;
         let mut b = server;
+        let mut now = now;
+
         let mut datagram = None;
         let is_done = |c: &mut Connection| match c.state() {
             State::Confirmed | State::Closing { .. } | State::Closed(..) => true,
             _ => false,
         };
+
         while !is_done(a) {
             let _ = maybe_authenticate(a);
-            let d = a.process(datagram, now());
+            let d = a.process(datagram, now);
             datagram = d.dgram();
+            now += rtt / 2;
             mem::swap(&mut a, &mut b);
         }
-        a.process(datagram, now());
+        a.process(datagram, now);
+        now
+    }
+
+    fn connect_with_rtt(
+        client: &mut Connection,
+        server: &mut Connection,
+        now: Instant,
+        rtt: Duration,
+    ) -> Instant {
+        let now = handshake(client, server, now, rtt);
+        assert_eq!(*client.state(), State::Confirmed);
+        assert_eq!(*client.state(), State::Confirmed);
+
+        assert_eq!(client.loss_recovery.rtt(), rtt);
+        assert_eq!(server.loss_recovery.rtt(), rtt);
+        now
     }
 
     fn connect(client: &mut Connection, server: &mut Connection) {
-        handshake(client, server);
-        assert_eq!(*client.state(), State::Confirmed);
-        assert_eq!(*server.state(), State::Confirmed);
+        connect_with_rtt(client, server, now(), Duration::new(0, 0));
     }
 
     fn assert_error(c: &Connection, err: ConnectionError) {
@@ -2477,7 +2500,7 @@ mod tests {
         .unwrap();
         let mut server = default_server();
 
-        handshake(&mut client, &mut server);
+        handshake(&mut client, &mut server, now(), Duration::new(0, 0));
         // TODO (mt): errors are immediate, which means that we never send CONNECTION_CLOSE
         // and the client never sees the server's rejection of its handshake.
         //assert_error(&client, ConnectionError::Transport(Error::CryptoAlert(120)));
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection.rs
```diff
@@ -535,6 +553,12 @@ impl Connection {
         }
         qinfo!([self], "resumption token {}", hex(token));
         let mut dec = Decoder::from(token);
+
+        let smoothed_rtt = match dec.decode_uint(4) {
```
martinthomson, line 557: Using a varint is probably better.

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2594,47 @@ mod tests {
...
+        let is_done = |c: &mut Connection| match c.state() {
+            State::Confirmed | State::Closing { .. } | State::Closed(..) => true,
+            _ => false,
+        };
+
+        while !is_done(a) {
+            let _ = maybe_authenticate(a);
+            let d = a.process(datagram, *now);
+            datagram = d.dgram();
+            *now += Duration::from_millis(20);
```
martinthomson, line 2610: Can you make a constant for this?

neqo-transport/src/recovery.rs
```diff
@@ -318,6 +318,14 @@ impl LossRecovery {
...
     }
 
+    pub fn smoothed_rtt(&self) -> Option<Duration> {
+        self.rtt_vals.smoothed_rtt
+    }
+
+    pub fn set_smoothed_rtt(&mut self, value: Duration) {
+        self.rtt_vals.smoothed_rtt = Some(value)
+    }
+
```
martinthomson, line 328: For this part, we need to not set the smoothed value, but instead overwrite the initial value (which is stored in latest_rtt).  That is currently a constant (INITIAL_RTT, right at the top of this file).

neqo-transport/src/connection.rs
```diff
@@ -506,6 +506,24 @@ impl Connection {
...
+                        Ok(smoothed_rtt) => smoothed_rtt,
+                        _ => {
+                            debug_assert!(
+                                false,
+                                "failed to cast smoothed_rtt value from ms to u32"
+                            );
+                            return None;
+                        }
+                    };
+                    enc.encode_uint(4, smoothed_rtt);
```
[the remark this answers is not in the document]
undef1nd: Thanks for the explanation. I've changed it. 

Initially I was not sure whether setting 0 as default in case if try_from fails or if smoothed_rtt is None is okay. So Dragana suggested to return None in such cases and add debug_asserts. Thus I ended up wrapping those into explicit matches.

neqo-transport/src/recovery.rs
undef1nd: Sorry, didn't get it first.
Fixed it. But I'm not sure that I did it right: I added set_latest_rtt for LossRecovery to overwrite the value in latest_rtt.

neqo-transport/src/connection.rs
```diff
@@ -506,6 +506,12 @@ impl Connection {
                 Some(ref t) => {
                     qtrace!("TLS token {}", hex(&t));
                     let mut enc = Encoder::default();
+                    let rtt = self
+                        .loss_recovery
+                        .smoothed_rtt()
+                        .map(|v| u64::try_from(v.as_millis()).unwrap_or(0))
+                        .unwrap_or(0);
```
martinthomson, line 513:
```suggestion
                        .map(|v| u64::try_from(v.as_millis()))
                        .flatten()
                        .unwrap_or(0);
```
Maybe this is clearer?  Either works.

neqo-transport/src/connection.rs
```diff
@@ -551,6 +563,8 @@ impl Connection {
         }
 
         self.tps.borrow_mut().remote_0rtt = Some(tp);
+        self.loss_recovery
+            .set_latest_rtt(Duration::from_millis(smoothed_rtt));
```
martinthomson, line 567: Given that 0 means that we don't have a smoothed RTT, it's probably best to do this:
```suggestion
        if smoothed_rtt > GRANULARITY {
          self.loss_recovery
              .set_latest_rtt(Duration::from_millis(smoothed_rtt));
        }
```
Note that this uses the timer granularity that we assume.

(Separately, we need to make that granularity 1ms.  You can avoid having to worry about that factoring in here by making the tests use an RTT estimate greater than 20ms for the moment.)

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2582,46 @@ mod tests {
         assert!(server.crypto.tls.info().unwrap().resumed());
     }
 
+    fn connect_with_rtt(client: &mut Connection, server: &mut Connection, now: &mut Instant) {
+        const TIME_INTERVAL: u64 = 20;
```
martinthomson, line 2586: I think that you should make this accessible on the outer scope and make it type `Duration`.  Maybe call it "RTT" and advance by half that value in this function.

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2582,46 @@ mod tests {
...
+        let is_done = |c: &mut Connection| match c.state() {
+            State::Confirmed | State::Closing { .. } | State::Closed(..) => true,
+            _ => false,
+        };
+
+        while !is_done(a) {
+            let _ = maybe_authenticate(a);
+            let d = a.process(datagram, *now);
+            datagram = d.dgram();
+            *now += Duration::from_millis(TIME_INTERVAL);
```
martinthomson, line 2600:
```suggestion
            *now += RTT / 2;
```

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2582,46 @@ mod tests {
...
+
+        // In exchange_ticket we only send a packet from a server to a client
+        // and client won't calculate rtt from that packet, so no need to increase time here.
+        let token = exchange_ticket(&mut client, &mut server);
+        let mut client = default_client();
+
+        client
+            .set_resumption_token(now(), &token[..])
+            .expect("should set token");
+        assert_eq!(client.loss_recovery.latest_rtt(), Duration::from_millis(40));
```
martinthomson, line 2622:
```suggestion
        assert_eq!(client.loss_recovery.latest_rtt(), RTT);
```

neqo-transport/src/recovery.rs
```diff
@@ -318,6 +318,19 @@ impl LossRecovery {
...
     }
 
+    #[cfg(test)]
+    pub fn latest_rtt(&self) -> Duration {
+        self.rtt_vals.latest_rtt
+    }
+
+    pub fn smoothed_rtt(&self) -> Option<Duration> {
+        self.rtt_vals.smoothed_rtt
+    }
```
martinthomson, line 328: The RttVals::rtt accessor contains the logic that you want to use here.  I would do this:

```suggestion
    pub fn rtt(&self) -> Duration {
        self.rtt_vals.rtt()
    }
```

And then you can drop the test-only `latest_rtt`, relying on this.  In the test you have, the RTT is fine.

neqo-transport/src/connection.rs
martinthomson: Or, you could pass the RTT as an argument.

There is a reason you might want to do that below.

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2582,46 @@ mod tests {
...
+        };
+
+        while !is_done(a) {
+            let _ = maybe_authenticate(a);
+            let d = a.process(datagram, *now);
+            datagram = d.dgram();
+            *now += Duration::from_millis(TIME_INTERVAL);
+            mem::swap(&mut a, &mut b);
+        }
+        a.process(datagram, *now);
```
martinthomson, line 2603: It is probably worth testing that the RTT is accurate here.

```suggestion
        a.process(datagram, *now);

        assert_eq!(a.loss_recovery.rtt(), RTT);
        assert_eq!(b.loss_recovery.rtt(), RTT);
```

neqo-transport/src/connection.rs
```diff
@@ -2568,6 +2582,46 @@ mod tests {
...
+        // In exchange_ticket we only send a packet from a server to a client
+        // and client won't calculate rtt from that packet, so no need to increase time here.
+        let token = exchange_ticket(&mut client, &mut server);
+        let mut client = default_client();
+
+        client
+            .set_resumption_token(now(), &token[..])
+            .expect("should set token");
+        assert_eq!(client.loss_recovery.latest_rtt(), Duration::from_millis(40));
+    }
```
martinthomson, line 2623: Once you have tested that the RTT is set when you set a resumption token, I would then test to see if that estimate affects the RTT estimate if the RTT changes.  For instance, you have a 40ms RTT simulated here.  If you make a new connection with a 50ms RTT, then it should report 50ms.  If you incorrectly added the 40ms value from the previous connection as a sample, then the rolling average will include that and you will likely get something like 48 or 49ms instead.

neqo-transport/src/recovery.rs
```diff
@@ -318,6 +318,19 @@ impl LossRecovery {
...
+
+    pub fn set_latest_rtt(&mut self, value: Duration) {
```
martinthomson, line 330: I would change the number of this function name to `set_initial_rtt()`.  Then add an assertion that you have no prior estimates.

```suggestion
    pub fn set_initial_rtt(&mut self, value: Duration) {
        debug_assert!(self.smoothed_rtt.is_none());
```

neqo-transport/src/connection.rs
undef1nd: I guess it will be reduced to 
```rust
let rtt = self.loss_recovery.rtt();
let rtt = u64::try_from(rtt.as_millis()).unwrap_or(0);
enc.encode_varint(rtt);
```
according to your comment https://github.com/mozilla/neqo/pull/519#discussion_r403774486

neqo-transport/src/connection.rs
undef1nd: I put GRANULARITY as a constant at the top of the file then.

neqo-transport/src/connection.rs
undef1nd: @martinthomson seems either I messed everything up while doing these changes...or I extending the test in a wrong way. What I've just tried to do is straight after these lines add smth like

```rust 
        // Add one more client to connect it with the new rtt
        let NEW_RTT  = Duration::from_millis(90);
        connect_with_rtt(&mut client, &mut server, &mut now_time, NEW_RTT);
        let token = exchange_ticket(&mut client, &mut server);
        // Again add one more client to check its rtt
        client
            .set_resumption_token(now(), &token[..])
            .expect("should set token");
        assert_eq!(client.loss_recovery.rtt(), NEW_RTT);
```

Suspect my understanding how to do it is wrong. Could you plz give me a hand with this? Maybe I've also broken smth with my last commit)

martinthomson: I think that what you want is something like (excuse lack of formatting):

```rust
const RTT1: Duration = Duration::from_millis(50);
const RTT2: Duration = Duration::from_millis(80);
let mut now = now();
now = connect_with_rtt(&mut client, &mut server, now, RTT1);
assert_eq!(client.loss_recovery.rtt(), RTT1);
let token = exchange_ticket(&mut client, &mut server);

client.set_resumption_token(now, &token[..]).expect("should set token");
assert_eq!(client.loss_recovery.rtt(), RTT1, "client should remember previous RTT");

now = connect_with_rtt(&mut client, &mut server, now, RTT2);
assert_eq!(client.loss_recovery.rtt(), RTT2, "previous RTT should be completely erased");
```

neqo-transport/src/connection.rs
```diff
@@ -53,6 +53,8 @@ const LOCAL_MAX_DATA: u64 = 0x3FFF_FFFF_FFFF_FFFF; // 2^62-1
 
 const MIN_CC_WINDOW: usize = 0x200; // let's not send packets smaller than 512
 
+const GRANULARITY: u64 = 1; // timer granularity in ms
```
martinthomson, line 56: There is already a constant in recovery.rs.  You can make that public and pull it in.  Sorry, I got my files mixed up.

undef1nd: Oh, thanks! Added. Had to cast it.

neqo-transport/src/connection.rs
undef1nd: Thanks. The last assert fails. So looks like it does not forget the previous RTT.

neqo-transport/src/connection.rs
```diff
@@ -551,6 +560,11 @@ impl Connection {
         }
 
         self.tps.borrow_mut().remote_0rtt = Some(tp);
+
+        if smoothed_rtt > GRANULARITY.as_millis() as u64 {
```
martinthomson, line 564:
```suggestion
        if Duration::from_millis(smoothed_rtt) > GRANULARITY {
```
Or you can take the above suggestion.

neqo-transport/src/connection.rs
```diff
@@ -535,6 +538,12 @@ impl Connection {
         }
         qinfo!([self], "resumption token {}", hex(token));
         let mut dec = Decoder::from(token);
+
+        let smoothed_rtt = match dec.decode_varint() {
+            Some(v) => v,
```
martinthomson, line 543:
```suggestion
            Some(v) => Duration::from_millis(v),
```

neqo-transport/src/connection.rs
martinthomson: Hmm, that's unexpected.  Can you push the code, maybe with that test commented out, and I'll take a look?

neqo-transport/src/connection.rs
undef1nd: Indeed. Thanks.

neqo-transport/src/connection.rs
undef1nd: @martinthomson I commented out the failing part of the test. It would be great if you could take a look at it.

martinthomson:
Thanks for giving this a good swing.  You got so close.

undef1nd:
Thanks for all the help, @martinthomson

PR #525, opened 2020-04-06
We can still have RecvStreamReadable after a stream has been read completely

ddragana:


ddragana:
This is one test that shows that bug https://bugzilla.mozilla.org/show_bug.cgi?id=1622233 can happen. There may be more cases. Http3 code relies on the fact that all RecvStreamReadable will be removed after all data has been read from a stream, but that is not true.

agrover:
Just to be clear, I think you're saying that now we have a test that shows the problem, we should be able to fix the problem so that Http3 code can continue to rely on its assumption, rather than removing the assumption (the panic()s) like this PR does. Is that what you're saying?

ddragana:
I wanted to land this one to, so that I can pull new version into m.-c. today and we can fix this later.

neqo-http3/src/connection_client.rs
```diff
@@ -3139,4 +3139,34 @@ mod tests {
         client.process(out.dgram(), now());
         assert_closed(&client, Error::HttpFrameUnexpected);
     }
+
+    #[test]
+    fn we_can_have_transort_stream_readable_event_after_we_have_read_all_data_from_a_stream() {
```
agrover, line 3144: I kinda like the long test function names that the other tests you added use, but this one is a little *too* verbose I think. Please omit "we_can_have", "we_have", "from_a_stream", and fix "transport" typo and this is good.

neqo-transport/src/events.rs
```diff
@@ -126,6 +126,11 @@ impl ConnectionEvents {
         self.insert(ConnectionEvent::ZeroRttRejected);
     }
 
+    pub fn remove_stream_readable_events(&self, stream_id: StreamId) {
```
agrover, line 129: Can we name this `recv_stream_complete` to match the existing `send_stream_complete` please?

neqo-transport/src/recv_stream.rs
```diff
@@ -505,6 +505,8 @@ impl RecvStream {
                 let bytes_read = recv_buf.read(buf)?;
                 let fin_read = recv_buf.buffered() == 0;
                 if fin_read {
+                    self.conn_events
+                        .remove_stream_readable_events(self.stream_id);
```
agrover, line 509: Let's put this call inside `set_state`? That's where we're already doing a similar thing.

neqo-transport/src/events.rs
```diff
@@ -126,6 +126,11 @@ impl ConnectionEvents {
...
+        // If stopped, no longer writable.
```
agrover, line 130: I think you mean "readable"

neqo-transport/src/recv_stream.rs
ddragana: I was thinking about it but decided not to do it to avoid calling it too many times.
It is a better place, I will change it.

PR #532, opened 2020-04-08
Remove prepending length from hex function

### LINKED ISSUE #524: hex() should just convert to hex, it shouldn't prepend the length in brackets
This is limiting its use cases. For the times we do want the length, maybe we can add a second function, maybe `hex_with_len()` or similar.

undef1nd:
Closes #524

### DIFF
--- a/neqo-common/src/lib.rs
+++ b/neqo-common/src/lib.rs
@@ -33,7 +33,7 @@ macro_rules! matches {
 
 #[must_use]
 pub fn hex(buf: &[u8]) -> String {
-    let mut ret = String::with_capacity(10 + buf.len() * 3);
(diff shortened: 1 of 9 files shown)


neqo-common/src/lib.rs
```diff
@@ -33,6 +33,15 @@ macro_rules! matches {
 
 #[must_use]
 pub fn hex(buf: &[u8]) -> String {
+    let mut ret = String::with_capacity(10 + buf.len() * 3);
```
agrover, line 36: I think you can use a smaller capacity here, maybe just buf.len() * 2.

undef1nd: Thanks. Changed.

PR #535, opened 2020-04-09
[WIP] Replace u64 for stream id with StreamId

undef1nd:
@agrover @martinthomson 
I tried to solve this issue #346 , but either I have changed too much or not enough.
Could you please look through? 

For example, there are places where it looks like the type should be StreamId: like this [Result](https://github.com/mozilla/neqo/blob/master/neqo-transport/src/connection.rs#L1936) or maybe even in some fields of this struct [Http3Connection](https://github.com/mozilla/neqo/blob/master/neqo-http3/src/connection.rs#L64)

I haven't touched tests yet. They are failing cause stream_id is expected to be u64.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -187,7 +187,7 @@ impl Handler for PostConnectHandler {
         client.process_http3(Instant::now());
         while let Some(event) = client.next_event() {
             match event {
-                Http3ClientEvent::HeaderReady { stream_id } => match self.streams.get(&stream_id) {
+                Http3ClientEvent::HeaderReady { stream_id } => match self.streams.get(&stream_id.as_u64()) {
                     Some(out_file) => {
                         let headers = client.read_response_headers(stream_id);
                         if out_file.is_none() {
@@ -201,7 +201,7 @@ impl Handler for PostConnectHandler {
                 },
                 Http3ClientEvent::DataReadable { stream_id } => {
                     let mut stream_done = false;
-                    match self.streams.get_mut(&stream_id) {
+                    match self.streams.get_mut(&stream_id.as_u64()) {
                         None => {
                             println!("Data on unexpected stream: {}", stream_id);
                             return Ok(false);
@@ -233,7 +233,7 @@ impl Handler for PostConnectHandler {
                     }
 
                     if stream_done {
-                        self.streams.remove(&stream_id);
+                        self.streams.remove(&stream_id.as_u64());
                         if self.streams.is_empty() {
                             client.close(Instant::now(), 0, "kthxbye!");
                             return Ok(false);
@@ -305,7 +305,7 @@ fn client(
             &to_headers(&args.header),
         )?;
 
-        let _ = client.stream_close_send(client_stream_id);
+        let _ = client.stream_close_send(client_stream_id.into());
 
         let out_file = if let Some(ref dir) = args.output_dir {
             let mut out_path = dir.clone();
@@ -483,7 +483,7 @@ mod old {
             while let Some(event) = client.next_event() {
                 match event {
                     ConnectionEvent::RecvStreamReadable { stream_id } => {
-                        if !self.streams.contains(&stream_id) {
+                        if !self.streams.contains(&stream_id.as_u64()) {
                             println!("Data on unexpected stream: {}", stream_id);
                             return Ok(false);
                         }
@@ -595,7 +595,7 @@ mod old {
             let client_stream_id = client.stream_create(StreamType::BiDi).unwrap();
             let req: String = "GET /10\r\n".to_string();
             client
-                .stream_send(client_stream_id, req.as_bytes())
+                .stream_send(client_stream_id.into(), req.as_bytes())
                 .unwrap();
             let mut h2 = PostConnectHandlerOld::default();
             h2.streams.insert(client_stream_id);
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -12,7 +12,7 @@ use crate::stream_type_reader::NewStreamTypeReader;
 use neqo_common::{matches, qdebug, qerror, qinfo, qtrace, qwarn};
 use neqo_qpack::decoder::{QPackDecoder, QPACK_UNI_STREAM_TYPE_DECODER};
 use neqo_qpack::encoder::{QPackEncoder, QPACK_UNI_STREAM_TYPE_ENCODER};
-use neqo_transport::{AppError, CloseError, Connection, State, StreamType};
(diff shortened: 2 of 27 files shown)


agrover:
Hi, if you haven't I'd take a look at @michiboo's PR, #347 and the comments there.

I think it's clear that this conversion touches a great many spots, in different layers, so it's hard to know when to stop: either don't switch over upper layers right away and use `as_u64` as needed, or do EVERYTHING but then the PR becomes so big that it needs frequent rebasing when other PRs land in the meantime.

I'd suggest being extremely methodical about this, and how the work is split up into multiple commits. Maybe the first commit is "Use StreamId in transport events", and that makes the one change and fixes up all the other spots as needed. Then the next commit eliminates some of the StreamId<-->u64 conversions by making something else (e.g. method arguments, container key type, etc) a StreamId, which will then cause ANOTHER wave of needed fixups. This makes it possible to maybe do a series of PRs, so at least the earlier commits you don't have to keep rebasing, or, if it stays as one PR, rebasing a series of smaller commits is easier than one huge commit.

undef1nd:
Yeah, I did look into that PR. But unfortunately, didn’t realize that it would result into THAT many changes (and still some changes to be made, especially in tests). 
Will try to choose another approach then.
Thanks for the feedback.

neqo-client/src/main.rs
```diff
@@ -595,7 +595,7 @@ mod old {
             let client_stream_id = client.stream_create(StreamType::BiDi).unwrap();
             let req: String = "GET /10\r\n".to_string();
             client
-                .stream_send(client_stream_id, req.as_bytes())
+                .stream_send(client_stream_id.into(), req.as_bytes())
```
martinthomson, line 598: This function can probably take StreamId.

neqo-http3/src/connection.rs
```diff
@@ -130,14 +130,14 @@ impl<T: Http3Transaction> Http3Connection<T> {
     fn create_qpack_streams(&mut self, conn: &mut Connection) -> Res<()> {
         qdebug!([self], "create_qpack_streams.");
         self.qpack_encoder
-            .add_send_stream(conn.stream_create(StreamType::UniDi)?);
+            .add_send_stream(conn.stream_create(StreamType::UniDi)?.into());
```
martinthomson, line 133: Having qpack know about StreamId isn't ideal, but I think that it probably should in the end.  Or, we could make these qpack public functions generic on `<T: Into<u64>>`.

neqo-http3/src/connection.rs
```diff
@@ -496,16 +500,16 @@ impl<T: Http3Transaction> Http3Connection<T> {
...
     }
 
-    pub fn stream_close_send(&mut self, conn: &mut Connection, stream_id: u64) -> Res<()> {
+    pub fn stream_close_send(&mut self, conn: &mut Connection, stream_id: StreamId) -> Res<()> {
         qinfo!([self], "Close sending side for stream {}.", stream_id);
         debug_assert!(self.state_active() || self.state_zero_rtt());
         let transaction = self
             .transactions
-            .get_mut(&stream_id)
+            .get_mut(&stream_id.as_u64())
```
martinthomson, line 508: transactions can be keyed on `StreamId`.

neqo-http3/src/connection_client.rs
```diff
@@ -426,30 +434,32 @@ impl Http3Client {
...
-    fn handle_goaway(&mut self, goaway_stream_id: u64) -> Res<()> {
+    fn handle_goaway(&mut self, goaway_stream_id: StreamId) -> Res<()> {
         qinfo!([self], "handle_goaway");
         // Issue reset events for streams >= goaway stream id
         for id in self
             .base_handler
             .transactions
             .iter()
-            .filter(|(id, _)| **id >= goaway_stream_id)
+            .filter(|(id, _)| **id >= goaway_stream_id.as_u64())
```
martinthomson, line 452: This is where you have to consider implementing `PartialCmp` for `StreamId`.  I think that's not a function you want to have implemented on the struct, so you would need to unbox in this function.

neqo-http3/src/connection.rs
undef1nd: @martinthomson Thanks for the comments!
In many case I just used `as_u64` to make the code compile till I get back to these places: I was not sure whether I needed to touch collections or not. 


I did too many changes at once and it's kinda difficult to make sense of them now. I was thinking about closing this PR and starting from scratch: maybe trying to localize changes within a single package first and temporary fixing other places with `as_u64` where stream_id is expected to be u64.

undef1nd: There was also one specific place I didn't know how to deal with.
If I change [these fields](https://github.com/mozilla/neqo/blob/master/neqo-qpack/src/decoder_instructions.rs#L19-L20) to `StreamId`, but leave `InsertCountIncrement { increment: u64 }` as is, how do I go about [this match](https://github.com/mozilla/neqo/blob/master/neqo-qpack/src/decoder_instructions.rs#L111) without adding arm for InsertCountIncrement separately?

martinthomson: Eww, you are right, that's ugly.  It might be a case of this being too hard and we can factor this differently, at a small cost in ugliness:

```rust
                        qtrace!([self], "varint read {}", val);
                        let set_none = |dec: &mut Self| {
                            dec.state = DecoderInstructionReaderState::ReadInstruction;
                            Ok(Some(mem::replace(
                                &mut self.instruction,
                                DecoderInstruction::NoInstruction,
                            )))
                        };
                        match &mut self.instruction {
                            DecoderInstruction::InsertCountIncrement { increment: v } => {
                                *v = val;
                                break set_none(&mut self);
                            }
                            DecoderInstruction::HeaderAck { stream_id: v }
                            | DecoderInstruction::StreamCancellation { stream_id: v } => {
                                *v = StreamId::from(val);
                                break set_none(&mut self);
                            }
                            _ => unreachable!(...);
                        }
```

Or you could avoid this weird code and use the simple version instead:

```rust
                        qtrace!([self], "varint read {}", val);
                        let ret = match self.instruction {
                            DecoderInstruction::InsertCountIncrement { .. } =>
                                DecoderInstruction::InsertCountIncrement { increment: val },
                            DecoderInstruction::HeaderAck { stream_id: v } =>
                                DecoderInstruction::HeaderAck { stream_id: StreamId::from(val) },
                            DecoderInstruction::StreamCancellation { stream_id: v } =>
                                DecoderInstruction::StreamCancellation { stream_id: StreamId::from(val) },
                            _ => unreachable!(...);
                        };
                       self.instruction = DecoderInstruction::NoInstruction;
                       ret
```

undef1nd: @martinthomson thanks a lot!

PR #536, opened 2020-04-09
Merge process_timers and process_output.

### LINKED ISSUE #159: Fold process_timer() into process_output()
see https://github.com/mozilla/neqo/issues/75#issuecomment-519387255

In thinking about this one, I think that all we need is two functions:

1.  absorb packet (called process_input today)
2.  do something that might trigger actions from the caller

The Output enum accurately captures those actions: nothing, wait, send a packet.

Now all you need to do as a caller is call 1 if you get a packet, or call 2 when you are done with the task neqo gave you. I think that's clean and simple. From a caller perspective, I don't see what is different in responding to a timeout as opposed to asking for more actions.

Internally, that means that we should be running timer hooks on process_output.

ddragana:
Fixed #159

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -153,10 +153,7 @@ fn process_loop(
         }
 
         match socket.recv(&mut buf[..]) {
-            Err(ref err) if err.kind() == ErrorKind::WouldBlock => {
-                // timer expired
-                client.process_timer(Instant::now());
-            }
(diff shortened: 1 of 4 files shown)


agrover:
I'm not in favor of doing this, as I explained in #159. What is the rationale?

agrover:
Ok I'm on board. :+1:  Please feel free to merge after rebasing/deconflicting. Thanks!

PR #538, opened 2020-04-10
Check if transaction is done(), do not check only fin before removing a transaction

ddragana:


agrover:
Is this fixing a bug? I could use a little more context because I'm not seeing the bug in the current code..

ddragana:
> Is this fixing a bug? I could use a little more context because I'm not seeing the bug in the current code..
> 
> EDIT: ok, we can't remove the transaction if sending is not complete, even if we've received the fin.


Yes, we should not remove transaction if we sill sending request.

> EDIT2: I'm still not sure why fin needs to be false to do the second branch..?

If fin is true the stream is completely read, there is nothing to read. The second branch is only reading from a single stream.

Issue #505, opened 2020-03-17
panic in recovery.rs check_loss_detection_timeout


ddragana:
https://firefoxci.taskcluster-artifacts.net/MSIbvMvgQoWSbG5vRCnSDQ/0/public/logs/live_backing.log

vonasek:
I've hit this locally and have a full log:
[http3server.log](https://github.com/mozilla/neqo/files/4446737/http3server.log)

```
#0  core::option::expect_failed () at src/libcore/option.rs:1188
#1  0x0000555f08519bc3 in core::option::Option<T>::expect (self=..., msg=...) at /rustc/f3e1a954d2ead4e2fc197c7da7d71e6c61bad196/src/libcore/option.rs:348
#2  0x0000555f0857c6f8 in neqo_transport::recovery::LossRecovery::check_loss_detection_timeout (self=0x555f0a45f9d0, now=...)
    at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/recovery.rs:679
#3  0x0000555f08527791 in neqo_transport::connection::Connection::process_timer (self=0x555f0a45f278, now=...)
    at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/connection.rs:654
#4  0x0000555f0852914f in neqo_transport::connection::Connection::process (self=0x555f0a45f278, dgram=..., now=...)
    at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/connection.rs:755
#5  0x0000555f084e9797 in neqo_transport::server::Server::process_connection (self=0x7ffcddf92fc8, c=..., dgram=..., now=...)
    at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/server.rs:245
#6  0x0000555f084ecf78 in neqo_transport::server::Server::process_next_output (self=0x7ffcddf92fc8, now=...)
    at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/server.rs:398
#7  0x0000555f084ee08e in neqo_transport::server::Server::process::{{closure}} () at /mnt/work/opt/moz/hg-central-2/third_party/rust/neqo-transport/src/server.rs:425
#8  0x0000555f0851c258 in core::option::Option<T>::or_else (self=..., f=...) at /rustc/f3e1a954d2ead4e2fc197c7da7d71e6c61bad196/src/libcore/option.rs:754
[...]

PR #545, opened 2020-04-15
Use the minimum idle_timeout value from both peers

martinthomson:
We have a fixed idle timeout, but we aren't using the value from the
peer to set our actual timeout.

This does that.  It chooses the minimum value from both endpoints.  But
it also is careful not to go lower than 3x the current PTO, which might
otherwise cause things to go pear-shaped.


ddragana:
@agrover, please look at this when you try to fix, understand #505. IdleTimeout may change because it depends on PTO. I want that all corner cases with timers are well understood.

PR #547, opened 2020-04-15
Turn on clippy pedantics for neqo-http3.

ddragana:
also Remove HttpEarlyResponse error code

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -4,6 +4,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#![allow(clippy::module_name_repetitions)]
+
 use crate::connection::Http3State;
 use neqo_common::matches;
 use neqo_transport::{AppError, StreamType};
@@ -44,19 +46,23 @@ pub struct Http3ClientEvents {
 }
 
 impl Http3ClientEvents {
-    pub fn header_ready(&self, stream_id: u64) {
+    /// Add a new `HeaderReady` event.
+    pub(crate) fn header_ready(&self, stream_id: u64) {
         self.insert(Http3ClientEvent::HeaderReady { stream_id });
     }
 
-    pub fn data_writable(&self, stream_id: u64) {
+    /// Add a new `DataWritable` event.
+    pub(crate) fn data_writable(&self, stream_id: u64) {
         self.insert(Http3ClientEvent::DataWritable { stream_id });
     }
 
-    pub fn data_readable(&self, stream_id: u64) {
+    /// Add a new `DataReadable` event
+    pub(crate) fn data_readable(&self, stream_id: u64) {
         self.insert(Http3ClientEvent::DataReadable { stream_id });
     }
 
-    pub fn stop_sending(&self, stream_id: u64, error: AppError) {
+    /// Add a new `StopSending` event
+    pub(crate) fn stop_sending(&self, stream_id: u64, error: AppError) {
         // Remove DataWritable event if any.
         self.remove(|evt| {
             matches!(evt, Http3ClientEvent::DataWritable {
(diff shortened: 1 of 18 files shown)


neqo-http3/src/transaction_server.rs
```diff
@@ -4,6 +4,9 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+// because of receive() function.
+#![allow(clippy::too_many_lines)]
```
martinthomson, line 8: Can we put this on the function directly?  And open a bug to determine whether fixing it is worthwhile?

neqo-http3/src/transaction_client.rs
```diff
@@ -122,15 +123,15 @@ enum TransactionSendState {
...
+pub(crate) struct TransactionClient {
     send_state: TransactionSendState,
     response_stream: ResponseStream,
     stream_id: u64,
     conn_events: Http3ClientEvents,
 }
 
 impl TransactionClient {
-    pub fn new(
+    pub(crate) fn new(
```
martinthomson, line 134: If the type is `pub(crate)`, then I think that we should just use the undecorated `pub` for methods.

neqo-http3/src/server.rs
```diff
@@ -35,6 +37,9 @@ impl ::std::fmt::Display for Http3Server {
 }
 
 impl Http3Server {
+    /// # Errors
+    /// Making a `neqo-tracaction::connection` may produce an error. This can only be a crypto error if
```
martinthomson, line 41: I'm still not sure that this is correct either.  There is no `neqo-transaction`, and if there was, it would be `neqo_transaction`.

```suggestion
    /// Making a `neqo-transaction::connection` may produce an error. This can only be a crypto error if
```

agrover: pretty sure this is meant to be "transport"

agrover:
That was easy??? A truly huge diff. Even if it tells you what it wants, you still have to fix each spot...

:+1:

martinthomson:
I didn't say easy.  I said easier than expected. :)

neqo-http3/src/server.rs
ddragana: this was suppose to be transport :)

neqo-http3/src/client_events.rs
```diff
@@ -44,19 +46,23 @@ pub struct Http3ClientEvents {
 }
 
 impl Http3ClientEvents {
-    pub fn header_ready(&self, stream_id: u64) {
+    /// Add a new `HeaderReady` event.
+    pub(crate) fn header_ready(&self, stream_id: u64) {
```
martinthomson, line 50: Are these still `pub(crate)` because the type is being exported outside of neqo?

neqo-http3/src/connection_client.rs
```diff
@@ -35,6 +35,9 @@ impl ::std::fmt::Display for Http3Client {
 }
 
 impl Http3Client {
+    /// # Errors
+    /// Making a `neqo-tracaction::connection` may produce an error. This can only be a crypto error if
```
martinthomson, line 39:
```suggestion
    /// Making a `neqo-transport::connection` may produce an error. This can only be a crypto error if
```

neqo-http3/src/connection_client.rs
```diff
@@ -2193,7 +2239,7 @@ mod tests {
         let e = client.events().next().unwrap();
         if let Http3ClientEvent::DataReadable { stream_id } = e {
             assert_eq!(stream_id, request_stream_id);
-            let mut buf = [0u8; 100];
+            let mut buf = [0_u8; 100];
```
martinthomson, line 2242: No need to change this now, but you probably only need to do this:

```suggestion
            let mut buf = [0; 100];
```

neqo-http3/src/connection_client.rs
```diff
@@ -2664,15 +2711,15 @@ mod tests {
         check_control_qpack_streams(&mut server);
 
         // Check that we can send a request and that the stream_id starts again from 0.
-        let request_stream_id = make_request(&mut client, false);
-        assert_eq!(request_stream_id, 0);
+        let request_stream_id_1 = make_request(&mut client, false);
+        assert_eq!(request_stream_id_1, 0);
```
martinthomson, line 2715:
```suggestion
        assert_eq!(make_request(&mut client, false), 0);
```

neqo-http3/src/client_events.rs
ddragana: yes, this as setters and they are only set by code in neqo-http3

Issue #513, opened 2020-03-25
Ack timers for discharged pn spaces


ddragana:
In the current implementation, it cannot happen that the ack_timer is set for a space for which we have discharged keys. I am afraid that we will reorder something in the code and produce a bug that we have a timer but we do not have keys. Such a state will cause us to return timer == 0ms until the end of a connection. This is not good.
We should make sure to disable an ack_timer when we remove keys.

martinthomson:
Note also that there are cases that require us to send when we have disposed of Initial keys.  That's massively awkward for us (that might be a separate bug).

PR #550, opened 2020-04-16
Initial cut at pacing

martinthomson:
This adds a packet pacer.

It also changes the default state for a server connection to Init.  I think that will help us with handling a multi-datagram client Initial if we need to.

This currently has no tests, so I'm not there yet, but I ran out of hours today.


ddragana:
I started to work on this... we should better coordinate.

neqo-transport/src/connection.rs
```diff
@@ -1230,9 +1253,18 @@ impl Connection {
                     false,
                     self.loss_recovery.cwnd_avail(),
                     PNSpace::Initial,
-                    false,
+                    paced,
                 )
             };
```
ddragana: I was thinking about this already before:
I do not like how we calculate recovery timers state, next_delay  and now "do not send because of pacing" in different functions. I would like that someone takes a look at that all.

Here I do not like that we need to return SendOption. Can we maybe extract part starting with "et (pto, cong_avail, min_pn_space, cc_limited) =" ? and combine it with next_delay? (I do not have a concrete solution, I would need o look at the code). This can be salved in a separate issue.

martinthomson: Not sure that I completely understand your comment here.

But if it is to point out that process_output now checks timers, then sends a packet (and checks timers as needed in that process), then finally checks timers again, that's a good observation.  That's a good justification for merging process_timer into process_output, because now we can start to simplify.

It seems to me that what you want to do is check all of our timers, and save anything that might affect the sending of packets or the final callback time we might return.  Pacing is a little different in that it doesn't cause an event to be generated; it's enough to call it opportunistically.  (More so as pacing only really applies if you aren't CC limited or sending a PTO.)

That's probably worth doing as a follow-up either way.  It would be good to reduce the number of times we call into the loss recovery code.

martinthomson:
This is failing now because of #574.  I'll need some more time to work out how to factor this, maybe based on the discussion there.

martinthomson:
OK, I have added a few things to this patch based on refactoring and rebasing.

1. This now sends ACKs when there is a PTO for a higher PN space.  We didn't do that before, but #513 demonstrated that if we don't allow ACKs to be sent, we crash.

2. The send profile and its usage is greatly simplified.

3. We send a PING in every space when there is a PTO for any space.  Based on discussion with others, it seems like people will likely interpret a PING as requiring immediate acknowledgment (new issue coming), so this is good.  The resulting code is now a lot simpler.

Issue #555, opened 2020-04-20
Handshake fails with custom client


piano-man:
I'm using a custom client to connect to the neqo server.
However, the handshake never completes and this is what I get on the server side :-

_Event: StateChange { conn: ActiveConnectionRef { c: RefCell { value: ServerConnectionState { c: Server Connection: Closing { error: Transport(CryptoAlert(120)), frame_type: 6, msg: "CryptoAlert(120)", timeout: Instant { tv_sec: 21793, tv_nsec: 997789700 } } None, last_timer: Instant { tv_sec: 21793, tv_nsec: 997789700 } } } }, state: Closing(Transport(376)) }_

Are there any obvious reasons this could be happening ?
NOTE :- The client being used successfully completes handshakes with several other QUIC Implementations.

agrover:
Just to be clear, neqo-server or neqo-http3-server?

piano-man:
neqo-http3-server

martinthomson:
`Transport(376)` is a failure to agree on ALPN.

That means that you are likely not advertising a value that we like for the application protocol name.  Tip of tree insists on 'h3-27'.

piano-man:
Ah, I updated the alpn to h3.
I now run into this :- 

thread 'main' panicked at 'assertion failed: `(left == right)`                                                                                                                
left: `20`,                                                                                                                                                                
right: `22`', neqo-transport/src/crypto.rs:159:13                                                                                                                          
stack backtrace:                                                                                                                                                               
0:     0x7f0ac6419294 - backtrace::backtrace::libunwind::trace::heb43798aede8bd30                                                                                                                       
at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-
0.3.40/src/backtrace/libunwind.rs:88                                               1:     0x7f0ac6419294 - 
backtrace::backtrace::trace_unsynchronized::had2ba7dec4bd2732                                                                                                                   
at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66                                                     
2:     0x7f0ac6419294 - std::sys_common::backtrace::_print_fmt::hda61f46e822731b2                                                                                                                       
at src/libstd/sys_common/backtrace.rs:84                                                                                                        
3:     0x7f0ac6419294 - <std::sys_common::backtrace::_print::DisplayBacktrace as 
core::fmt::Display>::fmt::hfe37fa5de6572965                                                                            at 
[...]

PR #556, opened 2020-04-21
Factor more of the crypto handshake

martinthomson:
This moves the handshake stuff into crypto.rs.

I also added a proper constant for the content type, which we had fixed
to a literal 22 in a few places.

I removed the generic implementation of `Into<Epoch>` from PNSpace as that was very specific to the way that we are using this.  That is now part of the handshake method in crypto.rs.


martinthomson:
This will change the way that we react to the error in https://github.com/mozilla/neqo/issues/555#issuecomment-616874345 but there is more to do to fix that properly.

PR #558, opened 2020-04-21
Sanitize CONNECTION_CLOSE from application

martinthomson:
This recent requirement ensures that application-specific details don't leak.

I tested this by having the connection close.  I also tweaked the server so that goes from an initial state to closed without going to closing.  That avoids creating extra state on servers that receive bogus Initial packets.


agrover:
maybe wait to rebase this again until after #583 goes in, since that will likely generate more conflicts.

martinthomson:
More rebasing.

agrover:
Changing `close()` to take an owned string seems unrelated to other, more worthwhile changes (and is also a source of conflicts). Maybe drop that, or at least give it its own commit?

martinthomson:
Superseded by #625.

Issue #507, opened 2020-03-18
DataReady events for removed streams


ddragana:
neqo-htp3 may hold DataRead events for already deleted streams.
app reads DataReady event -> reads some data, not all -> neqo will post another DataReady event -> app reads again from the stream and does not call next_event. this time it reads everything so stream will be deleted, but DataReady event is still in the event queue. 

neqo-htttp3 sould not post that second event I think.

ddragana:
fixed by #567

PR #570, opened 2020-04-22
A transaction should try to read the next frame when it is done reading a data frame

### LINKED ISSUE #568: read_response_data should not call receive, it should be handled by a transaction
This [line](https://github.com/mozilla/neqo/blob/45b963f41bd8ed6a44f2ab9b2b8afb436fc9ed1c/neqo-http3/src/connection_client.rs#L219) should not be needed.

ddragana:
fixes #568

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -153,14 +153,4 @@ impl Http3ClientEvents {
                 | Http3ClientEvent::StopSending { stream_id: x, .. } if *x == stream_id)
         });
     }
-
-    /// Remove header/data ready events for a stream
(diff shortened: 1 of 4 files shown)


ddragana:
This is also a better fix for #507

agrover:
This appears to be undoing #567, is that what you intended?

ddragana:
> This appears to be undoing #567, is that what you intended?

yes,

ddragana:
If a user of http/3 uses it properly (calls process, next_event in the way we assume it should be) with this patch #567 is not needed, but if it does not hen we need this. We should assume later, so I will undo removing #567

Issue #569, opened 2020-04-22
HeaderReady event should have headers as well


ddragana:
we do not need function read_response_headers.

PR #572, opened 2020-04-23
Make HeaderReady hold headers

ddragana:


ddragana:
Fixes #569

neqo-http3/src/client_events.rs
```diff
@@ -17,7 +18,11 @@ use std::rc::Rc;
 #[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone)]
 pub enum Http3ClientEvent {
     /// Space available in the buffer for an application write to succeed.
-    HeaderReady { stream_id: u64 },
+    HeaderReady {
+        stream_id: u64,
+        headers: Option<Vec<Header>>,
```
agrover, line 23: I feel like we've talked about this before but I don't remember. Do we need the `Option` here? If no headers are returned then isn't an empty Vec a reasonable way to represent that? 

Other than that, :+1: . I like attaching the data to the event.

ddragana: Let's resolve this in a follow up.
I am returning very verbose information here. We can have:
1) header frame that is empty -> that would be empty Vec
2) no header frame just a closed stream -> None

We probably do not need to report a second one. I would like to do that in a follow up. This one just mapped existing behavior.

agrover: ok cool.

PR #587, opened 2020-04-29
Enforce a correct stream id in a GOAWAY frame

ddragana:
Ensure GOAWAY frame stream_id is not increasing
 the stream_id must be a bidi client-initiated stream

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -16,7 +16,7 @@ use neqo_common::{
 };
 use neqo_crypto::{agent::CertificateInfo, AuthenticationStatus, SecretAgentInfo};
 use neqo_transport::{
-    AppError, Connection, ConnectionEvent, ConnectionIdManager, Output, StreamType,
+    AppError, Connection, ConnectionEvent, ConnectionIdManager, Output, StreamId, StreamType,
 };
 use std::cell::RefCell;
 use std::net::SocketAddr;
@@ -167,7 +167,7 @@ impl Http3Client {
         );
         // Requests cannot be created when a connection is in states: Initializing, GoingAway, Closing and Closed.
         match self.base_handler.state() {
-            Http3State::GoingAway | Http3State::Closing(..) | Http3State::Closed(..) => {
+            Http3State::GoingAway(..) | Http3State::Closing(..) | Http3State::Closed(..) => {
                 return Err(Error::AlreadyClosed)
             }
             Http3State::Initializing => return Err(Error::Unavailable),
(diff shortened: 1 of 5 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -1952,6 +1969,100 @@ mod tests {
...
+
+        // First send Goaway with an higher number
+        let _ = server
+            .conn
+            .stream_send(server.control_stream_id.unwrap(), &[0x7, 0x1, 0x8]);
+
+        let out = server.conn.process(None, now());
+        client.process(out.dgram(), now());
+
+        // Check that there is one reeset for stream_id 8
```
martinthomson, line 1993:
```suggestion
        // Check that there is one reset for stream_id 8
```

neqo-http3/src/connection_client.rs
```diff
@@ -1952,6 +1969,100 @@ mod tests {
...
+                    stream_reset += 1;
+                }
+                _ => {}
+            }
+        }
+
+        assert_eq!(stream_reset, 1);
+        assert_eq!(client.state(), Http3State::GoingAway(8));
+
+        // Server sends anothe GOAWAY
```
martinthomson, line 2009:
```suggestion
        // Server sends another GOAWAY
```

neqo-http3/src/connection_client.rs
```diff
@@ -472,7 +472,27 @@ impl Http3Client {
...
+            return Err(Error::HttpId);
+        }
+
+        match self.base_handler.state {
+            Http3State::Connected => {
+                self.base_handler.state = Http3State::GoingAway(goaway_stream_id);
+            }
+            Http3State::GoingAway(ref mut stream_id) => {
+                if goaway_stream_id > *stream_id {
+                    return Err(Error::HttpId);
```
martinthomson, line 488: Might be worth a new code for this.

martinthomson: Also, you don't test this branch.  You should.

neqo-http3/src/connection_client.rs
```diff
@@ -1952,6 +1969,100 @@ mod tests {
...
+                Http3ClientEvent::Reset { stream_id, error } => {
+                    assert_eq!(stream_id, request_stream_id_2);
+                    assert_eq!(error, Error::HttpRequestRejected.code());
+                    stream_reset += 1;
+                }
+                _ => {}
+            }
+        }
+
+        assert_eq!(stream_reset, 2);
```
martinthomson, line 2048: That `stream_reset` is cumulative threw me at first.

neqo-http3/src/connection_client.rs
ddragana: My keyboard starts to have problems with e as well. It started to write double e. (and it already have the same problem with 't') this is not good.

neqo-http3/src/connection_client.rs
ddragana: yes, I need test, sorry it slipped.
Te spec does not define an error for this. I want to write a spec issue (It was just too late for me yesterday)

Issue #539, opened 2020-04-10
Take another look at encoder.rs, decoder.rs and table.rs


ddragana:

mahakbansal2019:
hey @ddragana  !!
I would like to work on this 
Could you please explain what I need to do here?
Thanks

ddragana:
I do not know what exacly needs to be done here. The issue is to actually figure out that to do.

ddragana:
Closing this. There was some refactoring of this code already (#616 #598 #617 #611 #612 #614 )

PR #598, opened 2020-05-04
Improve table.rs

ddragana:
- define more descriptive errors,
- remove QPackSide,
-  do not use acked_inserts_cnt on the decoder side,
-  add comments,
-  define some struct pub(crate),
-  and some other small improvements.

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -60,10 +60,16 @@ impl QPackEncoder {
     /// a `SETTINGS_QPACK_MAX_TABLE_CAPACITY` setting parameter.
     /// # Errors
     /// `EncoderStream` if value is too big.
+    /// `ChangeCapacity` if table capacity cannot be reduced.
     pub fn set_max_capacity(&mut self, cap: u64) -> Res<()> {
         if cap > (1 << 30) - 1 {
             return Err(Error::EncoderStream);
         }
+
+        if cap == self.table.capacity() {
+            return Ok(());
+        }
+
         qdebug!(
             [self],
             "Set max capacity to {} {}.",
@@ -127,7 +133,9 @@ impl QPackEncoder {
     }
 
     fn insert_count_instruction(&mut self, increment: u64) -> Res<()> {
-        self.table.increment_acked(increment)?;
+        self.table
+            .increment_acked(increment)
+            .map_err(|_| Error::DecoderStream)?;
         self.recalculate_blocked_streams();
         Ok(())
     }
@@ -206,6 +214,7 @@ impl QPackEncoder {
     /// The function can return transport errors.
     /// `HeaderLookup` if `InsertWithNameRefStatic`, `InsertWithNameRefDynamic` or `Duplicate` is used and the
     /// reference entry cannot be found.
+    /// `DynamicTableFull` if the dynamic table does not have enough space for the entry.
     pub fn insert(&mut self, conn: &mut Connection, instruction: &EncoderInstruction) -> Res<u64> {
         qdebug!([self], "insert instruction {:?}.", instruction);
         self.send(conn)?;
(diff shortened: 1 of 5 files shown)


ddragana:
This is based on #576

ddragana:
This is part 1 fro fixing #539.

neqo-qpack/src/encoder.rs
```diff
@@ -438,6 +460,36 @@ mod tests {
         }
     }
 
+    fn connect_flow_control(max_data: u64) -> TestEncoder {
```
martinthomson, line 463: This is a lot of duplicate code.  Maybe you should have a baseline function like:

```rust
fn connect_generic<F>(f: F) -> TestEncoder 
where F : FnOnce(&mut Connection,&mut Connection) {
        let mut conn = default_client();
        let mut peer_conn = default_server(); 
        f(&mut conn, &mut peer_conn);
        
         // ...
}
```
This could then be:
```rust
fn connect_fc(max_data: u64) -> TestEncoder {
  connect_generic(|client, server| {
     server.set_local_tparam(...);
  })
```

neqo-transport/src/connection.rs
```diff
@@ -1478,6 +1478,11 @@ impl Connection {
         }
     }
 
+    // TODO: this should be #[cfg(test)], but it is not acceptable from a different crate.
+    pub fn handle_max_data_test(&mut self, maximum_data: u64) {
+        self.handle_max_data(maximum_data);
+    }
```
martinthomson, line 1484: We need a different way to handle this.

What should be happening in your test is that the client hits the flow control limit and sends a STREAM_DATA_BLOCKED frame.  The server should respond by increasing the flow control limits.  That should suffice for your test, meaning that you don't need this function at all.

neqo-qpack/src/table.rs
```diff
@@ -57,63 +58,67 @@ impl DynamicTableEntry {
 }
 
 #[derive(Debug)]
-pub struct HeaderTable {
-    qpack_side: QPackSide,
+pub(crate) struct HeaderTable {
```
martinthomson, line 61: Is this change necessary?  That is, is there anything in this module that needs to be used outside of the crate?

neqo-http3/src/transaction_client.rs
```diff
@@ -54,19 +54,25 @@ impl Request {
         r
     }
 
-    fn ensure_encoded(&mut self, encoder: &mut QPackEncoder, stream_id: u64) {
+    fn ensure_encoded(
```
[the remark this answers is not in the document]
ddragana: fine by me, I will change it.

neqo-transport/src/connection.rs
ddragana: I will leave it in this PR and change it #599.

neqo-qpack/src/table.rs
ddragana: No, there is nothing that uses this outside crate, that is why I limited its scope.

ddragana:
> There are lots of `map_err` calls in this change. It seems that the header table itself can't be the final decider on what the error code should be. Is it possible to build a single error code mapping function at a higher layer? It might need to be slightly smarter (i.e., use `match` to limit what errors it changes), but it might help contain this proliferation of error code mappings that seem brittle and hard to maintain.

Previously (long time ago) I tried to make decision in table.rs, but that was not working very nice.
encoder, decoder and header_block must set the final error.
I find it nicer to explicitly map errors, because the mapping is usually per function, e.g. execute_instruction will always return only EncoderStream.
I think I can reduce number of map_err or even have a mapping function, but I need to look at decoder.rs and encoder.rs first. This is just one step in making neqo-qpack final. I did not want to make a huge PR therefore this split, small steps. We can leave this open am merge it with other PR that are coming.

neqo-transport/src/connection.rs
ddragana: #599 has been updated.

PR #599, opened 2020-05-04
 Send DATA_BLOCKED and/or STREAM_DATA_BLOCKED if qpack cannot write complete instructions into a treansport stream

ddragana:


ddragana:
this fix a todo from #576 and it is based on that PR.

neqo-transport/src/send_stream.rs
```diff
@@ -699,6 +699,22 @@ impl SendStream {
...
+            self.flow_mgr
+                .borrow_mut()
+                .stream_data_blocked(self.stream_id, self.max_stream_data);
+        }
+
+        if self.flow_mgr.borrow().conn_credit_avail() < len {
+            self.flow_mgr.borrow_mut().data_blocked();
+        }
+        false
+    }
```
agrover, line 717: `SendStream::avail()` should do what you want -- call that and then compare against len

ddragana: I do that in the beginning of the function, the rest is to find out what is blocking and send the data_block stream_data_blocked frame

agrover: oh ok, hah, sorry.

agrover: Let's break this up just a little, since this function is kind of doing two things. In h3 code use `Connection::stream_avail_send_space()` (which already exists) and then if that is not enough compared to the needed len, call a new function -- `Connection::stream_needs_send_space(len)` that queues the fc frames as needed -- basically the last 2/3 of this function.

agrover: It might send partial bytes, and the upper layer doesn't want to deal with that.

ddragana: I could make an stream_send_atomic, but this will complicate code in qpack.

Currently qpack checks if it can send an instruction, if it cannot it just continues without adding anything into a table. If is can send the instruction it tries to add it into the dynamic table and if it cannot it just continues without adding anything into a table. Otherwise it adds it to the table and send an instruction.

If we want stream_send_atomic we will need to check if we can add a element to the dynamic table, but we should not add it, and than call stream_send_atomic. Only if stream_send_atomic succeeds we add the element to the dynamic table. My thoughts was that it it faster to check credits than calling inserts to the table twice, because an insert may need to perform a lookup and also walk the table a bit to see if it can evict enough elements.

But thinking about it, it is probably not that bad and maybe it is better to keep the code that is doing: "I try to do something, just to check if it would work" then do something else , then "I really do what I wanted" in one module.

I will update the PR

ddragana:
I will closed this one. I made a new PR #616

PR #600, opened 2020-05-05
Fix UdpSocket::recv() error handling

vonasek:


agrover:
Won't this cause ctrl-C to stop working as a way to quit the program?

vonasek:
> Won't this cause ctrl-C to stop working as a way to quit the program?

I tested it and ctrl-c still works. It obviously doesn't cause the function to throw error Interrupted. Instead the default SIGINT handler kills the process.

PR #601, opened 2020-05-05
Close a connection if critical streams are reset or a stop_sending is received

ddragana:


neqo-http3/src/connection.rs
```diff
@@ -323,6 +323,46 @@ impl<T: Http3Transaction> Http3Connection<T> {
         }
     }
 
+    pub fn is_critical_stream(&self, stream_id: u64) -> bool {
+        if let Some(id) = self.qpack_encoder.local_stream_id() {
```
martinthomson, line 327: There is a clever way of doing this, in case you were interested...

```rust
self.qpack_encoder.local_stream_id().iter().chain(
  self.qpack_encoder.remote_stream_id().iter()
) // .chain ...
.any(|id| stream_id == id)
```

ddragana: Thanks. I like it. I like how much I have learned rust with your comments.

neqo-http3/src/connection.rs
```diff
@@ -323,6 +323,25 @@ impl<T: Http3Transaction> Http3Connection<T> {
...
+                    self.qpack_decoder.local_stream_id().iter().chain(
+                        self.qpack_decoder.remote_stream_id().iter().chain(
+                            self.control_stream_local
+                                .stream_id()
+                                .iter()
+                                .chain(self.control_stream_remote.stream_id().iter()),
+                        ),
+                    ),
+                ),
+            )
```
martinthomson, line 341: Rather than nesting, how about:

```suggestion
            .chain(self.qpack_encoder.remote_stream_id().iter())
            .chain(self.qpack_decoder.local_stream_id().iter())
            .chain(self.qpack_decoder.remote_stream_id().iter())
            .chain(self.control_stream_local.stream_id().iter())
            .chain(self.control_stream_remote.stream_id().iter())
```

Issue #546, opened 2020-04-15
Enforce maximum values for stream offsets and stream ID limits


martinthomson:
STREAM and CRYPTO frames could include unreferenceable data if the sum of the offset and length exceed 2<sup>62</sup>-1.  We should check that.

MAX_STREAMS or the max_stream_limit transport parameter can encode values larger than 2<sup>60</sup>.  Those values are unusable and so they are not permitted in the protocol. 

We don't enforce either restriction, but we could and fairly easily.  The frame parser is easy to tweak to return an error if these values are too large.  Same for the handling of the transport parameter.

ddragana:
Fixed by #608

PR #608, opened 2020-05-07
Enforce maximum values for stream offsets and stream ID limits

abbcdfin:
Enforce maximum values for stream offsets and stream ID limits

As requested in #546, to make the implementation better comply with the protocol.

### DIFF
--- a/neqo-transport/src/frame.rs
+++ b/neqo-transport/src/frame.rs
@@ -594,9 +594,14 @@ impl Frame {
             }),
             FRAME_TYPE_CRYPTO => {
                 let o = dv!(dec);
+                let l = dv!(dec);
+                if (o + l) > (1 << 62 - 1) {
+                    return Err(Error::FrameEncodingError)
+                }
+                let len = Some(l);
                 Ok(Self::Crypto {
                     offset: o,
-                    data: d!(dec.decode_vvec()).to_vec(), // TODO(mt) unnecessary copy
+                    data: d!(dec.decode_checked(len)).to_vec(), // TODO(mt) unnecessary copy
                 })
             }
             FRAME_TYPE_NEW_TOKEN => {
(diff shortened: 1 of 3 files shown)


neqo-transport/src/frame.rs
```diff
@@ -594,9 +594,14 @@ impl Frame {
             }),
             FRAME_TYPE_CRYPTO => {
                 let o = dv!(dec);
+                let l = dv!(dec);
```
martinthomson, line 597: You can still use decode_vvec here.  Just check that `o + vec.len() < ((1<<62)-1)`.  You can do that before you call to_vec() and do the copy.

abbcdfin: Sure, I see. Thanks for letting me know.

abbcdfin: Hi @martinthomson ,

With the method you mentioned, I need to convert from usize to u64 `o + vec.len() as u64 < ((1<<62)-1)`. Though most likely this should be safe for quite a long while, but it is not really future proofing.

Is there any better way to handle this? Please help.

Thanks and Best Regards

martinthomson: The conversion should be ok. Packets don't get that big. As long as we check the conversion, we are ok. So far, when it comes to lengths, our policy is that we will accept up to whatever usize AND u64 can hold. Mostly that is the same thing, but even if that is only 2^32, that's a lot.

Use u64::try_from rather than `as`.

abbcdfin: Got it, thanks a lot.

neqo-transport/src/frame.rs
```diff
@@ -594,9 +594,13 @@ impl Frame {
             }),
             FRAME_TYPE_CRYPTO => {
                 let o = dv!(dec);
+                let data = d!(dec.decode_vvec());
+                if o + u64::try_from(data.len()).unwrap() > ((1 << 62) - 1) {
```
martinthomson, line 598: BTW, I have confirmed that this addition can't overflow as both operands are guaranteed to be < 2^62.

abbcdfin: :)

agrover:
I think this isn't being automerged because CI needs to run, and only runs for collaborators. I've invited @abbcdfin to the project, so hopefully once they accept automation will do the rest.

abbcdfin:
Hi @agrover,
Thanks for the invitation, I have accepted. However, the CI did not start automatically. Might needs an extra kick this time?

agrover:
Who knows what's up with the CI, I just went ahead and merged it. Thanks for your contribution!

abbcdfin:
Hi @agrover,
Thanks for the merging, it's my pleasure.

PR #610, opened 2020-05-08
Ignore timers while closing

martinthomson:
So I managed to remove the 0-length timers thing, but it took some doing.  There were three sources of 0-duration timers.

1. The PTO bug that I found in #609. 

2. We weren't servicing some timers when we were closing (what this PR aims to fix).

3. We kept old loss recovery states around after discarding keys and that was being tracked when reporting the loss recovery timer.  That is fixed in #586.  That fix is needed for the test that I'm adding here, so this turns out to be a bit of a ball of string.

I would like to land #586 first, then we can rebase this one.  The PTO bug is fairly self-contained and could land as well.

If we do that, the changelog on this will look a lot less messy.  The diff here isn't terrible, but it does include #586, which makes it look pretty bad.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -562,10 +562,10 @@ mod old {
                             .expect("Read should succeed");
 
                         let mut have_out_file = false;
-                        if let Some(out_file) = out_file {
+                        if let Some(Some(out_file)) = out_file {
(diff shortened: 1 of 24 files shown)


agrover:
merged #586 so it should be safe to rebase and merge this :+1:

PR #611, opened 2020-05-11
Add an encoder's max_table_size.

ddragana:
Currently we were using the any size a peer was returning in the SETTINGS_QPACK_MAX_TABLE_CAPACITY.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -12,6 +12,7 @@ use qlog::QlogStreamer;
 use neqo_common::{self as common, hex, matches, qlog::NeqoQlog, Datagram, Role};
 use neqo_crypto::{init, AuthenticationStatus};
 use neqo_http3::{self, Header, Http3Client, Http3ClientEvent, Http3State, Output};
+use neqo_qpack::QpackSettings;
 use neqo_transport::FixedConnectionIdManager;
 
 use std::cell::RefCell;
@@ -75,8 +76,11 @@ pub struct Args {
     #[structopt(short = "h", long, number_of_values = 2)]
     header: Vec<String>,
 
-    #[structopt(name = "max-table-size", short = "t", long, default_value = "128")]
-    max_table_size: u64,
+    #[structopt(name = "encoder-table-size", short = "e", long, default_value = "128")]
+    max_table_size_encoder: u64,
+
+    #[structopt(name = "decoder-table-size", short = "d", long, default_value = "128")]
+    max_table_size_decoder: u64,
 
     #[structopt(name = "max-blocked-streams", short = "b", long, default_value = "128")]
     max_blocked_streams: u16,
@@ -341,8 +345,11 @@ fn client(
         Rc::new(RefCell::new(FixedConnectionIdManager::new(0))),
         local_addr,
         remote_addr,
-        args.max_table_size,
-        args.max_blocked_streams,
+        QpackSettings {
+            max_table_size_encoder: args.max_table_size_encoder,
+            max_table_size_decoder: args.max_table_size_decoder,
+            max_blocked_streams: args.max_blocked_streams,
+        },
     )
     .expect("must succeed");
     client.set_qlog(qlog_new(args, origin)?);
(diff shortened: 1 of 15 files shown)


neqo-client/src/main.rs
```diff
@@ -75,8 +75,21 @@ pub struct Args {
     #[structopt(short = "h", long, number_of_values = 2)]
     header: Vec<String>,
 
-    #[structopt(name = "max-table-size", short = "t", long, default_value = "128")]
-    max_table_size: u64,
+    #[structopt(
+        name = "max-table-size_encoder",
```
martinthomson, line 79: I think that the mix of hyphens and underscores will be annoying.  I know that the German keyboard layout finds both equally annoying, but the typical US keyboard doesn't have a modifier key for hyphens.  I recommend just using `encoder-table-size`.

neqo-http3-server/src/main.rs
```diff
@@ -35,8 +35,11 @@ struct Args {
     #[structopt(default_value = "[::]:4433")]
     hosts: Vec<String>,
 
-    #[structopt(short = "t", long, default_value = "128")]
-    max_table_size: u64,
+    #[structopt(short = "e", long, default_value = "128")]
```
martinthomson, line 38: Please use consistent forms for this between server and client.

neqo-http3/src/connection.rs
```diff
@@ -84,21 +78,23 @@ impl<T: Http3Transaction> ::std::fmt::Display for Http3Connection<T> {
 
 impl<T: Http3Transaction> Http3Connection<T> {
     /// Create a new connection.
-    pub fn new(max_table_size: u64, max_blocked_streams: u16) -> Self {
-        if max_table_size > (1 << 30) - 1 {
+    pub fn new(local_qpack_settings: QpackSettings) -> Self {
+        if (local_qpack_settings.max_table_size_encoder > (1 << 30) - 1)
```
martinthomson, line 82:
```suggestion
        if local_qpack_settings.max_table_size_encoder >= (1 << 30)
```

martinthomson: A constant for that limit might also be good.

neqo-http3/src/lib.rs
```diff
@@ -39,6 +39,28 @@ pub use transaction_server::TransactionServer;
 
 type Res<T> = Result<T, Error>;
 
+#[derive(Debug, PartialEq, PartialOrd, Ord, Eq, Clone, Copy)]
+pub struct QpackSettings {
+    max_table_size_decoder: u64,
+    max_table_size_encoder: u64,
+    max_blocked_streams: u16,
```
martinthomson, line 46: This is one cases where it might be OK to have public fields.  It makes instantiating settings easier.  If you do that, maybe provide default values (`#[derive(Default)]` isn't entirely crazy here, even if that isn't necessarily optimal) so that people can do `QpackSettings { max_blocked_streams: 100, ...QpackSettings::default() }` or similar.  The advantage of public fields is that you get documentation.  The above is better than `QpackSettings::new(0, 0, 100)`, because you don't have to look up which argument goes where.

neqo-qpack/src/encoder.rs
```diff
@@ -41,10 +42,11 @@ pub struct QPackEncoder {
 
 impl QPackEncoder {
     #[must_use]
-    pub fn new(use_huffman: bool) -> Self {
+    pub fn new(max_table_size: u64, use_huffman: bool) -> Self {
```
martinthomson, line 45: You could make this take an argument of `QpackSettings`, I think.  That there are values this doesn't use isn't that big of a deal.

neqo-qpack/src/encoder.rs
```diff
@@ -66,13 +68,15 @@ impl QPackEncoder {
...
             cap,
-            self.table.capacity()
+            self.table.capacity(),
+            self.max_table_size,
         );
-        self.max_entries = (cap.to_f64().unwrap() / 32.0).floor().to_u64().unwrap();
-        // we also set our table to the max allowed. TODO we may not want to use max allowed.
-        self.change_capacity(cap)
+        let new_cap = std::cmp::min(self.max_table_size, cap);
+        self.max_entries = (new_cap.to_f64().unwrap() / 32.0).floor().to_u64().unwrap();
```
martinthomson, line 77: Since you are here, this should be:

```suggestion
        self.max_entries = min(self.max_table_size, cap) / 32;
```

Simple division truncates in the way we want and we don't need to round-trip to a float.

PR #613, opened 2020-05-11
Allow passing of String to close()

martinthomson:
This turns the close() function into a template so that we can accept
owned strings.  This doesn't change usage at all.

From #558.


martinthomson:
I changed my mind here.  It is better to pass in a reference as we need to allocate a `Vec<u8>` from the value we get.  It's nice to have a string API, but we ultimately want to own a reference to a sequence of octets.

(Some of this gets backed out with #558, but this should make the usage nicer until then.)

PR #616, opened 2020-05-11
The qpack encoder uses stream_send_atomic to send instructions

ddragana:


neqo-transport/src/send_stream.rs
```diff
@@ -1232,4 +1257,35 @@ mod tests {
...
+
+        let mut s = SendStream::new(4.into(), 0, Rc::clone(&flow_mgr), conn_events);
+        s.set_max_stream_data(2);
+
+        // Stream is initially blocked (conn:5, stream:2)
+        // and will not accept atomic write of 3 bytes.
+        assert_eq!(s.send_atomic(b"abc").unwrap(), 0);
+
+        // assert non-atomic write works
+        assert_eq!(s.send(b"abc").unwrap(), 2);
```
martinthomson, line 1275: This should still send a blocked frame.  This test doesn't appear to test that these arrive at the right moments.  Can you do that?

neqo-transport/src/send_stream.rs
```diff
@@ -629,6 +649,11 @@ impl SendStream {
             return Ok(0);
         }
 
+        if atomic && (can_send_bytes < buf.len() as u64) {
```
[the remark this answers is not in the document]
ddragana: I will make it more compact, but I will open another issue for making avail() return usize. That will not be some lines of code.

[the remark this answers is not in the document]
martinthomson: Yeah, getting the right types for things is hard.  We should agree on what the principles are.

neqo-qpack/src/encoder.rs
```diff
@@ -211,50 +211,49 @@ impl QPackEncoder {
...
+        qdebug!([self], "insert {:?} {:?}.", name, value);
         self.send(conn)?;
         if self.send_buf.len() != 0 {
             return Err(Error::EncoderStreamBlocked);
         }
 
+        let entry_size =
+            u64::try_from(name.len() + value.len() + ADDITIONAL_TABLE_ENTRY_SIZE).unwrap();
+
+        if !self.table.check_insert_possible(entry_size) {
```
martinthomson, line 229: This is another case where usize is probably a better type for the argument to check_insert_possible (I would call it just `insert_possible`).

neqo-qpack/src/table.rs
```diff
@@ -32,7 +34,7 @@ impl DynamicTableEntry {
     }
 
     pub fn size(&self) -> u64 {
```
martinthomson, line 36: sizes should be usize :)

neqo-transport/src/connection.rs
```diff
@@ -3380,10 +3385,14 @@ mod tests {
...
-            assert!(matches!(
-                frames[0],
-                (Frame::Stream { .. }, PNSpace::ApplicationData)
-            ));
+            assert_eq!(
+                frames
+                    .iter()
+                    .filter(|i| matches!(i, (Frame::Stream { .. }, PNSpace::ApplicationData)))
+                    .collect::<Vec<_>>()
+                    .len(),
```
martinthomson, line 3393: You can use .count() instead of .collect().len().

PR #617, opened 2020-05-11
Add a reference to a qpack dynamic table entry as soon as it is referred in a header block encoding

ddragana:
If header block consists of 2 headers and the first is encoded as a reference to an entry in the dynamic table and we do not add the reference immediately the following header may delete the entry the first header is referring to


ddragana:
The error at the end of the test will be fix by #614

PR #618, opened 2020-05-12
Pacing

### LINKED ISSUE #185: Implement Pacing
As mentioned [here](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#rfc.section.7.8) we should implement pacing, to spread packets we're allowed to send over the smoothed RTT. This is a nice-to-have feature if we're using NewReno congestion control, but my understanding is that BBR requires pacing, so this blocks if Neqo wanted to support BBR or BBR2 in the future.

martinthomson:
This adds a pacer.

There are a few related changes that I want to spend a bit of time trying to extricate before I call this "ready".

1. Because the pacer needs to be initialized, it turns out that leaving the server in Init state (rather than WaitInitial) is better.  That simplified some of the setup code, which is good.

2. Adding more stuff to `output_path` pushed it completely over the edge complexity-wise.  So I created a struct called `SendProfile` to hold a bunch of instructions from the recovery/cc code to the packet sending code.  That turned out to be more complex at first, but a little bit of cleaning allowed this to be cleaner overall.

3. As the pacer needs to communicate with the code that sets the timer (if we decide not to send a packet because we are pacing, we need to include the pacing timer in the calculation for when the next timer pops), the interface from `output` needs to be wider than just `Option<Datagram>`.  A new `SendOption` enum covers that off.

4. Part of the factoring resulting from use of `SendProfile` was that we send a PING in every active packet number space when there is a PTO.  Some implementations will interpret a PING as a cue to send an immediate ACK (we probably should too), so this might be a better way to elicit an ACK than skipping a packet number.  I mostly did this for the huge simplification of the code, but this is nice.

5. I added RTTs to a bunch of the core test methods we use.

6. The test function for filling the congestion window is now a lot more complicated.  Unfortunately, our API doesn't really reveal what timer was involved when we set `Output::Callback`, so we have to check internals to ensure that we keep advancing time when asked to pace, but not things like the PTO or loss recovery timer.

None of these are really that easy to separate from the core code.  The only one that isn't directly a result of adding pacing is the `SendProfile` factoring, which could be extracted into a separate patch.  It's still pretty closely bound to the pacer though.  If this is too daunting a review task, then let me know.

Closes #185.


neqo-transport/src/connection.rs
```diff
@@ -155,6 +153,20 @@ impl Output {
...
 
+/// Used by inner functions like Connection::output.
+enum SendOption {
+    /// Yes, please send this datagram.
+    Yes(Datagram),
+    /// Don't send.  If this was blocked on the pacer (the arg is true).
+    No(bool),
+}
+
+impl Default for SendOption {
```
martinthomson, line 164: Would this be better as a named function instead, like `no()`?

neqo-transport/src/cc.rs
```diff
@@ -190,4 +204,27 @@ impl CongestionControl {
         //TODO(agrover): how do we get this info??
         false
     }
+
+    pub fn start_pacer(&mut self, now: Instant) {
+        // Start the pacer with a small burst size of 2 packets.
```
martinthomson, line 209:
```suggestion
        // Start the pacer with a small burst size.
```

neqo-transport/src/connection.rs
```diff
@@ -1261,38 +1268,22 @@ impl Connection {
...
-                }
-            } else {
-                // A higher packet number space has a PTO; only add a PING.
-                builder.encode_varint(Frame::Ping.get_type());
-                (Vec::new(), true)
-            };
+            let limit = profile.limit() - tx.expansion();
+            let (tokens, ack_eliciting) =
+                self.add_frames(&mut builder, *space, limit, &profile, now);
+            if builder.is_empty() {
```
martinthomson, line 1282: It occurs to me that this is not a good check in the first place, because the builder will always include a header.  So I think we need to fix this test.

```suggestion
            if builder.len() == payload_start {
```

neqo-transport/src/connection.rs
```diff
@@ -3763,32 +3770,51 @@ mod tests {
         assert_eq!(pkt0.as_dgram_ref().unwrap().len(), PATH_MTU_V6);
     }
 
-    // Handle sending a bunch of bytes from one connection to another, until
-    // something stops us from sending.
-    fn send_bytes(src: &mut Connection, stream: u64, now: Instant) -> Vec<Datagram> {
+    /// This fills the congestion window from a single source.
+    /// As the pacer will interfere with this, this moves time forward
+    /// as `Output::Callback` is received.  Because it is hard to tell
+    /// from the return value whether a timeout is to ACK delay, PTO, or
```
martinthomson, line 3776:
```suggestion
    /// from the return value whether a timeout is an ACK delay, PTO, or
```

neqo-transport/src/connection.rs
```diff
@@ -3763,32 +3770,51 @@ mod tests {
...
+                src.loss_recovery.cwnd_avail(),
+                pkt
+            );
             match pkt {
                 Output::Datagram(dgram) => {
                     total_dgrams.push(dgram);
                 }
-                Output::Callback(_) => break,
+                Output::Callback(t) => {
+                    if src.loss_recovery.cwnd_avail() < ACK_ONLY_SIZE_LIMIT {
```
martinthomson, line 3808: This is where we need special info, which includes when we stop sending packets.  This is unfortunate, but I don't see how this can be avoided without creating a new API.  Maybe that API is something we want, but I didn't want to create one without first discussing it and having a use case for it that isn't just making our tests work.

neqo-transport/src/pace.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Spend credit.  This cannot fail; users of this API are expected to call
+    /// next() to determine when to spend.  This takes the current time (`now`),
+    /// an estimate of the round trip time (`rtt`), the estimated congestion
+    /// window (`cwnd`), and the number of bytes that were sent (`count`).
+    pub fn spend(&mut self, now: Instant, rtt: Duration, cwnd: usize, count: usize) {
+        qtrace!([self], "spend {} over {}, {:?}", count, cwnd, rtt);
+        // Increase the capacity by:
+        //    `(now - self.t) * PACER_SPEEDUP * cwnd / rtt`
+        // That is, the elapsed fraction of the RTT times rate that data is added.
+        let incr = (now - self.t)
```
martinthomson, line 93:
```suggestion
        let incr = now.saturating_duration_since(self.t)
```

Otherwise, if `now` is in the past relative to `self.t`, bad things happen.

Getting the overflows and underflows right on this is nasty.

neqo-transport/src/pace.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    #[test]
+    fn even() {
+        let mut n = now();
+        let p = Pacer::new(n, PACKET, PACKET);
+        assert_eq!(p.next(RTT, CWND), None);
+        p.spend(n, RTT, CWND, PACKET);
+        assert_eq!(p.next(RTT, CWND), Some(n + (RTT / 10)));
+    }
+}
```
martinthomson, line 136:
```suggestion

    #[test]
    fn backwards_in_time() {
        let mut n = now();
        let p = Pacer::new(n + RTT, PACKET, PACKET);
        assert_eq!(p.next(RTT, CWND), None);
        // Now spend some credit in the past using a time machine.
        p.spend(n, RTT, CWND, PACKET);
        assert_eq!(p.next(RTT, CWND), Some(n + (RTT / 10)));
    } 
}
```

neqo-transport/src/recovery.rs
```diff
@@ -97,6 +100,65 @@ impl RttVals {
...
+    limit: usize,
+    pto: Option<PNSpace>,
+    paced: bool,
+}
+
+impl SendProfile {
+    pub fn new_limited(limit: usize) -> Self {
+        // When the limit is too low, we only send ACK frames.
+        // Set the limit to `ACK_ONLY_SIZE_LIMIT - 1` to ensure that
+        // ACK-only packets are still limited in size.
```
martinthomson, line 115: Yes, I know that our ACK sender doesn't pay attention to the space it has available, but it probably should.

martinthomson:
Hey @agrover, this is starting to rot now, and I'd like to understand what you want to do with it.

agrover:
It's a rather large PR and pacing is a medium priority issue, and an enhancement, not a bug. (Well, it's arguable.)

I'm out next week but will get to it the following week. I know it's gathering conflicts as time passes but I'm happy to do the needed conflict resolution before I review it.

neqo-transport/src/pace.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// credit needed before a packet is sent.  This should be a substantial
+    /// fraction of the maximum packet size, if not the packet size.
+    pub fn new(now: Instant, m: usize, p: usize) -> Self {
+        assert!(m >= p, "maximum capacity has to be at least one packet");
+        Self { t: now, m, c: m, p }
+    }
+
+    /// Determine when the next packet will be available based on the provided RTT
+    /// and congestion window.  This doesn't update state.
+    /// This returns an option, with a value of `None` returned when there is
```
ddragana, line 57: "option"? this does not match the function.

neqo-transport/src/connection.rs
ddragana: I think it is fine as it is.

martinthomson:
"Rebased" and this should be OK for another round.  Nothing rotted in the meantime \o/.

PR #619, opened 2020-05-12
neqo-client: Handle UdpSocket::send() errors

vonasek:


vonasek:
> Are we going to do the same for the server?

I created an issue [#628](https://github.com/mozilla/neqo/issues/628)

PR #623, opened 2020-05-13
Ignore kept lost packets

### LINKED ISSUE #621: assertion failed: earliest > now
[neqo-client.log](https://github.com/mozilla/neqo/files/4616805/neqo-client.log)

I hit this almost all the time when running handshakeloss qns test. This log is from [pacing branch](https://github.com/martinthomson/neqo/tree/pacing) because in PR [#618](https://github.com/mozilla/neqo/pull/618) was mentioned that it should fix issue [#177](https://github.com/mozilla/neqo/issues/177), but it didn't. Latest master branch fails in the same way.

martinthomson:
This should fix the problems that we are seeing in interop testing.

Getting this test right was quite difficult, but I think that this is robust.

Fixes #621.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -4695,8 +4695,10 @@ mod tests {
 
     /// We should not be setting the loss recovery timer based on packets
     /// that are sent prior to the largest acknowledged.
-    /// This requires that there be a gap in one packet number space so that
-    /// the packets sent in another space can be counted.
+    /// Testing this requires that we construct a case where one packet
+    /// number space causes the loss recovery timer to be engaged.  At the same time,
+    /// there is a packet in another space that hasn't been acknowledged AND
+    /// that packet number space has not received acknowledgments for later packets.
     #[test]
     fn loss_time_past_largest_acked() {
         const RTT: Duration = Duration::from_secs(10);
@@ -4773,8 +4775,6 @@ mod tests {
         // mark packets as lost and retransmit, after which we should be on the PTO
         // timer.
         now += lr_time;
-        let retrans = client.process(None, now).dgram();
-        assert!(retrans.is_some());
         let delay = client.process(None, now).callback();
         assert_ne!(delay, Duration::from_secs(0));
         assert!(delay > lr_time);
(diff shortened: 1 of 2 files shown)


neqo-transport/src/recovery.rs
```diff
@@ -126,18 +126,23 @@ impl LossRecoverySpace {
...
-            self.sent_packets
-                .values()
-                .min_by_key(|sp| sp.time_sent)
-                .map(|sp| sp.time_sent)
-        );
-        earliest
+        // Sent packets should be ordered by packet number, so pick the first that isn't lost.
+        for sent in self.sent_packets.values() {
+            if !sent.lost() {
+                debug_assert_eq!(
```
martinthomson, line 135: @grover, do you think we need to keep this assertion?  I fixed it, but I'm content in relying on the iteration order of the BTreeMap for this.

neqo-transport/src/connection.rs
```diff
@@ -4611,4 +4611,172 @@ mod tests {
...
+    /// there is a packet in another space that hasn't been acknowledged AND
+    /// that packet number space has not received acknowledgments for later packets.
+    #[test]
+    fn loss_time_past_largest_acked() {
+        const RTT: Duration = Duration::from_secs(10);
+        const INCR: Duration = Duration::from_millis(1);
+        let mut client = default_client();
+        let mut server = default_server();
+
+        // Maintain a timer for each endpoint.
```
martinthomson, line 4709:
```suggestion
```

neqo-transport/src/recovery.rs
```diff
@@ -126,18 +129,12 @@ impl LossRecoverySpace {
         self.space
     }
 
+    /// Find the time we sent the first packet that is lower than the
+    /// largest acknowledged and that isn't yet declared lost.
+    /// The logic here needs to match the logic in `detect_lost_packets`.
```
martinthomson, line 134:
```suggestion
    /// Use the value we prepared earlier in `detect_lost_packets`.
```

agrover:
@martinthomson :+1: from me. Looks like you're still tidying around the edges but please merge when you're happy.

neqo-transport/src/recovery.rs
agrover: I think I was worried we might switch to a different Map impl but that really would break a lot of other stuff, so this isn't really needed and can go.

PR #625, opened 2020-05-13
Sanitize application close frames

martinthomson:
I realize that I was pushing to my fork, but the original PR was from the branch on the main repo that I created before getting the fork setup.

This patch sanitizes CONNECTION_CLOSE as required by the spec in case the close comes from the application.  We don't want to leak application details, but we still need to send a CONNECTION_CLOSE in some cases.  This does that.

See the commentary on #558 for more.


neqo-transport/src/connection.rs
```diff
@@ -292,17 +298,36 @@ impl StateSignaling {
...
     }
 
-    pub fn close(&mut self) {
-        *self = Self::ConnectionClose
+    /// If a close is pending, take it.
+    pub fn close_frame(&mut self) -> Option<Frame> {
+        if let Self::ConnectionClose(frame) = self {
+            let frame = mem::replace(frame, Frame::Padding);
+            *self = Self::CloseSent(frame.clone());
+            Some(frame)
```
agrover, line 319: Maybe try defining ConnectionClose as ConnectionClose(Option<Frame>) then use `Option::take()`? Less... weird. Same for `CloseSent` below.

That might also let you define data members in ConnectionClose as the data in a close frame rather than an actual `Frame`, then build the Close frame here? So there would be no possibility of another frame type slipping in there somehow. Dunno, just an idea.

neqo-transport/src/connection.rs
```diff
@@ -1383,13 +1397,14 @@ impl Connection {
...
-    pub fn close(&mut self, now: Instant, error: AppError, msg: &str) {
-        self.set_state(State::Closing {
-            error: ConnectionError::Application(error),
-            frame_type: 0,
-            msg: msg.into(),
-            timeout: self.get_closing_period_time(now),
-        });
+    pub fn close<S>(&mut self, now: Instant, app_error: AppError, msg: S)
+    where
+        S: AsRef<str>,
```
agrover, line 1402: you like this better than
```
    pub fn close(
        &mut self,
        error: ConnectionError,
        frame_type: FrameType,
        message: impl AsRef<str>,
```
I like `impl Trait` in argument position but I know opinions differ.

martinthomson: I don't know why I keep forgetting that.  I also prefer that.

neqo-transport/src/connection.rs
martinthomson: This one is not really something that `Option::take()` can support.  Because we aren't really taking the value.  We need to keep the value for later, in case we need to send another.  That said, this isn't 100% right now that I've thought this through again.  (This hasn't captured the difference between closing and draining properly.

neqo-transport/src/connection.rs
```diff
@@ -909,10 +979,10 @@ impl Connection {
                 State::Closing { .. } => {
                     // Don't bother processing the packet. Instead ask to get a
                     // new close frame.
-                    self.state_signaling.close();
+                    self.state_signaling.send_close();
                     return Ok(frames);
                 }
-                State::Closed(..) => {
+                State::Draining { .. } | State::Closed(..) => {
```
martinthomson, line 985: Note: in most cases Draining and Closing are lumped together.  Not here.

neqo-transport/src/lib.rs
```diff
@@ -96,7 +100,10 @@ pub enum Error {
 impl Error {
     pub fn code(&self) -> TransportError {
         match self {
-            Self::NoError | Self::IdleTimeout => 0,
+            Self::NoError
+            | Self::IdleTimeout
+            | Self::PeerError(_)
+            | Self::PeerApplicationError(_) => 0,
```
martinthomson, line 106: Our response to an error code from the other side should be NO_ERROR.

martinthomson:
@grover, After noticing that problem, I decided that we should probably implement the draining thing properly.  This does that.

So I'm asking for another review.

CI failures are caused by a very rare race condition that I have a fix on a parallel PR for.

martinthomson:
Rebased, and it's still working.

agrover:
:+1: once conflicts resolved.

PR #632, opened 2020-05-14
Add [self] to padding log

agrover:
We should probably do this more often.


agrover:
This PR is based on a trivial commit pushed to mozilla/neqo by martinthomson. I'm in the process of cleaning up stale branches and am just pushing this in.

Rebased and reformatted to pass CI.

PR #634, opened 2020-05-15
Updated rust stable and now it formats macros

martinthomson:
I want to create a PR for this so that I can be sure that the old rust
won't object to the change.


agrover:
okay then

PR #642, opened 2020-05-15
Fix option conflict and change default values

### LINKED ISSUE #638: -d option to neqo-http3-server is conflicted
In #611, we added a -d option, but that is used by the server code already to identify where the NSS database is.  We should pick a different option.  Do we need a short code for this option?

ddragana:
Fixes #638

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -76,13 +76,13 @@ pub struct Args {
     #[structopt(short = "h", long, number_of_values = 2)]
     header: Vec<String>,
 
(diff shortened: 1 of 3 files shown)


agrover:
alternate proposal: don't give these options short options.

agrover:
For the same values? They can go too.

martinthomson:
I don't care so much about the defaults, but it's probably good to have consistent defaults.

I do care that the arguments use the same short forms.

martinthomson:
FWIW, I'm with Andy on dropping the short tags, but this is fine.

PR #645, opened 2020-05-17
Add new macro that logs in all circumstances

valenting:
I've been looking into bug [1611990](https://bugzilla.mozilla.org/show_bug.cgi?id=1611990) but there doesn't seem to be an easy way to make all logging happen regardless if `release_max_level_info` is set.

Instead, it seems we can replace the [log!](https://github.com/rust-lang/log/blob/4fb398b68ea43f5317e6e0f04b5eacaa9e679218/src/macros.rs#L34) logging macro with do_log! that does pretty much the same but does not respect `STATIC_MAX_LEVEL`

### DIFF
--- a/neqo-common/src/log.rs
+++ b/neqo-common/src/log.rs
@@ -7,7 +7,6 @@
 use std::io::Write;
 use std::sync::Once;
 use std::time::Instant;
-extern crate log;
 
 #[macro_export]
 macro_rules! do_log {
@@ -24,12 +23,11 @@ macro_rules! do_log {
     ($lvl:expr, $($arg:tt)+) => ($crate::do_log!(target: ::log::__log_module_path!(), $lvl, $($arg)+))
 }
 
-#[macro_export(local_inner_macros)]
+#[macro_export]
 macro_rules! log_enabled {
     (target: $target:expr, $lvl:expr) => {{
         let lvl = $lvl;
-        lvl <= ::log::max_level()
-            && ::log::__private_api_enabled(lvl, $target)
+        lvl <= ::log::max_level() && ::log::__private_api_enabled(lvl, $target)
     }};
     ($lvl:expr) => {
         $crate::log_enabled!(target: ::log::__log_module_path!(), $lvl)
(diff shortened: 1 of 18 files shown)


neqo-common/src/log.rs
```diff
@@ -7,6 +7,34 @@
 use std::io::Write;
 use std::sync::Once;
 use std::time::Instant;
+extern crate log;
+
+#[macro_export]
+macro_rules! do_log {
+    (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
```
martinthomson, line 14: Please use log_enabled here .

neqo-common/src/log.rs
```diff
@@ -7,6 +7,34 @@
 use std::io::Write;
 use std::sync::Once;
 use std::time::Instant;
+extern crate log;
```
martinthomson, line 10: Don't we do `use ::log` instead?

neqo-common/src/log.rs
valenting: That's not what the log! macro does.
And that would actually make the performance of the logging twice as slow.

Right now when we log, it just checks the max_level, then goes into the logger implementation and [logs](https://searchfox.org/mozilla-central/rev/09b8072a543c145de2dc9bb76eddddd4a6c09adc/toolkit/library/rust/shared/lib.rs#160,191)

If we were to call log_enabled, then we'd also have to make sure [enabled](https://searchfox.org/mozilla-central/rev/09b8072a543c145de2dc9bb76eddddd4a6c09adc/toolkit/library/rust/shared/lib.rs#185) returns the proper runtime level, which means acquiring the [lock](https://searchfox.org/mozilla-central/rev/09b8072a543c145de2dc9bb76eddddd4a6c09adc/xpcom/rust/gecko_logger/src/lib.rs#49) on the hashmap. Except we'd do it twice, once for log_enabled and a second time for actually logging.

The only downside of doing it this way is that we're needlessly expanding the format arguments.

martinthomson: Ahh, I see now.  That's awkward, but I guess that's where we are at with this.

I wonder whether we should internalize the log_enabled!() macro and replace it with a log_subject!() macro that takes T:Display and produces either format!("{}", x) or "" as needed to satisfy the only use of log_enabled!() we have in our code.

valenting:
Thanks for the review!

PR #649, opened 2020-05-19
Reduce strictness of assertions while closing

martinthomson:
This patches over the assertion, essentially allowing us to mark something as closing.  I don't know if this is the best strategy, but it does stop a crash that is fairly reproducible from the server end.

The alternative is to try to find why things are attempting to set a closing state after going to closed.  If you think that is better, then I'm OK with deferring this and doing the analysis.  I'm just recording what I did to remove a road block.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -11,6 +11,7 @@ use std::cmp::{max, min, Ordering};
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt::{self, Debug};
+use std::mem;
 use std::net::SocketAddr;
 use std::rc::Rc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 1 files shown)


ddragana:
Is this happening with http3 server or quic server?

I actually want to understand why it is happening.

martinthomson:
I'm going to try to dig in and understand it first.  I'm seeing this with neqo-http3-server.

ddragana:
> I'm going to try to dig in and understand it first. I'm seeing this with neqo-http3-server.

I am not really happy with http3 server side. I will refactor it at some point (this is not the highest priority at the moment) so if you want you can land this and open an issue to fix this later.

neqo-transport/src/connection.rs
```diff
@@ -1817,8 +1818,11 @@ impl Connection {
                 _ => {}
             }
             self.events.connection_state_change(state);
-        } else {
-            assert_eq!(state, self.state);
+        } else if mem::discriminant(state) != mem::discriminant(self.state) {
```
martinthomson, line 1821:
```suggestion
        } else if mem::discriminant(&state) != mem::discriminant(&self.state) {
```

martinthomson:
I worked out what case was causing this.  The connection was busted, but then the local application decided to also call close(), which then asserted.

PR #652, opened 2020-05-19
Have the server send a session ticket

martinthomson:
This patch adds a few changes to resumption on the HTTP/3 server side.

The simple part is that the HTTP/3 server sends a session ticket (just one) when the connection is established.  This enables resumption (and 0-RTT) when possible.  That change required some changes to the code, but nothing dramatic.

(There is a performance cost here in sending a session ticket at the time when you are most likely needing to send responses.  If we wanted a truly performant server, we might consider a few optimizations for that: sending tickets in 0.5-RTT, which would require changes to NSS because that depends on predicting the client Finished; or, managing the priority of session tickets relative to responses.  Then there is the newly minted ticket request extension, which gives the client the ability to specify how many tickets it needs, which is ideally implemented in NSS, even if we could in theory implement it in this stack.  Those fixes can wait though; we're not looking to provide the very best performing server.)

The bigger change is that the HTTP/3 server saves its settings in the ticket and it checks those settings when the connection is attempted.  If they don't match up, then it rejects 0-RTT.  Plumbing those checks through required some changes to the structure of the checking that the transport does.

An important change to the API here is that 0-RTT is not automatically enabled for servers.  You no longer include everything necessary for 0-RTT in the server connection constructor.  0-RTT has to be separately enabled for single connections.  The server classes (those that collect multiple connections) do include this logic, so the preferred interface for the server end doesn't really change; it's mostly just something that affects our internal test configurations.  The main reason for splitting this off is that servers need anti-replay and 0-RTT configuration checks, and carrying extra arguments for those on the main API was getting unwieldy.


martinthomson:
I hit a regression in NSS with this, but that is the only reason that the tests are failing.  I'm working with Kevin on a fix for NSS, so we won't be able to land this until then.  Until then, we can use an older copy of NSS.

neqo-http3/src/settings.rs
```diff
@@ -113,3 +119,58 @@ impl Deref for HSettings {
...
+            .encode_varint(settings.max_blocked_streams);
+        enc.into()
+    }
+}
+
+impl ZeroRttChecker for HttpZeroRttChecker {
+    fn check(&self, token: &[u8]) -> ZeroRttCheckResult {
+        let mut dec = Decoder::from(token);
+
+        // Read and checm the version.
```
ddragana: typo

neqo-http3/src/server.rs
```diff
@@ -938,11 +966,91 @@ mod tests {
...
+                max_table_size_decoder: DEFAULT_SETTINGS.max_table_size_decoder + 1,
+                ..DEFAULT_SETTINGS
+            },
+            &ZeroRttState::AcceptedClient,
+        );
+    }
+
+    /// A smaller QPACK decoder table size prevents 0-RTT.
+    #[test]
+    fn zero_rtt_smaller_decoder_table() {
```
martinthomson, line 1011: @ddragana, not sure what you were asking for.  All the tests like this with Rejected are the ones where the 0-RTT check fails.

ddragana: sorry, my mistake.

PR #653, opened 2020-05-19
Preconnect for resumption

### LINKED ISSUE #429: 0-RTT in neqo-client

martinthomson:
Closes #429.

Depends on #651 landing first, so this won't build yet.


martinthomson:
@vonasek has a different approach.  This doesn't assume that you have >1 URL, but if you can, then the alternative approach is probably better.

vonasek:
> @vonasek has a different approach. This doesn't assume that you have >1 URL, but if you can, then the alternative approach is probably better.

I just did exactly what the test description says (see [resumption and zerortt](https://github.com/marten-seemann/quic-interop-runner#test-cases)).

PR #657, opened 2020-05-19
Use separate  send_message/recv_message instead of transactions

ddragana:
- make server and client use the same code, i.e. send_message/recv_message
- http3 connection use separate send_message/recv_message and not transaction. This will help with Push, which only has one side
- add a place holder for a PushController


ddragana:
I have split this PR into separate commits for a easier review

neqo-http3-server/src/main.rs
```diff
@@ -92,14 +92,17 @@ fn process_events(server: &mut Http3Server) {
 
                 let default_ret = b"Hello World".to_vec();
 
-                let response = match headers.iter().find(|&(k, _)| k == ":path") {
-                    Some((_, path)) if !path.is_empty() => {
-                        match path.trim_matches(|p| p == '/').parse::<usize>() {
-                            Ok(v) => vec![b'a'; v],
-                            Err(_) => default_ret,
+                let response = match headers {
```
martinthomson, line 95: Under what conditions do we allow headers to be None, as opposed to an empty set.

Also, we should reject anything without ':path' to keep this simple.

ddragana: If stream gets closed before headers we return None. If we get empty header frame we return empty set.
We probably should just ignore streams being closed without a header frame, I have not fix this yet, because there are some other small corner cases that I wanted to do at once (they are all about how much semantics neqo-htttp3 implements and what it lefts for the application)

for neqo-http3-server we should reject anything without a :path.

neqo-http3/src/send_message.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    pub fn send_body(&mut self, conn: &mut Connection, buf: &[u8]) -> Res<usize> {
+        qinfo!(
+            [self],
+            "send_request_body: state={:?} len={}",
+            self.state,
+            buf.len()
+        );
+        match self.state {
```
martinthomson, line 115: I don't think that we need to be as careful with framing at this layer.  If we get blocked, that only prevents us from sending other frames on the stream, which boils down to trailers (which we don't send) and control frames (which don't currently exist).

So I would suggest sending everything we have as possible.  Then you only have to track how much data you have left to send in the current frame (i.e., SendMessageState::SendingData now has a usize parameter).  The only probing you would then need to do is to work out whether the header of the frame fits in the available space, but you can do that with the atomic write.

An application that does have other things to send than data (in the future perhaps) might withhold data from the stack to make space, or we can revise this code and cap the size of writes in some way.

That should make this a lot less complex: Encode a DATA frame header, write it atomically then send whatever data from the provided buffer you can using the non-atomic send.

That also means that an application that attempts to send X bytes and only is able to send X-Y bytes must later send at least Y bytes or bad things will happen because we will otherwise be unable to fill the frame.  It's worth documenting that constraint.

neqo-http3/src/send_message.rs
```diff
@@ -0,0 +1,270 @@
...
+                match conn.stream_send(self.stream_id, &buf[..to_send]) {
+                    Ok(sent) => Ok(sent),
+                    Err(e) => Err(Error::TransportError(e)),
+                }
+            }
+            SendMessageState::Closed => Err(Error::AlreadyClosed),
+        }
+    }
+
+    pub fn is_sending_closed(&self) -> bool {
```
martinthomson, line 166: I think that you could attach this method (and the two following) to the state.

neqo-http3/src/send_message.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    pub fn done(&self) -> bool {
+        matches!(self.state, SendMessageState::Closed)
+    }
+
+    pub fn is_state_sending_data(&self) -> bool {
+        matches!(self.state, SendMessageState::SendingData)
+    }
+
+    fn ensure_response_encoded(
```
martinthomson, line 183: Is this code generic?  That is, should it be response-specific or does it apply to requests also?  It looks like it.  Is this better named `encode_header`?

neqo-http3/src/server.rs
```diff
@@ -111,31 +113,59 @@ impl Http3Server {
...
+                            stream_id,
+                            headers,
+                            fin,
+                        } => self.events.headers(
+                            ClientRequestStream::new(conn.clone(), handler.clone(), stream_id),
+                            headers,
+                            fin,
+                        ),
+                        Http3ServerConnEvent::DataReadable { stream_id } => {
+                            loop {
```
martinthomson, line 130: I realize that Rust might not let you do this, but this function is an epic.  Please consider moving this loop to a new function.

neqo-http3/src/server_connection_events.rs
```diff
@@ -38,6 +36,28 @@ pub(crate) struct Http3ServerConnEvents {
...
+
+    /// Add a new `DataReadable` event
+    fn data_readable(&self, stream_id: u64) {
+        self.insert(Http3ServerConnEvent::DataReadable { stream_id });
+    }
+}
+
+impl SendMessageEvents for Http3ServerConnEvents {
+    fn data_writable(&self, _stream_id: u64) {
+        // Curently not used on the server side.
```
martinthomson, line 57: Do you have an issue open for this?

neqo-http3/src/send_message.rs
```diff
@@ -0,0 +1,270 @@
...
+                }
+            } else {
+                let b = buf.split_off(sent);
+                *buf = b;
+            }
+        }
+        Ok(())
+    }
+
+    // TransactionClient owns headers and sends them. This method returns if
```
martinthomson, line 244: Check the title.

neqo-http3/src/send_message.rs
```diff
@@ -0,0 +1,270 @@
...
+    // they're still being sent. Request body (if any) is sent by http client
+    // afterwards using `send_request_body` after receiving DataWritable event.
```
martinthomson, line 246: I didn't see it anywhere, but we should intercept the DataWritable event while this method returns true.

neqo-http3/src/connection_client.rs
```diff
@@ -515,20 +536,39 @@ impl Http3Client {
...
+                    Some(*id)
+                } else {
+                    None
+                }
+            }))
+            .collect();
+
+        ids.dedup();
+        ids.iter()
+            .for_each(|id| self.events.reset(*id, Error::HttpRequestRejected.code()));
```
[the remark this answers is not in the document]
ddragana: Actually reset is already idempotent.

neqo-http3/src/recv_message.rs
```diff
@@ -230,24 +243,34 @@ impl ResponseStream {
...
+                                    .as_ref()
+                                    .ok_or(Error::HttpId)?
+                                    .borrow()
+                                    .new_push_promise(push_id, header_block)?,
+                                HFrame::DuplicatePush { push_id } => self
+                                    .push_handler
+                                    .as_ref()
+                                    .ok_or(Error::HttpId)?
+                                    .borrow()
+                                    .new_duplicate_push(push_id)?,
```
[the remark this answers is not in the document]
ddragana: I was just transforming code without changing it functionality. I will add PR #665 to this PR

neqo-http3/src/send_message.rs
ddragana: It is a good point. I will open a separate issue for this.

neqo-http3/src/server_connection_events.rs
ddragana: No, but the server side needs a lots of improvements and this is the first think that needs to change. When I am happy with the client side, i will look into the server side.

neqo-http3/src/send_message.rs
ddragana: ah, yes.it is generic.
it is not only headers. The server side may have data as well. That will be fix once the server side is fixed, but that is not the highest priority at the moment.

neqo-http3/src/send_message.rs
ddragana: We do already in connection_client.
The server side do not emit DataWritable events at the moment.

PR #661, opened 2020-05-20
Use NSS option to remove EndOfEarlyData

### LINKED ISSUE #659: Remove EOED from TLS transcript too
The hack I implemented to suppress EndOfEarlyData doesn't match the requirements in the spec, which also remove the message from the TLS transcript.

My bad, but this will require changes to NSS to fix :(  Until then, we don't get 0-RTT.

martinthomson:
Unfortunately, this depends on the change landing in NSS, so this is
incomplete.  It will need to be updated with changes that point to the
right NSS commit.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1639413 for the NSS changes.

Closes #659.


martinthomson:
Assuming that CI is OK with this, this is ready to go now as the code is present in gecko.  I've used the commit here, but that will eventually need to use the next RTM tag.

martinthomson:
I realize now that I haven't pushed the docker image, but I'd like to.  If nothing else, please review the docker piece.

PR #663, opened 2020-05-21
Add retry, resumption and zerortt tests

vonasek:
resumption and zerortt tests fail at this moment, but they should pass when issue #659 is fixed

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -118,6 +118,11 @@ pub struct Args {
     #[structopt(name = "qns-mode", long)]
     /// Enable special behavior for use with QUIC Network Simulator
     qns_mode: bool,
+
+    #[structopt(short = "r", long)]
+    /// Client attemps to resume connections when there are multiple connections made.
+    /// Use this for 0-RTT: the stack always attempts 0-RTT on resumption.
+    resume: bool,
 }
 
 trait Handler {
@@ -447,7 +452,12 @@ fn main() -> Res<()> {
                 args.use_old_http = true;
             }
             Ok(s) if s == "zerortt" || s == "resumption" => {
+                if args.urls.len() < 2 {
+                    eprintln!("Warning: resumption tests won't work without >1 URL");
+                    exit(127)
+                }
                 args.use_old_http = true;
+                args.resume = true;
                 resumption_test = true;
             }
             Ok(s) if s == "multiconnect" => {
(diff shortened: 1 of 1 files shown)


neqo-client/src/main.rs
```diff
@@ -492,13 +499,37 @@ fn main() -> Res<()> {
                 &urls,
             )?;
         } else if !args.download_in_series {
+            let mut token: Option<Vec<u8>> = None;
+
+            if resumption_test {
```
martinthomson, line 504: I think that you want to do this differently now that I think on it more.

Rather than take a single token and reuse that on every subsequent request, you should instead use the multiconnect option when resumption is requested.  That keeps the logic here simpler:

```rust
let mut resumption_token = None;
for url in urls {
  token = old::old_client(..., token);
}
```

neqo-client/src/main.rs
```diff
@@ -428,11 +428,17 @@ fn main() -> Res<()> {
...
     if args.qns_mode {
         match env::var("TESTCASE") {
             Ok(s) if s == "http3" => {}
-            Ok(s) if s == "handshake" || s == "transfer" => {
+            Ok(s) if s == "handshake" || s == "transfer" || s == "retry" => {
+                args.use_old_http = true;
+            }
+            Ok(s) if s == "zerortt" || s == "resumption" => {
                 args.use_old_http = true;
+                resumption_test = true;
```
martinthomson, line 441: Add a check here for the number of URLs rather than down below; no point in executing the test without this.

```rust
if resumption_test && args.urls.len < 2 {
    eprintln!("Warning: resumption won't work without >1 URL");
    exit(127); // Not sure about this one, but it seems right.
}
```

You probably still need to issue a warning further down, because there might be fewer URLs for a single host, but this should catch the worst problems.

neqo-client/src/main.rs
```diff
@@ -428,11 +428,17 @@ fn main() -> Res<()> {
 
     let mut args = Args::from_args();
 
+    let mut resumption_test = false;
```
martinthomson, line 431: I would add an option to `Args` for this.  There is general utility in the resumption flag, even if it only works for the multi-connect option right now.  Having to pass separate flags is an awkwardness I'd prefer to avoid.

neqo-client/src/main.rs
vonasek: > Rather than take a single token and reuse that on every subsequent request, you should instead use the multiconnect option when resumption is requested. That keeps the logic here simpler:

That's not how it works now. First URL is fetched to get the token, but all subsequent URL are fetched over a single connection.
 
> ```rust
> let mut resumption_token = None;
> for url in urls {
>   token = old::old_client(..., token);
> }
> ```

This wouldn't work for zerortt test because it downloads 40 files but expects exactly 2 handshakes. With this code we would have 40 handshakes.

martinthomson: I see.  Oh well, carry on.

neqo-client/src/main.rs
vonasek: I added "resume" option, but this option itself doesn't ensure that any additional connection will be made to obtain the token. So it should be combined with "download-in-series" option.

It's enabled implicitly when running "zerortt" or "resumption" test in QNS mode.

PR #665, opened 2020-05-25
Remove DUPLICATE_PUSH frame

ddragana:


ddragana:
This is based on #657

martinthomson:
I would merge this into #657 then.  It's a relatively small change.

ddragana:
Merged into #657

PR #667, opened 2020-05-26
Add test chacha20

vonasek:


neqo-client/src/main.rs
```diff
@@ -519,6 +526,7 @@ fn main() -> Res<()> {
                     remote_addr,
                     &format!("{}", host),
                     &[url],
+                    ciphers,
```
martinthomson, line 529: The API on this is starting to get uncomfortably wide.  I wonder if we need to start breaking this down more.

neqo-transport/src/connection.rs
```diff
@@ -511,6 +512,20 @@ impl Connection {
         Ok(())
     }
 
+    /// Enable a set of ciphers.
+    pub fn enable_ciphers(&mut self, ciphers: &[Cipher]) -> Res<()> {
+        if self.role() != Role::Client {
```
martinthomson, line 517: Please remove this check.  The server has the same needs.

neqo-transport/src/connection.rs
```diff
@@ -511,6 +512,20 @@ impl Connection {
...
+            qerror!([self], "Cannot enable ciphers in role {:?}", self.role);
+            return Err(Error::ConnectionState);
+        }
+        if self.state != State::Init {
```
martinthomson, line 521: Can you also add a similar check in SecretAgent::enable_ciphers?  Belt and braces.

neqo-client/src/main.rs
vonasek: Maybe. For now I've added allow(clippy::too_many_arguments) to make clippy happy.

PR #669, opened 2020-05-26
Drop packets with unknow versions

### LINKED ISSUE #668: Panic "Attempted to get space from wrong packet type" when running handshakecorruption test
[neqo-client.log](https://github.com/mozilla/neqo/files/4682953/neqo-client.log)

ddragana:
fixes #668

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -876,7 +876,9 @@ impl Connection {
                     self.handle_retry(packet)?;
                     return Ok(frames);
                 }
-                (PacketType::VersionNegotiation, ..) | (PacketType::Retry, ..) => {
+                (PacketType::VersionNegotiation, ..)
+                | (PacketType::Retry, ..)
+                | (PacketType::OtherVersion, ..) => {
                     qwarn!("dropping {:?}", packet.packet_type());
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection.rs
```diff
@@ -4791,4 +4793,21 @@ mod tests {
         assert_ne!(delay, Duration::from_secs(0));
         assert!(delay > lr_time);
     }
+
+    #[test]
+    fn unknow_version() {
```
martinthomson, line 4798:
```suggestion
    fn unknown_version() {
```

neqo-transport/src/connection.rs
```diff
@@ -4791,4 +4793,21 @@ mod tests {
...
+        let mut client = default_client();
+        // Start the handshake.
+        let _ = client.process(None, now()).dgram();
+
+        const UNKNOW_VERSION_PACKET: &[u8] = &[
```
martinthomson, line 4803:
```suggestion
        const UNKNOWN_VERSION_PACKET: &[u8] = &[
```

This could be smaller.  But it might be better to pad to 1200 so that we hit the full "initial" path.  For that, you might be better using `Vec::resize`

Issue #300, opened 2019-11-13
Implement server push, client side


ddragana:

ddragana:
Fixed by #677 and #695

Issue #342, opened 2019-12-06
qpack: Decode push_promise's header block 


ddragana:
This is what has remain to be implemented for the client side of server push (#341).
This needs changes to qpack, currently qpack remembers blocked stream by stream_id and push_promise does not need to block the stream it was received on. Therefore we should change qpack that it work with stream_id or push_id.

ddragana:
@KershawChang you can look at this one. We can check if needed, I can give you some more pointers.

PR #677, opened 2020-05-29
Server Push: client side only

ddragana:


ddragana:
This will fix #300

neqo-http3/src/push_controller.rs
```diff
@@ -25,15 +96,333 @@ impl Display for PushController {
...
+                PushState::Init => {
+                    self.conn_events
+                        .push_promise(push_id, ref_stream_id, header_block.clone());
+                    *push_state = PushState::PushPromise {
+                        headers: header_block,
+                    };
+                    Ok(())
+                }
+                PushState::PushPromise { headers } | PushState::Active { headers, .. } => {
+                    if header_block != *headers {
```
martinthomson, line 138: Are you asserting that the headers are exactly identical?  That's going to take a fair bit of work, right?

I say that because you will be creating an entry on line 117, and then comparing it against itself, which isn't ideal.  Can that drop out early?

neqo-http3/src/push_controller.rs
```diff
@@ -1,20 +1,91 @@
...
+}
 
 impl PushController {
-    pub fn new() -> Self {
-        PushController {}
+    pub fn new(max_concurent_push: u64, conn_events: Http3ClientEvents) -> Self {
+        PushController {
+            max_concurent_push,
+            current_max_push_id: 0,
+            push_streams: BTreeMap::new(),
```
martinthomson, line 85: I think that this would be better as a `VecDeque`.  It requires keeping one extra piece of state (the push_id of the first item), but it would be far more performant and some of the other utilities would be nicer.  You could wrap it in a small struct so that indexing is easy.

neqo-http3/src/push_controller.rs
```diff
@@ -1,20 +1,91 @@
...
+            next_push_id_to_open: 0,
```
martinthomson, line 86: If you do as I suggest, then you don't need to track this (the value is `first_push_id + push_streams.len()`).

neqo-http3/src/push_controller.rs
```diff
@@ -25,15 +96,333 @@ impl Display for PushController {
...
+        if push_id > self.current_max_push_id {
+            qerror!("Push id is greater than current_max_push_id.");
+            return Err(Error::HttpId);
+        }
+
+        while self.next_push_id_to_open <= push_id {
+            self.push_streams
+                .insert(self.next_push_id_to_open, PushState::Init);
+            self.next_push_id_to_open += 1;
+        }
```
martinthomson, line 214: You can use `VecDeque::resize` for this (but I would move the function to the collection wrapper.

neqo-http3/src/push_stream.rs
```diff
@@ -0,0 +1,162 @@
...
+impl PushStreamState {
+    pub fn push_id(&self) -> Option<u64> {
+        match self {
+            Self::ReadResponse { push_id, .. } => Some(*push_id),
+            _ => None,
+        }
+    }
+}
+
+// The `PushController` keeps information about all push streams. Each push stream is responsible for contacting the
```
martinthomson, line 35:
```suggestion
/// The `PushController` keeps information about all push streams. Each push stream is responsible for contacting the
```

neqo-http3/src/push_stream.rs
```diff
@@ -0,0 +1,162 @@
...
+impl RecvStream for PushStream {
+    fn receive(&mut self, conn: &mut Connection, decoder: &mut QPackDecoder) -> Res<()> {
+        loop {
+            match &mut self.state {
+                PushStreamState::ReadPushId(id_reader) => {
+                    let push_id = id_reader.get_type(conn, self.stream_id);
+                    let fin = id_reader.fin();
+                    if fin {
+                        self.state = PushStreamState::Closed;
+                        return Ok(());
```
martinthomson, line 97: This really isn't OK, is it?  Is there an event that we need to raise in this case?  It seems like this might be cause for a PushError event.

neqo-http3/src/push_stream.rs
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                    return Ok(());
+                }
+                _ => return Ok(()),
+            }
+        }
+    }
+
+    fn done(&self) -> bool {
+        matches!(self.state, PushStreamState::Closed)
```
martinthomson, line 137: It's probably better to use `==` rather than `matches!()` unless you really have to.

neqo-http3/src/client_events.rs
```diff
@@ -31,10 +31,24 @@ pub enum Http3ClientEvent {
...
+        headers: Vec<u8>,
+    },
+    /// A push response headers are ready.
+    PushHeaderReady {
+        push_id: u64,
+        headers: Option<Vec<Header>>,
+        fin: bool,
+    },
+    /// New bytes are available on a push stream for reading.
+    PushDataReadable { push_id: u64 },
```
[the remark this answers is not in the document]
ddragana: I made it on purpose that a push is only referred by push_id. I will open an issue and do it after this PR lands.
Note: necko will need to remember stream_id besides push_id, I wanted to avoid that.

neqo-http3/src/connection_client.rs
```diff
@@ -31,13 +33,13 @@ use std::time::Instant;
 use crate::{Error, Res};
 
 // This is used for filtering send_streams and recv_Streams with a stream_ids greater that a given id.
-fn id_gte<T, U>(base: T) -> impl FnMut((&T, &U)) -> Option<T> + 'static
+// Only bidirectional streams are filtered.
+fn id_gte_bidi<U>(base: u64) -> impl FnMut((&u64, &U)) -> Option<u64> + 'static
```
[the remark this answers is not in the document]
ddragana: I will leave it here.
Actually I am thinking of moving it back into filter_map. Thinking about this maybe thatt is the most understandable way to write this.

neqo-http3/src/connection_client.rs
```diff
@@ -3608,4 +3866,835 @@ mod tests {
...
+            &[3],
+            request_stream_id,
+        );
+        assert_eq!(client.state(), Http3State::Connected);
+    }
+
+    // Test receiving pushes out of order.
+    // Push_id 5 is received first and read so that it is removed from the list,
+    // therefore Push_id 3 will be in the PushState:Init state.
+    // Start push_id 3 by receiving a push_promise and then a push stream.
```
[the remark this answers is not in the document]
ddragana: I can remove this one.

neqo-http3/src/push_controller.rs
```diff
@@ -25,15 +96,333 @@ impl Display for PushController {
...
+        match self.push_streams.get_mut(&push_id) {
+            None => {
+                qtrace!(
+                    "Push has been closed already {} {}.",
+                    self.next_push_id_to_open,
+                    push_id
+                );
+                Ok(())
+            }
+            Some(push_state) => match push_state {
```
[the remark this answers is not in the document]
ddragana: clippy is turned on for this file. Actually for everything in neqo-http3 and neqo-qpack.
Clippy does not complain about this line.

neqo-http3/src/push_controller.rs
ddragana: Line 117 creates only an entry in Init state and in that state we do not compare headers. It is never compared against itself.

These headers are not decompress yet (Issue #342 ). This will not work correctly until the issue is fixed.

neqo-http3/src/push_stream.rs
ddragana: Implementing PartalEq for PushStreamState is not easy (NewStreamTypeReader, RecvMessage, HFrameReader, etc need to implement PartilaEq).
I could do a partial implementation of it, but I decided not to do it

neqo-http3/src/push_controller.rs
ddragana: For this I will need a Close state as well. It is doable.

neqo-http3/src/connection_client.rs
```diff
@@ -81,16 +93,25 @@ impl Http3Client {
                 quic_version,
             )?,
             qpack_settings,
+            max_concurent_push_streams,
```
martinthomson, line 96: concurrent

neqo-http3/src/connection_client.rs
```diff
@@ -299,14 +327,46 @@ impl Http3Client {
...
+
+    /// Push response data are read directly into a buffer supplied as a parameter of this function
+    /// to avoid copying data.
+    /// # Errors
+    /// It returns an error if a stream does not exist(`InvalidStreamId`) or an error has happened while
+    /// reading a stream, e.g. early close, protocol error, etc.
+    pub fn push_read_data(
+        &mut self,
+        now: Instant,
+        push_id: u64,
```
martinthomson, line 357: I think that we want a newtype for PushId, just like StreamId.

neqo-http3/src/connection_client.rs
```diff
@@ -2711,7 +3001,7 @@ mod tests {
 
         let out = server.conn.process(None, now());
         client.process(out.dgram(), now());
-        client.process(None, now());
+        //        client.process(None, now());
```
martinthomson, line 3004: Don't leave commented-out code lying around.

neqo-http3/src/lib.rs
```diff
@@ -99,6 +102,24 @@ impl Error {
             _ => 3,
         }
     }
+
+    // TODO: dragana look into http3 errors
```
martinthomson, line 106: Can you be more specific?

ddragana: I am looking into this #757 . I will write down which error can happen in which function and how and where to handle it (my last 2 patches are made while I was looking into this and some more are coming)

neqo-http3/src/connection_client.rs
```diff
@@ -30,21 +32,30 @@ use std::time::Instant;
 
 use crate::{Error, Res};
 
-// This is used for filtering send_streams and recv_Streams with a stream_ids greater that a given id.
-fn id_gte<T, U>(base: T) -> impl FnMut((&T, &U)) -> Option<T> + 'static
+// This is used for filtering send_streams and recv_Streams with a stream_ids greater than or equal a given id.
+// Only the same type (bidirectional or unidirectionsl) streams are filtered.
+fn id_gte<U>(base: StreamId) -> impl FnMut((&u64, &U)) -> Option<u64> + 'static
```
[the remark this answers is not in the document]
ddragana: the whole http3 code uses u64 instead of StreamId for the beginning(the reason was that I did not want to expose StreamId to applications using http3, and it was better, at the time to use u64 in whole http3 code....). We can change that, but let's do it in a different PR.

neqo-http3/src/connection_client.rs
ddragana: Let's do that in a separate PR, this one is too big and I really want to land it.
I have open issue #785

PR #681, opened 2020-06-02
Factor some of the ChaCha20 stuff

martinthomson:
Mainly this was so that I could construct a sample ChaCha20 header
protection example for the spec, but I think that this is cleaner
overall, and more usable.

There are some improvements to the neqo-client interface as well.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -10,10 +10,13 @@
 use qlog::QlogStreamer;
 
 use neqo_common::{self as common, hex, matches, qlog::NeqoQlog, Datagram, Role};
-use neqo_crypto::{init, AuthenticationStatus, Cipher, TLS_CHACHA20_POLY1305_SHA256};
+use neqo_crypto::{
+    constants::{TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256},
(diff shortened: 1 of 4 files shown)


martinthomson:
The documentation is a little vague, but it is careful to distinguish between *instances* and *values*.

It's best illustrated by an example.

If you have no limit to the number of values, then the flag works by consuming arguments until it detects another flag.  So `--one a b c d --two e f g` attributes `a` through `d` to `--one` and `e` through `g` to `--two`.  That means that if you also have positional parameters, you need to do something like `--one a b c d -- positional` (I think).  Even with this, you can also repeat the flag, so if you get `--one a --one b c --two d` has two instances of `--one`, the first with a single value of `a` and the second with two values (`b` and `c`).  The net effect being that you get `a` through `c` in the array.

If you limit the number of values (1 is a good limit, but we use two for headers) then once you hit that limit, the values stop accruing to the array and you get something sensible like `--alpn h3-27 https://example.com/` only putting a single value into `alpn: Vec<String>`.  That means that if want to provide multiple ALPN strings, you need to provide multiple instances, like so: `--alpn h3-27 --alpn h3-28 https://example.com/`.

Does that make sense?

martinthomson:
@vonasek, could you take a look at this please?  I'm happy if you defer to Andy if you don't feel like you can.

PR #682, opened 2020-06-02
neqo-client: fix multiplexing test

vonasek:
All URLs are first queued and then we start downloading as many URLs as possible at the same time. If we get StreamLimitError from Connection::stream_create() we pause the download queue until we receive ConnectionEvent::SendStreamCreatable.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -582,10 +582,11 @@ fn main() -> Res<()> {
 
 mod old {
     use std::cell::RefCell;
-    use std::collections::HashMap;
+    use std::collections::{HashMap, VecDeque};
     use std::fs::File;
     use std::io::{ErrorKind, Write};
     use std::net::{SocketAddr, UdpSocket};
+    use std::path::PathBuf;
     use std::process::exit;
     use std::rc::Rc;
     use std::time::Instant;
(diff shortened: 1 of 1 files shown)


neqo-client/src/main.rs
```diff
@@ -597,18 +598,22 @@ mod old {
...
     use super::{emit_datagram, get_output_file, Args};
 
     trait HandlerOld {
-        fn handle(&mut self, args: &Args, client: &mut Connection) -> Res<bool>;
+        fn handle(&mut self, client: &mut Connection) -> Res<bool>;
+    }
+
+    trait DownloadNext {
+        fn download_next(&mut self, client: &mut Connection) -> Res<bool>;
     }
```
agrover, line 612: I don't think you need this trait, because you're only calling `download_next` on the `PostConnectHandlerOld` concrete type. Instead, create an impl block for `PostConnectHandlerOld`:
```
    impl<'a> PostConnectHandlerOld<'a> {
         fn download_next(blah) { }
    }
```
after its struct definition and put the method there.

neqo-client/src/main.rs
```diff
@@ -678,21 +701,51 @@ mod old {
         }
     }
 
+    impl DownloadNext for PostConnectHandlerOld<'_> {
+        fn download_next(&mut self, client: &mut Connection) -> Res<bool> {
+            let url = self
+                .url_queue
+                .pop()
+                .expect("download_next called with empty queue");
```
agrover, line 709: This will create streams for URLs in LIFO order. If instead of a Vec `url_queue` was a VecDeque we could `push_front` & `pop_front`.

agrover:
Also please edit commit message to break it into a short (< 50-65 chars) summary line, then a blank line, then additional commitlog text.

PR #690, opened 2020-06-05
Ignore VersionNegotiation containing reserved values and the value of the used QUIC version

### LINKED ISSUE #675: We often fail to establish connection when handshake packet is corrupted
Test handshakecorruption fails very often:
[neqo-client.1.log](https://github.com/mozilla/neqo/files/4695444/neqo-client.1.log)
[neqo-client.2.log](https://github.com/mozilla/neqo/files/4695445/neqo-client.2.log)

ddragana:
fixes #675

This also add test server_receive_unknow_first_packet which was crashing when trying to get timers (I have accidentally made a wrong test that revealed the issue)

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -943,7 +943,7 @@ impl Connection {
         match self.output(now) {
             SendOption::Yes(dgram) => Output::Datagram(dgram),
             SendOption::No(paced) => match self.state {
-                State::Closed(_) => Output::None,
+                State::Init | State::Closed(_) => Output::None,
                 State::Closing { timeout, .. } | State::Draining { timeout, .. } => {
                     Output::Callback(timeout.duration_since(now))
                 }
@@ -1102,10 +1102,24 @@ impl Connection {
                     }
                 }
                 (PacketType::VersionNegotiation, State::WaitInitial, Role::Client) => {
-                    self.set_state(State::Closed(ConnectionError::Transport(
-                        Error::VersionNegotiation,
-                    )));
-                    return Err(Error::VersionNegotiation);
+                    match packet.supported_versions() {
+                        Ok(versions) => {
+                            if versions.is_empty() || versions.contains(&self.quic_version.as_u32())
+                            {
+                                // Ignore VersionNegotiation packets that contain the current version.
+                                self.stats.dropped_rx += 1;
+                                return Ok(frames);
+                            }
+                            self.set_state(State::Closed(ConnectionError::Transport(
+                                Error::VersionNegotiation,
+                            )));
+                            return Err(Error::VersionNegotiation);
+                        }
+                        Err(_) => {
+                            self.stats.dropped_rx += 1;
+                            return Ok(frames);
+                        }
+                    }
                 }
                 (PacketType::Retry, State::WaitInitial, Role::Client) => {
                     self.handle_retry(packet)?;
@@ -2495,6 +2509,7 @@ mod tests {
     use crate::cc::PACING_BURST_SIZE;
     use crate::cc::{INITIAL_CWND_PKTS, MIN_CONG_WINDOW};
     use crate::frame::{CloseError, StreamType};
+    use crate::packet::PACKET_BIT_LONG;
     use crate::path::PATH_MTU_V6;
     use crate::recovery::ACK_ONLY_SIZE_LIMIT;
     use crate::recovery::PTO_PACKET_COUNT;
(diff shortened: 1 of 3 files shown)


neqo-transport/tests/server.rs
```diff
@@ -717,14 +717,10 @@ fn version_negotiation() {
     }
     assert!(found, "valid version not found");
 
+    // Client ignores VN packet that contain negotiatied version.
```
agrover, line 720: typo: negotiated

neqo-transport/src/packet.rs
```diff
@@ -614,6 +614,20 @@ impl<'a> PublicPacket<'a> {
             Err(Error::DecryptError)
         }
     }
+
+    pub fn decode_version_negotiation(&self) -> Res<Vec<Version>> {
```
agrover, line 618: let's call this something a little less about the implementation details and more about the result. Maybe `supported_versions()`? Or you tell me.

neqo-transport/src/packet.rs
```diff
@@ -614,6 +614,20 @@ impl<'a> PublicPacket<'a> {
...
     }
+
+    pub fn supported_versions(&self) -> Res<Vec<Version>> {
+        assert_eq!(self.packet_type, PacketType::VersionNegotiation);
+        let mut decoder = Decoder::new(&self.data[self.header_len..]);
+        let mut res = Vec::new();
+        while decoder.remaining() > 0 {
+            let version = Version::try_from(Self::opt(decoder.decode_uint(4))?)?;
+            // Ignore reserved versions.
+            if (version & 0x0f0f_0f0f) != 0x0a0a_0a0a {
```
martinthomson, line 625: See above.

ddragana:
> I'm a little uncomfortable with the idea that we might ignore reserved versions. It goes against the spirit of having reserved versions. I think that if we receive a response that contains only reserved versions, we should just let the connection fail.

you are right.

neqo-transport/src/connection.rs
```diff
@@ -1026,6 +1026,15 @@ impl Connection {
                     self.crypto.states.init(self.role, &packet.dcid());
                 }
                 (PacketType::VersionNegotiation, State::WaitInitial, Role::Client) => {
+                    if let Ok(versions) = packet.supported_versions() {
+                        if versions.contains(&QUIC_VERSION) {
+                            // Ignore VersionNegotiation packets that contain the current version or only
+                            // contain reserved versions. (decode_version_negotiation will remove reserved
+                            // versions)
```
martinthomson, line 1033: Update this comment.

```suggestion
                            // Ignore VersionNegotiation packets that contain the current version.
```

neqo-transport/src/connection.rs
```diff
@@ -5173,4 +5183,141 @@ mod tests {
...
+    }
+
+    #[test]
+    fn version_negotiation_current_version() {
+        let mut client = default_client();
+
+        let vn = get_vn(&mut client, &[0x1a1_a1a1a, QUIC_VERSION]);
+
+        assert_eq!(
+            client.process(Some(Datagram::new(loopback(), loopback(), vn,)), now(),),
```
martinthomson, line 5240: This is an odd time to be processing a VN packet.  Can you make this a little more conventional in the sense that it is received after sending the first Initial?

neqo-transport/src/connection.rs
```diff
@@ -5173,4 +5183,141 @@ mod tests {
...
+            client.process(
+                Some(Datagram::new(loopback(), loopback(), &vn[..vn.len() - 1])),
+                now(),
+            ),
+            Output::None
+        );
+        match client.state() {
+            State::Closed(err) => {
+                assert_eq!(*err, ConnectionError::Transport(Error::VersionNegotiation))
+            }
```
martinthomson, line 5281: Shouldn't we be ignoring corrupted version negotiation packets too?

neqo-transport/src/connection.rs
```diff
@@ -5173,4 +5183,141 @@ mod tests {
...
+
+    #[test]
+    fn version_negotiation_empty() {
+        let mut client = default_client();
+
+        let vn = get_vn(&mut client, &[]);
+
+        assert_eq!(
+            client.process(
+                Some(Datagram::new(loopback(), loopback(), &vn[..vn.len() - 1])),
```
martinthomson, line 5294: Check the truncation here.

neqo-transport/src/connection.rs
```diff
@@ -5173,4 +5183,141 @@ mod tests {
...
+        match client.state() {
+            State::Closed(err) => {
+                assert_eq!(*err, ConnectionError::Transport(Error::VersionNegotiation))
+            }
+            _ => panic!("Invalid client state"),
+        }
+    }
+
+    #[test]
+    fn version_negotiation_empty() {
```
martinthomson, line 5287: Shouldn't this have the same effect as the corrupted one?  That is, shouldn't we ignore it?

neqo-transport/src/connection.rs
ddragana: The initial is sent inside get_vn(). I moved it there because we need the Intitial packet to retrieve cids.

agrover:
@ddragana please resolve conflicts.
@martinthomson please re-review
I'd like to get this in asap.

neqo-transport/src/connection.rs
martinthomson: OK, that's misunderstanding on my part.  Renaming get_vn to handle_vn might work, but you would have to have it return the result of the client.process() call. The advantage of that is that you save another line of code from every test.

neqo-transport/src/connection.rs
```diff
@@ -5173,4 +5187,131 @@ mod tests {
...
+        assert_eq!(
+            server.process(
+                Some(Datagram::new(
+                    loopback(),
+                    loopback(),
+                    unknown_version_packet,
+                )),
+                now(),
+            ),
+            Output::None
```
martinthomson, line 5207: Ahh, the reason that this is not generating a VN response is that we generate that in Server not Connection.  That took me longer to realize than I should be happy to admit.  Could you add a comment explaining that?

Issue #655, opened 2020-05-19
Occasional STOP_SENDING from client


martinthomson:
I am sometimes getting 0-RTT working, but other times the client doesn't recognize the QPACK encoder stream.  I'm getting this:

```
0s213ms TRACE [STREAM] stream 7 0->3 = [3]: 3f6102
...
0s215ms TRACE [Http3 connection] Readable stream 7.
0s215ms DEBUG Unknown stream.
```

That's the first three bytes as unrecognizable QPACK goop rather than the stream type (2 for encoder, 3 for decoder).  We have a decoder stream already (stream 11), so I'm guessing that this is the encoder stream.

This always seems to occur, even when the client is successful.  I guess it is down to timing as to whether the client is able to send the STOP_SENDING and get the RESET_STREAM back in time to affect the success of the connection.

martinthomson:
This appears to be the result of a bad ordering between the stream type and a capacity change.  That is an HPACK integer of 128 with a 3-bit prefix of 0b001 before the stream type.  This is the result of buffering instructions in send_buf in the encoder.

I think that we should not be buffering anything on this stream.  All writes should be atomic and all attempts to insert should fail until the stream header (type + capacity instruction) is written.  Capacity changes thereafter can fail if there isn't enough space to write the instruction.

martinthomson:
Oh, now I see how this might be my fault.  I'm doing 0-RTT so the client might be getting its settings in before the server can make the streams.  That results in buffering the capacity change instruction before the stream type.  That doesn't change the outcome: we shouldn't be buffering anything.  If we never get enough flow control to send type + capacity, then we don't need to use the dynamic table.

ddragana:
I use buffering just because I do not want to keep states. Basically the buffer is the state. Before using the dynamic table we check if send_buf is empty and if it is not we do not use dynamic table. You can change send_buf to be a state "DynamicTableUninitialize" and "DynamicTableInitialize" but there will not be a lot of change to the code.
Sending capacity change instruction after initialization will get more complicated (WantToChangeCapacity state or something). I still prefer just keeping send_buf

ddragana:
Fixed by #691

PR #691, opened 2020-06-05
The qpack encoder should write all instructions directly into a transport stream

ddragana:
Probably this is all that is needed for #655

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -14,7 +14,7 @@ use crate::stats::Stats;
 use crate::table::{HeaderTable, LookupResult, ADDITIONAL_TABLE_ENTRY_SIZE};
 use crate::Header;
 use crate::{Error, QpackSettings, Res};
-use neqo_common::{qdebug, qlog::NeqoQlog, qtrace};
+use neqo_common::{qdebug, qerror, qlog::NeqoQlog, qtrace};
 use neqo_transport::Connection;
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::convert::TryFrom;
@@ -100,7 +100,7 @@ impl QPackEncoder {
         let new_cap = std::cmp::min(self.max_table_size, cap);
         self.max_entries = new_cap / 32;
         // we also set our table to the max allowed.
-        self.change_capacity(new_cap);
+        self.need_change_capacity(new_cap);
         Ok(())
     }
 
@@ -278,20 +278,19 @@ impl QPackEncoder {
         }
     }
 
-    fn change_capacity(&mut self, value: u64) {
+    fn need_change_capacity(&mut self, value: u64) {
         qdebug!([self], "change capacity: {}", value);
         self.change_capacity = Some(value);
     }
 
     fn maybe_send_change_capacity(&mut self, conn: &mut Connection, stream_id: u64) -> Res<()> {
-        if let Some(cap) = self.change_capacity {
+        if let Some(cap) = self.change_capacity.take() {
             let mut buf = QPData::default();
             EncoderInstruction::Capacity { value: cap }.marshal(&mut buf, self.use_huffman);
             if !conn.stream_send_atomic(stream_id, &buf)? {
                 return Err(Error::EncoderStreamBlocked);
             }
             self.table.set_capacity(cap)?;
-            self.change_capacity = None;
         }
         Ok(())
     }
@@ -301,7 +300,10 @@ impl QPackEncoder {
     ///   returns `EncoderStream` in case of an error.
     pub fn send(&mut self, conn: &mut Connection) -> Res<()> {
         match self.local_stream {
-            LocalStreamState::NoStream => Ok(()),
+            LocalStreamState::NoStream => {
+                qerror!("Send call but there is no stream yet.");
+                Ok(())
+            }
             LocalStreamState::Uninitialized(stream_id) => {
                 let mut buf = QPData::default();
                 buf.encode_varint(QPACK_UNI_STREAM_TYPE_ENCODER);
@@ -499,6 +501,14 @@ mod tests {
         peer_conn: Connection,
     }
 
+    impl TestEncoder {
+        pub fn change_capacity(&mut self, capacity: u64) -> Res<()> {
+            self.encoder.set_max_capacity(capacity).unwrap();
+            // We will try to really change the table only when we send the change capacity instruction.
+            self.encoder.send(&mut self.conn)
+        }
+    }
+
     fn connect_generic<F>(huffman: bool, f: F) -> TestEncoder
     where
         F: FnOnce(&mut Connection, &mut Connection),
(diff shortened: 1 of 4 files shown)


neqo-qpack/src/encoder.rs
```diff
@@ -257,28 +271,42 @@ impl QPackEncoder {
...
+
+    fn maybe_send_change_capacity(&mut self, conn: &mut Connection, stream_id: u64) -> Res<()> {
+        if let Some(cap) = self.change_capacity {
+            let mut buf = QPData::default();
+            EncoderInstruction::Capacity { value: cap }.marshal(&mut buf, self.use_huffman);
+            if !conn.stream_send_atomic(stream_id, &buf)? {
+                return Err(Error::EncoderStreamBlocked);
+            }
+            self.table.set_capacity(cap)?;
+            self.change_capacity = None;
```
agrover, line 287: If you use `self.change_capacity.take()` ([ref](https://doc.rust-lang.org/std/option/enum.Option.html#method.take)) you don't need to assign to `None` at the bottom.

neqo-qpack/src/encoder.rs
```diff
@@ -81,7 +97,8 @@ impl QPackEncoder {
         let new_cap = std::cmp::min(self.max_table_size, cap);
         self.max_entries = new_cap / 32;
         // we also set our table to the max allowed.
-        self.change_capacity(new_cap)
+        self.change_capacity(new_cap);
```
agrover, line 100: So the max capacity might not actually change? Maybe we don't want to call `change_capacity()` if it hasn't actually increased?

neqo-qpack/src/encoder.rs
```diff
@@ -257,28 +271,42 @@ impl QPackEncoder {
...
-            let r = conn
-                .stream_send(stream_id, &self.send_buf[..])
-                .map_err(|_| Error::EncoderStream)?;
-            qdebug!([self], "{} bytes sent.", r);
-            self.send_buf.read(r as usize);
-            Ok(())
-        } else {
-            Ok(())
+        match self.local_stream {
+            LocalStreamState::NoStream => Ok(()),
```
agrover, line 297: Just wanted to check this should be `Ok(())`. Is calling `send()` allowed when there isn't even an uninitialized stream? Does this deserve to be an error or emit a log message?

neqo-qpack/src/encoder.rs
```diff
@@ -509,6 +537,12 @@ mod tests {
         })
     }
 
+    fn change_capacity(encoder: &mut TestEncoder, capacity: u64) -> Res<()> {
```
martinthomson, line 540: Can you make this a method on TestEncoder?

neqo-qpack/src/encoder.rs
```diff
@@ -257,28 +271,44 @@ impl QPackEncoder {
...
-    fn change_capacity(&mut self, value: u64) -> Res<()> {
+    fn need_change_capacity(&mut self, value: u64) {
         qdebug!([self], "change capacity: {}", value);
-        self.table.set_capacity(value)?;
-        EncoderInstruction::Capacity { value }.marshal(&mut self.send_buf, self.use_huffman);
+        self.change_capacity = Some(value);
+    }
+
+    fn maybe_send_change_capacity(&mut self, conn: &mut Connection, stream_id: u64) -> Res<()> {
+        if let Some(cap) = self.change_capacity.take() {
```
martinthomson, line 280: If this write fails, then you won't send the capacity change.  So you should defer clearing the value until the write succeeds.

neqo-qpack/src/encoder.rs
```diff
@@ -257,28 +271,44 @@ impl QPackEncoder {
...
+            let mut buf = QPData::default();
+            EncoderInstruction::Capacity { value: cap }.marshal(&mut buf, self.use_huffman);
+            if !conn.stream_send_atomic(stream_id, &buf)? {
+                return Err(Error::EncoderStreamBlocked);
+            }
+            self.table.set_capacity(cap)?;
```
martinthomson, line 286: As we have two fallible operations here, this is tricky.  I would suggest that you try to this (and attempt to perform evictions) first as that operation - if it succeeds - won't result in a half-committed state.  If we manage to set capacity, but not send the instruction, then we can still try to set the capacity again later before writing another instruction.  If we can't evict, but send the instruction, then we will have violated the protocol (probably).

neqo-qpack/src/encoder.rs
```diff
@@ -257,28 +271,44 @@ impl QPackEncoder {
         }
     }
 
-    fn change_capacity(&mut self, value: u64) -> Res<()> {
+    fn need_change_capacity(&mut self, value: u64) {
```
martinthomson, line 274: Names are hard, but I think that a function called `change_capacity` is very clear.  If anything, the variable name is what needs changing, to `next_capacity` or something perhaps.

neqo-qpack/src/encoder.rs
```diff
@@ -14,21 +14,36 @@ use crate::stats::Stats;
...
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::convert::TryFrom;
 
 pub const QPACK_UNI_STREAM_TYPE_ENCODER: u64 = 0x2;
 
+#[derive(Debug, PartialEq)]
+enum LocalStreamState {
+    NoStream,
+    Uninitialized(u64),
+    Initialized(u64),
```
martinthomson, line 28: If these are stream IDs, can you use `StreamId`.  If for no other reason than we have far too many u64s floating around.

neqo-qpack/src/encoder.rs
```diff
@@ -264,28 +278,56 @@ impl QPackEncoder {
...
+    fn change_capacity(&mut self, value: u64) {
         qdebug!([self], "change capacity: {}", value);
-        self.table.set_capacity(value)?;
-        EncoderInstruction::Capacity { value }.marshal(&mut self.send_buf, self.use_huffman);
+        self.next_capacity = Some(value);
+    }
+
+    fn maybe_send_change_capacity(&mut self, conn: &mut Connection, stream_id: u64) -> Res<()> {
+        if let Some(cap) = self.next_capacity {
+            // Check if itt is possible to reduce the capacity, e.g. if enough space can be make free for the reduction.
```
martinthomson, line 288:
```suggestion
            // Check if it is possible to reduce the capacity, e.g. if enough space can be make free for the reduction.
```

neqo-qpack/src/encoder.rs
```diff
@@ -264,28 +278,56 @@ impl QPackEncoder {
...
+            if cap < self.table.capacity() && !self.table.test_evict_to(cap) {
```
martinthomson, line 289: This might be better named `can_evict_to`

neqo-qpack/src/encoder.rs
```diff
@@ -264,28 +278,56 @@ impl QPackEncoder {
...
+            }
+            let mut buf = QPData::default();
+            EncoderInstruction::Capacity { value: cap }.marshal(&mut buf, self.use_huffman);
+            if !conn.stream_send_atomic(stream_id, &buf)? {
+                return Err(Error::EncoderStreamBlocked);
+            }
+            if self.table.set_capacity(cap).is_err() {
+                debug_assert!(
+                    false,
+                    "test_evict_to should have checked and make sure this operation is possible"
```
martinthomson, line 300: If you change the name:
```suggestion
                    "can_evict_to should have checked and make sure this operation is possible"
```

neqo-qpack/src/encoder.rs
ddragana: I will switch it to StreamId here.

PR #695, opened 2020-06-08
Decoder PushPromise headers

### LINKED ISSUE #342: qpack: Decode push_promise's header block

ddragana:
Fixes #342 

This is base on #677

### DIFF
--- a/neqo-http3/src/push_controller.rs
+++ b/neqo-http3/src/push_controller.rs
@@ -32,7 +32,7 @@ use std::slice::SliceIndex;
 enum PushState {
     Init,
     PushPromise {
-        headers: Vec<u8>,
+        headers: Vec<Header>,
     },
     OnlyPushStream {
         stream_id: u64,
@@ -40,7 +40,7 @@ enum PushState {
     },
     Active {
         stream_id: u64,
-        headers: Vec<u8>,
+        headers: Vec<Header>,
     },
     Closed,
 }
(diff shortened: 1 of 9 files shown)


neqo-http3/src/push_controller.rs
```diff
@@ -25,15 +170,324 @@ impl Display for PushController {
...
+                    qerror!(
+                        "A server should not send CANCEL_PUSH after a push stream has been opened."
+                    );
+                    let _ = base_handler.stream_reset(
+                        conn,
+                        stream_id,
+                        Error::HttpRequestCancelled.code(),
+                    );
+                    self.conn_events.remove_events_for_push_id(push_id);
+                    self.conn_events.push_canceled(push_id);
```
martinthomson, line 326: Any reason that you would not want to do both of these in OnlyPushStream as well?  Then you could collapse the two arms.

neqo-http3/src/push_controller.rs
```diff
@@ -25,15 +170,324 @@ impl Display for PushController {
...
+        match self.push_streams.get(push_id) {
+            Some(PushState::Active { stream_id, .. }) => Some(*stream_id),
+            _ => None,
+        }
+    }
+
+    pub fn maybe_send_max_push_id_frame(&mut self, base_handler: &mut Http3Connection) {
+        let push_done = self.push_streams.number_done();
+        if self.max_concurent_push > 0
+            && (self.current_max_push_id - push_done) <= (self.max_concurent_push / 2)
```
martinthomson, line 417: I think that we can probably be a lot more aggressive about sending this one.  In the extreme, it would be OK to send an update for every push completion.  But you could set a value that is min(self.max_concurrent_push, SMALL_NUMBER_LIKE_3) if you wanted to limit the rate a little.

ddragana:
> This is pretty good, but it would be better rebased and landed.

This depends on #677.
@martinthomson , does #677 need another review?

PR #701, opened 2020-06-10
Split retry tests

martinthomson:
This runs all the tests.  The only non-deterministic aspect is the code,
not the tests themselves.


martinthomson:
Woo, I managed to crash the linker with that last CI run.  I can, however, promise that the code builds and tests run and pass.

PR #702, opened 2020-06-10
Factor connection ID management

martinthomson:
Working on this uncovered a real bug in the server: we were passing the
wrong connection ID manager to new connections.

Most of this is stuff we discussed, but you didn't have time for.  In particular, I really didn't like that the server required an initial source connection ID as an extra parameter.  Getting that to fit in with all the path management stuff was a little fiddly, but I think that this is much cleaner in the end.


neqo-transport/src/server.rs
```diff
@@ -434,47 +458,42 @@ impl Server {
...
         }));
 
         let sconn = Connection::new_server(
             &self.certs,
             &self.protocols,
             &self.anti_replay,
-            self.cid_manager.clone(),
-            quic_version,
-            local_initial_source_cid.clone(),
+            cid_mgr.clone(),
```
martinthomson, line 481: The bug was here.  I didn't pass in the wrapper.  It was the underlying manager that was being passed.

That meant that any new connection IDs we might generate wouldn't be registered properly.

neqo-transport/src/connection.rs
```diff
@@ -611,25 +603,39 @@ impl Connection {
...
-        self.tps
-            .borrow_mut()
-            .local
-            .set_bytes(tparams::ORIGINAL_DESTINATION_CONNECTION_ID, odcid.to_vec());
-        self.remote_original_destination_cid = Some(odcid);
+    /// `odcid` is their original choice for our CID, which we get from the Retry token.
+    /// `remote_cid` is the value from the Source Connection ID field of
+    ///   an incoming packet: what the peer wants us to use now.
+    /// `retry_cid` is what we asked them to use when we sent the Retry.
+    pub(crate) fn set_retry_cids(
```
martinthomson, line 610: This wraps up several functions from before (those shown on the left).

One thing that this does not do is set `remote_original_destination_cid`.  The server doesn't use it.

PR #707, opened 2020-06-10
Add an interop test for qpack.

ddragana:


neqo-qpack/src/stats.rs
```diff
@@ -0,0 +1,13 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+// Tracking of some useful statistics.
+
+#[derive(Default, Debug)]
+/// `QPack` statistics
+pub struct Stats {
+    pub dynamic_table_used: bool,
```
martinthomson, line 12: Why not count the number of entries inserted instead?

One problem with this as it is is that you count a change to the maximum capacity as "use", which isn't really what you want.  You want the number of insertions AND the number of references to the table to be non-zero.

ddragana: I actually did not want to merge this as it is. I wrote this quickly for the interop. 
Yes all your comments are valid and should be change.

We need some stats like this anyway for telemetry (I am working on a doc that collects all probes we want.)

neqo-qpack/src/stats.rs
```diff
@@ -0,0 +1,14 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+// Tracking of some useful statistics.
+
+#[derive(Default, Debug)]
+/// `QPack` statistics
+pub struct Stats {
+    pub dynamic_table_inserts: u64,
+    pub header_blocks_ref_dynamic_table: u64,
```
martinthomson, line 13: Perhaps this would be clearer as `dynamic_table_references` with a comment explaining what the number represents.

martinthomson, line 13: Both of these are counts and so should be `usize`.

PR #716, opened 2020-06-12
Use display

### LINKED ISSUE #672: Macro for display functions
Having to write out the display function (which we use in logging) is annoying.  A macro that routed to `format!()` would make this a lot less painful.

martinthomson:
I started to use the code in #672 (which was great), but found that it needed a few tweaks to get into shape.  This is the conversion and the tweaks:

1. The type now accepts lifetime and type parameters.

2. More complex expressions are possible, not just member references.

3. I added a macro to forward to a `Debug` implementation, which we were using for `Error` implementations.

Closes #672 because it includes that code.

### DIFF
--- a/neqo-common/src/display.rs
+++ b/neqo-common/src/display.rs
@@ -1,0 +1,83 @@
+#[macro_export]
+macro_rules! display {
+
+    ($name: ident, $formatter: expr) => {
+        display!($name, self, $formatter,)
+    };
+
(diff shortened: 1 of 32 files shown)


agrover:
How does this compare to using https://docs.rs/parse-display/0.1.2/parse_display/ for this extremely (ahem) pressing and urgent problem?

martinthomson:
parse_display seems like a much better option.

PR #718, opened 2020-06-12
Save packets for which we don't have keys for later

### LINKED ISSUE #714: Don't discard first 1RTT data packet from ngtcp2
As shown in [this qlog](https://nghttp2.org/qlog/ad2af6b7fd03f17efc20c6c68b54699c08bc.qlog) at time 4 ngtcp2 is sending a 1RTT data stream right after handshake crypto frames, but the 1RTT packet is being dropped due to decryption failure. In chat, @martinthomson said an async cert validation thing might be the reason.

martinthomson:
In order to test this, I had to add support for the server sending prior
to the handshake completing.  The nice thing is that we can use
availability of transport parameters to drive this.  If we have them,
then the server will be able to send.

There is a change in the HTTP code that bothers me a little.  There is a
new event being generated.  I don't understand why that event wasn't
being produced before.

Closes #714.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -433,6 +433,17 @@ pub struct Connection {
     /// Checked against tparam value from peer
     remote_original_destination_cid: Option<ConnectionId>,
 
+    /// We sometimes save a datagram (just one) against the possibility that keys
+    /// will later become available.
+    /// One example of this is the case where a Handshake packet containing a
+    /// certificate is received coalesced with a 1-RTT packet.  The certificate
+    /// might need to authenticated by the application before the 1-RTT packet can
+    /// be processed.
+    /// The boolean indicates whether processing should be deferred: after saving,
+    /// we usually immediately try to process this, but that won't work until after
+    /// returning to the application.
+    saved_datagram: Option<(Datagram, bool)>,
+
     pub(crate) crypto: Crypto,
     pub(crate) acks: AckTracker,
     idle_timeout: IdleTimeout,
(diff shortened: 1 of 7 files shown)


agrover:
> There is a change in the HTTP code that bothers me a little. There is a
new event being generated. I don't understand why that event wasn't
being produced before.

Which event?

martinthomson:
The event that I was seeing was that "StreamWritable" for stream 0 (the request stream).  For some reason, the code was not expecting that event previously.  But I couldn't see any way that it would not be generated.  I was hoping that @ddragana could explain why that wasn't being seen previously.

agrover:
Just to record here from our talk yesterday, I'm still seeing consistent 1-2 decryption failure messages with quic.rocks:4433, which I wasn't expecting and wanted to understand better.

martinthomson:
OK, I checked quic.rocks and everything looks fine.  Without the patch we drop two packets, with the patch we drop one.

The first packet we drop is the 1-RTT packet that is coalesced with a Handshake packet.  With this change we eventually process that packet.

The second packet we drop is an Initial packet, but we have dropped the Initial keys when that arrives, so we should be dropping that.

I've rebased this and tweaked the logic slightly.

neqo-transport/src/connection.rs
```diff
@@ -5358,6 +5432,106 @@ mod tests {
         assert!(matches!(client.state(), State::Draining { .. }));
     }
 
+    /// Test that a server that coalesces 0.5 RTT with handshake packets
+    /// doesn't cause the client to drop application data.
```
martinthomson, line 5436:
```suggestion
    /// Test that a server can send 0.5 RTT application data.
```

agrover:
:+1: please merge once conflicts resolved.

PR #724, opened 2020-06-15
Remove reference cycle in server

### LINKED ISSUE #523: Refcounting has cycles between CidMgr and Connection?
When running http3-server, I'm never seeing connections dropped (i.e deallocated). I think it's because Connection contains a reference to CidMgr and ServerConnectionIdManager contains a reference to the connection.

martinthomson:
This uses `Weak` rather than `Rc` because the loop was holding
connections open.

In other news, Rust doesn't fix memory leaks arising from thoughtless
use of Rc.

Closes #523.

### DIFF
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -293,7 +293,7 @@ impl Server {
         match out {
             Output::Datagram(_) => {
                 qtrace!([self], "Sending packet, added to waiting connections");
-                self.waiting.push_back(c.clone());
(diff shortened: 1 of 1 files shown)


ddragana:
I think we have a issue for this? or was it another cycle?

martinthomson:
Just #523.

agrover:
With this fixed, it should also be possible to remove the small bit of ugliness at neqo-transport/server.rs line 327? If the connection is dropped, the NeqoQlog file for it should now also be automatically closed.

Issue #683, opened 2020-06-02
Rename enable_ciphers() to set_ciphers()


agrover:
opened as a minor tangent to reviewing #681..

If you don't call `Connection::enable_ciphers()`, it doesn't mean that ciphers aren't enabled (that would not work of course), it just means the default ciphers are chosen and used. So, not a bug but an unclear implication.

A clearer name would be `set_ciphers()` for the methods in both `Connection` and `Agent`.

i-am-jeetu:
Created a PR  #737 :). A tiny contribution.

PR #737, opened 2020-06-16
Renamed enable_ciphers() to set_ciphers()

i-am-jeetu:
Fixes Issue : #683 
Made changes, compiled and ran tests.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -838,7 +838,7 @@ mod old {
         }
 
         if let Some(cip) = ciphers {
-            client.enable_ciphers(cip).expect("Cannot enable ciphers");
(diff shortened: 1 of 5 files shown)


neqo-client/src/main.rs
```diff
@@ -838,7 +838,7 @@ mod old {
         }
 
         if let Some(cip) = ciphers {
-            client.enable_ciphers(cip).expect("Cannot enable ciphers");
+            client.set_ciphers(cip).expect("Cannot enable ciphers");
```
agrover, line 841: Just need to update the expect message too.

agrover:
Thanks for having a look and helping out, @i-am-jeetu !!

Issue #553, opened 2020-04-17
Turn on clippy pedantic for neqo-transport


agrover:
Like #547 did for neqo-http3, and the previous work on crypto and other of our crates.

abbcdfin:
Hi @agrover,

Good day. I was trying to work on this issue, but when I tried to do the clippy check I got this:
[jian@arch-dev neqo]$ cargo clippy -p neqo-transport -- -W clippy::pedantic
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s

However, when I use the same method to check neqo-server, seems it kind of worked.
[jian@arch-dev neqo]$ cargo clippy -p neqo-server -- -W clippy::pedantic
    Checking neqo-server v0.3.1 (/home/jian/openlab/neqo/neqo-server)
error: integer type suffix should be separated by an underscore
...

I am new to rust, would you please let me know the proper way to start the check?

Thanks and Best Regards

i-am-jeetu:
Yes, @abbcdfin is right. Apart from removing `#![deny(clippy::pedantic)]` from neqo-transport, there is nothing else to do.

martinthomson:
`cargo clippy -- -W clippy::pedantic` returns a lot of warnings for me (though I'm on rust 1.44 and so a little ahead of the curve).

Here's a taste:

```
warning: binding's name is too similar to existing binding
    --> neqo-transport/src/connection.rs:5387:13
     |
5387 |         let scid = dec.decode_vec(1).expect("client SCID");
     |             ^^^^
...
error: Unnecessary boolean `not` operation
   --> neqo-transport/src/cc.rs:183:9
    |
183 | /         if !self.in_congestion_recovery(sent_time) {
...
error: item name starts with its containing module's name
   --> neqo-transport/src/crypto.rs:239:1
    |
239 | / pub enum CryptoDxDirection {
...
error: adding items after statements is confusing, since items exist from the start of the scope
   --> neqo-transport/src/crypto.rs:292:9
    |
292 | /         const INITIAL_SALT_27: &[u8] = &[
```

PR #748, opened 2020-06-21
clippy pedantic enabled neqo-transport

i-am-jeetu:
Removed `#![deny(clippy::pedantic)]` from neqo-transport files 
Issue : #553

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -5,7 +5,6 @@
 // except according to those terms.
 
 // Pacer
-#![deny(clippy::pedantic)]
 
 use neqo_common::qtrace;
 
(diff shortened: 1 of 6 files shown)


i-am-jeetu:
@agrover Can you rerun the jobs as build failed during `neqo-crypto/tests/selfencrypt.rs` which was untouched. 
Also, locally builds are succesful using the same commands as CircleCI used, i.e `cargo build -v --all-targets --tests`

martinthomson:
This change has the effect of disabling warnings that were previously enabled.  So it is no surprise that it works.  The goal of the issue was to enable more warnings.

i-am-jeetu:
Oh! My bad. My inexperience with rust exposed :rofl:  
Closing this one. Will submit new by actually solving the issue, instead of doing exactly opposite.

PR #762, opened 2020-06-23
New token

martinthomson:
We don't consume and use NEW_TOKEN frames.  That might be a problem later.  Fix it.

Also, save a few session tickets.


agrover:

PR #766, opened 2020-06-24
Write the header block's prefix properly.

### LINKED ISSUE #459: qpack: Fix Header Block Prefix

ddragana:
Fixes #459

### DIFF
--- a/neqo-qpack/src/header_block.rs
+++ b/neqo-qpack/src/header_block.rs
@@ -16,6 +16,7 @@ use crate::table::HeaderTable;
 use crate::Header;
 use crate::{Error, Res};
 use neqo_common::qtrace;
+use std::mem;
 use std::ops::{Deref, Div};
 
 #[derive(Default, Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


neqo-qpack/src/header_block.rs
```diff
@@ -163,25 +121,41 @@ impl HeaderEncoder {
...
+            enc_insert_cnt,
+            delta,
+            prefix
+        );
+
+        self.buf
+            .encode_prefixed_encoded_int(NO_PREFIX, enc_insert_cnt);
+        self.buf.encode_prefixed_encoded_int(prefix, delta);
+
+        self.buf.write_bytes(&tmp);
```
martinthomson, line 158: Not that this needs fixing, but this is worth identifying as a potential performance improvement.

For instance, we could reserve space for the header and write into that space at the end, then return the appropriate slice.

PR #767, opened 2020-06-25
Actions CI image

martinthomson:
Of course GitHub actions aren't enabled in the mozilla org, so this won't run, but it does appear to function as advertised.  I can trigger this with pushes to my personal repo (you could do the same as well if you configure the necessary secrets).


agrover:

PR #769, opened 2020-06-29
Small changes:

ddragana:
- Set Http3Connection to the ZeroRtt state only after the initialization is done, because the initialization may fail.
- Ignore InvalidStreamId error from the transport when trying to close a stream.
- If decoding settings in set_resumption_token fails map error to InvalidResumptionToken

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -171,10 +171,10 @@ impl Http3Connection {
 
     /// We have a resumption token which remembers previous settings. Update the setting.
     pub fn set_0rtt_settings(&mut self, conn: &mut Connection, settings: HSettings) -> Res<()> {
-        self.state = Http3State::ZeroRtt;
         self.initialize_http3_connection(conn)?;
         self.set_qpack_settings(&settings)?;
         self.settings_state = Http3RemoteSettingsState::ZeroRtt(settings);
+        self.state = Http3State::ZeroRtt;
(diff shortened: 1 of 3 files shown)


neqo-http3/src/connection.rs
```diff
@@ -534,7 +534,9 @@ impl Http3Connection {
             .send_streams
             .get_mut(&stream_id)
             .ok_or(Error::InvalidStreamId)?;
-        send_stream.close(conn)?;
+        // The following funcion may return InvalidStreamId from the transport layer if the stream has been cloesd
+        // already. It is ok to ignore it here.
```
vonasek, line 538: Maybe we might put here an assertion that the error is InvalidStreamId for case that transport layer is changed and can throw other errors as well. But I'll leave it up to you...

ddragana: This is a good idea. I will add this assertion in a follow up.

neqo-http3/src/connection.rs
```diff
@@ -534,7 +534,9 @@ impl Http3Connection {
             .send_streams
             .get_mut(&stream_id)
             .ok_or(Error::InvalidStreamId)?;
-        send_stream.close(conn)?;
+        // The following funcion may return InvalidStreamId from the transport layer if the stream has been cloesd
```
martinthomson, line 537: "closed"

PR #770, opened 2020-06-29
Small fix 2

ddragana:


ddragana:
I want to look into splitting IncrementalDecoder in a separate PR.

neqo-http3/src/hframe.rs
```diff
@@ -105,20 +106,18 @@ impl HFrame {
...
-    GetType,
-    GetLength,
-    GetData,
-    UnknownFrameDischargeData,
-    Done,
+    BeforeFrame { decoder: IncrementalDecoder },
+    GetType { decoder: IncrementalDecoder },
+    GetLength { decoder: IncrementalDecoder },
+    GetData { decoder: IncrementalDecoder },
+    UnknownFrameDischargeData { decoder: IncrementalDecoder },
```
martinthomson, line 115: It seems like this should be a struct that holds a decoder and an enum rather than an enum with the same type attached to every alternative.

neqo-http3/src/hframe.rs
```diff
@@ -314,39 +317,26 @@ impl HFrameReader {
...
-                    _ => return Err(Error::NotEnoughData),
-                },
+                push_id: dec.decode_varint().ok_or(Error::HttpFrame)?,
             },
             H3_FRAME_TYPE_SETTINGS => {
                 let mut settings = HSettings::default();
-                settings.decode_frame_contents(&mut dec)?;
+                settings
+                    .decode_frame_contents(&mut dec)
+                    .map_err(|_| Error::HttpFrame)?;
```
martinthomson, line 326: You can avoid needing the mapping here by implementing From.

neqo-http3/src/hframe.rs
```diff
@@ -465,175 +451,101 @@ mod tests {
...
-        assert!(f.is_ok());
-        if let HFrame::Settings { settings } = f.unwrap() {
+        assert!(fr.process(&[0x40]).is_none());
+        assert!(fr.process(&[0x4]).is_none());
+        assert!(fr.process(&[0x6]).is_none());
+        assert!(fr.process(&[0x6]).is_none());
+        assert!(fr.process(&[0x40]).is_none());
+        assert!(fr.process(&[0x4]).is_none());
+        assert!(fr.process(&[0x8]).is_none());
+        assert!(fr.process(&[0x41]).is_none());
```
martinthomson, line 518:
```rust
for i in &[0x40, 0x04, ...] {
  assert!(fr.process(&[*i]).is_none());
}
```

neqo-http3/src/hframe.rs
```diff
@@ -465,175 +451,101 @@ mod tests {
...
+        if let HFrame::Settings { settings } = frame.unwrap() {
             assert!(settings.len() == 1);
             assert!(settings[0] == HSetting::new(HSettingType::MaxHeaderListSize, 4));
         } else {
             panic!("wrong frame type");
         }
     }
 
-    // Test PUSH_PROMISE
+    // Test receiving bytte by byte for a PUSH_PROMISE frame.
```
martinthomson, line 530:
```suggestion
    // Test receiving byte by byte for a PUSH_PROMISE frame.
```

neqo-http3/src/hframe.rs
```diff
@@ -465,175 +451,101 @@ mod tests {
...
-
-        assert!(fr.done());
-        let f = fr.get_frame();
-        assert!(f.is_ok());
+        assert!(fr.process(&[0x5]).is_none());
+        assert!(fr.process(&[0x5]).is_none());
+        assert!(fr.process(&[0x41]).is_none());
+        assert!(fr.process(&[0x1]).is_none());
+        assert!(fr.process(&[0x1]).is_none());
+        assert!(fr.process(&[0x2]).is_none());
```
martinthomson, line 541: loop

neqo-http3/src/hframe.rs
```diff
@@ -645,35 +557,21 @@ mod tests {
...
 
         // Read data frame 0003010203
-        conn_s
-            .stream_send(stream_id, &[0x0, 0x3, 0x1, 0x2, 0x3])
-            .unwrap();
-        let out = conn_s.process(None, now());
-        let _ = conn_c.process(out.dgram(), now());
-        assert_eq!(Ok(false), fr.receive(&mut conn_c, stream_id));
+        let frame = fr.process(&[0x0, 0x3, 0x1, 0x2, 0x3]).unwrap();
+        if let HFrame::Data { len } = frame {
```
martinthomson, line 564: assert!(matches!(frame, HFrame::Data { len } if len == 3));

neqo-http3/src/connection.rs
```diff
@@ -233,35 +233,24 @@ impl Http3Connection {
...
-            while self.control_stream_remote.frame_reader_done()
-                || self.control_stream_remote.recvd_fin()
-            {
-                if let Some(f) = self.handle_control_frame()? {
-                    control_frames.push(f);
+        assert!(self.control_stream_remote.is_recv_stream(stream_id));
+        let mut control_frames = Vec::new();
+
+        loop {
+            match self.control_stream_remote.receive(conn)? {
```
martinthomson, line 240:
```rust
if let Some(f) = self.control_stream_remote.receive(conn)? {
...
```

neqo-http3/src/connection.rs
```diff
@@ -233,35 +233,24 @@ impl Http3Connection {
...
+                Some(f) => {
+                    if let Some(f) = self.handle_control_frame(f)? {
+                        control_frames.push(f);
+                    }
+                }
+                None => {
+                    if control_frames.is_empty() {
+                        break Ok(HandleReadableOutput::NoOutput);
```
martinthomson, line 248: return is probably clearer than break here.

neqo-http3/src/connection.rs
```diff
@@ -543,28 +537,21 @@ impl Http3Connection {
 
     // If the control stream has received frames MaxPushId or Goaway which handling is specific to
     // the client and server, we must give them to the specific client/server handler..
```
martinthomson, line 539:
```suggestion
    // the client and server, we must give them to the specific client/server handler.
```

neqo-http3/src/hframe.rs
```diff
@@ -186,124 +201,112 @@ impl HFrameReader {
                 }
             };
 
-            let progress = self.decoder.consume(&mut input);
-            match self.state {
-                HFrameReaderState::BeforeFrame | HFrameReaderState::GetType => match progress {
+            match &mut self.state {
```
martinthomson, line 204: Since you are here, can you find a way to split this loop out into a new function?

neqo-http3/src/hframe.rs
ddragana: I want to split IncrementalDecoder into 3 decoders, I did not get to it yesterday. So each enum will hold a different decoder.

martinthomson: I see what you mean.  Yeah, that's a good idea.

neqo-http3/src/hframe.rs
ddragana: Ok, I am going to ask a stupid question: can you explain what you mean?

neqo-common/src/incrdecoder.rs
```diff
@@ -4,162 +4,34 @@
...
-/// You only get an error if the decoder completed its last instruction.
-#[derive(Debug, PartialEq)]
-pub enum IncrementalDecoderResult {
-    InProgress,
-    Uint(u64),
-    Buffer(Vec<u8>),
-    Ignored,
-    Error,
+#[derive(Clone, Debug)]
+pub struct IncrementalDecoderUint {
```
martinthomson, line 19: You can collapse this and put the implementation on the enum directly.

martinthomson: Or, you could make the struct simpler:

```rust
struct Blah {
   v: u64,
   remaining: usize,
}
```

This can have an automatic `Default` implementation.  If remaining is zero, then you are before any number.

neqo-http3/src/hframe.rs
martinthomson: `map_err()` is used to translate `Result<T, A>` to `Result<T, B>`.  However, the `?` operator also does the same automatically.  If the function returns `Result<T, B>` and there is an implementation of `Into<B>` on `A` or `From<A>` on `B`, then the conversion is automatic.

But all that doesn't apply here: because the two types are the same.  I think that this is case where you should just change `decode_frame_contents` to return the right type of error.  Then you can drop `map_err` altogether.

neqo-http3/src/hframe.rs
```diff
@@ -134,176 +134,161 @@ impl HFrameReader {
...
+            if !read {
+                // There was no new data, exit the loop.
+                break Ok((None, false));
+            }
+        }
+    }
+
+    fn consume(&mut self, mut input: Decoder) -> Res<Option<HFrame>> {
+        match &mut self.state {
+            HFrameReaderState::BeforeFrame { decoder } | HFrameReaderState::GetType { decoder } => {
```
martinthomson, line 208: Do you really need a separate state for "BeforeFrame"?

neqo-common/src/incrdecoder.rs
ddragana: remaining==0 also mean the end of decoding. I will put it as an Option

neqo-http3/src/hframe.rs
ddragana: It is used for one thing to decide if an end-of-stream is an error or not: in the middle of a frame it is an error otherwise it is not. I can indicate that in a different way as well.

neqo-common/src/incrdecoder.rs
martinthomson: But remaining == 0 implies that you have already received the next value.  Isn't that OK?  Anyone using this will take the Option::Some and use it.  With remaining == 0 then they can ask for the next number without making a new object (assuming that you set v = 0 on return, that is).

neqo-http3/src/hframe.rs
ddragana: That was the reason I was asking. I wanted to know if I miss some functionality of From (I do not know what I do not know). 
I will do this in a separate PR. I want to be sure that i am not breaking something.

neqo-common/src/incrdecoder.rs
ddragana: All of IncrementalDecoder* are meant to be used only once. I think I will leave it as Option<>

martinthomson: That being the case, you can use `Option<NonZeroUsize>` :)

PR #775, opened 2020-07-01
Try using lld in CI

martinthomson:
lld is better in some ways that GNU ld, so maybe this will fix our
issues with crashes in CI.

I don't really have any means of testing this out other than trying it out, but it can't be much worse than what we're suffering through.  With your permission, I'll try poking at this.


agrover:
Sure, yes please. Since this is intermittent I don't know a way to see if this makes a difference other than applying it and see if the OOM CI failures diminish.

Issue #753, opened 2020-06-22
Check pseudo-header fields


ddragana:
produce a stream error H3_GENERAL_PROTOCOL_ERROR if header block is malformed (section [4.1.3](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-malformed-requests-and-resp)):
- the presence of prohibited fields or pseudo-header fields
- the absence of mandatory pseudo-header fields
- invalid values for pseudo-header fields
- pseudo-header fields after fields
- an invalid sequence of HTTP messages
- the inclusion of uppercase field names
- the inclusion of invalid characters in field names or values

i-am-jeetu:
I was tackling this. Have a look at the draft below, and see if the checks are performed at the right place. Currently, only the inclusion of uppercase characters in field names are handled. #780

ddragana:
Note: some of this checks are perform in the firefox platform code, but they should be in neqo.

PR #780, opened 2020-07-01
Uppercase in header field

i-am-jeetu:
Draft for #753.

### DIFF
--- a/neqo-http3/src/recv_message.rs
+++ b/neqo-http3/src/recv_message.rs
@@ -130,9 +130,58 @@ impl RecvMessage {
         }
     }
 
+    #[allow(unreachable_code)]
     fn check_headers(headers: &[Header]) -> bool {
+        let mut regular_header_field_encountered = false;
+
+        let prohibited_header_list = vec![
+            "accept-charset",
+            "accept-encoding",
+            "access-control-request-headers",
+            "access-control-request-method",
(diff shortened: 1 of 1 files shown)


ddragana:
Looks good.

martinthomson:
Yep, but it needs some tests to demonstrate that it works and to ensure that we don't accidentally break it.

i-am-jeetu:
>produce a stream error H3_GENERAL_PROTOCOL_ERROR if header block is malformed 
> - the presence of prohibited fields or pseudo-header fields

Where can I find the list of prohibited header field names? It isn't this one right? [MDN](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name)

ddragana:
> > produce a stream error H3_GENERAL_PROTOCOL_ERROR if header block is malformed
> > 
> > * the presence of prohibited fields or pseudo-header fields
> 
> Where can I find the list of prohibited header field names? It isn't this one right? [MDN](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name)

this is about pseudo-headers fields. They are define in section "4.1.1.1.  Pseudo-Header Fields" in https://tools.ietf.org/html/draft-ietf-quic-http-29

agrover:

Issue #795, opened 2020-07-06
We get PROTOCOL_VIOLATION (0xA) error from google


ddragana:
First diagnostic of the problem can be found here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1630851#c11

There are 2 problems:
1) we get PROTOCOL_VIOLATION (0xA) error from a peer
2) necko never finds out about the error. Necko should receive a ConnectionClosing or ConnecttionClosed event. This may be a problem in necko as well.

ddragana:
PROTOCOL_VIOLATION (0xA) is necko problem :) I will fix it. It is mismatching alpn.

ddragana:
Fixed by #796

PR #796, opened 2020-07-06
Draining is also a closing state.

ddragana:


ddragana:
This will fix problem the second problem described in #795 
Fixes #795

PR #800, opened 2020-07-08
Document errors for neqo-transport::Connection::stream_recv

ddragana:


neqo-transport/src/connection.rs
```diff
@@ -2547,6 +2547,9 @@ impl Connection {
 
     /// Read buffered data from stream. bool says whether read bytes includes
     /// the final data on stream.
+    /// # Errors
+    /// `InvalidStreamId` if streeam does not exist.
```
martinthomson, line 2551:
```suggestion
    /// `InvalidStreamId` if the stream does not exist.
```

neqo-transport/src/connection.rs
```diff
@@ -2547,6 +2547,9 @@ impl Connection {
...
+    /// `NoMoreData` if data nd fin bit are already read by the application.
```
martinthomson, line 2552:
```suggestion
    /// `NoMoreData` if data and fin bit were previously read by the application.
```

neqo-transport/src/recv_stream.rs
```diff
@@ -514,6 +514,8 @@ impl RecvStream {
             .map_or(false, RxStreamOrderer::data_ready)
     }
 
+    /// # Errors
+    /// `NoMoreData` if data nd fin bit aree already read by the application.
```
martinthomson, line 518:
```suggestion
    /// `NoMoreData` if data and fin bit were previously read by the application.
```

PR #803, opened 2020-07-08
Handle properly errors from set_resumption_token

ddragana:


neqo-transport/src/connection.rs
```diff
@@ -821,6 +830,11 @@ impl Connection {
                 | State::Closed(err) => {
                     qwarn!([self], "Closing again after error {:?}", err);
                 }
+                State::Init => {
+                    // We have not even sent anything just close the connection without sending any error.
+                    // This may happen when clieeent_start fails.
```
agrover, line 835: Darn Mac keyboard.

Issue #799, opened 2020-07-07
QNS/client: Fix issues with handshakecorruption (C1) and handshakeloss (L1) tests


agrover:
https://interop.seemann.io/

We are periodically failing these tests against multiple server implementations.

@martinthomson says we may: `need to keep the PTO timer armed even after discarding Initial keys.`

agrover:
We now have per-test-failure issues so we can close this more general issue.

PR #822, opened 2020-07-15
Handshake recovery fixes

martinthomson:
There are two main changes here:

1. Keep the PTO timer armed for the Initial and Handshake packet number spaces.

This is based on the spec, which says that clients need to keep these armed, even if they haven't sent anything that is "in flight".  Doing this avoids a handshake deadlock when the server hasn't got its own PTO timer armed.  The server can't arm the PTO timer when it hasn't validated the client's address.

Note that we don't implement the anti-amplification limit, so our server doesn't really do this.  It arms the PTO timer when it shouldn't and ignores limits.  This is something we might need to fix eventually.

The other limitation here is that the PTO is armed past the point where it should be.  The spec says that the client can stop after the server has acknowledged Handshake or 1-RTT packets; or after the handshake is confirmed.  Not stopping means that there are some crazy corner cases where we'll keep probing in these spaces, but the impact should be limited.

(It would be good if you could also go through these corner cases and see if you could find one that is worth caring about.  I couldn't.)

2. Resend CRYPTO frames in Initial and Handshake packet number spaces if a useless CRYPTO frame is received (i.e., one that doesn't make more data available).  Also, if a PING is received in these spaces, do the same.

The CRYPTO frame thing is close to what is recommended in the spec, noting that this doesn't strictly adhere to the spec.  The spec says that only CRYPTO frames containing redundant data cause retransmission; this change causes any CRYPTO frame that doesn't make new data available.  This means that out-of-order CRYPTO frames will cause us to send more CRYPTO frames.  This could be a little wasteful. 

Assuming that we do this, it's probably worth opening an issue to look into making this a little less aggressive about re-sending.  As it is, this might eat a little of our congestion window if we get an out-of-order packet AND we don't receive an ACK for outstanding CRYPTO data (which is likely, as ACKs tend to be put with the first CRYPTO frames).  Fixing this might help with wastage, but then it might not depending on how often apparent out-of-order packets are a consequence of loss - re-sending immediately is good if the out-of-order packet really is an indication that the other data was lost.

The PING frame is another digression from spec, but it recognizes that a PING, sent on PTO in Handshake packets, is equivalent to receiving useless CRYPTO data.  If a PING is sent in Handshake or Initial packets, it means that the other side is probing (and likely can't send CRYPTO). Interpreting that as equivalent to a useless CRYPTO frame is safe and likely a good optimization for the case where server Handshake packets are lost.

---
I was planning to split this into two, and I could still do so, but the test that I wrote for the first bit didn't properly complete the handshake until I did the second part.  The problem there was that we were sending PING frames at the client when the Handshake PTO expired.  This elicited an ACK from the server, but the server didn't re-send CRYPTO frames.  The result was that we would have stalled until the server hit a PTO.  So I gave you both fixes.  Each is fairly distinct in the code.

Read the second commit for commentary about the tricky test.

There are some changes to logging that I found useful here as well, but none of those should really interfere.

### DIFF
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -235,6 +235,12 @@ impl Crypto {
         self.streams.lost(token);
     }
 
+    /// Mark any outstanding frames in the indicated space as "lost" so
+    /// that they can be sent again.
+    pub fn resend_frames(&mut self, space: PNSpace) {
+        self.streams.resend(space);
+    }
+
     /// Discard state for a packet number space and return true
     /// if something was discarded.
     pub fn discard(&mut self, space: PNSpace) -> bool {
@@ -996,13 +1002,23 @@ impl CryptoStreams {
         self.get_mut(token.space)
             .unwrap()
             .tx
-            .mark_as_acked(token.offset, token.length)
+            .mark_as_acked(token.offset, token.length);
     }
 
     pub fn lost(&mut self, token: &CryptoRecoveryToken) {
         // See BZ 1624800, ignore lost packets in spaces we've dropped keys
         if let Some(cs) = self.get_mut(token.space) {
-            cs.tx.mark_as_lost(token.offset, token.length)
+            cs.tx.mark_as_lost(token.offset, token.length);
+        }
+    }
+
+    /// Resend any Initial or Handshake CRYPTO frames that might be outstanding.
+    /// This can help speed up handshake times.
+    pub fn resend(&mut self, space: PNSpace) {
+        if space != PNSpace::ApplicationData {
+            if let Some(cs) = self.get_mut(space) {
+                cs.tx.unmark_sent();
+            }
         }
     }
 
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -237,8 +237,8 @@ impl Crypto {
 
     /// Mark any outstanding frames in the indicated space as "lost" so
     /// that they can be sent again.
-    pub fn resend_frames(&mut self, space: PNSpace) {
-        self.streams.resend(space);
+    pub fn resend_unacked(&mut self, space: PNSpace) {
+        self.streams.resend_unacked(space);
     }
 
     /// Discard state for a packet number space and return true
@@ -1014,7 +1014,7 @@ impl CryptoStreams {
 
     /// Resend any Initial or Handshake CRYPTO frames that might be outstanding.
     /// This can help speed up handshake times.
-    pub fn resend(&mut self, space: PNSpace) {
+    pub fn resend_unacked(&mut self, space: PNSpace) {
         if space != PNSpace::ApplicationData {
             if let Some(cs) = self.get_mut(space) {
                 cs.tx.unmark_sent();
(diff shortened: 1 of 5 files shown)


martinthomson:
See #799.  This implements the proposed fix mentioned there.  But I'm not certain that it will fix it completely, so I won't mark this as closing that issue.

agrover:
I've verified that this passes L1 and C1 against quic-go consistently now (passed 5/5 times), and does better against quant and picoquic, although still not passing against msquic and quiche. So yeah can't close #799 yet but making forward progress.

neqo-transport/src/send_stream.rs
```diff
@@ -271,6 +271,19 @@ impl RangeTracker {
...
+        let mut to_remove = SmallVec::<[_; 8]>::new();
+        for (off, (_, state)) in self.used.iter() {
+            if *state != RangeState::Acked {
+                to_remove.push(*off);
+            }
+        }
+        for r in to_remove {
+            self.used.remove(&r);
+        }
+    }
```
agrover, line 286: We already have this guy's big brother in `unmark_range()`. Instead of this I think you can call `self.ranges.unmark_range(0, self.ranges.highest_offset())` in `TxBuffer::unmark_sent()`, perhaps also turning down the qinfos in `unmark_range` to qdebug or qtrace.

neqo-transport/src/crypto.rs
```diff
@@ -996,13 +1002,23 @@ impl CryptoStreams {
         self.get_mut(token.space)
             .unwrap()
             .tx
-            .mark_as_acked(token.offset, token.length)
+            .mark_as_acked(token.offset, token.length);
```
agrover, line 1005: o...kay?

neqo-transport/src/crypto.rs
```diff
@@ -235,6 +235,12 @@ impl Crypto {
         self.streams.lost(token);
     }
 
+    /// Mark any outstanding frames in the indicated space as "lost" so
+    /// that they can be sent again.
+    pub fn resend_frames(&mut self, space: PNSpace) {
+        self.streams.resend(space);
+    }
```
agrover, line 242: "frames" doesn't seem right. `resend_unacked`?

neqo-transport/src/crypto.rs
```diff
@@ -235,6 +235,12 @@ impl Crypto {
         self.streams.lost(token);
     }
 
+    /// Mark any outstanding frames in the indicated space as "lost" so
```
agrover, line 238: outstanding frames -> unacked ranges?

neqo-transport/src/crypto.rs
```diff
@@ -996,13 +1002,23 @@ impl CryptoStreams {
...
         // See BZ 1624800, ignore lost packets in spaces we've dropped keys
         if let Some(cs) = self.get_mut(token.space) {
-            cs.tx.mark_as_lost(token.offset, token.length)
+            cs.tx.mark_as_lost(token.offset, token.length);
+        }
+    }
+
+    /// Resend any Initial or Handshake CRYPTO frames that might be outstanding.
+    /// This can help speed up handshake times.
+    pub fn resend(&mut self, space: PNSpace) {
```
agrover, line 1017: Like above, seems clearer to refer to unacked ranges than outstanding frames. Can rename this method to the same as the new method above.

PR #833, opened 2020-07-20
Rearm PTO timer after confirming

### LINKED ISSUE #828: QNS: handshakeloss (L1) failure with aioquic
[Complete QNS logs](https://interop.seemann.io/logs_2020-07-16T21:22:00UTC/aioquic_neqo/handshakeloss/).

Here's an excerpt of the connection that failed: [gist](https://gist.github.com/agrover/bd946c762a415a7e6ea004db3fa50566)

From looking at the server logs it never receives the http request on stream 0.

Analysis:
Neqo never receives a 1RTT ACK, either of the packet including the stream 0 request ([here ap-4](https://gist.github.com/agrover/bd946c762a415a7e6ea004db3fa50566#file-gistfile1-txt-L347)) that was dropped or for ap-5, ap-7, or ap-8. Server sends [HandshakeDone](https://gist.github.com/agrover/bd946c762a415a7e6ea004db3fa50566#file-gistfile1-txt-L483) and it looks like this disables PTO when it should only disable PTO for hs, not ap.

Proposed solution: Neqo should not disable the 1RTT PTO timer until it receives an 1RTT Ack. This would have caused Neqo to continue to 1RTT Ping the server, which would have generated an 1RTT ack indicating ap-4 was lost and this would cause Neqo to resend the http request data.

Also weird: Why all the "security library: received bad data" messages towards the end, and is this the sign of another issue?

@martinthomson does the above look right? Thanks.

martinthomson:
Don't fire the PTO timer for ApplicationData prior to confirming the
handshake.  However, ignoring this means that when we confirm the
handshake, if the PTO timer for ApplicationData should have fired, it
might need to fire immediately.  That means we need to recalculate the
timer when confirming.

This doesn't mark anything lost from ApplicationData when the PTO fires
to keep the amount of state transmission required minimal.  So we might
send new data, or only send a PING, if the PTO has expired when
confirming the handshake.  That seems acceptable; many other stacks work
that way.

Closes #828.  Maybe.  I haven't run the full analysis of that.  This is
only based on this being the correct behaviour, per-spec.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -1049,10 +1049,10 @@ impl Connection {
         Ok(())
     }
 
-    fn discard_keys(&mut self, space: PNSpace) {
+    fn discard_keys(&mut self, space: PNSpace, now: Instant) {
         if self.crypto.discard(space) {
             qinfo!([self], "Drop packet number space {}", space);
-            self.loss_recovery.discard(space);
+            self.loss_recovery.discard(space, now);
             self.acks.drop_space(space);
         }
     }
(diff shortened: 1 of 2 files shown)


martinthomson:
OK, I can now confirm that we have another issue in #828, but I can't unmark this as fixing that, so I will open a more specific issue for the other piece.

agrover:
closing, merged as part of #838.

PR #834, opened 2020-07-20
 Don't cap ACK Delay prior to confirmation

martinthomson:
This does two things:

1. Better caching of undecryptable packets.  Multiple packets are saved against the keys that we don't yet have.  Once those keys are available, those packets are restored.

As part of this the time that the packet was received is also saved.  That is the time we use when processing that packet later.

It took me a while to confirm that processing packets with a time that is potentially in the past relative to other actions we've taken is safe, but I think that it's OK to do that.  We don't assume that the arrow of time flows in one direction very much.  There are a few places where the results might be weird, but what I found was that the code is fairly robust against this sort of thing already.

That analysis isn't 100%.  Time gets everywhere, so there is some risk to this change.  The tests I've added mitigate against that well, but I don't think it is possible to cover every potential outcome here.  This is where better fuzzing (with the new simulator stuff) is likely necessary.  (There is a hole there for 0-RTT that probably needs plugging up.)

2. Removing the max_ack_delay cap on the ACK Delay field prior to confirmation.  This was the original goal of the change, but I realized that I needed some way to test this and saved packets was the only way.  The tests I've added save packets and delay their processing for considerable time, which would previously have resulted in a skewed RTT estimate as the max_ack_delay would have limited what is in the ACK Delay field.


martinthomson:
I've added more thorough caching for datagrams to this.  The last thing left is to add a test for caching of 1-RTT packets that arrive ahead of the last client Handshake packet.  That shouldn't take long, but the meeting starts soon.

agrover:

PR #850, opened 2020-07-23
Simulator

martinthomson:
This is a rough framework for some simulation.

Run this with `RUST_LOG=network=info` to get it to report how much (simulated) time elapsed.

I haven't set up any `#[bench]` tests, but the outcomes are entirely reproducible, so you shouldn't see much variation in execution time either.  I have noticed a huge overhead to debug logging, but some of the simple tests still take far longer than I would like to execute.


martinthomson:
The following test fails:

```
SIMULATION_SEED=270142275defc3966028e6305a9aa9939679f78d36832140b2ebfef06d87aa29 cargo test -p neqo-transport --test network -- transfer_delay_drop --exact --nocapture
```

Might be worth investigating.

ddragana:
Over all this looks good. I am too tired for a detailed review.
Thank you for putting this together.

neqo-transport/tests/sim/drop.rs
```diff
@@ -0,0 +1,70 @@
...
+    threshold: u64,
+    max: u64,
+    rng: Option<Rng>,
+}
+
+impl Drop {
+    /// Make a new random drop generator.  Each `drop` is called, this generates a
+    /// random value between 0 and `max` (exclusive).  If this value is less than
+    /// `threshold` a value of `true` is returned.
+    pub fn new(threshold: u64, max: u64) -> Self {
```
ddragana, line 26: nit: do this need to be public?

neqo-transport/tests/network.rs
```diff
@@ -69,6 +70,28 @@ simulate!(
     ],
 );
 
+simulate!(
+    connect_tail_drop_jitter,
+    [
+        ConnectionNode::new_client(boxed![ReachState::new(State::Confirmed)]),
+        TailDrop::dsl_uplink(),
+        Delay::new(ZERO..JITTER),
```
ddragana, line 100: is this going to add reordering? I need o check...

neqo-transport/tests/sim/delay.rs
```diff
@@ -0,0 +1,98 @@
...
+impl Delay {
+    pub fn new(bounds: Range<Duration>) -> Self {
+        Self {
+            random: RandomDelay::new(bounds),
+            queue: BTreeMap::default(),
+        }
+    }
+
+    fn insert(&mut self, d: Datagram, now: Instant) {
+        let mut t = now + self.random.next();
```
ddragana, line 65: This will introduce reordering. We need to have a delay, jitters that id not reordering packets as well.

neqo-transport/tests/sim/connection.rs
```diff
@@ -51,28 +64,44 @@ impl ConnectionNode {
...
+                        self.goals.remove(i);
+                    } else {
+                        active |= status == GoalStatus::Active;
+                        i += 1;
+                    }
                 }
             }
+            // Exit at this point if the connection produced a datagram.
+            // OR if one of the goals acted.
+            if matches!(res, Output::Datagram(_)) || !active {
```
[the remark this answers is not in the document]
martinthomson: No, I think that the comment explains this:

So if we get a datagram, that is easy.  We have to stop.

But we want to make sure that if any of the goals report activity and we don't have any datagram, that might have changed the connection state so that we will get a datagram (or at least we might get a different timeout).

I will amend the comment (and correct a logic error).

neqo-transport/tests/sim/drop.rs
martinthomson: It's not used right now, but the `percentage` function isn't really that good, as it doesn't allow for fractional percentages (for instance).

neqo-transport/tests/network.rs
martinthomson: Yes, this adds reordering.  As you say, we probably do want a random delay without reordering (and probably a queue where small packets can jump the queue).  We can add that later.

This reordering already turned out to be useful in uncovering a bug.

martinthomson:
Good point about having multiple nodes with the same timeout.  That is something I considered, but I don't think that it is worth worrying about.  We'll hit the first in sequence, which isn't great, but it is at least reproducible.

ddragana:
@martinthomson can we land this so that more of us can contribute and add tests?

martinthomson:
Sure.  All it needs is approval.

PR #855, opened 2020-07-23
Fix all the lints from clippy in 1.45

martinthomson:


martinthomson:
Making a PR in case the new overrides are bad in older versions.

PR #862, opened 2020-07-27
Faster ACK processing

martinthomson:
This improves performance by an absurd amount.  Transferring 100M with this implementation of ACK processing takes 2.3s (approximately) of wall clock time on my old machine.  Before the change it took 15.7s.

The flamegraphs show this clearly as well.

Before:
image

After:
image

Clearly we have a way to go before our performance is *good*, but this should help a lot.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -2967,6 +2967,7 @@ mod tests {
             let output = a.process(input, now).dgram();
             assert!(had_input || output.is_some());
             input = output;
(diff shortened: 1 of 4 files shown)


martinthomson:
This all depends on the simulator changes so that it could be tested.  Once I have worked out what the bug is with the extended fixed_seed test I will remove the simulator commits from this branch.

As I look into that, I've noticed that we likely have some work to do to optimize recv_stream.rs also.

PR #867, opened 2020-07-28
More complete testing of RxStreamOrderer

### LINKED ISSUE #866: Loss of received bytes when overlapping at the end
If we receive new bytes that overlap the end of existing bytes, we were dropping some of those existing bytes if the length of the newly received bytes was less than the existing bytes.  Worse, we could have crashed if the truly new bytes were shorter than the overlap.

martinthomson:
I discovered a bug in testing that was hard to reproduce otherwise.

The existing tests that would have caught this used a buffer of the same
length as the one that it overlapped with, hiding the error.

Rather than try to fix just that test case, I added a bunch of new
tests.  This probably overdoes it by quite a bit, but I would rather
have that than miss an error.

Closes #866.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -79,7 +79,7 @@ impl RxStreamOrderer {
                     // (In-order frames will take this path, with no overlap)
                     let overlap = prev_end.saturating_sub(new_start);
                     if overlap != 0 {
-                        let truncate_to = new_data.len() - overlap as usize;
+                        let truncate_to = prev_vec.len() - overlap as usize;
                         prev_vec.truncate(truncate_to)
                     }
                     qtrace!(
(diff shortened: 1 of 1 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -556,6 +556,131 @@ impl RecvStream {
...
+
+    /// An overlap with no new bytes.
+    #[test]
+    fn recv_overlap_duplicate() {
+        recv_ranges(&[10..20, 11..12, 0..10], 20);
+        recv_ranges(&[10..20, 10..15, 0..10], 20);
+        recv_ranges(&[10..20, 14..20, 0..10], 20);
+        // Now with the first range split.
+        recv_ranges(&[10..14, 14..20, 10..15, 0..10], 20);
+    }
```
ddragana, line 683: Let overdo it a bit more:
can you add a test that overlaps 3 ranges:
10..15, 16..20, 21..25, 10..25 ?

martinthomson: Happy to overdo things more :)

PR #875, opened 2020-07-29
Accurate byte counting

martinthomson:
This is slightly different than the specification currently requires,
but it is marginally better, I think.

### DIFF
--- a/neqo-transport/src/cc.rs
+++ b/neqo-transport/src/cc.rs
@@ -6,7 +6,7 @@
 
 // Congestion control
 
-use std::cmp::max;
+use std::cmp::{max, min};
 use std::fmt::{self, Display};
 use std::time::{Duration, Instant};
 
@@ -31,6 +31,7 @@ const PERSISTENT_CONG_THRESH: u32 = 3;
 pub struct CongestionControl {
     congestion_window: usize, // = kInitialWindow
     bytes_in_flight: usize,
+    acked_bytes: usize,
     congestion_recovery_start_time: Option<Instant>,
     ssthresh: usize,
     pacer: Option<Pacer>,
@@ -45,6 +46,7 @@ impl Default for CongestionControl {
         Self {
             congestion_window: INITIAL_WINDOW,
             bytes_in_flight: 0,
+            acked_bytes: 0,
             congestion_recovery_start_time: None,
             ssthresh: std::usize::MAX,
             pacer: None,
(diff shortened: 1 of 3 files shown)


martinthomson:
One consequence of this is that we generate fewer packets that are less than a full datagram in size.

neqo-transport/src/cc.rs
```diff
@@ -243,6 +254,7 @@ impl CongestionControl {
         if !self.in_congestion_recovery(sent_time, PacketState::Lost) {
             self.congestion_recovery_start_time = Some(now);
             self.congestion_window /= 2; // kLossReductionFactor = 0.5
+            self.acked_bytes /= 2;
```
ddragana, line 257: I think this should be set to congestion_window/2 (i.e. the new cwnd)

martinthomson: I don't think so.  acked_bytes is the number of bytes that we haven't deposited into the congestion window.  

In slow start, this value will always be zero (because we always transfer from acked_bytes to congestion_window), so setting a non-zero value will result in an increase to the congestion window.

In congestion avoidance, this value will be between 0 and the current congestion window (that is, once it goes above the congestion window, we drop the value down and add MAX_DATAGRAM_SIZE to the congestion window).

This is just scaling the `0..congestion_window` value to `0..(congestion_window/2)` to match the new value for congestion_window.

neqo-transport/src/cc.rs
```diff
@@ -117,31 +119,40 @@ impl CongestionControl {
...
-                    QlogMetric::BytesInFlight(self.bytes_in_flight),
-                ],
+                &mut self.qlog_curr_cong_state,
+                CongestionState::SlowStart,
+            );
+        }
+        // Congestion avoidance, above the slow start threshold.
+        if self.acked_bytes >= self.congestion_window {
+            self.congestion_window += MAX_DATAGRAM_SIZE;
+            self.acked_bytes -= MAX_DATAGRAM_SIZE;
```
ddragana, line 141: acked_bytes are increase on every packet by the size of the packet. These lines mean hat we are going to double cwnd on each RTT.
I think this should be set to 0.

neqo-transport/src/cc.rs
ddragana: Thinking about it a bit this should be sett to 0. we should start counting from 0 again.

neqo-transport/src/cc.rs
martinthomson: That's a good catch.  This should be self.acked_bytes -= congestion_window (so we keep any overage) and should be reordered with the line above.

martinthomson: I've added a test that catches the bug (and fixed the bug).

ddragana:
This looks good. Please set acked_bytes to 0 when we decrease cwnd (see comment above).  We need to start counting acked bytes from 0 after recovery period is done.

martinthomson:
I don't agree.  For reference, see the [pull request on the spec repo](https://github.com/quicwg/base-drafts/pull/3917/files), which does exactly as described.

The way I think of this is that the residual value in acked_bytes represents credit obtained by successful exchange of packets beyond the current congestion window.  Erasing that credit increases the penalty for a congestion event to more than 2x.  The division keeps that cost the same.  The previous algorithm always banked credits it got, so this is equivalent in that aspect.  Well, other than to note that constant banking of credit has the effect of reducing the value of future credits, reducing increases to less than 1MSS per round trip as intended, which is why this fix is worth doing.

ddragana:
That is reasonable, anyway this will improve performance for a bit, but not really much.

martinthomson:
Mergify saves the day.  Dragana, would you mind taking another look at the test in cc.rs?

neqo-transport/src/cc.rs
```diff
@@ -126,31 +128,40 @@ impl CongestionControl {
...
-            qlog::metrics_updated(
+            self.acked_bytes += pkt.size;
+        }
+        qtrace!([self], "ACK received, acked_bytes = {}", self.acked_bytes);
+
+        // Slow start, up to the slow start threshold.
+        if self.congestion_window < self.ssthresh {
+            let increase = min(self.ssthresh - self.congestion_window, self.acked_bytes);
+            self.congestion_window += increase;
+            self.acked_bytes -= increase;
```
junhochoi, line 139: I think no need to update acked_bytes during slow start (it applies only to congestion avoidance)

neqo-transport/src/cc.rs
```diff
@@ -126,31 +128,40 @@ impl CongestionControl {
...
                 &mut self.qlog,
-                &[
-                    QlogMetric::CongestionWindow(self.congestion_window),
-                    QlogMetric::BytesInFlight(self.bytes_in_flight),
-                ],
+                &mut self.qlog_curr_cong_state,
+                CongestionState::SlowStart,
             );
         }
+        // Congestion avoidance, above the slow start threshold.
```
junhochoi, line 147: We need `else` here:
```
if self.congestion_window < self.ssthresh {
...
} else { --> this is missing
...
```

neqo-transport/src/cc.rs
junhochoi: Also need to set acked_bytes = 0 when persistent congestion event.

neqo-transport/src/cc.rs
martinthomson: I'm using it as a scratch space/temporary variable in this case.  Also, if we overshoot ssthresh, any extra will stay around.

neqo-transport/src/cc.rs
martinthomson: That too was intentional.  If we hit ssthresh, the bytes up to that limit will count 1:1, the bytes after will be counted toward the next max_datagram_size increase.

agrover:

PR #884, opened 2020-07-31
1-RTT reordering test

martinthomson:
This was originally intended to be part of #834, but I ran into a [bug in NSS](https://bugzilla.mozilla.org/show_bug.cgi?id=1656429), so this will have to wait a bit.  So this looks like a lot of code, but really it's only the last commit.

This tests that reordering of the last handshake packet with 1-RTT is correctly saved AND the RTT estimate is correct.  It does that with 0-RTT, which is different, because 0-RTT allows the server to send an ACK frame.  This results in the saved packets generating RTT samples, which is important to test.  This works, but only once NSS is fixed..


agrover:

PR #886, opened 2020-07-31
Set granularity of the server timer to 1ms.

ddragana:
The default granularity is 100ms.

### DIFF
--- a/neqo-http3-server/src/main.rs
+++ b/neqo-http3-server/src/main.rs
@@ -372,7 +372,9 @@ fn main() -> Result<(), io::Error> {
 
     let (poll, mut sockets, mut servers) = init_poll(&hosts, &args)?;
 
-    let mut timer = Builder::default().build::<usize>();
+    let mut timer = Builder::default()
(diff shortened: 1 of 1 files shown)


martinthomson:
If we are going to do this, then we need to tell the stack about the timer granularity as well.

agrover:
The stack meaning neqo lower layers? I'm not clear why, can you explain a little more?

ddragana:
> If we are going to do this, then we need to tell the stack about the timer granularity as well.

The Timer that we use has a default granularity of 100ms which is bad. if we set timer to 199ms it will fire after 100ms and not fire again. Therefore server was actually relying on getting packets to trigger pacing timer. 
Wee need  to check OS level timer to actually get correct granularity. that would be a different PR.

martinthomson:
> if we set timer to 199ms it will fire after 100ms and not fire again

This is odd.  My understanding is that timers fire late and never early.  If we're getting early firing, then we'll wake up and do nothing, which I guess is OK, but if it happens all the time we'll be wasting a lot of effort doing nothing all the time.

PR #888, opened 2020-08-03
Write datagrams inot socket immediately.

ddragana:


ddragana:
This change reduces run time for local testing from 38s to 22s.

PR #891, opened 2020-08-04
Better caching of undecryptable packets

martinthomson:
Multiple packets are saved against the keys that we don't yet have. Once those keys are available, those packets are restored.

In support of this, a new concept is added: a crypto space, which indicates which keys are in use and corresponds to a packet type.  This made most of the crypto code a little simpler, as keying based on packet number space was likely wrong.  This is particularly true for reading packets, where we might previously have tried to read 0-RTT packets with 1-RTT keys and vice versa (I've no evidence of this happening, but it seems entirely possible with some reordering).

As part of this the time that the packet was received is also saved. That is the time we use when processing that packet later.  This is important because it stops strangeness when it comes to calculating RTT.

It took me a while to confirm that processing packets with a time that is potentially in the past relative to other actions we've taken is safe, but I think that it's OK to do that. We don't assume that the arrow of time flows in one direction very much. There are a few places where the results might be weird, but what I found was that the code is fairly robust against this sort of thing already.

That analysis isn't 100%. Time gets everywhere, so there is some risk to this change. The tests I've added mitigate against that well, but I don't think it is possible to cover every potential outcome here. This is where better fuzzing (with the new simulator stuff) is likely necessary. (There is a hole there for 0-RTT that needs plugging up, but I found a bug in NSS that will need to be fixed before I can land that test.)


agrover:
My current take is that this is cool but I'm worried the benefit -- less retransmits on links with reordering is that right? -- does not outweigh the added complexity.

neqo-transport/src/connection.rs
```diff
@@ -394,6 +397,44 @@ impl StateSignaling {
     }
 }
 
+struct SavedDatagram {
```
ddragana, line 407: Can we move this into a separate file?

ddragana (approved):

martinthomson:
I split a bunch of things from connection.rs.  I won't do any more to save on conflicts from other pull requests, but now seems like a good time to consider that.

martinthomson:
Closing in favour of the other, more complete, PR.

PR #894, opened 2020-08-04
1rtt reordering test

martinthomson:
This was originally intended to be part of #891, but I ran into a [bug in NSS](https://bugzilla.mozilla.org/show_bug.cgi?id=1656429), so this will have to wait a bit.  So this looks like a lot of code, but really it's only the last commit.

This tests that reordering of the last handshake packet with 1-RTT is correctly saved AND the RTT estimate is correct.  It does that with 0-RTT, which is different, because 0-RTT allows the server to send an ACK frame.  This results in the saved packets generating RTT samples, which is important to test.  This works, but only once NSS is fixed.


ddragana:
This looks good.

Can we rebase it and land?

martinthomson:
Happy to land the whole thing; it doesn't appear to need rebasing right now.

ddragana (approved):

PR #898, opened 2020-08-04
Send Qpack's StreamCancelled instruction

ddragana:


neqo-qpack/src/decoder.rs
```diff
@@ -152,6 +152,7 @@ impl QPackDecoder {
 
     pub fn cancel_stream(&mut self, stream_id: u64) {
         if self.table.capacity() > 0 {
+            self.blocked_streams.retain(|(id, _)| *id == stream_id);
```
agrover, line 155: Is this backwards? I think we want to retain when id != stream_id.

neqo-http3/src/recv_message.rs
```diff
@@ -323,10 +323,27 @@ impl RecvStream for RecvMessage {
...
+        {
+            decoder.cancel_stream(self.stream_id);
+        }
         self.conn_events.reset(self.stream_id, app_error);
     }
 
+    fn stream_reset(&self, decoder: &mut QPackDecoder) {
+        if !matches!(
+            self.state,
+            RecvMessageState::ClosePending | RecvMessageState::Closed
```
agrover, line 340: This `matches` is used a few times, would putting it in a little helper function named `closed()`(?) on `RecvMessageState` be worthwhile?

PR #906, opened 2020-08-05
In neqo-http3-server we should read from the sockets more often.

ddragana:


agrover:
Please give description, thanks.

neqo-http3-server/src/main.rs
```diff
@@ -335,6 +337,7 @@ fn init_poll(
     Ok((poll, sockets, servers))
 }
 
+#[allow(clippy::cognitive_complexity)]
```
martinthomson, line 340: Please factor the code rather than suppress warnings.  The loop on curr_active seems to be easily extracted.

neqo-http3-server/src/main.rs
```diff
@@ -424,14 +438,44 @@ fn main() -> Result<(), io::Error> {
...
+                            None,
+                            &mut timer,
+                            socket,
+                        ) {
+                            active_servers.insert(event.token().0);
+                        }
+                    }
+                }
+            }
+            let curr_active = mem::replace(&mut active_servers, HashSet::new());
```
martinthomson, line 454: You can use mem::take here.

neqo-http3-server/src/main.rs
```diff
@@ -424,14 +438,44 @@ fn main() -> Result<(), io::Error> {
...
+                        if process(
+                            &mut **server,
+                            svr_timeout,
+                            event.token().0,
+                            None,
+                            &mut timer,
+                            socket,
+                        ) {
+                            active_servers.insert(inx);
+                        }
```
martinthomson, line 478: This block repeats here and above (Lines 432 to 451).  That means it can be extracted.

neqo-http3-server/src/main.rs
```diff
@@ -385,7 +397,9 @@ fn main() -> Result<(), io::Error> {
                         if let Some((ref mut server, svr_timeout)) =
                             servers.get_mut(&socket.local_addr().unwrap())
                         {
-                            process(&mut **server, svr_timeout, inx, None, &mut timer, socket);
+                            if process(&mut **server, svr_timeout, inx, None, &mut timer, socket) {
+                                active_servers.insert(inx);
+                            }
```
martinthomson, line 402: Why is this block not the same as the block you use when datagrams arrive or when a server is active?

neqo-http3-server/src/main.rs
```diff
@@ -375,8 +378,17 @@ fn main() -> Result<(), io::Error> {
...
     loop {
-        poll.poll(&mut events, None)?;
+        poll.poll(
+            &mut events,
+            if active_servers.is_empty() {
+                None
+            } else {
+                Some(Duration::from_millis(0))
+            },
+        )?;
```
martinthomson, line 391: I assume that you want to poll here to pick up any datagrams that arrived while we were processing stuff from the last round.  It might be worth documenting that.

ddragana:
Doing a little bit of refactoring, did not look good. I decided to do more invasive refactoring of one part of the code.

neqo-http3-server/src/main.rs
```diff
@@ -207,234 +209,299 @@ fn process(
...
+            self.poll.register(
+                &socket,
+                Token(i),
+                Ready::readable() | Ready::writable(),
+                PollOpt::edge(),
+            )?;
+
+            self.sockets.push(socket);
+            self.servers
+                .insert(local_addr, (create_server(&self.args), None));
```
martinthomson, line 371: create_server can be an associated function, rather than a completely separate function.  That is `Self::create_server` or `self.create_server`.

neqo-http3-server/src/main.rs
```diff
@@ -207,234 +209,299 @@ fn process(
...
-                        servers.get_mut(&socket.local_addr().unwrap())
-                    {
-                        process(
+    fn process_and_events(&mut self, inx: usize, read_socket: bool) -> Result<(), io::Error> {
+        if let Some(socket) = self.sockets.get_mut(inx) {
+            if let Some((ref mut server, svr_timeout)) =
+                self.servers.get_mut(&socket.local_addr().unwrap())
+            {
+                if read_socket {
+                    loop {
```
martinthomson, line 386: Do we really want to loop in this way?

What this does is reads every available datagram from the same socket and then processes each in turn, including event processing.  Would it be better to loop over every available datagram first and then process the events in a single pass?  That would allow for a different factoring of this function (this still contains repetition on either side of the read_socket condition that this could erase).

neqo-http3-server/src/main.rs
```diff
@@ -207,234 +209,299 @@ fn process(
...
+        Ok(())
+    }
 
-                    if sz == 0 {
-                        eprintln!("zero length datagram received?");
-                    } else if let Some((ref mut server, svr_timeout)) =
-                        servers.get_mut(&socket.local_addr().unwrap())
-                    {
-                        process(
+    fn process_and_events(&mut self, inx: usize, read_socket: bool) -> Result<(), io::Error> {
```
martinthomson, line 380: The name here is strange.  I think that this is process_datagrams_and_events, but that's pretty lengthy.

PR #913, opened 2020-08-07
Ack frequency extension

martinthomson:
Tests are WIP, plus we shouldn't land this while the spec is unstable, because this will advertise support and we might choke on badly formed frames.

This assumes a structure to the ACK_FREQUENCY frame that doesn't exist in the draft, specifically:

```
ACK_FREQUENCY Frame {
  Type (i) = 0xaf,
  Sequence Number (i),
  Ack Delay (i),
  Packet Threshold (i),
  Loss Threshold (i),
}
```

This PR needs to sit until that discussion stabilizes a little more.  Shipping this would be irresponsible.

(One thing we might consider taking from this in the meantime is the code that advertises the MAX_ACK_DELAY transport parameter.)


martinthomson:
OBE

PR #918, opened 2020-08-10
Add tests with different network conditions and add throughput measuremens

ddragana:


ddragana:
I need to debug this a bit to understand throughput numbers better and also to tune it a bit, because it is taking too long to run, the transfer size is to high for some scenarios.  (I am planning to do this today.)

neqo-transport/tests/sim/taildrop.rs
```diff
@@ -71,6 +75,9 @@ impl TailDrop {
             dropped: 0,
             delivered: 0,
             maxq: 0,
+            start_time: None,
+            last_sent: None,
```
martinthomson, line 79: Maybe call these first_packet_time and last_packet_time so it is clear what these are for.

neqo-transport/tests/sim/taildrop.rs
```diff
@@ -171,8 +183,22 @@ impl Node for TailDrop {
 
     fn print_summary(&self, test_name: &str) {
         println!(
-            "{}: taildrop: rx {} drop {} tx {} maxq {}",
-            test_name, self.received, self.dropped, self.delivered, self.maxq,
+            "{}: taildrop: rx {} drop {} tx {} maxq {} thruoghput {}",
```
martinthomson, line 186: You should calculate throughput in a separate step.  Also:

```suggestion
            "{}: taildrop: rx {} drop {} tx {} maxq {} throughput {}",
```

neqo-transport/tests/sim/taildrop.rs
```diff
@@ -151,6 +158,9 @@ impl TailDrop {
 impl Node for TailDrop {
     fn process(&mut self, d: Option<Datagram>, now: Instant) -> Output {
         if let Some(dgram) = d {
+            if self.start_time.is_none() {
+                self.start_time = Some(now);
+            }
```
martinthomson, line 163: You should move this to the init() function, which exists for this purpose.  That won't capture when the first packet was injected into this link, but it is close enough (the server->client link will see a packet 1/2 RTT later in most cases).

neqo-transport/tests/network.rs
```diff
@@ -148,6 +149,147 @@ simulate!(
...
+}
+
+// bandwidth 10Mbps, 20Mbps, 50Mbps and 100Mbps
+// delay by 10ms, 50ms, 100ms and 250ms
+// loss 0%
+// With tail-drop queue
+#[test]
+fn diff_conditions_client_upload_taildrop() {
+    for b in &[10, 20, 50, 100] {
+        for delay in &[10, 25, 50, 100, 250] {
```
martinthomson, line 266: You don't seem to be using `b` here other than in debugging.  Also, with 20 different configurations, the large transfer is going to take a long time to run.

neqo-transport/tests/network.rs
```diff
@@ -148,6 +149,147 @@ simulate!(
...
+    }
+}
+
+// bandwidth 10Mbps, 20Mbps, 50Mbps and 100Mbps
+// delay by 10ms, 50ms, 100ms and 250ms
+// loss 0%
+// With tail-drop queue
+#[test]
+fn diff_conditions_client_download_tail_drop() {
+    for b in &[10, 20, 50, 100] {
```
martinthomson, line 231: See below regarding `b`.

neqo-transport/tests/network.rs
```diff
@@ -148,6 +149,147 @@ simulate!(
...
+        for delay in &[10, 25, 50, 100, 250] {
+            Simulator::new(
+                format!("diff conditions upload taildrop only {} {}", b, *delay),
+                boxed![
+                    ConnectionNode::new_client(boxed![
+                        ReachState::new(State::Confirmed),
+                        SendData::new(TRANSFER_AMOUNT_LARGE),
+                        ReachState::new(State::Closed(ConnectionError::Transport(
+                            Error::IdleTimeout
+                        )))
```
martinthomson, line 275: I don't think that you need to include the two ReachState conditions here.

neqo-transport/tests/network.rs
```diff
@@ -148,6 +149,147 @@ simulate!(
...
+}
+
+// bandwidth 10Mbps, 20Mbps, 50Mbps and 100Mbps
+// delay by 10ms, 50ms, 100ms and 250ms
+// loss 1%, 2%, 5%, 10% and 20%
+#[test]
+fn diff_conditions_client_upload() {
+    for b in &[10, 20, 50, 100] {
+        for delay in &[10, 25, 50, 100, 250] {
+            for loss in &[0, 1, 2, 5, 10, 20] {
```
martinthomson, line 196: Combinatorial explosion of 120 transfers.  Ouch.  I think that we might want to take these tests and put them in a new file behind a feature flag so that regular test runs don't run the whole matrix.

As far as I can see, this is identical to the later tests without loss, so maybe just drop those latter ones.

neqo-transport/tests/network.rs
```diff
@@ -148,6 +149,147 @@ simulate!(
     ],
 );
 
+// bandwidth 10Mbps, 20Mbps, 50Mbps and 100Mbps
+// delay by 10ms, 50ms, 100ms and 250ms
+// loss 1%, 2%, 5%, 10% and 20%
+// With tail-drop queue
+#[test]
+fn diff_conditions_client_download() {
```
martinthomson, line 157: How much do you think that client download and server upload will be different?  The server can send earlier than a client, but that's all I've got.

neqo-transport/tests/sim/taildrop.rs
```diff
@@ -71,6 +75,9 @@ impl TailDrop {
             dropped: 0,
             delivered: 0,
             maxq: 0,
+            start_time: None,
+            last_sent: None,
+            sent_bytes: 0,
```
martinthomson, line 80: Given the number of statistical items we now have, it might pay to factor them out into a TaildropStats struct.  Then the core functioning of this doesn't get swamped with all the stats.

ddragana:
I have not intended to land this as it is and especially not to  run in on every 'cargo test'.
Landing this behind a feature is good idea. 
I want to use this tests as a base for measuring/estimating performance of the library. I want to add  "expected throughput values" that would be the values TCP NewReno would achieve and that would be our goal.
I have tried with TCP equations, but they are not good, so it will be empirical.

neqo-transport/tests/network.rs
ddragana: Yes, they are the same, I should not upload the fist version of the patch, it is never good. I will upload an update soon.

martinthomson:
For me, I'm very happy to see early versions.  Bugs and all :)  It helps me understand what you are doing better, which I think helps overall.  That's why you are seeing drafts from me as well.

PR #925, opened 2020-08-12
Grease the QUIC bit

martinthomson:
This randomizes the value of the spin bit at the same time.

Based on draft-thomson-quic-bit-grease-00, which seems to have attracted quite a bit of attention rather quickly.  Pun intended.

### DIFF
--- a/neqo-transport/src/connection.rs
+++ b/neqo-transport/src/connection.rs
@@ -5874,8 +5874,10 @@ mod tests {
         corrupted[idx] ^= 0x76;
         let dgram = Datagram::new(d.source(), d.destination(), corrupted);
         server.process_input(dgram, now());
-        // The server should have received some packets, but dropped all of them.
-        assert_ne!(server.stats().packets_rx, 0);
-        assert_eq!(server.stats().packets_rx, server.stats().dropped_rx);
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection.rs
```diff
@@ -5874,8 +5874,10 @@ mod tests {
...
         let dgram = Datagram::new(d.source(), d.destination(), corrupted);
         server.process_input(dgram, now());
-        // The server should have received some packets, but dropped all of them.
-        assert_ne!(server.stats().packets_rx, 0);
-        assert_eq!(server.stats().packets_rx, server.stats().dropped_rx);
+        // The server should have received two packets,
+        // the first should be dropped, the second saved.
+        assert_eq!(server.stats().packets_rx, 2);
+        assert_eq!(server.stats().dropped_rx, 1);
+        // assert_eq!(server.stats().saved_datagram, 1);
```
ddragana, line 5881: Should this be a comment?

martinthomson: Yeah, that stat doesn't exist in `main` yet.  I wanted to capture this in a comment though, because it's important to understand what happened to the packet that wasn't dropped.

PR #926, opened 2020-08-14
Wrong check, we should check if a stream is already in he blocked list not if any stream <= streamId is in he list

ddragana:
We should not enter the same stream multiple times into the the block list. The check was wrong it should have checked whether the stream is in the list and not whether streams with id <= streamId are in the list.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -5419,7 +5419,6 @@ mod tests {
 
         // Headers on the second request are blocked as well are blocked
         // waiting fro the encoder instructions.
-        let header_ready_event = |e| matches!(e, Http3ClientEvent::HeaderReady { .. });
         assert!(!client.events().any(header_ready_event));
 
         // Now read headers.
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -1119,10 +1119,12 @@ mod tests {
     // The data frame payload from HTTP_RESPONSE_2 is:
     const EXPECTED_RESPONSE_DATA_2_FRAME_1: &[u8] = &[0x61, 0x62, 0x63];
 
-    fn connect_and_send_request(close_sending_side: bool) -> (Http3Client, TestServer, u64) {
-        let (mut client, mut server) = connect();
-        let request_stream_id = make_request(&mut client, close_sending_side);
-        assert_eq!(request_stream_id, 0);
+    fn make_request_and_exchange_pkts(
+        client: &mut Http3Client,
+        server: &mut TestServer,
+        close_sending_side: bool,
+    ) -> u64 {
+        let request_stream_id = make_request(client, close_sending_side);
 
         let out = client.process(None, now());
         let _ = server.conn.process(out.dgram(), now());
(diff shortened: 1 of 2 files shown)


ddragana:
I will write a test. I hit this while debugging something else on Firefox.

ddragana:
This bug will actually make some requests stall forever, which is not good.

ddragana:
I added a test.

neqo-http3/src/connection_client.rs
```diff
@@ -5358,4 +5368,64 @@ mod tests {
...
+        let header_ready_event = |e| matches!(e, Http3ClientEvent::HeaderReady { .. });
+        assert!(!client.events().any(header_ready_event));
+
+        // Make another request.
+        let request2 = make_request_and_exchange_pkts(&mut client, &mut server, true);
+        // Send response headers.
+        server_send_response_and_exchange_packet(&mut client, &mut server, request2, &d, true);
+
+        // Headers on the second request are blocked as well are blocked
+        // waiting fro the encoder instructions.
```
martinthomson, line 5415:
```suggestion
        // waiting for the encoder instructions.
```

neqo-http3/src/connection_client.rs
```diff
@@ -5358,4 +5368,64 @@ mod tests {
...
+        hframe.encode(&mut d);
+        server_send_response_and_exchange_packet(
+            &mut client,
+            &mut server,
+            request_stream_id,
+            &d,
+            true,
+        );
+
+        // Headers are blocked waiting fro the encoder instructions.
```
martinthomson, line 5405:
```suggestion
        // Headers are blocked waiting for the encoder instructions.
```

neqo-http3/src/connection_client.rs
```diff
@@ -5358,4 +5368,64 @@ mod tests {
...
+        // Make another request.
+        let request2 = make_request_and_exchange_pkts(&mut client, &mut server, true);
+        // Send response headers.
+        server_send_response_and_exchange_packet(&mut client, &mut server, request2, &d, true);
+
+        // Headers on the second request are blocked as well are blocked
+        // waiting fro the encoder instructions.
+        assert!(!client.events().any(header_ready_event));
+
+        // Now read headers.
```
martinthomson, line 5418:
```suggestion
        // Now make the encoder instructions available.
```

neqo-http3/src/connection_client.rs
```diff
@@ -5358,4 +5368,64 @@ mod tests {
...
+        let _ = client.process(encoder_insts.dgram(), now());
+
+        // find the new request/response stream and send frame v on it.
```
martinthomson, line 5421: I think that the comment is incorrect.

```suggestion
        // Header blocks for both streams should be ready.
```

neqo-http3/src/connection_client.rs
```diff
@@ -5358,4 +5368,64 @@ mod tests {
...
+        let mut count_responses = 0;
+        while let Some(e) = client.next_event() {
```
martinthomson, line 5423: I've been wondering whether we should provide an `Iterator` implementation for events so that we can do these things more efficiently.  The code you have here is roughly:

```rust
client.events().filter(|e| matches!(e, Http3ClientEvent::HeaderReady { .. })).count();
```

We could get that by doing:

```rust
trait ConnectionEventProvider {
    fn next_event(&mut self) -> Option<ConnectionEvent>;
}

struct<'a, T> ConnectionEventIter<T> {
    c: &'a T;
}
    
impl<'a, T> Iterator for ConnectionEventIter<'a, T> 
    where T: ConnectionEventProvider {
    type Item = ConnectionEvent;
    fn next(&mut self) -> Option<Self::Item> {
        self.c.next_event()
    }
}
```

martinthomson:
Oh, I forgot.  Fix the nits and land.  We can take the refactoring elsewhere.

neqo-http3/src/connection_client.rs
ddragana: too much copy-pasting :(

PR #928, opened 2020-08-17
Post an event when a connection changes to ZeroRtt state.

ddragana:
Also add a funcion is_zerortt().

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -185,7 +185,9 @@ impl Http3ClientEvents {
             // If closing, existing events no longer relevant.
             Http3State::Closing { .. } | Http3State::Closed(_) => self.events.borrow_mut().clear(),
             Http3State::Connected => {
-                self.remove(|evt| matches!(evt, Http3ClientEvent::StateChange(Http3State::ZeroRtt)));
+                self.remove(|evt| {
+                    matches!(evt, Http3ClientEvent::StateChange(Http3State::ZeroRtt))
+                });
             }
             _ => (),
         }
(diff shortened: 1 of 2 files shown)


ddragana:
We need this for necko. I am not sure which one will be better to use he event or the is_zerortt(), I need to test it a bit. The event will work for sure, but the function may make some things faster but it may be tricky because on the order of events... I just need to test it a bit.

neqo-http3/src/client_events.rs
```diff
@@ -184,6 +184,9 @@ impl Http3ClientEvents {
         // If closing, existing events no longer relevant.
```
martinthomson, line 184: Fix the commentary or move it.

neqo-http3/src/connection_client.rs
```diff
@@ -3246,6 +3253,9 @@ mod tests {
             .expect("Set resumption token.");
 
         assert_eq!(client.state(), Http3State::ZeroRtt);
+        let zerortt_event = |e| matches!(e, Http3ClientEvent::StateChange(Http3State::ZeroRtt));
+        assert!(client.events().any(zerortt_event));
+        assert!(client.is_zerortt());
```
martinthomson, line 3258: How does `is_zerortt` differ from `state() == Http3State::ZeroRtt` ?

neqo-http3/src/client_events.rs
```diff
@@ -184,6 +184,9 @@ impl Http3ClientEvents {
...
         match state {
             Http3State::Closing { .. } | Http3State::Closed(_) => self.events.borrow_mut().clear(),
+            Http3State::Connected => {
+                self.remove(|evt| matches!(evt, Http3ClientEvent::StateChange(Http3State::ZeroRtt)))
```
martinthomson, line 188:
```suggestion
                self.remove(|evt| matches!(evt, Http3ClientEvent::StateChange(Http3State::ZeroRtt)));
```

neqo-http3/src/connection_client.rs
ddragana: Ok, I can remove is_zerortt. I can implement it in neqo_glue code.

PR #933, opened 2020-08-17
More fixes for persistent congestion

### LINKED ISSUE #831: Defer start of persistent congestion
Persistent congestion cannot include packets that are sent prior to getting an RTT estimate.

We don't get an RTT estimate from Initial packets very often, as we often discard keys before getting that measurement.  Handshake packets often do produce an estimate, but that doesn't necessarily help with 0-RTT and 0.5-RTT packets.  In other words, RTT measurement is not a natural product of the handshake, so it cannot be tied to anything in the handshake.

Thus, the persistent congestion measurement needs to know the packet number of the first packet sent after receiving an RTT measurement.  This is complicated by there being multiple packet number spaces.  We can either track only the Application Data space, track the time that we received an RTT estimate, or change to a single space for packet numbers.  The last is something done in some implementations, but I don't think that it would suit ours well.  Recording the time of first estimate is likely best.

We can replace the counter I recently added with `Option<Instant>` as the counter being non-zero was the only valuable option.

martinthomson:
This avoids starting persistent congestion prior to having an RTT
sample and adds more tests.

Closes #831.

### DIFF
--- a/neqo-transport/src/cc.rs
+++ b/neqo-transport/src/cc.rs
@@ -164,6 +164,47 @@ impl CongestionControl {
         );
     }
 
+    fn detect_persistent_congestion(
+        &mut self,
+        prev_largest_acked_sent: Option<Instant>,
+        pto: Duration,
+        lost_packets: &[SentPacket],
+    ) {
+        let pc_period = pto * PERSISTENT_CONG_THRESH;
+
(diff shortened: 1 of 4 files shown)


martinthomson:
The build problem here is fixed in #935.  It is caused by me updating this file to use clippy::pedantic.

PR #935, opened 2020-08-18
Use usize::MAX

martinthomson:
Rust 1.43 added `usize::MAX` to replace `std::usize::MAX`.  Clippy now
warns about the latter.  So we should do the same.

(This is not consistently warning, but as the problem appears some
places, we need to make the change everywhere.)


martinthomson:
Subsumed by #933.

PR #936, opened 2020-08-18
Display the connection ID in logs

martinthomson:
Using the pointer seems to result in the same value from being used multiple times as the same memory location appears to be reused for holding different connections, so let's avoid doing that for connections at least.


martinthomson:
Overriding mergify, which is playing up.

Issue #929, opened 2020-08-17
Expiration time for a resumption token


ddragana:
Necko use the expiration time to decide if it should replace a token that is already in the cache.
The time is read from the resumption token. The token from neqo does no contain only a raw TLS token but it also has other parameters so to use nss function to extract the expiration time in necko would mean that we need to parse the token and find the start of the TLS token. I would like to avoid doing this. here is 2 option:
1) neqo does the parsing of the token and return the token with the expiration time
2) we assume that the new token is newer and we always replace a token in the cache

martinthomson:
So newer tokens will always be more relevant.  The question is whether expiration time might cause you not to consider a particular token.

ddragana:
I will ask Dana about this.

ddragana:
I should have looked more carefully. I was completely wrong. We use expiration time to prioritize removing tokens from he cache, because thee cache has a limited size.

We should extract the time in neqo.

PR #941, opened 2020-08-19
Expose the raw tls resumption token

ddragana:


ddragana:
Fixes #929

ddragana:
I have decided that extracting expiration time in neqo is not worth the trouble. I will just expose the raw tls token and necko can extract time.

martinthomson:
Before we do this, I think that you should look at #893.  That has the resumption token being pulled from where it lives.  That would fail here.

I think that the right approach is to have neqo-crypto call `SSL_GetResumptionTokenInfo` when it gets a new token from NSS and pass any relevant metadata along with the token.  After we wrap that token (in neqo-transport and neqo-http3) we can pass the meta-information along (or modify it).  That's more work, but cleaner.

While the patient is open, we can hook up some event reporting.

ddragana:
I reviewed #893 after making this PR :) And I know it would be hack to make it work. I will make a new PR.

PR #943, opened 2020-08-20
Extract the expiration time from a resumption token

### LINKED ISSUE #929: Expiration time for a resumption token

ddragana:
Fixes #929


ddragana:
This was easier than I thought.

martinthomson (commented):
I just found this review sitting here.  I think that this is generally good, but have some thoughts for improvement.

ddragana:
Most of this comments are already addressed if you look at the last version:
https://github.com/mozilla/neqo/commit/94c70031d3fed736355701f74edd9f1ebe896ab9

martinthomson (approved):

martinthomson:
Mergify is busted.  We need to decide whether to keep it or not, because I can't see how it should fail like this.

PR #944, opened 2020-08-20
Stop reading ranges when we have filled a buffer.

ddragana:


ddragana:
This is a corner case though.

martinthomson (changes requested):
A test would be great.

PR #945, opened 2020-08-21
Change capacity if possible before starting to encode a header block.

ddragana:
Currently it may happen that a capacity change is executed during encoding of a header block. The capacity will be changed by calling send(). If capacity changes some parameters like max_entries change as well and Encoder and HeaderEncoder are out of sync. Therefore call send() before starting to encode a header.

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -244,7 +244,6 @@ impl QPackEncoder {
         value: &[u8],
     ) -> Res<u64> {
         qdebug!([self], "insert {:?} {:?}.", name, value);
-        self.send(conn)?;
 
         let entry_size = name.len() + value.len() + ADDITIONAL_TABLE_ENTRY_SIZE;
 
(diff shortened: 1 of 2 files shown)


neqo-qpack/src/encoder.rs
```diff
@@ -359,6 +358,20 @@ impl QPackEncoder {
         stream_id: u64,
     ) -> Res<HeaderEncoder> {
         qdebug!([self], "encoding headers.");
+
```
martinthomson, line 361: This looks good.  I can see where this might interleave instructions.  But I couldn't work out which test case would have caused this to break.

ddragana: It is test zero_rtt_send_request. To trigger the bug:
- 0rtt needs to be used(so we have a capacity change waiting to be sent. Without 0rtt we usually send the capacity change before we try to send a request, but the bug may be triggered in that code path as well)
- a header that is not in dynamic or static table need to be present in a request, e.g. "newheader: newvalue".
In test zero_rtt_send_request I have added &[(String::from("myheaders"), String::from("myvalue"))].

martinthomson (approved):
Thanks.  I was having trouble seeing how this would be caused, but the explanation helped.

PR #946, opened 2020-08-25
Read buffer

martinthomson:
I originally just did this so that #944 could have a test case, but I then saw the inefficient code and decided to take a swing at that too.

This cleans up the code a little and makes the removal of ranges more efficient (using `BTreeMap::split_off` rather than iterative `BTreeMap::remove`).

It turns out that the extra break added in #944 isn't necessary for correctness, but it saves on doing a zero-length copy, so I've left it in (but moved it).  New tests ensure that this code path is exercised in tests.

Closes #944.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -225,9 +225,9 @@ impl RxStreamOrderer {
     fn read(&mut self, buf: &mut [u8]) -> usize {
         qtrace!("Reading {} bytes, {} available", buf.len(), self.buffered());
         let mut copied = 0;
-        let mut keep = None;
 
         for (&range_start, range_data) in &mut self.data_ranges {
+            let mut keep = false;
             if self.retired >= range_start {
                 // Frame data has new contiguous bytes.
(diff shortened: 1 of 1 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -733,6 +731,48 @@ mod tests {
...
+        let mut s = RxStreamOrderer::new();
+
+        // Add two chunks.
+        s.inbound_frame(0, vec![0; CHUNK_SIZE]).unwrap();
+        let offset = u64::try_from(CHUNK_SIZE).unwrap();
+        s.inbound_frame(offset, vec![0; EXTRA_SIZE]).unwrap();
+
+        let mut buf = vec![0; 1];
+        for _ in 0..CHUNK_SIZE + EXTRA_SIZE {
+            // Read, providing only enough space for some of the first chunk.
```
ddragana, line 769: Update this comment it is not correct.

ddragana (commented):
Looks good. fix the comment and land it.

PR #949, opened 2020-09-08
Drop frame type from application CONNECTION_CLOSE

martinthomson:
How this escaped notice for so long is somewhat shocking.


ddragana (approved):

martinthomson:
OK, mergify seems to have completely dropped the ball somewhere.  Not sure how.

PR #950, opened 2020-09-09
Post ResumptionToken events

ddragana:
The events are added when a new token or a ticket is received. A new token and ticket may not be sent/received at the same time, therefore we need to remember the last token and the last ticket to be able to add them to a new resumption token, e.g. if we only receive a new ticket we use the last value of the token to create the resumption token.


ddragana:
This adds a resumptionToken event. I will leave it as a draft for now, because I am not really convinced yet.

neqo-crypto/src/agent.rs
```diff
@@ -686,14 +686,20 @@ impl ::std::fmt::Display for SecretAgent {
     }
 }
 
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone)]
+pub struct ResumptionToken {
+    pub token: Vec<u8>,
+    pub expiration_time: i64,
+}
```
martinthomson, line 693: We can probably implement `AsRef<[u8]>` for this to get at the contents of the token more easily.  And modify those functions that take a token to take `impl AsRef<[u8]>` instead of `&[u8]`.  That's ergonomics and should avoid needing to make the fields public.

For the expiration time, this needs to be an `Instant`, not `PRTime` and I would still prefer to use an accessor.  We should not be exposing the internal structure of objects.

neqo-crypto/tests/handshake.rs
```diff
@@ -146,6 +146,6 @@ pub fn resumption_setup(mode: Resumption) -> (Option<AntiReplay>, Vec<u8>) {
 
     // `client` is about to go out of scope,
     // but we only need to keep the resumption token, so clone it.
-    let token = client.resumption_token().expect("token is present");
+    let token = client.resumption_token().expect("token is present").token;
     (anti_replay, token)
```
martinthomson, line 150: Let's return the `ResumptionToken` instance here.

neqo-http3/src/connection_client.rs
```diff
@@ -151,21 +151,16 @@ impl Http3Client {
...
-            } else {
-                None
-            }
-        } else {
-            None
+    fn create_resumption_token(&mut self, token: ResumptionToken) {
+        if let Some(settings) = self.base_handler.get_settings() {
+            let mut enc = Encoder::default();
+            settings.encode_frame_contents(&mut enc);
+            enc.encode(&token.token[..]);
```
martinthomson, line 159: Can we add a method on ResumptionToken to append/prepend some additional information?  We're doing this in a couple of places now and it would be good to have common code for creating a new token from an existing one with some extra stuff added in.

neqo-transport/src/addr_valid.rs
```diff
@@ -263,23 +259,23 @@ impl AddressValidation {
...
     }
 
     /// If this is a client, take a token if there is one.
     /// If this is a server, panic.
-    pub fn take_token(&mut self) -> Option<Vec<u8>> {
-        if let Self::Client(ref mut tokens) = self {
-            tokens.pop()
+    pub fn token(&mut self) -> Option<Vec<u8>> {
+        if let Self::Client(token) = self {
+            token.clone()
```
martinthomson, line 278: Please don't do this.  Cloning tokens risks using them twice.

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -136,11 +131,17 @@ fn two_tickets_and_tokens() {
...
     client.process_input(pkt, now());
-    let token1 = client.resumption_token();
-    let token2 = client.resumption_token();
-    assert_ne!(token1, token2);
-    assert!(client.resumption_token().is_none());
-
-    can_resume(token1, true);
-    can_resume(token2, true);
+    let tokens = get_tokens(&mut client);
+    assert_eq!(tokens.len(), 4);
```
martinthomson, line 135: This seems wrong.

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -89,11 +89,9 @@ fn address_validation_token_resume() {
     assert!(server.crypto.tls.info().unwrap().resumed());
 }
 
-fn can_resume(mut token: Option<Vec<u8>>, initial_has_token: bool) {
+fn can_resume(token: &[u8], initial_has_token: bool) {
```
martinthomson, line 92: Let's pass ResumptionToken here too.

neqo-transport/src/connection/mod.rs
```diff
@@ -1828,13 +1825,24 @@ impl Connection {
...
+                    let mut tokens = Vec::new();
+                    if let Agent::Client(ref mut c) = self.crypto.tls {
+                        while let Some(t) = c.resumption_token() {
+                            tokens.push(t);
+                        }
+                    }
+                    while let Some(t) = tokens.pop() {
+                        self.last_token = Some(t);
+                        self.create_resumption_token();
+                    }
```
martinthomson, line 1837: Let's factor this out somehow.  Probably best to make a new function in `crypto.rs` for getting any new tokens that have arrived.  That can deal with the `crypto.tls` internals.

neqo-transport/src/connection/mod.rs
```diff
@@ -1828,13 +1825,24 @@ impl Connection {
...
                 } else {
                     // If we get a useless CRYPTO frame send outstanding CRYPTO frames again.
                     self.crypto.resend_unacked(space);
                 }
             }
             Frame::NewToken { token } => {
                 self.new_token.save_token(token);
+                self.create_resumption_token();
```
martinthomson, line 1845: Ahh, I see how you are getting 4 token events.  I don't think that we want this.  See the meta-comment.

neqo-transport/src/connection/mod.rs
```diff
@@ -267,7 +267,7 @@ pub struct Connection {
     new_token: NewTokenState,
     stats: StatsCell,
     qlog: NeqoQlog,
-
+    last_token: Option<ResumptionToken>,
```
martinthomson, line 270: I don't think that we need to hold anything here.  If we do need to hold some state so that we can match up tickets with NEW_TOKEN, let's make a new object to manage it.

martinthomson (changes requested):
OK, I got through half of this and I don't think that it is going to work very well.  The problem here is that under ideal circumstances we want to get both the TLS session ticket and the NEW_TOKEN from the server and put those together.

A purely event-based API won't work for this because we might get one without the other.  We can put the event on a timer and release session tickets without NEW_TOKEN if that pops, I guess, but that seems clunky.

The other option is to enqueue events for session tickets and opportunistically match NEW_TOKEN to those.  We can buffer NEW_TOKEN (say up to 4, as current code does) and pull from that queue as session tickets arrive.  And if NEW_TOKEN arrives late, we can match them up to session ticket events that haven't been taken yet.  That's a little asymmetric in that late NEW_TOKEN might not get through if they are sent after tickets and events are polled fairly aggressively.

ddragana:
I know this is not working well, that why it is only draft. I also do not like any of ideas I have how to fix it. 
I do not like using timers to send events without NEW_TOKEN.
My last idea was to send events when we have both. If we do not have a token we do not send an event. The app can call get_resumption_token if it needs a resumption token and it did not get an event (e.g. just before closing a connection or when an app want to create a new connection in paralel) and in that case we will return a resumption token without a TOKEN only with a ticket.
The second idea I had is to send a resumption_token_event without a TOKEN just before a CONNECTION_CLOSE event, but that may be too late for an app.

This PR contains PR #943 as well. I would like to land #943 separately because it is easier and add the event in this PR. (I have updated #943 according to your comments in this PR)

PR #956, opened 2020-09-14
Return an error if a reserved frame type is received

ddragana:


neqo-http3/src/connection_client.rs
```diff
@@ -5458,4 +5458,15 @@ mod tests {
         }
         assert_eq!(count_responses, 2);
     }
+
+    #[test]
+    fn reseerved_frames() {
```
martinthomson, line 5463:
```suggestion
    fn reserved_frames() {
```

martinthomson (commented):
Looks good.  Fix the typo and land it.

PR #958, opened 2020-09-14
Receiving a reserved settings type should return a connection error H3_SETTINGS_ERROR

ddragana:


neqo-http3/src/connection_client.rs
```diff
@@ -5458,4 +5458,23 @@ mod tests {
...
+    #[test]
+    fn send_reserved_settings() {
+        for s in H3_RESERVED_SETTINGS {
+            let (mut client, mut server) = connect_only_transport();
+            // Create server control stream.
+            let control_stream = server.conn.stream_create(StreamType::UniDi).unwrap();
+            // Send the control stream type(0x0), a settings frame(0x4) of length 2.
+            let mut sf = vec![0x0, 0x4, 0x2];
+            // The settings frame contains a reserved settings type and some value (0x1).
+            sf.push(*s as u8);
```
martinthomson, line 5471: Use try_from
```suggestion
            sf.push(u8::try_from(*s).unwrap());
```

neqo-http3/src/connection_client.rs
```diff
@@ -5458,4 +5458,23 @@ mod tests {
         }
         assert_eq!(count_responses, 2);
     }
+
+    #[test]
+    fn send_reserved_settings() {
+        for s in H3_RESERVED_SETTINGS {
+            let (mut client, mut server) = connect_only_transport();
+            // Create server control stream.
```
martinthomson, line 5466:
```suggestion
```

neqo-http3/src/connection_client.rs
```diff
@@ -5458,4 +5458,23 @@ mod tests {
...
+            let control_stream = server.conn.stream_create(StreamType::UniDi).unwrap();
+            // Send the control stream type(0x0), a settings frame(0x4) of length 2.
+            let mut sf = vec![0x0, 0x4, 0x2];
```
martinthomson, line 5469: It might be better to use Encoder here and just write this out in full.  Even if that means a little more code.

martinthomson (commented):
Good.  Fix and land.

PR #959, opened 2020-09-15
Add stats that show how many PTOs a connection had

ddragana:
Count single PTO, 2 PTOs in a row, 3 PTOs in a row, etc separately.

### DIFF
--- a/neqo-transport/src/recovery.rs
+++ b/neqo-transport/src/recovery.rs
@@ -552,9 +552,8 @@ struct PtoState {
 }
 
 impl PtoState {
-    pub fn new(space: PNSpace, probe: PNSpaceSet, stats: &mut Stats) -> Self {
+    pub fn new(space: PNSpace, probe: PNSpaceSet) -> Self {
         debug_assert!(probe[space]);
-        stats.add_pto_count(1);
         Self {
             space,
             count: 1,
@@ -563,17 +562,20 @@ impl PtoState {
         }
     }
 
-    pub fn pto(&mut self, space: PNSpace, probe: PNSpaceSet, stats: &mut Stats) {
+    pub fn pto(&mut self, space: PNSpace, probe: PNSpaceSet) {
         debug_assert!(probe[space]);
         self.space = space;
         self.count += 1;
-        stats.add_pto_count(self.count);
         self.packets = PTO_PACKET_COUNT;
         self.probe = probe;
     }
 
     pub fn count(&self) -> usize {
         self.count
+    }
+
+    pub fn count_pto(&self, stats: &mut Stats) {
+        stats.add_pto_count(self.count);
     }
 
     /// Generate a sending profile, indicating what space it should be from.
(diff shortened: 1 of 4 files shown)


neqo-transport/src/stats.rs
```diff
@@ -13,6 +13,8 @@ use std::fmt::{self, Debug};
 use std::ops::Deref;
 use std::rc::Rc;
 
+pub(crate) const MAX_PTO_COUNTS: usize = 10;
```
martinthomson, line 16: Not sure if this can be fully public.  It's going to be visible via pto_counts, which is also public.  Note that this means we are committing to this value.

This works out being enough PTOs for a connection to hit an idle timeout at around a 16ms base PTO.  That seems like it might be high, but it's not very high.

neqo-transport/src/stats.rs
```diff
@@ -56,6 +62,23 @@ impl Stats {
             self.dropped_rx
         )
     }
+
+    pub fn add_pto_count(&mut self, count: usize) {
+        if count > 0 {
+            let inx = if count < MAX_PTO_COUNTS {
+                count - 1
+            } else {
+                MAX_PTO_COUNTS - 1
```
martinthomson, line 71: IMPORTANT: When this saturates, your assertion below will fail.  Try:

```rust
if count >= MAX_PTO_COUNTS {
    // We can't move this count any further, so stop.
    return;
}
self.pto_counts[count - 1] += 1;
if count > 1 {
    self.pto_counts[count - 2] -= 1;
}
...
```

neqo-transport/src/recovery.rs
```diff
@@ -904,9 +907,13 @@ impl LossRecovery {
...
     fn fire_pto(&mut self, pn_space: PNSpace, allow_probes: PNSpaceSet) {
         if let Some(st) = &mut self.pto_state {
-            st.pto(pn_space, allow_probes);
+            st.pto(pn_space, allow_probes, &mut *self.stats.borrow_mut());
         } else {
-            self.pto_state = Some(PtoState::new(pn_space, allow_probes));
+            self.pto_state = Some(PtoState::new(
+                pn_space,
+                allow_probes,
+                &mut *self.stats.borrow_mut(),
```
martinthomson, line 915: Save on duplication and add a function to PtoState that accumulates stats like:

```rust
self.pto_state.as_mut().unwrap().count_pto(&mut *self.stats.borrow_mut());
```
Then you don't count in two places and invoke in two places.  The only icky bit is the unwrap() call.

martinthomson (commented):
There's a bug here that will cause crashes, but it's trivial to fix.  The code is good otherwise.

neqo-transport/src/stats.rs
```diff
@@ -56,6 +62,20 @@ impl Stats {
             self.dropped_rx
         )
     }
+
+    pub fn add_pto_count(&mut self, count: usize) {
+        if count > 0 {
```
martinthomson, line 67: Can we instead assert that count > 0?

martinthomson (approved):

Issue #953, opened 2020-09-11
Events for resumption tokens


martinthomson:
@ddragana, we should probably discuss design options in an issue rather than a pull request.

On #950 you said: 

>  My last idea was to send events when we have both. If we do not have a token we do not send an event. The app can call get_resumption_token if it needs a resumption token and it did not get an event (e.g. just before closing a connection or when an app want to create a new connection in paralel) and in that case we will return a resumption token without a TOKEN only with a ticket.
>
> The second idea I had is to send a resumption_token_event without a TOKEN just before a CONNECTION_CLOSE event, but that may be too late for an app.

These are both interesting takes on this problem.

The first fails because you don't guarantee that you ever get NEW_TOKEN for every session ticket.  There are legitimate reasons for a server to only send tickets (they don't need address validation, because they don't do 0-RTT, for instance), so that isn't great.  And we certainly can't assume that NEW_TOKEN will always be sent before the session ticket (our code sends NEW_TOKEN second).

The second suggests a way we can avoid getting completely trapped by that, but for many cases that isn't going to work well either.  We could keep our current interface, but make it clear that this only works when NEW_TOKEN isn't available (that is, we provide events when we have both and you can manually pull a session ticket).

Frankly, I think that the timer is the best option out of those presented.  Hold the event for a PTO (or three) and then let it out if we don't see NEW_TOKEN before then.

We can then drop the function for retrieving tokens.

ddragana:
> @ddragana, we should probably discuss design options in an issue rather than a pull request.
> 
> On #950 you said:
> 
> > My last idea was to send events when we have both. If we do not have a token we do not send an event. The app can call get_resumption_token if it needs a resumption token and it did not get an event (e.g. just before closing a connection or when an app want to create a new connection in paralel) and in that case we will return a resumption token without a TOKEN only with a ticket.
> > The second idea I had is to send a resumption_token_event without a TOKEN just before a CONNECTION_CLOSE event, but that may be too late for an app.
> 
> These are both interesting takes on this problem.
> 
> The first fails because you don't guarantee that you ever get NEW_TOKEN for every session ticket. There are legitimate reasons for a server to only send tickets (they don't need address validation, because they don't do 0-RTT, for instance), so that isn't great. And we certainly can't assume that NEW_TOKEN will always be sent before the session ticket (our code sends NEW_TOKEN second).
> 
[...]

PR #960, opened 2020-09-16
Post ResumptionToken events

ddragana:
If both resumption a ticket and a token have been received post a ResumptionToken event immediately.
If only a ticket has been received wait for a token for 3 * PTO before posting a ResumptionToken event.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -511,38 +511,39 @@ impl Connection {
         );
 
         while self.crypto.has_resumption_token() && self.new_token.has_token() {
-            if let Some(token) = self.crypto.create_resumption_token(
-                self.new_token.take_token(),
-                self.tps
-                    .borrow()
-                    .remote
-                    .as_ref()
-                    .expect("should have transport parameters"),
-                u64::try_from(self.loss_recovery.rtt().as_millis()).unwrap_or(0),
-            ) {
-                self.events.client_resumption_token(token)
-            } else {
-                unreachable!("This is a client and there must be a ticket.");
-            }
-        }
-
-        // If we have a resumption ticket check or set a timer.
-        if self.crypto.has_resumption_token() {
-            let arm = if let Some(expiration_time) = self.release_resumption_token_timer {
-                if expiration_time <= now {
-                    if let Some(token) = self.crypto.create_resumption_token(
-                        None,
+            self.events.client_resumption_token(
+                self.crypto
+                    .create_resumption_token(
+                        self.new_token.take_token(),
                         self.tps
                             .borrow()
                             .remote
                             .as_ref()
                             .expect("should have transport parameters"),
                         u64::try_from(self.loss_recovery.rtt().as_millis()).unwrap_or(0),
-                    ) {
-                        self.events.client_resumption_token(token);
-                    } else {
-                        unreachable!("This is a client and there must be a ticket.");
-                    }
+                    )
+                    .unwrap(),
+            );
+        }
+
+        // If we have a resumption ticket check or set a timer.
+        if self.crypto.has_resumption_token() {
+            let arm = if let Some(expiration_time) = self.release_resumption_token_timer {
+                if expiration_time <= now {
+                    self.events.client_resumption_token(
+                        self.crypto
+                            .create_resumption_token(
+                                None,
+                                self.tps
+                                    .borrow()
+                                    .remote
+                                    .as_ref()
+                                    .expect("should have transport parameters"),
+                                u64::try_from(self.loss_recovery.rtt().as_millis()).unwrap_or(0),
+                            )
+                            .unwrap(),
+                    );
+
                     self.release_resumption_token_timer = None;
 
                     // This means that we release one session ticket every 3 PTOs
@@ -836,10 +837,10 @@ impl Connection {
             }
         }
 
-        if let Some(t) = self.release_resumption_token_timer {
-            qtrace!([self], "Resumption token event timer {:?}", t);
-            delays.push(t);
-        }
+        // `release_resumption_token_timer` is not considered here, because
+        // it is not important enough to force the application to set a
+        // timeout for it  It is expected thatt other activities will
+        // drive it.
 
         let earliest = delays.into_iter().min().unwrap();
         // TODO(agrover, mt) - need to analyze and fix #47
@@ -2543,6 +2544,11 @@ impl Connection {
     /// previously-queued events, or cause new events to be generated.
     pub fn next_event(&mut self) -> Option<ConnectionEvent> {
         self.events.next_event()
+    }
+
+    #[cfg(test)]
+    pub fn get_pto(&self) -> Duration {
+        self.loss_recovery.pto_raw(PNSpace::ApplicationData)
     }
 }
 
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -267,6 +267,8 @@ pub struct Connection {
     new_token: NewTokenState,
     stats: StatsCell,
     qlog: NeqoQlog,
+    /// A session ticket was received without NEW_TOKEN,
+    /// this is when that turns into an event without NEW_TOKEN.
     release_resumption_token_timer: Option<Instant>,
     quic_version: QuicVersion,
 }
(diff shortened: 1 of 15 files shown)


ddragana:
Fixes #953

neqo-transport/src/connection/mod.rs
```diff
@@ -495,36 +496,53 @@ impl Connection {
...
+            }
+            if self.release_resumption_token_timer.unwrap() > now {
+                return;
+            }
+        }
+
+        // Create reumpion tokens:
+        //  1) create only one if we have not receiveed a new_token,
+        //  2) otherwise create as many as possible that contain ticket and token.
+        loop {
```
martinthomson, line 522: A few things:

The crypto code shouldn't know about the events.  It can create a token without dispatching it.

You duplicate the timer update code, but that is the same in both places; I would duplicate the token generation piece instead.

That suggests a different structure:

```rust
if self.role == Role::Server || self.state < State::Connected {
    return;
}
while self.crypto.has_resumption_token() && self.new_token.has_token() {
    let tok = self.crypto.create_resumption_token(
        &self.new_token.take_token(),
        self.tps.borrow().remote.as_ref().expect("..."),
        u64::try_from(self.loss_recovery.rtt().as_millis()).unwrap_or(0)),
    );
    self.events.resumption_token(tok);
}
if self.crypto.has_resumption_token() {
    let arm = if let Some(t) = self.resumption_token_timer {
        if t < now {
            let tok = self.crypto.create_resumption_token(
                &[],
                self.tps.borrow().remote.as_ref().expect("..."),
                u64::try_from(self.loss_recovery.rtt().as_millis()).unwrap_or(0)),
            );
            self.events.resumption_token(tok);
            // This means that we release one session ticket every 3 PTOs
            // if no NEW_TOKEN frame is received.
            self.crypto.has_resumption_token()
        } else {
            false
        }
    } else {
        true
    };
    self.resumption_token_timer = if arm {
        Some(3 * self.loss_recovery.pto_raw(PNSpace::ApplicationData))
    } else {
        None
    };
}
```

I considered the option of moving this elsewhere, but this seems like the right trade-off.

A different design would have a partial event created for each session ticket that we receive.  That would be withheld unless that event was too old or a NEW_TOKEN completed it.  However, that means that next_event needs to know what time it is.  We also hit next_event more often than we hit process_output.  The cost of this design is that process_output needs to be run in tests, which has odd side effects that we need to manage, but the tests where we use tokens are fairly limited and we can just deal with the side effects.

neqo-transport/src/connection/mod.rs
```diff
@@ -267,7 +267,7 @@ pub struct Connection {
     new_token: NewTokenState,
     stats: StatsCell,
     qlog: NeqoQlog,
-
+    release_resumption_token_timer: Option<Instant>,
```
martinthomson, line 270:
```suggestion
    /// A session ticket was received without NEW_TOKEN, 
    /// this is when that turns into an event without NEW_TOKEN.
    resumption_token_timer: Option<Instant>,
```

neqo-transport/src/crypto.rs
```diff
@@ -243,6 +245,41 @@ impl Crypto {
...
+                enc.encode_varint(rtt);
+                enc.encode_vvec_with(|enc_inner| {
+                    tps.encode(enc_inner);
+                });
+                let token = new_token.take_token();
+                enc.encode_vvec(token.as_ref().map_or(&[], |t| &t[..]));
+                enc.encode(t.as_ref());
+                qinfo!("resumption token {}", hex_snip_middle(&enc[..]));
+                events
+                    .client_resumption_token(ResumptionToken::new(enc.into(), t.expiration_time()));
```
martinthomson, line 269: See previous comment about minimizing knowledge of events in this module.  This can just pass out the token.

neqo-transport/src/connection/mod.rs
```diff
@@ -2108,6 +2132,7 @@ impl Connection {
         self.crypto.install_application_keys(now + pto)?;
         self.process_tps()?;
         self.set_state(State::Connected);
+        self.create_resumption_token(now + pto);
```
martinthomson, line 2135: You shouldn't be adding PTO here.  Chances are that we don't have any resumption tokens, so this is unlikely to do anything, but it's still the wrong time to pass in (the above should probably be changed to `(now, pto)`...

neqo-transport/src/connection/mod.rs
```diff
@@ -742,6 +760,10 @@ impl Connection {
         let lost = self.loss_recovery.timeout(now);
         self.handle_lost_packets(&lost);
         qlog::packets_lost(&mut self.qlog, &lost);
+
+        if self.release_resumption_token_timer.is_some() {
+            self.create_resumption_token(now);
+        }
```
martinthomson, line 766: Do we need to add this to `next_delay()`?  If we don't then it's worth explicitly noting that and the reason why.  We should ensure that all of our timers are accounted for in both places.

neqo-transport/src/connection/tests/mod.rs
```diff
@@ -323,3 +327,19 @@ fn split_datagram(d: &Datagram) -> (Datagram, Option<Datagram>) {
         b.map(|b| Datagram::new(d.source(), d.destination(), b)),
     )
 }
+
+fn get_tokens(client: &mut Connection, now: Option<Instant>) -> Vec<ResumptionToken> {
+    if let Some(t) = now {
+        let _ = client.process_output(t);
+    }
```
martinthomson, line 334: Could we make the Instant argument mandatory AND check that process_output does not produce a datagram?  I'm thinking that this is almost always going to generate an ACK: receiving a session ticket will often be the first packet received by a client after HANDSHAKE_DONE.  connect_force_idle forces that to be acknowledged, but if we aren't using that, the client will want to send an ACK.

If we can't ensure that there is no packet generated, we should log any that are dropped as a result so that there are no surprises.

neqo-transport/tests/server.rs
```diff
@@ -284,7 +284,16 @@ fn get_ticket(server: &mut Server) -> ResumptionToken {
 
     // Calling active_connections clears the set of active connections.
     assert_eq!(server.active_connections().len(), 1);
-    client.resumption_token().unwrap()
+    client
+        .events()
+        .fold(None, |res, e| {
```
martinthomson, line 289: Using `fold` here rather than find_map() is a little non-intuitive.  Is there any reason why you need the last rather than the first?

martinthomson (changes requested):
This is pretty good.  Don't let the comments convince you otherwise :)

It took me a while to understand why all your tests had timeouts in them.

I think that we can avoid this by having the test server configuration call `Connection::set_validation()`.  The challenge is that this is a weak pointer to something that we don't have a place to hold.  For that, I propose a thread local.  You'll have to do that in two places though as this code is duplicated in both the fixture and the neqo-transport::connection tests.  You might be able to hold the thread local in just the fixture and reuse it in the connection tests.

Then you can back out all the changes with timers in them.

So that the timer is tested, I would then add a test that overwrites the address validation, explicitly drops the reference to the original copy, and then you can test that the resumption token is withheld.

neqo-transport/tests/server.rs
ddragana: Sorry, this was needed in the previous patch (in my old draft version I needed to get the last ticket) and I just copied it.

neqo-transport/src/connection/mod.rs
ddragana: I did not meant to add pto. ... may I did copy-paste a wrong line...

neqo-transport/src/connection/mod.rs
ddragana: I did not add it to next_delay because I did not want to trigger a process() call only for a token, but this will not be trigger that often so it is ok to add it to next_delay().

neqo-transport/src/connection/mod.rs
```diff
@@ -497,52 +499,66 @@ impl Connection {
...
+        );
 
-        // Create reumpion tokens:
-        //  1) create only one if we have not receiveed a new_token,
-        //  2) otherwise create as many as possible that contain ticket and token.
-        loop {
-            self.crypto.create_resumption_token(
-                &mut self.new_token,
+        while self.crypto.has_resumption_token() && self.new_token.has_token() {
+            if let Some(token) = self.crypto.create_resumption_token(
```
martinthomson, line 514: I would just use unwrap() or expect() for this case.

neqo-transport/src/connection/mod.rs
```diff
@@ -497,52 +499,66 @@ impl Connection {
...
-            // If we have more tickets wait for more tokens.
-            // We do not remember when tickets have arrived, therefore we release tokens
-            // without new_token slowly the most one every 3 * PTO.
-            self.release_resumption_token_timer =
-                Some(now + 3 * self.loss_recovery.pto_raw(PNSpace::ApplicationData));
-        } else {
-            self.release_resumption_token_timer = None;
+            let arm = if let Some(expiration_time) = self.release_resumption_token_timer {
+                if expiration_time <= now {
+                    if let Some(token) = self.crypto.create_resumption_token(
```
martinthomson, line 533: As before, unwrap() is better.

neqo-transport/src/connection/mod.rs
```diff
@@ -820,6 +836,11 @@ impl Connection {
             }
         }
 
+        if let Some(t) = self.release_resumption_token_timer {
+            qtrace!([self], "Resumption token event timer {:?}", t);
+            delays.push(t);
+        }
```
martinthomson, line 842: So I think I might not have been clear.  I agree with you about not adding this to the set of delays here.  But, I think that it is worth adding a comment here explaining that you aren't considering that timer as it is not important enough to force the application to set a timer for (and you expect other activity to drive it).

It's not that this needs to be considered, it's that we need to establish a pattern whereby you update both process_timeout and next_delay together when adding new timers.

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -107,16 +107,27 @@ fn two_tickets() {
...
-    // Increase the current time with a large number to trigger
-    // the resumption_token_timer before calling get_tokens.
-    let mut now = now() + Duration::from_millis(140);
-    let token1 = get_tokens(&mut client, Some(now)).pop().unwrap();
-    now += Duration::from_millis(140);
-    let token2 = get_tokens(&mut client, Some(now)).pop().unwrap();
-    // There is no more tokens.
-    now += Duration::from_millis(140);
-    assert_eq!(get_tokens(&mut client, Some(now)).len(), 0);
+    // process() will return a ack first
```
martinthomson, line 110:
```suggestion
    // process() will return an ack first
```

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -107,16 +107,27 @@ fn two_tickets() {
...
+    assert!(client.process(Some(pkt), now()).dgram().is_some());
+    // We do not have a ResumptionToken evennt yet, because NEW_TOKEN was not sent.
```
martinthomson, line 112:
```suggestion
    // We do not have a ResumptionToken event yet, because NEW_TOKEN was not sent.
```

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -107,16 +107,27 @@ fn two_tickets() {
...
+    assert_eq!(get_tokens(&mut client).len(), 0);
+    // process() will return a Callback that has a value of the resumption token timer.
+    let callback = client.process(None, now());
```
martinthomson, line 115: This would seem to depend on the resumption token timer.  If you would prefer to keep the extra timer to keep testing easy, then it's not bad that you have the extra code there.

(We should use an iterator/generator though...)

martinthomson (commented):
LGTM.  Think about what you want to do with the next_delay() code (I think what you have here is fine if it makes testing easier) and land it.

PR #961, opened 2020-09-16
Update to draft-30

ddragana:


neqo-transport/src/packet/mod.rs
```diff
@@ -100,13 +101,14 @@ impl QuicVersion {
...
             Self::Draft29 => 0xff00_0000 + 29,
+            Self::Draft30 => 0xff00_0000 + 30,
         }
     }
 }
 
 impl Default for QuicVersion {
     fn default() -> Self {
-        Self::Draft29
+        Self::Draft30
```
martinthomson, line 111: See meta-comment.

```suggestion
        Self::Draft29
```

neqo-client/src/main.rs
```diff
@@ -897,7 +898,8 @@ mod old {
         let (quic_protocol, alpn) = match args.alpn.as_str() {
             "hq-27" => (QuicVersion::Draft27, "hq-27"),
             "hq-28" => (QuicVersion::Draft28, "hq-28"),
-            _ => (QuicVersion::Draft29, "hq-29"),
+            "hq-29" => (QuicVersion::Draft28, "hq-29"),
+            _ => (QuicVersion::Draft30, "hq-30"),
```
martinthomson, line 902:
```suggestion
            "hq-30" => (QuicVersion::Draft30, "hq-30"),
            _ => (QuicVersion::Draft29, "hq-29"),
```

Note that you had a typo here on -29 previously.

neqo-client/src/main.rs
```diff
@@ -76,7 +76,7 @@ type Res<T> = Result<T, ClientError>;
     about = "A basic QUIC HTTP/0.9 and HTTP/3 client."
 )]
 pub struct Args {
-    #[structopt(short = "a", long, default_value = "h3-29")]
+    #[structopt(short = "a", long, default_value = "h3-30")]
```
martinthomson, line 79:
```suggestion
    #[structopt(short = "a", long, default_value = "h3-29")]
```

martinthomson (commented):
Based on what I'm seeing from other implementations, I would prefer to just add -30 and not change the default.  I've made suggestions that I think capture all of that.

Other than that, this is good to go.

PR #963, opened 2020-09-22
Update QNS

martinthomson:
This has a few changes that I tested together, mostly because I'm lazy.  Let me know if you would prefer these broken into separate pull requests:

1. Update QNS image so that it reads certificates from /certs (otherwise, we won't work with Chrome).

2. Factor the QNS image a little differently, update the runner script.  Some of this was necessary for the previous step, but I went a little further.

3. Pass QNS parameters (TESTCASE and QLOGDIR) to neqo-client and neqo-server via the command line rather than pull them from the environment.

4. Enable the "retry" (S) testcase in our server.  We currently claim to support this, but we don't.  It's a few lines of code.

I've tested a few different test cases in the interop runner between neqo and quant in both roles and this seems to work fine.


ddragana (approved):
Looks good.

PR #964, opened 2020-09-23
Make the congestion control configurable

### LINKED ISSUE #443: Configurable congestion control
It would be nice to be able to experiment with, or have the client define, the CC algorithm to be used. This would also help us do ongoing work on more modern CC algos without causing disruption in the default CC.

ddragana:
fixes #443


neqo-transport/src/cc/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+pub enum CongestionControlAlgorithm {
+    NewReno,
+}
+
+impl CongestionControlAlgorithm {
+    pub fn create(&self) -> Box<dyn CongestionControl> {
+        match self {
+            Self::NewReno => Box::new(NewReno::default()),
+        }
+    }
```
martinthomson, line 26: I don't think that it makes sense to structure the creation like this.  Instead, I think that the thing that holds a congestion controller holds `Box<dyn CongestionControl>`, but you do `Box::new(NewReno::new())` to initialize that variable as necessary.

Under what circumstances do you think that we need to have an enum that identifies the type of congestion controller?  The NewReno Debug implementation can say "NewReno" so that logging shows what is going on.  Command-line tools can build their own mappings so that we can pick the CC variant.

neqo-transport/src/cc/mod.rs
```diff
@@ -0,0 +1,33 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+// Congestion control
+#![deny(clippy::pedantic)]
+
+pub mod cc;
```
martinthomson, line 10: I think that you should move the interface definition into this file.   Otherwise we have neqo_transport::cc::cc::CongestionControl as the canonical name for the interface, which is a little weird.

martinthomson (commented):
This is good, just a question about the glue piece.

Separately, I have had the idea that we stop having loss recovery own the congestion controller.  I want to move it to the path and pass it to the loss recovery code as needed.  That should help us with any eventual need to do migration.

neqo-transport/src/cc/mod.rs
ddragana: I want to be able to change congestion control at run time not at compile time. The piece that is missing is API, i.e. Connection::new(..) will have a parameter CongestionControlAlgorithm as well.

I did not want to expose NewReno outside of cc therefore I made this function, but I do not have a strong opinion about this. I will add cubic(because it is easy to add) and the API today and we can discuss about this function whether it make sense or not.

martinthomson: This is probably fine then.  One thing is that Connection::new() shouldn't take a CC instance, but a CC factory, because eventually we'll want a new CC for a new path.

neqo-transport/src/cc/new_reno_cc.rs
```diff
@@ -286,6 +235,83 @@ impl CongestionControl {
...
+            Some(
+                self.pacer
+                    .as_ref()
+                    .unwrap()
+                    .next(rtt, self.congestion_window),
+            )
+        } else {
+            None
+        }
+    }
```
martinthomson, line 259: I think that ultimately we want the pacer to be separate from the CC.  But that can be done later.

neqo-transport/src/cc/cc.rs
```diff
@@ -0,0 +1,49 @@
...
+        lost_packets: &[SentPacket],
+    );
+
+    fn discard(&mut self, pkt: &SentPacket);
+
+    fn on_packet_sent(&mut self, pkt: &SentPacket, rtt: Duration);
+
+    fn start_pacer(&mut self, now: Instant);
+
+    fn next_paced(&self, rtt: Duration) -> Option<Instant>;
```
martinthomson, line 48: As below, I think that we ultimately want a pacer that is independent of the congestion controller.

neqo-transport/src/cc/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+            Self::NewReno => Box::new(NewReno::default()),
+        }
+    }
+}
+
+impl Default for CongestionControlAlgorithm {
+    fn default() -> Self {
+        Self::NewReno
+    }
+}
```
martinthomson, line 33: I don't think that we need a Default implementation for this; indeed, I think that we probably want to drop the default LossRecovery constructor as well on the assumption that it will be provided with a CC instance as needed (or, for now, the type).

martinthomson (commented):
I think that we should get this in soon.  I've another pull request coming.

I'd like to unwind the ::cc::cc::cc thing and I don't think that we can drop the Default implementations, but the shape of this is OK.

neqo-transport/src/cc/mod.rs
ddragana: I think I will do this when cubic lands or as part of that PR.

ddragana:
I will do all refactoring of cc in this PR and only add Cubic algorithm in a separate PR. I will address some comments from #980

martinthomson (commented):
This is bigger than I anticipated.  Not necessarily in a bad way.  OK if this takes a few days?

ddragana:
> This is bigger than I anticipated. Not necessarily in a bad way. OK if this takes a few days?

not at all, I have other stuff o do for http3/quic.

neqo-transport/src/recovery.rs
```diff
@@ -590,36 +591,55 @@ impl PtoState {
...
+            spaces: LossRecoverySpaces::default(),
+            qlog: NeqoQlog::default(),
+            stats: StatsCell::default(),
+        }
+    }
+}
+
 impl LossRecovery {
-    pub fn new(stats: StatsCell) -> Self {
+    pub fn new_with_stats(alg: CongestionControlAlgorithm, stats: StatsCell) -> Self {
```
martinthomson, line 623: There is no `new` with no stats, so I don't think that you need to rename this one.

neqo-transport/src/recovery.rs
```diff
@@ -590,36 +591,55 @@ impl PtoState {
...
+            confirmed_time: None,
+            pto_state: None,
+            rtt_vals: RttVals::default(),
+            packet_sender: PacketSender::new(CongestionControlAlgorithm::default()),
+            spaces: LossRecoverySpaces::default(),
+            qlog: NeqoQlog::default(),
+            stats: StatsCell::default(),
+        }
+    }
+}
```
martinthomson, line 620: As I said previously, I don't think that we need a Default implementation for LossRecovery, or CongestionAlgorithm.  The latter is part of the constructor to Connection and we should be able to make a decision.  All of those `::default()` calls can be replaced by `::NewReno`.

neqo-transport/src/cc/new_reno.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl Default for NewReno {
+    fn default() -> Self {
+        Self {}
+    }
+}
+
+impl Display for NewReno {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "NewReno",)?;
```
martinthomson, line 26: autoformat for macros is poor:

```suggestion
        write!(f, "NewReno")?;
```

neqo-transport/src/cc/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub const MAX_DATAGRAM_SIZE: usize = PATH_MTU_V6;
+
+pub trait CongestionControl: Display + Debug {
+    fn set_qlog(&mut self, qlog: NeqoQlog);
+
+    fn cwnd(&self) -> usize;
+
+    #[cfg(test)]
+    fn ssthresh(&self) -> usize;
```
martinthomson, line 31: Is this something we can leave on the specific implementation?  That is, this is an internal detail that we might want to look at, but do we need to have it generically?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -73,98 +72,59 @@ impl State {
 }
 
 #[derive(Debug)]
-pub struct CongestionControl {
+pub struct ClassicCongestionControl {
+    cc_algorithm: Box<dyn CwndFn>,
```
martinthomson, line 76: I think that you can template this one.  That is `ClassCongestionControl<T>` and condition the implementation on `T:CwndFn`.

neqo-transport/src/cc/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    fn discard(&mut self, pkt: &SentPacket);
+
+    fn on_packet_sent(&mut self, pkt: &SentPacket);
+}
+
+trait CwndFn: Display + Debug {
+    fn on_packets_acked(&mut self, curr_cwnd: usize, acked_bytes: usize) -> (usize, usize);
+    fn on_congestion_event(&mut self, curr_cwnd: usize, acked_bytes: usize) -> (usize, usize);
+}
```
martinthomson, line 57: This is only needed for the ClassicCC, right?  I think that you can move it there.  And maybe call it WindowAdjustment or something like that.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -342,6 +247,100 @@ impl CongestionControl {
...
 
+    /// Whether a packet can be sent immediately as a result of entering recovery.
+    #[must_use]
+    fn recovery_packet(&self) -> bool {
+        self.state == State::RecoveryStart
+    }
+}
+
+impl ClassicCongestionControl {
+    pub fn new(alg: CongestionControlAlgorithm) -> Self {
```
martinthomson, line 258: If you template based on the algorithm, then you don't need to match anywhere here (the matching happens outside).  You might need a static `::new` function on the trait though, I think.

neqo-transport/src/recovery.rs
ddragana: This default is used only by tests. I can remove the default.

ddragana: To be precise the default for LossRecovery.

neqo-transport/src/cc/mod.rs
ddragana: This was used by some tests. This should not be exposed. I will remove it.

neqo-transport/src/recovery.rs
ddragana: I made CongestionAlgorithm implements Default to make it easier to update to a new cc without changing cc everywhere.

neqo-transport/src/recovery.rs
```diff
@@ -605,21 +606,34 @@ pub(crate) struct LossRecovery {
 }
 
 impl LossRecovery {
-    pub fn new(stats: StatsCell) -> Self {
+    #[cfg(test)]
+    pub fn new(alg: &CongestionControlAlgorithm) -> Self {
```
martinthomson, line 610: Is it so hard for the tests to provide an extra `StatsCell::default()` argument?  I'm concerned that this function and the next now need to be synchronized, all to save a tiny bit of typing (or copy-paste) in tests.

neqo-transport/src/tracking.rs
```diff
@@ -6,7 +6,7 @@
 
 // Tracking of received packets and generating acks thereof.
 
-#![deny(clippy::pedantic)]
+//#![deny(clippy::pedantic)]
```
martinthomson, line 9: ?

martinthomson (commented):
I like this.  I have a few minor niggles, but you should land once you have decided what to do with each.

neqo-transport/src/recovery.rs
ddragana: I will remove new_with_stats.

PR #970, opened 2020-09-29
Add key update for network simulator

martinthomson:
I haven't been able to get a copy of wireshark that supports the latest version, so I need to test this.

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -363,6 +363,12 @@ impl Debug for Encoder {
     }
 }
 
+impl AsRef<[u8]> for Encoder {
+    fn as_ref(&self) -> &[u8] {
+        self.buf.as_ref()
+    }
+}
+
 impl<'a> From<Decoder<'a>> for Encoder {
     #[must_use]
     fn from(dec: Decoder<'a>) -> Self {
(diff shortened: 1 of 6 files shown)


martinthomson:
I've confirmed that this results in a key update and the test almost runs perfectly.  Installing the right version of wireshark helped a lot, even if that took an unnecessarily large amount of effort.

The reason that the test fails is a curious one.  The test insists on packets being sent with two key phases.  This is reasonable, but annoying.  We only get one request to make, and it appears as though our stack does not immediately send SETTINGS when the connection is established.  I think that it is probably worth fixing that problem to have the test pass.

martinthomson:
Marten has graciously updated the test conditions to allow for this one case.  As it turns out, you can't require either client or server to send with key phase 0.

neqo-common/src/codec.rs
```diff
@@ -363,6 +363,12 @@ impl Debug for Encoder {
     }
 }
 
+impl AsRef<[u8]> for Encoder {
```
ddragana, line 366: do we need this for this PR?

ddragana (approved):

neqo-common/src/codec.rs
martinthomson: No, I mixed the streams, but it's not a big deal if it is here.

PR #975, opened 2020-10-01
Send a packet on entering recovery

### LINKED ISSUE #938: Send a packet immediately after entering recovery
The spec permits an endpoint to send just one packet in excess of the newly reduced congestion window when entering recovery.  The way this works in QUIC is subtly different than in TCP, but there is still a period where sending halts.  This one packet can ensure that the identified loss is repaired without waiting for that dead period to end.

We probably can engage our PTO logic for this so that we don't need a ton of extra code, but it will need something extra in the congestion controller side of things to support this.

martinthomson:
This uses the PTO send profile and probes in every active packet number space immediately after entering recovery.  This is intended to ensure that we exit recovery as quickly as possible as an acknowledgment for this packet will cause an exit to recovery.

This factors some of the recovery code to add an explicit state rather than add another flag.  I realized that the recovery start time couldn't be eliminated after running some tests, but I tried to clean that up to.

I exploited the extra flag to remove an argument from a function.

In debugging this, I realized that some of our existing tests were unrealistic.  So I tweaked those a little, mostly just to add an RTT. Those changes can be considered separately.

In terms of performance, in the worst case, this might improve things a bit.  Running this test:

```
cargo test -p neqo-transport --test network transfer_fixed_seed -- --nocapture
```

On this branch takes a simulated time of 15.952991493s.  On main it takes16.581412134s.  This is very intentionally a pathological test with terrible conditions, including unrealistic amounts of reordering, so it's not really representative.

Other tests have less favourable comparisons, but there is a lot of noise.  We send slightly more packets in some tests, which sometimes helps, sometimes not.  I've seen 2s of variation on a 7s test, so it's not clear what, if any, effect this truly has.  It very much depends on the seed chosen apparently.  I don't have the tools to get any sort of statistical testing yet and I'm not sure that the tests we have are good enough to say that running that test would produce useful information.

Closes #938.

### DIFF
--- a/neqo-transport/src/cc.rs
+++ b/neqo-transport/src/cc.rs
@@ -13,7 +13,7 @@ use std::time::{Duration, Instant};
 
 use crate::pace::Pacer;
 use crate::path::PATH_MTU_V6;
-use crate::qlog::{self, CongestionState, QlogMetric};
+use crate::qlog::{self, QlogMetric};
 use crate::tracking::SentPacket;
 use neqo_common::{const_max, const_min, qdebug, qinfo, qlog::NeqoQlog, qtrace};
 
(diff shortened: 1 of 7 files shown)


ddragana (commented):
The refactoring looks very reasonable.
Sending packet when recovery starts should not hurt.

Please land it and I will rebase my PRs

PR #977, opened 2020-10-01
Send a session ticket when doing HTTP/0.9

martinthomson:
This might help with passing a few more QNS tests.

Haven't tested this, but it seems useful enough to just try it out.


ddragana (approved):

ddragana:
Let's try it out. I is a small change.

PR #980, opened 2020-10-01
Cubic

ddragana:


ddragana:
I wanted to marked this as draft because it is not done yet, but github UI somehow does not work.

This is implementation is something between the RFC and linux implementation. 

@martinthomson do you know a crate for converting between f64 and usize?

ddragana:
It did converted it to draft.

martinthomson (commented):
Only looking at the structural stuff for the moment...

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,205 @@
...
+        now: Instant,
+        rtt: Duration,
+        acked_bytes: &mut usize,
+        newly_acked_bytes: usize,
+    ) {
+        let curr_cwnd = *congestion_window as f64;
+        self.last_phase_was_tcp = false;
+        let curr_cwnd_f64 = *congestion_window as f64;
+        if self.ca_epoch_start.is_none() {
+            // This is the first time the congestion avoidance has been entered.
```
martinthomson, line 84: "first time" implies that this happens only once, but this is just the start of the epoch, which happens every time we exit recovery.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,205 @@
...
+            self.last_max_cwnd, self.k, self.w_max, self.ca_epoch_start
+        )?;
+        Ok(())
+    }
+}
+
+impl Cubic {
+    // W_cubic(t) = C*(t-K)^3 + W_max (Eq. 1)
+    fn w_cubic(&self, t: f64) -> f64 {
+        CUBIC_C * (t - self.k).powi(3) * MAX_DATAGRAM_SIZE_F64 + self.w_max
```
martinthomson, line 64: What are the units of this?  If our value of W_max is counted in bytes, why then do you multiply out by MAX_DATAGRAM_SIZE again later?

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        self.estimated_tcp_cwnd +=
+            *acked_bytes as f64 * CUBIC_ALFA * MAX_DATAGRAM_SIZE_F64 / self.estimated_tcp_cwnd;
+
+        let time_ca = (now + rtt - self.ca_epoch_start.unwrap()).as_secs_f64();
+
+        let w_cubic = self.w_cubic(time_ca);
+        if w_cubic > curr_cwnd_f64 {
+            let inc = ((w_cubic - curr_cwnd_f64) / curr_cwnd_f64 * MAX_DATAGRAM_SIZE_F64) as usize;
```
martinthomson, line 106: You already multiple by MAX_DATAGRAM_SIZE in w_cubic.

martinthomson (commented):
I'm really struggling with this one now that I've started to look at Cubic more closely.  Why are all the congestion control specifications so bad?

As far as I can tell, you are implementing something else, but I'm having trouble following the places that you deviate from spec.

ddragana:
This is really really a draft. I have posted it just to show how newReno and Cubic will leave together.
I am planning to explain what I have implemented. It is not following the spec completely  (I am not really trusting the spec because the spec was actually never implemented, it was suppose to document what implementations have implemented but it is not doing that). This implementation is lean on google's implementation, but it is not correct yet and I still need to debug it and fix it.
This is not a high priority, I am planning to get back to it soon.

ddragana:
Replaced with #1074

PR #981, opened 2020-10-02
Don't skip secrets on key update

### LINKED ISSUE #979: It's possible our key update doesn't work
This needs to be tested more thoroughly, but I haven't managed to get a single interop with it yet.

martinthomson:
This was nasty.  Each set of application data read/write keys holds a
secret.  This is not the secret that is currently in use (secret N,
which is thrown away), but the next one (secret N+1).  When we update,
we should use that secret (N+1) and then create the next one after that
(secret N+2).

We were using secret N+2 and keeping it too.  That meant that we were
effectively skipping the second set of secrets (those for the first
switch to key phase 1) entirely.

Simple fix, but impossible to test without interoperating with other
implementations.  Which I have now done.

Closes #979.

### DIFF
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -584,7 +584,7 @@ impl CryptoDxAppData {
         }
         let next_secret = Self::update_secret(self.cipher, &self.next_secret)?;
         Ok(Self {
-            dx: self.dx.next(&next_secret, self.cipher),
+            dx: self.dx.next(&self.next_secret, self.cipher),
             cipher: self.cipher,
             next_secret,
         })


neqo-transport/src/crypto.rs
```diff
@@ -584,7 +584,7 @@ impl CryptoDxAppData {
         }
         let next_secret = Self::update_secret(self.cipher, &self.next_secret)?;
         Ok(Self {
-            dx: self.dx.next(&next_secret, self.cipher),
+            dx: self.dx.next(&self.next_secret, self.cipher),
```
martinthomson, line 587: I keep asking myself whether Rust's semantics could have helped here at all.  The answer is "maybe".  If we passed by value we would have been forced to clone rather than reference secrets.  That might have helped.

ddragana (approved):

PR #982, opened 2020-10-02
Add AEAD limits

martinthomson:
This uses new numbers.

Testing this will be challenging.  I've got to do that though, so this is not done yet.


neqo-transport/src/crypto.rs
```diff
@@ -306,8 +319,11 @@ pub struct CryptoDxState {
     /// for verifying that packet numbers before a key update are strictly lower
     /// than packet numbers after a key update.
     used_pn: Range<PacketNumber>,
-    /// This is the minimum allowed.
+    /// This is the minimum packet numbers that is allowed.
```
martinthomson, line 322:
```suggestion
    /// This is the minimum packet number that is allowed.
```

ddragana (approved):

Issue #973, opened 2020-09-30
We have too many DataReadable events


ddragana:
I was looking at some Firefox logs and we have too many DataReadable events. They are received even after a stream has been closed.

I seen this on blog.cloudflare.com, but Io have only necko logging. I will try to reproduce it with neqo logging. should not be hard.

ddragana:
I looked at this. neqo-transport is posting too many events.
A long time ago Andy and I discussed this and the events should be edge triggered. In this case they are not. The neqo-http3 code and m.-c. work with premises that the evens are edge triggered. 
The code can tolerate too many events but still we execute code that we do not need to execute. The events for stream are sent for each QUIC packet that contains data from that stream and we should fix it.
The issue is in neqo-transport code. 
I will write a test and a patch for this.

PR #987, opened 2020-10-08
Use a generic event provider

### LINKED ISSUE #934: Provide an `Iterator` implementation for events
I've been wondering whether we should provide an `Iterator` implementation for events so that we can do these things more efficiently.  The code you have here is roughly:

```rust
client.events().filter(|e| matches!(e, Http3ClientEvent::HeaderReady { .. })).count();
```

We could get that by doing:

```rust
trait ConnectionEventProvider {
    fn next_event(&mut self) -> Option<ConnectionEvent>;
}

struct<'a, T> ConnectionEventIter<T> {
    c: &'a T;
}
    
impl<'a, T> Iterator for ConnectionEventIter<'a, T> 
    where T: ConnectionEventProvider {
    type Item = ConnectionEvent;
    fn next(&mut self) -> Option<Self::Item> {
        self.c.next_event()
    }
}
```

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/926_

martinthomson:
This doesn't really reduce code size much, but it unifies the interface
somewhat.  It also makes the iterator cleaner for those cases where it
is used.  As it now takes a mutable borrow, it is also safer to use.
Previously, the events that were retrieved could be out of date.

I did run into #973 when working on this.  `Iterator::any` stops when it
hits the first success and there were multiple DataReadable events
queued.  I've flagged that in the code.  It should be easy to fix up
when it comes time (what is hard is getting the function signatures
right; `Iterator::any` takes a value, `Iterator::filter` takes a
reference.

Closes #934.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -9,7 +9,7 @@
 
 use qlog::QlogStreamer;
 
-use neqo_common::{self as common, hex, qlog::NeqoQlog, Datagram, Role};
+use neqo_common::{self as common, event::Provider, hex, qlog::NeqoQlog, Datagram, Role};
 use neqo_crypto::{
     constants::{TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256},
     init, AuthenticationStatus, Cipher, ResumptionToken,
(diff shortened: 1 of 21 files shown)


ddragana (approved):
This looks better than it use to be.

PR #989, opened 2020-10-08
Cleanup streams before granting stream credit

### LINKED ISSUE #871: Don't give client more bidi streams until both sides are done
As soon as the client sends a request on a bidi stream, we are increasing MAX_STREAMS, even though we have not sent back response data. Instead, don't give them another one until both send and recv streams are in a terminal state.

### LINKED ISSUE #988: Receiving an ACK for stream end doesn't persist
An acknowledgment for the end of a stream only works if we have also received acknowledgments for all of the stream.  If the stream was written into two packets and both are acknowledged at once, we will send the end of stream again.

This is only problematic because we will send a stream end more than necessary, but it is worth fixing.

It interfered with a test that I wanted to write for #871.  A fix is bundled with that change.

martinthomson:
Andy's original comment:

Currently, the client can have open streams exceeding MAX_STREAMS
because we give them more streams credit when the recv stream is
complete rather than when both the send and recv stream are complete.

Clear up terminal send streams first. Then, only remove bidi recv
streams if send stream is not present.

This keeps the recv stream hanging around longer than strictly necessary
but I think this is more straightforward than logic to remove it, and
then increase max_stream_bidi when the send stream is cleaned up.

My additions:

I wrote a different test for this, and I'm really glad that I did.  It
turned up a different bug.  If SendStream received an ACK for the end of
a stream before it received an ACK for all the data, it would not
consider the stream to be acknowledged, and so it wouldn't clean itself
up.  As this is a very common situation that results from
send>packet>close>packet, I think that it is worth fixing.

I also ensured that if the FIN is lost after being acknowledged, that
wouldn't cause us to send an extra frame.

And I moved the stream cleanup so that it runs when you run
process_input or process_output, but it doesn't run twice when you run
process.  For process and process_output, it is now with the other
housekeeping stuff in process_timer; process_input calls it specially.

Closes #871.
Closes #988.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -1470,11 +1470,11 @@ mod tests {
         s.close();
         s.mark_as_sent(len_u64, 0, true);
 
-        // Ack the fin, then the data.
+        // Ack the fin, then mark it lost.
         s.mark_as_acked(len_u64, 0, true);
-        s.mark_as_lost(len_u64, 0, true); // Should be a noop.
+        s.mark_as_lost(len_u64, 0, true);
 
-        // Sending should result in no new frame being sent here.
+        // No frame should be sent here.
         let mut builder = SendStreams(IndexMap::default());
         builder.insert(id, s);
         assert!(builder.get_frame(PNSpace::ApplicationData, 1000).is_none());
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -394,6 +394,7 @@ enum SendStreamState {
         send_buf: TxBuffer,
         final_size: u64,
         fin_sent: bool,
+        fin_acked: bool,
     },
     DataRecvd {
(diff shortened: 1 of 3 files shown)


neqo-transport/src/send_stream.rs
```diff
@@ -1409,4 +1426,57 @@ mod tests {
...
+        let mut s = SendStream::new(id, 0, Rc::clone(&flow_mgr), conn_events);
+        s.set_max_stream_data(len_u64);
+
+        // Send all the data, then the fin.
+        let _ = s.send(MESSAGE).unwrap();
+        s.mark_as_sent(0, MESSAGE.len(), false);
+        s.close();
+        s.mark_as_sent(len_u64, 0, true);
+
+        // Ack the fin, then the data.
```
ddragana, line 1473: Fixed the comment.

ddragana (commented):
Looks good. Fix the comment and land it.

PR #991, opened 2020-10-09
Resend Initial in response to undecryptable packet

### LINKED ISSUE #990: Resend Initial if we get Handshake
The spec suggests that this helps sometimes.  We might also send again in response to ANY packet.

In both cases, we need restrictions:

* We should only send if the connection ID matches and for the first "packet" in a datagram, as potential padding of server packets would accidentally trigger this logic if we didn't.

* We should count the number of times we do this and stop.  Otherwise, we could be forced into a fast loop until we give up (and we don't give up necessarily, depending on the failure mode).

martinthomson:
This should help make our handshake faster.

I found that the code for handling packets was getting a little large,
so I factored it down a little.

Closes #990.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -71,10 +71,6 @@ struct Packet(Vec<u8>);
 pub const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
 pub const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
 
-/// The number of Initial packets that the client will send in response
-/// to receiving an undecryptable packet during the early part of the
-/// handshake.  This is a hack, but a useful one.
-const EXTRA_INITIALS: usize = 4;
 const LOCAL_MAX_DATA: u64 = 0x3FFF_FFFF_FFFF_FFFF; // 2^62-1
 
 #[derive(Debug, PartialEq, Eq)]
@@ -1131,7 +1127,7 @@ impl Connection {
                 // when there is a short RTT and losses.
                 if first
                     && self.is_valid_cid(packet.dcid())
-                    && self.stats.borrow().saved_datagrams <= EXTRA_INITIALS
+                    && self.loss_recovery.take_extra_initial()
                 {
                     self.crypto.resend_unacked(PNSpace::Initial);
                 }
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -71,6 +71,10 @@ struct Packet(Vec<u8>);
 pub const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
 pub const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
 
+/// The number of Initial packets that the client will send in response
+/// to receiving an undecryptable packet during the early part of the
+/// handshake.  This is a hack, but a useful one.
+const EXTRA_INITIALS: usize = 4;
 const LOCAL_MAX_DATA: u64 = 0x3FFF_FFFF_FFFF_FFFF; // 2^62-1
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1048,10 +1062,135 @@ impl Connection {
...
 
+    /// Perform any processing that we might have to do on packets prior to
+    /// attempting to remove protection.
+    fn preprocess(
+        &mut self,
+        packet: &PublicPacket,
+        first: bool,
+        now: Instant,
+    ) -> Res<PreprocessResult> {
+        match (packet.packet_type(), &self.state, &self.role) {
```
martinthomson, line 1073: This is almost a perfect copy, except for the return codes, which are converted from the values used at the call site and the tweaks I've highlighted below.

neqo-transport/src/connection/mod.rs
```diff
@@ -1048,10 +1062,135 @@ impl Connection {
...
+                        packet.dcid().to_vec(),
+                    )
+                }
+            }
+            (PacketType::VersionNegotiation, State::WaitInitial, Role::Client) => {
+                match packet.supported_versions() {
+                    Ok(versions) => {
+                        if versions.is_empty() || versions.contains(&self.quic_version.as_u32()) {
+                            // Ignore VersionNegotiation packets that contain the current version.
+                            self.stats.borrow_mut().pkt_dropped("Invalid VN");
```
martinthomson, line 1106: This wasn't using the helper function, so I fixed it here and just below.

neqo-transport/src/connection/mod.rs
```diff
@@ -1048,10 +1062,135 @@ impl Connection {
...
+                // Resend Initial CRYPTO frames immediately a few times just
+                // in case.  As we don't have an RTT estimate yet, this helps
+                // when there is a short RTT and losses.
+                if first
+                    && self.is_valid_cid(packet.dcid())
+                    && self.stats.borrow().saved_datagrams <= EXTRA_INITIALS
+                {
+                    self.crypto.resend_unacked(PNSpace::Initial);
+                }
+            }
```
martinthomson, line 1138: This is the new code.

neqo-transport/src/connection/mod.rs
```diff
@@ -1048,10 +1062,135 @@ impl Connection {
...
+            (PacketType::VersionNegotiation, ..)
```
martinthomson, line 1139: This case is a little whacky.  I was tempted to break it out specially, because you really want to stop processing packets if you get a Version Negotiation packet.  But those consume the entire packet and it should be dropped, so this is benign.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -479,10 +479,11 @@ fn reorder_handshake() {
...
     assert!(s_hs.is_some());
 
-    // Pass just the handshake packet in and the client can't handle it.
+    // Pass just the handshake packet in and the client can't handle it yet.
+    // It can only send another Initial packet.
     now += RTT / 2;
-    let res = client.process(s_hs, now);
-    assert_ne!(res.callback(), Duration::new(0, 0));
+    let dgram = client.process(s_hs, now).dgram();
+    assertions::assert_initial(&dgram.as_ref().unwrap(), false);
```
martinthomson, line 486: This is the one case where the new code actually gets hit.  Good to know we're getting almost enough tests to have something fail in response to every change.

neqo-transport/src/packet/mod.rs
```diff
@@ -572,7 +572,7 @@ impl<'a> PublicPacket<'a> {
         self.quic_version
     }
 
-    pub fn packet_len(&self) -> usize {
+    pub fn len(&self) -> usize {
```
martinthomson, line 575: This one just irked me, so I threw it in.  It's not really needed.

ddragana (approved):

martinthomson:
Taken as part of #992.

PR #994, opened 2020-10-11
Some minor tweaks to error reporting

martinthomson:
Added when trying to diagnose #993. (I still don't understand how that helps, by the way.)


ddragana (approved):
I rhink the commit message is wrong. Do you really mean #993 ?

martinthomson:
Yes, I did.  I don't see how the patch in #993 works.

PR #996, opened 2020-10-12
Do not post multiple unnecessary  DataReadable events

### LINKED ISSUE #973: We have too many DataReadable events

ddragana:
Fixes #973

### DIFF
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -271,8 +271,8 @@ fn do_not_accept_data_after_stop_sending() {
         server.stream_stop_sending(stream_id, Error::NoError.code())
     );
 
-    // Receive the second data frame. The frame should be ignored and now
-    // DataReadable events should be posted.
+    // Receive the second data frame. The frame should be ignored and
+    // DataReadable events shouldn't be posted.
     let out = server.process(out_second_data_frame.dgram(), now());
     assert!(!server.events().any(stream_readable));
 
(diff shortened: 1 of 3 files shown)


ddragana:
There is some code in necko that is executed on each of this events and it is really not needed.

This is bothers me because Firefox is producing to much logging for so many unnecessary events.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -503,3 +503,59 @@ fn max_streams_after_bidi_closed() {
...
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+    client.stream_send(stream_id, &[0x00]).unwrap();
+    let out = client.process(None, now());
+    let _ = server.process(out.dgram(), now());
+
+    // We have a data_readable event.
+    let stream_readable = |e| matches!(e, ConnectionEvent::RecvStreamReadable {..});
+    assert!(server.events().any(stream_readable));
+
+    // Send one more datta frame from client. The previous stream data has not been read yet,
```
martinthomson, line 523:
```suggestion
    // Send one more data frame from client. The previous stream data has not been read yet,
```

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -503,3 +503,59 @@ fn max_streams_after_bidi_closed() {
...
+    // therefore there should not be a new DataReadable event.
+    client.stream_send(stream_id, &[0x00]).unwrap();
+    let out_second_data_frame = client.process(None, now());
+    let _ = server.process(out_second_data_frame.dgram(), now());
+    assert!(!server.events().any(stream_readable));
+
+    // One more frame with a fin will not produce a new DataReadable event, becasue the
```
martinthomson, line 530:
```suggestion
    // One more frame with a fin will not produce a new DataReadable event, because the
```

martinthomson (commented):
Approved.  Thanks for fixing this.

I only found a couple of typos.

PR #1002, opened 2020-10-15
Don't save datagrams with a bad connection ID

martinthomson:
Our funny padding means that we get a lot of packets that look like they
might be 1-RTT packets.  We save that padding.

We probably shouldn't.  A simple way we can avoid that is to check that
the connection ID is right.


martinthomson:
Oh, I think that I can do better than this... Checking connection ID before attempting to decrypt will save a lot of pain.

PR #1005, opened 2020-10-15
Accept interim 1xx HTTP responses

ddragana:


ddragana:
This will make resolving #753 easy, because this pPr adds all proper error handling.

neqo-http3/src/server_events.rs
```diff
@@ -86,7 +86,7 @@ pub enum Http3ServerEvent {
     /// Headers are ready.
     Headers {
         request: ClientRequestStream,
-        headers: Option<Vec<Header>>,
+        headers: Vec<Header>,
```
martinthomson, line 89: 💯

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
         }
     }
+
+    #[test]
+    fn response_w_1xx() {
+        let (mut client, mut server, request_stream_id) = connect_and_send_request(true);
+
+        setup_server_side_encoder(&mut client, &mut server);
+
+        let headers1xx = vec![(String::from(":status"), String::from("101"))];
```
martinthomson, line 5666: 101 might be a bad example.  It's not valid in h3.

Should we reset streams if we get 101?

```suggestion
        let headers1xx = vec![(String::from(":status"), String::from("103"))];
```

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+
+        let encoded_headers = server
```
martinthomson, line 5668:
```suggestion
        let encoded_1xx = server
```

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+        };
+        let mut d = Encoder::default();
+        hframe.encode(&mut d);
+        let headers200 = vec![
+            (String::from(":status"), String::from("200")),
+            (String::from("my-header"), String::from("my-header")),
+            (String::from("content-length"), String::from("3")),
+        ];
+
+        let encoded_headers = server
```
martinthomson, line 5683:
```suggestion
        let encoded_200 = server
```


... the alternative is to define a function that takes a slice of header tuples and a mutable reference to the encoder and encodes those headers into the encoder.  That might be easier to read.

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+            request_stream_id,
+            &d,
+            false,
+        );
+
+        server_send_response_and_exchange_packet(
+            &mut client,
+            &mut server,
+            request_stream_id,
+            &[0x0, 0x0, 0x0],
```
martinthomson, line 5704: You probably need to explain what this means.  It took me a while to understand it.  Isn't this an empty DATA frame and one more byte?  Why don't we choke on that?

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+                    stream_id,
+                    interim,
+                    headers,
+                    ..
+                } = e
+                {
+                    Some((stream_id, interim, headers))
+                } else {
+                    None
+                }
```
martinthomson, line 5740: Both of your closures here are the same.  You can use `filter_map()` instead and use `.next().unwrap()` to get each for testing.  As a bonus, you can add `assert!(iter.next().is_none())` at the end.

```rust
   let events= client.events().filter_map(|e| {
                if let Http3ClientEvent::HeaderReady {
                    stream_id,
                    interim,
                    headers,
                    ..
                } = e
                {
                    Some((stream_id, interim, headers))
                } else {
                    None
                }
);
  let (stream_id_1xx, interim_1xx, headers_1xx) = events.next().unwrap();
assert_eq!((stream_id_1xx, interim_1xx, headers_1xx), (request_stream_id, true, headers_1xx));
  let (stream_id_200, interim_200, headers_200) = events.next().unwrap();
assert_eq!((stream_id_200, interim_200, headers_200), (request_stream_id, false, headers_200));
assert!(events.next().is_none());
```

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+        let hframe = HFrame::Headers {
+            header_block: encoded_headers.to_vec(),
+        };
+        let mut d = Encoder::default();
+        hframe.encode(&mut d);
+        let headers200 = vec![
+            (String::from(":status"), String::from("200")),
+            (String::from("my-header"), String::from("my-header")),
+            (String::from("content-length"), String::from("3")),
+        ];
```
martinthomson, line 5835: See above regarding the encoding of headers...

neqo-http3/src/connection_client.rs
```diff
@@ -5562,4 +5656,311 @@ mod tests {
...
+            })
+            .unwrap();
+        assert_eq!(push_id_200, FIRST_PUSH_ID);
+        assert!(!interim_200);
+        assert_eq!(headers_200, headers200);
+    }
+
+    // Client: receive a push stream
+    #[test]
+    fn push_single_wo_status() {
```
martinthomson, line 5905: Is there another test here?  A PushPromise that doesn't have `:authority` or similar should do the same, right?

neqo-http3/src/lib.rs
```diff
@@ -264,7 +269,14 @@ pub trait RecvStream: Debug {
...
 
 pub(crate) trait RecvMessageEvents: Debug {
-    fn header_ready(&self, stream_id: u64, headers: Option<Vec<Header>>, fin: bool);
+    fn header_ready(&self, stream_id: u64, headers: Vec<Header>, interim: bool, fin: bool);
     fn data_readable(&self, stream_id: u64);
-    fn reset(&self, stream_id: u64, error: AppError);
+    fn reset(&self, stream_id: u64, error: AppError, local: bool);
+}
+
+#[derive(PartialEq)]
```
martinthomson, line 277:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```

neqo-http3/src/recv_message.rs
```diff
@@ -132,32 +135,47 @@ impl RecvMessage {
...
+        decoder: &mut QPackDecoder,
+    ) -> Res<()> {
+        match (fin, self.is_interim(&headers)?) {
+            (true, true) => Err(Error::HttpGeneralProtocolStream),
+            (false, true) => {
+                self.conn_events
+                    .header_ready(self.stream_id, headers, true, fin);
+                self.state = RecvMessageState::WaitingForResponseHeaders {
+                    frame_reader: HFrameReader::new(),
+                };
```
martinthomson, line 151: Is this just a case of needing to reset the value of frame_reader?  Can you comment to that effect?

neqo-http3/src/recv_message.rs
```diff
@@ -132,32 +135,47 @@ impl RecvMessage {
...
-            self.state = RecvMessageState::WaitingForData {
-                frame_reader: HFrameReader::new(),
-            };
+    fn add_headers(
+        &mut self,
+        headers: Vec<Header>,
+        fin: bool,
+        decoder: &mut QPackDecoder,
+    ) -> Res<()> {
+        match (fin, self.is_interim(&headers)?) {
```
martinthomson, line 144: I think that your match here isn't really helping much.  Every arm of this is essentially the same code.  I think that you can do:

```rust
let interim = Self::is_interim(&headers)?;
if fin && interim {
  return Err(Error:HttpGeneralProtocolStream);
}
self.conn_events.header_ready(self.stream_id, headers, true, fin);
self.state = if interim {
RecvMessageState::WaitingForResponseHeaders {
                    frame_reader: HFrameReader::new(),
                }
} else {
 RecvMessageState::WaitingForData {
                    frame_reader: HFrameReader::new(),
                }
};
if fin {
  self.set_closed(decoder);
}

File: neqo-http3/src/recv_message.rs [hunk 12]
```diff
@@ -304,6 +321,23 @@ impl RecvMessage {
             RecvMessageState::ClosePending | RecvMessageState::Closed
         )
     }
+
+    fn is_interim(&self, headers: &[Header]) -> Res<bool> {
+        match self.message_type {
+            MessageType::Response => {
+                let status = headers.iter().find(|e| e.0 == ":status");
+                if let Some(s) = status {
```
martinthomson, line 329: Don't use numeric tuple fields if you can avoid it.

```suggestion
                if let Some((_name, value)) = status {
```

neqo-http3/src/recv_message.rs
```diff
@@ -304,6 +321,23 @@ impl RecvMessage {
...
+                    let status_code =
+                        s.1.parse::<i32>()
```
martinthomson, line 331:
```suggestion
                        value.parse::<i32>()
```

neqo-http3/src/recv_message.rs
```diff
@@ -304,6 +321,23 @@ impl RecvMessage {
             RecvMessageState::ClosePending | RecvMessageState::Closed
         )
     }
+
+    fn is_interim(&self, headers: &[Header]) -> Res<bool> {
+        match self.message_type {
+            MessageType::Response => {
+                let status = headers.iter().find(|e| e.0 == ":status");
```
martinthomson, line 328: This is probably is a case where find_map doesn't make the code more readable, sadly.

```suggestion
                let status = headers.iter().find(|(name, _value)| name == ":status");
```

martinthomson (commented):
I have a few suggestions, but this looks generally OK.

I'm a little concerned that some of the code is starting to get hard to read.  If you get the time, it might pay to spend a bit of time cleaning up receive_internal in recv_message.rs in particular.  A little refactoring can go a long way.

neqo-http3/src/connection_client.rs
ddragana: We should. I will open an issue.

neqo-http3/src/connection_client.rs
ddragana: This data is not important. I will change it to be something that make sense.
This is legal. it is DATA frame of length 0 and another incomplete DATA frame.

neqo-http3/src/connection_client.rs
ddragana: This is still missing #753

Issue #1004, opened 2020-10-15
Listen to IPv6 address as well


ddragana:

martinthomson:
Closed in #1008.

PR #1008, opened 2020-10-15
Listen to IPv6 as well

ddragana:


ddragana:
Fixes #1004

martinthomson (approved):
I find this somehow unsatisfactory, even though I know that this works.  Why does the v4 address listen only on v4 and the v6 address listens on both?

PR #1010, opened 2020-10-16
Write frames directly

martinthomson:
OK, first, I have to apologize for how big this is.  It's largely mechanical, but it doesn't look that way.

The individual commits should be fairly clear, because I did this almost a frame at a time.

Each source of frames now writes directly.  So the code from `Frame::marshal` and related functions are moved.  Rather than making a frame it just writes the data out after working out if there is space.

This is less architecturally pure in the sense that there is no single place where all the frame writing occurs, and it means that we have less direct testing of this code, but we now have plenty of testing elsewhere.

The biggest change here is moving all of the STREAM frame stuff from frame.rs to send_stream.rs.  I had to refactor the code a lot.  Same for the tests, though all the existing tests are still there, just with different fixtures.

(The next step is to make `Frame` borrow buffers rather than taking copies.  That turns out to be less of a big win, but it saves a bit of extra copying in a few narrow cases.)

### DIFF
--- a/neqo-transport/src/addr_valid.rs
+++ b/neqo-transport/src/addr_valid.rs
@@ -13,7 +13,7 @@ use neqo_crypto::{
 };
 
 use crate::cid::ConnectionId;
-use crate::frame::Frame;
+use crate::packet::PacketBuilder;
 use crate::recovery::RecoveryToken;
 use crate::Res;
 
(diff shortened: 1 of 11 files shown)


ddragana (approved):

martinthomson:
Merged as part of #1011 .

PR #1018, opened 2020-10-22
Migration

martinthomson:
This is the groundwork for migration.  I haven't done NEW_CONNECTION_ID yet (so migration doesn't work yet), but I wanted to put this up so that you can see where it is headed.

There are a few pieces of this that I want to extract and put in other pull requests as well.  The neqo-common stuff in particular can be separated.

The scary part of this change is that I ended up with a numeric reference to the current path.  This is hacktastic and I'm planning to store paths with `Rc<RefCell<Path>>` instead.  Then I can pass around a reference to the path instead; I can even use a weak reference in cases where the path isn't worth keeping (mostly related to closing).  That has some overheads involved with access, but it avoids the scary and brittle reference logic, which probably costs more anyway, both in terms of CPU cost and code cost.

One part of this design that I like is that datagrams are matched up with a path as they are received.  If it is a new path, then that path is "temporary".  If we like the packet, then we make the path a permanent one that we probe.  This eventually displaces old paths, as we only keep a few paths around.  Active paths don't go away and so it is the path that was active the longest time ago that gets the chop.  I like this design so I plan to keep that bit.

Next steps:
* build out NEW_CONNECTION_ID, which should be easy to do
* sending of probing packets needs special logic, because we don't want to commit too much to probing packets; right now, if a path wants to probe, then it gets all the frames
* the cleanup around path references
* cut this into manageable chunks
* and of course, going through and ensuring that all the important pieces have good test coverage


neqo-transport/src/connection/mod.rs
```diff
@@ -1225,21 +1270,17 @@ impl Connection {
...
+                    let space = PNSpace::from(payload.packet_type());
+                    if self.acks.get_mut(space).unwrap().is_duplicate(payload.pn()) {
+                        qdebug!([self], "Duplicate packet {}-{}", space, payload.pn());
+                        self.stats.borrow_mut().dups_rx += 1;
+                        continue;
                     }
+
+                    qlog::packet_received(&mut self.qlog, &packet, &payload);
+                    let migrate = self.process_packet(path, &payload, now)?;
+                    self.postprocess(path, &packet, migrate, now)?;
```
ddragana: Can we have a more descriptive name instead of postprocess?

ddragana (commented):
I did not get through all patches yet and I will need another pass through all patches.

These are just some comments.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -299,30 +261,32 @@ fn simultaneous_stop_sending_and_reset() {
     let stream_id = client.stream_create(StreamType::BiDi).unwrap();
     client.stream_send(stream_id, &[0x00]).unwrap();
     let out = client.process(None, now());
-    let _ = server.process(out.dgram(), now());
+    let ack = server.process(out.dgram(), now()).dgram();
 
-    let stream_readable = |e| matches!(e, ConnectionEvent::RecvStreamReadable {..});
+    let stream_readable =
+        |e| matches!(e, ConnectionEvent::RecvStreamReadable { stream_id: id } if id == stream_id);
```
martinthomson, line 267: With extra streams being created, extra filtering was needed here.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -299,30 +261,32 @@ fn simultaneous_stop_sending_and_reset() {
...
-    // Call stop sending.
-    assert_eq!(
-        Ok(()),
-        server.stream_stop_sending(stream_id, Error::NoError.code())
-    );
+    client.stream_reset_send(stream_id, 0).unwrap();
+    let out_reset_frame = client.process(ack, now()).dgram();
+
+    // Send something out of order to force the server to generate an
+    // acknowledgment at the next opportunity.
```
martinthomson, line 276: The changed timing of acknowledgments was a little tricky here too.

neqo-transport/src/path.rs
```diff
@@ -18,35 +33,447 @@ use neqo_common::Datagram;
...
+                if p.borrow().received_on(local, remote, false) {
+                    Some(Rc::clone(p))
+                } else {
+                    None
+                }
+            })
+            .unwrap_or_else(|| Rc::new(RefCell::new(Path::temporary(local, remote))))
+    }
+
+    pub fn find_path_with_rebinding(&self, local: SocketAddr, remote: SocketAddr) -> PathRef {
```
martinthomson, line 81: Add a comment here.

neqo-transport/src/path.rs
```diff
@@ -18,35 +33,447 @@ use neqo_common::Datagram;
...
+            .min()
+    }
+
+    /// Set the identified path to be primary.
+    /// This panics if `make_permanent` hasn't been called.
+    pub fn handle_migration(&mut self, path: &PathRef, remote: SocketAddr, now: Instant) {
+        qtrace!([self.primary().borrow()], "handle_migration");
+        for p in &self.paths {
+            qtrace!("Path {}", p.borrow());
+        }
```
martinthomson, line 254:
```suggestion
```

neqo-transport/src/path.rs
```diff
@@ -18,35 +33,447 @@ use neqo_common::Datagram;
...
+        if path.borrow().is_primary() {
+            // Update when the path was last regarded as valid.
+            path.borrow_mut().update(now);
+            return;
+        }
+
+        if let Some(old_path) = self.select_primary(path) {
+            // Need to probe the old path if the peer migrates.
+            old_path.borrow_mut().probe();
+            // TODO(mt) - suppress probing if the path was valid within 3PTO.
```
martinthomson, line 270: I'm leaving this as a TODO because until the path owns the RTT estimator, it is difficult to plumb through the necessary state.

neqo-transport/src/path.rs
```diff
@@ -106,4 +535,197 @@ impl Path {
...
+    pub fn amplification_limit(&self) -> usize {
+        if matches!(self.state, ProbeState::Failed) {
+            0
+        } else if self.is_valid() {
+            usize::MAX
+        } else if let Some(limit) = self.received_bytes.checked_mul(3) {
+            let budget = if limit == 0 {
+                // If we have received absolutely nothing thus far, then this endpoint
+                // is the one initiating communication on this path.  Allow enough space for probing.
+                self.mtu() * 5
```
martinthomson, line 705: There's a gotcha here: we're going to probe any path we migrate to and send a padded datagram.  We should get 1200 bytes or more back.  If we don't, we're going to be bound by the amplification limit as a client.  This isn't ideal, but I think that it's tolerable.  We will ignore this amplification limit for any PTO that we send, so we aren't completely deadlocked.  It's a little annoying, that's all.

neqo-transport/src/recovery.rs
```diff
@@ -581,13 +588,14 @@ impl PtoState {
...
+    /// This takes a packet from the supply if one remains, or returns `None`.
+    pub fn send_profile(&mut self, mtu: usize) -> Option<SendProfile> {
         if self.packets > 0 {
+            // This is a PTO, so ignore the limit.
             self.packets -= 1;
-            SendProfile::new_pto(self.space, mtu, self.probe)
+            Some(SendProfile::new_pto(self.space, mtu, self.probe))
         } else {
-            SendProfile::new_limited(0)
+            None
```
martinthomson, line 598: Note that this now returns `None` rather than returning a profile that only permits ACK frames.

This allows us to send probes on a path between the first and second PTO.  I found that we would not send anything at that time because - although we had plenty of congestion control window available - we were sending on PTO and therefore blocked.

This change might be a help if we have new data to send when the PTO timer is otherwise active.  This might be the source of some of our stalls.

neqo-transport/src/server.rs
```diff
@@ -292,19 +300,22 @@ impl Server {
...
+                    if let Ok(p) = packet {
+                        let retry = Datagram::new(dgram.destination(), dgram.source(), p);
+                        Some(retry)
+                    } else {
+                        qerror!([self], "unable to encode retry, dropping packet");
+                        None
+                    }
                 } else {
-                    qerror!([self], "unable to encode retry, dropping packet");
                     None
```
martinthomson, line 319: Need to at least log an error here.  The generator shouldn't run dry.

neqo-transport/src/server.rs
```diff
@@ -384,6 +395,30 @@ impl Server {
...
+        if let Some(odcid) = orig_dcid {
+            // There was a retry, so set the connection IDs for.
+            c.set_retry_cids(odcid, initial.src_cid, initial.dst_cid);
+        }
+        c.set_validation(Rc::clone(&self.address_validation));
+        if let Some(spa) = &self.preferred_address {
+            if c.set_preferred_address(spa).is_err() {
+                qwarn!([self], "Unable to set preferred address");
+            }
+        }
```
martinthomson, line 418: This is the new stuff.  The rest is just factored out.

neqo-transport/src/tparams.rs
```diff
@@ -455,6 +587,232 @@ mod tests {
...
+    //         if let IpAddr::V6(addr) = a.ip() {
+    //             assert!(!addr.is_unspecified());
+    //         } else {
+    //             panic!("invalid address type for v6 address");
+    //         }
+    //         assert_ne!(a.port(), 0);
+    //     }
+    //     Self { v4, v6 }
+    // }
+
```
martinthomson, line 752:
```suggestion
```

neqo-transport/src/path.rs
ddragana: ok.

ddragana (commented):
This is huge. 
It look fine and there is a lot of tests. 
Please address comment as you see fit and land it. 
Please do not land it before I create the next tag.

ddragana:
All the comments and commits messages were very useful, thanks.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -90,94 +86,65 @@ fn pto_works_full_cwnd() {
 }
 
 #[test]
-#[allow(clippy::cognitive_complexity)]
 fn pto_works_ping() {
     let mut client = default_client();
     let mut server = default_server();
     connect_force_idle(&mut client, &mut server);
```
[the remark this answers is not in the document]
martinthomson: Only those tests that count bytes (which are all in cc.rs) need to use that function.  It is harder to use `connect_rtt_idle` so it doesn't make sense to use it everywhere.

neqo-transport/src/path.rs
```diff
@@ -36,12 +36,7 @@ const MAX_PATH_PROBES: usize = 3;
 /// The maximum number of paths that `Paths` will track.
 const MAX_PATHS: usize = 15;
 
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub enum PathId {
-    Primary,
-    Temporary,
-    Id(u32),
-}
+pub type PathRef = Rc<RefCell<Path>>;
```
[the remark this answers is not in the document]
martinthomson: So do I.  It took me a little while to get there, but this seems to work pretty nicely.

neqo-transport/src/path.rs
```diff
@@ -168,11 +180,37 @@ impl Paths {
...
+    ) {
+        while let Some(seqno) = self.to_retire.pop() {
+            if builder.remaining() < 1 + Encoder::varint_len(seqno) {
+                self.to_retire.push(seqno);
+                break;
+            }
+            builder.encode_varint(FRAME_TYPE_RETIRE_CONNECTION_ID);
+            builder.encode_varint(seqno);
+            tokens.push(RecoveryToken::RetireConnectionId(seqno));
+            stats.retire_connection_id += 1;
```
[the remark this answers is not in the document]
martinthomson: Good call.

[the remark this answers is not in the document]
martinthomson: Hang on.  We generally count frames sent, including retransmissions.  Counting when we decide to retire would give us a different count as it could be some time until we put the frame out.  I'd rather get a precise accounting of the frames we send, and maybe look into counting losses so that we can get information on retransmissions - or remove them from counts - in a different way.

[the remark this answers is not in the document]
ddragana: I think we need another stats that counts retired connectionId not frames.

martinthomson:
@ddragana can you give the [last few commits](https://github.com/mozilla/neqo/compare/0c5199f..cec7ebfcc6750ad8dc42932cee6efe40efd925f0) a sanity check?  I merged the latest code and moved the preferred address configuration to use ConnectionParameters.

In doing so, I integrated #1071 and then discovered that wasn't enough.  I needed to save the value of `ConnectionParameters` because we were not setting the stream limits properly in `StreamIndexes`.  So you probably shouldn't expect the current code on `main` to work with non-default values, especially if 0-RTT is rejected.

I've defaulted to having preferred addresses enabled at the client, but you can turn them off with `ConnectionParameters::disable_preferred_address`.

I've also moved the constants for initial stream limits to `ConnectionParameters`.  Code that needs to know the default value can inspect a default `ConnectionParameters`.

Oh, and I moved `StreamType` to the `stream_id` module.  I needed it for the new module, but it didn't make sense to reference `frame`.  I just made helper functions for `Frame` to use.

neqo-transport/src/connection/mod.rs
```diff
@@ -350,15 +348,36 @@ impl Connection {
         );
         tps.set_empty(tparams::DISABLE_MIGRATION);
         tps.set_empty(tparams::GREASE_QUIC_BIT);
+    }
 
-        tps.set_integer(
+    /// Read connection parameters and update transport parameters.
+    fn read_parameters(&mut self) -> Res<()> {
```
ddragana, line 354: I am wondering, shouldn't `read_parameters` return `Res<TransportParameters>` and be responsible for creating `TransportParameters`, setting its default and reading `self.conn_params`?
I do not like that there are 2 steps and in between `TransportParameters` are not correct.

ddragana (commented):
This looks good.
My comments re only suggestions, this is good for landing.

neqo-transport/src/connection/mod.rs
martinthomson: That's good feedback, but it's a fairly large restructuring.  Opened #1073 for that.  This is definitely large enough to warrant follow-up.

ddragana: I agree, thanks.

neqo-transport/src/connection/mod.rs
```diff
@@ -2543,7 +2554,10 @@ impl Connection {
 
         self.send_streams.clear();
         self.recv_streams.clear();
-        self.indexes = StreamIndexes::new();
+        self.indexes = StreamIndexes::new(
+            self.conn_params.get_max_streams(StreamType::BiDi),
+            self.conn_params.get_max_streams(StreamType::UniDi),
```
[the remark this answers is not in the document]
martinthomson: Opened #1072.  If I get around to it before landing...

PR #1022, opened 2020-10-27
Check header fields

### LINKED ISSUE #753: Check pseudo-header fields

KershawChang:
Fix #753 

Implement some checks against response headers.
1. Pseudo header fields shoud not after fields.
2. Contains no uppercase characters in field names.
3. Have undefined or excluded header.

### DIFF
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -62,7 +62,6 @@ pub enum Error {
     HttpRequestIncomplete,
     HttpConnect,
     HttpVersionFallback,
-    HttpInvalidHeader,
     QpackError(neqo_qpack::Error),
 
     // Internal errors from here.
@@ -83,6 +82,7 @@ pub enum Error {
     TransportStreamDoesNotExist,
     InvalidInput,
     FatalError,
+    InvalidHeader,
 }
 
 impl Error {
@@ -105,7 +105,6 @@ impl Error {
             Self::HttpRequestIncomplete => 0x10d,
             Self::HttpConnect => 0x10f,
             Self::HttpVersionFallback => 0x110,
-            Self::HttpInvalidHeader => 0x111,
             Self::QpackError(e) => e.code(),
             // These are all internal errors.
             _ => 3,
@@ -133,10 +132,7 @@ impl Error {
 
     #[must_use]
     pub fn stream_reset_error(&self) -> bool {
-        matches!(
-            self,
-            Self::HttpGeneralProtocolStream | Self::HttpInvalidHeader
-        )
+        matches!(self, Self::HttpGeneralProtocolStream | Self::InvalidHeader)
     }
 
     #[must_use]
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -90,7 +90,9 @@ impl Error {
     pub fn code(&self) -> AppError {
         match self {
             Self::HttpNoError => 0x100,
-            Self::HttpGeneralProtocol | Self::HttpGeneralProtocolStream => 0x101,
+            Self::HttpGeneralProtocol | Self::HttpGeneralProtocolStream | Self::InvalidHeader => {
+                0x101
+            }
             Self::HttpInternal => 0x102,
             Self::HttpStreamCreation => 0x103,
             Self::HttpClosedCriticalStream => 0x104,
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -62,6 +62,7 @@ pub enum Error {
     HttpRequestIncomplete,
     HttpConnect,
     HttpVersionFallback,
+    HttpInvalidHeader,
     QpackError(neqo_qpack::Error),
 
     // Internal errors from here.
@@ -104,6 +105,7 @@ impl Error {
             Self::HttpRequestIncomplete => 0x10d,
             Self::HttpConnect => 0x10f,
             Self::HttpVersionFallback => 0x110,
+            Self::HttpInvalidHeader => 0x111,
             Self::QpackError(e) => e.code(),
             // These are all internal errors.
             _ => 3,
@@ -131,7 +133,10 @@ impl Error {
 
     #[must_use]
     pub fn stream_reset_error(&self) -> bool {
-        matches!(self, Self::HttpGeneralProtocolStream)
+        matches!(
+            self,
+            Self::HttpGeneralProtocolStream | Self::HttpInvalidHeader
+        )
     }
 
     #[must_use]
--- a/neqo-http3/src/recv_message.rs
+++ b/neqo-http3/src/recv_message.rs
@@ -332,10 +332,10 @@ impl RecvMessage {
             MessageType::Response => {
                 let status = headers.iter().find(|(name, _value)| name == ":status");
                 if let Some((_name, value)) = status {
-                    let status_code = value.parse::<i32>().map_err(|_| Error::HttpInvalidHeader)?;
+                    let status_code = value.parse::<i32>().map_err(|_| Error::InvalidHeader)?;
                     Ok(status_code >= 100 && status_code < 200)
                 } else {
-                    Err(Error::HttpInvalidHeader)
+                    Err(Error::InvalidHeader)
                 }
             }
             MessageType::Request => Ok(false),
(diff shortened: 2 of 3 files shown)


martinthomson (dismissed):
Thanks for writing this up.  I have a few issues, some of which are important enough to require another go-around.

neqo-http3/src/recv_message.rs
```diff
@@ -336,6 +336,58 @@ impl RecvMessage {
...
+                ];
+                for header in headers {
+                    let mut is_colon_header = false;
+                    for (i, c) in header.0.chars().enumerate() {
+                        if i == 0 && c == ':' {
+                            if seen_non_colon_header {
+                                qdebug!([self], "headers_valid - pseudo header fields after fields");
+                                return Err(Error::HttpGeneralProtocolStream);
+                            }
+                            is_colon_header = true;
```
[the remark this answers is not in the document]
KershawChang: I'll try to use `starts_with(':')`.

neqo-http3/src/recv_message.rs
```diff
@@ -336,6 +336,58 @@ impl RecvMessage {
...
+
+                    seen_non_colon_header = true;
+                    if excluded_headers.iter().any(|&excluded| excluded == header.0) {
+                        qdebug!([self], "headers_valid - found an excluded header:{}", header.0);
+                        return Err(Error::HttpGeneralProtocolStream);
+                    }
+                }
+                Ok(true)
+            }
+            MessageType::Request => Ok(false),
```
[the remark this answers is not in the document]
KershawChang: But the code here is only for checking response, right?
I'll extend this function to do some basic checks for request. The request specific checks will not be covered completely in this function.

martinthomson (commented):
Tests look great.  I think that you can lean more on rust matching here though.

neqo-http3/src/recv_message.rs
```diff
@@ -336,6 +336,118 @@ impl RecvMessage {
...
+            MessageType::Response => {
+                pseudo_header_map.entry(":status").or_insert(false);
+            }
+            MessageType::Request => {
+                pseudo_header_map.entry(":method").or_insert(false);
+                pseudo_header_map.entry(":scheme").or_insert(false);
+                pseudo_header_map.entry(":authority").or_insert(false);
+                pseudo_header_map.entry(":path").or_insert(false);
+            }
+        }
```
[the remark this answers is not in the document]
KershawChang: Thanks. This is really neat!

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
+                    | "te"
+                    | "transfer-encoding"
+                    | "upgrade"
+                    | "accept-encoding"
+            ) {
                 return Err(Error::HttpGeneralProtocolStream);
             }
         }
 
+        pseudo_state ^= REGULAR_HEADER;
```
martinthomson, line 415:
```suggestion
        pseudo_state &= !REGULAR_HEADER;
```

If you want to clear the bit, then this is safer.  Your code relies on there being a regular header (which seems likely, but not guaranteed).

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
 
-                if let hash_map::Entry::Occupied(mut e) = pseudo_header_map.entry(&header.0) {
-                    if *e.get_mut() {
-                        return Err(Error::HttpGeneralProtocolStream);
-                    }
-                    *e.get_mut() = true;
-                } else {
+            let mut bytes = header.0.bytes();
+            if is_pseudo {
+                if (pseudo_state & (PSEUDO_HEADER_UNDEFINED | REGULAR_HEADER)) != 0 {
```
martinthomson, line 386: This can be tracked in track_pseudo, right?

The former can just return an error (you don't need the bit then), the latter can be added ahead of the tuple in the main if statement (I'll mark it).

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
-            MessageType::Response => {
-                pseudo_header_map.entry(":status").or_insert(false);
-            }
-            MessageType::Request => {
-                pseudo_header_map.entry(":method").or_insert(false);
-                pseudo_header_map.entry(":scheme").or_insert(false);
-                pseudo_header_map.entry(":authority").or_insert(false);
-                pseudo_header_map.entry(":path").or_insert(false);
-            }
+    fn track_pseudo(&self, name: &str, state: &mut u8) -> Res<bool> {
```
martinthomson, line 348: This can be an associated function (`Self::track_pseudo` rather than `self.track_pseudo`) as it doesn't use `self`.

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
+        let (pseudo, bit) = if name.starts_with(':') {
+            (
```
martinthomson, line 350: You can check `(*state & REGULAR_HEADER) != 0` here and return an error.

I would also assign the output of the match to a variable rather than wrap all of this in the tuple, so
```rust
  let bit = match ... { ... };
  (true, bit)
```

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
+                true,
+                match self.message_type {
+                    MessageType::Response => match name {
+                        ":status" => PSEUDO_HEADER_STATUS,
+                        _ => PSEUDO_HEADER_UNDEFINED,
```
martinthomson, line 355:
```suggestion
                        _ => return Err(Error::HttpGeneralProtocolStream),
```

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
-
-                if excluded_header_byte
-                    .iter()
-                    .any(|&excluded| excluded == byte)
-                {
-                    qdebug!([self], "headers_valid - found an excluded byte");
-                    return Err(Error::HttpGeneralProtocolStream);
-                }
+            if bytes.any(|b| matches!(b, 0 | 0x10 | 0x13 | 0x3a | 0x41..=0x5a)) {
+                return Err(Error::HttpGeneralProtocolStream); // illegal characters.
```
martinthomson, line 397: I think that we need a better error for this header validation.  It can be treated as a stream error and the same error code, but it would be good to have better precision regarding local errors.  If we had `Error::InvalidHeaders` or similar we could at least trace errors back to this code.

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
+                    | "accept-encoding"
+            ) {
                 return Err(Error::HttpGeneralProtocolStream);
             }
         }
 
+        pseudo_state ^= REGULAR_HEADER;
         match self.message_type {
+            MessageType::Response => {
+                let response_mask = PSEUDO_HEADER_STATUS;
```
martinthomson, line 418: const ?

neqo-http3/src/recv_message.rs
```diff
@@ -337,115 +345,93 @@ impl RecvMessage {
...
+                if pseudo_state & response_mask != response_mask {
+                    return Err(Error::HttpGeneralProtocolStream);
+                }
```
martinthomson, line 421: You can move this block outside of the match statement.

martinthomson (commented):
I'm liking how this is shaping up.

neqo-http3/src/recv_message.rs
KershawChang: I'll add a new error code `HttpInvalidHeader`.

neqo-http3/src/lib.rs
```diff
@@ -103,6 +104,7 @@ impl Error {
             Self::HttpRequestIncomplete => 0x10d,
             Self::HttpConnect => 0x10f,
             Self::HttpVersionFallback => 0x110,
+            Self::HttpInvalidHeader => 0x111,
```
martinthomson, line 107: IMPORTANT: this code is not in the spec, so you can't rely on it.  We have to use a default error.

```suggestion
```

Unfortunately, this means that you can't check that the code is correct in the Reset events above.

I will enquire about whether it makes sense to have a new error code here and get back to you.

neqo-http3/src/lib.rs
```diff
@@ -130,7 +132,10 @@ impl Error {
 
     #[must_use]
     pub fn stream_reset_error(&self) -> bool {
-        matches!(self, Self::HttpGeneralProtocolStream)
+        matches!(
+            self,
+            Self::HttpGeneralProtocolStream | Self::HttpInvalidHeader
```
martinthomson, line 137: This bit is good.

martinthomson (commented):
*Almost* perfect.  Very nice.  Just the error code thing in the way of landing this.

neqo-http3/src/lib.rs
KershawChang: Thanks.

martinthomson: Negative on getting the error code.  This maps to 3.

KershawChang: I'll add a new internal error `InvalidHeader`.

ddragana: You should use HttpGeneralProtocolStream. The cod should be 0x103.

neqo-http3/src/lib.rs
```diff
@@ -104,7 +104,6 @@ impl Error {
             Self::HttpRequestIncomplete => 0x10d,
             Self::HttpConnect => 0x10f,
             Self::HttpVersionFallback => 0x110,
-            Self::HttpInvalidHeader => 0x111,
```
ddragana: Map the new error to 0x101. In the same way as HttpGeneralProtocolStream above.

ddragana:
Fix the small comment above and land this. This looks really good.

PR #1024, opened 2020-10-28
Defer processing of saved datagrams

martinthomson:
This ensures that we don't jump into handling Handshake packets before
we have finished processing the current datagram.

This doesn't do much, but the moved assertion in start_handshake()
should highlight why this was not great.  Without the changes, that
assertion fails.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -713,12 +713,18 @@ impl Connection {
         self.crypto.tls.peer_certificate()
     }
 
-    /// Call by application when the peer cert has been verified
+    /// Call by application when the peer cert has been verified.
+    ///
+    /// This panics if there is no active peer.  It's OK to call this
+    /// when authentication isn't needed, that will likely only cause
+    /// the connection to fail.  However, if no packets have been
+    /// exchanged, it's not OK.
     pub fn authenticated(&mut self, status: AuthenticationStatus, now: Instant) {
         qinfo!([self], "Authenticated {:?}", status);
         self.crypto.tls.authenticated(status);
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1030,12 +1037,16 @@ impl Connection {
...
-    fn process_saved(&mut self, cspace: CryptoSpace) {
-        if self.crypto.states.rx_hp(cspace).is_some() {
-            for saved in self.saved_datagrams.take_saved(cspace) {
-                qtrace!([self], "process saved @{:?}: {:?}", saved.t, saved.d);
-                self.process_input(saved.d, saved.t);
+    /// Process any saved datagrams that might be available for processing.
+    fn process_saved(&mut self, now: Instant) {
+        while let Some(cspace) = self.saved_datagrams.available() {
+            qdebug!([self], "process saved for space {:?}", cspace);
+            if self.crypto.states.rx_hp(cspace).is_some() {
```
ddragana, line 1044: I do not like that this is theory could loop if we have mismatch between making a space available and `self.crypto.states.rx_hp(cspace).is_some()` returning false. Can we do it a bit different?

martinthomson: Hmm.  I think that we have a guarantee that we won't ever mark something as available if the crypto space isn't available.  I think that what we really need here is an assertion rather than an if statement, like:

```rust
debug_assert!(self.crypto.states.rx_hp(cspace).is_some());
```

Would that make you more comfortable?

ddragana: let's do that.

Please add the assertion and land the PR.

PR #1025, opened 2020-10-28
Expose HFrameReader outside of the crate. It is needed for necko tests

ddragana:


neqo-qpack/src/lib.rs
```diff
@@ -27,6 +27,8 @@ mod static_table;
 pub mod stats;
 mod table;
 
+pub use decoder::QPackDecoder;
```
martinthomson, line 30: If you are going to expose this this way, then you should also expose the encoder.

martinthomson (commented):
Looks good.  Tweak the qpack thing and land it.

PR #1028, opened 2020-10-29
Script to update neqo version

ddragana:
Some time ago I wrote a script to update neqo version. I will just post the PR as a draft, I do not want to add it to repo.

### DIFF
--- a/update_version
+++ b/update_version
@@ -1,0 +1,13 @@
+#! /bin/bash
+
+cd "$(dirname "$0")"
+set -e
+
+[[ -n "$1" ]] || { echo "Usage: $0 <version>"; exit 1; }
+
+for entry in `find . -mindepth 2 -name Cargo.toml -print`; do
+    echo $entry
+    line=$(grep -n -m1 "version" $entry | cut -d: -f2)
+    CURRENT=$(echo ${line} | awk -F'"' '{print $2}')
+    sed -i "" "s/$CURRENT/$1/g" $entry
+done
--- a/update_version
+++ b/update_version
@@ -1,13 +1,16 @@
-#! /bin/bash
+#! /usr/bin/env bash
 
-cd "$(dirname "$0")"
+pushd "$(dirname "$0")" > /dev/null
 set -e
 
 [[ -n "$1" ]] || { echo "Usage: $0 <version>"; exit 1; }
 
-for entry in `find . -mindepth 2 -name Cargo.toml -print`; do
-    echo $entry
-    line=$(grep -n -m1 "version" $entry | cut -d: -f2)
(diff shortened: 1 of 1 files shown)


update_version
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
```
martinthomson, line 2:
```suggestion

cd "$(dirname "$0")"
set -e

[[ -n "$1" ]] || ! echo "Usage: $0 <version>"
```

update_version
```diff
@@ -0,0 +1,9 @@
...
+for entry in `ls ./*/Cargo.toml`; do
```
martinthomson, line 3:
```suggestion
for entry in `find . -mindepth 2 -name Cargo.toml -print`; do
```

martinthomson (commented):
I don't see why you wouldn't put this in-tree.  It's useful to have.

That said, I would just put the new version number in as "$1" and use that.  This might work better though:

```sh
find . -mindepth 2 -name Cargo.toml -exec \
  sed -i~ -e '0,/^version =/s/^version =.*/version = "'"$1"'"/' {} \+
```

update_version
```diff
@@ -0,0 +1,13 @@
...
+cd "$(dirname "$0")"
+set -e
+
+[[ -n "$1" ]] || { echo "Usage: $0 <version>"; exit 1; }
+
+for entry in `find . -mindepth 2 -name Cargo.toml -print`; do
+    echo $entry
+    line=$(grep -n -m1 "version" $entry | cut -d: -f2)
+    CURRENT=$(echo ${line} | awk -F'"' '{print $2}')
+    sed -i "" "s/$CURRENT/$1/g" $entry
```
martinthomson, line 12: I think that you need a -e here (and the `-i ""` might fail on mac, I can only remember that `-i~` works everywhere.

ddragana: This works om my mac.

mb: -i "" fails on linux for me. I suggest removing the space between `-i` and `""` changing it to

```
sed -i"" "s/$CURRENT/$1/g" $entry
``` 

Which only doesn't works on MacOS up to 10.8[[1]](https://stackoverflow.com/a/4247319), which had end of support around September 2016[[2]](https://en.wikipedia.org/wiki/MacOS_version_history#Releases). So it works on all platforms I think we care about.

martinthomson: OK, good.  Note that `sed -i""` is exactly equivalent to just `sed -i`.

mb: Yes, on linux it is equivalent. On macos only `sed -i` doesn't work if I understand it correctly.

larseggert:
I pushed a version that addresses the issues.

mb (approved):
Thanks. Looks good and works for me :+1:

Issue #804, opened 2020-07-08
Handle the case when we cannot create critical HTTP3 streams 


ddragana:
Probably we should just close a connection if server does not allow us to create control, and qpack streams due to the max streamId limit.

KershawChang:
@ddragana , I am not really sure what to do here.
It seems that the connection will be closed when `initialize_http3_connection` returns an error. Or maybe I am looking at the wrong place?

martinthomson:
@KershawChang, it might be worth testing that this is what happens.  Closing the connection would be the right reaction, but we should at least test that case.

ddragana:
Fixed by #1040

PR #1040, opened 2020-11-10
add tests to test the case when server doesn't allow to create critic…

KershawChang:
For #804 
Test the case that the server doesn't allow to create control and qpack streams.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -737,8 +737,9 @@ mod tests {
     use neqo_common::{event::Provider, Datagram, Decoder, Encoder};
     use neqo_crypto::{AllowZeroRtt, AntiReplay, ResumptionToken};
     use neqo_qpack::encoder::QPackEncoder;
+    use neqo_transport::tparams::{self, TransportParameter};
     use neqo_transport::{
-        CloseError, CongestionControlAlgorithm, ConnectionEvent, FixedConnectionIdManager,
+        CloseError, CongestionControlAlgorithm, ConnectionEvent, FixedConnectionIdManager, Output,
         QuicVersion, State, RECV_BUFFER_SIZE, SEND_BUFFER_SIZE,
     };
     use std::convert::TryFrom;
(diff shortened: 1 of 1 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5945,27 @@ mod tests {
         };
         assert!(server.conn.events().any(stop_sending_event));
     }
+
+    fn do_test_client_critical_stream_create_failed(max_stream: u64) {
```
martinthomson, line 5949: I think that this is a case where a generic function might be better.  That is, it might be better to have these two functions shaped like this:

```rust
#[test]
fn client_control_stream_create_fail() {
       let mut client = default_http3_client();
       let mut server = TestServer::new();
       server.set_max_uni_stream(max_stream);
       handshake_client_error(&mut client, &Error::StreamLimitError);
}
```

What do you think?  It's a little more duplicated code (you run the setup steps twice), but it means that we get a generic "handshake breaks" handler.

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5945,27 @@ mod tests {
...
+        let mut server = TestServer::new();
+        server.set_max_uni_stream(max_stream);
+        let out = do_handshake_only(&mut client, &mut server);
+        client.process(out.dgram(), now());
+        assert_closed(&client, &Error::StreamLimitError);
+    }
+
+    #[test]
+    fn test_client_control_stream_create_failed() {
+        // Client fails to create a control stream, since server
```
martinthomson, line 5960: Rather than use `//` comments inside the function, it is probably better to use `///` comments above.

martinthomson (commented):
Looks good.  Some nits and suggestions that you can take or leave.

neqo-http3/src/connection_client.rs
KershawChang: Yeah, I agree this looks clearer.
My origin code `do_test_client_critical_stream_create_failed(0);` does not look easy to understand.

ddragana (approved):

Issue #1042, opened 2020-11-12
Some qpack instructions can be sent before the stream type is sent


ddragana:
This is a server side problem:
If 0RTT is used and we receive a stream with encoded header block that uses the dynamic table, a header-ack instruction will be queued for sending immediately. This will happened before stream is created. The decoder-stream id is only encoded and queued for sending after the stream is created and thereby the header-ack instruction will be sent before the decoder-stream id.

This was causing some necko test to Intermittent test failure.

ddragana:
Fixed by #1044

PR #1044, opened 2020-11-13
Make sure to send a stream type before sending any other data.

ddragana:


ddragana:
In the test there is a description how this bug is triggered.

Fixes #1042

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
     }
+
+    fn maybe_authenticate(conn: &mut Http3Client) {
+        let authentication_needed = |e| matches!(e, Http3ClientEvent::AuthenticationNeeded);
+        if conn.events().any(authentication_needed) {
+            conn.authenticated(AuthenticationStatus::Ok, now());
+        }
+    }
+
+    // Test that decoder stream type is always sent beofer any other instruction also
```
martinthomson, line 5949:
```suggestion
    // Test that decoder stream type is always sent before any other instruction also
```

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+
+        let mut datagram = None;
+        let is_done = |c: &Http3Client| matches!(c.state(), Http3State::Connected);
+        while !is_done(&mut client) {
+            maybe_authenticate(&mut client);
+            datagram = client.process(datagram, now()).dgram();
+            datagram = server.process(datagram, now()).dgram();
+        }
+
+        // exchannge qpack settings, server will send a token as well.
```
martinthomson, line 5978:
```suggestion
        // exchange qpack settings, server will send a token as well.
```

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+        let token = client
+            .events()
+            .find_map(|e| {
+                if let Http3ClientEvent::ResumptionToken(token) = e {
+                    Some(token)
+                } else {
+                    None
+                }
+            })
+            .unwrap();
```
martinthomson, line 5992: Everything above here is basically setup for 0-RTT.  Make a function that runs client and server and spits out a token.  (It might have to take the server as an argument, but the client can be constructed in that function.

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+            DEFAULT_ALPN_H3,
+            anti_replay(),
+            Rc::new(RefCell::new(FixedConnectionIdManager::new(5))),
+            QpackSettings {
+                max_table_size_encoder: MAX_TABLE_SIZE,
+                max_table_size_decoder: MAX_TABLE_SIZE,
+                max_blocked_streams: MAX_BLOCKED_STREAMS,
+            },
+        )
+        .unwrap();
```
martinthomson, line 5968: This looks fairly generic.  Can it be factored out?

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+        assert_eq!(client.state(), Http3State::ZeroRtt);
+        let zerortt_event = |e| matches!(e, Http3ClientEvent::StateChange(Http3State::ZeroRtt));
+        assert!(client.events().any(zerortt_event));
+
+        // Send a request using 0RTT data. This will trigger sending a HEADER_ACK on
+        // the decoder stream before stream id is written.
+        // The client will read HEADER_ACK (value: 0x80) as a stream type. To decode the stream
+        // type we will need 3 more byte (0x80 starts with 10 which is 4 byte encoded verint).
+        // 3 more bytes wil be the corect stream type (0x3) which will be added and to more
+        // requests that will trigger 2 HEADER_ACK. The client will cloed the stream and
```
martinthomson, line 6010:
```suggestion
        // requests that will trigger 2 HEADER_ACK. The client will close the stream and
```

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+        // the server will trigger closing with error HttpClosedCriticalStream.
```
martinthomson, line 6011: Is this a comment about how this *used* to work?  Because what I see here is the request and response just working.

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
+        let out = client.process(out.dgram(), now());
+        let _ = server.process(out.dgram(), now());
```
martinthomson, line 6047: There are no real assertions here regarding the requests that were made.  You need that to show that the requests are getting through, they are getting processed successfully, and responses are received.

neqo-http3/src/control_stream_local.rs
```diff
@@ -25,6 +25,18 @@ impl ::std::fmt::Display for ControlStreamLocal {
 }
 
 impl ControlStreamLocal {
+    pub fn new() -> Self {
+        let mut buf = Vec::new();
+        let mut enc = Encoder::default();
+        enc.encode_varint(HTTP3_UNI_STREAM_TYPE_CONTROL);
+        buf.append(&mut enc.into());
```
martinthomson, line 32: You don't need `buf` here.  The return value from `enc.into()` is a vector.  Or you could just use `vec![u8::try_from(HTTP3_UNI_STREAM_TYPE_CONTROL).unwrap()]` and save a lot of work.

neqo-http3/src/connection_client.rs
```diff
@@ -5930,4 +5938,113 @@ mod tests {
...
+        );
+
+        let out = client.process(None, now());
+        let out = server.process(out.dgram(), now());
+
+        // Send a response
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
+        let out = client.process(out.dgram(), now());
+        let out = server.process(out.dgram(), now());
```
martinthomson, line 6025: There is a lot of back and forth here, but no real clues as to what is going on.  Some assertions would help.  Some comments where assertions don't make sense.

martinthomson (changes requested):
I think that test could use a little more work before we land this.  The fix seems good though.

neqo-http3/src/connection_client.rs
ddragana: Yes, I will update the comment

neqo-http3/src/connection_client.rs
```diff
@@ -5946,26 +5946,11 @@ mod tests {
...
-    // Test that decoder stream type is always sent beofer any other instruction also
-    // in case when 0RTT is used.
-    #[test]
-    fn zerortt_request_use_dynamic_table() {
-        const MAX_TABLE_SIZE: u64 = 65536;
-        const MAX_BLOCKED_STREAMS: u16 = 5;
+    const MAX_TABLE_SIZE: u64 = 65536;
+    const MAX_BLOCKED_STREAMS: u16 = 5;
+
+    fn get_token(server: &mut Http3Server) -> ResumptionToken {
```
martinthomson, line 5952:
```suggestion
    fn get_resumption_token(server: &mut Http3Server) -> ResumptionToken {
```

Probably best to say what sort.

71
```diff
@@ -0,0 +1,2840 @@
+warning: couldn't execute `llvm-config --prefix` (error: No such file or directory (os error 2))
```
martinthomson, line 1: oops

neqo-http3/src/connection_client.rs
```diff
@@ -6002,49 +6009,29 @@ mod tests {
...
+        // TODO this may be optimize at Http3Server.
+        let out = client.process(None, now()).dgram();
+        let out = server.process(out, now()).dgram();
+        let out = client.process(out, now()).dgram();
+        let out = server.process(out, now()).dgram();
+        let out = client.process(out, now()).dgram();
+        let out = server.process(out, now()).dgram();
+        let out = client.process(out, now()).dgram();
+        let out = server.process(out, now()).dgram();
+        let _ = client.process(out, now());
```
martinthomson, line 6032: There is enough here that you might want to loop instead.

martinthomson (commented):
Fix the oops and land it.

PR #1046, opened 2020-11-16
Expose the transport stats and clone the qpack stats.

ddragana:


martinthomson (approved):

neqo-http3/src/connection_client.rs
```diff
@@ -689,15 +689,20 @@ impl Http3Client {
...
     #[must_use]
-    pub fn qpack_encoder_stats(&self) -> &Stats {
+    pub fn qpack_encoder_stats(&self) -> QStats {
         self.base_handler.qpack_encoder.stats()
     }
 
+    #[must_use]
+    pub fn transport_stats(&self) -> TStats {
+        self.conn.stats()
+    }
```
martinthomson, line 704: I think that it would be better overall to have a single stats object for the HTTP layer that bundles up all of the stats.  That means a larger clone, but I think that it would be easier to use.

neqo-http3/src/connection_client.rs
```diff
@@ -18,10 +18,10 @@ use neqo_common::{
     Datagram, Decoder, Encoder, Role,
 };
 use neqo_crypto::{agent::CertificateInfo, AuthenticationStatus, ResumptionToken, SecretAgentInfo};
-use neqo_qpack::{stats::Stats, QpackSettings};
+use neqo_qpack::{QpackSettings, Stats as QStats};
```
martinthomson, line 21:
```suggestion
use neqo_qpack::{QpackSettings, Stats as QpackStats};
```

neqo-http3/src/connection_client.rs
```diff
@@ -18,10 +18,10 @@ use neqo_common::{
...
 use neqo_transport::{
     AppError, CongestionControlAlgorithm, Connection, ConnectionEvent, ConnectionId,
-    ConnectionIdManager, Output, QuicVersion, StreamId, StreamType, ZeroRttState,
+    ConnectionIdManager, Output, QuicVersion, Stats as TStats, StreamId, StreamType, ZeroRttState,
```
martinthomson, line 24:
```suggestion
    ConnectionIdManager, Output, QuicVersion, Stats as TransportStats, StreamId, StreamType, ZeroRttState,
```

martinthomson (commented):
These are suggestions only.

neqo-http3/src/connection_client.rs
ddragana: I will leave it like this for now. For use with necko there is no much difference.

PR #1048, opened 2020-11-17
Header accept-encoding should be allowed

ddragana:


KershawChang (approved):
Just wondering should we replace `accept-encoding` to `sec-websocket-key`?
https://searchfox.org/mozilla-central/rev/277ab3925eae21b419167a34624ec0ab518d0c94/netwerk/socket/neqo_glue/src/lib.rs#276

ddragana:
I think that header is there because of our weird nsHttpTransport implementation. We should refactor that at some point. DevTool will show wrong headers as  well because of the same issue. DevTools will show that we have sent connection headers even though we are not sending them.

PR #1054, opened 2020-12-08
truncate reason string

### LINKED ISSUE #1053: Crash: assert!(body.len() <= 2048); in neqo-transport/src/crypto.rs#l549
0 xul.dll RustMozCrash mozglue/static/rust/wrappers.cpp:16
1 xul.dll mozglue_static::panic_hook mozglue/static/rust/lib.rs:89
2 xul.dll core::ops::function::Fn::call<fn ../18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/ops/function.rs:227
3 xul.dll std::panicking::rust_panic_with_hook ../18bf6b4f01a6feaf7259ba7cdae58031af1b7b39//library/std/src/panicking.rs:573
4 xul.dll std::panicking::begin_panic::{{closure}}<str*> ../18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:498
5 xul.dll std::sys_common::backtrace::__rust_end_short_backtrace<closure-0, !> ../18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:153
6 xul.dll std::panicking::begin_panic<str*> ../18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:497
7 xul.dll neqo_transport::crypto::CryptoDxState::encrypt third_party/rust/neqo-transport/src/crypto.rs:552
8 xul.dll static neqo_transport::packet::PacketBuilder::build third_party/rust/neqo-transport/src/packet/mod.rs:320
9 xul.dll neqo_transport::connection::Connection::output third_party/rust/neqo-transport/src/connection/mod.rs:1408

KershawChang:
To fix #1053.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -7,7 +7,7 @@
 // The class implementing a QUIC connection.
 
 use std::cell::RefCell;
-use std::cmp::max;
+use std::cmp::{max, min};
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt::{self, Debug};
@@ -1531,7 +1531,8 @@ impl Connection {
                 if let CloseError::Transport(_) = error_code {
                     builder.encode_varint(*frame_type);
                 }
-                builder.encode_vvec(reason_phrase);
+                let reason_len = min(min(reason_phrase.len(), 256), builder.remaining());
+                builder.encode_vvec(&reason_phrase[..reason_len]);
             } else {
                 unreachable!();
             }
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1531,7 +1531,7 @@ impl Connection {
                 if let CloseError::Transport(_) = error_code {
                     builder.encode_varint(*frame_type);
                 }
-                let reason_len = min(min(reason_phrase.len(), 256), builder.remaining());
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1531,7 +1531,8 @@ impl Connection {
                 if let CloseError::Transport(_) = error_code {
                     builder.encode_varint(*frame_type);
                 }
-                builder.encode_vvec(reason_phrase);
+                let reason_len = min(min(reason_phrase.len(), 256), builder.remaining());
+                builder.encode_vvec(&reason_phrase[..reason_len]);
```
martinthomson, line 1535: This is going to overflow the remaining space as a result of the length field.  As the length is limited to 256 here, that means you might need two bytes for that length.

```suggestion
                let reason_len = min(min(reason_phrase.len(), 256), builder.remaining() - 2);
                builder.encode_vvec(&reason_phrase[..reason_len]);
```

martinthomson (changes requested):
One important change.

Can you add a test case for this?  It should be fairly easy to make a connection, connect, and close it with a very large reason string.

ddragana (approved):

PR #1055, opened 2020-12-10
Add possibilty to set INITIAL_MAX_STREAMS_BIDI/UNI limits.

ddragana:
Also set the limits to 100 for the QNS Zerortt test.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -503,7 +503,7 @@ fn client(
         Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
         local_addr,
         remote_addr,
-        &CongestionControlAlgorithm::NewReno,
+        CongestionControlAlgorithm::NewReno,
         quic_protocol,
     )?;
     let ciphers = args.get_ciphers();
@@ -998,7 +998,7 @@ mod old {
             Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
             local_addr,
             remote_addr,
-            &CongestionControlAlgorithm::NewReno,
+            CongestionControlAlgorithm::NewReno,
             quic_protocol,
         )?;
 
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -20,7 +20,7 @@ use neqo_http3::{
 use neqo_qpack::QpackSettings;
 use neqo_transport::{
     Connection, ConnectionId, ConnectionParameters, Error as TransportError,
-    FixedConnectionIdManager as EmptyConnectionIdGenerator, QuicVersion,
+    FixedConnectionIdManager as EmptyConnectionIdGenerator, QuicVersion, StreamType,
 };
 
 use std::cell::RefCell;
@@ -177,6 +177,9 @@ pub struct Args {
     /// The set of TLS cipher suites to enable.
     /// From: TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256.
     ciphers: Vec<String>,
+
+    #[structopt(subcommand)]
+    quic_parameters: QuicParameters,
 }
 
 impl Args {
@@ -190,6 +193,27 @@ impl Args {
                 _ => None,
             })
             .collect::<Vec<_>>()
+    }
+}
+
+#[derive(Debug, StructOpt)]
+struct QuicParameters {
+    #[structopt(long, default_value = "16")]
+    /// Set the MAX_STREAMS_BIDI limit.
+    max_streams_bidi: u64,
+
+    #[structopt(long, default_value = "16")]
+    /// Set the MAX_STREAMS_UNI limit.
+    max_streams_uni: u64,
+}
+
+impl QuicParameters {
+    fn get(&self) -> ConnectionParameters {
+        ConnectionParameters::default()
+            .max_streams(StreamType::BiDi, self.max_streams_bidi)
+            .unwrap()
+            .max_streams(StreamType::UniDi, self.max_streams_uni)
+            .unwrap()
     }
 }
 
@@ -503,7 +527,7 @@ fn client(
         Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
         local_addr,
         remote_addr,
-        &ConnectionParameters::default().quic_version(quic_protocol),
+        &args.quic_parameters.get().quic_version(quic_protocol),
     )?;
     let ciphers = args.get_ciphers();
     if !ciphers.is_empty() {
@@ -721,9 +745,8 @@ mod old {
     use neqo_common::{event::Provider, Datagram};
     use neqo_crypto::{AuthenticationStatus, ResumptionToken};
     use neqo_transport::{
-        Connection, ConnectionEvent, ConnectionParameters, Error,
-        FixedConnectionIdManager as EmptyConnectionIdGenerator, Output, QuicVersion, State,
-        StreamType,
+        Connection, ConnectionEvent, Error, FixedConnectionIdManager as EmptyConnectionIdGenerator,
+        Output, QuicVersion, State, StreamType,
     };
 
     use super::{emit_datagram, get_output_file, Args};
@@ -997,7 +1020,7 @@ mod old {
             Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
             local_addr,
             remote_addr,
-            &ConnectionParameters::default().quic_version(quic_protocol),
+            &args.quic_parameters.get().quic_version(quic_protocol),
         )?;
 
         if let Some(tok) = token {
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -19,7 +19,7 @@ use neqo_http3::{
 };
 use neqo_qpack::QpackSettings;
 use neqo_transport::{
-    CongestionControlAlgorithm, Connection, ConnectionId, Error as TransportError,
+    Connection, ConnectionId, ConnectionParameters, Error as TransportError,
     FixedConnectionIdManager as EmptyConnectionIdGenerator, QuicVersion,
 };
 
@@ -503,8 +503,7 @@ fn client(
         Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
         local_addr,
         remote_addr,
-        CongestionControlAlgorithm::NewReno,
-        quic_protocol,
+        &ConnectionParameters::default().quic_version(quic_protocol),
     )?;
     let ciphers = args.get_ciphers();
     if !ciphers.is_empty() {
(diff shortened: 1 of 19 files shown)


ddragana:
Fixes 1050.

ddragana:
I has this already implemented before your comment on the issue, therefore I am posting it for a review.

neqo-client/src/main.rs
```diff
@@ -503,7 +503,7 @@ fn client(
         Rc::new(RefCell::new(EmptyConnectionIdGenerator::new(0))),
         local_addr,
         remote_addr,
-        &CongestionControlAlgorithm::NewReno,
+        CongestionControlAlgorithm::NewReno,
```
martinthomson, line 506: This seems unrelated.  Perhaps we need a rust 1.48 PR that does all the clippy cleanup.

neqo-server/src/main.rs
```diff
@@ -90,6 +90,10 @@ struct Args {
     /// Use http 0.9 instead of HTTP/3
     use_old_http: bool,
 
+    #[structopt(name = "max-streams", long)]
+    /// Set MAX_STREAMS limits.
+    max_streams: Option<u64>,
```
martinthomson, line 95: This isn't enough, as it doesn't allow you to control all of the different limits independently.  We should at least have an idea of how we want to manage that.

neqo-transport/src/cc/mod.rs
```diff
@@ -49,6 +49,7 @@ pub trait CongestionControl: Display + Debug {
     fn on_packet_sent(&mut self, pkt: &SentPacket);
 }
 
+#[derive(Copy, Clone)]
```
martinthomson, line 52: This is good, but it's confounding the review of this PR.

neqo-transport/src/connection/mod.rs
```diff
@@ -225,6 +225,12 @@ impl AddressValidationInfo {
     }
 }
 
+struct ConnectionParameters {
+    pub quic_version: QuicVersion,
+    pub cc_algorithm: CongestionControlAlgorithm,
+    pub max_streams: Option<u64>,
+}
```
martinthomson, line 232: I think that you might want to double-down on this model more heavily by making this struct public.  However, I would not expose the member fields or we get worse compatibility problems as we upgrade.  I know that having a bunch of getters and setters is annoying, but I think that is the right strategy here.

Using setters allows us to have defaults attached to this struct.  You can use the builder pattern with functions that use the `fn max_streams_remote_bidi(self, v: u64) -> Self` signature to make using this easy.  Then you don't need to manage `Option<T>` instances everywhere.

Also, can you have this implement `Clone`.  The server class will need that if we make it a public interface.

Also, can you split the stream limit up by type?  (We'll need to do the same for flow control, but that can come later.)

neqo-transport/src/connection/mod.rs
```diff
@@ -335,17 +344,21 @@ impl Connection {
...
         Self::new(
             Role::Server,
             Server::new(certs)?.into(),
             cid_manager,
             protocols,
             None,
-            cc_algorithm,
-            quic_version,
+            &ConnectionParameters {
+                cc_algorithm: cc_algorithm,
```
martinthomson, line 358:
```suggestion
                cc_algorithm,
```

martinthomson (changes requested):
Thanks for doing this.  I think that we need something a little more complete here.

neqo-transport/src/cc/mod.rs
ddragana: This is need for the PR to compile.

ddragana:
I have not finished this yet. I will make it a draft, because I am still not happy with the PR and I do not have time to finish it today.

ddragana:
ok, it is ready for review.

neqo-server/src/main.rs
```diff
@@ -90,6 +90,10 @@ struct Args {
     /// Use http 0.9 instead of HTTP/3
     use_old_http: bool,
 
+    #[structopt(name = "max-streams", long, default_value = "16")]
```
martinthomson, line 93: A point solution here is probably OK, but you only include this one transport parameter and only for the server.  Can you open an issue to expand this further?

Any generic solution should probably include a function to generate connection parameters.

neqo-server/src/main.rs
```diff
@@ -382,6 +386,7 @@ impl ServersRunner {
                     &[args.alpn.clone()],
                     anti_replay,
                     cid_mgr,
+                    args.max_streams_bidi,
```
martinthomson, line 389: It would be better if this were passed ConnectionParameters.  You can add a function to Args to generate a set.  Then it will be easier to add new options.

neqo-transport/src/connection/mod.rs
```diff
@@ -225,6 +225,69 @@ impl AddressValidationInfo {
     }
 }
 
+#[derive(Clone)]
```
martinthomson, line 228: A little bit of documentation on this struct would be very helpful.  It shouldn't need to be very detailed.

neqo-transport/src/connection/mod.rs
```diff
@@ -225,6 +225,69 @@ impl AddressValidationInfo {
...
+            max_streams_bidi: LOCAL_STREAM_LIMIT_BIDI,
+            max_streams_uni: LOCAL_STREAM_LIMIT_UNI,
+        }
+    }
+}
+
+impl ConnectionParameters {
+    pub fn get_quic_version(&self) -> QuicVersion {
+        self.quic_version
+    }
```
martinthomson, line 250: Do you think we need these getters?  You haven't used these consistently, and it might be enough that we can access the members directly.

neqo-transport/src/connection/mod.rs
```diff
@@ -424,13 +492,13 @@ impl Connection {
...
             state_signaling: StateSignaling::Idle,
-            loss_recovery: LossRecovery::new(cc_algorithm, stats.clone()),
+            loss_recovery: LossRecovery::new(conn_params.cc_algorithm, stats.clone()),
             events: ConnectionEvents::default(),
             new_token: NewTokenState::new(role),
             stats,
             qlog: NeqoQlog::disabled(),
             release_resumption_token_timer: None,
-            quic_version,
+            quic_version: conn_params.quic_version,
```
martinthomson, line 501: Would it make sense to keep a copy of the parameters instead?  They are immutable, so there is no risk of them become inconsistent with anything that used them.

neqo-transport/src/server.rs
```diff
@@ -407,8 +410,7 @@ impl Server {
             &self.certs,
             &self.protocols,
             Rc::clone(&cid_mgr) as _,
-            &CongestionControlAlgorithm::NewReno,
-            initial.quic_version,
+            &self.conn_params.clone().quic_version(initial.quic_version),
```
martinthomson, line 413: This is another case where passing by value might make sense.

neqo-transport/tests/server.rs
```diff
@@ -939,3 +941,76 @@ fn closed() {
     let res = server.process(None, now() + Duration::from_secs(60));
     assert_eq!(res, Output::None);
 }
+
+#[test]
+fn max_streams() {
+    const MAX_STREAMS: u64 = 100;
```
martinthomson, line 947: You can probably use a smaller value safely.

```suggestion
    const MAX_STREAMS: u64 = 40;
```

neqo-transport/tests/server.rs
```diff
@@ -939,3 +941,76 @@ fn closed() {
...
+    let mut server = Server::new(
+        now(),
+        test_fixture::DEFAULT_KEYS,
+        test_fixture::DEFAULT_ALPN,
+        test_fixture::anti_replay(),
+        Box::new(AllowZeroRtt {}),
+        Rc::new(RefCell::new(FixedConnectionIdManager::new(9))),
+        ConnectionParameters::default()
+            .max_streams_bidi(MAX_STREAMS)
```
martinthomson, line 956: Would it be better to have this use StreamType?  That is, 

        ConnectionParameters::default().max_streams(StreamType:BiDi, MAX_STREAMS)

neqo-transport/tests/server.rs
```diff
@@ -939,3 +941,76 @@ fn closed() {
...
+    .expect("should create a server");
+
+    let mut client = default_client();
+    connect(&mut client, &mut server);
+
+    // Make sure that we can create MAX_STREAMS uni- and bidirectional streams.
+    for _i in 0..MAX_STREAMS {
+        client.stream_create(StreamType::UniDi).unwrap();
+        client.stream_create(StreamType::BiDi).unwrap();
+    }
```
martinthomson, line 970: You can factor out a function here:

```rust
fn can_create_streams(c: &mut Connection, t: StreamType, n: usize) {
    for _ in 0..n {
        c .stream_create(t).unwrap();
    }
    assert_eq!(client.stream_create(t), Err(Error::StreamLimitError));
}
```

neqo-transport/src/connection/mod.rs
martinthomson, line 228: This is small now, but it is going to grow.  Can you put this in a separate file?  You can `pub use` it so that it is available without the extra long name.

martinthomson (commented):
I like this.  A few things to think about cleaning up, but this looks solid.  And please file a followup (good-first-bug) issue for adding more command-line options to both server and client.  And to add controls for initial flow control limits.

neqo-transport/src/connection/mod.rs
ddragana: I will leave them. All the members are private, therefore we need them.

neqo-transport/src/connection/mod.rs
ddragana: We only need one member of ConnectionParameters  (soon we will need cc_algorithm as well) to be remembered in Connection. Therefore let's not hold the complete ConnectionParameters.

neqo-transport/src/connection_parameters.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    pub fn get_max_streams(&self, stream_type: StreamType) -> u64 {
+        match stream_type {
+            StreamType::BiDi => self.max_streams_bidi,
+            StreamType::UniDi => self.max_streams_uni,
+        }
+    }
+
+    pub fn max_streams(mut self, stream_type: StreamType, v: u64) -> Res<Self> {
```
martinthomson, line 60: I think that we can return `Self` here instead and just panic if the value is > 2^60.  That simplifies usage, and we can expect most users to respect the limit.  (Especially if you put in documentation that explains when this panics.)

That will remove a bunch of `.unwrap()` calls in usage.

martinthomson (commented):
Just one small thing to maybe fix, but please land this.  It's good.

martinthomson:
@ddragana, I want to add a switch to disable preferred address to this new structure, so I would like to have this landed before I land #1018.  Aside from the tiny tweak I suggest - which you could decide not to take - this is good to go.

PR #1057, opened 2020-12-10
Move client_close_with_long_reason_string to neqo-transport tests

KershawChang:
As Martin mentioned here [1], move this test to the proper place.

[1] https://github.com/mozilla/neqo/pull/1054#commitcomment-44961858


martinthomson (approved):
Thanks

PR #1059, opened 2020-12-10
Final version numbers

martinthomson:
This is a draft only because I don't want to land this now.  Everything checks out.


martinthomson:
I can see that the neqo-glue in Firefox has its own notion of [what versions are acceptable](https://searchfox.org/mozilla-central/rev/b32d4ca055ca9cf717be480df640f8970724a0ce/netwerk/socket/neqo_glue/src/lib.rs#82).  I think then we can merge this and stop maintaining the branch.  That is, assuming that this is OK.

ddragana (approved):

PR #1060, opened 2020-12-11
Test tparam parsing to prevent bugs like #705 (#706)

hawkinsw:
Add a test that checks for decoding failure when the ACTIVE_CONNECTION_ID_LIMIT is less than two.


martinthomson (approved):
Thanks Will, this is great.

PR #1064, opened 2020-12-14
Add a new http3 test server that always rejects zero rtt

KershawChang:
This patch simply adds `disallow_zerortt_server` to facilitate 0-Rtt reject test.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -868,6 +868,15 @@ impl ZeroRttChecker for AllowZeroRtt {
 }
 
 #[derive(Debug)]
+pub struct DisallowZeroRtt {}
+impl ZeroRttChecker for DisallowZeroRtt {
+    fn check(&self, _token: &[u8]) -> ZeroRttCheckResult {
+        qwarn!("DisallowZeroRtt reject 0-RTT");
+        ZeroRttCheckResult::Reject
+    }
+}
+
+#[derive(Debug)]
 struct ZeroRttCheckState {
     fd: *mut ssl::PRFileDesc,
     checker: Pin<Box<dyn ZeroRttChecker>>,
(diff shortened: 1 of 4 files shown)


neqo-transport/src/connection/tests/mod.rs
```diff
@@ -77,6 +77,22 @@ pub fn default_server() -> Connection {
     c
 }
 
+pub fn disallow_zerortt_server() -> Connection {
```
martinthomson, line 80: One-off things like this don't need functions to be added to the base fixture.  Just include this stuff inline.

neqo-transport/src/connection/tests/zerortt.rs
```diff
@@ -129,22 +125,8 @@ fn zero_rtt_send_reject() {
     client
         .enable_resumption(now(), token)
         .expect("should set token");
-    let mut server = Connection::new_server(
```
martinthomson, line 132: This is a different test case.  I would prefer that you add a new test case for your new thing.

neqo-crypto/src/agent.rs
```diff
@@ -867,6 +867,15 @@ impl ZeroRttChecker for AllowZeroRtt {
     }
 }
 
+#[derive(Debug)]
+pub struct DisallowZeroRtt {}
+impl ZeroRttChecker for DisallowZeroRtt {
```
martinthomson, line 872: This doesn't need to be public.  There is no situation, except for testing, where you would enable 0-RTT but then reject it.

martinthomson (changes requested):
If you are looking to implement rejection of 0-RTT for neqo-server, then the implementation of the rejection code should be local to that file.

ddragana:
This has bin resolved by #1150

Issue #1063, opened 2020-12-14
Update qlog crate version to 0.4.0


hawkinsw:
In order to solve #607 , we will need to update to the latest version of the qlog crate. I will make a PR for this change, if that's okay!

ddragana:
go ahead.

ddragana:
Fixed by #1066

Issue #607, opened 2020-05-07
qlog: Add mode that includes payload


agrover:
@vonasek mentioned that it might be useful if qlog included payload. This could help make using tools like Wireshark less needed.

Since payload is 1) large and 2) possibly sensitive, we probably don't want to do this by default when doing qlog.

hawkinsw:
Working on this. Currently blocked by https://github.com/cloudflare/quiche/issues/785

ddragana:
Thanks!

hawkinsw:
FYI: Patch accepted to fix qlog issue. Upstream will roll a new crate asap and then we can continue work on this issue.

larseggert:
We're using qlog much less than expected, so I'm not sure if this is worthwhile to pursue.

PR #1066, opened 2020-12-15
Update qlog to 0.4.0 (#1063)

hawkinsw:
In order to solve Issue #607, upgrade to qlog 0.4.0.


martinthomson (approved):
Thanks.  This was the easy part.  (Doing this as part of another change would also be fine.)

PR #1067, opened 2020-12-15
Rust 1.48 clippy changes

martinthomson:
There was one here that was genuinely useful, but some of the others were just annoying.  The insistence on using `Option::map_or` made some of the code worse, so I suppressed it.  Also, qpack usage deliberately overwrites specific error codes with `Result::map_err`, which now generates warnings that - rather than suppressing directly - I created a wrapper function for.

### DIFF
--- a/neqo-crypto/src/prio.rs
+++ b/neqo-crypto/src/prio.rs
@@ -5,7 +5,11 @@
 // except according to those terms.
 
 #![allow(dead_code, non_upper_case_globals, non_snake_case)]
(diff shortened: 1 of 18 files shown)


martinthomson:
Note that this includes a minor functional change.  There was a warning resulting from a usage of `Result::map_err` that turned out to be a violation of [this requirement](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-7.2.3-7).  I've updated the test to match.

ddragana (approved):

Issue #1052, opened 2020-12-02
Internal error


ddragana:
~1.8% of HTTP3 connection is closed with the INTERNAL_ERROR code. We should look into this.

martinthomson:
Is this down to us encountering an internal error, or receiving CONNECTION_CLOSE?

ddragana:
I think it is us (I need to investigate a bit).  This error is more common when necko gets ConnectionClosing event not ConnectionClosed event.

Looking at [this](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2020-12-01&include_spill=0&keys=closed!closing!__none__!__none__&max_channel_version=beta%252F84&measure=HTTP3_CONNECTTION_CLOSE_CODE&min_channel_version=null&processType=*&product=Firefox&sanitize=0&sort_by_value=0&sort_keys=submissions&start_date=2020-11-16&table=0&trim=0&use_submission_date=0).

ddragana:
I was thinking about adding a "comment" to each place we set this error and collect telemetry on this comments. Comments can be just a enumeration.

ddragana:
PR #1070 should help use narrow down this problem.

ddragana:
We have separated this errors a bit better and the general InternalErrors are not a problem any more.

PR #1069, opened 2020-12-18
Tweaks to frame writing code

martinthomson:
And some clippy fixups to make it build cleanly

The bug is in flow_mgr.  There is a guard in packet/mod.rs.  The rest is clippy complaining.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -9,7 +9,7 @@ use crate::cid::{ConnectionId, ConnectionIdDecoder, ConnectionIdRef, MAX_CONNECT
 use crate::crypto::{CryptoDxState, CryptoSpace, CryptoStates};
 use crate::{Error, Res};
 
-use neqo_common::{hex, hex_with_len, qtrace, Decoder, Encoder};
+use neqo_common::{hex, hex_with_len, qtrace, qwarn, Decoder, Encoder};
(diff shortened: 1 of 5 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -304,6 +304,12 @@ impl PacketBuilder {
 
     /// Build the packet and return the encoder.
     pub fn build(mut self, crypto: &mut CryptoDxState) -> Res<Encoder> {
+        if self.len() > self.limit {
```
ddragana, line 307: I think we should put a similar check and debug_assert! in every place we write into the buffer. If anyone catch the assert we will know exactly where it is happening. Also add a message to the assertion with useful values, e.g. self.len() and self.limit.
Also if we want to catch it sooner we could add  a different message to every place we close with InternalError and then we collect telemetry about it (that was my idea how to debug #1052 ). We can collect that telemetry on Nightly for a couple of weeks to catch all internal error or the most frequent errors.

ddragana:
Let's land this to stop the crash while we figure out what is causing it.

martinthomson:
The problem with adding a per-write check is that you have to overwrite all the `Encoder` methods in `PacketBuilder` as the latter knows the limit, but the methods are all currently accessed via `Deref` on the former.  We can do that, but let's do a follow-up for that.

PR #1070, opened 2020-12-18
Split internal errors for debugging purposes.

ddragana:
neqo-http3 events should use ConnectionError not Close error. This will give us more information in necko.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,13 @@ jobs:
             cargo run --bin neqo-client -- 'https://127.0.0.1:4433/'
 
       - run:
+          name: Internal Errors Check
+          command: |
+            all_internal_errors=($(find . -name '*.rs' -print | xargs -I {} sed '/Error::\(InternalError\|HttpInternal(\)/s/.*\(Error::[a-zA-Z]*([0-9]*)\)/\1/')); \
+            unique_internal_errors=($(for e in "${all_internal_errors[@]}"; do echo "$e"; done | sort | uniq)); \
+            [[ "${#all_internal_errors[@]}" -eq "${#unique_internal_errors[@]}" ]]
+
+      - run:
           name: "Prepare for Caching"
           command: "git gc --auto"
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,9 +82,7 @@ jobs:
       - run:
           name: Internal Errors Check
           command: |
-            all_internal_errors=($(find . -name '*.rs' -print | xargs -I {} sed '/Error::\(InternalError\|HttpInternal(\)/s/.*\(Error::[a-zA-Z]*([0-9]*)\)/\1/')); \
-            unique_internal_errors=($(for e in "${all_internal_errors[@]}"; do echo "$e"; done | sort | uniq)); \
-            [[ "${#all_internal_errors[@]}" -eq "${#unique_internal_errors[@]}" ]]
+            ./check_internal_errors
 
       - run:
           name: "Prepare for Caching"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,8 +81,7 @@ jobs:
 
       - run:
           name: Internal Errors Check
-          command: |
-            ./check_internal_errors
+          command: ./check_internal_errors
 
       - run:
           name: "Prepare for Caching"
--- a/check_internal_errors
+++ b/check_internal_errors
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd "$(git rev-parse --show-toplevel)"
+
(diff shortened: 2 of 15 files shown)


neqo-http3/src/connection.rs
```diff
@@ -47,8 +47,8 @@ pub enum Http3State {
     ZeroRtt,
     Connected,
     GoingAway(u64),
-    Closing(CloseError),
-    Closed(CloseError),
+    Closing(ConnectionError),
+    Closed(ConnectionError),
```
ddragana, line 51: I do not know why this was not using ConnectionError. It should have been.

neqo-http3/src/lib.rs
```diff
@@ -49,6 +49,8 @@ pub enum Error {
     HttpGeneralProtocol,
     HttpGeneralProtocolStream, //this is the same as the above but it should only close a stream not a connection.
     HttpInternal,
+    HttpInternal1,
+    HttpInternal2,
```
martinthomson, line 53: This is going to need some commentary.

neqo-transport/src/lib.rs
```diff
@@ -59,7 +59,12 @@ const ERROR_AEAD_LIMIT_REACHED: TransportError = 15;
 #[allow(clippy::pub_enum_variant_names)]
 pub enum Error {
     NoError,
-    InternalError,
+    InternalError1,
+    InternalError2,
+    InternalError3,
+    InternalError4,
+    InternalError5,
+    InternalError6,
```
martinthomson, line 67: As above.

martinthomson (changes requested):
I think that you might want to parameterize these rather than have InternalError\[N\].  Then if you really want to split things up, we can add a lint check that verifies that each is only used once.

martinthomson:
What I'm talking about here is probably having a definition like:

```rust
enum Error {
    // ...
    InternalError(String, u32),
}
```

Then you can have a macro in lib.rs that goes something like:

```rust
macro_rules! internal_error {
  () => { ::neqo_http3::Error::InternalError(file!(), line!()) }
}
```

Then it is relatively easy to add these as necessary and we capture the context automatically.

ddragana:
My main goal was to make this easy to capture in a telemetry probe.
You are right it is better to have something like InternalError(u32)
I am not sure about putting the context, because we will need to parse them and classify them before recording telemetry probe. We would also need to update that code every time we update neqo.

neqo-http3/src/connection.rs
martinthomson: Easy mistake to make :)

martinthomson (commented):
I think that we are going to need some enforcement for this.

How about adding this to CI:

```
all_internal_errors=($(find . -name '*.rs' -print | xargs -I {} sed '/Error::\(InternalError\|HttpInternal(\)/s/.*\(Error::[a-zA-Z]*([0-9]*)\)/\1/')); \
unique_internal_errors=($(for e in "${all_internal_errors[@]}"; do echo "$e"; done | sort | uniq)); \
[[ "${#all_internal_errors[@]}" -eq "${#unique_internal_errors[@]}" ]]
```

(Not completely tested.)

neqo-http3/src/lib.rs
```diff
@@ -209,7 +211,7 @@ impl From<QpackError> for Error {
     fn from(err: QpackError) -> Self {
         match err {
             QpackError::ClosedCriticalStream => Error::HttpClosedCriticalStream,
-            QpackError::InternalError => Error::HttpInternal,
+            QpackError::InternalError(i) => Error::HttpInternal(10 + i),
```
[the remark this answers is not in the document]
ddragana: it does not need to be contiguous.
I do not expect that w will have a very high number of internal error.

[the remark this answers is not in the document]
martinthomson: If it doesn't need to be contiguous, just make it a big number.

.circleci/config.yml
```diff
@@ -79,6 +79,11 @@ jobs:
             cargo run --bin neqo-server -- '0.0.0.0:4433' & spid=$!; \
             cargo run --bin neqo-client -- 'https://127.0.0.1:4433/'
 
+      - run:
+          name: Internal Errors Check
+          command: |
+            ./check_internal_errors
```
martinthomson, line 85:
```suggestion
          command: ./check_internal_errors
```

check_internal_errors
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
martinthomson, line 1: This might be better added to the `hooks` directory.

Can you add a call to this to `hooks/pre-commit` ?  It shouldn't take long to run, so it's a good check to have there.

martinthomson (commented):
Some very minor comments only now.

PR #1074, opened 2020-12-21
Cubic

ddragana:


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -180,11 +191,21 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         if self.congestion_window >= self.ssthresh {
-            let (cwnd, acked_bytes) = self
-                .cc_algorithm
-                .on_packets_acked(self.congestion_window, self.acked_bytes);
-            self.congestion_window = cwnd;
-            self.acked_bytes = acked_bytes;
+            let bytes =
+                self.cc_algorithm
+                    .on_packets_acked(self.congestion_window, new_acked, min_rtt, now);
+            // If enough creadit has been accumulated already, apply them gently.
```
ddragana, line 197: This is how the linux kernel is handling this.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,178 @@
...
+            curr_cwnd_f64 / (target - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64
+        } else {
+            100.0 * curr_cwnd_f64
+        };
+
+        let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
+        while self.tcp_acked_bytes > tcp_cnt {
+            self.tcp_acked_bytes -= tcp_cnt;
+            self.estimated_tcp_cwnd += MAX_DATAGRAM_SIZE_F64;
+        }
```
ddragana, line 135: I have decided to count bytes instead of a proportional increase.
(side effect: it was easier to calculate rounds in tcp phase in tests.)

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,178 @@
...
+            let tcp_cnt =
+                curr_cwnd_f64 / (self.estimated_tcp_cwnd - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64;
+            if cnt > tcp_cnt {
+                cnt = tcp_cnt;
+            }
+        }
+
+        // Limit increas to max 1 MSS per 2 ack packets.
+        cnt = cnt.max(2.0 * MAX_DATAGRAM_SIZE_F64);
+        cnt as usize
```
ddragana, line 147: This 'as' can be fix when FloatToInt is available in rust stable. Other way to avoid this is to count bytes in f64 as well(in classic_cc).

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    // This will start with TCP phase.
+    // in this phase cwnd is increase by CUBIC_ALPHA every RTT. We can look at it as
+    // increase of MAX_DATAGRAM_SIZE every 1 / CUBIC_ALPHA RTTs.
+    // The phase will end when cwnd calculated with cubic equation is equal to TCP estimate:
+    // CUBIC_C * (n * RTT / CUBIC_ALPHA)^3 * MAX_DATAGRAM_SIZE = n * MAX_DATAGRAM_SIZE
+    // from this n = sqrt(CUBIC_ALPHA^3/ (CUBIC_C * RTT^3)).
+    let num_tcp_increases = (CUBIC_ALPHA.powi(3) / (CUBIC_C * RTT.as_secs_f64().powi(3)))
+        .sqrt()
+        .round() as u64;
```
ddragana, line 94: I decided to leave it as 'as u64'. This can be avoid by using while loop, but I think it is not worth it.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,182 @@
...
+                now
+            })
+        .as_secs_f64();
+        let target = self.w_cubic(time_ca);
+
+        let mut cnt = if target > curr_cwnd_f64 {
+            curr_cwnd_f64 / (target - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64
+        } else {
+            100.0 * curr_cwnd_f64
+        };
```
martinthomson, line 133: I'm not really following the point of `cnt` here.  As we get larger windows that amount is going to be small.  So the resulting value is going to be quite large on the upper branch.  How does 100 relate to the MTU?

If this is the increase in the window, this seems to be backwards.  RFC 8312 says that you increase the window by `(W_cubic(t+RTT) - cwnd)/cwnd`, but this equation is the inverse of that: `cwnd/(W_cubic(t+RTT) - cwnd)`.

See https://ntap.github.io/rfc8312bis/draft-eggert-tcpm-rfc8312bis.html#name-convex-region

BTW, if `cnt` is the increase to the congestion window, then call it `cwnd_increase` or something like that.  As the output of this function is not documented, a more descriptive variable would help.

neqo-transport/src/cc/new_reno.rs
```diff
@@ -29,16 +29,25 @@ impl Display for NewReno {
...
-        }
-        (curr_cwnd, acked_bytes)
+    fn on_packets_acked(
+        &mut self,
+        curr_cwnd: usize,
+        _acked_bytes: usize,
+        _min_rtt: Duration,
+        _now: Instant,
+    ) -> usize {
+        curr_cwnd
```
martinthomson, line 39: OK, I clearly don't understand what is going on here, because the same function for Cubic returns a value that is limited to two datagrams.

martinthomson (commented):
The plumbing here looks pretty good, but I have some questions about the tests and the main Cubic function.

I haven't reviewed the cubic-specific tests in detail.  I want to understand what is going on more first.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[allow(clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_sign_loss)]
+    fn on_packets_acked(
+        &mut self,
+        curr_cwnd: usize,
+        acked_bytes: usize,
+        min_rtt: Duration,
+        now: Instant,
+    ) -> usize {
+        let curr_cwnd_f64 = f64::try_from(u32::try_from(curr_cwnd).unwrap()).unwrap();
```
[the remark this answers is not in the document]
ddragana: I was thinking about add such a conversion function, but I do not really think we can overflow u32. I expect that we will hit some other limits, e.g. CPU, socket buffers, etc. (we will hit this limit with something like 8GB * 250ms)

... ok, I will add the conversion. The other workaround this problem that I could think of was to keep curr_cwnd separate for u64 and f64 to avoid the conversion. The curr_cwnd is increase and decrease in one place so it would not be hard to keep them sync.... this is not worth the effort.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,182 @@
...
+        self.tcp_acked_bytes += f64::try_from(u32::try_from(acked_bytes).unwrap()).unwrap();
```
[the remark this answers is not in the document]
ddragana: yes this is only newly acked byte, so this conversion is fine.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,182 @@
...
+            } else {
+                now
+            })
+        .as_secs_f64();
+        let target = self.w_cubic(time_ca);
+
+        let mut cnt = if target > curr_cwnd_f64 {
+            curr_cwnd_f64 / (target - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64
+        } else {
+            100.0 * curr_cwnd_f64
```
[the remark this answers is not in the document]
ddragana: This is from the linux implementation. It is a corner case, I do not know how to trigger it. This should cause a very small increase in cwnd.
I will explain cnt than this will be more understandable:

cnt in NewReno is always cwnd, i.e. a congestion controller need to receive cwnd of acked bytes to increase cwnd by 1 MSS. For cubic it is calculated depending on W_cubic and tcp estimate, but is still "how many acked bytes it needs to receive to increase cnwd by 1 MSS.
This is leaned on the linux implementation. I may change it to be more similar to the RFC, I think that will make increase slightly slower(tiny, tiny slower).

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+#[test]
+fn tcp_phase() {
+    let mut cubic = ClassicCongestionControl::new(Cubic::default());
+
+    // change to congestion avoidance state.
+    cubic.set_ssthresh(1);
+
+    let mut now = now();
+    let start_time = now;
+    // helper variables to remember the next packet number ro be send/acked.
```
martinthomson, line 82:
```suggestion
    // helper variables to remember the next packet number to be sent/acked.
```

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    while cwnd_rtt_start == cubic.cwnd() {
+        num_acks += 1;
+        ack_packet(&mut cubic, next_pn_ack, now);
+        next_pn_ack += 1;
+        next_pn_send = fill_cwnd(&mut cubic, next_pn_send, now);
+        now += time_increase;
+    }
+
+    // Make sure that the increase is not according to TCP equation, i.e tha it took
```
martinthomson, line 133:
```suggestion
    // Make sure that the increase is not according to TCP equation, i.e., that it took
```

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+        now(),             // time sent
+        true,              // ack eliciting
+        Vec::new(),        // tokens
+        MAX_DATAGRAM_SIZE, // size
+    );
+    cc.on_packets_lost(None, None, PTO, &[p_lost]);
+}
+
+#[test]
+fn tcp_phase() {
```
martinthomson, line 74: This test does too much.  It tests that the TCP phase works, then it tests that the increases is faster than TCP when it exits, then it tests that it increases at precisely the cubic rate.

This is at least three tests.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        assert_eq!(cubic.cwnd() - cwnd_rtt_start, MAX_DATAGRAM_SIZE);
+    }
+
+    // The next increase will be according to the cubic equation.
+
+    let cwnd_rtt_start = cubic.cwnd();
+    // calculate new time_increase.
+    let time_increase = RTT / u32::try_from(cwnd_rtt_start / MAX_DATAGRAM_SIZE).unwrap();
```
martinthomson, line 122: This is an odd formulation.  This seems to be the RTT divided into the number of whole datagrams in the congestion window.  Explain why this is right, as opposed to something else.  It seems like you are trickling acknowledgements over an RTT until the congestion window increases once.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    while cwnd_rtt_start_after_tcp == cubic.cwnd() {
+        num_acks2 += 1;
+        ack_packet(&mut cubic, next_pn_ack, now);
+        next_pn_ack += 1;
+        next_pn_send = fill_cwnd(&mut cubic, next_pn_send, now);
+        now += time_increase;
+    }
+
+    now -= time_increase; // "- time_increase" because it is added above.
```
martinthomson, line 160: I can't see how this value is used anywhere.  If you want to avoid the extra increment, then add the extra to `now` it before injecting the acknowledgment.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+        / CUBIC_ALPHA)
+        .round() as u64;
+    assert!(num_acks < expected_ack_tcp_increase);
+
+    // This first increase after a TCP phase may be shorter than what it would take by a regular cubic phase,
+    // because of the proper byte counting and the credit it already had before entering this phase. Therefore
+    // We will perform another round and compare it to expected increase using the cubic equation.
+
+    let cwnd_rtt_start_after_tcp = cubic.cwnd();
+    let elapse_time = now - start_time - time_increase; // "- time_increase" because it is added above.
```
martinthomson, line 146: As below, you can avoid the extra subtraction by increasing now before calling ack_packet/fill_cwnd.  That also avoids having the first ACK being right on top of the packet being sent.

Also "elapsed_time".

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+
+#[test]
+fn cubic_phase() {
+    let mut cubic = ClassicCongestionControl::new(Cubic::default());
+    // Set last_max_cwnd to a higher number make sure that cc is the cubic phase (cwnd is calculated by the cubic equation).
+    cubic.set_last_max_cwnd(CWND_INITIAL_10_F64);
+    // Set ssthresh to something small to make sure that cc is in the congection avoidance phase.
+    cubic.set_ssthresh(1);
+    let mut now = now();
+    // helper variables to remember the next packet number ro be send/acked.
```
martinthomson, line 194:
```suggestion
```

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+                .round() as usize;
+
+        assert_close(cubic.cwnd(), expected);
+    }
+    assert_eq!(cubic.cwnd(), CWND_INITIAL_10);
+}
+
+fn assert_close(cwnd: usize, expects_cwnd: usize) {
+    if cwnd >= expects_cwnd {
+        assert!(cwnd <= expects_cwnd + MAX_DATAGRAM_SIZE);
```
martinthomson, line 229: I think that you want `<` rather than `<=` here.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+            (CUBIC_C * ((now - epoch_start).as_secs_f64() - k).powi(3) * MAX_DATAGRAM_SIZE_F64
+                + CWND_INITIAL_10_F64)
+                .round() as usize;
+
+        assert_close(cubic.cwnd(), expected);
+    }
+    assert_eq!(cubic.cwnd(), CWND_INITIAL_10);
+}
+
+fn assert_close(cwnd: usize, expects_cwnd: usize) {
```
martinthomson, line 227: Maybe `assert_within(cwnd, expected, MAX_DATAGRAM_SIZE)` instead.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+#[test]
+fn congestion_event_slow_start() {
+    let mut cubic = ClassicCongestionControl::new(Cubic::default());
+
+    let _ = fill_cwnd(&mut cubic, 0, now());
+    ack_packet(&mut cubic, 0, now());
+
+    assert!((cubic.last_max_cwnd() - 0.0).abs() < f64::EPSILON);
```
martinthomson, line 242: You can use assert_within here if you make it generic.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+    // Set last_max_cwnd to someting higher than cwnd so that the fast convergence is triggered.
+    cubic.set_last_max_cwnd(CWND_INITIAL_10_F64);
+
+    let _ = fill_cwnd(&mut cubic, 0, now());
+    ack_packet(&mut cubic, 0, now());
+
+    assert!((cubic.last_max_cwnd() - CWND_INITIAL_10_F64).abs() < f64::EPSILON);
+    assert_eq!(cubic.cwnd(), CWND_INITIAL);
+
+    // Trigger a congestion_event in slow start phase
```
martinthomson, line 294: This comment is not accurate.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+#[test]
+fn congestion_event_congestion_avoidance_2() {
+    let mut cubic = ClassicCongestionControl::new(Cubic::default());
+
+    // Set ssthresh to something small to make sure that cc is in the congection avoidance phase.
+    cubic.set_ssthresh(1);
+
+    // Set last_max_cwnd to someting higher than cwnd so that the fast convergence is triggered.
```
martinthomson, line 285:
```suggestion
    // Set last_max_cwnd to something higher than cwnd so that the fast convergence is triggered.
```

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+#[test]
+fn congestion_event_congestion_avoidance() {
+    let mut cubic = ClassicCongestionControl::new(Cubic::default());
+
+    // Set ssthresh to something small to make sure that cc is in the congection avoidance phase.
+    cubic.set_ssthresh(1);
+
+    // Set last_max_cwnd to someting smaller than cwnd so that the fast convergence is not triggered.
```
martinthomson, line 263:
```suggestion
    // Set last_max_cwnd to something smaller than cwnd so that the fast convergence is not triggered.
```

martinthomson (commented):
This is generally looking solid.  I have a bunch of comments, but mostly this will be about cleanup and making the code easier to understand.  Cubic is pretty simple, but it has a history of terrible code and documentation.  I'd rather not have that continue in our code.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,193 @@
...
+        if self.estimated_tcp_cwnd > curr_cwnd_f64 && self.estimated_tcp_cwnd > target {
+            let cnt_tcp_equation =
+                curr_cwnd_f64 / (self.estimated_tcp_cwnd - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64;
+            if cnt > cnt_tcp_equation {
+                cnt = cnt_tcp_equation;
+            }
+        }
+
+        // Limit increase to max 1 MSS per 2 ack packets.
+        cnt = cnt.max(2.0 * MAX_DATAGRAM_SIZE_F64);
```
martinthomson, line 155: I missed this last time around, but it might be worth pointing out that this limit is a direct result of the limit of 1.5cwnd in the target calculation.  You can draw a straight line from that to this, but it's non-obvious.  Also, I'm not sure that 2 is necessarily the only value we need to consider here, so it needs a constant.

BTW, this changes the function into an exponential function.  The increase goes from linear (the TCP region), to cubic (the W_max calculation), to exponential.  It's not really accurate to call it cubic, other than to note that a cubic function is used to smooth the transition from linear to exponential.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -180,11 +198,24 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
-                .cc_algorithm
-                .on_packets_acked(self.congestion_window, self.acked_bytes);
-            self.congestion_window = cwnd;
-            self.acked_bytes = acked_bytes;
+            // The following function return the amount acked bytes a controller needs
+            // to collect to be allowed to increase its cwnd by MAX_DATAGRAM_SIZE.
+            let byte_cnt =
+                self.cc_algorithm
+                    .on_packets_acked(self.congestion_window, new_acked, min_rtt, now);
+            // If enough credit has been accumulated already, apply them gradually.
```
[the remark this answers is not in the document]
ddragana: I prefer leaving at is this. I find it more understandable: If a rate suddenly changes the old accumulative bytes should increase rate at most by 1MSS. Newly accumulated byte are respected.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -562,153 +514,186 @@ mod tests {
...
-            false
-        } else if cc.cwnd() == CWND_MIN {
-            true
-        } else {
-            panic!("unexpected cwnd");
+    fn persistent_congestion(lost_packets: &[SentPacket], persistent_expected: bool) {
+        for cc_alg in &[
+            CongestionControlAlgorithm::NewReno,
+            CongestionControlAlgorithm::Cubic,
+        ] {
```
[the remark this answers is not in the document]
ddragana: I will leave as it is. I implemented something similar before, but some of the test functions that use the function got a bit hard to read because the function is called with a different  cc algorithm and multiple different configurations.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.last_max_cwnd = if curr_cwnd_f64 + MAX_DATAGRAM_SIZE_F64 < self.last_max_cwnd {
+            curr_cwnd_f64 * CUBIC_FAST_CONVERGENCE
+        } else {
+            curr_cwnd_f64
+        };
+        self.ca_epoch_start = None;
+        (
+            max(
+                curr_cwnd * CUBIC_BETA_USIZE_QUOTIENT / CUBIC_BETA_USIZE_DIVISOR,
+                CWND_MIN,
```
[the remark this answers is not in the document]
ddragana: oh, the caller already does that.

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,193 @@
...
+            .as_secs_f64();
+        let target = self.w_cubic(time_ca);
+
+        let mut cnt = if target > curr_cwnd_f64 {
+            curr_cwnd_f64 / (target - curr_cwnd_f64) * MAX_DATAGRAM_SIZE_F64
+        } else {
+            100.0 * curr_cwnd_f64
+        };
+
+        let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
```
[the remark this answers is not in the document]
ddragana: the RFC says that additive factor is CUBIC_ALPHA segments per RTT, this means that you need 1/CUBIC_ALPHA * RTT to increase cwnd by a MSS, or in byte count cwnd/CUBIC_ALPHA

neqo-transport/src/cc/tests/cubic.rs
ddragana: it is not always easy to construct the starting point of a test (we will need to expose more functions to modify cc internal values). 
I wrote the test to show that: tcp phase is exited exactly at the point expected and that the increase after that is not using tcp equation, but cubic equation. Therefore you need all 3 pieces in one test.

neqo-transport/src/cc/tests/cubic.rs
ddragana: it simulates that packets and their ack are evenly distributed over a RTT. time_increase is time between 2 acks if they are sent RTT/num_of_full_packets_per_rtt apart.

neqo-transport/src/cc/tests/cubic.rs
ddragana: I forgot to remove it. I was calculating expected_ack_cubic_increase a bit previously and now was used for that.
I will remove it.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,301 @@
...
+        .sqrt()
+        .round() as u64;
+
+    for _ in 0..num_tcp_increases {
+        let cwnd_rtt_start = cubic.cwnd();
+        //Expected acks during a period of RTT / CUBIC_ALPHA.
+        let acks = (f64::try_from(i32::try_from(cwnd_rtt_start).unwrap()).unwrap()
+            / MAX_DATAGRAM_SIZE_F64
+            / CUBIC_ALPHA)
+            .round() as u64;
```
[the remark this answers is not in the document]
ddragana: num_tcp_increases should be floor.

The other one should be round. Because it is not always an exact number and there is a couple of round of acks we need to use round, otherwise if we use ceil every time we will exit too early.

neqo-transport/src/cc/classic_cc.rs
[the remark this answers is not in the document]
martinthomson: The problem with this code is that it is unclear that it limits the increase.  It looks like it increases the rate by far more than 1MSS.  If there is left over bytes, you get 1 increase.  And then you have d*MSS.  You have to look at the congestion controller to understand that this is never going to be anything other than 0 or 1.

(You don't need two iterations of the `acked_bytes > bytes_for_increase` in the above.  One is enough.  I've edited for that.)

So I find the above far clearer about the important property: that we only increase by at most 1MSS.

neqo-transport/src/cc/classic_cc.rs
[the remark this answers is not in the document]
martinthomson: I'm not sure if we aren't talking past each other, so I will propose a change.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -562,153 +519,186 @@ mod tests {
...
+            let persistent = if cc.cwnd() == expected_cwnd_no_persistent_congestion {
+                false
+            } else if cc.cwnd() == CWND_MIN {
+                true
+            } else {
+                panic!("unexpected cwnd");
+            };
+            assert_eq!(persistent, persistent_expected);
         }
     }
```
martinthomson, line 560:
```suggestion
    fn persistent_congestion_by_algorithm(cc_alg: CongestionControlAlgorithm, reduced_cwnd: usize, lost_packets: &[SentPacket], persistent_expected: bool) {
        let mut cc = congestion_control(*cc_alg);
        for p in lost_packets {
            cc.on_packet_sent(p);
        }

        cc.on_packets_lost(Some(now()), None, PTO, lost_packets);
        let persistent = if cc.cwnd() == reduced_cwnd {
             false
        } else if cc.cwnd() == CWND_MIN {
            true
        } else {
            panic!("unexpected cwnd");
        };
        assert_eq!(persistent, persistent_expected);
    }

    fn persistent_congestion(lost_packets: &[SentPacket], persistent_expected: bool) {
        persistent_congestion_by_algorithm(CongestionControlAlgorithm::NewReno, CWND_INITIAL / 2, lost_packets, persistent_expected);
        persistent_congestion_by_algorithm(CongestionControlAlgorithm::Cubic, CWND_INITIAL * CUBIC_BETA_USIZE_QUOTIENT / CUBIC_BETA_USIZE_DIVISOR, lost_packets, persistent_expected);
    }
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -76,8 +76,25 @@ impl State {
...
+    /// to collect before increasing its cwnd by `MAX_DATAGRAM_SIZE`.
+    fn bytes_for_cwnd_increase(
+        &mut self,
+        curr_cwnd: usize,
+        new_acked_bytes: usize,
+        min_rtt: Duration,
+        now: Instant,
+    ) -> usize;
+    /// This function is called when a congestion event has beed detected and it
+    /// returns new (decreased) values of `curr_cwnd` and `acked_bytes`.
```
martinthomson, line 90:
```suggestion
    /// returns new (decreased) values of `curr_cwnd` and `acked_bytes`.
    /// This value can be very small; the calling code is responsible for ensuring that the
    /// congestion window doesn't drop below the minimum of `CWND_MIN`.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,197 @@
...
+
+// CUBIC_BETA = 0.7;
+pub const CUBIC_BETA_USIZE_QUOTIENT: usize = 7;
+pub const CUBIC_BETA_USIZE_DIVISOR: usize = 10;
+
+/// The fast convergence ratio further reduces the congestion window when a congestion event
+/// occurs before reaching the previous `W_max`.
+pub const CUBIC_FAST_CONVERGENCE: f64 = 0.85; // (1.0 + CUBIC_BETA) / 2.0;
+
+const MIN_INCREASE: f64 = 2.0;
```
martinthomson, line 31:
```suggestion
/// The minimum number of multiples of the datagram size that need
/// to be received to cause an increase in the congestion window.
/// When there is no loss, Cubic can return to exponential increase, but
/// this value reduces the magnitude of the resulting growth by a constant factor.
/// A value of 1.0 would mean a return to the rate used in slow start.
const EXPONENTIAL_GROWTH_REDUCTION: f64 = 2.0;
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,197 @@
...
+        // in the specification (Sections 4.4 and 4.5).
+        // The amount of data required therefore reduces asymptotically as the target increases.
+        // If the target is not significantly higher than the congestion window, require a very large
+        // amount of acknowledged data (effectively block increases).
+        let mut acked_to_increase =
+            MAX_DATAGRAM_SIZE_F64 * curr_cwnd_f64 / (target_cwnd - curr_cwnd_f64).max(1.0);
+
+        // Limit increase to max 1 MSS per MIN_INCREASE ack packets.
+        // This is basicaly limiting target to (1 + 1/MIN_INCREASE)cwnd.
+        acked_to_increase = acked_to_increase.max(MIN_INCREASE * MAX_DATAGRAM_SIZE_F64);
```
martinthomson, line 160:
```suggestion
        acked_to_increase = acked_to_increase.max(EXPONENTIAL_GROWTH_REDUCTION * MAX_DATAGRAM_SIZE_F64);
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,197 @@
...
+        // of `MAX_DATAGRAM_SIZE` to match the increase of `target - cwnd / cwnd` as defined
+        // in the specification (Sections 4.4 and 4.5).
+        // The amount of data required therefore reduces asymptotically as the target increases.
+        // If the target is not significantly higher than the congestion window, require a very large
+        // amount of acknowledged data (effectively block increases).
+        let mut acked_to_increase =
+            MAX_DATAGRAM_SIZE_F64 * curr_cwnd_f64 / (target_cwnd - curr_cwnd_f64).max(1.0);
+
+        // Limit increase to max 1 MSS per MIN_INCREASE ack packets.
+        // This is basicaly limiting target to (1 + 1/MIN_INCREASE)cwnd.
```
martinthomson, line 159:
```suggestion
        // Limit increase to max 1 MSS per EXPONENTIAL_GROWTH_REDUCTION ack packets.
        // This effectively limits target_cwnd to `(1 + 1/EXPONENTIAL_GROWTH_REDUCTION)*cwnd`.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -0,0 +1,197 @@
...
+        acked_to_increase = acked_to_increase.max(MIN_INCREASE * MAX_DATAGRAM_SIZE_F64);
+        acked_to_increase as usize
+    }
+
+    fn reduce_cwnd(&mut self, curr_cwnd: usize, acked_bytes: usize) -> (usize, usize) {
+        let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
+        // Fast Convergence
+        // If congestion event occurs before the maximum congestion window before the last congestion event,
+        // we reduce the the maximum congestion window and therby w_max.
```
martinthomson, line 168:
```suggestion
        // we reduce the the maximum congestion window and thereby W_max.
```
or `w_max()`.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,303 @@
...
+            (CUBIC_C * ((now - epoch_start).as_secs_f64() - k).powi(3) * MAX_DATAGRAM_SIZE_F64
+                + CWND_INITIAL_10_F64)
+                .round() as usize;
+
+        assert_within(cubic.cwnd(), expected, &MAX_DATAGRAM_SIZE);
+    }
+    assert_eq!(cubic.cwnd(), CWND_INITIAL_10);
+}
+
+fn assert_within<T: Sub<Output = T> + PartialOrd>(value: T, expected: T, margin: &T) {
```
martinthomson, line 224: You can make `T: Sub<...> + PartialOrd + Copy` and avoid taking a reference for `margin`.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        assert_eq!(cubic.cwnd() - cwnd_rtt_start, MAX_DATAGRAM_SIZE);
+    }
+
+    // The next increase will be according to the cubic equation.
+
+    let cwnd_rtt_start = cubic.cwnd();
+    // cwnd_rtt_start has change, therefore calculate new time_increase (the time
+    // between acks if they are ideally paced over a RTT.).
```
martinthomson, line 127:
```suggestion
    // between acks if they are ideally paced over a RTT).
```

neqo-transport/src/cc/tests/cubic.rs
martinthomson: I think that I can live with this.  It's not ideal, but it's also not that long relative to some of the other tests.

martinthomson (commented):
Let's do this.  I have some suggestions, but nothing that can block landing.

Land once you are happy with it.

neqo-transport/src/cc/classic_cc.rs
ddragana: We were talking past each other. I will take this suggestion.

PR #1082, opened 2021-01-13
Try loopback

martinthomson:
The last one was generating an undetected error, which isn't great.  Test this out as a draft.


martinthomson:
This was weird: I couldn't use IPv6 for reasons I couldn't trace down, but IPv4 works fine.  I confirmed this in the docker image locally; nothing I could find seemed to indicate why this shouldn't work.

ddragana (approved):

PR #1085, opened 2021-01-19
Add a sepparate InternalError for every step of encoding a packet.

ddragana:


martinthomson (approved):
This is why we needed the linting rule :) Looks good.

I hope to be able to remove these at some point.  Can you open an issue to track backing this change out?

ddragana:
I have opened issue #1086

PR #1094, opened 2021-02-11
Sketch of priority API

martinthomson:
This adds the ability to prioritize streams.

The most important aspect of this is that the default behaviour here
prioritizes retransmission over transmission.

This comes at a small cost: we iterate through all the active streams 5
times rather than just once.  I've tried to mitigate that by adding an
early exit: if we fill up the packet, we stop looking to add frames.

This isn't complete as it doesn't include any tests that verify that
prioritization works as expected.  And the HTTP fetch API needs priority
hooks too.

### DIFF
--- a/neqo-transport/src/connection/tests/priority.rs
+++ b/neqo-transport/src/connection/tests/priority.rs
@@ -1,0 +1,212 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use super::super::{Connection, Error, Output};
+use super::{connect, default_client, default_server, fill_cwnd, maybe_authenticate};
+use crate::send_stream::{RetransmissionPriority, TransmissionPriority};
+use crate::{ConnectionEvent, StreamType};
+
+use neqo_common::event::Provider;
+use test_fixture::{self, now};
+
+const BLOCK_SIZE: usize = 4_096;
+
+fn fill_stream(c: &mut Connection, id: u64) {
+    loop {
+        if c.stream_send(id, &[0x42; BLOCK_SIZE]).unwrap() < BLOCK_SIZE {
+            return;
+        }
+    }
+}
+
+/// A receive stream cannot be prioritized (yet).
+#[test]
+fn receive_stream() {
+    const MESSAGE: &[u8] = b"hello";
+    let mut client = default_client();
+    let mut server = default_server();
+    connect(&mut client, &mut server);
+
+    let id = client.stream_create(StreamType::UniDi).unwrap();
+    assert_eq!(MESSAGE.len(), client.stream_send(id, MESSAGE).unwrap());
+    let dgram = client.process_output(now()).dgram();
+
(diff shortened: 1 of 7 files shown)


martinthomson:
I added some tests.  This should at least be ready for feedback.

neqo-transport/src/send_stream.rs
```diff
@@ -29,6 +30,97 @@ use crate::{AppError, Error, Res};
...
+            },
+            RetransmissionPriority::Same => self,
+        }
+    }
+}
+
+/// If data is lost, this determines the priority that applies to retransmissions
+/// of that data.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RetransmissionPriority {
```
ddragana, line 109: This also looks reasonable.

neqo-transport/src/send_stream.rs
```diff
@@ -29,6 +30,97 @@ use crate::{AppError, Error, Res};
 
 pub const SEND_BUFFER_SIZE: usize = 0x10_0000; // 1 MiB
 
+/// The priority that is assigned to sending data for the stream.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TransmissionPriority {
```
ddragana, line 35: This order looks reasonable to me.

neqo-transport/src/connection/tests/priority.rs
```diff
@@ -0,0 +1,212 @@
...
+    let _ = server.events().count(); // Drain events.
+
+    let normal = client.stream_create(StreamType::UniDi).unwrap();
+    fill_stream(&mut client, normal);
+
+    let dgram = client.process(ack, now).dgram();
+    assert_eq!(client.stats().lost, 1); // Client should have noticed the loss.
+    server.process_input(dgram.unwrap(), now);
+
+    // Only the "normal" stream has data as the retransmission of the data from
```
ddragana: here it should be "low"

neqo-transport/src/connection/tests/priority.rs
```diff
@@ -78,10 +81,60 @@ fn relative() {
...
     for e in server.events() {
         if let ConnectionEvent::RecvStreamReadable { stream_id } = e {
-            assert_ne!(stream_id, normal);
+            assert_ne!(stream_id, id_normal);
+        }
+    }
+}
+
+/// Higher priority streams get sent ahead of lower ones, even when
+/// the higher priority stream is written to later.
```
ddragana, line 93: Please update this comment. It is the same as in the previous test, but something else is tested here.

neqo-transport/src/connection/tests/priority.rs
```diff
@@ -110,27 +163,27 @@ fn repairing_loss() {
...
+    let id_normal = client.stream_create(StreamType::UniDi).unwrap();
+    fill_stream(&mut client, id_normal);
 
     let dgram = client.process(ack, now).dgram();
     assert_eq!(client.stats().lost, 1); // Client should have noticed the loss.
     server.process_input(dgram.unwrap(), now);
 
-    // Only the "normal" stream has data as the retransmission of the data from
-    // the lost packet is more important than the high priority stream.
+    // Only the normal priority stream has data as the retransmission of the data
```
ddragana, line 181: this should be "low priority stream".

ddragana (commented):
Looks good, fix comments and land.

PR #1095, opened 2021-02-15
Dynamic timer resolution increase

### LINKED ISSUE #593: Reduce timer granularity constant
from https://github.com/mozilla/neqo/pull/591#issuecomment-622251047

> We should also look to reducing our timer granularity constant. It's currently 20ms, which is far too high.

martinthomson:
This is done automatically, with a little hysteresis to ensure that the
updates aren't performed too often.  Each connection updates its timer
resolution independently based on the timer values it reports
(currently, this is a quarter of the delay we determine) and that is
propagated to the thread level, where the relevant API for setting
timers are called.

I did a fair bit of digging here to learn how to do this.  There is some
contention about the value of the different APIs here, but the general
sense I get is that some people would prefer that timer resolution not
be increased unnecessarily.  This tries to avoid that, but in doing so
it might hit the various APIs more often than is ideal.

To reduce the number of potential system calls, there is a tiny amount
of hysteresis here: the last 10 timer resolution values are kept and the
second-lowest is used.  That way, we will tend not immediately peg the
resolution at the lowest value unless we have two events that justify
it.

Based on our default configuration, a high resolution is probably going
to be driven by acknowledgment delays or pacing.  With the 1/4, that
means our timer resolution will drive to 6ms when acknowledging packets,
as we use the default 25ms.  If we are sending, then pacing will
probably push this a lot lower much more readily as all it takes is 6
packets over a 50ms RTT to get to a 1ms resolution (1/4 for the delay,
plus we send at twice the current rate when pacing).  That will back
right off when we have nothing to send.

Closes #593.

### DIFF
--- a/neqo-common/Cargo.toml
+++ b/neqo-common/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.4.21"
 authors = ["Bobby Holley <bobbyholley@gmail.com>"]
 edition = "2018"
 license = "MIT/Apache-2.0"
+build = "build.rs"
 
 [dependencies]
 log = {version = "0.4.0", default-features = false}
--- a/neqo-common/build.rs
+++ b/neqo-common/build.rs
@@ -1,0 +1,4 @@
+fn main() {
+    #[cfg(windows)]
+    println!("cargo:rustc-link-lib=winmm");
+}
--- a/neqo-common/src/hrtime.rs
+++ b/neqo-common/src/hrtime.rs
@@ -299,23 +299,29 @@ impl Time {
     #[allow(clippy::unused_self)] // Only on some platforms is it unused.
     fn start(&self) {
         #[cfg(target_os = "macos")]
-        if let Some(p) = self.active {
-            mac::set_realtime(p.scaled(self.scale));
-        } else {
-            mac::set_thread_policy(self.deflt.clone());
+        {
+            if let Some(p) = self.active {
+                mac::set_realtime(p.scaled(self.scale));
+            } else {
+                mac::set_thread_policy(self.deflt.clone());
+            }
         }
 
         #[cfg(windows)]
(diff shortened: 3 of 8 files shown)


ddragana:
This looks fine, but it will conflict with necko setting it(I think it was event bucket). One may overwrite the other.
I see 2 options for fixing this:
1) have an API that returns current desired time granularity, so that necko can set it when the event bucket resets it.  necko can have a list of desired timer granularity one from neqo and one from the event bucket code. Necko will not do anything if the event bucket is not used.
2) have an API that force setting the timer again. So that necko can call it when the event bucket is done controlling the timers (the event bucket always sets it to 1ms)

I prefer the first one.

I do not have data about how often the event buckets sets the timer granularity.

martinthomson:
My understanding is that this doesn't conflict with other settings.  At least as far as the windows `timerBeginPeriod()` API goes.  That is, the OS maintains a count for each value that has been set and decrements the counter when `timerEndPeriod()` is called with that value.  If the count reaches zero, it moves to the next timer resolution that has a count (that is, the same algorithm I have here for aggregating).

As the timerBeginPeriod API was OS-level, and not process- or thread-level (I think that it is now process-level based on the experiments I ran and the blogs I've been reading), the OS has to handle this elegantly or this would fail spectacularly in other ways.  The automated tests are evidence of this (if you run multiple tests concurrently with `--nocapture`, you can see that the resolution improves for all threads even if only one thread invokes the API).

So if necko is calling `timerBeginPeriod()` independently, we should be fine.  We'll get 1ms resolution then.

The Mac thing does conflict with other changes on the same thread, but that API is thread-confined.

ddragana:
Thank toy for the explanation. That make sense.

necko uses this only on windows so that is fine.

ddragana (approved):

PR #1098, opened 2021-02-26
add_packet_number may return an error if there is not enough place in a packet. This is not an fatal error

ddragana:


neqo-transport/src/connection/mod.rs
```diff
@@ -1987,11 +1987,17 @@ impl Connection {
...
                 tx,
                 self.loss_recovery.largest_acknowledged_pn(*space),
-            )?;
+            ) {
+                pn
+            } else {
+                // No space for a packet of this type.
+                encoder = builder.abort();
+                continue;
+            };
```
martinthomson, line 2000: I think that it would be better to remove the `Res<T>` return on add_packet_number.  Just add a note that any overflow will be caught later (it's line 2005 on the right, so it's just a few extra lines of code that execute).

martinthomson (commented):
Take a look at #1100.  I think that we should delete code rather than add it here.

neqo-transport/src/connection/mod.rs
ddragana: I wrote a PR the same as #1100, but decided that I do not like having a code that overflows a buffer(i.e. buffer limits) even if it catch later.

martinthomson: We already do that with the header, I believe.  I think that the right answer here is to add a check at the point we add the header instead.  That check can ensure that the header, plus the 4 bytes needed for the packet number, are free in the buffer.

ddragana: We should do that. Let's land this.

PR #1101, opened 2021-03-01
Add API to increase a stream receiver buffer.

ddragana:


ddragana:
Necko used a similar API in HTTP2 to start off with a small receiver buffer per stream and only increases it later when it finds that a request is important (e.g. images start with a small receiver buffer and only increase later or if they are in the viewport). This is not a perfect way to do prioritization(I do not remember of top of my head all the detail why we did this in this way for http2). I am planning to do a bit of investigation about the order we send requests, so let's add this API to have this option open as well. 
This API only increases receiver buffer, it does not decrease it.

martinthomson:
I assume that we can close this now that we have #1125.

ddragana:
Yes, this is not needed any more.

PR #1100, opened 2021-03-01
An alternative to #1098

martinthomson:
This one just deletes a bunch of code.

Closes #1098.
Closes #1097.

PR #1102, opened 2021-03-01
Deal with failure to fit a packet header better

### LINKED ISSUE #1097: Remove InternalError(19) it is not an error
https://github.com/mozilla/neqo/blob/879d82ddd987c323de65a2426e384308eecc2b26/neqo-transport/src/packet/mod.rs#L310

martinthomson:
This turns out to be fiddly.  What we really want to do is just abort if
you can't create the header.  But the packet builder takes ownership of
the underlying buffer, and those places where it doesn't have enough
space is where we want that buffer back again.  So, rather than use a
`Res` return and have the buffer lost, set the limit to 0 if there isn't
enough space.  This causes the `remaining()` function to return 0 also
(after tweaking it), so that the packet can't be built.

I've collapsed some of the internal errors here as well.  Now that we
have better confidence that these are not causing problems, we should be
able to rely on less precise targeting of those errors.  So this walks back
some of the changes in #1070.

Closes #1100.
Closes #1098.
Closes #1097.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1759,6 +1759,7 @@ impl Connection {
                 tx,
                 self.loss_recovery.largest_acknowledged_pn(*space),
             );
+            // The builder will set the limit to 0 if there isn't enough space for the header.
             if builder.remaining() < 2 {
                 encoder = builder.abort();
                 break;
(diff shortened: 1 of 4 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1987,31 +1988,28 @@ impl Connection {
...
                 grease_quic_bit,
-            )?;
+            );
             let pn = Self::add_packet_number(
                 &mut builder,
                 tx,
                 self.loss_recovery.largest_acknowledged_pn(*space),
             );
-            let payload_start = builder.len();
+            if builder.remaining() < 2 {
```
ddragana, line 1997: Please add a comment here as well that build_packet_header and add_packet_number will set the builder limit to  0 if there is not enough space for a packet header.

ddragana (commented):
This is good to land. Thank you.

PR #1109, opened 2021-03-04
Send a PING with ACK occasionally

### LINKED ISSUE #830: Send a PING with ACK once per PTO
If we don't do this, then there is a risk that we accumulate a large number of acknowledgments.  I've seen this in traces and it gets unwieldy.

To do this, we should track when our last ack-eliciting packet was sent.  Then, if we are sending a packet and more than a PTO has elapsed since last sending an ack-eliciting packet, we can add a PING.  This should be more than PTO so that we don't send PING along with ACK too often.

martinthomson:
I decided to make this marginally more complex, sending slightly earlier
if we have lots of outstanding packets.  The logic is now that we send a
PING after 1 PTO if there are either 200 packets outstanding or after 2
PTOs with at least 16 packets outstanding, otherwise no PING.

These numbers are arbitrary, except for the minimum of 1 PTO.  The goal
is to make the larger high enough to encourage good ACK sending from
even the fastest peers; and to make the smaller enough that bad ACK
sending from a very slow sender doesn't lead to us asking for ACK when
we don't really need it.  Either threshold could be much larger or
smaller as you like, though the smaller probably shouldn't be lower than
a reasonable ACK threshold.

Closes #830.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1896,12 +1896,12 @@ impl Connection {
         let mut tokens = Vec::new();
         let primary = path.borrow().is_primary();
 
-        let ack_token = if primary {
-            self.acks
-                .write_frame(space, now, builder, &mut self.stats.borrow_mut().frame_tx)?
-        } else {
-            None
-        };
+        if primary {
+            let stats = &mut self.stats.borrow_mut().frame_tx;
+            if let Some(t) = self.acks.write_frame(space, now, builder, stats)? {
+                tokens.push(t);
+            }
+        }
         let ack_end = builder.len();
 
         // Avoid sending probes until the handshake completes,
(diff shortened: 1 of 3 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1896,12 +1896,12 @@ impl Connection {
...
-        let ack_token = if primary {
-            self.acks
-                .write_frame(space, now, builder, &mut self.stats.borrow_mut().frame_tx)?
-        } else {
-            None
-        };
+        if primary {
+            let stats = &mut self.stats.borrow_mut().frame_tx;
+            if let Some(t) = self.acks.write_frame(space, now, builder, stats)? {
+                tokens.push(t);
```
martinthomson, line 1902: Oh, with the change to use `ack_end`, we don't need to hold the ACK token back any more.  I forgot to make that change with the last patch.  This is far less brittle now.

ddragana (approved):

PR #1112, opened 2021-03-04
Extract flow control for receiver.

ddragana:
Add an interface for setting the initial stream flow control limits. This was needed to make encoder_flow_controlled_blocked work. The test was exploiting the fact that the flow control updates for the connection limit were sent any time a blocked frame is received. In this test send transport parameter and actual flow control limi did not match.


ddragana:
Fixes #1099

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
 
+#[derive(Debug)]
+pub struct ReceiverFlowControl<T>
+where
+    T: Debug + Sized,
+{
+    /// The thing that we're counting for.
+    subject: T,
+    /// Max amount of outstanding data
+    max_wnd: u64,
```
martinthomson, line 175:
```suggestion
    /// The maximum amount of items that can be active (e.g., the size of the receive buffer).
    max_active: u64,
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+        if retired <= self.retired {
+            return;
+        }
+
+        self.retired = retired;
+        let maybe_new_max = self.retired + self.max_wnd;
+        if maybe_new_max > (self.max_wnd / 2) + self.max_data {
+            self.max_data = maybe_new_max;
+            self.frame_pending = true;
+        }
```
martinthomson, line 215: So this is the same code as in send_flowc_update(), with an extra variable.  Maybe you can add an internal function that takes a value (x) where `x = self.max_wnd / 2` or `x = 0` as appropriate.

However, I think that we can make this better by not tracking max_data in quite the same way here.  That is, rather than only update the maximum when we see a need to do so (we retire more than half the window or we receive a blocked frame) instead, we can update the value at the time that the frame is sent.

That is, this function will just work out if an update might be needed.  That is, if we are retiring/reading data, we would set frame_pending to true if retired is within buffer_size/2 of the current maximum; or if we receive a blocked frame, we update if retired is less than buffer_size from the current maximum.  Then, when we are asked to produce a frame we calculate the new maximum as retired + buffer_size and pass the value to max_data_sent(), which clears frame_pending and sets max_data to that value.

```suggestion
        if self.retired + self.max_active / 2 > self.max_data {
            self.frame_pending = true;
        }
```

The effect of this should be that multiple reads will result in sending the largest MAX_DATA we can.  And, if we lose the packet containing MAX_DATA, we will send an even bigger value in the next packet if there were more reads in the meantime.

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+    }
+
+    pub fn lost(&mut self, maximum_data: u64) {
+        if maximum_data == self.max_data {
+            self.frame_pending = true;
+        }
+    }
+
+    fn max_data_sent(&mut self) {
+        self.frame_pending = false;
```
martinthomson, line 243:
```suggestion
    fn max_data_sent(&mut self, new_max: u64) {
        self.frame_pending = false;
        self.max_data = new_max;
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+        let new_max = self.retired + self.max_wnd;
+        if new_max > self.max_data {
+            self.max_data = new_max;
+            self.frame_pending = true;
+        }
+    }
+
+    pub fn max_data_needed(&self) -> Option<u64> {
+        if self.frame_pending {
+            Some(self.max_data)
```
martinthomson, line 230:
```suggestion
            Some(self.retired + self.max_active)
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+            self.frame_pending = true;
+        }
+    }
+
+    /// This function is called when STREAM_DATA_BLOCKED frame is received.
+    /// The flow control willl try to send an update if possible.
+    pub fn send_flowc_update(&mut self) {
+        let new_max = self.retired + self.max_wnd;
+        if new_max > self.max_data {
+            self.max_data = new_max;
```
martinthomson, line 223:
```suggestion
        if self.retired + self.max_active > self.max_data {
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+
+impl<T> ReceiverFlowControl<T>
+where
+    T: Debug + Sized,
+{
+    /// Make a new instance with the initial value and subject.
+    pub fn new(subject: T, max_bytes: u64) -> Self {
+        Self {
+            subject,
+            max_wnd: max_bytes,
```
martinthomson, line 191:
```suggestion
            max_active: max_bytes,
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+    pub fn check_allowed(&self, new_end: u64) -> bool {
+        new_end <= self.max_data
+    }
+
+    /// Some data has been read, retired them and maybe send flow control
+    /// update.
+    pub fn retired(&mut self, retired: u64) {
+        if retired <= self.retired {
+            return;
+        }
```
martinthomson, line 208: 👍 I was going to suggest this be an assertion, but this is better.  This way, we can use this for stream credit, which might not always retire streams in order.

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+        &mut self,
+        builder: &mut PacketBuilder,
+        tokens: &mut Vec<RecoveryToken>,
+        stats: &mut FrameStats,
+    ) -> Res<()> {
+        if let Some(max_data) = self.max_data_needed() {
+            if write_varint_frame(builder, &[FRAME_TYPE_MAX_DATA, max_data])? {
+                stats.max_data += 1;
+                tokens.push(RecoveryToken::MaxData(max_data));
+                self.max_data_sent();
```
martinthomson, line 258:
```suggestion
                self.max_data_sent(max_data);
```

neqo-transport/src/fc.rs
```diff
@@ -161,9 +164,131 @@ impl SenderFlowControl<StreamId> {
...
+            if write_varint_frame(
+                builder,
+                &[FRAME_TYPE_MAX_STREAM_DATA, self.subject.as_u64(), max_data],
+            )? {
+                stats.max_stream_data += 1;
+                tokens.push(RecoveryToken::MaxStreamData {
+                    stream_id: self.subject,
+                    max_data,
+                });
+                self.max_data_sent();
```
martinthomson, line 282:
```suggestion
                self.max_data_sent(max_data);
```

neqo-transport/src/recv_stream.rs
```diff
@@ -557,12 +509,42 @@ impl RecvStream {
...
     }
+
+    /// Maybe write a `MAX_STREAM_DATA` frame.
+    pub fn write_frame(
+        &mut self,
+        builder: &mut PacketBuilder,
+        tokens: &mut Vec<RecoveryToken>,
+        stats: &mut FrameStats,
+    ) -> Res<()> {
+        // Send MAX_STREAM_DATA at normal priority always.
```
martinthomson, line 520:
```suggestion
```
No point in mentioning this.  (We could prioritize receive, but it doesn't seem like it would help that much.)

test-fixture/src/lib.rs
```diff
@@ -133,6 +133,25 @@ pub fn default_server() -> Connection {
     make_default_server(DEFAULT_ALPN)
 }
 
+/// Create a transport server with an initial `max_stream_data` value.
+#[must_use]
+pub fn default_server_max_stream_data(max_stream_data: u64) -> Connection {
```
martinthomson, line 138: I think that it might be better to have this accept ConnectionParameters instead.  That seems more broadly useful.

martinthomson (changes requested):
I think that I put suggestions in for all the pieces that I care about.  Though I think we should have a test or two that confirms that the changes work as intended (one with extra retire calls between the retire call that causes the frame to be sent and when the frame is finally sent, and another one with a retire call between the sending of the frame and a loss).

neqo-transport/src/fc.rs
ddragana: This is a better name.

neqo-transport/src/recv_stream.rs
ddragana: yes, I decided to leave it at normal priority.

neqo-qpack/src/encoder.rs
```diff
@@ -590,7 +590,11 @@ mod tests {
     fn connect_generic(huffman: bool, max_data: Option<u64>) -> TestEncoder {
         let mut conn = default_client();
         let mut peer_conn = if let Some(max) = max_data {
-            default_server_max_stream_data(max)
+            default_server_max_stream_data(
```
martinthomson, line 593: Can you change the name of this?  `configure_server()` or something like that might be better.  It's not entirely a default now.

neqo-transport/src/connection/mod.rs
```diff
@@ -1834,12 +1836,7 @@ impl Connection {
             return Ok(());
         }
 
-        for stream in self.recv_streams.values_mut() {
-            stream.write_frame(builder, tokens, stats)?;
-            if builder.remaining() < 2 {
-                return Ok(());
-            }
-        }
+        recv_streams_write_frames(&mut self.recv_streams, builder, tokens, stats)?;
```
martinthomson, line 1839: Why is this not `self.recv_streams.write_frames(...)` ?

neqo-transport/src/fc.rs
```diff
@@ -171,8 +171,8 @@ where
 {
     /// The thing that we're counting for.
     subject: T,
-    /// Max amount of outstanding data
-    max_wnd: u64,
+    /// The maximum amount of items that can be active (e.g., the size of the receive buffer).
+    max_acive: u64,
```
martinthomson, line 175: typo: `max_ac[t]ive`
```suggestion
    max_active: u64,
```

neqo-transport/src/recv_stream.rs
```diff
@@ -33,6 +33,21 @@ pub const RECV_BUFFER_SIZE: usize = RX_STREAM_DATA_WINDOW as usize;
 
 pub(crate) type RecvStreams = BTreeMap<StreamId, RecvStream>;
```
martinthomson, line 34:
```suggestion
pub(crate) struct RecvStreams(BTreeMap<StreamId, RecvStream>);
```

I know that this means a bunch more factoring, so I'd be OK with a separate change, but it would be a good change.

martinthomson (commented):
Fix the naming and this is good to land.  Think about a newtype for RecvStreams, either here or elsewhere.  I haven't looked at how much work it would be to do, but it seems like it is time to do that.

neqo-qpack/src/encoder.rs
ddragana: I should not write code after 11PM after a long day

neqo-transport/src/fc.rs
ddragana: it was late evening  :(

PR #1113, opened 2021-03-04
Move sending of stop_sending frames into recv_stream

ddragana:


neqo-transport/src/recv_stream.rs
```diff
@@ -282,6 +283,10 @@ enum RecvStreamState {
         recv_buf: RxStreamOrderer,
     },
     DataRead,
+    ResetSending {
```
martinthomson, line 286: StopSending or AbortReading instead?

neqo-transport/src/recv_stream.rs
```diff
@@ -517,12 +529,27 @@ impl RecvStream {
...
-        if let RecvStreamState::Recv { fc, .. } = &mut self.state {
-            fc.write_frames(builder, tokens, stats)
-        } else {
-            Ok(())
+        match &mut self.state {
+            // Maybe send MAX_STREAM_DATA
+            RecvStreamState::Recv { fc, .. } => fc.write_frames(builder, tokens, stats)?,
+            // Maybe send STOP_SENDING
+            RecvStreamState::ResetSending { pending, err } => {
+                if *pending
```
martinthomson, line 537: "pending" might not be the right word here.  It could mean that you sending the frame is pending, or it could mean that the reset is pending.  Here, you have it meaning that the frame is pending, but as it is attached to something called "ResetSending", it would be easy to misinterpret.  I would go with frame_needed or something like that.

martinthomson (commented):
Very good.  Think about the names you are using for the different pieces and then land this.

PR #1114, opened 2021-03-04
Create TransportParameters at one place

### LINKED ISSUE #1073: Initialize transport parameters with valid local values
Right now, we have an odd arrangement where transport parameters are created, then modified to have default values, then the connection is constructed, then the connection parameters are used to update the transport parameters a third time.

This could be made to be a lot cleaner, perhaps with a method on ConnectionParameters that produced a value for local transport parameters.

ddragana:
Fixes #1073

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -4,10 +4,11 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use crate::connection::{ConnectionIdManager, Role, LOCAL_ACTIVE_CID_LIMIT, LOCAL_IDLE_TIMEOUT};
 use crate::recv_stream::RECV_BUFFER_SIZE;
 use crate::stream_id::{StreamIndex, StreamType};
-use crate::tparams::PreferredAddress;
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -125,4 +128,62 @@ impl ConnectionParameters {
...
+            .set_integer(tparams::INITIAL_MAX_DATA, LOCAL_MAX_DATA);
+        tps.local.set_integer(
+            tparams::IDLE_TIMEOUT,
+            u64::try_from(LOCAL_IDLE_TIMEOUT.as_millis()).unwrap(),
+        );
+        tps.local.set_integer(
+            tparams::ACTIVE_CONNECTION_ID_LIMIT,
+            u64::try_from(LOCAL_ACTIVE_CID_LIMIT).unwrap(),
+        );
+        tps.local.set_empty(tparams::DISABLE_MIGRATION);
```
martinthomson, line 153: Should we wrap this with `if role == Role::Server {}` ?

martinthomson (dismissed):
Much better.

Issue #1118, opened 2021-03-10
I am hitting an assertion in hrtime.rs


ddragana:
https://github.com/mozilla/neqo/blob/d086a329fd4ac17c9f6d146ac72d3ac5e3f0ace9/neqo-common/src/hrtime.rs#L179

martinthomson:
Maybe we can just remove those assertions.  If this doesn't work, the consequence isn't that severe.  Is this reproducible?

ddragana:
yes, I am debugging it.

I found out that this may fail, e.g. if system is busy. But the failure is very reproducible on 2 neqo tests, so I am no sure if it is the problem with the system being busy.

PR #1120, opened 2021-03-11
Fix thread_time_constraint_policy:

ddragana:
- call to this function may fail if a system is busy and we do not want to assert on that
- computation and constraint parameter of thread_time_constraint_policy should be a fraction of time se for the period


ddragana:
Fixes #1118

ddragana:
From he debugging: the function call was always failing if `computation` was above certain number.

martinthomson:
Thanks for confirming.  As this is best-effort, this is a good approach.

PR #1121, opened 2021-03-12
Update 0-RTT flow control limits

### LINKED ISSUE #1119: initial_max_stream_data increases after 0-RTT
If the initial stream data limits increase in server transport parameters over the value that was assumed in 0-RTT, we need to pretend that we received MAX_STREAM_DATA for each stream that is open.  We don't currently do this and we could become unnecessarily blocked as a result.

martinthomson:
After accepting 0-RTT, the server might increase flow control limits on
streams.  We need to ensure that these limits are applied to open
streams.

As part of this, I had to change the API for setting initial flow
control limits as we only had controls for bidirectional streams in the
wrong direction for the test (a server could only limit the streams it
created).  The API is more complex as a result, but it now works.  Note
that the qpack encoder test is now more general as a result, though I
think that the bidirectional limits aren't really relevant to that test.

Closes #1119.

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -43,7 +43,7 @@ pub struct ConnectionParameters {
     max_stream_data_uni: u64,
     /// Initial limit on bidirectional streams that the peer creates.
     max_streams_bidi: StreamIndex,
-    /// Initial limit on bidirectional streams that this endpoint creates.
+    /// Initial limit on unidirectional streams that this endpoint creates.
     max_streams_uni: StreamIndex,
     preferred_address: PreferredAddressConfig,
 }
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -33,10 +33,17 @@ pub enum PreferredAddressConfig {
...
-    max_stream_data_bidi: u64,
+    /// Initial flow control limit for receiving data on bidirectional streams that the peer creates.
+    max_stream_data_bidi_remote: u64,
+    /// Initial flow control limit for receiving data on bidirectional streams that this endpoint creates.
+    max_stream_data_bidi_local: u64,
+    /// Initial flow control limit for receiving data on unidirectional streams that the peer creates.
     max_stream_data_uni: u64,
+    /// Initial limit on bidirectional streams that the peer creates.
     max_streams_bidi: StreamIndex,
+    /// Initial limit on bidirectional streams that this endpoint creates.
```
ddragana, line 46: this should be "unidirectional"

ddragana (commented):
Fix comment and land.

PR #1125, opened 2021-03-12
Add API to increase a stream receiver buffer

ddragana:


ddragana:
I want to add more tests

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -549,3 +551,75 @@ fn no_dupdata_readable_events_empty_last_frame() {
...
+    let msd_before = server.stats().frame_rx.max_stream_data;
+    let out2 = client.process(None, now());
+    let out3 = server.process(out2.dgram(), now());
+    let expected = if RECV_BUFFER_START < new_fc { 1 } else { 0 };
+    assert_eq!(
+        server.stats().frame_rx.max_stream_data,
+        msd_before + expected
+    );
+
+    // If the flow control window has been increase, server can write more data.
```
martinthomson, line 588:
```suggestion
    // If the flow control window has been increased, server can write more data.
```

neqo-transport/src/fc.rs
```diff
@@ -239,6 +239,11 @@ where
         self.max_data = new_max;
         self.frame_pending = false;
     }
+
+    pub fn change_max_stream_data(&mut self, max_data: u64) {
```
martinthomson, line 243: This needs to be renamed to something about active, like set_active(max).  Then we can use it for other things.

neqo-transport/src/fc.rs
```diff
@@ -239,6 +239,11 @@ where
...
+        self.frame_pending |= self.max_active < max_data;
```
martinthomson, line 244: It seems to me like this is going to send MAX_STREAM_DATA sooner than we would otherwise.  I guess that's good, but you should document that intent.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -549,3 +551,75 @@ fn no_dupdata_readable_events_empty_last_frame() {
...
+    } else {
+        assert_eq!(written, 0);
+    }
+
+    // Exchange packets so that client gets all data.
+    let out4 = client.process(out3.dgram(), now());
+    let out5 = server.process(out4.dgram(), now());
+    let out6 = client.process(out5.dgram(), now());
+    let out7 = server.process(out6.dgram(), now());
+    let _ = client.process(out7.dgram(), now());
```
martinthomson, line 601: You don't need to do this if you set very small limits.

martinthomson (commented):
Looks good.

PR #1127, opened 2021-03-15
Refactor the flow control for streams

ddragana:


ddragana:
I will move the stream flow controls into RecvStreams and SendStreams in a follow up PR.

neqo-http3/src/connection_client.rs
```diff
@@ -692,6 +692,11 @@ impl Http3Client {
         Ok(())
     }
 
+    pub fn change_max_stream_data(&mut self, stream_id: u64, max_data: u64) -> Res<()> {
+        self.conn.change_max_stream_data(stream_id, max_data)?;
+        Ok(())
+    }
```
martinthomson, line 698: Is this changing the value of MAX_DATA or the size of the receive window?  I think that it's the latter, in which case this should be `c.set_receive_window(stream_id, window)`

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -549,3 +551,75 @@ fn no_dupdata_readable_events_empty_last_frame() {
     let _ = server.process(out_second_data_frame.dgram(), now());
     assert!(!server.events().any(stream_readable));
 }
+
+fn change_flow_control(stream_type: StreamType, new_fc: u64) {
+    const RECV_BUFFER_START: u64 = 3000;
```
martinthomson, line 556: This is a little easier if you start with a value that is smaller than a single packet.  That way you don't have to worry about exchanging more than one packet when you write something.

neqo-transport/src/fc.rs
```diff
@@ -239,6 +239,11 @@ where
         self.max_data = new_max;
         self.frame_pending = false;
     }
+
+    pub fn change_max_stream_data(&mut self, max_data: u64) {
```
martinthomson, line 274: set_active(active) is probably the right terminology to use.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -549,3 +551,75 @@ fn no_dupdata_readable_events_empty_last_frame() {
...
+    assert_eq!(u64::try_from(written).unwrap(), RECV_BUFFER_START);
+
+    // Send the stream to the client.
+    let out = server.process(None, now());
+    let _ = client.process(out.dgram(), now());
+
+    // change max_stream_data for stream_id.
+    client.change_max_stream_data(stream_id, new_fc).unwrap();
+
+    // server should receive a MAX_SREAM_DATA frame if the flow control window is updated.
```
martinthomson, line 578: MAX_S[T}TREAM_DATA and s/updated/increased

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -549,3 +551,75 @@ fn no_dupdata_readable_events_empty_last_frame() {
...
+fn increase_decrease_flow_control() {
+    const RECV_BUFFER_NEW_BIGGER: u64 = 4000;
+    const RECV_BUFFER_NEW_SMALLER: u64 = 2000;
+
+    change_flow_control(StreamType::UniDi, RECV_BUFFER_NEW_BIGGER);
+    change_flow_control(StreamType::BiDi, RECV_BUFFER_NEW_BIGGER);
+
+    change_flow_control(StreamType::UniDi, RECV_BUFFER_NEW_SMALLER);
+    change_flow_control(StreamType::BiDi, RECV_BUFFER_NEW_SMALLER);
+}
```
martinthomson, line 625: Four separate tests seem like they might be easier to run.  Or just two: one for increase, one for decrease.  There is no reason to run everything at the same time.

martinthomson (commented):
Check the title of this PR.  This is less about refactoring and more about setting the size of the receive buffer.

Other than the naming things, this is a good change.  Tweak and land it.

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+
+    pub fn is_new_stream(&mut self, stream_id: StreamId) -> Res<bool> {
+        let stream_idx = stream_id.as_u64() >> 2;
+        if !self.streams_fc.check_allowed(stream_idx) {
+            return Err(Error::StreamLimitError);
+        }
+        Ok(stream_idx >= self.next_stream)
+    }
+
+    pub fn add_stream(&mut self) -> Option<StreamId> {
```
martinthomson, line 150: This might be called `next_stream_id()`.  Why can't this just return `StreamId`?

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+
+    pub fn send_flowc_update(&mut self) {
+        self.streams_fc.send_flowc_update();
+    }
+
+    pub fn max_streams_lost(&mut self, max_streams: u64) {
+        self.streams_fc.lost(max_streams);
+    }
+
+    pub fn add_retired(&mut self, removed: u64) {
```
martinthomson, line 165: "removed" isn't clear here that it is a count.  Maybe "removed_count" instead.

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+        Self {
+            streams_fc: ReceiverFlowControl::new(stream_type, max_streams),
+            next_stream: 0,
+            stream_type_bits: type_val + role_val,
+        }
     }
 
-    pub fn as_u64(self) -> u64 {
-        self.0
+    pub fn write_frames(
```
martinthomson, line 133: There are lots of functions that forward to `self.fc`.  Maybe you can implement Deref and avoid the extra boilerplate.

neqo-transport/src/fc.rs
```diff
@@ -213,6 +240,10 @@ where
         }
     }
 
+    pub fn add_retired(&mut self, retired: u64) {
```
martinthomson, line 243: As I've mentioned, you might be able to replace this with an accessor for self.retired instead (though if you move the code as I suggest, you can just access the variable directly).

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+    pub fn add_retired(&mut self, removed: u64) {
+        self.streams_fc.add_retired(removed);
+        if removed > 0 {
+            // Send a update immediately.
+            self.streams_fc.send_flowc_update();
+        }
+    }
+}
+
+pub struct RemoteStreamsFlowControls([RemoteStreamsFlowControl; 2]);
```
martinthomson, line 174: I think that it might be better to have a StreamLimits object that has both local and remote limits all together.  That should take more of the load from Connection, which has a lot of work to do.

I was considering whether RecvStreams and SendStreams could handle this, but you are right to not put anything there; these limits don't work there as they deal with stream components, not the entire stream.

A wrapper means that you can't use Index[Mut], but I think that is OK.  `c.stream_limits.local_mut(UniDi)` and `c.stream_limits.remote(BiDi)` are pretty clear still.

Also, you don't need to use a fixed-size array for this.  It's OK to use named variables.

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+                SenderFlowControl::new(StreamType::UniDi, 0),
+            ],
+            role_bit: match role {
+                Role::Server => 1,
+                Role::Client => 0,
+            },
+        }
+    }
+
+    pub fn add_stream(&mut self, stream_type: StreamType) -> Option<StreamId> {
```
martinthomson, line 224: take_next_id() ?

neqo-transport/tests/server.rs
```diff
@@ -1109,8 +1108,8 @@ fn max_streams_after_0rtt_rejection() {
         Box::new(RejectZeroRtt {}),
         Rc::new(RefCell::new(CountingConnectionIdGenerator::default())),
         ConnectionParameters::default()
-            .max_streams(StreamType::BiDi, StreamIndex::new(MAX_STREAMS))
-            .max_streams(StreamType::UniDi, StreamIndex::new(MAX_STREAMS)),
+            .max_streams(StreamType::BiDi, MAX_STREAMS)
+            .max_streams(StreamType::UniDi, MAX_STREAMS),
```
martinthomson, line 1112: One thing to think about with these is the possibility that values can get switched.  It might be worth using different values for unidirectional and bidirectional stream limits in these tests.

neqo-transport/src/stream_id.rs
```diff
@@ -127,52 +105,172 @@ impl ::std::fmt::Display for StreamId {
...
+    pub fn send_flowc_update(&mut self) {
+        self.streams_fc.send_flowc_update();
+    }
+
+    pub fn max_streams_lost(&mut self, max_streams: u64) {
+        self.streams_fc.lost(max_streams);
+    }
+
+    pub fn add_retired(&mut self, removed: u64) {
+        self.streams_fc.add_retired(removed);
```
martinthomson, line 166: Maybe you don't need to define add_retired in the base FC class, but add an accessor instead and this function can do the addition.

Note that this line can go inside the if statement too.

neqo-transport/src/recv_stream.rs
```diff
@@ -1121,17 +1099,11 @@ mod tests {
 
     #[test]
     fn no_stream_flowc_event_after_exiting_recv() {
-        let flow_mgr = Rc::new(RefCell::new(FlowMgr::default()));
```
martinthomson, line 1118: Much nicer without this.

neqo-transport/src/connection/mod.rs
```diff
@@ -2932,101 +2898,68 @@ impl Connection {
...
-                    .to_stream_id(StreamType::UniDi, self.role);
-                self.indexes.remote_next_stream_uni += 1;
-                let initial_max_stream_data = self
-                    .tps
-                    .borrow()
-                    .remote()
-                    .get_integer(tparams::INITIAL_MAX_STREAM_DATA_UNI);
+        match self.local_streams_fc.add_stream(st) {
+            None => {
+                self.local_streams_fc[st].blocked();
```
martinthomson, line 2903: It seems to me like the blocked() call could be made from the take_next_id()/add_stream() call.

neqo-transport/src/connection/mod.rs
```diff
@@ -2932,101 +2898,68 @@ impl Connection {
...
+                Err(Error::StreamLimitError)
+            }
+            Some(new_id) => Ok(match st {
+                StreamType::UniDi => {
+                    let initial_max_stream_data = self
```
martinthomson, line 2908: This could use some factoring.

```rust
Some(new_id) => {
let send_limit_tp = match st {
  StreamType::UniDi => tparams::INITIAL_MAX_STREAM_DATA_UNI,
  StreamType::BiDi => tparams::INITIAL_MAX_STREAM_DATA_BIDI_REMOTE,
};
let send_limit = self.tps.borrow().remote().get_integer(send_limit_tp);
self.send_streams.insert(...);
if st == StreamType::BiDi {
   // ...
}
Ok(new_id.as_u64())
}
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2846,32 +2826,23 @@ impl Connection {
...
                         .local
                         .get_integer(tparams::INITIAL_MAX_STREAM_DATA_UNI)
                 };
 
-                loop {
-                    let next_stream_id =
-                        next_stream_idx.to_stream_id(stream_id.stream_type(), stream_id.role());
+                while self.remote_streams_fc[stream_id.stream_type()].is_new_stream(stream_id)? {
+                    let next_stream_id = self.remote_streams_fc[stream_id.stream_type()]
+                        .add_stream()
```
martinthomson, line 2837: I think that you can collapse these two calls into one, like:

```rust
while self.stream_limits.remote_mut(stream_id.stream_type()).add_stream(stream_id)? {
  // ...
}
```

It's probably not so terrible here to get the transport parameter value more than once.  Opening multiple streams is rare enough that the extra calls won't hurt.

neqo-transport/src/connection/mod.rs
```diff
@@ -2471,14 +2476,7 @@ impl Connection {
...
                 self.stats.borrow_mut().frame_rx.streams_blocked += 1;
-                let local_max = match stream_type {
-                    StreamType::BiDi => &mut self.indexes.local_max_stream_bidi,
-                    StreamType::UniDi => &mut self.indexes.local_max_stream_uni,
-                };
-
-                self.flow_mgr
-                    .borrow_mut()
-                    .max_streams(*local_max, stream_type)
+                self.remote_streams_fc[stream_type].send_flowc_update();
```
martinthomson, line 2479: I don't think that we need to include this call here.  We probably never needed the original code either.

Every time we retire a stream we already send an update.  So this isn't going to do anything.  Maybe replace this with a comment say that much instead.

neqo-transport/src/connection/mod.rs
```diff
@@ -1795,9 +1795,20 @@ impl Connection {
...
+        }
+
+        self.local_streams_fc[StreamType::BiDi].write_frames(builder, tokens, stats)?;
+        if builder.remaining() < 2 {
+            return Ok(());
+        }
+        self.local_streams_fc[StreamType::UniDi].write_frames(builder, tokens, stats)?;
         if builder.remaining() < 2 {
             return Ok(());
         }
```
martinthomson, line 1814: If you do as I suggest, then this block of code can be moved to the new StreamLimits object.

neqo-transport/src/connection/mod.rs
```diff
@@ -2122,10 +2133,10 @@ impl Connection {
...
         let tps = self.tps.borrow();
         let remote = tps.remote();
-        self.indexes.remote_max_stream_bidi =
-            StreamIndex::new(remote.get_integer(tparams::INITIAL_MAX_STREAMS_BIDI));
-        self.indexes.remote_max_stream_uni =
-            StreamIndex::new(remote.get_integer(tparams::INITIAL_MAX_STREAMS_UNI));
+        let _ = self.local_streams_fc[StreamType::BiDi]
+            .update(remote.get_integer(tparams::INITIAL_MAX_STREAMS_BIDI));
+        let _ = self.local_streams_fc[StreamType::UniDi]
+            .update(remote.get_integer(tparams::INITIAL_MAX_STREAMS_UNI));
```
martinthomson, line 2139: I couldn't work this out for a while, but then I realized that the remote streams object was returning the wrapper type, but the local streams one returns the underlying type.  This might be confusing.  At least until you get the Deref thing sorted out properly, I guess.

martinthomson (commented):
This is a good start, but I have a few suggestions about how this might be structured a little differently.

neqo-http3/src/connection_client.rs
ddragana: > Check the title of this PR. This is less about refactoring and more about setting the size of the receive buffer.
> 

I forgot to note that this is based on #1125 so the first change set will land in that PR and this one is only about streams flow control.

neqo-transport/src/stream_id.rs
ddragana: First I implemented StreamLimits (I think it really had the same name), but I did not like it. Somehow it felt like a half a way and had too complex interface. I think I wanted to move stream creation into that function. 
I need to think about it, maybe we need Streams object that holds all streams, the limits and it is responsible for creating streams and remove them. To make the usage easier recv_streams and send_streams can be public.

ddragana:
This change creates streams.rs that handles everything related to streams. I find this version better. Let me know what you think.

Only think I am not sure, but I think I should move RemoteStreamLimit and LocalStreamsFlowControls into streams.rs

neqo-transport/src/stream_id.rs
```diff
@@ -107,6 +87,17 @@ impl StreamId {
     pub fn is_recv_only(self, my_role: Role) -> bool {
         self.is_uni() && self.is_remote_initiated(my_role)
     }
+
+    pub fn next(&mut self) {
+        self.0 += 4;
+    }
+
+    pub fn role_bit(role: Role) -> u64 {
```
martinthomson, line 95: Please comment that this is the bit shared by all streams created by endpoints in the identified role.

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+// Stream ID and stream index handling.
```
martinthomson, line 7:
```suggestion
// Stream management for a connection.
```

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+use neqo_common::{qtrace, qwarn, Role};
+use std::cell::RefCell;
+use std::rc::Rc;
+
+pub struct Streams {
+    role: Role,
+    tps: Rc<RefCell<TransportParametersHandler>>,
+    events: ConnectionEvents,
+    flow_control_sender: Rc<RefCell<SenderFlowControl<()>>>,
+    flow_control_receiver: Rc<RefCell<ReceiverFlowControl<()>>>,
```
martinthomson, line 31:
```suggestion
    sender_fc: Rc<RefCell<SenderFlowControl<()>>>,
    receiver_fc: Rc<RefCell<ReceiverFlowControl<()>>>,
```

I think that we can tolerate shorter names in this context.  And it's consistent with the other ones you added.

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+    remote_streams_fc: RemoteStreamLimits,
+    local_streams_fc: LocalStreamsFlowControls,
+    pub(crate) send_streams: SendStreams,
+    pub(crate) recv_streams: RecvStreams,
```
martinthomson, line 35: This I'm less sure of, but it seems like it avoids some redundancy.

```suggestion
    pub(crate) send: SendStreams,
    pub(crate) recv: RecvStreams,
```

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+            flow_control_sender: Rc::new(RefCell::new(SenderFlowControl::new((), 0))),
+            flow_control_receiver: Rc::new(RefCell::new(ReceiverFlowControl::new((), max_data))),
+            remote_streams_fc: RemoteStreamLimits::new(limit_bidi, limit_uni, role),
+            local_streams_fc: LocalStreamsFlowControls::new(role),
+            send_streams: SendStreams::default(),
+            recv_streams: RecvStreams::default(),
+        }
+    }
+
+    pub fn client_0rtt_rejected(&mut self) {
```
martinthomson, line 66:
```suggestion
    pub fn zero_rtt_rejected(&mut self) {
        assert_eq!(self.role, Role::Client);
```

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+        self.send_streams.clear();
+        self.recv_streams.clear();
+        self.remote_streams_fc = RemoteStreamLimits::new(
+            self.tps
+                .borrow()
+                .local
+                .get_integer(tparams::INITIAL_MAX_STREAMS_BIDI),
```
martinthomson, line 73: Do we really need to update the remote stream limits here?  We won't have read anything from the server that would have caused us to change the limits.  Would a (debug) assertion that they haven't change suffice?

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    pub fn input_frame(&mut self, frame: Frame, stats: &mut FrameStats) -> Res<()> {
+        match frame {
+            Frame::ResetStream {
+                stream_id,
+                application_error_code,
+                ..
+            } => {
+                // TODO(agrover@mozilla.com): use final_size for connection MaxData calc
```
martinthomson, line 90: I think that we can drop this comment now.

```suggestion
```

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+
+            self.local_streams_fc[StreamType::BiDi].write_frames(builder, tokens, stats)?;
+            if builder.remaining() < 2 {
+                return Ok(());
+            }
+
+            self.local_streams_fc[StreamType::UniDi].write_frames(builder, tokens, stats)?;
+            if builder.remaining() < 2 {
+                return Ok(());
+            }
```
martinthomson, line 216: I would factor this out again into its own function.  Maybe `write_maintenance_frames`.

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+            RecoveryToken::Stream(st) => self.send_streams.lost(&st),
+            RecoveryToken::ResetStream { stream_id } => self.send_streams.reset_lost(*stream_id),
+            RecoveryToken::StreamDataBlocked { stream_id, limit } => {
+                self.send_streams.blocked_lost(*stream_id, *limit)
+            }
+            RecoveryToken::MaxStreamData {
+                stream_id,
+                max_data,
+            } => {
+                if let Ok((_, Some(rs))) = self.obtain_stream(*stream_id) {
```
martinthomson, line 234: Do we need to use obtain_stream() in this context?  Can't we use `self.recv.get_mut(id)` instead?  We're really only safeguarding against the stream being already dropped here, we don't need to call all the logic involving `self.send`.

neqo-transport/src/streams.rs
```diff
@@ -0,0 +1,464 @@
...
+    pub fn acked(&mut self, token: &RecoveryToken) {
+        match token {
+            RecoveryToken::Stream(st) => self.send_streams.acked(st),
+            RecoveryToken::ResetStream { stream_id } => self.send_streams.reset_acked(*stream_id),
+            RecoveryToken::StopSending { stream_id } => {
+                if let Ok((_, Some(rs))) = self.obtain_stream(*stream_id) {
+                    rs.stop_sending_acked();
+                }
+            }
+            _ => unreachable!("This is not a stream RecoveryToken"),
```
martinthomson, line 269: Should this be more tolerant of tokens that *are* for streams?  Like DataBlocked and Maxdata?

One way to avoid special matching logic is to have a method on RecoveryToken called `is_stream()` that the connection uses to route to here.  That saves us from having to replicate the matching logic in connection for stream-related tokens.

neqo-transport/src/connection/mod.rs
```diff
@@ -2591,39 +2431,21 @@ impl Connection {
...
+                    RecoveryToken::Stream(_)
+                    | RecoveryToken::ResetStream { .. }
+                    | RecoveryToken::StreamDataBlocked { .. }
+                    | RecoveryToken::MaxStreamData { .. }
+                    | RecoveryToken::StopSending { .. }
+                    | RecoveryToken::StreamsBlocked { .. }
+                    | RecoveryToken::MaxStreams { .. }
+                    | RecoveryToken::DataBlocked(_)
+                    | RecoveryToken::MaxData(_) => {
+                        self.streams.lost(token);
```
martinthomson, line 2448: I mention this later, but maybe RecoveryToken::stream_related() is worth defining so that this match block doesn't appear in multiple places.

martinthomson (commented):
I really like this structure.  The change is good and the way that this fits together is much better.  This neatly lifts out a lot of the bulk of the connection module into a coherent unit.

I'm going to suggest that we do a follow-up to make streams a module in its own right, including stream_id.rs, send_stream.rs, and recv_stream.rs.  But let's take this one step at a time.

Lots of comments here, but those are mostly just suggestions for improvements.  I would be happy to review again if you feel like making larger changes than my suggestions, but you can land this once it is fixed up.

neqo-transport/src/fc.rs
```diff
@@ -466,4 +668,209 @@ mod test {
         fc.retired(136);
         assert_eq!(fc.max_data_needed(), Some(196));
     }
+
+    #[test]
+    fn local_streams_flow_control_new_stream_client() {
```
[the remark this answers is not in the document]
ddragana: In the original version (before my refactoring) local and remote were switch, I do not know why. I wrote these tests before I switched the names.

neqo-transport/src/streams.rs
ddragana: I will do this in a follow up. This PR is trying not to change behavior, just to move things around.

neqo-transport/src/streams.rs
ddragana: I was thinking about adding something like is_stream() to RecoveryToken, but decided to leave it for a followup.

neqo-transport/src/fc.rs
```diff
@@ -347,11 +338,20 @@ impl ReceiverFlowControl<StreamType> {
...
                 });
-                self.max_data_sent(max_streams);
+                self.frame_sent(max_streams);
             }
         }
         Ok(())
     }
+
+    /// Retire given amount of additional data.
+    /// This function will sen flow updates immediately.
```
martinthomson, line 348:
```suggestion
    /// This function will send flow updates immediately.
```

neqo-transport/src/streams.rs
```diff
@@ -27,12 +25,12 @@ pub struct Streams {
...
     events: ConnectionEvents,
-    flow_control_sender: Rc<RefCell<SenderFlowControl<()>>>,
-    flow_control_receiver: Rc<RefCell<ReceiverFlowControl<()>>>,
+    sender_fc: Rc<RefCell<SenderFlowControl<()>>>,
+    receiver_fc: Rc<RefCell<ReceiverFlowControl<()>>>,
     remote_streams_fc: RemoteStreamLimits,
-    local_streams_fc: LocalStreamsFlowControls,
-    pub(crate) send_streams: SendStreams,
-    pub(crate) recv_streams: RecvStreams,
+    local_streams_fc: LocalStreamLimits,
```
martinthomson: Should remote_streams_fc be remote_stream_limits instead?

martinthomson (approved):

martinthomson:
Bah, I think that we should just disable mergify.

PR #1131, opened 2021-03-17
Enable ECH

martinthomson:
There is a lot of plumbing here, but it's all pretty simple stuff.

Testing only checks that the integration works properly.  I've run manual tests with the client and server with ECH on and that works fine.

I pulled in a bunch of code from my work on HPKE/Oblivious HTTP (as there are commonalities there).  That means some improvements to neqo-crypto, in particular, the types we use for wrapping NSS objects have been enhanced a little.  Some of that is to support the use of the ECH API (PublicKey and PrivateKey in particular needed a bunch of new functions), but I also simplified the creation of the scoped pointer wrappers and SECItem got a new wrapper too.

The only thing I didn't do was handle an automatic retry for the test client when the ECH retry is used.  It is tested to ensure that it works properly, but I didn't do a retry in the client.  We don't currently retry things in the client, so building that capability is probably best to defer.

Edit: NSS changes have landed.


neqo-crypto/src/p11.rs
```diff
@@ -31,9 +32,16 @@ macro_rules! scoped_ptr {
...
 
         impl $scoped {
-            #[must_use]
-            pub fn new(ptr: NonNull<$target>) -> Self {
-                Self { ptr: ptr.as_ptr() }
+            /// Create a new instance of `$scoped` from a pointer.
+            ///
+            /// # Errors
+            /// When passed a null pointer generates an error.
+            pub fn from_ptr(ptr: *mut $target) -> Result<Self, crate::err::Error> {
```
ddragana, line 39: this is a good change.

ddragana (commented):
This generally looks good. Please take a look a my comments.

ddragana:
Thanks for the test, they are better now.
This looks good and it can land when nss is ready.

PR #1133, opened 2021-03-18
Use winapi for high resolution timer when using gecko

KershawChang:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1698438#c6.
The linker failed to link `timeBeginPeriod` and `timeEndPeriod` in win32 platforms. Switching to use `winapi` seems to fix this issue.


ddragana (approved):
gecko is already using it let's just use it.

PR #1137, opened 2021-03-25
Add feature that disables encryption

### LINKED ISSUE #854: For Fuzzing we need to disable encryption

ddragana:
This is needed for fuzzing.
This also disables tests:
- aead tests
- selfencrypt tests
- tests that use corrupted initial, because they relay on a fact that initial cannot be decrypted
- tests that compare packets to precoded samples
- tests that relay on retry token validation, because that does not work anymore

Fixes #854


neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+        _aad: &[u8],
+        input: &[u8],
+        output: &'a mut [u8],
+    ) -> Res<&'a [u8]> {
+        let l = input.len();
+        // Check that:
+        // 1) expansion is all zeros and
+        // 2) if the encrypted data is also supplied that at least some values
+        //    are no zero (otherwise padding will be interpreted as a valid packet)
+        if input[l - 16..l].iter().all(|x| *x == 0x0) && (l == 16  || input[..l - 16].iter().any(|x| *x != 0x0)) {
```
martinthomson, line 51:
```suggestion
        if tag.len() >= FIXED_TAG.len() && &input[input.len() - FIXED_TAG.len()..] == FIXED_TAG {
```

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+            output[..l].copy_from_slice(input);
```
martinthomson, line 52:
```suggestion
            output[..input.len() - FIXED_TAG.len()].copy_from_slice(input[..input.len() - FIXED_TAG.len()]);
```
...starting to think you want a temporary variable for `input.len() - FIXED_TAG.len()` here.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+            Ok(&output[..l - 16])
+        } else {
+            Err(Error::NssError {
+                name: "SEC_ERROR_BAD_DATA".to_string(),
+                code: -8190,
+                desc: "security library: received bad data.".to_string(),
+            })
+        }
```
martinthomson, line 60: Please use `Error::from(SEC_ERROR_BAD_DATA)` rather than copying this stuff.  You might need to find the right module path to that variable.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+impl fmt::Debug for Aead {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "[AEAD Context]")
```
martinthomson, line 66: Put some nice text here like [FUZZING AEAD] or something.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::constants::{Cipher, Version};
+use crate::err::{Error, Res};
+use crate::p11::SymKey;
+use std::fmt;
+
+pub struct Aead {}
+
+impl Aead {
+    pub fn new(_version: Version, _cipher: Cipher, _secret: &SymKey, _prefix: &str) -> Res<Self> {
+        Ok(Self {})
```
martinthomson, line 16: Can you add something that causes this code to explode if it is run outside of a debug build?

```rust
#[cfg(not(debug_assertions))]
panic!("Trying to run a non-debug fuzzing build.");
```

neqo-crypto/tests/aead.rs
```diff
@@ -1,116 +1,120 @@
...
-    let expected_ciphertext: &[u8] = &[
-        0x5f, 0x01, 0xc4, 0xc2, 0xa2, 0x30, 0x3d, 0x29, 0x7e, 0x3c, 0x51, 0x9b, 0xf6, 0xb2, 0x23,
-        0x86, 0xe3, 0xd0, 0xbd, 0x6d, 0xfc, 0x66, 0x12, 0x16, 0x77, 0x29, 0x80, 0x31, 0x04, 0x1b,
-        0xb9, 0xa7, 0x9c, 0x9f, 0x0f, 0x9d, 0x4c, 0x58, 0x77, 0x27, 0x0a, 0x66, 0x0f, 0x5d, 0xa3,
-        0x62, 0x07, 0xd9, 0x8b, 0x73, 0x83, 0x9b, 0x2f, 0xdf, 0x2e, 0xf8, 0xe7, 0xdf, 0x5a, 0x51,
-        0xb1, 0x7b, 0x8c, 0x68, 0xd8, 0x64, 0xfd, 0x3e, 0x70, 0x8c, 0x6c, 0x1b, 0x71, 0xa9, 0x8a,
-        0x33, 0x18, 0x15, 0x59, 0x9e, 0xf5, 0x01, 0x4e, 0xa3, 0x8c, 0x44, 0xbd, 0xfd, 0x38, 0x7c,
-        0x03, 0xb5, 0x27, 0x5c, 0x35, 0xe0, 0x09, 0xb6, 0x23, 0x8f, 0x83, 0x14, 0x20, 0x04, 0x7c,
-        0x72, 0x71, 0x28, 0x1c, 0xcb, 0x54, 0xdf, 0x78, 0x84,
+#[cfg(not(feature = "fuzzing"))]
```
martinthomson, line 4: Use `#![cfg(not(feature = "fuzzing"))]` if the goal is to disable the entire module.

neqo-crypto/tests/selfencrypt.rs
```diff
@@ -1,91 +1,94 @@
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![warn(clippy::pedantic)]
 
-use neqo_crypto::constants::{TLS_AES_128_GCM_SHA256, TLS_VERSION_1_3};
-use neqo_crypto::{init, selfencrypt::SelfEncrypt, Error};
+#[cfg(not(feature = "fuzzing"))]
```
martinthomson, line 4: As above.

neqo-http3/Cargo.toml
```diff
@@ -20,3 +20,4 @@ test-fixture = { path = "../test-fixture" }
 [features]
 default = ["deny-warnings"]
 deny-warnings = []
+fuzzing_h = ["neqo-transport/fuzzing_t", "neqo-crypto/fuzzing"]
```
martinthomson, line 23: why the different name?

neqo-transport/src/packet/mod.rs
```diff
@@ -1225,7 +1240,14 @@ mod tests {
...
         // So only do a full check when that randomness matches up.
         if retry[0] == sample_retry[0] {
+            #[cfg(not(feature = "fuzzing_t"))]
             assert_eq!(&retry, &sample_retry);
+
+            #[cfg(feature = "fuzzing_t")]
+            assert_eq!(
+                &retry[..retry.len() - 16],
+                &sample_retry[..sample_retry.len() - 16]
+            );
```
martinthomson, line 1250: I'd say that you shouldn't bother with this check.  Just disable the test (or the above assert).

neqo-transport/src/packet/mod.rs
```diff
@@ -866,10 +866,16 @@ impl Deref for DecryptedPacket {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::crypto::{CryptoDxState, CryptoStates};
+    #[cfg(not(feature = "fuzzing_t"))]
```
martinthomson, line 869: I think that we can just disable all of these tests.  Far too much of this is not relevant to fuzzing (it's internals) anyway.  We'll get more value from fuzzing the end-to-end testing anyway.  Picking individual pieces out is too hard to maintain.

neqo-transport/tests/conn_vectors.rs
```diff
@@ -7,281 +7,299 @@
 // Tests with the test vectors from the spec.
 #![deny(clippy::pedantic)]
 
-use neqo_common::Datagram;
-use neqo_transport::{
-    Connection, ConnectionParameters, QuicVersion, RandomConnectionIdGenerator, State,
-};
-use test_fixture::{self, addr, now};
+#[cfg(not(feature = "fuzzing_t"))]
```
martinthomson, line 10: As before.

neqo-transport/tests/server.rs
```diff
@@ -461,6 +464,7 @@ fn new_token_0rtt() {
     assert!(client.tls_info().unwrap().resumed());
 }
 
+#[cfg(not(feature = "fuzzing_t"))]
```
martinthomson, line 467: Maybe you can move the Retry testing to a new file and just disable that instead.

martinthomson (changes requested):
Real fuzzers don't do crypto.

I think that this could be far less targeted when it comes to disabling tests.  The only place where being selective might be useful is Retry, where we do some very crypto-specific stuff.

neqo-http3/Cargo.toml
ddragana: I got confuse reading rust rules. The same name is allowed here. There was something else that does not allow same name.

martinthomson (commented):
I'm sorry, I had these comments written, but I failed to send them.  This is generally good, but I think that there are more changes needed.

The one thing I would like to see here is a simple test that checks that  we don't encrypt.  One way to do that is to make a connection, write a particular pattern to a stream and then check that the bytes you wrote are visible in the resulting packet.

ddragana:
I forgot to add test.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -0,0 +1,73 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use crate::constants::{Cipher, Version};
+use crate::err::{sec::SEC_ERROR_BAD_DATA, Error, Res};
+use crate::p11::SymKey;
+use std::fmt;
+
+const FIXED_TAG: &[u8] = &[0x0; 16];
```
martinthomson, line 12:
```suggestion
const FIXED_TAG: &[u8] = &[0x0a; 16];
```

Let's make it non-zero.

neqo-transport/src/connection/tests/fuzzing.rs
```diff
@@ -0,0 +1,49 @@
...
+#![cfg_attr(feature = "deny-warnings", deny(warnings))]
+#![warn(clippy::pedantic)]
+#![cfg(feature = "fuzzing")]
+
+use super::{connect_force_idle, default_client, default_server};
+use crate::StreamType;
+use test_fixture::now;
+
+#[test]
+fn packes_no_encrypted() {
```
martinthomson, line 16:
```suggestion
fn no_encryption() {
```

neqo-transport/src/connection/tests/fuzzing.rs
```diff
@@ -0,0 +1,49 @@
...
+    connect_force_idle(&mut client, &mut server);
+
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+
+    client.stream_send(stream_id, DATA_CLIENT).unwrap();
+    let client_pkt = client.process_output(now()).dgram().unwrap();
+    assert_eq!(
+        &client_pkt[HEADER_LEN..HEADER_LEN + DATA_CLIENT.len()],
+        DATA_CLIENT
+    );
```
martinthomson, line 31: it might be better to just search for the data.  That will be more robust.  I think that means iterating and using `starts_with()`.

neqo-transport/tests/server.rs
```diff
@@ -7,92 +7,34 @@
...
-    assert!(dgram.is_some()); // ACK + HANDSHAKE_DONE + NST
-
-    // Have the client process the HANDSHAKE_DONE.
-    let dgram = client.process(dgram, now()).dgram();
-    assert!(dgram.is_none());
-    assert_eq!(*client.state(), State::Confirmed);
-
-    connected_server(server)
-}
+mod common;
```
martinthomson, line 32: Maybe move this above the `use` statements.

martinthomson (commented):
Looks good.  A few small nits only.  Land when you are done.

neqo-transport/src/connection/tests/fuzzing.rs
```diff
@@ -0,0 +1,42 @@
...
+    const DATA_SERVER: &[u8] = &[3; 50];
+    let mut client = default_client();
+    let mut server = default_server();
+    connect_force_idle(&mut client, &mut server);
+
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+
+    client.stream_send(stream_id, DATA_CLIENT).unwrap();
+    let client_pkt = client.process_output(now()).dgram().unwrap();
+    assert!(client_pkt.iter().star_wih(DATA_CLIENT));
```
martinthomson, line 27: I'm guessing that you haven't run this.

neqo-transport/src/crypto.rs
```diff
@@ -1088,6 +1089,7 @@ impl CryptoStates {
         }
     }
 
+    #[cfg(not(feature = "fuzzing"))]
     #[cfg(test)]
```
martinthomson, line 1093: Maybe we should be more consistent about using two `#[cfg(...)]` statements vs. `#[cfg(all(...))]`.

PR #1141, opened 2021-04-07
Lots of clippy changes this time.

martinthomson:
The main offenders here are the insistence that we document panics.

There is also a rename of QPData and PNSpace to comply with updating
naming conventions.

And there was a warning about initializing vectors followed by a push.

And all Into implementations are inverted and replaced with From implementations.

And all `as` casting for pointers is changed to pointer::cast(), with turbofish as needed.

I had to suppress some warnings, which was annoying.  Particularly since
suppressing the unknown lint warning requires three warning suppressions
now: the new way of suppressing, the old way of suppressing, and a new
lint that warns about the old way of suppressing still being there.

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -35,7 +35,9 @@ impl<'a> Decoder<'a> {
         self.offset
     }
 
-    /// Skip n bytes.  Panics if `n` is too large.
+    /// Skip n bytes.
+    /// # Panics
+    /// If the remaining quantity is less than `n`.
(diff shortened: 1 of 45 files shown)


ddragana:
> I had to suppress some warnings, which was annoying. Particularly since
> suppressing the unknown lint warning requires three warning suppressions
> now: the new way of suppressing, the old way of suppressing, and a new
> lint that warns about the old way of suppressing still being there.

This is funny.

ddragana (approved):

PR #1143, opened 2021-04-08
Distinguish RecoveryTokens and Frames that er for streams.

ddragana:


neqo-transport/src/connection/mod.rs
```diff
@@ -2288,141 +2288,138 @@ impl Connection {
...
-                    let read = self.crypto.streams.read_to_end(space, &mut buf);
-                    qdebug!("Read {} bytes", read);
-                    self.handshake(now, space, Some(&buf))?;
-                    self.create_resumption_token(now);
-                } else {
-                    // If we get a useless CRYPTO frame send outstanding CRYPTO frames again.
+        if frame.is_stream() {
+            self.streams
+                .input_frame(frame, &mut self.stats.borrow_mut().frame_rx)?;
+        } else {
```
martinthomson, line 2294: Rather than use `else` here, it might be better to return early.

neqo-transport/src/connection/mod.rs
```diff
@@ -2491,23 +2484,21 @@ impl Connection {
...
+                } else {
+                    match token {
+                        RecoveryToken::Ack(at) => self.acks.acked(at),
+                        RecoveryToken::Crypto(ct) => self.crypto.acked(ct),
+                        RecoveryToken::NewToken(seqno) => self.new_token.acked(*seqno),
+                        RecoveryToken::NewConnectionId(entry) => self.cid_manager.acked(entry),
+                        RecoveryToken::RetireConnectionId(seqno) => {
+                            self.paths.acked_retire_cid(*seqno)
+                        }
+                        // We only worry when these are acked
```
martinthomson, line 2498:
```suggestion
                        // We only worry when these are lost
```

neqo-transport/src/frame.rs
```diff
@@ -242,6 +242,21 @@ impl<'a> Frame<'a> {
         }
     }
 
+    pub fn is_stream(&self) -> bool {
+        match self {
```
martinthomson, line 246: use `matches!()`

neqo-transport/src/recovery.rs
```diff
@@ -80,6 +80,23 @@ pub enum RecoveryToken {
     },
 }
 
+impl RecoveryToken {
+    pub fn is_stream(&self) -> bool {
+        match self {
```
martinthomson, line 85: `matches!()`

neqo-transport/src/connection/mod.rs
```diff
@@ -2288,141 +2288,138 @@ impl Connection {
...
+                Frame::HandshakeDone => {
+                    self.stats.borrow_mut().frame_rx.handshake_done += 1;
+                    if self.role == Role::Server || !self.state.connected() {
+                        return Err(Error::ProtocolViolation);
+                    }
+                    self.set_state(State::Confirmed);
+                    self.discard_keys(PacketNumberSpace::Handshake, now);
+                    self.migrate_to_preferred_address(now)?;
+                }
+                _ => unreachable!("The rest of frames are for streams"),
```
martinthomson, line 2420:
```suggestion
                _ => unreachable!("All other frames are for streams"),
```

martinthomson (commented):
Looks good.  Fix and land.

PR #1144, opened 2021-04-08
Small fixes

ddragana:
Rename max_data in ReceiverFlowControl to reflect other use-cases.
Remote limits do not need o be reinitialise after ZeroRtt has been rejected.


martinthomson (approved):
You'll need to deal with the variable renaming, but this is all good.

PR #1150, opened 2021-04-28
Expose HttpZeroRttChecker, to be able to test ZeroRtt being rejected

ddragana:


ddragana:
This is not the best API for a Http3Server, but I avoided to make something more complex just  to enable testing of ZeroRtt being rejected.

neqo-server/src/main.rs
```diff
@@ -466,6 +466,11 @@ impl ServersRunner {
                     max_table_size_decoder: args.max_table_size_decoder,
                     max_blocked_streams: args.max_blocked_streams,
                 },
+                Some(HttpZeroRttChecker::new(QpackSettings {
+                    max_table_size_encoder: args.max_table_size_encoder,
+                    max_table_size_decoder: args.max_table_size_decoder,
+                    max_blocked_streams: args.max_blocked_streams,
+                })),
```
martinthomson, line 473: How is this different than `None`?

martinthomson (commented):
Did you plan to add a test that used this new option?  I don't see that code.

neqo-server/src/main.rs
ddragana: It is not. I used that for testing and wanted to remove it, but forgot.

ddragana:
> Did you plan to add a test that used this new option? I don't see that code.

I need this option for Necko testing, but I will add tests here as well.

neqo-http3/src/server.rs
```diff
@@ -54,14 +54,19 @@ impl Http3Server {
...
+        zero_rtt_checker: Option<Box<dyn ZeroRttChecker>>,
     ) -> Res<Self> {
         Ok(Self {
             server: Server::new(
                 now,
                 certs,
                 protocols,
                 anti_replay,
-                Box::new(HttpZeroRttChecker::new(qpack_settings)),
+                if let Some(c) = zero_rtt_checker {
```
martinthomson, line 65: This should use unwrap_or_else

neqo-http3/src/server.rs
```diff
@@ -446,8 +457,8 @@ mod tests {
     }
 
     // Connect transport, send and receive settings.
-    fn connect() -> (Http3Server, PeerConnection) {
-        let (mut hconn, mut neqo_trans_conn) = connect_and_receive_settings();
+    fn create_client_and_connect_with_server(server: &mut Http3Server) -> PeerConnection {
```
martinthomson, line 460: A shorter name might be easier.

```suggestion
    fn connect_to(server: &mut Http3Server) -> PeerConnection {
```

martinthomson (commented):
Fix and land.  Thanks for updating the test.  Good to see that doing that had the right effect.

PR #1154, opened 2021-05-18
Implement NEW_CONNECTION_ID::retire_prior_to

martinthomson:
This is the implementation part.  I'm fairly sure that it's good, but I
haven't tested it.  And testing this promises to be tricky as there are
interactions with paths.  And we don't use Retire Prior To for our own
purposes.  It might be possible to do something invasive; I'll see what
I can come up with.


martinthomson:
This has some testing now.  It's not completely horrible even.

ddragana (approved):
This looks good. Thank you for testing corner cases.

PR #1158, opened 2021-05-28
Use RecvStream trait for control streams as well

ddragana:


ddragana:
I want to have another look before a review.

neqo-http3/src/connection_server.rs
```diff
@@ -216,24 +217,23 @@ impl Http3ServerHandler {
...
+        let recv_stream = self
+            .base_handler
+            .recv_streams
+            .get_mut(&stream_id)
+            .ok_or(Error::InvalidStreamId)?;
+
+        let hs = recv_stream.http_stream().ok_or(Error::InvalidStreamId)?;
+        match hs.read_data(conn, buf) {
+            Ok((amount, fin)) => {
+                if recv_stream.done() {
```
martinthomson, line 229: If you are able to use `hs.done()` here (I think that is possible), then you only need one variable.

neqo-http3/src/control_stream_remote.rs
```diff
@@ -24,45 +25,51 @@ impl ::std::fmt::Display for ControlStreamRemote {
...
-                self.fin = true;
-                Err(Error::HttpClosedCriticalStream)
+    fn receive(&mut self, conn: &mut Connection) -> Res<ReceiveOutput> {
+        let mut control_frames = Vec::new();
+
+        loop {
+            if let Some(f) = self.receive_single(conn)? {
+                control_frames.push(f);
+            } else if control_frames.is_empty() {
+                return Ok(ReceiveOutput::NoOutput);
```
martinthomson, line 57: Do you need to worry about this?  If there are no control frames, that's OK, right?  The caller will just do nothing.

neqo-http3/src/qpack_decoder_receiver.rs
```diff
@@ -0,0 +1,52 @@
...
+impl RecvStream for DecoderRecvStream {
+    fn stream_reset(&self, _error: AppError, _reset_type: ResetType) -> Res<()> {
+        Err(Error::HttpClosedCriticalStream)
+    }
+
+    fn receive(&mut self, conn: &mut Connection) -> Res<ReceiveOutput> {
+        let unblocked_streams = self.decoder.borrow_mut().receive(conn, self.stream_id)?;
+        if unblocked_streams.is_empty() {
+            Ok(ReceiveOutput::NoOutput)
+        } else {
```
martinthomson, line 36: As before, I don't think that it matters if this returns an empty vector rather than `NoOutput`.

neqo-http3/src/recv_message.rs
```diff
@@ -467,14 +453,42 @@ impl RecvStream for RecvMessage {
...
+                self.push_handler
+                    .as_ref()
+                    .ok_or(Error::HttpFrameUnexpected)?
+                    .borrow_mut()
+                    .new_push_promise(p.push_id, self.stream_id, headers)?;
+                self.blocked_push_promise.pop_front();
+            }
+        }
+
+        if self.blocked_push_promise.is_empty() {
```
martinthomson, line 485: This check is redundant.  The loop above guarantees it.

neqo-http3/src/recv_message.rs
```diff
@@ -467,14 +453,42 @@ impl RecvStream for RecvMessage {
...
+                .decode_header_block(&p.header_block, self.stream_id)?
+            {
+                self.push_handler
+                    .as_ref()
+                    .ok_or(Error::HttpFrameUnexpected)?
+                    .borrow_mut()
+                    .new_push_promise(p.push_id, self.stream_id, headers)?;
+                self.blocked_push_promise.pop_front();
+            }
+        }
```
martinthomson, line 483: I know that this is old code, but...

What happens if a header is unblocked but not all of the push promises can be unblocked?  It looks like this loops forever in that case.  If decode_header_block returns Ok(None) this is an infinite loop.

martinthomson (commented):
The change is good.  I have a few questions though.

neqo-http3/src/connection.rs
```diff
@@ -258,76 +233,67 @@ impl Http3Connection {
...
+                }
+                Ok(ReceiveOutput::NoOutput)
+            }
+            ReceiveOutput::ControlFrames(mut control_frames) => {
+                let mut rest = Vec::new();
+                for cf in control_frames.drain(..) {
+                    if let Some(not_handled) = self.handle_control_frame(cf)? {
+                        rest.push(not_handled);
+                    }
+                }
```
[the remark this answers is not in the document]
ddragana: The problem with retain is that handle_control_frame can return an error. I hav tried with filter_map, but to make that all work I would need to collect vectors of Res<..>. The code looks better in this way.

neqo-http3/src/connection.rs
```diff
@@ -258,76 +233,67 @@ impl Http3Connection {
...
 
     fn is_critical_stream(&self, stream_id: u64) -> bool {
         self.qpack_encoder
+            .borrow()
             .local_stream_id()
             .iter()
-            .chain(self.qpack_encoder.remote_stream_id().iter())
-            .chain(self.qpack_decoder.local_stream_id().iter())
-            .chain(self.qpack_decoder.remote_stream_id().iter())
+            .chain(self.qpack_encoder.borrow().local_stream_id().iter())
```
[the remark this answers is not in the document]
ddragana: It should be only once, I will fix it.
The remote streams implement RecvStream now and resetting them is handled in stream_reset that return error. Therefore they are not needed here anymore.

neqo-http3/src/connection.rs
```diff
@@ -424,73 +388,65 @@ impl Http3Connection {
...
-    fn decode_new_stream(
+    /// Returns true if the stream is a steam handled by this structture.
+    /// A push stream is the only stream that is not handled here.
+    fn handle_new_stream(
         &mut self,
         conn: &mut Connection,
         stream_type: u64,
         stream_id: u64,
     ) -> Res<bool> {
         match stream_type {
```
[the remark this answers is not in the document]
ddragana: This is a good suggestion.

neqo-http3/src/connection.rs
```diff
@@ -516,21 +472,30 @@ impl Http3Connection {
     ) -> Res<()> {
         qinfo!([self], "Reset stream {} error={}.", stream_id, error);
 
-        let mut found = self.send_streams.remove(&stream_id).is_some();
+        // Make sure that an application cannot reset a control stream.
+        // self.send_streams holds only http streams therefore if the stream
+        // is present in self.send_streams we don not need to check recv_streams.
+        if self.send_streams.get(&stream_id).is_none()
```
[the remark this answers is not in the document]
ddragana: This is an pp using HTTP/3 resetting a stream. In our common case is necko resetting a stream(make sure that necko only can reset a push stream or a http stream). The stream can be UniDi (push stream). I will actually remove check for send_stream and only check the type of  recv_stream.

neqo-http3/src/connection.rs
```diff
@@ -516,21 +472,30 @@ impl Http3Connection {
...
+        }
+        self.send_streams.remove(&stream_id);
         if let Some(s) = self.recv_streams.remove(&stream_id) {
-            s.stream_reset(error, &mut self.qpack_decoder, ResetType::App);
-            found = true;
+            s.stream_reset(error, ResetType::App)?;
         }
 
-        // Stream may be already be closed and we may get an error here, but we do not care.
+        // Stream maybe already be closed and we may get an error here, but we do not care.
```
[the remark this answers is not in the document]
ddragana: merging error....

neqo-http3/src/connection_server.rs
ddragana: I should make HttpRecvStream extend RecvStream trait than it should work. Thanks.

neqo-http3/src/control_stream_remote.rs
ddragana: not really. We can just return an empty vec.

neqo-http3/src/recv_message.rs
ddragana: Good catch. Push is not enabled so this is not a problem at the moment, but it needs fixing.

neqo-http3/src/connection_client.rs
```diff
@@ -5186,20 +5206,41 @@ mod tests {
...
         stream_id: u64,
         push_id: u64,
     ) -> Option<Datagram> {
-        let headers = vec![
+        send_push_promise_using_encoder_with_custom_headers(
+            client,
+            server,
+            stream_id,
+            push_id,
+            (String::from("my-header"), String::from("my-header")),
```
martinthomson, line 5224: Just to ensure that we have no mistakes, make the name and value different.

neqo-http3/src/connection_client.rs
```diff
@@ -5378,6 +5419,83 @@ mod tests {
         assert!(check_header_ready_and_push_promise(&mut client));
     }
 
+    // In this test there are 2 push promises that are blocked and the response header is
+    // blocked as well. After a packet is received olny the first push promises is unblocked.
```
martinthomson, line 5423: typo: only

neqo-http3/src/connection_client.rs
```diff
@@ -5378,6 +5419,83 @@ mod tests {
...
+
+        let request_stream_id = make_request_and_exchange_pkts(&mut client, &mut server, true);
+
+        // Send a PushPromise that is blocked until encoder_inst_pkt2 is process by the client.
+        let encoder_inst_pkt1 = send_push_promise_using_encoder_with_custom_headers(
+            &mut client,
+            &mut server,
+            request_stream_id,
+            0,
+            (String::from("my1"), String::from("my1")),
```
martinthomson, line 5445: myn1, myv1 perhaps?

neqo-http3/src/connection.rs
```diff
@@ -400,54 +374,56 @@ impl Http3Connection {
         }
     }
 
-    /// Returns true if the stream is a steam handled by this structture.
-    /// A push stream is the only stream that is not handled here.
+    /// If the new stream is a control stream, this function creates a proper handler
+    /// and perform a read.
+    /// if the new stream is a push stream, the function returns ReceiveOutput::PushStream
+    /// and the coller will handl it.
```
martinthomson, line 380: typo: handle, caller.

neqo-http3/src/push_stream.rs
```diff
@@ -93,35 +93,43 @@ impl RecvStream for PushStream {
...
-                            ));
+                    let to_read = id_reader.min_remaining();
+                    let mut buf = vec![0; to_read];
+                    match conn.stream_recv(self.stream_id, &mut buf[..])? {
+                        (_, true) => {
                             self.state = PushStreamState::Closed;
-                            return Ok(ReceiveOutput::NoOutput);
+                            break Ok(ReceiveOutput::NoOutput);
+                        }
+                        (0, false) => break Ok(ReceiveOutput::NoOutput),
```
martinthomson, line 103: I think you want `return Ok(...)` here.

neqo-http3/src/push_stream.rs
```diff
@@ -93,35 +93,43 @@ impl RecvStream for PushStream {
...
+                        (amount, false) => {
+                            if let Some(p) = id_reader.consume(&mut Decoder::from(&buf[..amount])) {
```
martinthomson, line 105: So this is starting to push up against the right hand side of the screen too much for me.  Find some chunk of this that can be split out, I think.

neqo-http3/src/push_stream.rs
```diff
@@ -93,35 +93,43 @@ impl RecvStream for PushStream {
...
-                                self.stream_id,
-                                Error::HttpRequestCancelled.code(),
-                            ));
+                    let to_read = id_reader.min_remaining();
+                    let mut buf = vec![0; to_read];
+                    match conn.stream_recv(self.stream_id, &mut buf[..])? {
+                        (_, true) => {
                             self.state = PushStreamState::Closed;
-                            return Ok(ReceiveOutput::NoOutput);
+                            break Ok(ReceiveOutput::NoOutput);
```
martinthomson, line 101: So if we get a push and the push ID is truncated, we don't close the connection?  That seems like it might be worth raising an error for.  I get that we ignore resets, but this is a FIN.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -69,16 +65,13 @@ impl RecvStream for NewStreamTypeReader {
...
-            Ok(ReceiveOutput::NoOutput)
-        }
+        Ok(self
+            .get_type(conn)
+            .map_or(ReceiveOutput::NoOutput, |t| ReceiveOutput::NewStream(t)))
     }
 
     fn done(&self) -> bool {
-        self.fin
+        *self == NewStreamTypeReader::Done
```
martinthomson, line 75: You can get the same result without implementing PartialEq on the incremental reader using: `matches!(self, NewStreamTypeReader::Done)`.

neqo-common/src/incrdecoder.rs
```diff
@@ -9,7 +9,7 @@ use std::mem;
 
 use crate::codec::Decoder;
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq)]
```
martinthomson, line 12: See below.

martinthomson (commented):
Looks good.  Using PartialEq seems unnecessary though.  Not sure about the code duplication either, but it's tiny enough.  A little cleanup where it is used is probably good sense though.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -82,3 +83,124 @@ impl RecvStream for NewStreamTypeReader {
...
+
+    #[test]
+    fn done_decoding() {
+        let mut t = Test::new();
+        t.decode(
+            0x3fff,
+            ReceiveOutput::NewStream(Http3StreamType::Unknown),
+            true,
+        );
+        t.decode(HTTP3_UNI_STREAM_TYPE_PUSH, ReceiveOutput::NoOutput, true);
```
martinthomson, line 197: I think that you need to include the truncated identifier option also.  That means literally encoding 0xff and nothing more.  Maybe the way to do that is to pass a pre-encoded byte array for the identifier.

martinthomson (commented):
Good.

ddragana:
I am going to do a follow up PR:
I think I am going to move push_id decoding into NewStreamTypeReader (we need that for WebTransport (see [this](https://github.com/ddragana/neqo/blob/79ca09d35e1a6f5d66da637afe41a8d456319e72/neqo-http3/src/stream_type_reader.rs#L24))). I am thinking about creating a generic incremental reader for reading frames and new stream type(I will hack something that I have in mind).

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,91 +6,211 @@
...
-                    return None;
-                }
-                Ok((0, false)) => {
-                    return None;
-                }
-                Ok((amount, false)) => {
-                    let res = self.reader.consume(&mut Decoder::from(&buf[..amount]));
-                    if res.is_some() {
-                        return res;
+            match self {
```
martinthomson, line 33: You can use `if let NewStreamTypeReader::Read { } =` here.

martinthomson (commented):
LGTM

PR #1163, opened 2021-06-16
Processing saved packet may overflow cubic cc

ddragana:
Saved packets are process with a timestamp that correspond to time they have been received. This may overflow the calculation of the time spent in ca.

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -139,7 +139,13 @@ impl WindowAdjustment for Cubic {
 
         let time_ca = self
             .ca_epoch_start
-            .map_or(min_rtt, |t| now + min_rtt - t)
+            .map_or(min_rtt, |t| {
+                if now + min_rtt < t {
+                    min_rtt
+                } else {
+                    now + min_rtt - t
+                }
+            })
             .as_secs_f64();
         let target_cubic = self.w_cubic(time_ca);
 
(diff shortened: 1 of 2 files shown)


ddragana:
This should fix bug [1716588](https://bugzilla.mozilla.org/show_bug.cgi?id=1716588) (the reports that involve cubic).

This is triggered when saved-packets are process (see [https://crash-stats.mozilla.org/report/index/288b61d0-dd5d-4fb7-a9ef-ab2510210609]).
I only have an artificial test, I could not make a test that is triggering this with a quic connection and saved packets.

neqo-transport/src/cc/cubic.rs
```diff
@@ -139,7 +139,13 @@ impl WindowAdjustment for Cubic {
 
         let time_ca = self
             .ca_epoch_start
-            .map_or(min_rtt, |t| now + min_rtt - t)
+            .map_or(min_rtt, |t| {
+                if now + min_rtt < t {
+                    min_rtt
```
martinthomson, line 144: It might be worth commenting here that this only happens when processing old packets that were saved and replayed with old timestamps.  In that case, using a fixed value is safe.

It's a shame that we can't use `saturating_sub()` and all that with `Duration` yet.  Those could make this slightly nicer.

martinthomson (commented):
Looks good.

We have quite a few time-traveling tests that could have tripped this, but they all use Reno.  Maybe we should consider switching our default CC in tests (maybe using a feature flag so that we can just build and run tests under two different configurations).

PR #1172, opened 2021-06-22
Do not cancel a stream if disallowed headers are present, just remove them

### LINKED ISSUE #1171: Do not fail stream if disalloed headers are present
We have seen a report that there is a server that returns a `Host `header. The header is a request header not a response header. neqo will reset the stream which carries the header in the response and Firefox will show an error. HTTP/2 stack does not close a stream in such a case, it just  silently removes the header and never show them to the upper stack or a web page.

The developers get confuse that that a page is working in HTTP/2 but is not working in HTTP/3.
I think we can mimic behavior of HTTP/2.

ddragana:
Fixes #1171

### DIFF
--- a/neqo-common/src/header.rs
+++ b/neqo-common/src/header.rs
@@ -1,0 +1,24 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone)]
+pub struct Header(pub String, pub String);
+
+impl Header {
+    #[must_use]
+    pub fn is_allowed_for_respone(&self) -> bool {
+        !matches!(
+            self.0.as_str(),
+            "connection"
+                | "host"
+                | "keep-alive"
+                | "proxy-connection"
+                | "te"
+                | "transfer-encoding"
+                | "upgrade"
+        )
+    }
+}
--- a/neqo-common/src/header.rs
+++ b/neqo-common/src/header.rs
@@ -5,13 +5,24 @@
 // except according to those terms.
 
 #[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone)]
-pub struct Header(pub String, pub String);
+pub struct Header {
+    name: String,
+    value: String,
+}
 
 impl Header {
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn new(name: impl ToString, value: impl ToString) -> Self {
+        Self {
+            name: name.to_string(),
+            value: value.to_string(),
+        }
+    }
+
     #[must_use]
-    pub fn is_allowed_for_respone(&self) -> bool {
+    pub fn is_allowed_for_response(&self) -> bool {
         !matches!(
-            self.0.as_str(),
+            self.name.as_str(),
             "connection"
                 | "host"
                 | "keep-alive"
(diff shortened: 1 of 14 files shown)


neqo-http3/src/recv_message.rs
```diff
@@ -377,10 +377,10 @@ impl RecvMessage {
         }
     }
 
-    fn headers_valid(&self, headers: &[Header]) -> Res<bool> {
+    fn sanitize_headers(&self, headers: &mut Vec<Header>) -> Res<()> {
```
martinthomson, line 380: I would create a new function for the stuff you extracted out.  Then this function doesn't need to take a mutable value.

You could even move the validation to `Header::is_allowed` and just call: `headers.retain(Header::is_allowed)` in `add_headers` after validating.

martinthomson (commented):
I don't much like that we need to do this, but it seems like that is how we get interop.

Maybe try the factoring I suggest.  But land away.

neqo-http3/src/connection_client.rs
```diff
@@ -296,10 +296,10 @@ impl Http3Client {
...
         // Transform pseudo-header fields
         let mut final_headers = vec![
-            (":method".into(), method.to_owned()),
-            (":scheme".into(), scheme.to_owned()),
-            (":authority".into(), host.to_owned()),
-            (":path".into(), path.to_owned()),
+            Header(":method".into(), method.to_owned()),
+            Header(":scheme".into(), scheme.to_owned()),
+            Header(":authority".into(), host.to_owned()),
+            Header(":path".into(), path.to_owned()),
```
martinthomson, line 302: You might do better with `Header::new(impl ToString, impl ToString)` here so that you can avoid all the .into() or .to_owned() calls.  That will create new instances as needed (or just convert an already owned value without copying).

You wouldn't have to change all those call sites, but you could.

neqo-common/src/header.rs
```diff
@@ -0,0 +1,24 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone)]
+pub struct Header(pub String, pub String);
+
+impl Header {
+    #[must_use]
+    pub fn is_allowed_for_respone(&self) -> bool {
```
martinthomson, line 12: typo in name

neqo-http3/src/recv_message.rs
```diff
@@ -377,10 +379,10 @@ impl RecvMessage {
         }
     }
 
-    fn sanitize_headers(&self, headers: &mut Vec<Header>) -> Res<()> {
+    fn headers_valid(&self, headers: &[Header]) -> Res<()> {
         let mut method_value: Option<&String> = None;
         let mut pseudo_state = 0;
-        for header in headers.iter() {
+        for header in headers {
             let is_pseudo = Self::track_pseudo(&header.0, &mut pseudo_state, &self.message_type)?;
```
martinthomson: Maybe for later, but large chunks of this function can now move to Header.  Think about a function that returns `Res<enum {
  Pseudo(MessageType, enum { Method, Authority, ... } ),
  Regular,
}>`

neqo-common/src/header.rs
```diff
@@ -0,0 +1,24 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone)]
+pub struct Header(pub String, pub String);
```
martinthomson, line 8: Future change here too: `{ pub name: String, pub value: String }`

martinthomson (commented):
LGTM.

Lots of opportunities for improvement however.  I'll leave it to you to raise issues for those or fix them.  They are largely mechanical.

ddragana:
I have already planned to implement all this changes that you suggested and also some more, e.g. maybe add Headers structure and extract all verification from recv_message.

I have stopped here, because I have another rather urgent necko bug that I want to figure out and fix before some people leave for PTO today.
I will implement this improvement in a separate PR.

martinthomson (approved):
That's a huge improvement.  We can track the other things in issues.

PR #1177, opened 2021-07-14
Set idle

martinthomson:
This builds on #1160 as a discrete change.

PR #1160, opened 2021-06-11
Keep-alives

martinthomson:
No tests, and not enough time to do anything.

Is this a good design?

PR #1178, opened 2021-07-14
HTTP portion of keep-alive

martinthomson:
This uses the transport keep-alive functions to ensure that connections
don't time out when there are outstanding requests.

I've also added some keep-alive capability for server push, but this is
limited.  Promised pushes won't do.  Only those promises that the client
chooses to read from will be used.  This ensures that ignored pushes
don't keep things open unnecessarily.

This builds on #1177 (which in turn builds on #1160).  There is just one new commit here.


martinthomson:
Oh, I forgot to add notes on the test changes.  In order to force the client and server into an idle state, I had to make a few small changes in the tests.  Mostly this just avoids generating datagrams that are dropped by the test framework.  This was generally OK in most tests, but I did need to fix one that assumed an ACK that was previously being generated in response to apparent packet loss.  Without loss, there is no ACK (at least today).

I needed to modify the initial request exchange to allow for the "Set Capacity" QPACK encoder instruction to be sent along with the request.  I'm not entirely sure why this was not being sent previously, but it might have been in one some of the dropped packets previously, so it wasn't being received.  The result is a tiny increase in complexity, but an increase in reliability, so I think that it is fine.

neqo-http3/src/connection_client.rs
```diff
@@ -4404,6 +4458,30 @@ mod tests {
...
+        client.process_input(dgram.unwrap(), now());
+
+        let mut buf = [0; 16];
+        let (read, fin) = client.push_read_data(now(), 0, &mut buf).unwrap();
+        assert!(read < buf.len());
+        assert!(!fin);
+
+        force_idle(&mut client, &mut server);
+        let idle_timeout = ConnectionParameters::default().get_idle_timeout();
+        assert_eq!(client.process_output(now()).callback(), idle_timeout / 2);
```
ddragana, line 4482: This test is no really testing the push stream. connect_and_send_request(..) will open a stream tha well also need keeping alive. You need to send response for request_stream_id.

neqo-http3/src/connection_client.rs
```diff
@@ -4404,6 +4458,30 @@ mod tests {
         assert_eq!(client.cancel_push(0), Err(Error::InvalidStreamId));
     }
 
+    /// We can't keep the connection alive on the basis of a push promise,
+    /// nor do we want to if the push promise is not interesting to the client.
+    /// We do the next best thing, which is keep any push stream alive if the
+    /// client reads from it.
+    #[test]
```
ddragana, line 4465: We also need a test that  push_promise will no cause keep live to be active. Probably also that the keep alive is not active if client does not call read_push_data.

martinthomson: Yeah, I think that I can manage that in the same test: promise push, send response, read response, client is idle, send (incomplete) push data, read push data, client is not idle

ddragana (dismissed):

PR #1186, opened 2021-07-27
Increase flow control credit on streams

martinthomson:
We already allow flow control credit to be customized.  This goes a little further and makes the stream flow control limits expand as they increase, up to the connection-level flow control limit.  This can be overridden on a per-stream basis.

The exact configuration and algorithm are largely TBD, just looking for feedback for now.


neqo-transport/src/fc.rs
```diff
@@ -286,36 +300,46 @@ impl ReceiverFlowControl<()> {
...
+            self.frame_sent(self.next_limit());
         }
     }
 }
 
 impl ReceiverFlowControl<StreamId> {
+    /// Double the active limit every time we need to send an update.
+    fn increase_active(&mut self) {
+        if self.active < self.max_active {
+            self.active = min(self.max_active, self.active * 2);
```
ddragana, line 318: I think we need to have a higher initial limit, otherwise we will have a slow download for a couple of RTTs and probably get into app-limited cc state.
My initial idea was to have a low initial value in the transport parameter, but to increase it immediately to something high for the streams from the focus tab and/or high priority items. This kind of change will need investigation.

I would suggest just to bump the initial value to something high.(maybe 12MB that is HTTP2 limit; HTTP2 also was the same session limit). This is simpler and I do not see a need for more complex solution.

PR #1187, opened 2021-07-29
Use session flow control

ddragana:
We set the flow control limit to the max possible and therefore we have not implemented actually calculating retired data.
This is fine with the limit set to max possible, but this is needed if we set it to something smaller.

This changes adds calculation of retired bytes.
Ensuring that the other side does not exceed the limit will be added in the follow up commit.

### DIFF
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -863,3 +863,61 @@ fn session_flow_control_stop_sending_state_data_recvd() {
         SMALL_MAX_DATA
     );
 }
+
+#[test]
+fn session_flow_control_affects_all_streams() {
+    const SMALL_MAX_DATA: usize = 1024;
+
+    let mut client = default_client();
+    let mut server = new_server(
+        ConnectionParameters::default().max_data(u64::try_from(SMALL_MAX_DATA).unwrap()),
+    );
+
+    connect(&mut client, &mut server);
+
+    let stream_id = client.stream_create(StreamType::UniDi).unwrap();
+    assert_eq!(
+        client.stream_avail_send_space(stream_id).unwrap(),
+        SMALL_MAX_DATA
+    );
+
+    let stream_id2 = client.stream_create(StreamType::UniDi).unwrap();
+    assert_eq!(
+        client.stream_avail_send_space(stream_id2).unwrap(),
+        SMALL_MAX_DATA
+    );
+
+    assert_eq!(
+        client
+            .stream_send(stream_id, &[b'a'; SMALL_MAX_DATA / 2 + 1])
+            .unwrap(),
+        SMALL_MAX_DATA / 2 + 1
+    );
+
+    assert_eq!(
+        client.stream_avail_send_space(stream_id).unwrap(),
+        SMALL_MAX_DATA / 2 - 1
+    );
+    assert_eq!(
+        client.stream_avail_send_space(stream_id2).unwrap(),
+        SMALL_MAX_DATA / 2 - 1
+    );
+
+    exchange_data(&mut client, &mut server);
+
+    let mut buf = [0x0; SMALL_MAX_DATA];
+    let (read, _) = server.stream_recv(stream_id, &mut buf).unwrap();
+    assert_eq!(read, SMALL_MAX_DATA / 2 + 1);
+
+    exchange_data(&mut client, &mut server);
+
+    assert_eq!(
+        client.stream_avail_send_space(stream_id).unwrap(),
+        SMALL_MAX_DATA
+    );
+
+    assert_eq!(
+        client.stream_avail_send_space(stream_id2).unwrap(),
+        SMALL_MAX_DATA
+    );
+}
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -152,7 +152,7 @@ fn exchange_data(client: &mut Connection, server: &mut Connection) {
 
 #[test]
 fn sending_max_data() {
-    const SMALL_MAX_DATA: usize = 16383;
+    const SMALL_MAX_DATA: usize = 2048;
 
     let mut client = default_client();
     let mut server = new_server(
@@ -171,7 +171,7 @@ fn sending_max_data() {
 
     assert_eq!(
         client
-            .stream_send(stream_id, &vec![b'a'; RECV_BUFFER_SIZE].into_boxed_slice())
+            .stream_send(stream_id, &[b'a'; SMALL_MAX_DATA + 1])
             .unwrap(),
         SMALL_MAX_DATA
     );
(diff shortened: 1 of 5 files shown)


ddragana:
I forgot to deal with cancelled streams.

ddragana:
This PR build on top of what we already have. I may do a bit of refactoring of this code. But I was thinking to do that in a separate PR.
This PR is blocking  a bug hat I am working on to control the flow control limits and experiment with them.

martinthomson (commented):
Yikes, we missed quite a bit by not implementing session-level flow control properly.

I have a bunch of comments, but I think that the general shape of this is right.

martinthomson:
Oh, I don't see any tests here that show that sending on one stream results in flow control for another stream being affected.  Do we already have that test?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -618,3 +724,145 @@ fn increase_decrease_flow_control() {
...
+            .unwrap(),
+        SMALL_MAX_DATA
+    );
+
+    let out1 = client.process(None, now()).dgram();
+    // Delay his packet and lett the server receive fin first (it will enter SizeKnown state).
+    client.stream_close_send(stream_id).unwrap();
+    let out2 = client.process(None, now()).dgram();
+
+    mem::drop(server.process(out2, now()).dgram());
```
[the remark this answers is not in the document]
ddragana: I just do not care about the packet.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -618,3 +724,145 @@ fn increase_decrease_flow_control() {
...
+
+    server
+        .stream_stop_sending(stream_id, Error::NoError.code())
+        .unwrap();
+
+    let out = server.process(out1, now()).dgram();
+    let _out = client.process(out, now()).dgram();
+
+    // The flow control should have been updated and the clinet caan again send
+    // SMALL_MAX_DATA.
```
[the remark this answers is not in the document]
ddragana: session_flow_control_stop_sending_state_recv tests the case when we need to wait for the reset. Because of the way the frames are send the state transfers from AbortReading -> WaitingForReset(while processing acks) -> Recvd (when reset is processed). In this test we call exchange_packets because we need the reset frame. MAX_DATA is send after RESET is received.

session_flow_control_stop_sending_state_size_known tests the case where we have the final size and we do not need a RESET frame. Here already sop_sending() call releases flow control and MAX_DATA is send together with STOP_SENDING frame.

I will add some comments to the test.

ddragana:
session_flow_control_stop_sending_state_recv(the new version that will be push shortly) tests that sending on one stream affects other streams, but this is tests only as a side effect.

I will add a test that tests only hat interaction.

neqo-transport/src/recv_stream.rs
```diff
@@ -1146,18 +1249,511 @@ mod tests {
...
+        assert_eq!(fc.borrow().retired(), SW / 2);
+
+        // Receiving duplicate frames (already consumed data) will not cause an error or
+        // change fc.
+        s.inbound_stream_frame(true, 0, &[0; SW_US / 2]).unwrap();
+        // the stream does not have fc any more. We can only check the session fc.
+        assert_eq!(fc.borrow().consumed(), SW / 2);
+        assert_eq!(fc.borrow().retired(), SW / 2);
+
+        // Receiving frame pass the final size of a stream will NOT return an error.
```
martinthomson, line 1625:
```suggestion
        // Receiving frame past the final size of a stream will NOT return an error.
```

neqo-transport/src/recv_stream.rs
```diff
@@ -1146,18 +1249,511 @@ mod tests {
...
+        s.inbound_stream_frame(true, 0, &[0; SW_US / 2 + 1])
+            .unwrap();
+
+        // Receiving frame pass the stream's fc limit will not return an error.
```
martinthomson, line 1629:
```suggestion
        // Receiving frame past the stream's fc limit will not return an error.
```
you could just combine the comments though.

martinthomson (commented):
I like the new tests.  I still don't think that the final size belongs on the flow control objects.

neqo-transport/src/recv_stream.rs
```diff
@@ -1146,18 +1249,511 @@ mod tests {
...
+        session_c: u64,
+        session_r: u64,
+        s1_c: u64,
+        s1_r: u64,
+    ) {
+        assert_eq!(session_fc.borrow().consumed(), session_c);
+        assert_eq!(s1.fc().unwrap().consumed(), s1_c);
+        assert_eq!(session_fc.borrow().retired(), session_r);
+        assert_eq!(s1.fc().unwrap().retired(), s1_r);
+    }
```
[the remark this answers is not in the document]
ddragana: Yeah, I was not really sure which version was easier to read. I have try a couple.  I will use your suggestion.

neqo-transport/src/fc.rs
```diff
@@ -318,6 +355,34 @@ impl ReceiverFlowControl<StreamId> {
...
+            if consumed > self.max_allowed {
+                qtrace!("Stream RX window exceeded: {}", consumed);
+                return Err(Error::FlowControlError);
+            }
+            let new_consumed = consumed - self.consumed;
+            self.consumed = consumed;
+            Ok(new_consumed)
+        } else {
+            Ok(0)
+        }
```
martinthomson, line 378: maybe you can just do: 
```rust
    let x = consumed.saturating_sub(self.consumed); 
    self.consume(x)?; 
    Ok(x)
```

neqo-transport/src/recv_stream.rs
```diff
@@ -374,16 +384,62 @@ impl RecvStreamState {
...
+            (false, true) => {
+                if consumed > fc.consumed() {
+                    return Err(Error::FinalSizeError);
+                }
+            }
+            (_, false) => {
+                if fin && consumed < fc.consumed() {
+                    return Err(Error::FinalSizeError);
+                }
+            }
```
martinthomson, line 433:
```suggestion
            (true, false) => {
                if consumed < fc.consumed() {
                    return Err(Error::FinalSizeError);
                }
            }
            (false, false) => (),
```

martinthomson: Another option:

```rust
let final_size_ok = match (fin, final_size_reached) {
  (true, true) => consumed == fc.consumed(),
  (true, false) => consumed >= fc.consumed(),
  (false, true) => consumed <= fc.consumed(),
  (false, false) => true,
};
if !final_size_ok {
  return Err(Error::FinalSizeError);
}

martinthomson (dismissed):
This looks good.  Thanks for all the extra testing.  We might even have TOO much testing now, but that's probably not a problem.

neqo-transport/src/fc.rs
ddragana: his does not work because consume is not implemented for ReceiverFlowControl<StreamId> because it is not needed. ReceiverFlowControl<StreamId> only uses set_consumed.

martinthomson: Flip it the other way around then.

PR #1194, opened 2021-08-06
Ignore acknowledgment of keep-alive PING

martinthomson:
Should fix the gecko crasher.


martinthomson:
Now to do the hard part: add a test so that this never happens again.

ddragana:
There is still one problem with keep-alive code and that is that if PING is not lost keep_alive_outstanding will never set o false again and we will only once send keep-alive ping for he whole connection(path).
Can you look at #1193?

martinthomson:
This was incomplete.

PR #1198, opened 2021-08-12
Split stream-related recovery tokens out

martinthomson:
In part, the reason we had issues with the keep-alive patch was that we
had `_ => panic!()` clauses in our `match` statements.  This patch moves
the stream-related pieces of the enum into a new structure.  This might
add an extra discriminator to the layout, but the overall effect is
otherwise minimal.  And we no longer have catch-all clauses in our match
statements.


ddragana (approved):
Thanks for doing this!
I was planning o do something similar as well to avoid  catch-all clauses.

Issue #1197, opened 2021-08-09
HTTP Priorities


ddragana:
https://httpwg.org/http-extensions/draft-ietf-httpbis-priority.html

ddragana:
For structured headers we can use https://crates.io/crates/sfv

larseggert:
It appears this landed in #1200.

PR #1200, opened 2021-08-13
Client API for Extensible Prioritization Scheme for HTTP

mb:
Specified in [httpbis-priority]

Start of #1197.

Extending the fetch function with one parameter to specify the priority.
The Priority type is easy to use if the user doesn't want to specify a priority, they can pass Priority::default().
And the priority struct can be used on the server side as well, so it might be good to move it to `neqo-common` soon.

An alternative would be to change the fetch calling convention to make it easier to call when not all parameters should be specified, but since I think that almost always all parameters are specified I didn't take this route:

```rust
/* decision: see diff */
client.fetch(..., Priority::default()); // default priority
client.fetch(..., Priority:new(4, false)); // specified priority

/* alternative */
struct Request { ... }

fn fetch(now: Instant, Resquest) -> { ... }

client.fetch(now(), Request::new("GET", "https", "something.com", "/")); // default priority
client.fetch(now(), Request::new("GET", "https", "something.com", "/").with_priority(4, false)); // specified priority
client.fetch(now(), Request::new("GET", "https", "something.com", "/").with_headers(...).with_priority(4, false));
```

The `priority_update` only takes changing Priorities, because keeping track of each priority of each stream doesn't take much resources.
It takes all changing resources at once, so that it can send the `PRIORITY_UPDATE` frame instantly without sending too many  `PRIORITY_UPDATE`-frames when multiple streams change the priority.

Alternative, it could take two slices. The downside of this would be, that it has to check, that both slices have the same length and the caller has to create the `priority` slice anyway, so creating a (stream_id, priority) slice doesn't differ much:

```rust
fn priority_update(&mut self, update: &[(u64, Priority)]) -> Res<()> { todo!() } // decision
fn priority_update(&mut self, stream_ids: &[u64], priorities: &[Priority)]) -> Res<()> { // alternative
    assert_eq!(stream_ids.len(), priorities.len());
    todo!()
}
```

[httpbis-priority]: https://datatracker.ietf.org/doc/html/draft-kazuho-httpbis-priority-04


neqo-http3/src/connection_client.rs
```diff
@@ -335,6 +340,23 @@ impl Http3Client {
         Ok(id)
     }
 
+    /// Send an [`PRIORITY_UPDATE`-frame][1]. Only priorities that changes have to be passed to this
+    /// function.
+    /// # Error
+    /// `InvalidStreamId` if the stream does not exist
+    /// [1]: https://datatracker.ietf.org/doc/html/draft-kazuho-httpbis-priority-04#section-5.2
+    pub fn priority_update(&mut self, client_id: u64, priority: Priority) -> Res<()> {
```
martinthomson, line 348:
```suggestion
    pub fn priority_update(&mut self, stream_id: u64, priority: Priority) -> Res<()> {
```

neqo-http3/src/connection.rs
```diff
@@ -604,12 +611,14 @@ impl Http3Connection {
...
         send_stream: SendMessage,
         recv_stream: Box<dyn RecvStream>,
+        priority: Priority,
     ) {
         if send_stream.has_data_to_send() {
             self.streams_have_data_to_send.insert(stream_id);
         }
         self.send_streams.insert(stream_id, send_stream);
         self.recv_streams.insert(stream_id, recv_stream);
+        self.stream_priorities.insert(stream_id, priority);
```
martinthomson, line 621: Priority *signaling* is a property of a recv stream, isn't it?  So this should apply to push streams also.

neqo-http3/src/connection.rs
```diff
@@ -94,6 +97,8 @@ impl Http3Connection {
             streams_have_data_to_send: BTreeSet::new(),
             send_streams: HashMap::new(),
             recv_streams: HashMap::new(),
+            stream_priorities: HashMap::new(),
+            send_priority_update: false,
```
martinthomson, line 101: I'm not seeing where this is read.  I assume that comes next.

martinthomson (commented):
The API piece looks good.

I see some missing pieces:
* Signaling for changes in priority
* Setting send priority at the transport layer

I think that I'd prefer if the priority piece were attached to RecvStream.  I know that complicates things as it goes through the trait, but it would simplify management.  You could add a getter and setter to the trait, or you could make a wrapper with a `Priority` and a `Box<dyn RecvStream>` that the base handler holds instead (the wrapper can implement Deref for RecvStream to smooth the change).  There isn't much point in having priority for control streams (we should just set a fixed value in the transport, probably higher than the default), but there is little value in isolating only HTTP and push streams.  The extra two bytes for three streams is a trivial cost to carry.

neqo-http3/src/connection.rs
ddragana: I think this should be on the control stream and we actually change priority of a stream only when its update get sent to the peer.

neqo-http3/src/connection.rs
ddragana: Note: I suggested that push stream are less important at the moment, because they are not used in Firefox at the moment.
You are right that Property should be a property of a RecvStream.

ddragana:
> The API piece looks good.
> 
> I see some missing pieces:
> 
>     * Signaling for changes in priority

Yes, I missed that, we should have an event (i.e. CientEvent and ServerEvent) when the peer changes a priority

> 
>     * Setting send priority at the transport layer
> 

I suggested to do this as a follow up, therefore it is missing in this PR. I will open an issue for this because I would like to discuss the proper behavior.

neqo-http3/src/connection.rs
martinthomson: Yes, the frame is sent on the control stream, but we can make changes to our sending piece immediately (without waiting for the control stream frames to be sent).

ddragana: That is true.
We need to find a way to handle:
priority was Priority{3, false}
priority_update(0, Priority{4, false}) -> this update has not been sent yet, before the following update
priority_update(0, Priority{3, false}) -> now neqo does not need to send an update anymore.

martinthomson: Oh, of course.  It might be easiest to just have a list of pending priority updates as `(u64, Priority)` tuples attached to the control stream.  That would mean a linear search every priority update, but it shouldn't be HUGE numbers of items in that list.  Indeed, you could put a hard limit on the length of the list and throw out old priority updates if the list got too long; we don't want to spend all our uplink on sending priority updates.

mb: I'm using VecDeque<u64> now and checking which to Priority to update to from the `recv_conn` variable. I didn't implement anything to throw out some priority updates if there are too many of them.

neqo-http3/src/control_stream_local.rs
```diff
@@ -39,8 +43,35 @@ impl ControlStreamLocal {
     }
 
     /// Send control data if available.
-    pub fn send(&mut self, conn: &mut Connection) -> Res<()> {
+    pub fn send(
+        &mut self,
+        conn: &mut Connection,
+        recv_conn: &mut HashMap<u64, Box<dyn RecvStream>>,
```
mb, line 49: control_stream now depends on the recv_conn attribute of `Http3Connection`

neqo-http3/src/lib.rs
```diff
@@ -321,6 +323,15 @@ pub trait RecvStream: Debug {
     fn done(&self) -> bool;
     fn stream_type(&self) -> Http3StreamType;
     fn http_stream(&mut self) -> Option<&mut dyn HttpRecvStream>;
+    fn priority(&self) -> Priority;
+
+    /// only relevant for streams supporting priority updates
```
mb, line 328: I've added some traits that are only relevant for requests and push streams

neqo-http3/src/priority.rs
```diff
@@ -0,0 +1,85 @@
...
+        write!(priority, "{}", self).unwrap();
+
+        HFrame::PriorityUpdateRequest {
+            element_id: stream_id,
+            priority,
+        }
+    }
+}
+
+impl fmt::Display for Priority {
```
mb, line 53: Is implementing Display a okay trait for this purpose?

mb (commented):
I think functionality wise this could be done (client side, request only). I'm not sure about some design decisions and added comments for them.

neqo-http3/src/control_stream_local.rs
ddragana: It is ok. I think this is a good approach.

ddragana (commented):
Thanks. This is moving into the right direction.

neqo-http3/src/priority.rs
martinthomson: I think that this works nicely as it is.

neqo-http3/src/priority.rs
```diff
@@ -0,0 +1,74 @@
...
+                urgency: 3,
+                incremental: false,
+            } => None,
+            other => Some(Header::new("priority", format!("{}", other))),
+        }
+    }
+
+    pub fn encode_request_frame(self, stream_id: u64) -> HFrame {
+        let mut priority = Vec::new();
+        write!(priority, "{}", self).unwrap();
```
martinthomson, line 44: I would copy *self* into the HFrame.  It's small enough and you then don't need to worry about copying a Vec instance around.

neqo-http3/src/lib.rs
```diff
@@ -321,6 +323,15 @@ pub trait RecvStream: Debug {
...
+    #[allow(unused_variables)]
```
martinthomson, line 329: You shouldn't need default implementations for these if you do as Dragana suggests.  Which means that you don't need to suppress warnings (you should avoid doing that anyway; renaming `priority` to `_priority` would work here).

martinthomson (commented):
Another piece that might be worth thinking about: what would we do on our server to implement priorities?  My sense is that we would map the numbers to our internal priority scheme, with a little compression, ignoring the `i`.  (That can go into a follow-up.)

neqo-http3/src/lib.rs
mb: Yea, `HttpRecvStream` looks appropriate for these functions. I even moved priority() there too.

neqo-http3/src/control_stream_local.rs
```diff
@@ -39,8 +43,35 @@ impl ControlStreamLocal {
...
+                let mut enc = Encoder::new();
+                hframe.encode(&mut enc);
+                if conn.stream_send_atomic(stream_id, &enc)? {
+                    update_stream.priority_update_sent();
+                } else {
+                    self.outstanding_priority_update.push_front(update_id);
+                }
+            }
+
             if !self.buf.is_empty() {
```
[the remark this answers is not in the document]
mb: I've moved that part below: https://github.com/mozilla/neqo/pull/1200/commits/553366156ecc0e6c44e5f31db006f8805572197a

neqo-http3/src/connection.rs
```diff
@@ -158,7 +158,8 @@ impl Http3Connection {
     )] // Until we require rust 1.53 we can't use or_patterns.
     pub fn process_sending(&mut self, conn: &mut Connection) -> Res<()> {
         // check if control stream has data to send.
-        self.control_stream_local.send(conn)?;
+        self.control_stream_local
+            .send(conn, &mut self.recv_streams)?;
```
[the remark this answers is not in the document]
mb: There is still some other stuff happening in that function. I could split the function into two: `send` and `send_priority_updates` if that would make it clearer

neqo-http3/src/priority.rs
mb: Done. But unfortunately I don't know if I can get rid of the two allocations when encoding: https://github.com/mozilla/neqo/pull/1200/commits/e1c6398920b9989ef0031ae580e3528fc1c7990b

neqo-http3/src/recv_message.rs
```diff
@@ -76,6 +76,8 @@ pub(crate) struct RecvMessage {
     conn_events: Box<dyn RecvMessageEvents>,
     push_handler: Option<Rc<RefCell<PushController>>>,
     stream_id: u64,
+    priority: Priority,
+    last_send_priority: Priority,
```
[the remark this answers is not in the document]
mb: I think it is good to have Priority type represent only one priority. Maybe having https://github.com/mozilla/neqo/pull/1200/files#diff-bcde5dab41d27b601e3435a616c94b01dfe7c210d0396aa3763ba16addeb9a47R336 return both the last send priority and the current priority would archive the same?

neqo-http3/src/connection.rs
mb: >  Indeed, you could put a hard limit on the length of the list and throw out old priority updates if the list got too long; we don't want to spend all our uplink on sending priority updates.

Do we want to put a limit somewhere or somehow only send priority_updates as long as we don't have something else to send?

mb:
Thanks for the suggestions! I finished with my changes again for client side priority_update for RecvStreams again.

neqo-http3/src/connection.rs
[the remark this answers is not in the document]
ddragana: it is fine to leave he name, send().

neqo-http3/src/control_stream_local.rs
```diff
@@ -55,9 +55,10 @@ impl ControlStreamLocal {
                     Some(update_stream) => update_stream,
                     None => continue,
                 };
-                if update_stream.stream_type() != Http3StreamType::Http {
-                    continue;
-                }
+                // can unwrap here, because
```
ddragana: please end the sentence. :)

neqo-http3/src/connection_client.rs
```diff
@@ -357,10 +358,11 @@ impl Http3Client {
             stream.priority_update(priority);
             self.base_handler
                 .control_stream_local
```
ddragana, line 360: please also do not expose control_stream_local.
Create a function queue_update_priority in Http3Connection that calls `self.control_stream_local.queue_update_priority().`

neqo-http3/src/lib.rs
```diff
@@ -341,6 +332,12 @@ pub trait HttpRecvStream: RecvStream {
     /// # Errors
     /// An error may happen while reading a stream, e.g. early close, protocol error, etc.
     fn read_data(&mut self, conn: &mut Connection, buf: &mut [u8]) -> Res<(usize, bool)>;
+
+    fn priority(&self) -> Priority;
+
+    fn priority_update(&mut self, priority: Priority);
+    fn priority_update_outstanding(&self) -> bool;
+    fn priority_update_sent(&mut self);
```
ddragana, line 340: these 4 functions manipulate priorities. 
What about having a PriorityHandler that implements last 3 functions? and HttpRecvStream has only:

`priority(&mut self) -> &'a mut PriorityHandler`

neqo-http3/src/recv_message.rs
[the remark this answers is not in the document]
ddragana: Take a look at my suggestion to have PriorityHandler.

ddragana (commented):
This part is almost done, except maybe adding PriorityHandler.
There are some stuff missing:
- reading PRIORITY_UPDATE frames
- sending events when a priority has changed (this will be useful for tests)
- tests
- changing stream priority in neqo-transport (I suggested to put this is a separate PR)

neqo-http3/src/connection.rs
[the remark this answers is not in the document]
martinthomson: Though I always appreciate functional decomposition when functions get big.

neqo-http3/src/connection.rs
```diff
@@ -630,8 +630,9 @@ impl Http3Connection {
             .http_stream()
             .ok_or(Error::InvalidStreamId)?;
 
-        if stream.priority() != priority {
-            stream.priority_update(priority);
+        let priority_handler = stream.priority_handler_mut();
+        if priority_handler.priority() != priority {
+            priority_handler.priority_update(priority);
```
ddragana, line 635: This can all be in a single function maybe_update_priority that returns true if priority is updated otherwise false.

ddragana:
Thanks for he update. 
My comments are only extracting code hat belongs together into separate functions. It will make code easier to read.

neqo-http3/src/priority.rs
```diff
@@ -37,6 +38,30 @@ impl Priority {
...
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<Priority, &'static str> {
+        let dict = Parser::parse_dictionary(bytes)?;
+        let urgency = match dict.get("u") {
+            Some(ListEntry::Item(Item {
+                bare_item: BareItem::Integer(u),
+                ..
+            })) if (0i64..=7).contains(u) => *u as u8,
+            None => 3,
+            _ => return Err("priority: invalid urgency value"),
```
ddragana, line 50: please return Error::HttpFrame.

If you want a more specific error, you can add a new one, but at the end this will be mapped into Error::HttpFrame.

neqo-http3/src/priority.rs
```diff
@@ -37,6 +38,30 @@ impl Priority {
...
+        };
+        let incremental = match dict.get("i") {
+            Some(ListEntry::Item(Item {
+                bare_item: BareItem::Boolean(i),
+                ..
+            })) => *i,
+            None => false,
+            _ => return Err("priority: invalid incremental value"),
```
ddragana, line 58: same here.

neqo-http3/src/control_stream_local.rs
```diff
@@ -78,25 +78,18 @@ impl ControlStreamLocal {
...
-                    } else {
-                        self.outstanding_priority_update.push_front(update_id);
+                    if let Some(hframe) = priority_handler.maybe_encode_frame(update_id) {
+                        let mut enc = Encoder::new();
+                        hframe.encode(&mut enc);
+                        if conn.stream_send_atomic(stream_id, &enc)? {
+                            priority_handler.priority_update_sent();
+                        } else {
+                            self.outstanding_priority_update.push_front(update_id);
+                        }
```
ddragana, line 92: Nit: everything inside `if !self.buf.is_empty() { .... }` can be a separate function

ddragana (commented):
This looks good.
There is a suggestion to factor out the writing priorities into a separate function.

neqo-http3/src/push_controller.rs
```diff
@@ -475,6 +475,10 @@ impl RecvMessageEvents for RecvPushEvents {
         );
     }
 
+    fn priority_update(&self, stream_id: u64, priority: Priority) {
+        self.push_handler.borrow_mut().new_stream_event(self.push_id, Http3ClientEvent::PriorityUpdate { stream_id, priority});
+    }
+
```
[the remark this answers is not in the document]
ddragana: why you need this. client cannot receive priority updates for push streams.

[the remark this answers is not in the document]
mb: I've added it so that the client can terminate the connection instead of ignoring the frame. From [Section 7.2](https://httpwg.org/http-extensions/draft-ietf-httpbis-priority.html#section-7.2)

> PRIORITY_UPDATE frames of either type are only sent by clients. If a client receives a PRIORITY_UPDATE frame, this MUST be treated as a connection error of type H3_FRAME_UNEXPECTED.

ddragana (commented):
This looks good!

Just one comment.

mb:
Thanks for the review :). I've made my last change now and think that it is ready for a last look and merge.

neqo-transport/src/fc.rs
```diff
@@ -448,6 +448,11 @@ impl RemoteStreamLimit {
         }
     }
 
+    pub fn is_in_limit(&self, stream_id: StreamId) -> bool {
```
ddragana, line 451: maybe call it is_allowed()?

neqo-transport/src/fc.rs
```diff
@@ -448,6 +448,11 @@ impl RemoteStreamLimit {
...
+        let stream_idx = stream_id.as_u64() >> 2;
+        self.streams_fc.check_allowed(stream_idx)
+    }
+
     pub fn is_new_stream(&mut self, stream_id: StreamId) -> Res<bool> {
         let stream_idx = stream_id.as_u64() >> 2;
         if !self.streams_fc.check_allowed(stream_idx) {
```
ddragana, line 458: You can use is_in_limit() function here.

neqo-transport/src/streams.rs
```diff
@@ -61,6 +61,10 @@ impl Streams {
         }
     }
 
+    pub fn is_in_remote_bidi_limit(&self, stream_id: StreamId) -> bool {
+        self.remote_stream_limits[StreamType::BiDi].is_in_limit(stream_id)
```
ddragana, line 65: maybe call the function `is_stream_id_allowed`
You are not checking the type of the stream in this function. The caller is responsible for that, that is fine, but you can make it generic:
```
pub fn is_stream_id_allowed(&self, stream_id: StreamId) -> bool {
    self.remote_stream_limits[stream_id.stream_type()].is_in_limit(stream_id)
}
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2174,6 +2174,10 @@ impl Connection {
         }
     }
 
+    pub fn is_in_remote_bidi_limit(&self, stream_id: StreamId) -> bool {
```
ddragana, line 2177: maybe `is_stream_id_allowed`?

neqo-http3/src/connection_server.rs
```diff
@@ -193,13 +193,20 @@ impl Http3ServerHandler {
...
                         }
                         HFrame::PriorityUpdateRequest { element_id, priority } => {
-                            // TODO: check that the element_id references a request stream
-                            //       within the client-sided bidirectional stream limit
+                            // check that the element_id references a request stream
+                            // within the client-sided bidirectional stream limit
+                            let element_stream_id = StreamId::new(element_id);
+                            if !element_stream_id.is_bidi()
+                                || !element_stream_id.is_client_initiated()
+                                || !conn.is_in_remote_bidi_limit(element_stream_id)
```
ddragana, line 205: Can you write a test for this?
You can copy his one: https://github.com/mozilla/neqo/blob/37bb62fa7b43ac624406090ba4f94f4cc082790b/neqo-http3/src/server.rs#L560
and send PRIORITY_UPDATE with a wrong ID.

neqo-http3/src/control_stream_local.rs
```diff
@@ -51,6 +63,45 @@ impl ControlStreamLocal {
...
+            let priority_handler = update_stream.http_stream().unwrap().priority_handler_mut();
+
+            // in case multiple priority_updates were issued, ignore now irrelevant
+            if let Some(hframe) = priority_handler.maybe_encode_frame(update_id) {
+                let mut enc = Encoder::new();
+                hframe.encode(&mut enc);
+                if conn.stream_send_atomic(stream_id, &enc)? {
+                    priority_handler.priority_update_sent();
+                } else {
+                    self.outstanding_priority_update.push_front(update_id);
```
ddragana, line 102: You need to exit loop here, otherwise this will loop for ever.

Please add a test for this. (you can set a small max_data limit by setting the parameter in ConnectionParameters: `ConnectionParameters::default().max_data(...)` and you can fill the flow control by sending stream data.

neqo-http3/src/server.rs
```diff
@@ -960,6 +970,16 @@ mod tests {
...
 
+    /*
+    #[test]
+    fn test_client_priority_update() {
+        let (mut hconn, mut peer_conn) = connect();
+        let client = Http3ClientEvent::new()
+        let stream_id = peer_conn.stream_create(StreamType::BiDi).unwrap();
+        peer_conn.stream_send(stream_id, REQUEST_WITH_BODY).unwrap();
+    }
+    */
```
ddragana, line 981: Do you need this test?

neqo-http3/tests/priority.rs
```diff
@@ -0,0 +1,171 @@
...
+        Http3ServerEvent::PriorityUpdate {
+            stream_id: update_id,
+            priority,
+        } => {
+            assert_eq!(update_id, stream_id);
+            assert_eq!(priority, update_priority);
+        }
+        other => panic!("unexpected server event: {:?}", other),
+    }
+}
```
ddragana, line 171: Please add test for the multiple priority update:
- the initial priority is 5 and priority_update() is called with value 5 -> there should not be any priority frame sent
- the initial priority is 5 and priority_update() is called with value 6 -> there should be a priority frame sent
- the initial priority is 5 and priority_update() is called with value 6 then once again with 5 -> there should not be any  priority frame sent
- the initial priority is 5 and priority_update() is called with value 6 then once again with 7 -> there should be a  priority frame sent with value 7
- the initial priority is 5 and priority_update() is called with value 6 and the stream is reset -> there should not be any  priority frame sent
- the initial priority is 5 and priority_update() is called with value 5 and a changed increment value  -> there should be a priority frame sent

ddragana (commented):
I went through the whole change. The most important comments are about adding more tests.

neqo-http3/src/server.rs
mb: Yeah, forgot to delete it. It got replaced by `neqo-http3/tests/priority.rs`

neqo-http3/src/connection_server.rs
```diff
@@ -132,6 +132,7 @@ impl Http3ServerHandler {
                             Rc::clone(&self.base_handler.qpack_decoder),
                             Box::new(self.events.clone()),
                             None,
+                            Priority::default(),
```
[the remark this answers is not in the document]
mb: Opened #1216

mb:
Thanks! This test case is now also working. Prettier diff including both commits creating this test case: https://github.com/mozilla/neqo/compare/051290f70cc863d70a184ef5163198607c0dfd2d...af227d6d2d353fbef9b02ab3fc0e31ef3beb1b5e

neqo-http3/src/connection_client.rs
```diff
@@ -6521,4 +6612,32 @@ mod tests {
...
+        // Send a lot of data
+        client
+            .send_request_body(request_stream_id, &[0; 3000])
+            .unwrap();
+
+        // now change priority
+        assert!(client
+            .priority_update(request_stream_id, Priority::new(6, false))
+            .unwrap());
+
```
ddragana, line 6633: Can we extend this test? Currently it only tests that there is not loop.
Can we  do the following:
- after this line, let the client and server exchange packets and make sure that the sever does not receive a priority update (to check this it will be difficult because the server in this tests does not have http layer. You can read data from the control stream(I think the control stream will have stream_id 2) and check thaa it does not have any data or maybe only a SETTINGS frame).
- read stream data on he server side -> this will cause the server to queue a max_data update frame,
-  the client and server exchange packets again so that the client receives the max_data update frame,
- now server will receive a priority update (here as well you need to read the control stream and check that i contains a priority frame).

neqo-http3/src/priority.rs
```diff
@@ -0,0 +1,188 @@
...
+        assert!(p.maybe_encode_frame(4).is_none());
+    }
+
+    #[test]
+    fn multiple_priority_updates_send_update() {
+        let mut p = PriorityHandler::new(false, Priority::new(5, false));
+        assert!(p.maybe_update_priority(Priority::new(6, false)));
+        assert!(p.maybe_update_priority(Priority::new(7, false)));
+        // updating two times with a different -> there should be a priority frame sent
+        assert!(p.maybe_encode_frame(4).is_some());
```
ddragana, line 178: Can you check that this frame is for priority 7 not for priority 6.

neqo-http3/src/priority.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    #[test]
+    fn priority_updates_incremental() {
+        let mut p = PriorityHandler::new(false, Priority::new(5, false));
+        assert!(p.maybe_update_priority(Priority::new(5, true)));
+        // updating the incremental parameter -> there should be a priority frame sent
+        assert!(p.maybe_encode_frame(4).is_some());
```
ddragana, line 186: and also check here that the PRIORITY frame that it is correct (that i has "i" in it).

ddragana (commented):
This looks good. Just fix the tests and this is good to go. Yuo can land it withouth waiting form my review.

neqo-http3/src/connection_client.rs
mb: This doesn't work as I expected for me. The priority_update seems to get send (in the second packet I think):

doing the http request and priority update before the following packet exchange:

1. client -> server (partial http request, but priority_update doesn't fit)
2. server -> client (ack I guess)
3. client -> server (priority_update)

afterwards, reading the control_stream returns the priority update

I think this is because each stream has a separate flow control https://http3-explained.haxx.se/en/quic/quic-streams#flow-control
And I'm just restricting the flow control, client control stream has still free buffer to fill.

ddragana: There is per-stream flow control and per-complete quic session. Max_data should change the session's flow control.
Maybe something is freeing flow control credits, but I cannot say what without seeing the test....
when I think about it it is possible that enough flow control  credits is freed by reading SETTINGS frame and the request HEADER frame. And in addition the client sends MAX_DATA_BLOCKED frame that will cause server to free credits immediately. (my original thinking was wrong about this)

So please write the test as you have describe it above. It is fine as well.
In the step 2 (from your description) add a check that max_data frame has been sent:
```
let md_before = server.stats().frame_tx.max_data;
let out = server.process(..);
assert_eq!(server.stats().frame_tx.max_data, md_before + 1);
```

neqo-http3/src/connection_client.rs
```diff
@@ -6615,29 +6615,57 @@ mod tests {
...
     fn priority_update_during_full_buffer() {
-        // set a lower max_data on the server side to restrict the data the client can send
+        // set a lower MAX_DATA on the server side to restrict the data the client can send
         let (mut client, mut server) =
             connect_with_connection_parameters(ConnectionParameters::default().max_data(1200));
+
+        assert!(client.process(None, now()).dgram().is_none());
+
+        let mut buf = [0u8; 32];
+        server.conn.stream_recv(2, &mut buf).unwrap();
```
ddragana, line 6625: Add a comment here that you are reading SETTINGS frame.

neqo-http3/src/connection_client.rs
```diff
@@ -6615,29 +6615,57 @@ mod tests {
...
+        let out = client.process(out, now()).dgram();
+        let out = server.conn.process(out, now()).dgram();
+
+        // the server increased the max_data during the second read if that isn't the case
+        // in the future and therefore this asserts fails, the request data on stream 0 could be read
+        // to cause a max_update frame
+        assert_eq!(md_before + 1, server.conn.stats().frame_tx.max_data);
+
+        // make sure that the server didn't receive a priority_update yet
+        // exchange packets including settings on client control stream (stream_id 2)
```
ddragana, line 6658: This comment is not correct. You already read settings earlier.

neqo-http3/src/connection_client.rs
```diff
@@ -6615,29 +6615,57 @@ mod tests {
...
+        assert_eq!(server.conn.stream_recv(2, &mut buf), Ok((0, false)));
+
+        // the client now sends the priority update
+        let out = client.process(out, now()).dgram();
+        println!("3rd len: {}", out.as_ref().unwrap().len());
```
ddragana, line 6663: please remove println.

neqo-http3/src/connection_client.rs
```diff
@@ -6615,29 +6615,57 @@ mod tests {
...
+        let _out = server.conn.process(out, now()).dgram();
```
ddragana, line 6664: since we do not care about the output you can use process_input(..) instead of process()

neqo-http3/src/connection_client.rs
```diff
@@ -6615,29 +6615,57 @@ mod tests {
...
+
+        // make sure that the server didn't receive a priority_update yet
+        // exchange packets including settings on client control stream (stream_id 2)
+        let num_read = server.conn.stream_recv(2, &mut buf).unwrap();
```
ddragana, line 6668: The comment is not correct, please update it.

ddragana (dismissed):
Looks good.
Please land it

PR #1201, opened 2021-08-16
Datagrams

ddragana:
https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram

API:
- Event Datagram(Vec<u8>) - gives a new datagram to the application (TODO: we should limit number of datagrams that are queued in the Events)
- Events DatagramLost/DatagramAcked - this event is set when a datagram was acked or marked lost. It is not clear how much use the application will have from this signal. (TODO the datagrams need identifiers, otherwise these events have a limited use;)
- max_dgram_size() - returns the current max size of a datagram. This is calculated from the peer’s supplied limit and MTU and the exact current QUIC packet header size
- send_dgram() - stores a datagram into neqo-transport::Connection. The datagrams will be sent as soon as possible depending on the congestion control and pacing. If send_dgram() before the previous datagram has been sent, the previous datagram will be dismissed. (TODO: currently only one datagram is stored even if it is not filling a QUIC packet. We should store multiple datagrams if all can fit into a QUIC packet)

Other changes:
select_tx() returns a mutable reference to CryptoDxState. His function is renamed into select_tx_mut() and select_tx() returns an immutable reference to CryptoDxState. The new function is used to calculate QUIC packet header size.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -99,7 +99,6 @@ impl Http3Client {
         http3_parameters: &Http3Parameters,
         now: Instant,
     ) -> Res<Self> {
-        assert!(conn_params.get_max_datagram_frame_size() == 0);
         Ok(Self::new_with_conn(
             Connection::new_client(
                 server_name,
@@ -613,7 +612,7 @@ impl Http3Client {
                 }
                 ConnectionEvent::Datagram { .. }
                 | ConnectionEvent::DatagramAcked
-                | ConnectionEvent::DatagramLost => panic!("Datagrams are not enabled"),
+                | ConnectionEvent::DatagramLost => {}
             }
         }
         Ok(())
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -99,6 +99,7 @@ impl Http3Client {
         http3_parameters: &Http3Parameters,
         now: Instant,
     ) -> Res<Self> {
+        assert!(conn_params.get_max_datagram_frame_size() == 0);
         Ok(Self::new_with_conn(
             Connection::new_client(
                 server_name,
@@ -610,6 +611,9 @@ impl Http3Client {
                 ConnectionEvent::ResumptionToken(token) => {
                     self.create_resumption_token(&token);
                 }
+                ConnectionEvent::Datagram { .. }
+                | ConnectionEvent::DatagramAcked
+                | ConnectionEvent::DatagramLost => panic!("Datagrams are not enabled"),
             }
         }
         Ok(())
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -39,12 +39,13 @@ use crate::dump::*;
 use crate::events::{ConnectionEvent, ConnectionEvents};
 use crate::frame::{
     CloseError, Frame, FrameType, FRAME_TYPE_CONNECTION_CLOSE_APPLICATION,
-    FRAME_TYPE_CONNECTION_CLOSE_TRANSPORT, FRAME_TYPE_DATAGRAM, FRAME_TYPE_DATAGRAM_WITH_LEN,
+    FRAME_TYPE_CONNECTION_CLOSE_TRANSPORT,
 };
 use crate::packet::{
     DecryptedPacket, PacketBuilder, PacketNumber, PacketType, PublicPacket, QuicVersion,
 };
 use crate::path::{Path, PathRef, Paths};
+use crate::quic_datagrams::QuicDatagrams;
 use crate::recovery::{LossRecovery, RecoveryToken, SendProfile};
 use crate::rtt::GRANULARITY;
 pub use crate::send_stream::{RetransmissionPriority, TransmissionPriority};
@@ -246,10 +247,9 @@ pub struct Connection {
     /// In particular, this occurs when asynchronous certificate validation happens.
     saved_datagrams: SavedDatagrams,
 
+    /// This is responsible for the QuicDatagrams' handling:
     /// https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram
-    /// Holds one Datagram that will be sent as soon as possible. If a new datagram is
-    /// added before the previous has been sent, the old Datagram will be lost.
-    datagram: Option<Vec<u8>>,
+    quic_datagrams: QuicDatagrams,
 
     pub(crate) crypto: Crypto,
     pub(crate) acks: AckTracker,
@@ -368,6 +368,11 @@ impl Connection {
 
         let stats = StatsCell::default();
         let events = ConnectionEvents::default();
+        let quic_datagrams = QuicDatagrams::new(
+            conn_params.get_datagram_size(),
+            conn_params.get_queued_datagrams(),
+            events.clone(),
+        );
 
         let c = Self {
             role,
@@ -395,7 +400,7 @@ impl Connection {
             release_resumption_token_timer: None,
             conn_params,
             hrtime: hrtime::Time::get(Self::LOOSE_TIMER_RESOLUTION),
-            datagram: None,
+            quic_datagrams,
             #[cfg(test)]
             test_frame_writer: None,
         };
@@ -1806,6 +1811,9 @@ impl Connection {
             }
         }
 
+        // Check if there is a Datagram to be written
+        self.quic_datagrams.write_frames(builder, tokens, stats);
+
         self.streams
             .write_frames(TransmissionPriority::Critical, builder, tokens, stats);
         if builder.is_full() {
@@ -1923,11 +1931,6 @@ impl Connection {
         let primary = path.borrow().is_primary();
         let mut ack_eliciting = false;
 
-        // Check if there is a Datagram to be written
-        if primary && space == PacketNumberSpace::ApplicationData && !profile.ack_only(space) {
-            ack_eliciting |= self.maybe_write_datagram(builder, &mut tokens);
-        }
-
         if primary {
             let stats = &mut self.stats.borrow_mut().frame_tx;
             self.acks
@@ -1987,33 +1990,6 @@ impl Connection {
 
         stats.all += tokens.len();
         Ok((tokens, ack_eliciting, padded))
-    }
-
-    fn maybe_write_datagram(
-        &mut self,
-        builder: &mut PacketBuilder,
-        tokens: &mut Vec<RecoveryToken>,
-    ) -> bool {
-        if let Some(data) = &self.datagram {
-            let len = data.len();
-            if builder.remaining() >= len + 1 {
-                // + 1 for Frame type
-                let length_len = Encoder::varint_len(u64::try_from(len).unwrap());
-                if builder.remaining() > 1 + length_len + len {
-                    builder.encode_varint(FRAME_TYPE_DATAGRAM_WITH_LEN);
-                    builder.encode_vvec(&data);
-                } else {
-                    builder.encode_varint(FRAME_TYPE_DATAGRAM);
-                    builder.encode(&data);
-                }
-                debug_assert!(builder.len() <= builder.limit());
-                self.datagram = None;
-                self.stats.borrow_mut().frame_tx.datagram += 1;
-                tokens.push(RecoveryToken::Datagram);
-                return true;
-            }
-        }
-        false
     }
 
     /// Build a datagram, possibly from multiple packets (for different PN
@@ -2211,6 +2187,13 @@ impl Connection {
             self.idle_timeout
                 .set_peer_timeout(Duration::from_millis(peer_timeout));
         }
+
+        self.quic_datagrams.set_remote_datagram_size(
+            self.tps
+                .borrow()
+                .remote()
+                .get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE),
+        );
     }
 
     pub fn is_stream_id_allowed(&self, stream_id: StreamId) -> bool {
@@ -2535,16 +2518,8 @@ impl Connection {
                     .ack_freq(seqno, tolerance - 1, delay, ignore_order);
             }
             Frame::Datagram { data, .. } => {
-                let max_size = self
-                    .tps
-                    .borrow()
-                    .local
-                    .get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE);
-                if max_size < u64::try_from(data.len()).unwrap() {
-                    return Err(Error::ProtocolViolation);
-                }
-                self.events.datagram(data);
                 self.stats.borrow_mut().frame_rx.datagram += 1;
+                self.quic_datagrams.handle_datagram(data)?;
             }
             _ => unreachable!("All other frames are for streams"),
         };
@@ -2846,32 +2821,17 @@ impl Connection {
         self.streams.keep_alive(stream_id.into(), keep)
     }
 
-    fn max_dgram_size_configured(&self) -> Res<u64> {
-        let tph = self.tps.borrow();
-        if let Some(r) = &tph.remote {
-            Ok(r.get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE))
-        } else if let Some(r) = &tph.remote_0rtt {
-            Ok(r.get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE))
-        } else {
-            Err(Error::NotAvailable)
-        }
-    }
-
-    #[cfg(test)]
-    pub fn set_max_dgram_size(&mut self, v: u64) {
-        self.tps
-            .borrow_mut()
-            .remote
-            .as_mut()
-            .unwrap()
-            .set_integer(tparams::MAX_DATAGRAM_FRAME_SIZE, v);
+    pub fn remote_datagram_size(&self) -> u64 {
+        self.quic_datagrams.remote_datagram_size()
     }
 
     /// Returns the current max size of a datagram that can fit into a packet.
+    /// The value will change over time depending on the encoded size of the
+    /// packet number, ack frames, etc.
     /// # Error
     /// The function returns `NotAvailable` if datagrams are not enabled.
-    pub fn max_dgram_size(&self) -> Res<u64> {
-        let max_dgram_size = self.max_dgram_size_configured()?;
+    pub fn max_datagram_size(&self) -> Res<u64> {
+        let max_dgram_size = self.quic_datagrams.remote_datagram_size();
         if max_dgram_size == 0 {
             return Err(Error::NotAvailable);
         }
@@ -2885,7 +2845,7 @@ impl Connection {
         } else {
             return Err(Error::NotAvailable);
         };
-        let path = self.paths.select_path().ok_or(Error::NotAvailable)?;
+        let path = self.paths.primary_fallible().ok_or(Error::NotAvailable)?;
         let mtu = path.borrow().mtu();
         let encoder = Encoder::with_capacity(mtu);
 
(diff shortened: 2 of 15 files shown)


ddragana:
There are some TODOs that I can fix in a follow up PR.

neqo-transport/src/connection/mod.rs
```diff
@@ -2491,6 +2530,18 @@ impl Connection {
                 self.acks
                     .ack_freq(seqno, tolerance - 1, delay, ignore_order);
             }
+            Frame::Datagram { data, .. } => {
+                let max_size = self
+                    .tps
+                    .borrow()
+                    .local
+                    .get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE);
```
martinthomson, line 2538: We should be able to get this from the connection parameters instead.  That's immutable, so it won't be different.  And it doesn't need this whole borrow/lookup thing.

You should be able to copy the value to a new object if you move this processing elsewhere, but this check could just stay here.

neqo-transport/src/connection/mod.rs
```diff
@@ -2491,6 +2530,18 @@ impl Connection {
...
+                if max_size < u64::try_from(data.len()).unwrap() {
+                    return Err(Error::ProtocolViolation);
+                }
+                self.events.datagram(data);
```
martinthomson, line 2542: The copy that happens here is a little unfortunate.  It might be a good idea to mark this for future investigation.  I don't think that a copy is going to be so bad for us, but we might want to enable a more efficient handling process.  That's a project though, so this is fine.

neqo-transport/src/connection/mod.rs
```diff
@@ -2491,6 +2530,18 @@ impl Connection {
...
+                self.stats.borrow_mut().frame_rx.datagram += 1;
```
martinthomson, line 2543: I'd put this at the top, ahead of any checks, just like all the other frames.

neqo-transport/src/connection/mod.rs
```diff
@@ -2562,6 +2614,7 @@ impl Connection {
                     RecoveryToken::RetireConnectionId(seqno) => self.paths.acked_retire_cid(*seqno),
                     RecoveryToken::AckFrequency(rate) => self.paths.acked_ack_frequency(rate),
                     RecoveryToken::KeepAlive => self.idle_timeout.ack_keep_alive(),
+                    RecoveryToken::Datagram => self.events.datagram_acked(),
```
martinthomson, line 2617: Do we really care about that?

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+            self.loss_recovery
+                .largest_acknowledged_pn(PacketNumberSpace::ApplicationData),
+        );
+
+        let data_len_possible =
+            u64::try_from(mtu.saturating_sub(tx.expansion() + builder.len() + 1)).unwrap();
+        Ok(min(data_len_possible, max_dgram_size))
+    }
+
+    /// Returns if there was an unsend datagram that has been dismissed.
```
martinthomson, line 2909:
```suggestion
    /// Returns if there was an unsent datagram that has been dismissed.
```

Note that this could - and maybe should - change to returning any datagram we dropped.

neqo-transport/src/connection/mod.rs
```diff
@@ -1977,6 +1989,33 @@ impl Connection {
...
+                let length_len = Encoder::varint_len(u64::try_from(len).unwrap());
+                if builder.remaining() > 1 + length_len + len {
+                    builder.encode_varint(FRAME_TYPE_DATAGRAM_WITH_LEN);
+                    builder.encode_vvec(&data);
+                } else {
+                    builder.encode_varint(FRAME_TYPE_DATAGRAM);
+                    builder.encode(&data);
+                }
+                debug_assert!(builder.len() <= builder.limit());
+                self.datagram = None;
```
martinthomson, line 2010: When we can't send a datagram, we might consider generating an event.  That event shouldn't contain the datagram, but it could include the length of the datagram and a reason (mostly just expiration).  That might be something to track in a new issue rather than doing it as part of this though.

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+        let (cspace, tx) = if let Some(crypto) = self
+            .crypto
+            .states
+            .select_tx(PacketNumberSpace::ApplicationData)
+        {
+            crypto
+        } else {
+            return Err(Error::NotAvailable);
+        };
+        let path = self.paths.select_path().ok_or(Error::NotAvailable)?;
```
martinthomson, line 2884: IMPORTANT: You don't want to use select_path().  That's for deciding what to send next.  You want the primary path (use primary_fallible).

```suggestion
        let path = self.paths.primary_fallible().ok_or(Error::NotAvailable)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+            Ok(r.get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE))
+        } else if let Some(r) = &tph.remote_0rtt {
+            Ok(r.get_integer(tparams::MAX_DATAGRAM_FRAME_SIZE))
+        } else {
+            Err(Error::NotAvailable)
+        }
+    }
+
+    #[cfg(test)]
+    pub fn set_max_dgram_size(&mut self, v: u64) {
```
martinthomson, line 2857: Do you really need this?  Or can you just set the local transport parameter?  (Note to self: check the tests that use this.)

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+            .borrow_mut()
+            .remote
+            .as_mut()
+            .unwrap()
+            .set_integer(tparams::MAX_DATAGRAM_FRAME_SIZE, v);
+    }
+
+    /// Returns the current max size of a datagram that can fit into a packet.
+    /// # Error
+    /// The function returns `NotAvailable` if datagrams are not enabled.
```
martinthomson, line 2868: The problem with this function is that it is a prediction that is based on things that might change.  I have no problem with providing this function, but you should make it clear that this is an *estimate* of available capacity.  That goes with my comments about how DATAGRAM is prioritized for writing.

In practice, that means that sending DATAGRAM frames close to this limit is unwise.  We should skip any frames that don't fit.  We should avoid checking size in the send_datagram() function, absent a simple, hard maximum of 2^16-20 or thereabouts.

That means considering the "we dropped your datagram" event I already mentioned.  (That goes with a "your datagram was lost" event.)

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+        let data_len_possible =
+            u64::try_from(mtu.saturating_sub(tx.expansion() + builder.len() + 1)).unwrap();
+        Ok(min(data_len_possible, max_dgram_size))
+    }
+
+    /// Returns if there was an unsend datagram that has been dismissed.
+    /// # Error
+    /// The function returns `TooMuchData` if the supply buffer cannot fit into a packet.
+    /// The function returns `NotAvailable` if datagrams are not enabled.
+    pub fn send_dgram(&mut self, buf: &[u8]) -> Res<bool> {
```
martinthomson, line 2913: No point in a contraction here.

```suggestion
    pub fn send_datagram(&mut self, buf: &[u8]) -> Res<bool> {
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
...
+        let data_limit = self.max_dgram_size()?;
```
martinthomson, line 2914: Please don't run all that code here.  This is likely to be a hot path; instead expect users of the API to make this call and do their own checking.  Then add checks on sending, where we already have all the necessary context.

neqo-transport/src/connection/params.rs
```diff
@@ -81,6 +82,7 @@ impl Default for ConnectionParameters {
             ack_ratio: DEFAULT_ACK_RATIO,
             idle_timeout: DEFAULT_IDLE_TIMEOUT,
             preferred_address: PreferredAddressConfig::Default,
+            max_datagram_frame_size: 0,
```
martinthomson, line 85: I think that we want the default for this to be enabled.  I realize that you don't want this in the HTTP/3 part just yet (I haven't worked out why yet), but I think - long term - we will want to negotiate this feature for those connections too.

neqo-transport/src/connection/mod.rs
```diff
@@ -2788,6 +2841,84 @@ impl Connection {
     pub fn stream_keep_alive(&mut self, stream_id: u64, keep: bool) -> Res<()> {
         self.streams.keep_alive(stream_id.into(), keep)
     }
+
+    fn max_dgram_size_configured(&self) -> Res<u64> {
```
martinthomson, line 2845:
```suggestion
    fn remote_datagram_size(&self) -> Res<u64> {
```

neqo-transport/src/connection/mod.rs
martinthomson: Adding an additional parameter for the expiration time of the datagram would be useful, I think.  We might add a new function (`send_datagram_before()`) for that though.

I'd be happy to have that tracked in an issue.

neqo-transport/src/connection/params.rs
```diff
@@ -209,6 +211,15 @@ impl ConnectionParameters {
         self.idle_timeout
     }
 
+    pub fn get_max_datagram_frame_size(&self) -> u64 {
+        self.max_datagram_frame_size
+    }
+
+    pub fn max_datagram_frame_size(mut self, v: u64) -> Self {
```
martinthomson, line 218:
```suggestion
    pub fn datagram_size(mut self, v: u64) -> Self {
```

Shorter names = good.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+#[test]
+fn mtu_limit() {
+    let (client, server) = create_and_connect_with_dgrams();
+
+    assert_eq!(client.max_dgram_size(), Ok(DATAGRAM_LEN_MTU));
+    assert_eq!(server.max_dgram_size(), Ok(DATAGRAM_LEN_MTU));
+}
+
+#[test]
+fn limit_data_size() {
```
martinthomson, line 113: If you take my suggestion about limiting size, then you will need to test that the datagrams are sent rather than just attempting to send.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+    let out = client.process_output(now()).dgram().unwrap();
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent + 1);
+
+    server.process_input(out, now());
+    let datagram =
+        |e| matches!(e, ConnectionEvent::Datagram(data) if data == DATA_SMALLER_THAN_MTU);
+    assert!(server.events().any(datagram));
+}
+
+fn create_and_connect_with_dgrams() -> (Connection, Connection) {
```
martinthomson, line 95:
```suggestion
fn connect_datagram() -> (Connection, Connection) {
```

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    let dgram_sent = client.stats().frame_tx.datagram;
+    assert_eq!(client.send_dgram(DATA_SMALLER_THAN_MTU), Ok(false));
+    let out = client.process_output(now()).dgram();
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent + 1);
+
+    let dgram_received = server.stats().frame_rx.datagram;
+    server.process_input(out.unwrap(), now());
+    assert_eq!(server.stats().frame_rx.datagram, dgram_received + 1);
+    let now = now() + Duration::from_millis(1);
```
martinthomson, line 139: Don't put fixed timeouts in.  Is this the ack delay?

I think that this test might be fragile.  It relies on the server needing to send an ACK for other reasons.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+    let _out = client.process_output(now()).dgram(); // This packet will be lost.
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent + 1);
+
+    // Wait for PTO
+    let now = now() + AT_LEAST_PTO;
+    let dgram_sent = client.stats().frame_tx.datagram;
+    let out = client.process_output(now).dgram();
+    assert!(out.is_some()); //PING probing
+                            // Datagram is not sent again.
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent);
```
martinthomson, line 169: You should just test for PING too.  That is safer than a comment.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent + 1);
+
+    // Call process_output again should not send any new Datagram.
+    let dgram_sent = client.stats().frame_tx.datagram;
+    assert!(client.process_output(now()).dgram().is_none());
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent);
+}
+
+#[test]
+fn dgram_no_allowed() {
```
martinthomson, line 191: This is where you tweak the remote transport parameters.

I would use the test_frame_writer function for this instead.  You can write the DATAGRAM frame yourself.  It's trivial to do.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+    client.process_input(out, now());
+
+    assert_error(
+        &client,
+        &ConnectionError::Transport(Error::ProtocolViolation),
+    );
+}
+
+#[test]
+fn dgram_too_big() {
```
martinthomson, line 209: Same applies as the previous one, except that this one depends on a larger frame.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -0,0 +1,243 @@
...
+    client.process_input(out, now());
+
+    assert_error(
+        &client,
+        &ConnectionError::Transport(Error::ProtocolViolation),
+    );
+}
+
+#[test]
+fn overwrite_datagram() {
```
martinthomson, line 230: Not sure what this test is adding.

martinthomson (commented):
There are a lot of open issues here.  Some of which you already covered, which is good.  I have a few extra to add.

neqo-http3/src/connection_client.rs
```diff
@@ -99,6 +99,7 @@ impl Http3Client {
         http3_parameters: &Http3Parameters,
         now: Instant,
     ) -> Res<Self> {
+        assert!(conn_params.get_max_datagram_frame_size() == 0);
```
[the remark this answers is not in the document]
ddragana: This is only temporarily. I have added it because if someone accidentally turn it on the code may panic if it receive a Datagram event later. This is just to make code panic non start not later.
I will add a note.

neqo-transport/src/connection/mod.rs
```diff
@@ -246,6 +246,11 @@ pub struct Connection {
     /// In particular, this occurs when asynchronous certificate validation happens.
     saved_datagrams: SavedDatagrams,
 
+    /// https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram
+    /// Holds one Datagram that will be sent as soon as possible. If a new datagram is
+    /// added before the previous has been sent, the old Datagram will be lost.
+    datagram: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
ddragana: Yes, I was planning to do exactly that.

neqo-transport/src/connection/mod.rs
```diff
@@ -1915,6 +1921,12 @@ impl Connection {
     ) -> Res<(Vec<RecoveryToken>, bool, bool)> {
         let mut tokens = Vec::new();
         let primary = path.borrow().is_primary();
+        let mut ack_eliciting = false;
+
+        // Check if there is a Datagram to be written
+        if primary && space == PacketNumberSpace::ApplicationData && !profile.ack_only(space) {
+            ack_eliciting |= self.maybe_write_datagram(builder, &mut tokens);
+        }
```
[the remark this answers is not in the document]
ddragana: I for got to add the note why I did it in this way. I decided to send datagram first because ACKs can be sent even if there is a cc limit.

Anyway, I will reorder them.

neqo-transport/src/connection/mod.rs
ddragana: I have already spent some time thinking about this. I do not have a solution at the moment. Especially with rust-c++ glue code this will need another copy. I will open an issue, maybe we will have a better idea.

neqo-transport/src/connection/mod.rs
ddragana: Probably not, I have added for now, just to have it when we start implemented the protocol on top of it.

neqo-transport/src/connection/tests/datagram.rs
ddragana: I was searching for it and I could not find it. I knew we have something :)

neqo-transport/src/connection/tests/datagram.rs
ddragana: Test that a previous datagram is dropped if a new one is added and the old one is not sent yet.

martinthomson: Thanks.  I completely missed that.  It's pretty obvious when I look again.

neqo-http3/src/connection_client.rs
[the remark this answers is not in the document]
ddragana: I have change my mind: I will remove this assertion and just ignore Datagram* events.

neqo-transport/src/connection/mod.rs
ddragana: Open issue #1203

neqo-transport/src/connection/params.rs
ddragana: I will leave the feature to be disabled by default for now. 
I remove the assertion in HTTP/3 code.

neqo-transport/src/connection/tests/datagram.rs
ddragana: The server needs to send an ack for the received datagram. I will use AT_LEAST_PTO time here.

neqo-transport/src/events.rs
```diff
@@ -152,6 +158,44 @@ impl ConnectionEvents {
         self.remove(|evt| matches!(evt, ConnectionEvent::RecvStreamReadable { stream_id: x } if *x == stream_id.as_u64()));
     }
 
+    fn check_datagram_queued(&self, max_queued_datagrams: usize) {
```
KershawChang, line 161: Since what this function does is not just `checking`, could you add some comments to describe this a bit?

neqo-transport/src/events.rs
```diff
@@ -152,6 +158,44 @@ impl ConnectionEvents {
...
+                remove = Some(d.0);
+            }
+        }
+        if let Some(r) = remove {
+            q.remove(r);
+            q.push_back(ConnectionEvent::DatagramLost);
+        }
+    }
+
+    pub fn datagram(&self, max_queued_datagrams: usize, data: &[u8]) {
```
KershawChang, line 186: Could you rename this function for the sake of readability? Probably `add_datagram` or something else.

neqo-transport/src/events.rs
```diff
@@ -62,6 +62,12 @@ pub enum ConnectionEvent {
     /// Any data written to streams needs to be written again.
     ZeroRttRejected,
     ResumptionToken(ResumptionToken),
+    Datagram(Vec<u8>),
+    // TODO: datagrams probably need some identifiers. The packet number they
+    // are sent in would be enough. In that case we need an even DaagramSent
```
KershawChang, line 67: nit: event

KershawChang (approved):

neqo-transport/src/connection/mod.rs
```diff
@@ -2792,6 +2820,69 @@ impl Connection {
...
+        );
+        let _ = Self::add_packet_number(
+            &mut builder,
+            tx,
+            self.loss_recovery
+                .largest_acknowledged_pn(PacketNumberSpace::ApplicationData),
+        );
+
+        let data_len_possible =
+            u64::try_from(mtu.saturating_sub(tx.expansion() + builder.len() + 1)).unwrap();
```
mxinden, line 2869: @martinthomson @larseggert does either of you know what the `+ 1` represents here?

The whole function (i.e. `max_datagram_size`) along with the `+ 1` was introduced in this pull request and has been untouched since. See latest `main`:

https://github.com/mozilla/neqo/blob/220ef85f624b361d87efaf662bf8758c4fca6fcd/neqo-transport/src/connection/mod.rs#L3848-L3850

Context: We have a [bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1992919) where `neqo-transport` packs 1 byte too much into a GSO segment. While here we take both `tx.expansion()` and the unknown `+ 1` into account, we only take `tx.expansion()` into account when actually building a packet.

https://github.com/mozilla/neqo/blob/220ef85f624b361d87efaf662bf8758c4fca6fcd/neqo-transport/src/connection/mod.rs#L2671-L2678

My best bet is that `+ 1` is added for the crypto expansion varint length:

https://github.com/mozilla/neqo/blob/220ef85f624b361d87efaf662bf8758c4fca6fcd/neqo-transport/src/packet/mod.rs#L462-L464

mxinden: Will be resolved in https://github.com/mozilla/neqo/pull/1201/files#r2414614536.

PR #1202, opened 2021-08-17
Move decoding of PushId into NewStreamTypeReader

ddragana:
This is extracted from my work on WebTransport and simplifies PushStream implementation. PushStream is not any more responsible for reading PushId and therefore it is transformed into a very simple wrapper around RecvMessage.

This PR also changes the behavior regarding closing a stream too early, e.g. if a stream is closed before the stream id is read it will return HttpStreamCreation.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -13,9 +13,9 @@ use crate::qpack_decoder_receiver::DecoderRecvStream;
 use crate::qpack_encoder_receiver::EncoderRecvStream;
 use crate::send_message::SendMessage;
 use crate::settings::{HSetting, HSettingType, HSettings, HttpZeroRttChecker};
-use crate::stream_type_reader::NewStreamTypeReader;
+use crate::stream_type_reader::NewStreamHeadReader;
 use crate::{Http3StreamType, NewStreamType, Priority, ReceiveOutput, RecvStream, ResetType};
-use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn};
+use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn, Role};
 use neqo_qpack::decoder::QPackDecoder;
 use neqo_qpack::encoder::QPackEncoder;
 use neqo_qpack::QpackSettings;
@@ -203,11 +203,11 @@ impl Http3Connection {
         }
     }
 
-    pub fn handle_new_unidi_stream(&mut self, stream_id: u64, push_stream_allowed: bool) {
+    pub fn handle_new_unidi_stream(&mut self, stream_id: u64, role: Role) {
         qtrace!([self], "A new stream: {}.", stream_id);
         self.recv_streams.insert(
             stream_id,
-            Box::new(NewStreamTypeReader::new(stream_id, push_stream_allowed)),
+            Box::new(NewStreamHeadReader::new(stream_id, role)),
         );
     }
 
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -14,7 +14,7 @@ use crate::qpack_encoder_receiver::EncoderRecvStream;
 use crate::send_message::SendMessage;
 use crate::settings::{HSetting, HSettingType, HSettings, HttpZeroRttChecker};
 use crate::stream_type_reader::NewStreamTypeReader;
-use crate::{Http3StreamType, Priority, ReceiveOutput, RecvStream, ResetType};
+use crate::{Http3StreamType, NewStreamType, Priority, ReceiveOutput, RecvStream, ResetType};
 use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn};
 use neqo_qpack::decoder::QPackDecoder;
 use neqo_qpack::encoder::QPackEncoder;
@@ -203,10 +203,12 @@ impl Http3Connection {
         }
     }
 
-    pub fn handle_new_unidi_stream(&mut self, stream_id: u64) {
+    pub fn handle_new_unidi_stream(&mut self, stream_id: u64, push_stream_allowed: bool) {
         qtrace!([self], "A new stream: {}.", stream_id);
-        self.recv_streams
-            .insert(stream_id, Box::new(NewStreamTypeReader::new(stream_id)));
+        self.recv_streams.insert(
+            stream_id,
+            Box::new(NewStreamTypeReader::new(stream_id, push_stream_allowed)),
+        );
     }
 
     #[allow(unknown_lints, renamed_and_removed_lints, clippy::unknown_clippy_lints)] // Until we require rust 1.51.
@@ -286,6 +288,7 @@ impl Http3Connection {
                 }
                 Ok(ReceiveOutput::ControlFrames(rest))
             }
+            ReceiveOutput::NewStream(NewStreamType::Push(_)) => Ok(output),
             ReceiveOutput::NewStream(_) => {
                 unreachable!("NewStream should have been handled already")
             }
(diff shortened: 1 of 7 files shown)


ddragana:
This is extracted from my work on WebTransport, because it can be reviewed  and landed separately.

neqo-http3/src/connection_client.rs
```diff
@@ -628,25 +630,43 @@ impl Http3Client {
...
-                    Rc::clone(&self.push_handler),
-                    Rc::clone(&self.base_handler.qpack_decoder),
-                    self.events.clone(),
-                )),
+    fn handle_new_push_stream(&mut self, stream_id: u64, push_id: u64) -> Res<()> {
+        if !self.push_handler.borrow().can_receive_push() {
+            return Err(Error::HttpId);
+        }
+
+        // Add a new pish stream to `PushController`. `add_new_push_stream` may return an error
```
martinthomson, line 638:
```suggestion
        // Add a new push stream to `PushController`. `add_new_push_stream` may return an error
```

neqo-http3/src/connection_server.rs
```diff
@@ -136,7 +136,7 @@ impl Http3ServerHandler {
                     ),
                     StreamType::UniDi => self
                         .base_handler
-                        .handle_new_unidi_stream(stream_id.as_u64()),
+                        .handle_new_unidi_stream(stream_id.as_u64(), false),
```
martinthomson, line 139: Would it be better to pass `Role` here instead of a boolean?  The effect in terms of code is the same, but the code might be more readable as a result.

neqo-http3/src/lib.rs
```diff
@@ -290,14 +290,28 @@ pub enum Http3StreamType {
...
+    Encoder,
+    Http,
+    Push(u64),
+    Unknown,
+}
+
+impl NewStreamType {
+    fn decode_stream_type(
+        stream_type: u64,
+        push_stream_allowed: bool,
```
martinthomson, line 306: See above regarding Role.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
-                ref mut reader,
-                stream_id,
-            } = self
-            {
+    fn read(&mut self, conn: &mut Connection) -> Res<(Option<u64>, bool)> {
+        match self {
+            NewStreamTypeReader::ReadType {
+                reader, stream_id, ..
+            }
+            | NewStreamTypeReader::ReadSessionId { reader, stream_id } => loop {
```
martinthomson, line 44: I think that you want to use `if let N.... = self {` here.  I completely missed the loop bit on my first read.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
+        loop {
+            let (output, fin) = self.read(conn)?;
+            if output.is_none() {
+                if fin {
+                    *self = NewStreamTypeReader::Done;
+                    break Err(Error::HttpStreamCreation);
+                }
+                break Ok(None);
+            }
+            let output = output.unwrap();
```
martinthomson, line 71: I think that we can do:

```rust
let output = if let Some(o) = output {
    o
} else {
    if fin { ... }
    break Ok(None);
};
```
Also, if breaking out of a loop produces a return value, I would prefer that you use "return" rather than "break" as it makes it easier to understand that this is the end.  With "break" you need to check that the loop is the only statement in the function, "return" is unambiguous.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
+                    push_stream_allowed,
+                    stream_id,
+                    ..
+                } => {
+                    let res = NewStreamType::decode_stream_type(output, *push_stream_allowed);
+                    if res.is_err() {
+                        *self = NewStreamTypeReader::Done;
+                        return res;
+                    }
+                    let res = res.unwrap();
```
martinthomson, line 84: The same pattern can be used here to avoid using `unwrap()`.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
+                        *self = NewStreamTypeReader::Done;
+                        return self.map_stream_fin(res);
+                    }
+                    qtrace!("Decoded stream type {:?}", res);
+                    if res.is_some() {
                         *self = NewStreamTypeReader::Done;
-                        return None;
+                        break Ok(res);
+                    }
+                    // This is a WebTransportStream stream and it needs more data to be decoded.
```
martinthomson, line 94: This isn't true yet.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
-use neqo_common::{Decoder, IncrementalDecoderUint};
+use crate::{
+    AppError, Error, Http3StreamType, HttpRecvStream, NewStreamType, ReceiveOutput, RecvStream,
+    Res, ResetType,
+};
+use neqo_common::{qtrace, Decoder, IncrementalDecoderUint};
 use neqo_transport::Connection;
 
 #[derive(Debug)]
 pub enum NewStreamTypeReader {
```
martinthomson, line 17: You should write some documentation for this.  I'm not seeing how you distinguish between (unidirectional) streams that just have a type and streams that have a varint type plus a Push ID or WT Session ID.

This is also more than a type reader.  It reads a stream header.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
+                        *self = NewStreamTypeReader::Done;
+                        return res;
+                    }
+                    let res = res.unwrap();
+                    if fin {
+                        *self = NewStreamTypeReader::Done;
+                        return self.map_stream_fin(res);
+                    }
+                    qtrace!("Decoded stream type {:?}", res);
+                    if res.is_some() {
```
martinthomson, line 90: It's not clear from context here what `res` is.  Naming this variable better (and the decode_stream_type function) would make this easier to understand.  The function returns two things: a concrete stream type in the case where the stream header is complete, or `None` if additional information is required from the header to produce a final type.  You could rename it `final_stream_type`.

Some commentary might help as well.  That is, "if the stream type isn't final, we need to read an extra identifier from the stream to know what the stream is for".

neqo-http3/src/stream_type_reader.rs
```diff
@@ -6,57 +6,120 @@
...
     }
+
+    fn map_stream_fin(&self, decoded: Option<NewStreamType>) -> Res<Option<NewStreamType>> {
+        match decoded {
+            Some(NewStreamType::Control)
+            | Some(NewStreamType::Encoder)
+            | Some(NewStreamType::Decoder) => Err(Error::HttpClosedCriticalStream),
+            None => Err(Error::HttpStreamCreation),
+            Some(NewStreamType::Unknown) => Ok(decoded),
+            _ => unreachable!("PushStream is mapped to None at this stage."),
```
martinthomson, line 120: I would enumerate the options here.  Avoid `_` where possible.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -139,78 +208,172 @@ mod tests {
...
+        t.decode(
+            &[QPACK_UNI_STREAM_TYPE_DECODER],
+            false,
+            Ok(ReceiveOutput::NoOutput),
+            true,
+        );
+    }
+
+    #[test]
+    fn stream_fin() {
```
martinthomson, line 324: No reason not to make this 7 different test functions.  I would only merge the two push tests.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -139,78 +208,172 @@ mod tests {
...
+            &[0x3fff_ffff_ffff_ffff],
+            true,
+            Ok(ReceiveOutput::NewStream(NewStreamType::Unknown)),
+            true,
+        );
+
+        let mut t = Test::new(true);
+        // stream_id 0x3fff_ffff_ffff_ffff is encoded into [0xff, 8].
+        // For this test the last byte is no delivered before a fin.
+        // This should cause aan error.
```
martinthomson, line 376:
```suggestion
        // A stream ID of 0x3fff_ffff_ffff_ffff is encoded into [0xff; 8].
        // For this test the stream type is truncated.
        // This should cause an error.
```

martinthomson (commented):
Good changes.  I've made a few suggestions about naming and other things that should keep things comprehensible.  I struggled a little understanding it and I think that some of that is down to naming.

neqo-http3/src/stream_type_reader.rs
ddragana: I did no remove all signs of WebTransport :)

neqo-http3/src/connection.rs
```diff
@@ -413,48 +416,57 @@ impl Http3Connection {
...
+                qinfo!(
+                    [self],
+                    "A new push stream {} push_id:{}.",
+                    stream_id,
+                    push_id
+                );
             }
-            Http3StreamType::Decoder => {
+            NewStreamType::Decoder => {
                 qinfo!([self], "A new remote qpack encoder stream {}", stream_id);
```
KershawChang, line 438: nit: should be "decoder" stream

neqo-http3/src/connection.rs
```diff
@@ -413,48 +416,57 @@ impl Http3Connection {
...
-                )))
+                    Box::new(DecoderRecvStream::new(
+                        stream_id,
+                        Rc::clone(&self.qpack_decoder),
+                    )),
+                );
             }
-            Http3StreamType::Encoder => {
+            NewStreamType::Encoder => {
                 qinfo!([self], "A new remote qpack decoder stream {}", stream_id);
```
KershawChang, line 449: "encoder" stream?

neqo-http3/src/connection.rs
KershawChang: never mind. I think this is correct.

KershawChang (approved):

PR #1213, opened 2021-09-01
Avoid panicing when substracting Instants

### LINKED ISSUE #1209: Bug 1727148
Bug [1727148](https://bugzilla.mozilla.org/show_bug.cgi?id=1727148)

This may be:
https://searchfox.org/mozilla-central/rev/4cca5d2f257c6f1bcef50a0debcbd66524add703/third_party/rust/neqo-transport/src/recovery.rs#651

ddragana:
Fixes #1209

### DIFF
--- a/neqo-transport/src/recovery.rs
+++ b/neqo-transport/src/recovery.rs
@@ -640,8 +640,9 @@ impl LossRecovery {
         ack_delay: Duration,
     ) {
         let confirmed = self.confirmed_time.map_or(false, |t| t < send_time);
-        let sample = now - send_time;
-        rtt.update(&mut self.qlog, sample, ack_delay, confirmed, now);
+        if let Some(sample) = now.checked_duration_since(send_time) {
+            rtt.update(&mut self.qlog, sample, ack_delay, confirmed, now);
+        }
     }
 
     /// Returns (acked packets, lost packets)


ddragana:
I also did an audit of the code and I could not fined any other case of substracting time.

neqo-transport/src/recovery.rs
```diff
@@ -640,8 +640,9 @@ impl LossRecovery {
         ack_delay: Duration,
     ) {
         let confirmed = self.confirmed_time.map_or(false, |t| t < send_time);
-        let sample = now - send_time;
-        rtt.update(&mut self.qlog, sample, ack_delay, confirmed, now);
+        if now >= send_time {
+            rtt.update(&mut self.qlog, now - send_time, ack_delay, confirmed, now);
```
martinthomson, line 644: I think that this is more idiomatic.

```suggestion
        if Some(sample) = now.checked_sub(send_time) {
            rtt.update(&mut self.qlog, sample, ack_delay, confirmed, now);
```

martinthomson (commented):
LGTM

neqo-transport/src/recovery.rs
ddragana: I have tried that, but it does not work checked_sub needs Duration; the output of checked_sub is Instant, but update() needs Duration.

martinthomson: My mistake, I should have checked types.  What I meant was:

```suggestion
        if Some(sample) = now.checked_duration_since(send_time) {
            rtt.update(&mut self.qlog, sample, ack_delay, confirmed, now);
```

That is in rust 1.39, so it available in our minimum version.

Issue #1156, opened 2021-05-27
Build failed in macOS


Rouzip:
I have followed the steps in [this](https://github.com/mozilla/neqo/issues/208#issuecomment-535158410). But I cannot build neqo successfully. And I also try to use extern nss, but still get this:
```bash
error[E0432]: unresolved imports `crate::p11::HpkeAeadId`, `crate::p11::HpkeKdfId`, `crate::p11::HpkeKemId`
  --> neqo-crypto/src/ech.rs:19:22
   |
19 | pub use crate::p11::{HpkeAeadId as AeadId, HpkeKdfId as KdfId, HpkeKemId as KemId};
   |                      ^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^ no `HpkeKemId` in `p11`
   |                      |                     |
   |                      |                     no `HpkeKdfId` in `p11`
   |                      no `HpkeAeadId` in `p11`

error[E0432]: unresolved import `crate::ssl::HpkeSymmetricSuite`
  --> neqo-crypto/src/ech.rs:20:9
   |
20 | pub use crate::ssl::HpkeSymmetricSuite as SymmetricSuite;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `HpkeSymmetricSuite` in `ssl`

error[E0432]: unresolved import `crate::err::ssl::SSL_ERROR_ECH_RETRY_WITH_ECH`
 --> neqo-crypto/src/ech.rs:7:18
  |
7 | use crate::err::{ssl::SSL_ERROR_ECH_RETRY_WITH_ECH, Error, Res};
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `SSL_ERROR_ECH_RETRY_WITH_ECH` in `err::codes::SSLErrorCodes`

error[E0425]: cannot find function, tuple struct or tuple variant `PK11_HPKE_Serialize` in this scope
  --> neqo-crypto/src/p11.rs:86:13
   |
86 |             PK11_HPKE_Serialize(
   |             ^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `ssl_preinfo_ech` in module `ssl`
   --> neqo-crypto/src/agent.rs:161:39
    |
[...]

PR #1217, opened 2021-09-07
Fixed Build NSS command on Linux build document

dasixi34:
When I followed Linux build document, I faced same problem as #1156.
The cause of this problem was `-u c7a1c91cd9be` which means to use `NSS_3_53_BETA1`.


martinthomson:
Thanks

PR #1219, opened 2021-09-09
PushStream does not need to hold a PriorityHandler

ddragana:
PR #1202 made PushStream a simpler wrapper around RecvMessage and that made possible to remove PropertyHandler rom PushStream


KershawChang (approved):
nit in summary: replace `PropertyHandler` to `PriorityHandler`

Issue #1207, opened 2021-08-17
When we can't send a datagram, we might consider generating an event


ddragana:
When we can't send a datagram, we might consider generating an event.  That event shouldn't contain the datagram, but it could include the length of the datagram and a reason (mostly just expiration).  That might be something to track in a new issue rather than doing it as part of this though.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1201#discussion_r689964376_

ddragana:
The last change in #1201 generates a DatagramLost event without any info. To fix his issue we should add information o DatagramLost or maybe add a separate event DatagramNotSent.

ddragana:
Fixed by #1220

PR #1220, opened 2021-09-09
Queue multiple outgoing Quic datagrams

### LINKED ISSUE #1203: Datagram follow up: queue multiple datagrams for transmission
Currently we only store one datagram, we should store multiple datagrams.

ddragana:
Additional changes:
- The IncomingQuicDatagramDropped event is queued when a queued QuicDatagram event gets dropped due to the number of queued datagrams exceeding the limit. The limit is controlled by a connection parameter.
- The OutgoingQuicDatagramOutcome event is queued when the outcome of the outgoing of a QuicDatagram is known. It can be: lost, asked, dropped because the outgoing queue of datagrams is full(The limit is controlled by a connection parameter) or it can be dropped because it does not fit into a quic packet.
- Change name of some parameters and events from datagram/Datagram to quic_datagram/QuicDatagram for consistency. This does not apply to the QuicDatagram structure because the structure is about QuicDatagrams.
- Fix a bug: multiple Datagrams containing the same data were not queued, just silently dropped because insert() function was used that checks for duplicate events. In this case we want to queue duplicate events.

Fixes #1203 and #1207

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -36,7 +36,7 @@ use crate::cid::{
 
 use crate::crypto::{Crypto, CryptoDxState, CryptoSpace};
 use crate::dump::*;
-use crate::events::{ConnectionEvent, ConnectionEvents};
+use crate::events::{ConnectionEvent, ConnectionEvents, OutgoingQuicDatagramOutcome};
 use crate::frame::{
     CloseError, Frame, FrameType, FRAME_TYPE_CONNECTION_CLOSE_APPLICATION,
     FRAME_TYPE_CONNECTION_CLOSE_TRANSPORT,
@@ -369,8 +369,9 @@ impl Connection {
         let stats = StatsCell::default();
         let events = ConnectionEvents::default();
         let quic_datagrams = QuicDatagrams::new(
-            conn_params.get_datagram_size(),
-            conn_params.get_queued_datagrams(),
+            conn_params.get_quic_datagram_size(),
+            conn_params.get_queued_outgoing_quic_datagrams(),
+            conn_params.get_queued_incoming_quic_datagrams(),
             events.clone(),
         );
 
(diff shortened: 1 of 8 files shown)


martinthomson (commented):
This seems like a useful addition, though I'm not sure about using the length as an identifier.  That's a halfway that is sometimes useful for identifying what happened, but it could also end up being useless.  I think we either want to reduce information (and simplify) or attach an identifier to datagrams so that we can do more precise tracking.  I might prefer the latter, but maybe the right answer is to do that in a follow-up change.

The other thing that I don't like much is the addition of "quic" to all of the names.

ddragana:
In the first part of the implemetation of Datagrams that has already landed in some cases I use quic_datagram/QuicDatagram because "Datagram" was also used for udp packet(or maybe quic packet, e.g. saved_datagrams) and that got a bit confusing.
In this change I wanted to make it consistent,  butt that makes name longer as well. I will change it back to shorter names.

neqo-transport/src/connection/mod.rs
```diff
@@ -2880,8 +2885,8 @@ impl Connection {
     /// to check the estimated max datagram size and to use smaller datagrams.
     /// `max_datagram_size` is just a current estimate and will change over
     /// time depending on the encoded size of the packet number, ack frames, etc.
-    pub fn add_datagram(&mut self, buf: &[u8]) -> Res<bool> {
-        self.quic_datagrams.add_datagram(buf)
+    pub fn add_datagram(&mut self, buf: &[u8], id: Option<u64>) -> Res<()> {
```
KershawChang, line 2888: The comment of this function needs to be updated.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -10,37 +10,61 @@ use crate::frame::{FRAME_TYPE_DATAGRAM, FRAME_TYPE_DATAGRAM_WITH_LEN};
...
 use std::cmp::min;
+use std::collections::VecDeque;
 use std::convert::TryFrom;
+use std::ops::Deref;
 
 pub const MAX_QUIC_DATAGRAM: u64 = 65535;
 
+struct QuicDatagram {
+    data: Vec<u8>,
+    id: Option<u64>,
```
KershawChang, line 24: Why this `id` is optional? I think this should be always presented.

ddragana: We are not sure how useful it is and if it will be used, therefore it is optional.

KershawChang (approved):
lgtm

PR #1221, opened 2021-09-09
Change rust version

ddragana:


martinthomson:
Firefox 91 is the latest ESR and [the minimum rust version there is 1.51.0](https://searchfox.org/mozilla-central/source/docs/writing-rust-code/update-policy.md).  That should be fine for our purposes.

(There are a few things we can clean up now that we require this version, but that can come in time.)

Issue #1218, opened 2021-09-09
Do not use events for Datagrams


ddragana:
Currently Datagrams are delivered as events, I think we should have a separate API for datagrams so that they can be handled independent of other connection events.

mxinden:
As long as we are statically allocating each datagram in a `Vec<u8>` anyways, what is the benefit of serving them outside of our event system? We need a notification (i.e. event) of a datagram being available anyways. Why not just bundle the event with the datagram itself?

PR #1222, opened 2021-09-10
Dgrams stats

ddragana:


ddragana:
This is based on #1220 and only the last push is new.

neqo-transport/src/stats.rs
```diff
@@ -127,6 +127,17 @@ pub struct Stats {
...
+    /// The number of incoming datagrams dropped due to reaching the limit
+    /// of the incoming queue.
+    pub incoming_datagram_dropped: usize,
+    /// The number of datagrams declared lost.
+    pub outgoing_datagram_lost: usize,
+    /// The number of datagrams dropped due to being too large.
+    pub outgoing_datagram_dropped_too_big: usize,
+    /// The number of datagrams dropped due to reaching the limit of the
+    /// outgoing queue.
+    pub outgoing_datagram_dropped_queue_full: usize,
```
martinthomson, line 140: It seems like we have a datagram_tx struct needed here to capture the different outcomes.  We won't avoid a train wreck of names at the usage site, but we might be able to limit some of the names from getting too unwieldy.

neqo-transport/src/events.rs
```diff
@@ -168,7 +168,7 @@ impl ConnectionEvents {
 
     // The number of datagrams in the events queue is limited to max_queued_datagrams.
     // This function ensure this and deletes the oldest datagrams if needed.
-    fn check_datagram_queued(&self, max_queued_datagrams: usize) {
+    fn check_datagram_queued(&self, max_queued_datagrams: usize, stats: &mut Stats) {
         let mut q = self.events.borrow_mut();
         let mut remove = None;
         if q.iter()
```
martinthomson, line 174: Before we ship this, we really need to ensure that we don't keep this bit.  We should just be able to do a simple comparison of `self.datagrams.len()` (or something similarly named) with a target value.

martinthomson (commented):
This part looks good.  The naming is again hard, but the code is solid.

neqo-transport/src/events.rs
ddragana: Yes, open #1218 to fix this. That is the next on my list to fix.

PR #1223, opened 2021-09-10
Continue trying to write datagrams if a datagram is dropped for being too large

ddragana:


ddragana:
This is based on #1220. I notice it while self-reviewing #1220.

neqo-transport/src/connection/mod.rs
```diff
@@ -2880,8 +2885,8 @@ impl Connection {
     /// to check the estimated max datagram size and to use smaller datagrams.
     /// `max_datagram_size` is just a current estimate and will change over
     /// time depending on the encoded size of the packet number, ack frames, etc.
-    pub fn add_datagram(&mut self, buf: &[u8]) -> Res<bool> {
-        self.quic_datagrams.add_datagram(buf)
+    pub fn add_datagram(&mut self, buf: &[u8], id: Option<u64>) -> Res<()> {
```
martinthomson, line 2888: I wonder if we can make a new type for this, just to improve the readability of call sites:

```rust
enum DatagramTracking {
  None,
  Id(u64),
}
```

We can implement `Into<Option<u64>>` and `From<Option<u64>>` for people and even make this signature something like this so that `Some(20)` works (and your tests don't need to change again...):

```suggestion
    pub fn add_datagram(&mut self, buf: &[u8], id: impl Into<DatagramTracking>) -> Res<()> {
```

(Unrelated thought...) we have send_stream(), so why not just send_datagram() ?  "adding" a datagram is more how the operation works that what the caller wants to have happen.  They want to send a datagram.  That we have to add it to a queue is our business.

neqo-transport/src/connection/params.rs
```diff
@@ -223,12 +226,23 @@ impl ConnectionParameters {
         self
     }
 
-    pub fn get_queued_datagrams(&self) -> usize {
-        self.queued_datagrams
+    pub fn get_outgoing_datagrams_queue(&self) -> usize {
```
martinthomson, line 229: You have a datagram queue or a queue of datagrams:

```suggestion
    pub fn get_outgoing_datagram_queue(&self) -> usize {
```

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -43,12 +45,12 @@ fn datagram_disabled_both() {
...
     assert_eq!(server.max_datagram_size(), Err(Error::NotAvailable));
     assert_eq!(
-        client.add_datagram(DATA_SMALLER_THAN_MTU),
+        client.add_datagram(DATA_SMALLER_THAN_MTU, Some(1)),
         Err(Error::TooMuchData)
     );
     assert_eq!(server.stats().frame_tx.datagram, 0);
     assert_eq!(
-        server.add_datagram(DATA_SMALLER_THAN_MTU),
+        server.add_datagram(DATA_SMALLER_THAN_MTU, Some(1)),
```
martinthomson, line 53: If you are using the same ID, then there is no reason to use the ID at all, right?

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -130,24 +136,53 @@ fn limit_data_size() {
     assert!(u64::try_from(DATA_BIGGER_THAN_MTU.len()).unwrap() > DATAGRAM_LEN_MTU);
     // Datagram can be queued because they are smaller than allowed by the peer,
     // but they cannot be sent.
-    assert_eq!(client.add_datagram(DATA_BIGGER_THAN_MTU), Ok(false));
-    assert_eq!(server.add_datagram(DATA_BIGGER_THAN_MTU), Ok(false));
+    assert_eq!(client.add_datagram(DATA_BIGGER_THAN_MTU, Some(1)), Ok(()));
+    assert_eq!(server.add_datagram(DATA_BIGGER_THAN_MTU, Some(1)), Ok(()));
```
martinthomson, line 140: I don't think that you need to test both client and server here (I missed this earlier).  That these use the same identifier is a bit of a warning flag.

If you do test them both like this, then moving the add_datagram() calls down to the other endpoint-specific code will make this clearer (it's easy to miss that these actions are on different endpoints).

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -130,24 +136,53 @@ fn limit_data_size() {
...
+    let datagram_dropped = |e| {
+        matches!(
+        e,
+        ConnectionEvent::OutgoingDatagramOutcome { id, outcome } if id == 1 && outcome == OutgoingDatagramOutcome::DroppedTooBig)
+    };
+
+    let dgram_sent_c = client.stats().frame_tx.datagram;
+    assert!(client.process_output(now()).dgram().is_some());
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent_c + 1);
+    assert!(client.events().any(datagram_dropped));
```
martinthomson, line 177: Events here should occur in a specific order, right?

Can you not just do

```rust
assert!(matches!(client.next_event(), ConnectionEvent::... if id == 1 ...));
assert!(matches!(client.next_event(), ConnectionEvent::... if id == 2 && outcome == OutgoingDatagramOutcome::???));

File: neqo-transport/src/connection/tests/datagram.rs [hunk 6]
```diff
@@ -245,24 +288,45 @@ fn dgram_too_big() {
...
+    );
+    // The outgoing datagram queue limit is 2, therefore the datagram with id 1
+    // will be dropped after adding one more datagram.
+    assert_eq!(client.add_datagram(DATA_MTU, Some(3)), Ok(()));
+    let datagram_dropped = |e| {
+        matches!(
+        e,
+        ConnectionEvent::OutgoingDatagramOutcome{ id , outcome } if id == 1 && outcome == OutgoingDatagramOutcome::DroppedQueueFull)
+    };
+    assert!(client.events().any(datagram_dropped));
```
martinthomson, line 308: Can this use next_event() ?

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -362,3 +426,25 @@ fn too_many_datagram_events() {
     assert_eq!(datagrams.next().unwrap(), FOURTH_DATAGRAM);
     assert!(datagrams.next().is_none())
 }
+
+#[test]
+fn multiple_quic_datagrams_in_one_packet() {
+    let (mut client, mut server) = connect_datagram();
+
+    let dgram_sent = client.stats().frame_tx.datagram;
+    // Que 2 datagrams that can fit in a single packet.
```
martinthomson, line 435:
```suggestion
    // Enqueue 2 datagrams that can fit in a single packet.
```

neqo-transport/src/quic_datagrams.rs
```diff
@@ -10,37 +10,61 @@ use crate::frame::{FRAME_TYPE_DATAGRAM, FRAME_TYPE_DATAGRAM_WITH_LEN};
...
         conn_events: ConnectionEvents,
     ) -> Self {
         Self {
             local_datagram_size,
             remote_datagram_size: 0,
-            max_queued_datagrams,
-            datagram: None,
+            max_queued_outgoing_datagrams,
+            max_queued_incoming_datagrams,
+            datagrams: VecDeque::new(),
```
martinthomson, line 67:
```suggestion
            datagrams: VecDeque::with_capacity(max_queued_outgoing_datagrams),
```

The extra overhead of this allocation will be tiny and we can avoid the overhead of allocation on the first call.

neqo-transport/src/events.rs
```diff
@@ -181,21 +189,23 @@ impl ConnectionEvents {
         }
         if let Some(r) = remove {
             q.remove(r);
-            q.push_back(ConnectionEvent::DatagramLost);
+            q.push_back(ConnectionEvent::IncomingDatagramDropped);
         }
     }
 
     pub fn add_datagram(&self, max_queued_datagrams: usize, data: &[u8]) {
```
martinthomson, line 196: Why is the number of enqueued datagrams not a property of `self`?  (Something to consider when moving events to their own queue.)

martinthomson (commented):
Much better. I have a few niggles, but nothing that should prevent this from landing.

neqo-transport/src/connection/mod.rs
ddragana: This is a good suggestion. I will make this change.

neqo-transport/src/connection/tests/datagram.rs
ddragana: I will move this line and also add a comment.

neqo-transport/src/events.rs
ddragana: I was thinking about moving incoming datagram queue into QuicDatagrams structure. That will not have this issue. We can discuss that in issue #1218

PR #1225, opened 2021-09-15
WebTransport

ddragana:


ddragana:
This is still missing part for closing wt streams. This is still not ready.

martinthomson:
it's a big one.  I'll need to carve some time out for this.  let me know what sort of time frame you would like to operate on.

neqo-http3/src/connection_server.rs
```diff
@@ -49,12 +51,112 @@ impl Http3ServerHandler {
...
+                );
+                self.base_handler
+                    .insert_streams_have_data_to_send(stream_id);
+                Ok(())
+            }
+        }
+    }
+
+    /// Create a WebTraansport stream.
+    pub fn wt_create_stream(
```
martinthomson, line 114: check naming

neqo-http3/src/connection_server.rs
```diff
@@ -49,12 +51,112 @@ impl Http3ServerHandler {
...
+        stream_type: StreamType,
+    ) -> Res<u64> {
+        let wt_session = self
+            .base_handler
+            .send_streams
+            .get_mut(&stream_id)
+            .ok_or(Error::InvalidStreamId)?
+            .get_wt_session()
+            .ok_or(Error::InvalidStreamId)?;
+        match WebTransportSession::create_new_stream_local(wt_session, stream_type, conn)? {
```
martinthomson, line 127: Note to self: this split probably could be improved.  We might need different handling entirely based on the stream type rather than have this awkward return value handling.

neqo-http3/src/connection_server.rs
```diff
@@ -182,11 +270,34 @@ impl Http3ServerHandler {
...
+                let res = self.base_handler.handle_stream_readable(conn, stream_id)?;
+                assert_eq!(ReceiveOutput::NoOutput, res);
+                Ok(())
+            }
             ReceiveOutput::ControlFrames(control_frames) => {
                 for f in control_frames {
                     match f {
-                        HFrame::MaxPushId { .. } => {
+                        HFrame::MaxPushId { .. } | HFrame::Settings { .. } => {
+                            // Settings frame is only handled by the client.
```
martinthomson, line 300: This is a worrying comment.  I think that you are saying that on the client needs to have anything other than the base handler process a SETTINGS frame; or that the base handler logic is sufficient for a server.

neqo-http3/src/connection_server.rs
```diff
@@ -235,22 +346,37 @@ impl Http3ServerHandler {
         now: Instant,
         stream_id: u64,
         buf: &mut [u8],
+        wt: bool,
```
martinthomson, line 349:
```suggestion
        webtransport: bool,
```

... though I don't think that you want this.  More below.

neqo-http3/src/connection_server.rs
```diff
@@ -235,22 +346,37 @@ impl Http3ServerHandler {
...
-            .ok_or(Error::InvalidStreamId)?
-            .http_stream()
-            .ok_or(Error::InvalidStreamId)?;
-
-        let res = recv_stream.read_data(conn, buf);
-        if let Err(e) = &res {
-            self.close(conn, now, e);
-        } else if recv_stream.done() {
-            self.base_handler.recv_streams.remove(&stream_id);
+        let (res, recv_stream) = if !wt {
```
martinthomson, line 352: Did you run this through clippy?  It should complain about this.

neqo-http3/src/connection_server.rs
```diff
@@ -235,22 +346,37 @@ impl Http3ServerHandler {
...
+            let recv_stream = self
+                .base_handler
+                .recv_streams
+                .get_mut(&stream_id)
+                .ok_or(Error::InvalidStreamId)?;
+            let http = recv_stream.http_stream().ok_or(Error::InvalidStreamId)?;
```
martinthomson, line 358: Given that we know the type of stream is involved, I think we should be conditionally extracting the stream type, as in:

```rust
            let recv_stream = self
                .base_handler
                .recv_streams
                .get_mut(&stream_id)
        let res = if let Some(wt) = recv_stream.wt_stream() {
           wt.read_data(conn, buf)
        } else if let Some(http) = recv_stream.http_stream() {
           http.read_data(conn, buf)
        } else {
            return Err(Error::InvalidStreamId);
        };
        match res { ...
```

Then you don't need the `wt` argument.

neqo-http3/src/lib.rs
```diff
@@ -325,6 +332,43 @@ pub(crate) trait RecvMessageEvents: Debug {
...
+
+pub trait SendStream: Debug {
+    fn send(&mut self, conn: &mut Connection) -> Res<()>;
+    fn has_data_to_send(&self) -> bool;
+    fn stream_writable(&self);
+    fn done(&self) -> bool;
+    fn stop_sending(&mut self, error: AppError);
+    fn http_stream(&mut self) -> Option<&mut dyn HttpSendStream>;
+    fn get_wt_session(&self) -> Option<Rc<RefCell<WebTransportSession>>>;
+    fn wt_stream(&mut self) -> Option<&mut dyn WtSendStream>;
```
martinthomson, line 365: Another case where we might want to spell "webtransport" out in full.

neqo-http3/src/lib.rs
```diff
@@ -325,6 +332,43 @@ pub(crate) trait RecvMessageEvents: Debug {
...
+    fn clone_box(&self) -> Box<dyn WtEvents>;
+}
+
+pub trait SendStream: Debug {
+    fn send(&mut self, conn: &mut Connection) -> Res<()>;
+    fn has_data_to_send(&self) -> bool;
+    fn stream_writable(&self);
+    fn done(&self) -> bool;
+    fn stop_sending(&mut self, error: AppError);
+    fn http_stream(&mut self) -> Option<&mut dyn HttpSendStream>;
```
martinthomson, line 363: This (and the two after), could benefit from a default implementation here now that we have a few more places that this is implemented as:

```suggestion
    fn http_stream(&mut self) -> Option<&mut dyn HttpSendStream> { None }
```

neqo-http3/src/recv_message.rs
```diff
@@ -94,11 +100,16 @@ impl RecvMessage {
         conn_events: Box<dyn RecvMessageEvents>,
         push_handler: Option<Rc<RefCell<PushController>>>,
         priority_handler: PriorityHandler,
+        header_frame_type_read: bool, // TODO Use stream PEEK for decoding a stream type instread of consuming data.
     ) -> Self {
+        let mut frame_reader = HFrameReader::new();
+        if header_frame_type_read {
+            let mut d = Encoder::default();
+            d.encode_varint(H3_FRAME_TYPE_HEADERS);
+            frame_reader.consume(Decoder::from(&d)).unwrap();
```
martinthomson, line 109: This seems like it is wasteful.  We should just tell HFrameReader to skip straight to the bit where it reads the length.

neqo-http3/src/recv_message.rs
```diff
@@ -161,12 +172,26 @@ impl RecvMessage {
             return Err(Error::HttpGeneralProtocolStream);
         }
 
-        self.conn_events
-            .header_ready(self.stream_id, headers, interim, fin);
+        let is_web_transport = self.message_type == MessageType::Request
+            && headers
+                .iter()
+                .any(|h| h.name() == ":method" && h.value() == "CONNECT")
```
martinthomson, line 178: An idea that might be worth doing in future, but maybe it's worth it anyway...

When we look for pseudo-header fields, it would be much faster if we stopped looking once we got a non-pseudo-header field...

```rust
headers.iter().take_until(|h| h[0] != ':').any(|h| ....)
```

Note:
1. This assumes that we validate headers and ensure that !h.is_empty() somewhere.
1. That also speaks to having a struct for headers rather than a dumb `Vec<Header>`.

neqo-http3/src/recv_message.rs
```diff
@@ -368,6 +400,7 @@ impl RecvMessage {
                     ":scheme" => PSEUDO_HEADER_SCHEME,
                     ":authority" => PSEUDO_HEADER_AUTHORITY,
                     ":path" => PSEUDO_HEADER_PATH,
+                    ":protocol" => PSEUDO_HEADER_PROTOCOL,
```
martinthomson, line 403: We should also check that this does not appear on anything other CONNECT.  Down below vvv in headers_valid, which appears to need some validation.

neqo-http3/src/server.rs
```diff
@@ -33,6 +35,7 @@ const MAX_EVENT_DATA_SIZE: usize = 1024;
 pub struct Http3Server {
     server: Server,
     qpack_settings: QpackSettings,
+    enable_wt: bool,
```
martinthomson, line 38: HttpServerParameters time?

neqo-http3/src/server.rs
```diff
@@ -230,21 +283,36 @@ fn prepare_data(
...
-                    fin,
-                );
+                if !wt {
+                    events.data(
+                        ClientRequestStream::new(conn.clone(), handler.clone(), stream_id),
+                        data,
+                        fin,
+                    );
+                } else {
+                    events.web_transport_stream_data(
```
martinthomson, line 310: I wonder if events.data() might be made generic so that it could take a webtransport event.  You would need to make this function take generic also, so maybe it's not worth it (`FnMut(&mut ActiveConnectionRef, &HandlerRef, u64) -> impl DataEvent`) but all these "is webtransport" checks are not a good sign as it makes the code much harder to reason about.

neqo-http3/src/server.rs
```diff
@@ -876,8 +945,7 @@ mod tests {
                         .unwrap();
                     data_received += 1;
                 }
-                Http3ServerEvent::StateChange { .. } => {}
-                Http3ServerEvent::PriorityUpdate { .. } => {}
+                _ => {}
```
martinthomson, line 948: This is why you want to mark off webtransport-specific events.  We should not be using _ in matches like this (it makes things more brittle).

neqo-http3/src/server_connection_events.rs
```diff
@@ -36,6 +35,27 @@ pub(crate) enum Http3ServerConnEvent {
     //Reset { stream_id: u64, error: AppError },
     /// Connection state change.
     StateChange(Http3State),
+    WebTransportNewSession {
```
martinthomson, line 38: Carving out a separate enum for webtransport events might help here too.

neqo-http3/src/server_events.rs
```diff
@@ -53,24 +78,22 @@ impl ClientRequestStream {
...
     pub fn set_response(&mut self, headers: &[Header], data: &[u8]) -> Res<()> {
         qinfo!([self], "Set new response.");
-        self.handler
-            .borrow_mut()
-            .set_response(self.stream_id, headers, data)
+        self.stream_handler.handler.borrow_mut().set_response(
+            self.stream_handler.stream_id,
+            headers,
+            data,
+        )
```
martinthomson, line 85: This part is HTTP-specific, so it can go here, but you might want to think about how much this object needs to manipulate the contents of the stream_handler's handler in order to get things done.

neqo-http3/src/server_events.rs
```diff
@@ -53,24 +78,22 @@ impl ClientRequestStream {
...
-            self.stream_id,
-            app_error
-        );
-        self.conn
+        qdebug!([self], "stop sending error={}.", app_error);
+        self.stream_handler
+            .conn
             .borrow_mut()
-            .stream_stop_sending(self.stream_id, app_error)?;
+            .stream_stop_sending(self.stream_handler.stream_id, app_error)?;
```
martinthomson, line 96: Please move some of this to the new struct, particularly these parts that are generic.

neqo-http3/src/server_events.rs
```diff
@@ -79,30 +102,202 @@ impl ClientRequestStream {
...
+                stream_id,
+            },
+        }
+    }
+
+    pub fn stream_id(&self) -> u64 {
+        self.stream_handler.stream_id
+    }
+
+    /// Send data ona WebTransport Stream.
```
martinthomson, line 235:
```suggestion
    /// Send data on a WebTransport Stream.
```

neqo-http3/src/server_events.rs
```diff
@@ -79,30 +102,202 @@ impl ClientRequestStream {
...
+    /// # Errors
+    /// It may return `InvalidStreamId` if a stream does not exist anymore or if
+    /// the stream is only a receiver stream.
+    pub fn send_data(&mut self, data: &[u8]) -> Res<usize> {
+        qinfo!([self], "Send data.");
+        self.stream_handler
+            .handler
+            .borrow_mut()
+            .wt_stream_send_data(
```
martinthomson, line 244: I would like to understand why "send data on stream X" doesn't work for this.

neqo-http3/src/server_events.rs
```diff
@@ -79,30 +102,202 @@ impl ClientRequestStream {
...
+            .handler
+            .borrow_mut()
+            .wt_session_response(
+                &mut self.stream_handler.conn.borrow_mut(),
+                self.stream_handler.stream_id,
+                accept,
+            )
+    }
+
+    /// Create a WebTraansport stream.
```
martinthomson, line 169:
```suggestion
    /// Create a WebTransport stream.
```

martinthomson (commented):
I had a little time, but nowhere near enough to get through all of this.

When you think that this is more baked, I'd appreciate a writeup of some of the major design decisions and the rationale.  There are a few things here for which it is hard to trace the reasons for the design choices you have made.

ddragana:
Thanks for a review, but this is really not done. I am not sure about some stuff here. Also I am not happy about some parts of it either. 
I need some more time to work on it before this is ready for a review.  I am also sure that some details are not compliant with the spec at the moment.

I will look at your comments, but just ignore he PR until I think it is ready for a first look.

neqo-http3/src/client_events.rs
```diff
@@ -138,6 +178,36 @@ impl SendMessageEvents for Http3ClientEvents {
...
+
+    fn web_transport_data_writable(&self, stream_id: u64) {
+        self.insert(Http3ClientEvent::WebTransportDataWritable { stream_id });
+    }
+
+    fn web_transport_stream_stop_sending(&self, stream_id: u64, error: AppError) {
+        self.insert(Http3ClientEvent::WebTransportStreamStopSending { stream_id, error });
+    }
+
+    fn clone_box(&self) -> Box<dyn WtEvents> {
```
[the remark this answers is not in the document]
ddragana: This is some of the parts that I do not remember why I did it (I wrote this code 2 months ago). I need to do a self review of the code, I only did a rebase yesterday.

neqo-http3/src/connection.rs
```diff
@@ -107,18 +111,43 @@ impl Http3Connection {
...
+                    HSetting {
+                        setting_type: HSettingType::MaxTableCapacity,
+                        value: self.qpack_decoder.borrow().get_max_table_size(),
+                    },
+                    HSetting {
+                        setting_type: HSettingType::BlockedStreams,
+                        value: self.qpack_decoder.borrow().get_blocked_streams().into(),
+                    },
+                ])
+            },
```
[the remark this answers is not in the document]
ddragana: I did not want to allocate a Vec. But this should be changed.

ddragana:
We took a bit different approach.

PR #1227, opened 2021-09-16
Lots of lints

martinthomson:
It seems like a lot of code landed without any of these checks being
run.  Hopefully we can be stricter about this in future.  There were
some minor (but relevant) problems in there.

(Lots of missing semicolons, which I blame rustfmt for; and lots of
extra `&`, which I might have to own up to.)

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -263,7 +263,7 @@ impl Http3Client {
     /// is created. A response body may be added by calling `send_request_body`.
     /// # Errors
     /// If a new stream cannot be created an error will be return.
-    #[allow(clippy::too_many_arguments)] //
+    #[allow(clippy::too_many_arguments)] // https://github.com/mozilla/neqo/issues/1226
     pub fn fetch(
         &mut self,
         now: Instant,
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -172,7 +172,7 @@ impl Http3Client {
     /// the same origin.
     #[must_use]
     pub fn connection_id(&self) -> &ConnectionId {
-        &self.conn.odcid().expect("Client always has odcid")
+        self.conn.odcid().expect("Client always has odcid")
     }
 
     /// A resumption token encodes transport and settings parameter as well.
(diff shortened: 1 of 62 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -263,6 +263,7 @@ impl Http3Client {
     /// is created. A response body may be added by calling `send_request_body`.
     /// # Errors
     /// If a new stream cannot be created an error will be return.
+    #[allow(clippy::too_many_arguments)] //
```
ddragana, line 266: nit: remove "//"

neqo-transport/src/lib.rs
```diff
@@ -59,7 +59,7 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
 const ERROR_AEAD_LIMIT_REACHED: TransportError = 15;
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
-#[allow(clippy::pub_enum_variant_names)]
+// #[allow(clippy::pub_enum_variant_names)]
```
ddragana, line 62: you want to remove this?

neqo-transport/src/quic_datagrams.rs
```diff
@@ -111,7 +111,7 @@ impl QuicDatagrams {
     ) {
         while let Some(dgram) = self.datagrams.pop_front() {
             let len = dgram.len();
-            if builder.remaining() >= len + 1 {
+            if builder.remaining() > len {
```
ddragana, line 114: this "+ 1" was for he frame type. I had a comment, i looks like I have lost it in a iteration.

Please add a comment that more than len bytes is needed because of the frame type.

ddragana (commented):
A couple of nits. Please land the PR after fixing them.

neqo-transport/src/lib.rs
martinthomson: Yeah, for reasons I don't understand, this doesn't generate warnings.  But including the override fails in 1.54.0.  🤷‍♀️

neqo-transport/src/quic_datagrams.rs
martinthomson: The comment is there on the next line.  I'll revise it slightly.  I found a bug here though.  Opening an issue for you.

PR #1231, opened 2021-09-17
Define and use RequestTarget for fetch

### LINKED ISSUE #1226: Fetch function has too many arguments.

martinthomson:
This does a little generic magic in order to allow for three basic forms
of request target natively:

* A simple string (this is the most costly, because we have to parse the
URL)
* A Url (in case one exists, as is the case with neqo-client)
* A three-tuple of things that turn into `str`

More of these can be implemented if that turns out to be convenient for
the caller.  Monomorphism ensures that only those forms that are used
are paid for.

This was a little fiddly to implement.  I would have liked to make the
string form more generic, ideally `AsRef<str>` but because we take
`&Url` and because `Url` implements `AsRef<str>` that causes rust to get
confused.  I tried fixing it, but you end up needing extra type
parameters on every invocation, which is less friendly than just having
people call `as_ref()` themselves.

Closes #1226.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -10,6 +10,7 @@ use crate::hframe::HFrame;
 use crate::push_controller::PushController;
 use crate::push_stream::PushStream;
 use crate::recv_message::{MessageType, RecvMessage};
+use crate::request_target::{AsRequestTarget, RequestTarget};
 use crate::send_message::{SendMessage, SendMessageEvents};
 use crate::settings::HSettings;
 use crate::{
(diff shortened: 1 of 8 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -1248,9 +1245,7 @@ mod tests {
             .fetch(
                 now(),
                 "GET",
-                "https",
-                "something.com",
-                "/",
+                "https://something.com/",
```
martinthomson, line 1248: This is the "easy" form of the interface, but it is the least efficient as the string needs to be parsed into a URL.  This hides that cost somewhat, which is un-rust-like, but ergonomics >> purity.

ddragana (approved):

Issue #1168, opened 2021-06-18
Update minimum rust version to 1.51.0


martinthomson:
The next release of Firefox looks like it might move the minimum rust version up to 1.51.0.  This is good because some of the new lints in rust 1.53.0 insist on changes that break in earlier versions.  And it is about time we cut 1.43.0 loose.

martinthomson:
OK, maybe the linting situation needs to be fixed some other way (--cap-lint perhaps) as 1.51 rejects or-patterns where 1.53 requires them.

martinthomson:
I've started a branch (named gha) that uses GitHub Actions for CI.  I've had success with that on another project and it should help avoid the current docker-bound mess we have that isn't really maintainable.

PR #1232, opened 2021-09-21
Set clippy lint msrv to rust 1.51

mb:
So it doesn't have to be annotated in the code. Solves lint problem mentioned in #1168.


neqo-crypto/src/replay.rs
```diff
@@ -15,12 +15,7 @@ use std::ptr::null_mut;
 use std::time::{Duration, Instant};
 
 // This is an opaque struct in NSS.
-#[allow(
-    unknown_lints,
-    renamed_and_removed_lints,
-    clippy::unknown_clippy_lints,
-    clippy::upper_case_acronyms
-)] // Until we require rust 1.51.
+#[allow(clippy::upper_case_acronyms)] // Until we require rust 1.51.
```
martinthomson, line 18:
```suggestion
#[allow(clippy::upper_case_acronyms)]
```

We already require 1.51 :)

neqo-crypto/src/err.rs
```diff
@@ -34,7 +29,7 @@ pub mod nspr {
 pub type Res<T> = Result<T, Error>;
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
-#[allow(clippy::pub_enum_variant_names)]
+#[allow(renamed_and_removed_lints, clippy::pub_enum_variant_names)] // until we require rust 1.54
```
martinthomson, line 32: This will need another fix once we get to 1.54.  Maybe instead:

```suggestion
#[allow(renamed_and_removed_lints, clippy::pub_enum_variant_names)] // rust 1.54 will require a different approach
```

martinthomson (commented):
Thanks for cleaning this up.  Just a couple of comments on comments.

ddragana (approved):

PR #1236, opened 2021-09-27
Fast PTOs

martinthomson:
Fastly have an option that enables faster PTOs.  This code adds a tuning parameter that scales PTOs.  It allows for tuning the base scaling of the timer, with a value of 100 being normal; 200 doubles the PTO; 50 halves it.

This deliberately doesn't affect other things that depend on the base PTO timer, like persistent congestion or idle timeouts.  Though those might be affected by a higher or lower rate of "retransmission", changing the timer substantially there might have unforeseen consequences.

The idea here is to allow people to set a value of ~50 to improve latency in cases of tail loss.  The trade-off being that there are more packets sent that might be redundant.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -393,7 +393,7 @@ impl Connection {
             streams: Streams::new(tphandler, role, events.clone()),
             connection_ids: ConnectionIdStore::default(),
             state_signaling: StateSignaling::Idle,
-            loss_recovery: LossRecovery::new(stats.clone()),
+            loss_recovery: LossRecovery::new(stats.clone(), conn_params.get_fast_pto()),
             events,
             new_token: NewTokenState::new(role),
             stats,
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -5,6 +5,7 @@
 // except according to those terms.
 
 use crate::connection::{ConnectionIdManager, Role, LOCAL_ACTIVE_CID_LIMIT};
+pub use crate::recovery::FAST_PTO_SCALE;
 use crate::recv_stream::RECV_BUFFER_SIZE;
 use crate::rtt::GRANULARITY;
 use crate::stream_id::StreamType;
@@ -70,6 +71,7 @@ pub struct ConnectionParameters {
     datagram_size: u64,
     outgoing_datagram_queue: usize,
     incoming_datagram_queue: usize,
+    fast_pto: u8,
 }
 
 impl Default for ConnectionParameters {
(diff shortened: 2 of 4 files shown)


martinthomson:
Interested in whether you think we should do this for Firefox @ddragana.  (I hear that Fastly set this to 1/4; that seems aggressive, but we might get some benefit from 1/2.)

ddragana:
> Interested in whether you think we should do this for Firefox @ddragana. (I hear that Fastly set this to 1/4; that seems aggressive, but we might get some benefit from 1/2.)

We can try this. I do not see that there will be a lot of damage. With H1.1 we needed to be more careful, because we open 6 connection and with high packet loss depending how you open this connection (connection pacing) you may get something through or not. I am  not worried in h3 case. We should try it out and watch telemetry.

ddragana (approved):

mxinden:
We're [about to enable Fast PTO](https://phabricator.services.mozilla.com/D306763) on all Firefox QUIC connections, and I want to make sure I understand the trade-off before we do: we're trading extra bytes sent against faster tail-loss recovery, right?

The case I'm unsure about is sequential request/response on a stable, moderate-RTT path (request → wait → response → idle → repeat). Whenever the scaled PTO is below the RTT, each request seems to cost one spurious probe:

- Given: `smoothed_rtt` = 100 ms, `4*rttvar` = 20 ms, `max_ack_delay` = 25 ms, `fast_pto = 50` (`F = 0.5`)
- Spec PTO = `145 ms` (> RTT). Scaled: **72.5 ms, below the 100 ms RTT.**
- `t=0`: send request, arm PTO at 72.5 ms.
- `t=72.5`: ACK can't be back yet → **spurious probe**, backoff to 145 ms.
- `t≈100`: real ACK arrives, disarms, resets count.

Idle between requests keeps the RTT estimate fixed, so this repeats per request: N requests → N wasted probes, none recovering real loss. Triggers whenever `RTT > 25 ms + 4*rttvar`.

@martinthomson @larseggert is this analysis correct, and is it a concern worth mitigating, or is the overhead small enough in practice that we don't care?

Related: https://github.com/mozilla/neqo/pull/3662

mxinden:
Discussion with Martin on metrics:

- time to dead-connection-detection
- time of tail loss resend

larseggert:
@mxinden I think you're correct that this causes a spurious probe for longer RTTs. I don't have a good feeling for whether we should care. IIRC Chrome does something like this, too, but maybe they have a different mechanism.

larseggert:
IIRC Chrome does `kPtoRttvarMultiplier = 2` (and not 4). Maybe that is something we should consider...

martinthomson:
Reducing the effect of RTTVAR would not necessarily fix the issue.  It might even make it worse in this case.

PR #1238, opened 2021-09-28
For an easier fuzzing we need a fix initial CID

ddragana:


ddragana:
After chatting with truber. We may not need this. I will leave it as a draft for now.

larseggert:
This is redundant with the `disable-random` feature in #1764.

PR #1240, opened 2021-09-29
Change to receiving streams traits to better fit WebTransport and some code cleanup

ddragana:


neqo-http3/src/connection.rs
```diff
@@ -210,11 +210,9 @@ impl Http3Connection {
...
 
         if let Some(recv_stream) = self.recv_streams.get_mut(&stream_id) {
-            let output = recv_stream.receive(conn);
-            if recv_stream.done() {
-                self.recv_streams.remove(&stream_id);
-            }
-            output
+            let res = recv_stream.receive(conn);
+            self.handle_stream_manipulation_output(res, stream_id, conn)
+                .map_or_else(Err, |(output, _)| Ok(output))
```
martinthomson, line 215: Won't a simple map() work here?

neqo-http3/src/connection_client.rs
```diff
@@ -676,12 +647,13 @@ impl Http3Client {
...
+            Box::new(RecvMessage::new(
+                MessageType::Response,
                 stream_id,
-                push_id,
-                Rc::clone(&self.push_handler),
                 Rc::clone(&self.base_handler.qpack_decoder),
-                Priority::default(),
+                Box::new(RecvPushEvents::new(push_id, Rc::clone(&self.push_handler))),
+                None,
+                PriorityHandler::new(true, Priority::default()),
```
martinthomson, line 656: Can you flag this with a TODO: it would be good to examine what priority we set push streams relative to the request.

neqo-http3/src/recv_message.rs
```diff
@@ -420,68 +421,30 @@ impl RecvMessage {
 
         Ok(())
     }
+
+    fn done(&self) -> bool {
```
martinthomson, line 425: Is this only called from RecvStream::receive?  It could be inlined if it is.

neqo-http3/src/recv_message.rs
```diff
@@ -29,7 +29,7 @@ const PSEUDO_HEADER_AUTHORITY: u8 = 0x8;
 const PSEUDO_HEADER_PATH: u8 = 0x10;
 const REGULAR_HEADER: u8 = 0x80;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
martinthomson, line 32:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```

I'm not sure about when we should use Eq, but it seems like this is a good case for that.  And this type really needs to implement Copy.

neqo-http3/src/lib.rs
```diff
@@ -298,33 +297,47 @@ pub enum ReceiveOutput {
...
+impl Default for ReceiveOutput {
+    fn default() -> Self {
+        Self::NoOutput
+    }
+}
+
 pub trait RecvStream: Debug {
+    /// # Errors
+    /// An error may happen while reading a stream, e.g. early close, protocol error, etc.
+    fn receive(&mut self, conn: &mut Connection) -> Res<(ReceiveOutput, bool)>;
```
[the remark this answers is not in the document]
martinthomson: I know that Result is automatically marked as `#[must_use]`.  However, there is a risk that the value of an Ok result is ignored.  I think that we should mark `ReceiveOutput` as `#[must_use]` also.

martinthomson (commented):
Another good set of changes here.

I was a little uncertain about collapsing the RecvStream::done() into a tuple on the return of receive (and friends), but when I looked at the tests, it seems like it is a good way to ensure that the value isn't ignored.

Take a look at the comments, but land when you are done.

neqo-http3/src/connection_client.rs
ddragana: I will add a TODO.
(I do not bother with ServerPush that much at the moment, since I am not sure when we will turn it on in Firefox)

PR #1241, opened 2021-09-29
Create a BufferedStream

ddragana:
BufferedStream may have some data in a buffer and it will always make use to send the buffered data first.

### DIFF
--- a/neqo-http3/src/buffered_send_stream.rs
+++ b/neqo-http3/src/buffered_send_stream.rs
@@ -22,7 +22,7 @@ impl Default for BufferedStream {
 
 impl ::std::fmt::Display for BufferedStream {
     fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        write!(f, "BufferedStream {:?}", self.stream_id())
+        write!(f, "BufferedStream {:?}", Option::<u64>::from(self))
     }
 }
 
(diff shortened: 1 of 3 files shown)


neqo-http3/src/control_stream_local.rs
```diff
@@ -112,12 +94,15 @@ impl ControlStreamLocal {
     /// Create a control stream.
     pub fn create(&mut self, conn: &mut Connection) -> Res<()> {
         qtrace!([self], "Create a control stream.");
-        self.stream_id = Some(conn.stream_create(StreamType::UniDi)?);
+        self.stream
+            .init(StreamId::from(conn.stream_create(StreamType::UniDi)?));
```
martinthomson, line 98: I wonder if implementing `From` would be better.

neqo-http3/src/buffered_send_stream.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    #[must_use]
+    pub fn stream_id(&self) -> Option<u64> {
+        match self {
+            Self::Uninitialized => None,
+            Self::Initialized { stream_id, .. } => Some(stream_id.as_u64()),
+        }
+    }
```
martinthomson, line 105: This is almost `impl From<BufferedStream> for Option<u64>`.

martinthomson (commented):
A good cleanup.  Fix and land.

PR #1242, opened 2021-09-29
Change the  stream side trait to better fit WebTransport

ddragana:


ddragana:
This change is based on #1240 . Only last to commits are new.

KershawChang (approved):
looks good!

PR #1243, opened 2021-09-29
Decode WebTransport streams' head

ddragana:


neqo-http3/src/connection.rs
```diff
@@ -476,6 +471,53 @@ impl Http3Connection {
         self.recv_streams.clear();
     }
 
+    fn handle_stream_manipulation_output<U>(
```
KershawChang, line 474: Sorry, I found this function is not easy to understand for me. Could you add some comments?

neqo-http3/src/client_events.rs
```diff
@@ -83,24 +81,19 @@ pub struct Http3ClientEvents {
...
-    fn header_ready(&self, stream_id: u64, headers: Vec<Header>, interim: bool, fin: bool) {
-        self.insert(Http3ClientEvent::HeaderReady {
-            stream_id,
-            headers,
-            interim,
-            fin,
-        });
-    }
-
+impl RecvStreamEvents for Http3ClientEvents {
```
KershawChang, line 84: Sorry for a stupid question: is `RecvStreamEvents` only used by webtransport? what's the difference with `HttpRecvStreamEvents`?

ddragana: `RecvStreamEvents` contain functions that are more general and common for http streams and webtransport streams.
`HttpRecvStreamEvents` extends `RecvStreamEvents` and adds functions that are only specific to http streams

neqo-http3/src/connection.rs
ddragana: I have added some explanations.

KershawChang (approved):

PR #1246, opened 2021-09-30
Refactor stream closing events

ddragana:
The stream closing events, e.g. application close, RESET frame, etc., will be used by WebTransport. This patch tries to cleanup that code so that it would be easier to extend it for WebTransport. ServerPush already uses the events trait, RecvStreamEvents, for the bookkeeping. For example, when a push stream is canceled or finished, the push stream listener that implements RecvStreamEvents, calls appropriate PushController functions to remove the information about the stream, etc. WebTransport will use the events in a similar way.

RecvStream’s function reset() and SendStream’s function stop_sending() now use CloseType. The two functions will be used to inform a stream that it is abruptly closed, either by application(CloseType::AppError), stream error (e.g. unallowed request headers;CloseType::ResetLocal(error)) or the peer reset(CloseType::ResetRemote(error)).

Reset and done functions from RecvStreamEvents are replaced by one function recv_closed that also uses CloseType. Similarly, stop_sending and done functions from SendStreamEvents are replaced by send_closed().
CloseType values are:
Done - stream has finished without error.
AppError(error) - the application closed the stream.
ResetLocal(error) - there was an error on the stream that is not the session error, e.g. unallowed headers.
ResetRemote(error) - the peer has sent RESET/STOP_SENDING frame.

How this all work:
SendStreams will be closed by calling close()(closing without an error) or stop_sending(). These functions will call the events listener function send_closed with appropriate CloseType. If close() is called on a sending stream send_closed will be called immediately, even though not all data is sent (We may need to change this in a separate PR).
RecvSteam will be closed by reset() or by receiving a fin. The fin will be received during execution of functions receive() or read_data(). These functions will call the listener function recv_closed with appropriate CloseType.

Additional changes:
- remove_recv_side_events_for_stream_id and similar functions are not used any more outside Http3ClientEvent and Http3ServerConnEvent. This is a cleaner design.  

- stream_reset API function was used to reset HTTP requests. In the HTTP case it is usual that we want to close both sides, sending and receiving side, of a stream. For WebTransport this will not be the case. This patch renames stream_reset into cancel_http_request and prepares 2 additional functions stream_stop_sending and stream_reset_send. The API for here 2 functions will be added when WebTransport is implemented.

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -89,14 +89,20 @@ impl RecvStreamEvents for Http3ClientEvents {
 
     /// Add a new `Reset` event.
     fn recv_closed(&self, stream_id: u64, close_type: CloseType) {
-        if close_type == CloseType::Done {
-            return;
-        }
-        self.remove_recv_side_events_for_stream_id(stream_id);
         let (local, error) = match close_type {
-            CloseType::ResetApp(_) | CloseType::Done => return,
-            CloseType::ResetRemote(e) => (false, e),
-            CloseType::LocalError(e) => (true, e),
+            CloseType::ResetApp(_) => {
+                self.remove_recv_stream_events(stream_id);
+                return;
+            }
+            CloseType::Done => return,
+            CloseType::ResetRemote(e) => {
+                self.remove_recv_stream_events(stream_id);
+                (false, e)
+            }
+            CloseType::LocalError(e) => {
+                self.remove_recv_stream_events(stream_id);
+                (true, e)
+            }
         };
 
         self.insert(Http3ClientEvent::Reset {
@@ -126,7 +132,7 @@ impl SendStreamEvents for Http3ClientEvents {
     }
 
     fn send_closed(&self, stream_id: u64, close_type: CloseType) {
-        self.remove_send_side_events_for_stream_id(stream_id);
+        self.remove_send_stream_events(stream_id);
         if let CloseType::ResetRemote(error) = close_type {
             self.insert(Http3ClientEvent::StopSending { stream_id, error });
         }
@@ -212,7 +218,7 @@ impl Http3ClientEvents {
     }
 
     /// Remove all events for a stream
-    fn remove_recv_side_events_for_stream_id(&self, stream_id: u64) {
+    fn remove_recv_stream_events(&self, stream_id: u64) {
         self.remove(|evt| {
             matches!(evt,
                 Http3ClientEvent::HeaderReady { stream_id: x, .. }
@@ -222,7 +228,7 @@ impl Http3ClientEvents {
         });
     }
 
-    fn remove_send_side_events_for_stream_id(&self, stream_id: u64) {
+    fn remove_send_stream_events(&self, stream_id: u64) {
         self.remove(|evt| {
             matches!(evt,
                 Http3ClientEvent::DataWritable { stream_id: x }
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -7,7 +7,7 @@
 #![allow(clippy::module_name_repetitions)]
 
 use crate::connection::Http3State;
-use crate::{Header, HttpRecvStreamEvents, RecvStreamEvents, ResetType, SendStreamEvents};
+use crate::{CloseType, Header, HttpRecvStreamEvents, RecvStreamEvents, SendStreamEvents};
 
 use neqo_common::event::Provider as EventProvider;
 use neqo_crypto::ResumptionToken;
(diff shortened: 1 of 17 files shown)


ddragana:
Only the last change is the new one. This is based on #1243

neqo-http3/src/connection.rs
```diff
@@ -520,37 +517,99 @@ impl Http3Connection {
...
 
         // Stream may be already be closed and we may get an error here, but we do not care.
-        mem::drop(conn.stream_reset_send(stream_id, error));
-        // Stream may be already be closed and we may get an error here, but we do not care.
-        mem::drop(conn.stream_stop_sending(stream_id, error));
+        conn.stream_stop_sending(stream_id, error)?;
+        Ok(())
+    }
+
+    pub fn cancel_http_request(
```
martinthomson, line 569: how about cancel_fetch(), to match fetch()

neqo-http3/src/client_events.rs
```diff
@@ -88,19 +88,17 @@ impl RecvStreamEvents for Http3ClientEvents {
...
     /// Add a new `Reset` event.
-    fn reset(&self, stream_id: u64, error: AppError, reset_type: ResetType) {
-        let local = match reset_type {
-            ResetType::App => return,
-            ResetType::Remote => false,
-            ResetType::Local => true,
+    fn recv_closed(&self, stream_id: u64, close_type: CloseType) {
+        if close_type == CloseType::Done {
+            return;
+        }
```
martinthomson, line 94: If you move the remove_recv_side_events_for_stream_id (that name is too long) AFTER the match, then you don't need this if statement.

neqo-http3/src/connection.rs
```diff
@@ -150,26 +150,33 @@ impl Http3Connection {
...
-
+    fn send_non_control_streams(&mut self, conn: &mut Connection) -> Res<()> {
         let to_send = mem::take(&mut self.streams_have_data_to_send);
         for stream_id in to_send {
-            let mut remove = false;
-            if let Some(s) = &mut self.send_streams.get_mut(&stream_id) {
+            let done = if let Some(s) = &mut self.send_streams.get_mut(&stream_id) {
                 s.send(conn)?;
                 if s.has_data_to_send() {
                     self.streams_have_data_to_send.insert(stream_id);
```
martinthomson, line 159: Just noticed that this should be called streams_with_data_to_send, streams_with_pending_data, or similar.

martinthomson, line 159: Also, should this function use retain() instead?  Or is it better to assume that we won't be needing to put streams back on the pending queue?

neqo-http3/src/connection.rs
```diff
@@ -520,37 +517,99 @@ impl Http3Connection {
...
+
+                // Stream may be already be closed and we may get an error here, but we do not care.
+                mem::drop(self.stream_stop_sending(conn, stream_id, error));
+            }
+            (Some(s), Some(r)) => {
+                let st = s.stream_type();
+                let rt = r.stream_type();
+                if st != rt {
+                    panic!("The sender and receive stream types must be the same");
+                }
```
martinthomson, line 603: `debug_assert_eq!(s.stream_type(), r.stream_type());` should be enough here.  This is not a problem we would need to check at runtime in the general case.  If this has ever happened to you, then a runtime (Internal) error might be OK, but I am guessing that we can't get into this state.

neqo-http3/src/connection_client.rs
```diff
@@ -357,12 +344,10 @@ impl Http3Client {
     /// Both sides, sending and receiving side, will be closed.
     /// # Errors
     /// An error will be return if a stream does not exist.
-    pub fn stream_reset(&mut self, stream_id: u64, error: AppError) -> Res<()> {
-        qinfo!([self], "reset_stream {} error={}.", stream_id, error);
+    pub fn cancel_http_request(&mut self, stream_id: u64, error: AppError) -> Res<()> {
+        qinfo!([self], "reset_:stream {} error={}.", stream_id, error);
```
martinthomson, line 348:
```suggestion
        qinfo!([self], "reset_stream {} error={}.", stream_id, error);
```

neqo-http3/src/lib.rs
```diff
@@ -369,15 +368,14 @@ pub trait HttpSendStream: SendStream {
...
-    fn remove_send_side_event(&self, _stream_id: u64) {}
-    fn done(&self, _stream_id: u64) {}
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum ResetType {
-    App,
-    Remote,
-    Local,
+pub enum CloseType {
```
martinthomson, line 376: Some documentation would be helpful here.

martinthomson (commented):
Seems good.  Fix and land.

neqo-http3/src/client_events.rs
ddragana: That does not work because it needs to be called if type is ResetApp but not if it is Done. Both are exited in that match statement. It is not perfect. I will move the call to match statement.

neqo-http3/src/connection.rs
ddragana: This should happen very rarely.
This code will be touch when fixing Issue #261 (I will put the issue on my list to fix.).

neqo-http3/src/connection.rs
ddragana: No we cannot get into this state at all.

Issue #1239, opened 2021-09-29
Crash when running in FIPS-mode


msirringhaus:
We are currently having the following problem with FF91esr: When run in FIPS-mode, Firefox crashes in neqo.
As it turns out the call to `hkdf::import_key(TLS_VERSION_1_3, cipher, salt)` in `new_initial()` in `third_party/rust/neqo-transport/src/crypto.rs` fails, because it tries to create a key from raw key material, which is not allowed in FIPS-mode.

msmeissn:
It does PK_ImportSymKey to store key material, but a raw key import is not allowed in NSS in FIPS mode.

This seems just to do key derivation if I see this correctly, the key is discarded after the hdkf::expand without any further use.
perhaps this could be done without involving temporary keys?

martinthomson:
Confirmed as working.

PR #1247, opened 2021-10-01
Use PK11_ImportDataKey for FIPS safety

martinthomson:
When FIPS is enabled PK11_ImportSymKey can fail.

Note that I had to change the pre-commit hook here as the stash
invocation touches files.

For #1239. I still need to verify that this fixes the issue completely, but it can't make things worse.


ddragana (approved):

martinthomson:
Is it too late to get this into the next Firefox build?

ddragana:
No, I have not submitted a patch yet. I will update neqo version. Sorry, I should have asked before creating version 0.5.2.

PR #1254, opened 2021-10-15
fix panic from slice::from_raw_parts

12101111:
Sometime  `sct_ptr` is not null, but `sct_ptr.data` is null. In this case, use `slice::from_raw_parts` is UB and violate [safety guarantee](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) (In debug build, this will hit a `debug_assert` failure).


martinthomson (approved):
Thanks for catching this.  It's a common problem in NSS that is easy to miss.

PR #1259, opened 2021-11-03
Convert API to use StreamId

### LINKED ISSUE #346: StreamId should be in the public interface

martinthomson:
This makes basically all uses of u64 for stream IDs use the wrapper type.

This will affect gecko, but I hope not that much.  I've been meaning to do this for a while and it wasn't particularly hard or time-consuming.

Closes #346.


ddragana (dismissed):
Thanks.
I meant to do this as well. I started to use StreamId for WebTransport  already.

PR #1262, opened 2021-11-14
Improve client

heftig:
Improvements to the client made in the process of trying to debug something.

- **Create directory for client's output file**
- **Support connecting only to IPv4 or IPv6**

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -26,7 +26,7 @@ use neqo_transport::{
 use std::cell::RefCell;
 use std::collections::{HashMap, VecDeque};
 use std::fmt::{self, Display};
-use std::fs::{File, OpenOptions};
+use std::fs::{create_dir_all, File, OpenOptions};
 use std::io::{self, ErrorKind, Write};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, ToSocketAddrs, UdpSocket};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


neqo-client/src/main.rs
```diff
@@ -702,8 +711,22 @@ fn main() -> Res<()> {
...
-        let remote_addr = *addrs.first().unwrap();
+        let addrs: Vec<_> = format!("{}:{}", host, port)
+            .to_socket_addrs()?
+            .filter(|addr| {
+                !matches!(
+                    (addr, args.ipv4_only, args.ipv6_only),
+                    (SocketAddr::V4(..), false, true) | (SocketAddr::V6(..), true, false)
+                )
+            })
+            .collect();
```
martinthomson, line 722: You shouldn't need to collect() just to get the first() value.

```rust
        let remote_addr = format!("{}:{}", host, port)
            .to_socket_addrs()?
            .filter(|addr| {
                !matches!(
                    (addr, args.ipv4_only, args.ipv6_only),
                    (SocketAddr::V4(..), false, true) | (SocketAddr::V6(..), true, false)
                )
            })
            .first().expect(format!("No compatible address found for {}", host));
```
(That's a panic instead of eprintln!() and exit(), but you get the idea.  You can use `let remote_addr = if let Some(a) = ... .first() { *a } { eprintln!( ... ); exit(1); }` if you prefer.)

martinthomson (commented):
Thanks for this, this is a good improvement!  Just one nit.

neqo-client/src/main.rs
heftig: Replaced with `find`. Kept the `match` separate as it's more readable.

martinthomson (approved):

PR #1269, opened 2021-11-18
SendMessage refactor

ddragana:
There are to change sets.
1) encode_header_block does not need to return an error:
This will simplify SendMessage code. Any error during the encode_header_block call will be ignored and they will be picked up by the main loop, i.e. process() calls.

2) Encode headers when they are set:
The function encode_header_block does not produce any errors that need to be picked up by the main processing loop. This was change in the previous change set. Therefore the encoding can be performed immediately.

This PR will change current behavior regarding the order in which the encoder instructions and header blocks are sent.

### DIFF
--- a/neqo-http3/src/send_message.rs
+++ b/neqo-http3/src/send_message.rs
@@ -118,13 +118,13 @@ impl SendMessage {
     /// # Errors
     /// `ClosedCriticalStream` if the encoder stream is closed.
     /// `InternalError` if an unexpected error occurred.
-    fn ensure_encoded(&mut self, conn: &mut Connection) -> Res<()> {
+    fn ensure_encoded(&mut self, conn: &mut Connection) {
         if let SendMessageState::Initialized { headers, data, fin } = &self.state {
             qdebug!([self], "Encoding headers");
             let header_block =
                 self.encoder
                     .borrow_mut()
-                    .encode_header_block(conn, headers, self.stream_id)?;
+                    .encode_header_block(conn, headers, self.stream_id);
             let hframe = HFrame::Headers {
                 header_block: header_block.to_vec(),
             };
@@ -144,7 +144,6 @@ impl SendMessage {
                 fin: *fin,
             };
         }
-        Ok(())
     }
 }
 
@@ -231,7 +230,7 @@ impl SendStream for SendMessage {
     /// `TransportStreamDoesNotExist` if the transport stream does not exist (this may happen if `process_output`
     /// has not been called when needed, and HTTP3 layer has not picked up the info that the stream has been closed.)
     fn send(&mut self, conn: &mut Connection) -> Res<()> {
-        self.ensure_encoded(conn)?;
+        self.ensure_encoded(conn);
 
         let label = if ::log::log_enabled!(::log::Level::Debug) {
             format!("{}", self)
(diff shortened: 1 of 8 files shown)


neqo-http3/src/send_message.rs
```diff
@@ -118,33 +104,30 @@ impl SendMessage {
...
+        conn: &mut Connection,
+        stream_id: StreamId,
+    ) -> Vec<u8> {
+        qdebug!("Encoding headers");
+        let header_block = encoder.encode_header_block(conn, headers, stream_id);
+        let hframe = HFrame::Headers {
+            header_block: header_block.to_vec(),
+        };
+        let mut d = Encoder::default();
+        hframe.encode(&mut d);
```
martinthomson, line 120: At some point we should probably try to avoid this bit where we make a vector, put stuff in it, then make another vector and put the other vector in that at an almost fixed offset.  That's unnecessary work.

Our encoder has a neat `encode_vvec_with` function that we could use for encode_header_block so that we encode the frame just once.

neqo-qpack/src/encoder.rs
```diff
@@ -348,20 +348,24 @@ impl QPackEncoder {
         conn: &mut Connection,
         h: &[Header],
         stream_id: StreamId,
-    ) -> Res<HeaderEncoder> {
+    ) -> HeaderEncoder {
         qdebug!([self], "encoding headers.");
 
         let mut encoder_blocked = false;
         // Try to send capacity instructions if present.
         match self.send(conn) {
```
martinthomson, line 356: I think that you can just do this:

```rust
let mut encoder_blocked = self.send(conn).is_err();
```

neqo-qpack/src/encoder.rs
```diff
@@ -404,21 +408,24 @@ impl QPackEncoder {
...
+                } else {
+                    // The errors can be:
+                    //   1) `EncoderStreamBlocked` - this is an error that
+                    //      can occur.
+                    //   2) `DynamicTableFull` - this is an error that
+                    //      can occur.
+                    //   3) `InternalError` - this is unexpected error.
+                    //   4) `ClosedCriticalStream` - this is error that should
+                    //      close the HTTP/3 session.
+                    // The last 2 errors are ignored here and will be pcked up
```
martinthomson, line 424:
```suggestion
                    // The last 2 errors are ignored here and will be picked up
```

I think that the point is that other code is responsible for closing the connection in response to those errors.  This code doesn't try to deal with errors, it just tries to write to the encoder stream AND if it can't uses literal instructions.  It is thus infallible in the sense that it doesn't care about encoder stream errors.

neqo-qpack/src/encoder.rs
```diff
@@ -348,20 +348,24 @@ impl QPackEncoder {
...
             Ok(()) => {}
-            Err(Error::EncoderStreamBlocked) => {
+            Err(_) => {
+                // The errors can be:
+                //   1) `EncoderStreamBlocked` - this is an error that
+                //      can occur.
+                //   2) `InternalError` - this is unexpected error.
+                //   3) `ClosedCriticalStream` - this is error that should
+                //      close the HTTP/3 session.
+                // The last 2 errors are ignored here and will be pcked up
```
martinthomson, line 365:
```suggestion
                // The last 2 errors are ignored here and will be picked up
```

neqo-qpack/src/encoder.rs
martinthomson, line 356: Also, I think that you want to rename this "send" function to something more appropriate: update_encoder_stream() or send_encoder_updates() spring to mind as options.  I like terse names, but this is a little too terse.

martinthomson (dismissed):
This is a good refactoring.  Making the header block encoding infallible is a good move and removing the state is very much cleaner.  Comments on style only.

PR #1270, opened 2021-11-18
Implement sending of 1xx responses and refactor SendMessage

ddragana:
The biggest change is that SendMessage uses BufferedStream and does not need states. The header blocks are added using send_headers and they are buffered into the BufferedStream. There may be multiple header blocks, but the code does not enforce or check that only server can send multiple headers blocks or that the multiple 1xx header blocks are followed by a non-1xx header block. This is left for the application to enforce.

The server and client API are unified, both sides use send_header, send_data  and close_send functions.

### DIFF
--- a/neqo-common/src/headers.rs
+++ b/neqo-common/src/headers.rs
@@ -1,0 +1,153 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use crate::{Error, Header, MessageType, Res};
+use std::ops::Deref;
+use std::ops::DerefMut;
+
+const PSEUDO_HEADER_STATUS: u8 = 0x1;
+const PSEUDO_HEADER_METHOD: u8 = 0x2;
+const PSEUDO_HEADER_SCHEME: u8 = 0x4;
+const PSEUDO_HEADER_AUTHORITY: u8 = 0x8;
+const PSEUDO_HEADER_PATH: u8 = 0x10;
+const REGULAR_HEADER: u8 = 0x80;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Headers {
+    headers: Vec<Header>,
+}
+
+impl Default for Headers {
+    fn default() -> Self {
+        Self {
+            headers: Vec::new(),
+        }
+    }
+}
+
+impl Headers {
+    pub fn new(headers: &[Header]) -> Self {
+        Self {
+            headers: headers.to_vec(),
+        }
+    }
+
+    pub fn is_interim(&self) -> Res<bool> {
+        let status = self.headers.iter().find(|h| h.name() == ":status");
+        if let Some(h) = status {
+            #[allow(clippy::map_err_ignore)]
+            let status_code = h.value().parse::<i32>().map_err(|_| Error::InvalidHeader)?;
+            Ok((100..200).contains(&status_code))
+        } else {
+            Err(Error::InvalidHeader)
+        }
+    }
+
+    fn track_pseudo(name: &str, state: &mut u8, message_type: MessageType) -> Res<bool> {
+        let (pseudo, bit) = if name.starts_with(':') {
+            if *state & REGULAR_HEADER != 0 {
+                return Err(Error::InvalidHeader);
+            }
+            let bit = match message_type {
+                MessageType::Response => match name {
+                    ":status" => PSEUDO_HEADER_STATUS,
+                    _ => return Err(Error::InvalidHeader),
+                },
+                MessageType::Request => match name {
+                    ":method" => PSEUDO_HEADER_METHOD,
+                    ":scheme" => PSEUDO_HEADER_SCHEME,
+                    ":authority" => PSEUDO_HEADER_AUTHORITY,
+                    ":path" => PSEUDO_HEADER_PATH,
+                    _ => return Err(Error::InvalidHeader),
+                },
+            };
+            (true, bit)
+        } else {
+            (false, REGULAR_HEADER)
+        };
+
+        if *state & bit == 0 || !pseudo {
+            *state |= bit;
+            Ok(pseudo)
+        } else {
+            Err(Error::InvalidHeader)
+        }
+    }
+
+    pub fn headers_valid(&self, message_type: MessageType) -> Res<()> {
+        let mut method_value: Option<&str> = None;
+        let mut pseudo_state = 0;
+        for header in &self.headers {
+            let is_pseudo = Self::track_pseudo(header.name(), &mut pseudo_state, message_type)?;
+
+            let mut bytes = header.name().bytes();
+            if is_pseudo {
+                if header.name() == ":method" {
+                    method_value = Some(header.value());
+                }
+                let _ = bytes.next();
+            }
+
+            if bytes.any(|b| matches!(b, 0 | 0x10 | 0x13 | 0x3a | 0x41..=0x5a)) {
+                return Err(Error::InvalidHeader); // illegal characters.
+            }
+        }
+        // Clear the regular header bit, since we only check pseudo headers below.
+        pseudo_state &= !REGULAR_HEADER;
+        let pseudo_header_mask = match message_type {
+            MessageType::Response => PSEUDO_HEADER_STATUS,
+            MessageType::Request => {
+                if method_value == Some(&"CONNECT".to_string()) {
+                    PSEUDO_HEADER_METHOD | PSEUDO_HEADER_AUTHORITY
+                } else {
+                    PSEUDO_HEADER_METHOD | PSEUDO_HEADER_SCHEME | PSEUDO_HEADER_PATH
+                }
+            }
+        };
+        if pseudo_state & pseudo_header_mask != pseudo_header_mask {
+            return Err(Error::InvalidHeader);
+        }
+
+        Ok(())
+    }
+
+    pub fn retain_valid_for_response(&mut self) {
+        self.headers.retain(Header::is_allowed_for_response);
+    }
+}
+
+impl From<&[Header]> for Headers {
+    fn from(h: &[Header]) -> Self {
+        Headers::new(h)
+    }
+}
+
+impl From<Vec<Header>> for Headers {
+    fn from(headers: Vec<Header>) -> Self {
+        Self { headers }
+    }
+}
+
+impl DerefMut for Headers {
+    #[must_use]
+    fn deref_mut(&mut self) -> &mut Vec<Header> {
+        &mut self.headers
+    }
+}
+
+impl Deref for Headers {
+    type Target = Vec<Header>;
+    #[must_use]
+    fn deref(&self) -> &Self::Target {
+        &self.headers
+    }
+}
+
+impl AsRef<[Header]> for Headers {
+    fn as_ref(&self) -> &[Header] {
+        &self.headers
+    }
+}
--- a/neqo-common/src/headers.rs
+++ b/neqo-common/src/headers.rs
@@ -41,7 +41,7 @@ impl Headers {
     /// Returns an error if response headers do not contain
     /// a status header or if the value of the header cannot be parsed.
     pub fn is_interim(&self) -> Res<bool> {
-        let status = self.headers.iter().find(|h| h.name() == ":status");
+        let status = self.headers.iter().take(1).find(|h| h.name() == ":status");
         if let Some(h) = status {
             #[allow(clippy::map_err_ignore)]
             let status_code = h.value().parse::<i32>().map_err(|_| Error::InvalidHeader)?;
@@ -56,18 +56,13 @@ impl Headers {
             if *state & REGULAR_HEADER != 0 {
                 return Err(Error::InvalidHeader);
             }
-            let bit = match message_type {
-                MessageType::Response => match name {
-                    ":status" => PSEUDO_HEADER_STATUS,
-                    _ => return Err(Error::InvalidHeader),
-                },
-                MessageType::Request => match name {
-                    ":method" => PSEUDO_HEADER_METHOD,
-                    ":scheme" => PSEUDO_HEADER_SCHEME,
-                    ":authority" => PSEUDO_HEADER_AUTHORITY,
-                    ":path" => PSEUDO_HEADER_PATH,
-                    _ => return Err(Error::InvalidHeader),
-                },
+            let bit = match (message_type, name) {
+                (MessageType::Response, ":status") => PSEUDO_HEADER_STATUS,
+                (MessageType::Request, ":method") => PSEUDO_HEADER_METHOD,
+                (MessageType::Request, ":scheme") => PSEUDO_HEADER_SCHEME,
+                (MessageType::Request, ":authority") => PSEUDO_HEADER_AUTHORITY,
+                (MessageType::Request, ":path") => PSEUDO_HEADER_PATH,
+                (_, _) => return Err(Error::InvalidHeader),
             };
             (true, bit)
         } else {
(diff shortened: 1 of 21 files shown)


ddragana:
This is based on #1269 

This needs a follow up to improve neqo-server. set_response use to guaranty to take all response data, but send_data does not, so the server needs to account for that.

neqo-http3/src/connection_client.rs
```diff
@@ -292,14 +296,21 @@ impl Http3Client {
         }
         final_headers.extend_from_slice(headers);
 
+        let mut send_message = SendMessage::new(
+            id,
+            self.base_handler.qpack_encoder.clone(),
+            Box::new(self.events.clone()),
+        );
+
+        // SendMessage implements http_stream so ith will not panic.
```
martinthomson, line 305:
```suggestion
```
Once is enough.

neqo-http3/src/lib.rs
```diff
@@ -376,14 +377,10 @@ pub trait SendStream: Stream {
 }
 
 pub trait HttpSendStream: SendStream {
+    /// This function sets informational response.
```
martinthomson, line 380: It sends final response header blocks also, right?  And, from what I can see, trailers and middlers (see my comment about that below).

neqo-http3/src/send_message.rs
```diff
@@ -213,37 +150,20 @@ impl SendStream for SendMessage {
...
+            if self.fin {
+                Error::map_error(
+                    conn.stream_close_send(self.stream_id()),
+                    Error::HttpInternal(6),
+                )?;
+                qtrace!([self], "done sending request");
             } else {
-                let b = buf.split_off(sent);
-                *buf = b;
+                self.conn_events.data_writable(self.stream_id());
```
martinthomson, line 166: Why don't we call self.stream_writable() here?

There's a corner case here that I'm not sure about: if we buffer the headers and that takes *all* the remaining flow control window for the stream, we'll get to this point and the stream will not be writable.  Do we need to ask the connection if there is space available?

neqo-http3/src/send_message.rs
```diff
@@ -253,25 +173,23 @@ impl SendStream for SendMessage {
...
+            conn.stream_close_send(self.stream_id())?;
         }
 
         self.conn_events
-            .send_closed(self.stream_id, CloseType::Done);
+            .send_closed(self.stream_id(), CloseType::Done);
         Ok(())
     }
 
     fn stop_sending(&mut self, close_type: CloseType) {
```
martinthomson, line 190: This function is called when we receive a STOP_SENDING right?  And we're relying on the transport to send the RESET_STREAM in response.  Reading it here in isolation I found the reaction a little confusing.  (The function could be read to be an instruction as in "please stop sending on this stream" as opposed to a reaction "you just got a STOP_SENDING")  A comment on the function might avoid that.  Renaming to handle_stop_sending() might be better.

neqo-http3/src/send_message.rs
```diff
@@ -138,71 +79,67 @@ impl Stream for SendMessage {
...
     fn done(&self) -> bool {
-        self.state.done()
+        !self.stream.has_buffered_data() && self.fin
     }
 
     fn stream_writable(&self) {
-        if self.state.is_state_sending_data() {
-            self.conn_events.data_writable(self.stream_id);
+        if !self.stream.has_buffered_data() && !self.fin {
+            self.conn_events.data_writable(self.stream_id());
```
martinthomson, line 142: See the comment below in "send()".

neqo-http3/src/server_events.rs
```diff
@@ -48,17 +48,32 @@ impl ClientRequestStream {
...
+    }
+
+    /// Supply a response header to a request.
+    /// # Errors
+    /// It may return `InvalidStreamId` if a stream does not exist anymore.
+    pub fn send_headers(&mut self, headers: &[Header]) -> Res<()> {
+        self.handler
+            .borrow_mut()
+            .send_headers(self.stream_id, headers, &mut self.conn.borrow_mut())
+    }
```
martinthomson, line 68: Maybe reorder these functions so that they follow the natural order:
1. send_headers
2. send_data
3. close_send

neqo-http3/src/send_message.rs
```diff
@@ -281,29 +199,19 @@ impl SendStream for SendMessage {
...
-    fn set_message(
-        &mut self,
-        headers: &[Header],
-        data: Option<&[u8]>,
-        conn: &mut Connection,
-    ) -> Res<()> {
-        if !matches!(self.state, SendMessageState::Uninitialized) {
-            return Err(Error::AlreadyInitialized);
-        }
+    fn send_headers(&mut self, headers: &[Header], conn: &mut Connection) {
```
martinthomson, line 202: This isn't doing anything to ensure that headers are injected at a sensible position in the stream.  I think that we want to do that so that we don't do allow people using this code to do silly things, either on purpose or by accident.

That is, we should validate that:

* there aren't multiple requests (with methods and so forth)
* there aren't multiple final status codes
* informational status codes don't follow final status codes
* requests and response blocks (those with methods or status codes) don't follow data
* there is just one block of trailers
* data doesn't follow trailers

That suggests a state machine of sorts.  There might be a point to specializing into RequestSendStream and ResponseSendStream to do that, but I'm not sure if that really helps.

I'm happy for that to exist at a higher level than this class or in a specialization of this class, but as we're basically exposing this function directly, we should put some safety rails on it.

neqo-http3/src/connection_client.rs
```diff
@@ -292,15 +296,21 @@ impl Http3Client {
...
+        send_message
+            .http_stream()
+            .unwrap()
```
martinthomson, line 308: Do you need these lines?  At a minimum you should be able to do `(send_message as HttpStream).send_headers()` as the thing you have is fine.

martinthomson (changes requested):
I like this new structure for SendStream.  You have a buffer of things you want to send that includes some bytes and maybe a FIN.  And you buffer headers using that.  But you don't buffer data, instead only passing what can be written directly to the underlying stream.

However, I think we need some guardrails on the use of the new breakouts.  That also means more tests than what you have here.

I've also identified what I think is a corner case that will need to be addressed and tested.

neqo-http3/src/lib.rs
ddragana: Yes, this is left over of the first design.

neqo-http3/src/send_message.rs
ddragana: This is fine. DataWritable is just a signal to try writing and if it fails it is fine. This can already happen in other ways, e.g. if session flow control opens up, all active stream will get this signal, but first streams that get the signal can eat up all the credits and later streams will not be able to write data.

neqo-http3/src/send_message.rs
ddragana: I will rename it to handle_stop_sending and add a comment.

neqo-http3/src/send_message.rs
ddragana: I purposely decided not to do that. The main reason was to easy testing the client side, by letting  server do silly things. I did not want to have asymmetry, i.e. client is strict and sever is not. The previous version had states that are appropriate for the client side(the states fit well the Firefox use case), but not really working for the server. 

I will try to implement checks and keep one SendMessage and if it too ugly I will split them.

neqo-http3/src/send_message.rs
martinthomson: A comment to that effect would be helpful.

neqo-http3/tests/send_message.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    assert_eq!(send_trailers(&mut request), Err(Error::InvalidInput));
+
+    request.close_send().unwrap();
+    exchange_packets(&mut hconn_c, &mut hconn_s);
+    process_client_events(&mut hconn_c);
+}
+
+#[test]
+fn data_after_tralier() {
```
martinthomson, line 219:
```suggestion
fn data_after_trailer() {
```

neqo-http3/tests/send_message.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    exchange_packets(&mut hconn_c, &mut hconn_s);
+    process_client_events(&mut hconn_c);
+}
+
+#[test]
+fn multiple_response_headers() {
+    let (mut hconn_c, mut hconn_s, mut request) = connect_send_and_receive_request();
+    request
+        .send_headers(&[Header::new(":status", "200"), Header::new("something", "3")])
```
martinthomson, line 251: Can you make this header block a constant?  RESPONSE_HEADERS or something.

neqo-http3/tests/send_message.rs
```diff
@@ -0,0 +1,291 @@
...
+        .unwrap();
+
+    assert_eq!(
+        request.send_headers(&[Header::new(":status", "200"), Header::new("something", "3"),]),
+        Err(Error::InvalidHeader)
+    );
```
martinthomson, line 257: Can you add another test with an information status code after a final status code?

neqo-http3/tests/send_message.rs
```diff
@@ -0,0 +1,291 @@
...
+fn data_before_headers() {
+    let (mut hconn_c, mut hconn_s, mut request) = connect_send_and_receive_request();
+    assert_eq!(request.send_data(RESPONSE_DATA), Err(Error::InvalidInput));
+
+    send_headers(&mut request).unwrap();
+    request.send_data(RESPONSE_DATA).unwrap();
+    request.close_send().unwrap();
+    exchange_packets(&mut hconn_c, &mut hconn_s);
+    process_client_events(&mut hconn_c);
+}
```
martinthomson, line 291: How about data after an informational status code?

neqo-http3/tests/send_message.rs
```diff
@@ -0,0 +1,291 @@
...
+    request.send_headers(&[
+        Header::new("something1", "something"),
+        Header::new("something2", "3"),
+    ])
+}
+
+fn send_headers(request: &mut ClientRequestStream) -> Res<()> {
+    request.send_headers(&[
+        Header::new(":status", "200"),
+        Header::new("content-length", "3"),
```
martinthomson, line 62: See below: this should probably be a constant somehow.  Maybe you need to use lazy_static!() because I don't think that you can make a new Vec (as this does) without that.

martinthomson (dismissed):
A few somewhat major concerns for which I don't want to block landing this (because it's good):

1. For *sending*, we don't need Headers to take ownership of the values.  That's inefficient.  For *receiving* we need ownership, so it's not easy. We will need to think a bit about how we might work that out.
2. I'm now of the opinion that we need a neqo-http crate for the headers piece, rather than dumping this all in neqo-common.
3. We'll need a plan for dealing with `:protocol` as a pseudo-header field in this validation piece.

Other than that, the tests here are solid and it is a worthwhile improvement overall.

ddragana:
> A few somewhat major concerns for which I don't want to block landing this (because it's good):
> 
>     1. For _sending_, we don't need Headers to take ownership of the values.  That's inefficient.  For _receiving_ we need ownership, so it's not easy. We will need to think a bit about how we might work that out.

This is the biggest problem here. I have not found an easy solution. The only other think I though doing is to remove Headers structure and make check functions not belong to any structure and take &[Header] as an argument. 

I think I will go back to that version, but make a separate PR.

> 
>     2. I'm now of the opinion that we need a neqo-http crate for the headers piece, rather than dumping this all in neqo-common.
> 
>     3. We'll need a plan for dealing with `:protocol` as a pseudo-header field in this validation piece.

This is already implemented in one of other patches for WebTransport (I need to double check with Kershaw if I can land them.)

> 
> 
> Other than that, the tests here are solid and it is a worthwhile improvement overall.

neqo-common/src/headers.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    /// Checks if request/response headers are well formed, i.e. contain
+    /// allowed pseudo headers and in a right order, etc.
+    /// # Errors
+    /// Returns an error if headers are not well formed.
+    pub fn headers_valid(&self, message_type: MessageType) -> Res<()> {
+        let mut method_value: Option<&str> = None;
+        let mut pseudo_state = 0;
```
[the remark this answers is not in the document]
ddragana: I will leave this fora follow up.

neqo-http3/src/connection_client.rs
ddragana: This does not work. I have tried it.

neqo-common/src/headers.rs
```diff
@@ -0,0 +1,175 @@
...
+                let _ = bytes.next();
+            }
+
+            if bytes.any(|b| matches!(b, 0 | 0x10 | 0x13 | 0x3a | 0x41..=0x5a)) {
+                return Err(Error::InvalidHeader); // illegal characters.
+            }
+        }
+        // Clear the regular header bit, since we only check pseudo headers below.
+        pseudo_state &= !REGULAR_HEADER;
+        let pseudo_header_mask = match message_type {
```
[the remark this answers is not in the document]
ddragana: I will do this in a follow up.

neqo-http3/src/send_message.rs
```diff
@@ -25,77 +25,102 @@ const MAX_DATA_HEADER_SIZE_3_LIMIT: usize = MAX_DATA_HEADER_SIZE_3 + 5; // 16383
...
-    pub fn is_sending_closed(&self) -> bool {
-        match self {
-            Self::SendingInitialMessage { fin, .. } => *fin,
-            Self::SendingData => false,
-            _ => true,
+impl MessageState {
+    fn new_headers(&mut self, headers: &Headers, message_type: MessageType) -> Res<()> {
+        match &self {
+            Self::WaitingForHeaders => {
+                debug_assert!(headers.headers_valid(message_type).is_ok());
```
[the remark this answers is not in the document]
ddragana: yes, because of that. I will add a comment.

neqo-http3/src/lib.rs
```diff
@@ -43,14 +43,14 @@ pub use conn_params::Http3Parameters;
...
-pub use neqo_common::Header;
+pub use neqo_common::{Error as CommonError, Header, Headers, MessageType};
 pub use priority::Priority;
 pub use server::Http3Server;
 pub use server_events::{ClientRequestStream, Http3ServerEvent};
 pub use settings::HttpZeroRttChecker;
 pub use stream_type_reader::NewStreamType;
 
-type Res<T> = Result<T, Error>;
+pub type Res<T> = Result<T, Error>;
```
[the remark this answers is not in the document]
ddragana: This is needed for test, I will use Result<..> directly.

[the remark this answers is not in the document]
ddragana: The tests in ./neqo-http3/tests/ directory.

neqo-http3/src/connection_server.rs
```diff
@@ -40,20 +41,45 @@ impl Http3ServerHandler {
...
             .ok_or(Error::InvalidStreamId)?
             .http_stream()
             .ok_or(Error::InvalidStreamId)?
-            .set_message(headers, Some(data), conn)?;
+            .send_headers(Headers::from(headers), conn)?;
+        self.base_handler.stream_has_pending_data(stream_id);
+        Ok(())
+    }
+
+    /// This is call when application is done sending a request.
```
KershawChang, line 77: nit: `called`

neqo-http3/src/connection_server.rs
```diff
@@ -40,20 +41,45 @@ impl Http3ServerHandler {
...
+    /// # Errors
+    /// An error will be return if stream does not exist.
```
KershawChang, line 79: nit: `returned`

neqo-http3/src/lib.rs
```diff
@@ -369,21 +378,21 @@ pub trait SendStream: Stream {
     /// # Errors
     /// It may happen that the transport stream is already close. This is unlikely.
     fn close(&mut self, conn: &mut Connection) -> Res<()>;
-    fn stop_sending(&mut self, close_type: CloseType);
+    /// This function is called when sending side is closed abruptly by the pear or
```
KershawChang, line 381: nit: `peer`

KershawChang (dismissed):
This looks good to me. I only have some nits.

PR #1275, opened 2021-11-24
Remove the Headers structure

ddragana:


neqo-common/src/lib.rs
```diff
@@ -4,14 +4,13 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![cfg_attr(feature = "deny-warnings", deny(warnings))]
-#![warn(clippy::pedantic)]
+//#![cfg_attr(feature = "deny-warnings", deny(warnings))]
+//#![warn(clippy::pedantic)]
```
martinthomson, line 8: ?

martinthomson (commented):
WFM

neqo-common/src/lib.rs
ddragana: My clippy is making problems with neqo-common. I need to fix it, but meanwhile I run it locally with disabled clippy in neqo-common. Forgot to remove this.

PR #1277, opened 2021-11-29
Add the DataWritable event to the server side

### LINKED ISSUE #1251: Have Http3Parameters include ConnectionParameters
Any sense in having Http3Parameters include ConnectionParameters?  You could make it impl DerefMut<Target=ConnectionParameters> so that you can use the same interface to manage everything.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1250#discussion_r723761738_

ddragana:
PR #1270 refactored the server side API. Instead of set_response that used to accept all data supply, now there are 3 functions: send_headers, send_data and stream_close_send. Important change is also that send_data does not accept more data than the flow control allows. To be able to write larger responses DataWritable events are needed to to trigger writes when more flow control credit is available.

Additional changes:
To write a test for the event it was necessary to change transport level parameters when creating a Http3Server. To avoid hitting the limit of the number of parameters in a function, ConnectionParameters are added to Http3Parameters. This will fix #1251.

### DIFF
--- a/neqo-http3/tests/httpconn.rs
+++ b/neqo-http3/tests/httpconn.rs
@@ -221,6 +221,7 @@ fn test_data_writable_events() {
 
     mem::drop(connect_peers(&mut hconn_c, &mut hconn_s));
 
+    // Create a request.
     let req = hconn_c
         .fetch(
             now(),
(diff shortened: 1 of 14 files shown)


neqo-http3/tests/webtransport/mod.rs
```diff
@@ -456,23 +455,29 @@ impl WtTest {
...
+        let stream_reset = |e| {
+            matches!(
+                e,
+                Http3ServerEvent::StreamReset {
+                    stream,
+                    error
+                } if stream.stream_id() == expected_stream_id && error == expected_error
+            )
+        };
+        assert!(self.server.events().any(stream_reset));
```
KershawChang, line 467: Can we find the exact event instead of using `any`?

neqo-http3/tests/httpconn.rs
```diff
@@ -191,3 +208,58 @@ fn test_103_response() {
...
+            Http3ServerEvent::DataWritable {
+                stream
+            } if stream.stream_id() == request.stream_id()
+        )
+    };
+    assert!(hconn_s.events().any(data_writable));
+
+    // Data can be sent again.
+    let sent2 = request.send_data(&buf[sent1..]).unwrap();
+    assert!(sent2 > 0);
```
KershawChang, line 264: Should we test if all data are sent out?

KershawChang (approved):
I only have two comments about tests, but I think we can still land this.

neqo-http3/tests/webtransport/mod.rs
ddragana: This will fine the exact event with expected steam_id(expected_stream_id) and expected_error.

PR #1281, opened 2021-12-02
A WebTransport session should be rejected with 404

ddragana:


neqo-http3/src/connection.rs
```diff
@@ -854,7 +854,7 @@ impl Http3Connection {
             (Some(s), Some(_r), false) => {
                 if s.http_stream()
                     .ok_or(Error::InvalidStreamId)?
-                    .send_headers(&[Header::new(":status", "400")], conn)
+                    .send_headers(&[Header::new(":status", "404")], conn)
```
martinthomson, line 857: No test for this?

ddragana: We currently do not propagate the response code. Therefore it is not possible to test this.

We should propagate the response code. I will submit an update.

KershawChang (approved):

PR #1297, opened 2021-12-16
Make FrameReader generic so that it is possible to use it for WebTransport frames

ddragana:


ddragana:
This is base on top of #1281 . The last to change set are new.

KershawChang (approved):

Issue #1286, opened 2021-12-02
Closing WebTransport cleanly


ddragana:
This is new in draft-02. 
A WebTranspor session can be closed by sending CLOSE_WEBTRANSPORT_SESSION frame (capsule) on the CONNECT stream.

> Upon learning that the session has been terminated, the endpoint MUST
   reset all of the streams associated with the session; it MUST NOT
   send any new datagrams or open any new streams.

...

> A CLOSE_WEBTRANSPORT_SESSION capsule MUST be followed by a FIN on the
   sender side.  If any additional stream data is received on the
   CONNECT stream after CLOSE_WEBTRANSPORT_SESSION, the stream MUST be
   reset with code H3_MESSAGE_ERROR.  The recipient MUST close the
   stream upon receiving a FIN.  If the sender of
   CLOSE_WEBTRANSPORT_SESSION does not receive a FIN after some time, it
   SHOULD send STOP_SENDING on the CONNECT stream.

> Cleanly terminating a CONNECT stream without a
   CLOSE_WEBTRANSPORT_SESSION capsule SHALL be semantically equivalent
   to terminating it with a CLOSE_WEBTRANSPORT_SESSION capsule that has
   an error code of 0 and an empty error string.

ddragana:
Fixed by #1300

PR #1300, opened 2021-12-19
Send and receive CloseSession frame

ddragana:


ddragana:
I want to explore a different architecture for WebTransport CONNECT stream.

I do not like that Http3Connection (connection.rs) needs to know about WebTransport CONNECT streams to be able to manage them. 

I am planing to try creating a special WebTransport stream that holds RecvMessage and SendMessage. With this design the WebTransport stream would be put into the Http3Connection's send_streams and recv_streams and they will manage their own live time. so we do not need a hashmap of WebTransport sessions in ExtendedConnectFeature. Also Http3Connection does not need a special handling for WebTransport streams.

ddragana:
One more think that is missing is that CloseSession frame needs to be send whole. The current implementation does not guarantee that.

ddragana:
This also needs more tests.

ddragana:
This fixes #1286

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -0,0 +1,513 @@
...
+            frame_reader: FrameReader::new(),
+            events,
+            send_streams: BTreeSet::new(),
+            recv_streams: BTreeSet::new(),
+            role,
+        }
+    }
+
+    /// # Errors
+    /// The above function can only fail if supplied headers are not valid http headers.
```
KershawChang, line 137: nit: which `above` function? I think you mean `send_request`, right?

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -0,0 +1,513 @@
...
+    fn has_data_to_send(&self) -> bool {
+        self.control_stream_send.has_data_to_send()
+    }
+
+    fn done(&self) -> bool {
+        self.state == SessionState::Done
+    }
+
+    fn handle_stop_sending(&mut self, close_type: CloseType) {
+        self.close(close_type);
```
KershawChang, line 211: This function does exactly the same as `reset`. Maybe we can remove one of them.

neqo-http3/src/connection_client.rs
```diff
@@ -436,6 +435,22 @@ impl Http3Client {
         output
     }
 
+    /// Close `WebTransport` cleanly
+    /// # Errors
+    /// `InvalidStreamId` if thee stream does not exist,
```
KershawChang, line 440: nit: `the`

KershawChang (approved):
I have some nits and minor comments, but I think this is good to land.

neqo-http3/src/recv_message.rs
```diff
@@ -468,15 +473,30 @@ impl HttpRecvStream for RecvMessage {
...
+    fn priority_update_frame(&mut self) -> Option<HFrame> {
+        self.priority_handler.maybe_encode_frame(self.stream_id)
+    }
+
+    fn priority_update_sent(&mut self) {
+        self.priority_handler.priority_update_sent();
     }
 
     fn set_new_listener(&mut self, conn_events: Box<dyn HttpRecvStreamEvents>) {
+        self.state = RecvMessageState::WaitingForData {
```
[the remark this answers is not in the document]
ddragana: I am planning to refactor this a bit. My idea is: when the code figures out that this is a CONNECT request it should transform it into a ExtendedConnect stream. This is not easy to do in rust therefore I have not doe int in this change.

I will leave it for now.

neqo-http3/src/send_message.rs
```diff
@@ -306,8 +320,13 @@ impl HttpSendStream for SendMessage {
     }
 
     fn set_new_listener(&mut self, conn_events: Box<dyn SendStreamEvents>) {
+        self.stream_type = Http3StreamType::ExtendedConnect;
```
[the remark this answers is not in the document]
ddragana: This  will part of the same refactoring as above.

neqo-http3/src/features/extended_connect/webtransport_session.rs
ddragana: Nice catch, both of them can be remove. They only call close().

PR #1302, opened 2021-12-29
Update preferred address config for servers

martinthomson:
This was still using an old method of configuration, which meant that it
didn't work.


martinthomson:
CI failure can't be reproduced locally, see #1304.

ddragana (approved):

PR #1309, opened 2022-01-05
reset stream when receiving 101

### LINKED ISSUE #1017: Treat response code 101 as an error and reset the stream

KershawChang:
This will fix #1017

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -6233,7 +6233,7 @@ mod tests {
         let push_stream_id = server.conn.stream_create(StreamType::UniDi).unwrap();
 
         let mut d = Encoder::default();
-        let headers1xx: &[Header] = &[Header::new(":status", "101")];
+        let headers1xx: &[Header] = &[Header::new(":status", "100")];
         server.encode_headers(push_stream_id, headers1xx, &mut d);
 
         let headers200: &[Header] = &[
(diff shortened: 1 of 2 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -6728,4 +6728,42 @@ mod tests {
...
     }
+
+    #[test]
+    fn response_w_101() {
+        let (mut client, mut server, request_stream_id) = connect_and_send_request(true);
+
+        setup_server_side_encoder(&mut client, &mut server);
+
+        let mut d = Encoder::default();
+        let headers1xx: &[Header] = &[Header::new(":status", "101")];
```
martinthomson, line 6739:
```suggestion
        let headers1xx = &[Header::new(":status", "101")];
```

neqo-http3/src/connection_client.rs
```diff
@@ -6728,4 +6728,42 @@ mod tests {
...
+        server.encode_headers(request_stream_id, headers1xx, &mut d);
+
+        let headers200: &[Header] = &[
```
martinthomson, line 6742:
```suggestion
        let headers200 = &[
```

martinthomson: Do you even need a 200 response?

neqo-http3/src/headers_checks.rs
```diff
@@ -48,12 +48,12 @@ impl TryFrom<(MessageType, &str)> for PseudoHeaderState {
...
 /// Returns an error if response headers do not contain
 /// a status header or if the value of the header cannot be parsed.
-pub fn is_interim(headers: &[Header]) -> Res<bool> {
+pub fn is_interim(headers: &[Header]) -> Res<(bool, i32)> {
     let status = headers.iter().take(1).find(|h| h.name() == ":status");
     if let Some(h) = status {
         #[allow(clippy::map_err_ignore)]
         let status_code = h.value().parse::<i32>().map_err(|_| Error::InvalidHeader)?;
-        Ok((100..200).contains(&status_code))
+        Ok(((100..200).contains(&status_code), status_code))
```
martinthomson, line 56: Would it be possible to do this instead?

```suggestion
        if status_code == 101 {
            Err(Error::InvalidHeader)
        } else {
            Ok((100..200).contains(&status_code))
        }
```

That would integrate better into existing code.  It means generating an error when sending a message, but I think that's good too.

KershawChang: Right, doing this means less change for the current code.  Thanks.

neqo-http3/src/connection_client.rs
KershawChang: Obviously, no. I'll remove this.

ddragana (approved):

PR #1310, opened 2022-01-05
Generate ConnectionEvent::SendStreamCreatable at startup

### LINKED ISSUE #842: Generate ConnectionEvent::SendStreamCreatable at startup
When a new stream is created, there are no new events that indicate that a stream can be created; the event is only generated when the stream limit increases.  This event should be generated whenever the stream limit is set so that using code can drive stream creation off the event alone.

This is probably the right place to do that:

https://github.com/mozilla/neqo/blob/37374c4232fb25304a30127380f70c0e66a1f4cf/neqo-transport/src/connection.rs#L1735-L1738

KershawChang:
To fix #842

### DIFF
--- a/neqo-transport/src/streams.rs
+++ b/neqo-transport/src/streams.rs
@@ -467,6 +467,13 @@ impl Streams {
                 .remote()
                 .get_integer(tparams::INITIAL_MAX_DATA),
         );
+
+        if self.local_stream_limits[StreamType::BiDi].available() > 0 {
+            self.events.send_stream_creatable(StreamType::BiDi);
+        }
+        if self.local_stream_limits[StreamType::UniDi].available() > 0 {
+            self.events.send_stream_creatable(StreamType::UniDi);
+        }
     }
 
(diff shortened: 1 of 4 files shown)


neqo-transport/src/streams.rs
```diff
@@ -467,6 +467,9 @@ impl Streams {
                 .remote()
                 .get_integer(tparams::INITIAL_MAX_DATA),
         );
+
+        self.events.send_stream_creatable(StreamType::BiDi);
+        self.events.send_stream_creatable(StreamType::UniDi);
```
martinthomson, line 472: This can't be done unconditionally as you might not be able to create these streams if the limit is zero.  Check that the transport parameter is non-zero here.

Given that, you should create a test case (in neqo-transport) that tests for these events not being generated when the initial values are 0.

I'd also like to see a test in neqo-transport for the case where the stream limits are non-zero, rather than tacking an assertion onto something in neqo-http3.

martinthomson (changes requested):

ddragana (approved):

PR #1315, opened 2022-01-11
zero-length unknown frame payloads

martinthomson:
Don't try to read the payload of empty frames, it doesn't work.

The real fix is in neqo-http3/src/frames/reader.rs

This includes a few other changes:

* I made the incremental ignoring decoder not work for zero-length
things.  This is the source of the problem.  Fixing this doesn't help in
this case, but it should block similar mistakes.

* Enable RUST_LOG for running in CI.  The test runner now captures
output, so we are able to get per-test logs when they fail.

* Use || instead of | for a boolean condition.

* Fixed a typo in comments/logs.

* Improved exchange_packets.

* Streamlined the event handling in priority_update.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -92,6 +92,7 @@ jobs:
         env:
            LD_LIBRARY_PATH: ${{ github.workspace }}/dist/Debug/lib
            RUST_BACKTRACE: 1
+           RUST_LOG: neqo=debug
 
       - name: Check formatting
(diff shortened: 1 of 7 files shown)


KershawChang (approved):
Locally tested and I can't reproduce this issue any more with this patch.
Thanks.

ddragana (approved):
Thanks.
This is a regression of my changes that landed at the end of December. This is good, i.e. mozilla-central is not affected.

martinthomson:
Thanks for checking Dragana. Very happy that this didn't make it into deployment.

PR #1318, opened 2022-01-21
Use the mozbuild crate when building in gecko

glandium:
See [bug 1751331](https://bugzilla.mozilla.org/show_bug.cgi?id=1751331).

### DIFF
--- a/neqo-crypto/Cargo.toml
+++ b/neqo-crypto/Cargo.toml
@@ -15,6 +15,7 @@ bindgen = {version = "0.56", default-features = false, features= ["runtime"]}
 serde = "1.0"
 serde_derive = "1.0"
 toml = "0.4"
+mozbuild = {version = "0.1", optional = true}
 
 [dev-dependencies]
 test-fixture = { path = "../test-fixture" }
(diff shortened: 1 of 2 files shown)


glandium:
The clippy error comes from the fact the branch is derived from v0.5.6 rather than main, because I wanted it to be merged rather than rebased, so that we could update mozilla-central without pulling the changes since v0.5.6.

martinthomson (approved):
Thanks for this @glandium, this is a very nice change.  I can confirm the clippy error is not your fault (it's mine).

martinthomson:
Going YOLO on the merge to main.  @glandium, do you want to land a version with 0.5.6 plus your changes in m-c?  (Maybe @ddragana can help with the details as I just timed out for the week.)

glandium:
I was hoping you'd do an actual merge, such that I could just use the revision from this PR, but it's too late for that now...

glandium:
One possibility would be create a branch for f1cd349.

martinthomson:
@ddragana is in the best position to work that out.  If we're landing an 0.5.7 with new features soon, that might be easier.

ddragana:
Sorry for the delay. I will create 0.5.7 and landed it into m.-c. today.

PR #1319, opened 2022-01-21
Rust 1.58 has a nasty clippy false positive

martinthomson:


.github/workflows/check.yml
```diff
@@ -99,5 +99,5 @@ jobs:
         if: ${{ success() || failure() }}
 
       - name: Clippy
-        run: cargo +${{ matrix.rust-toolchain }} clippy -v --tests -- -D warnings
+        run: cargo +${{ matrix.rust-toolchain }} clippy -v --tests -- -D warnings $([ $(cargo version | cut -f2 -d" " -) = "1.58.0" ] && echo -A clippy::question_mark)
```
martinthomson, line 102: This right here is disgusting, so I hope we can remove it soon.

ddragana (approved):

Issue #1086, opened 2021-01-19
Remove InternalErrors added in #1085


ddragana:
Maybe we can remove all InternalErrors

Grubba27:
This issue still can be done ? is just removing the added erros made in this [pull request](https://github.com/mozilla/neqo/pull/1085/files) ?

PR #1323, opened 2022-01-27
chore: Removed InternalErrors added in #1085 

Grubba27:
I've just removed all of the internal errors as issue  #1086 don't know if we need to remove other code, the issue description was not clear to me. 
If is needed to make any changes I gladly would do it

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -293,7 +293,7 @@ impl QPackEncoder {
                     false,
                     "can_evict_to should have checked and make sure this operation is possible"
                 );
-                return Err(Error::InternalError(1));
+                return Err(Error::InternalError);
             }
             self.max_entries = cap / 32;
             self.next_capacity = None;
@@ -505,7 +505,7 @@ fn map_stream_send_atomic_error(err: &TransportError) -> Error {
         }
         _ => {
             debug_assert!(false, "Unexpected error");
-            Error::InternalError(2)
+            Error::InternalError
         }
     }
 }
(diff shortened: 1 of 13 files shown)


martinthomson:
I guess my question is for @ddragana: what does the telemetry for these look like?

If we're OK with removing all of them, we should remove the enum variant from `Error` and the CI checks as well.

ddragana:
> I guess my question is for @ddragana: what does the telemetry for these look like?
> 

Looking at the recent data: 
InternalError(5) occurred 2 times in a week and InternalError(4) occurred once. 
This is very low. I would prefer that these numbers are 0 but we will not be able to investigate this.  
We can remove it.

> If we're OK with removing all of them, we should remove the enum variant from `Error` and the CI checks as well.

larseggert:
@martinthomson @KershawChang is this still something we want to do? Or is this OBE?

martinthomson:
We're not really using them, so it would be cleaner not to have them.

larseggert:
@Grubba27 want to take a stab at resolving the conflicts that have accumulated?

larseggert:
I fixed the conflicts. I'll note that there are still occurrences of `InternalError` in the code in this PR. Not sure if that is to be expected.

martinthomson:
Well, we could just purge the lot.  I think that the telemetry has served its purpose, so that might be OK.

This patch doesn't build because it means that a few functions go from returning both Ok or Error in different circumstances to always returning `Ok(())`, which clippy rightfully complains about.  There also appears to be some other problems.

martinthomson:
It probably requires a little bit of analysis, but these are all places where we would have previously had an assertion (`assert!()` or maybe `debug_assert!()`.  I would be OK collapsing this all onto a single item and removing the number, which is an awkward thing to maintain.

larseggert (approved):

PR #1327, opened 2022-02-22
Bump minimum rust version to 1.57

### LINKED ISSUE #1325: rust_decimal crate uses features that were unstable in rust 1.51
1.51 is our current minimum supported version; we should pin our use of rust_decimal to 1.20.  Unfortunately, we pick this up via the sfv crate.  sfv 0.9.1 (our dependency) specifies rust_decimal ^1.20.0, which selects 1.21.0, which doesn't compile.

We need to include it directly (which doesn't really guarantee that we get the right version, but seems to work).

martinthomson:
This fixes all the lints that newly fail and removes some overrides that
were previously necessary.

Fixes #1325.
Closes #1326.


mb:
If we bump the version here, you could also include the msrv in the Cargo.toml: https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html#cargo-rust-version. We could do it in a separate PR if you think it should be done separately.

ddragana (approved):

PR #1337, opened 2022-04-06
Don't send small Initial packets

### LINKED ISSUE #1336: msquic will drop our second flight
If we send an Initial packet coalesced with a Handshake packet, we don't pad the datagram to 1200 bytes.

The spec says that the client is required to pad these packets to 1200 bytes.  We don't.  That is because we don't need to: sending the Handshake packet ensures that the server will be able to perform address validation (if it can't use the connection ID for that).

The spec doesn't say that the server can drop these.  [The packet handling text says](https://quicwg.org/base-drafts/rfc9000.html#section-5.2-2):

> Endpoints try to associate a packet with an existing connection. If the packet has a non-zero-length Destination Connection ID corresponding to an existing connection, QUIC processes that packet accordingly. Note that more than one connection ID can be associated with a connection; see [Section 5.1](https://quicwg.org/base-drafts/rfc9000.html#connection-id).

Or [later text](https://quicwg.org/base-drafts/rfc9000.html#section-5.2.2-3) with the same effect:

> Packets with a supported version, or no Version field, are matched to a connection using the connection ID or -- for packets with zero-length connection IDs -- the local address and port. These packets are processed using the selected connection; otherwise, the server continues as described below.

msquic drops datagrams if they contain an Initial packet and they are smaller than 1200 bytes.  This is overzealous enforcement of the datagram size requirements.  (I can't say if the lack of a "MUST" rule for servers is deliberate, but I do know that we considered making an exception for the sort of behaviour neqo has and ultimately decided that it was too complicated to write down; it's possible then that the omission was deliberate; that would seem to be a mistake in retrospect).  However, it will result in a performance regression for connection establishment.

We only send this datagram once and discard the Initial keys after sending it.  Subsequent datagrams will therefore only contain a Handshake packet, which will be accepted.  However, with msquic the net effect is that we will burn a PTO on connection establishment.

Padding the Initial to avoid this problem isn't free.  The typical size of the second flight is small enough that we gain about a kilobyte of extra data from the congestion window to use on requests.  Given asynchronous certificate handling delays, we are likely unable to add application data to the same datagram, but we might be able to send additional packets immediately afterwards.  Padding the initial would give us one fewer packet to send.

martinthomson:
This turned out to be a little fiddly, but it should also improve
handshake reliability (in rare scenarios, admittedly).

Changes:

1. At the client, discard the Initial packet number space when the
Handshake keys are installed.

2. Because Initial packets from the client are not sent in its second
flight, the server won't get an ACK, which otherwise drives the RTT
estimate.  Rather than disrupt all the tests, when the server discards
its Initial packet number space it will add an implicit RTT sample.

Note: The server will not increase its congestion window as a result of
this.  We might consider doing something about that later.

3. Because the client might only receive Initial packets from the
server, if it discards the Initial packet number space there will be
nothing to send.  No Initial packets and no Handshake packets, not even
acknowledgments.  So, to prevent deadlocks, if *either* endpoint
receives a packet from a space in which it has discarded keys, it will
probe in the next available space when it next sends something.  It will
only do this prior to connection establishment, meaning that it won't
probe with Application packets.

Closes #1336.

### DIFF
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -397,7 +397,9 @@ fn reorder_05rtt_with_0rtt() {
     now += RTT / 2;
     server.process_input(c4.unwrap(), now);
     assert_eq!(*server.state(), State::Confirmed);
-    assert_eq!(server.paths.rtt(), RTT);
+    // Don't check server RTT as it will be massively inflated by a
+    // poor initial estimate received when the server dropped the
+    // Initial packet number space.
 }
 
 /// Test that a server that coalesces 0.5 RTT with handshake packets
@@ -521,7 +523,9 @@ fn reorder_handshake() {
     now += RTT / 2;
     let s3 = server.process(c3, now).dgram();
     assert_eq!(*server.state(), State::Confirmed);
-    assert_eq!(server.paths.rtt(), RTT);
+    // Don't check server RTT estimate as it will be inflated due to
+    // it making a guess based on retransmissions when it dropped
+    // the Initial packet number space.
 
     now += RTT / 2;
     client.process_input(s3.unwrap(), now);
@@ -565,9 +569,8 @@ fn reorder_1rtt() {
     now += RTT / 2;
     let s2 = server.process(c2, now).dgram();
     // The server has now received those packets, and saved them.
-    // The three additional are: an Initial ACK, a Handshake,
-    // and a 1-RTT (containing NEW_CONNECTION_ID).
-    assert_eq!(server.stats().packets_rx, PACKETS * 2 + 5);
+    // The two additional are a Handshake and a 1-RTT (w/ NEW_CONNECTION_ID).
+    assert_eq!(server.stats().packets_rx, PACKETS * 2 + 4);
     assert_eq!(server.stats().saved_datagrams, PACKETS);
     assert_eq!(server.stats().dropped_rx, 1);
     assert_eq!(*server.state(), State::Confirmed);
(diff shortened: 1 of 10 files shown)


neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1012,3 +1015,72 @@ fn bad_min_ack_delay() {
...
+    let server_dgram2 = server.process_output(now + AT_LEAST_PTO).dgram();
+
+    // Only pass on the Initial from the first.  We should get a Handshake in return.
+    let (initial, handshake) = split_datagram(&server_dgram1.unwrap());
+    assert!(handshake.is_some());
+
+    // The client will not acknowledge the Initial as it discards keys.
+    // It sends a Handshake probe instead, containing just a PING frame.
+    assert_eq!(client.stats().frame_tx.ping, 0);
+    let probe = client.process(Some(initial), now).dgram();
```
ddragana, line 1040: can you check client.stats().dropped_rx here? it should be 0.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1012,3 +1015,72 @@ fn bad_min_ack_delay() {
...
+    // Pass the Handshake packet and complete the handshake.
+    now += AT_LEAST_PTO;
+    client.process_input(handshake.unwrap(), now);
+    maybe_authenticate(&mut client);
+    let dgram = client.process_output(now).dgram();
+    let dgram = server.process(dgram, now).dgram();
+    client.process_input(dgram.unwrap(), now);
+
+    assert_eq!(*client.state(), State::Confirmed);
+    assert_eq!(*server.state(), State::Confirmed);
```
ddragana, line 1064: Can we test that a probe for untracted_packets will not be sent after switching to the Confirmed state?

ddragana (approved):

martinthomson:
Before I merge this, can you check the new test?

ddragana (approved):

PR #1338, opened 2022-04-24
Minor improvement to neqo-crypto build script

dlrobertson:
If the NSS_DIR environment variable is absolute the user may be
presented with confusing error messages later in the build
script.

 - The NSS_DIR environment variable (if relative) is expected to be relative to `neqo-crypto`
 - The path added to the search path will not be correct, so the user will see a failure to find a needed lib

We could throw a warning and rely on the user to set the `LD_LIBRARY_PATH`
correctly, but it seems easier to just expect a absolute path.


ddragana (approved):
This looks good to me.

martinthomson (approved):

Issue #1331, opened 2022-03-11
Don't grease VN with the version from the client Initial


martinthomson:
If the client sends 0x?a?a?a?a, we should not include that version in the VN packet.  A client is required to ignore a VN where the server includes the version they attempted, which applies here too.

The easiest way to fix this is to pass the client packet version into PacketBuilder::version_negotiation and ensure that we don't repeat that when greasing (we can just add 0x10 to the version to avoid a collision).  We should include a test for this, which will be probabilistic, but at 2<sup>-16</sup> odds, we will get test failures often enough if we run the test a few hundred times.

The impact of this is minor: the client should ignore the VN packet and continue sending its Initial.  We probably won't make the same mistake multiple times in a row, but it will add unnecessary latency.

PR #1339, opened 2022-04-24
Version Negotiation do not repeat client initial version

dlrobertson:
Avoid repeating the clients initial version in the supported versions
listed in the Version Negotiation packet.

Fixes: #1331

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -437,11 +437,16 @@ impl PacketBuilder {
     }
 
     /// Make a Version Negotiation packet.
-    pub fn version_negotiation(dcid: &[u8], scid: &[u8], versions: &[Version]) -> Vec<u8> {
+    pub fn version_negotiation(
+        dcid: &[u8],
+        scid: &[u8],
+        client_version: u32,
+        versions: &[Version],
+    ) -> Vec<u8> {
         let mut encoder = Encoder::default();
-        let mut grease = random(5);
+        let mut grease = random(4);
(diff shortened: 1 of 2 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -1357,6 +1367,14 @@ mod tests {
         assert_eq!(&vn, &SAMPLE_VN);
     }
 
+    #[test]
+    fn vn_do_not_repeat_client_grease() {
+        fixture_init();
+        let vn =
+            PacketBuilder::version_negotiation(SERVER_CID, CLIENT_CID, 0x0a0a0a0a, &Version::all());
+        assert_ne!(&vn[SAMPLE_VN.len() - 4..], &[0x0a, 0x0a, 0x0a, 0x0a]);
```
dlrobertson, line 1375: Would it make sense to use something like [mockall](https://docs.rs/mockall/latest/mockall/) to mock `random`?

ddragana: We can try that. Can you make a patch?

dlrobertson: :+1: I'll give it a shot. My current thought is to try to [use automock on the module](https://docs.rs/mockall/latest/mockall/#modules), but it could require some amount of reorganization for the `neqo-crypto` interface for `random`.

dlrobertson: Took some time to look at automock and at this time I don't think they'd fit our needs here. Perhaps something to look into in the future though.

martinthomson (approved):

PR #1340, opened 2022-04-26
Suppress borrow_as_ptr warnings

martinthomson:
bindgen makes a hash of these, but so did we in a few places


dlrobertson:
FWIW https://github.com/rust-lang/rust-bindgen/pull/2203 will fix the bindgen cippy errors.

dlrobertson:
Also sorry for introducing a clippy error in https://github.com/mozilla/neqo/commit/f0f578ef5f761ddba41081786824e60bff1182f9 😬

martinthomson:
Hah.  clippy doesn't always pick up problems in build.rs.  I make the same sort of mistake all the time.

ddragana (approved):

PR #1344, opened 2022-04-30
Use AsRef and AsMut instead of Deref

dlrobertson:
Use the `AsRef` and `AsMut` traits instead of `Deref` to access the `Encoder`
and `Decoder` internal buffers.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -715,7 +715,7 @@ impl Connection {
                 tps.borrow().local.encode(enc_inner);
             });
             enc.encode(extra);
-            let records = s.send_ticket(now, &enc)?;
+            let records = s.send_ticket(now, enc.as_ref())?;
             qinfo!([self], "send session ticket {}", hex(&enc));
             self.crypto.buffer_records(records)?;
         } else {
@@ -2128,13 +2128,20 @@ impl Connection {
                 continue;
             }
 
-            dump_packet(self, path, "TX ->", pt, pn, &builder[payload_start..]);
+            dump_packet(
+                self,
+                path,
+                "TX ->",
+                pt,
+                pn,
+                &builder.as_ref()[payload_start..],
+            );
             qlog::packet_sent(
                 &mut self.qlog,
                 pt,
                 pn,
                 builder.len() - header_start + aead_expansion,
-                &builder[payload_start..],
+                &builder.as_ref()[payload_start..],
             );
 
             self.stats.borrow_mut().packets_tx += 1;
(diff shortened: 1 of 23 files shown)


dlrobertson:
There are a lot of changes here for a debatable improvement, so no worries if this isn't a change y'all want to take on. Just an idea I had.

neqo-transport/src/connection/mod.rs
```diff
@@ -2909,7 +2918,7 @@ impl Connection {
         let val = self
```
martinthomson, line 2918: Just above is one place where `data: impl AsRef<u8>` is probably necessary.  You have the `.as_ref()` call here, but that's a noop here without the change.

neqo-transport/src/packet/mod.rs
```diff
@@ -210,7 +210,7 @@ impl PacketBuilder {
     }
 
     fn is_long(&self) -> bool {
-        self[self.header.start] & 0x80 == PACKET_BIT_LONG
+        self.as_ref()[self.header.start] & 0x80 == PACKET_BIT_LONG
```
martinthomson, line 213:
```suggestion
        self.encoder.as_ref()[self.header.start] & 0x80 == PACKET_BIT_LONG
```

Consistency...

martinthomson (commented):
Hmm, in going through this I realized that I don't quite know how far to propagate the `impl AsRef<u8>` call signatures.  There's a case for only letting those go as far as the public API, to avoid monomorphism costs in code that is deeper in our codebase, but that's hard to quantify.

For instance, you have done this for public interfaces on neqo-common and neqo-crypto, but those crate boundaries are really only for our convenience.  It might be more disruptive to use `&[u8]` there, but it seems like it might be better to avoid the potential code bloat that arises from any potential deep monomorphism that might arise from using `impl`  unnecessarily.

dlrobertson:
> Hmm, in going through this I realized that I don't quite know how far to propagate the impl AsRef<u8> call signatures. There's a case for only letting those go as far as the public API, to avoid monomorphism costs in code that is deeper in our codebase, but that's hard to quantify.

Great point. I largely chose this boundary arbitrarily and based on what would be convenient at the time (limiting `s/encoder/encoder.as_ref()/`). Based on my very limited knowledge, I think the main place we'd receive benefits from using `AsRef` is really in the codec. I'm not sure how much we benefit from extending it much beyond that. We might end up with a lot more `.as_ref()` calls, but that is a pretty low cost (could even make things more readable/explicit).

> For instance, you have done this for public interfaces on neqo-common and neqo-crypto, but those crate boundaries are really only for our convenience. It might be more disruptive to use &[u8] there, but it seems like it might be better to avoid the potential code bloat that arises from any potential deep monomorphism that might arise from using impl unnecessarily.

Yeah, this makes a lot of sense. I think as a first implementation I might have gotten overexuberant with my usage of `AsRef`. I'll make an attempt at limiting this in a second changeset and see what happens.

martinthomson (approved):

PR #1347, opened 2022-05-20
Fix changes to VN

martinthomson:
CI not working really hid a major problem here.  Let's not let this rot
so much again...

The code didn't work here because we weren't saving the wire version
when we parsed the packet header for a packet that had an unsupported
version.  Now we do, so we can correctly generate a VN packet.

I took the liberty of renaming as_u32 to wire_version so that we are
more consistent with the naming of the version functions.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -188,7 +188,7 @@ impl PacketBuilder {
             && 11 + dcid.as_ref().len() + scid.as_ref().len() < limit - encoder.len()
         {
             encoder.encode_byte(PACKET_BIT_LONG | PACKET_BIT_FIXED_QUIC | pt.to_byte(version) << 4);
-            encoder.encode_uint(4, version.as_u32());
+            encoder.encode_uint(4, version.wire_version());
             encoder.encode_vec(1, dcid.as_ref());
             encoder.encode_vec(1, scid.as_ref());
         } else {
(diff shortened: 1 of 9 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -503,7 +503,7 @@ pub struct PublicPacket<'a> {
     /// The size of the header, not including the packet number.
     header_len: usize,
     /// Protocol version, if present in header.
-    version: Option<Version>,
+    version: Option<WireVersion>,
```
martinthomson, line 506: I want to flag this, as this is the main change.

neqo-transport/src/packet/mod.rs
```diff
@@ -686,7 +686,12 @@ impl<'a> PublicPacket<'a> {
     }
 
     pub fn version(&self) -> Option<Version> {
-        self.version
+        self.version.and_then(|v| Version::try_from(v).ok())
+    }
+
+    pub fn wire_version(&self) -> WireVersion {
```
martinthomson, line 692: A new accessor here is used only when we need the raw version.

neqo-transport/src/server.rs
```diff
@@ -569,11 +569,11 @@ impl Server {
...
 
-            qdebug!([self], "Unsupported version: {:?}", packet.version());
+            qdebug!([self], "Unsupported version: {:x}", packet.wire_version());
             let vn = PacketBuilder::version_negotiation(
                 packet.scid(),
                 packet.dcid(),
-                packet.version().unwrap().as_u32(),
+                packet.wire_version(),
                 self.conn_params.get_versions().all(),
             );
```
martinthomson, line 578: This is the other substantive change.  Debugging is cleaner and there is no more unwrapping involved.

martinthomson (commented):
Some reviewing notes, as there are lots of renaming changes that hide the real changes.

ddragana (approved):

PR #1353, opened 2022-07-10
`aead_fuzzing::Aead::new`: Don't panic in release builds.

jimblandy:
The original rationale for this panic was to ensure we never ship users a Firefox with a crypto-disabled version of aead. However, it is common for Firefox fuzzing builds to be optimized: fuzzing is often combined with LLVM's Address Sanitizer, whose docs recommend optimized builds. Such Firefox builds crash the moment anything tries to initialize Aead.

There are other safeguards in place that prevent Mozilla from shipping fuzzing-enabled builds, so this one can be removed.

[edit:] This solution to the problem was discussed in [Bug 1743672](https://bugzilla.mozilla.org/show_bug.cgi?id=1743672#c11).


jimblandy:
I believe the CI failures here are addressed by #1354.

martinthomson (approved):

PR #1354, opened 2022-07-10
Placate clippy.

jimblandy:
Use the `write!` macro to extend strings with formatted text.

Allow clippy::module_name_repetitions in the `log` module.


martinthomson (approved):

martinthomson:
Thanks @jimblandy !

PR #1355, opened 2022-07-12
Allow encryption to be enabled/disabled at runtime when fuzzing is enabled.

jschwartzentruber:
This allows quic to be functional in fuzzing builds, unless the connection parameter is explicitly set to disable encryption for fuzzing.


martinthomson (changes requested):
I can't take something that is this intrusive.  There are a number of places where this takes relatively simple code (say 3 lines) and explodes into something that is very hard to understanding (30 lines).  More seriously, having to test `#[cfg(feature = "fuzzing")]` in so many places is not acceptable.

I understand the requirement, but I think that we need to do some software engineering in order to make it happen without these externalities.

Let's say for example that each connection has a central place where instantiations of the Aead type happen.  Maybe we can call that an AeadFactory so that we can get our Java Enterprise Programming Certificate, but the exact name doesn't matter.

In a regular build, AeadFactory can just reduce to Aead::new.  In a fuzzing-enabled build, AeadFactory can be a closure that holds a boolean and returns a new type of object that holds that `bool`.  That object can present the same interface as Aead, so it compiles nicely (the same way we did with the aead_fuzzing.rs file) but it uses its boolean (or maybe it matches on two arms of an enum) to determine which code to run.

The advantage there is that aead.rs doesn't include any `#[cfg(...)]` pollution.  It can concentrate on being correct and performant.  The rest of the code has to carry the burden of passing the AeadMaker around, but that is generic.  In a regular build, that should optimize out because it is fixed.

A separate fix is needed for Retry packets.  That is harder to do and we do want to fuzz them, but the surface area we're protecting is much smaller.  I would suggest we tackle that in a separate change.

There is also header protection, which we haven't got any fuzzing-specific code for, which is a bit of a problem.

I will also suggest that you not touch the self encryption code.  Yes, that uses an AEAD, and yes we will always appreciate more testing, but *we* provide the inputs there, not some random stranger on the Internet.  The AEAD protects our code from adversarial inputs.  Fuzzing that code isn't going to do much good.  Also, that is only used by our server code, which we aren't shipping in Gecko.

jschwartzentruber:
Thanks for the feedback. I originally wanted to abstract Aead & aead_fuzzing::Aead as you describe, but I wasn't sure how to make make the types work. So to be clear, this will be almost as intrusive, but it will be more generic because the "AeadFactory" passed in place of the bool doesn't need to be behind `#[cfg(feature = "fuzzing")]` everywhere, and it has the advantage of not complicating the real Aead.

Why is enabling more tests with the fuzzing feature problematic? I enabled all of the tests because if this code will be used in fuzzing builds, I don't see why you wouldn't want to run the existing tests.

martinthomson:
> Why is enabling more tests with the fuzzing feature problematic? 

If I implied that, that's my mistake.  I like that this tests everything with fuzzing on when the fuzzing feature is enabled.

larseggert:
@jschwartzentruber are you going to rework this PR? Or should we close it?

larseggert:
Closing as OBE

PR #1360, opened 2022-09-07
Do not calculate keepalive's timeout when a keepalive is outstanding.

### LINKED ISSUE #1359: Assertion eariest > now
I hit [this assertion](https://github.com/mozilla/neqo/blob/e71e860489447fed25b64bd8ec31a804dccb0dad/neqo-transport/src/connection/mod.rs#L976) while working with firefox. 
I think the reason was that the server stopped responding and i think sam pings have a wrong sent time or something. First, I need to try to recreate the test case using only neqo.

ddragana:
Fixes #1359

### DIFF
--- a/neqo-http3/tests/httpconn.rs
+++ b/neqo-http3/tests/httpconn.rs
@@ -6,7 +6,7 @@
 
 #![allow(unused_assignments)]
 
-use neqo_common::{event::Provider, Datagram};
+use neqo_common::{event::Provider, qtrace, Datagram};
 use neqo_crypto::{AuthenticationStatus, ResumptionToken};
 use neqo_http3::{
     Header, Http3Client, Http3ClientEvent, Http3OrWebTransportStream, Http3Parameters, Http3Server,
@@ -112,7 +112,7 @@ fn connect_peers(hconn_c: &mut Http3Client, hconn_s: &mut Http3Server) -> Option
     out.dgram()
 }
 
-fn connect_peers_with_rtt(
+fn connect_peers_with_network_propagation_delay(
     hconn_c: &mut Http3Client,
     hconn_s: &mut Http3Server,
     net_delay: u64,
@@ -126,7 +126,8 @@ fn connect_peers_with_rtt(
     now += net_delay;
     let out = hconn_c.process(out.dgram(), now); // ACK
     now += net_delay;
-    mem::drop(hconn_s.process(out.dgram(), now)); //consume ACK
+    let out = hconn_s.process(out.dgram(), now); //consume ACK
+    assert!(out.dgram() == None);
     let authentication_needed = |e| matches!(e, Http3ClientEvent::AuthenticationNeeded);
     assert!(hconn_c.events().any(authentication_needed));
     now += net_delay;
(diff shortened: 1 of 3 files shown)


ddragana:
If an ack-eliciting packet is outstanding and we send a keepalive packet, the IdleTimeoutState will not change, therefore anytime we calculate the keepalive timer it will return the same time as when the keepalive packet has been sent which is in the past.

neqo-http3/tests/httpconn.rs
```diff
@@ -379,3 +412,45 @@ fn zerortt() {
...
 }
+
+#[test]
+fn test_fetch_noresponse_will_idletimeout() {
+    let mut hconn_c = default_http3_client();
+    let mut hconn_s = default_http3_server();
+
+    let (dgram, mut now) = connect_peers_with_rtt(&mut hconn_c, &mut hconn_s, 10);
+
+    eprintln!("-----client");
```
martinthomson, line 423: Use `qtrace!()` for these.

neqo-http3/tests/httpconn.rs
```diff
@@ -111,6 +112,38 @@ fn connect_peers(hconn_c: &mut Http3Client, hconn_s: &mut Http3Server) -> Option
     out.dgram()
 }
 
+fn connect_peers_with_rtt(
+    hconn_c: &mut Http3Client,
+    hconn_s: &mut Http3Server,
+    net_delay: u64,
+) -> (Option<Datagram>, Instant) {
+    let net_delay = Duration::from_millis(net_delay);
```
martinthomson, line 120: API quibble: you are using `net_delay`, which is the one way propagation delay rather than RTT, but the function is named "\_with\_rtt".

neqo-http3/tests/httpconn.rs
```diff
@@ -111,6 +112,38 @@ fn connect_peers(hconn_c: &mut Http3Client, hconn_s: &mut Http3Server) -> Option
...
+    assert_eq!(hconn_c.state(), Http3State::Initializing);
+    let mut now = now();
+    let out = hconn_c.process(None, now); // Initial
+    now += net_delay;
+    let out = hconn_s.process(out.dgram(), now); // Initial + Handshake
+    now += net_delay;
+    let out = hconn_c.process(out.dgram(), now); // ACK
+    now += net_delay;
+    mem::drop(hconn_s.process(out.dgram(), now)); //consume ACK
```
martinthomson, line 129: Rather than mem::drop() here, you should probably just use process_input() at the client, so that it doesn't generate the ACK and the server doesn't need to process it.  Then we don't drop packets in the test. Dropping stuff can affect other things, which might not alter this particular test, but it's probably worthwhile. You'll need to unwrap the datagram to do that, but I think that it's better.

You can also remove some of the extra delays then.

neqo-http3/tests/httpconn.rs
```diff
@@ -111,6 +112,38 @@ fn connect_peers(hconn_c: &mut Http3Client, hconn_s: &mut Http3Server) -> Option
...
+    let out = hconn_c.process(None, now); // Handshake
+    assert_eq!(hconn_c.state(), Http3State::Connected);
+    now += net_delay;
+    let out = hconn_s.process(out.dgram(), now); // Handshake
+    now += net_delay;
+    let out = hconn_c.process(out.dgram(), now);
+    now += net_delay;
+    let out = hconn_s.process(out.dgram(), now);
+    now += net_delay;
+    let out = hconn_c.process(out.dgram(), now);
```
martinthomson, line 143: What do all these other ones do?  HANDSHAKE_DONE?  Control streams?

neqo-http3/tests/httpconn.rs
```diff
@@ -379,3 +412,45 @@ fn zerortt() {
     exchange_packets(&mut hconn_c, &mut hconn_s, out.dgram());
     process_client_events(&mut hconn_c);
 }
+
+#[test]
+fn test_fetch_noresponse_will_idletimeout() {
```
martinthomson, line 417:
```suggestion
/// When a client has an outstanding fetch, it will send keepalives.
/// Test that it will successfully run until the connection times out.
fn fetch_noresponse_will_idletimeout() {
```

martinthomson (commented):
Can we have a test in neqo-transport for this?  The bug is there, so we really should be testing it there as well.

neqo-http3/tests/httpconn.rs
ddragana: We do not need to drop this packet because it is not a datagram. I will add an assertion.

neqo-http3/tests/httpconn.rs
ddragana: I will add a description, but yes HANDSHAKE_DONE and control and qpack streams.

KershawChang (approved):
Looks good to me.

PR #1385, opened 2022-11-23
Support redirects in WebTransport session accept responses

### LINKED ISSUE #1384: Support 3XX responses for Webtransport requests
In order to support testing of redirects for [Bug 1792678](https://bugzilla.mozilla.org/show_bug.cgi?id=1792678), neqo http3 server must be able to send 3XX responses and location headers for a webtransport request.

MayyaSunil:
This PR is intended to fix #1384

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -53,6 +53,22 @@ where
     pub priority: Priority,
 }
 
+pub enum WebTransportSessionAcceptAction {
+    Accept,
+    Reject,
+    Redirect
+}
+
+impl ::std::fmt::Display for WebTransportSessionAcceptAction {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        match self {
+            WebTransportSessionAcceptAction::Accept => write!(f, "Accept"),
+            WebTransportSessionAcceptAction::Reject => write!(f, "Reject"),
+            WebTransportSessionAcceptAction::Redirect => write!(f, "Redirect"),
+        }
+    }
+}
+
 #[derive(Debug)]
 enum Http3RemoteSettingsState {
     NotReceived,
@@ -1100,9 +1116,9 @@ impl Http3Connection {
         conn: &mut Connection,
         stream_id: StreamId,
         events: Box<dyn ExtendedConnectEvents>,
-        accept: bool,
+        accept_res: WebTransportSessionAcceptAction,
     ) -> Res<()> {
-        qtrace!("Respond to WebTransport session with accept={}.", accept);
+        qtrace!("Respond to WebTransport session with accept={}.", accept_res);
         if !self.webtransport_enabled() {
             return Err(Error::Unavailable);
         }
@@ -1119,14 +1135,14 @@ impl Http3Connection {
 
         let send_stream = self.send_streams.get_mut(&stream_id);
 
-        match (send_stream, recv_stream, accept) {
+        match (send_stream, recv_stream, accept_res) {
             (None, None, _) => Err(Error::InvalidStreamId),
             (None, Some(_), _) | (Some(_), None, _) => {
                 // TODO this needs a better error
                 self.cancel_fetch(stream_id, Error::HttpRequestRejected.code(), conn)?;
                 Err(Error::InvalidStreamId)
             }
-            (Some(s), Some(_r), false) => {
+            (Some(s), Some(_r), WebTransportSessionAcceptAction::Reject) => {
                 if s.http_stream()
                     .ok_or(Error::InvalidStreamId)?
                     .send_headers(&[Header::new(":status", "404")], conn)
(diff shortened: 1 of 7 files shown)


neqo-http3/src/connection.rs
```diff
@@ -53,6 +53,22 @@ where
     pub priority: Priority,
 }
 
+pub enum WebTransportSessionAcceptAction {
+    Accept,
+    Reject,
+    Redirect
```
KershawChang, line 59: I think we might want to treat `Redirect` as `Reject`, so maybe change this to:
```
pub enum WebTransportSessionAcceptAction {
    Accept,
    Reject(Vec<Header>)
}
```

neqo-http3/src/connection.rs
```diff
@@ -1166,6 +1182,34 @@ impl Http3Connection {
...
+            (Some(s), Some(_r), WebTransportSessionAcceptAction::Redirect) => {
+                if s.http_stream()
+                    .ok_or(Error::InvalidStreamId)?
+                    .send_headers(&[Header::new(":status", "302"), 
+                                  Header::new("location", "/")],
+                                  conn)
+                    .is_ok()
+                {
+                    let extended_conn =
+                        Rc::new(RefCell::new(WebTransportSession::new_with_http_streams(
```
KershawChang, line 1194: I think we don't want to create a WebTransport session, since we already redirect the client to another server.

KershawChang (changes requested):
I think we also need to add tests for this PR.

neqo-http3/src/connection.rs
martinthomson: It might be OK to signal redirects separately, but we'd also want to know *where* the redirect was to.

neqo-http3/src/connection.rs
```diff
@@ -53,6 +53,22 @@ where
...
+    Reject,
+    Redirect
+}
+
+impl ::std::fmt::Display for WebTransportSessionAcceptAction {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        match self {
+            WebTransportSessionAcceptAction::Accept => write!(f, "Accept"),
+            WebTransportSessionAcceptAction::Reject => write!(f, "Reject"),
+            WebTransportSessionAcceptAction::Redirect => write!(f, "Redirect"),
```
martinthomson, line 67:
```suggestion
            WebTransportSessionAcceptAction::Accept => f.write_str("Accept"),
            WebTransportSessionAcceptAction::Reject => f.write_str("Reject"),
            WebTransportSessionAcceptAction::Redirect => f.write_str("Redirect"),
```

Or even
```rust
        f.write_str(match self {
            WebTransportSessionAcceptAction::Accept => "Accept",
            WebTransportSessionAcceptAction::Reject => "Reject",
            WebTransportSessionAcceptAction::Redirect => "Redirect",
        })
```

No need to use macros for this one.

neqo-http3/src/connection.rs
martinthomson: Definitely.  The right answer here is - unfortunately - send this right back to the caller.  We don't do connection management and so we have to bump this entirely.

martinthomson (commented):
I think that @KershawChang covered the high points here already.

Are there really any need to for changes to the API here in light of the need to pass the redirect back to the caller?  Do we just need a way to signal the headers if the session is rejected?

MayyaSunil:
> I think we also need to add tests for this

MayyaSunil:
Thank you for the comments Kershaw and Martin. 
I have incorporated your comments. 
Please let me know if there are more issues that need to be addressed.

neqo-http3/src/connection.rs
```diff
@@ -53,6 +53,22 @@ where
     pub priority: Priority,
 }
 
+pub enum WebTransportSessionAcceptAction {
+    Accept,
+    Reject,
+    Redirect(Vec<Header>),
```
KershawChang, line 59: Can we just do this?
```
Reject(Vec<Header>),
```

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -154,6 +157,46 @@ fn wt_session_respone_with_1xx() {
...
+            wt_server_session = Some(session);
+        }
+    }
+
+    let mut wt_server_session = wt_server_session.unwrap();
+
+    let accept_res: WebTransportSessionAcceptAction = WebTransportSessionAcceptAction::Redirect(
+        [Header::new(":status", "302"), Header::new("location", "/")].to_vec(),
+    );
+    wt_server_session.response(&accept_res).unwrap();
```
KershawChang, line 194: Why not reuse the current function `negotiate_wt_session` to send response?
It should be possible to do:
```
 let (wt_session_id, _wt_session) =
    wt.negotiate_wt_session(&accept_res);
```

MayyaSunil: Done

KershawChang (approved):
Looks good to me. Thanks.

PR #1389, opened 2022-12-01
Mark WebTransport streams keep alive

### LINKED ISSUE #1369: Mark WebTransport streams keep-alive
We need to propagate neqo_transport::Connection to  this function and call this function:
 https://github.com/mozilla/neqo/blob/25a862e97072a06b2ac93a6d5035e0202c0c7d6f/neqo-http3/src/connection.rs#L1048

MayyaSunil:
This PR is intended to resolve #1369


martinthomson (changes requested):
See my comments on the issue.

martinthomson (changes requested):
I don't think you have the right places for these calls.

You need one for the server and one for the client.

I think that the client one goes [in webtransport_create_session](https://github.com/mozilla/neqo/blob/b045c4a7033feffb133d05f86acea74dc50084ec/neqo-http3/src/connection.rs#L1054) and the server one goes [in webtransport_accept_session](https://github.com/mozilla/neqo/blob/b045c4a7033feffb133d05f86acea74dc50084ec/neqo-http3/src/connection.rs#L1098).

The points you have identified are about streams created within the session, not the session stream.

Finally, I would like to see test cases for this, showing that keep-alives (PING) are sent if the connection is left to go idle for too long, with only a webtransport session active.  There are test cases for keep-alive that you can base this on.

larseggert:
@MayyaSunil are you still intending to land this change?

larseggert:
@MayyaSunil any update on when you might get to this?

MayyaSunil:
> @MayyaSunil any update on when you might get to this?

I will resume this early next week and hope to complete by the end of the week

larseggert:
OK. Making this a draft PR until it's ready for review, please change it back then.

larseggert:
@MayyaSunil is this good to go functionality-wise? Does it still need a test?

MayyaSunil:
@larseggert Its not okay functionality wise and needs a test

larseggert:
CC @jesup

larseggert:
Webtransport newbie here. Trying to help wrap this up.

> I think that the client one goes [in webtransport_create_session](https://github.com/mozilla/neqo/blob/b045c4a7033feffb133d05f86acea74dc50084ec/neqo-http3/src/connection.rs#L1054)

The call to `create_bidi_transport_stream` in that function already enabled keepalives. Hence, are we done for the client?

> and the server one goes [in webtransport_accept_session](https://github.com/mozilla/neqo/blob/b045c4a7033feffb133d05f86acea74dc50084ec/neqo-http3/src/connection.rs#L1098).

I've made that addition, and added a test. WDYT?

KershawChang (approved):
This looks good to me.
Thanks!

neqo-http3/tests/webtransport.rs
```diff
@@ -229,6 +229,15 @@ fn receive_data_server(
     wt_stream.unwrap()
 }
 
+#[test]
+fn wt_keepalive() {
+    let (mut client, mut server) = connect();
+    let _wt_session = create_wt_session(&mut client, &mut server);
+    let idle_timeout = ConnectionParameters::default().get_idle_timeout();
+    assert_eq!(client.process_output(now()).callback(), idle_timeout / 2);
+    assert_eq!(server.process_output(now()).callback(), idle_timeout / 2);
```
mxinden, line 238:
```suggestion
    // Expect client and server to send PING after half of the idle timeout in order to keep connection alive.
    assert_eq!(client.process_output(now()).callback(), idle_timeout / 2);
    assert_eq!(server.process_output(now()).callback(), idle_timeout / 2);
```

Feel free to ignore.

mxinden (approved):

larseggert:
@martinthomson could you clear your old review, or better yet re-review?

martinthomson (approved):
This is certainly a lot cleaner than I remember it being.

Issue #1391, opened 2022-12-08
Turn on TLS greasing


martinthomson:
When https://phabricator.services.mozilla.com/D161806 lands, we should enable greasing.

This means either tolerating a failure to set the option or bumping the NSS version requirement (and waiting for an NSS release).  I might lean toward the former for the moment to avoid a pipeline stall.

martinthomson:
Ah, what I meant was rather than adding to [configure](https://github.com/mozilla/neqo/blob/68c8186d7d5c74a57c614ca21aef1ef9b5d1ea61/neqo-crypto/src/agent.rs#L398) as follows:

```rust
self.set_option(Opt::Grease, true)?;
```

We instead do:

```rust
// Ignore whether this succeeds or not, until NSS 3.?? is a runtime requirement.
let _ = self.set_option(Opt::Grease, true);
```

...or maybe we just watch for an error and log it.

```rust
if let Err(e) = self.set_option(Opt::Grease, true) {
    // Until NSS 3.?? is a requirement for running, it's OK to fail here.
    qinfo!([self], "Failed to enable greasing {:?}", e);
}
```

We currently have an [NSS version check](https://github.com/mozilla/neqo/blob/68c8186d7d5c74a57c614ca21aef1ef9b5d1ea61/neqo-crypto/src/lib.rs#L112) in the `neqo-crypto` crate that we could update to require a version of NSS with greasing.  But I see no reason to demand that yet.  We should be able to compile against a newer NSS version and rely on the NSS ABI compatibility promise to ensure that setting this option fails cleanly.

dennisjackson:
This has landed and will be in NSS 3.89, expected to release on March 9th.

martinthomson:
ψ(｀∇´)ψ  thanks @dennisjackson (and thanks also to Leander); that's pretty significant work.

PR #1393, opened 2022-12-13
enable tls greasing

KershawChang:
For issue #1391 .


martinthomson (approved):
This is one for which there is no point in writing a test case.  We can't even validate that the extension was successfully negotiated as we simply ignore its presence.  Let's just trust that NSS does its job.

One minor nit to fix though.

Issue #1395, opened 2022-12-13
gecko build issue with version v0.6.2


KershawChang:
I encountered this error when building gecko with v0.6.2.
```
 0:16.07    Compiling neqo_glue v0.1.0 (/Users/changkershaw/work/gecko/netwerk/socket/neqo_glue)
 0:16.17 error[E0603]: module `features` is private
 0:16.17    --> netwerk/socket/neqo_glue/src/lib.rs:11:5
 0:16.17     |
 0:16.17 11  |     features::extended_connect::SessionCloseReason, Error as Http3Error, Http3Client,
 0:16.17     |     ^^^^^^^^ private module
 0:16.17     |
 0:16.17 note: the module `features` is defined here
 0:16.17    --> /Users/changkershaw/work/neqo/neqo/neqo-http3/src/lib.rs:146:1
 0:16.17     |
 0:16.17 146 | mod features;
 0:16.17     | ^^^^^^^^^^^^^
 0:16.21 For more information about this error, try `rustc --explain E0603`.
 0:16.21 error: could not compile `neqo_glue` due to previous error
```

PR #1396, opened 2022-12-13
make 'features' module public

KershawChang:
To address issue #1395

### DIFF
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -143,7 +143,7 @@ mod connection_client;
 mod connection_server;
 mod control_stream_local;
 mod control_stream_remote;
-mod features;
+pub mod features;
 mod frames;
 mod headers_checks;
 mod priority;


martinthomson (approved):
As a hygiene matter, this means that you probably need to look at the module and decide what inside that module need to be public.

My sense is that the NegotiatingState needs to be public, but the functions on it do not.  Is that right?

If that's the case, maybe change pub to pub(crate) on those functions.

KershawChang:
> As a hygiene matter, this means that you probably need to look at the module and decide what inside that module need to be public.
> 
> My sense is that the NegotiatingState needs to be public, but the functions on it do not. Is that right?
> 
> If that's the case, maybe change pub to pub(crate) on those functions.

It's  [features::extended_connect::SessionCloseReason](https://searchfox.org/mozilla-central/rev/43cb6eca1c3069d46d589d52ab4949257e630d19/netwerk/socket/neqo_glue/src/lib.rs#11) needs to be accessed, but it's inside a private module.

PR #1400, opened 2023-01-09
Expose SendStreamStats

KershawChang:
This is for #issue 1379

### DIFF
--- a/neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
+++ b/neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
@@ -98,7 +98,8 @@ fn connect_with(client: &mut Http3Client, server: &mut Http3Server) {
     let out = client.process(out.dgram(), now());
     let out = server.process(out.dgram(), now());
     let out = client.process(out.dgram(), now());
-    std::mem::drop(server.process(out.dgram(), now()));
+    let out = server.process(out.dgram(), now());
+    std::mem::drop(client.process(out.dgram(), now()));
 }
 
 fn connect(
@@ -201,13 +202,10 @@ impl WtTest {
         loop {
             now += RTT / 2;
             out = self.client.process(out, now).dgram();
+            let client_none = out.is_none();
             now += RTT / 2;
             out = self.server.process(out, now).dgram();
-            now += RTT / 2;
-            out = self.client.process(out, now).dgram();
-            now += RTT / 2;
-            out = self.server.process(out, now).dgram();
-            if out.is_none() {
+            if client_none && out.is_none() {
                 break;
             }
         }
--- a/neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
+++ b/neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
@@ -13,13 +13,14 @@ use neqo_common::event::Provider;
 use crate::{
     features::extended_connect::SessionCloseReason, Error, Header, Http3Client, Http3ClientEvent,
     Http3OrWebTransportStream, Http3Parameters, Http3Server, Http3ServerEvent, Http3State,
-    WebTransportEvent, WebTransportRequest, WebTransportServerEvent,
+    SendStreamStats, WebTransportEvent, WebTransportRequest, WebTransportServerEvent,
     WebTransportSessionAcceptAction,
 };
 use neqo_crypto::AuthenticationStatus;
 use neqo_transport::{ConnectionParameters, StreamId, StreamType};
 use std::cell::RefCell;
 use std::rc::Rc;
+use std::time::Duration;
 
 use test_fixture::{
     addr, anti_replay, fixture_init, now, CountingConnectionIdGenerator, DEFAULT_ALPN_H3,
@@ -194,10 +195,18 @@ impl WtTest {
     }
 
     fn exchange_packets(&mut self) {
+        const RTT: Duration = Duration::from_millis(10);
         let mut out = None;
+        let mut now = now();
         loop {
-            out = self.client.process(out, now()).dgram();
-            out = self.server.process(out, now()).dgram();
+            now += RTT / 2;
+            out = self.client.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.server.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.client.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.server.process(out, now).dgram();
             if out.is_none() {
                 break;
             }
@@ -302,6 +311,10 @@ impl WtTest {
             data.len()
         );
         self.exchange_packets();
+    }
+
+    fn send_stream_stats(&mut self, wt_stream_id: StreamId) -> Result<SendStreamStats, Error> {
+        self.client.webtransport_send_stream_stats(wt_stream_id)
     }
 
     fn receive_data_client(
--- a/neqo-http3/src/features/extended_connect/webtransport_streams.rs
+++ b/neqo-http3/src/features/extended_connect/webtransport_streams.rs
@@ -13,7 +13,6 @@ use neqo_common::Encoder;
 use neqo_transport::{Connection, StreamId};
 use std::cell::RefCell;
 use std::rc::Rc;
-use std::time::{SystemTime, UNIX_EPOCH};
 
 pub const WEBTRANSPORT_UNI_STREAM: u64 = 0x54;
 pub const WEBTRANSPORT_STREAM: u64 = 0x41;
@@ -92,9 +91,6 @@ pub(crate) struct WebTransportSendStream {
     events: Box<dyn SendStreamEvents>,
     session: Rc<RefCell<WebTransportSession>>,
     session_id: StreamId,
-    bytes_written: u64,
-    bytes_sent: u64,
-    bytes_non_app_data: u64,
 }
 
 impl WebTransportSendStream {
@@ -125,9 +121,6 @@ impl WebTransportSendStream {
             events,
             session_id,
             session,
-            bytes_written: 0,
-            bytes_sent: 0,
-            bytes_non_app_data: 0,
         }
     }
 
@@ -152,11 +145,6 @@ impl SendStream for WebTransportSendStream {
     fn send(&mut self, conn: &mut Connection) -> Res<()> {
         if let WebTransportSenderStreamState::SendingInit { ref mut buf, fin } = self.state {
             let sent = conn.stream_send(self.stream_id, &buf[..])?;
-            // We only want to count bytes when the state is SendingData.
-            // However, the underlying connection is not aware of the state
-            // here. That's why we need to count the bytes sent before
-            // SendingData state.
-            self.bytes_non_app_data += sent as u64;
             if sent == buf.len() {
                 if fin {
                     conn.stream_close_send(self.stream_id)?;
@@ -190,9 +178,7 @@ impl SendStream for WebTransportSendStream {
     fn send_data(&mut self, conn: &mut Connection, buf: &[u8]) -> Res<usize> {
         self.send(conn)?;
         if self.state == WebTransportSenderStreamState::SendingData {
-            self.bytes_written += buf.len() as u64;
             let sent = conn.stream_send(self.stream_id, buf)?;
-            self.bytes_sent += sent as u64;
             Ok(sent)
         } else {
             Ok(0)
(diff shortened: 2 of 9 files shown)


neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
```diff
@@ -189,10 +190,18 @@ impl WtTest {
...
-            out = self.client.process(out, now()).dgram();
-            out = self.server.process(out, now()).dgram();
+            now += RTT / 2;
+            out = self.client.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.server.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.client.process(out, now).dgram();
+            now += RTT / 2;
+            out = self.server.process(out, now).dgram();
```
martinthomson, line 204: Can you add some explanation as to why you need two iterations (client->server->client->server, then loop) rather than one (client->server, loop)?  It's not obvious to me why this is necessary.

There's another version of this that looks like this:

```rust
            out = self.client.process(out, now()).dgram();
            let client_none = out.is_none();
            out = self.server.process(out, now()).dgram();
            if client_none && out.is_none() {
                break;
            }
```
That might be clearer (and less likely to be wasteful).

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -91,6 +92,9 @@ pub(crate) struct WebTransportSendStream {
     events: Box<dyn SendStreamEvents>,
     session: Rc<RefCell<WebTransportSession>>,
     session_id: StreamId,
+    bytes_written: u64,
+    bytes_sent: u64,
+    bytes_non_app_data: u64,
```
martinthomson, line 97: We should instead track everything necessary at the transport level.  This is a generic capability that would be useful for many other purposes.  The only correction you need specific to web transport is the removal of any stream header.  We can correct for with a simple calculation; see below.

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -145,6 +152,11 @@ impl SendStream for WebTransportSendStream {
     fn send(&mut self, conn: &mut Connection) -> Res<()> {
         if let WebTransportSenderStreamState::SendingInit { ref mut buf, fin } = self.state {
             let sent = conn.stream_send(self.stream_id, &buf[..])?;
+            // We only want to count bytes when the state is SendingData.
+            // However, the underlying connection is not aware of the state
+            // here. That's why we need to count the bytes sent before
+            // SendingData state.
+            self.bytes_non_app_data += sent as u64;
```
martinthomson, line 159: Given that we can calculate this value based on the encoded size of the session ID (varint length of the session ID / 4), do we really need to track this in this way?

neqo-transport/src/send_stream.rs
```diff
@@ -841,6 +847,7 @@ impl SendStream {
                 ref mut send_buf, ..
             } => {
                 send_buf.mark_as_acked(offset, len);
+                self.bytes_acked += len as u64;
```
martinthomson, line 850: I don't think that this is right.   The spec says:

> Only sequential bytes up to, but not including, the first non-acknowledged byte, are counted. 

You need to access `TxBuffer::retired` for this value.  Unfortunately, this means navigating the `SendStreamState` enum, which doesn't always retain the transmission buffer (or the final size).  Adding a `final_size` to the `DataRecvd` state should cover the success cases.  We will lose the acknowledged data if there is a reset, but you could add another field to each of the `Reset*` states.  The other cases have zero data sent.

Then define a `SendStream::received` function.  That function can check the state and return `retired` if there is a `TxBuffer`, `final_size` if the state is `DataRecvd`, the acknowledged size for `Reset*`, or zero otherwise.

neqo-http3/src/lib.rs
```diff
@@ -543,6 +544,47 @@ trait HttpRecvStreamEvents: RecvStreamEvents {
     fn extended_connect_new_session(&self, _stream_id: StreamId, _headers: Vec<Header>) {}
 }
 
+// See https://www.w3.org/TR/webtransport/#send-stream-stats.
+#[derive(Debug, Clone)]
+pub struct SendStreamStats {
+    pub timestamp: Duration,
```
martinthomson, line 550: Is this really a `Duration`?  What is the reference time?

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -178,7 +190,9 @@ impl SendStream for WebTransportSendStream {
     fn send_data(&mut self, conn: &mut Connection, buf: &[u8]) -> Res<usize> {
         self.send(conn)?;
         if self.state == WebTransportSenderStreamState::SendingData {
+            self.bytes_written += buf.len() as u64;
```
martinthomson, line 193: This will count rejected bytes, which isn't the intent of the statistic.

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -199,4 +213,18 @@ impl SendStream for WebTransportSendStream {
...
         Ok(())
     }
+
+    fn stats(&mut self, conn: &mut Connection) -> Res<SendStreamStats> {
+        let mut acked = conn.stream_bytes_acked(self.stream_id)?;
+        acked -= self.bytes_non_app_data;
+        let stats = SendStreamStats::new(
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("SystemTime before UNIX EPOCH!"),
```
martinthomson, line 223: We should not be invoking `SystemTime::now()` in this code (or `Instant::now()`).  We always have the caller tell us what time it is.

martinthomson (changes requested):

neqo-http3/src/features/extended_connect/webtransport_streams.rs
KershawChang: OK, so maybe we don't need to have `timestamp` in `SendStreamStats` here.

neqo-http3/src/features/extended_connect/webtransport_streams.rs
KershawChang: I'll change this. Thanks.

neqo-http3/src/lib.rs
KershawChang: I'll remove this.

neqo-transport/src/send_stream.rs
KershawChang: I'll try to do this. Thanks.

neqo-http3/src/features/extended_connect/webtransport_streams.rs
KershawChang: Will do.

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
KershawChang: I'll need to dig into this. I can only tell that one iteration does not work for now.

neqo-http3/src/features/extended_connect/webtransport_streams.rs
KershawChang: Sorry, I am not sure I really understand this.
@martinthomson  Could you explain a bit more? Thanks.

martinthomson: It looks like you are trying to calculate the data sent on a stream, excluding the stream header (in this case, that is the identifier we send to indicate that the stream belongs to a particular webtransport session).  We track the total number of bytes sent on a stream, but subtract the size of this header from the value that we report.

But we know what that overhead is and can calculate it when we report the webtransport-specific stats.  So rather than keep a variable that only ever holds a value of 1, we can calculate the value instead (with `Encoder::varint_len()` and the value from here, which we should know).

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -199,4 +199,28 @@ impl SendStream for WebTransportSendStream {
         }
         Ok(())
     }
+
+    fn stats(&mut self, conn: &mut Connection) -> Res<SendStreamStats> {
+        let stream_header_size = if self.stream_id.is_client_initiated() {
+            let id_len = if self.stream_id.is_uni() {
+                Encoder::varint_len(WEBTRANSPORT_UNI_STREAM)
+            } else {
+                Encoder::varint_len(WEBTRANSPORT_STREAM)
```
martinthomson, line 208: These are constants, right?  I just checked and `Encoder::varint_len` can be made `const`.  That might push toward:

```rust
const TYPE_LEN_UNI: usize = Encoder::varint_len(WEBTRANSPORT_UNI_STREAM);
const TYPE_LEN_BIDI: usize = Encoder::varint_len(WEBTRANSPORT_STREAM);
let id_len = if self.stream_id.is_uni() { TYPE_LEN_UNI } else { TYPE_LEN_BIDI };
```

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -199,4 +199,28 @@ impl SendStream for WebTransportSendStream {
...
+        } else {
+            0
+        };
+        let subtract_non_app_bytes = |count: u64| -> u64 {
+            if count >= stream_header_size {
+                count - stream_header_size
+            } else {
+                0
+            }
+        };
```
martinthomson, line 220: Maybe instead:

```rust
let subtract_non_app_bytes = |count: u64| -> u64 {
    count.saturating_sub(stream_header_size)
};
```

You might find that it is easier to inline that instead.

neqo-transport/src/send_stream.rs
```diff
@@ -552,6 +561,8 @@ pub struct SendStream {
     priority: TransmissionPriority,
     retransmission_priority: RetransmissionPriority,
     retransmission_offset: u64,
+    bytes_written: u64,
+    bytes_sent: u64,
```
martinthomson, line 565: Do we need to track this here?

Edit: see comments below...

bytes_written => retired + buf.len(), then you need to track a value
bytes_sent => fc.used(), then final_size

neqo-transport/src/send_stream.rs
```diff
@@ -596,6 +609,25 @@ impl SendStream {
...
+    }
+
+    pub fn bytes_sent(&self) -> u64 {
+        self.bytes_sent
+    }
+
+    pub fn bytes_acked(&self) -> u64 {
+        match &self.state {
+            SendStreamState::Send { send_buf, .. } => send_buf.retired(),
+            SendStreamState::DataSent { send_buf, .. } => send_buf.retired(),
```
martinthomson, line 623:
```suggestion
            SendStreamState::Send { send_buf, .. }
            | SendStreamState::DataSent { send_buf, .. } => send_buf.retired(),
```

neqo-transport/src/send_stream.rs
```diff
@@ -596,6 +609,25 @@ impl SendStream {
...
+            SendStreamState::DataRecvd { retired } => *retired,
+            SendStreamState::ResetSent { final_retired, .. } => *final_retired,
+            SendStreamState::ResetRecvd { final_retired } => *final_retired,
```
martinthomson, line 626:
```suggestion
            SendStreamState::ResetSent { final_retired, .. }
            | SendStreamState::ResetRecvd { final_retired } => *final_retired,
```

neqo-transport/src/send_stream.rs
```diff
@@ -718,7 +750,7 @@ impl SendStream {
                 builder.encode_vvec(&data[..length]);
             }
             debug_assert!(builder.len() <= builder.limit());
-
+            self.bytes_sent += length as u64;
```
martinthomson, line 753: This will overcount if you have packet loss.

That suggests that you want to add a test case where you need to retransmit lost data, so that you can ensure that this value correctly matches the amount of data that has been sent.

I think that the right answer - if we can't track this some other way - is:

```rust
self.bytes_sent = max(self.bytes_sent, offset + u64::try_from(length).unwrap());
```

Edit:  I just realized that you don't need a variable for this either.  This is already tracked by our flow control.  `fc.used()` should give you the same value (ignore the connection-level flow control and just look at the stream level).  That value is carried over into the final size.

neqo-transport/src/send_stream.rs
```diff
@@ -992,6 +1029,7 @@ impl SendStream {
                 let sent = send_buf.send(buf);
                 fc.consume(sent);
                 conn_fc.borrow_mut().consume(sent);
+                self.bytes_written += sent as u64;
```
martinthomson, line 1032: You should be able to track this on a send buffer as `self.retired + self.send_buf.len()` (with appropriate casting) without the additional variable.

If you need to retain this, then keeping it on `SendState`, like you have done for the retired value, should be fine.  It's a little more work to get, but less risk of something getting desynchronized.

neqo-transport/src/send_stream.rs
```diff
@@ -596,6 +609,25 @@ impl SendStream {
         }
     }
 
+    pub fn bytes_written(&self) -> u64 {
```
martinthomson, line 612: I think that you should document what these functions return a little more thoroughly.  They aren't 100% obvious.  (Don't rely on a link to the webtransport spec for the definitions, though the text is pretty good and so it can probably be copied over.)

neqo-transport/src/connection/mod.rs
```diff
@@ -2932,6 +2932,21 @@ impl Connection {
         Ok(())
     }
 
+    pub fn stream_bytes_written(&mut self, stream_id: StreamId) -> Res<u64> {
```
martinthomson, line 2935: Doing three stream lookups like this might push me toward having the stats returned as a struct from here.   That is:

```rust
pub fn stream_stats(&self, stream_id: StreamId) -> Res<Option<SendStreamStats>> {
   Ok(self.streams.get_send_stream(stream_id).map(|s| s.stats()))
}
```

The webtransport code will need to unwrap this and tweak the values, but they will own the answer, so it should be OK.

martinthomson (changes requested):
This is better, but I think that we can cut down further on duplicated state.

neqo-transport/src/send_stream.rs
KershawChang: It seems that using `fc.used()` is not correct here, because `used` value is increased when [consume](https://searchfox.org/mozilla-central/rev/aec3a901e6f6b3041b5ec457c9111a042cef1fb1/third_party/rust/neqo-transport/src/send_stream.rs#993) is called.
I'll put this line in `mark_as_sent`, since it's clearer and easier to write a test for it.
```
self.bytes_sent = max(self.bytes_sent, offset + u64::try_from(length).unwrap());
```

KershawChang:
> Very nice. Fix nits and you can land this one. See [b31a2d3#comments](https://github.com/mozilla/neqo/commit/b31a2d329406977dd409b48feeadc67d39b72291#comments)

Thanks for the review!

PR #1402, opened 2023-01-16
Add idle timeout flag to command-line tools

martinthomson:
Based on a request I'm fielding about this, it seems like it would be helpful to add this.

For a later date, we might consider consolidating some of this code.  The duplication is a problem.  We'd probably want to add that to the mainline code, but under a feature flag, so that we don't pull in all the dependencies unless we're building the command-line tools as well.


marcblanchet:
thanks for doing that. very useful for my testing of long delay networks.

KershawChang (approved):

Issue #1045, opened 2020-11-16
Failure to load dynamic library when running neqo-server


yaseco:
from` /neqo/target/debug` I'm running the following command:

`./neqo-server [::]:12345
`
I'm getting the following error:

thread 'main' panicked at 'NSS_Initialize failed: NssError { name: "PR_LOAD_LIBRARY_ERROR", code: -5977, desc: "Failure to load dynamic library" }', /root/neqo/neqo-crypto/src/lib.rs:134:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

How to fix that?

martinthomson:
You need to point to the NSS binaries using LD_LIBRARY_PATH (or DYLD_LIBRARY_PATH if you are on a mac).  It's a limitation of our build setup that we can't write rpath into binaries.  I don't know where to find that, but you might be able to do something like this:

```sh
export LD_LIBRARY_PATH="$(dirname "$(find . -name libssl3.so -print)")"
```

or for a mac 

```sh
export DYLD_LIBRARY_PATH="$(dirname "$(find . -name libssl3.dylib -print)")"
```

I haven't tested either.

`cargo run` probably doesn't suffer from this problem unless you use an external NSS.

marcblanchet:
ran into same problem. response of @martinthomson works. Suggest to add it to the readme explicitly.

valenting:
I think `cargo run --bin neqo-server -- [::]:12345 --db ./test-fixture/db` and `cargo run --bin neqo-client [URL]` should work instead.

marcblanchet:
> I think `cargo run --bin neqo-server -- [::]:12345 --db ./test-fixture/db` and `cargo run --bin neqo-client [URL]` should work instead.

fine. would be useful to update the README.

PR #1406, opened 2023-01-17
add instructions on how to solve the failure to load dynamic library at run time

marcblanchet:
my take is this issue happens often to users installing and using this software. And one get an hint with the info in the next paragraph, but that is not clear it is also for any kind of installation. And it is not easy to do the right command as the solution is finding the temp dir used at build time. Hence the need to specify the information clearly for people.

reference to issue #1045

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@ To run test HTTP/3 programs (neqo-client and neqo-server):
 * `./target/debug/neqo-server [::]:12345 --db ./test-fixture/db`
 * `./target/debug/neqo-client http://127.0.0.1:12345/`
 
+If a "Failure to load dynamic library" error happens at runtime, do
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,9 @@ To run test HTTP/3 programs (neqo-client and neqo-server):
 * `./target/debug/neqo-server [::]:12345 --db ./test-fixture/db`
 * `./target/debug/neqo-client http://127.0.0.1:12345/`
 
+If a "Failure to load dynamic library" error happens at runtime, do
+export [DY]LD\_LIBRARY\_PATH="$(dirname "$(find . -name libssl3.so -print)")"
```
martinthomson, line 12:
````suggestion
```
export LD_LIBRARY_PATH="$(dirname "$(find . -name libssl3.so -print | head -1)")"
```
On a mac, use `DYLD_LIBRARY_PATH` instead.
````

martinthomson (approved):

PR #1408, opened 2023-01-21
Update gecko build docs

dlrobertson:
- Update the gecko build docs.
 - Update failing build.rs when built with gecko.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,44 +49,27 @@ Some examples:
 
 ### Trying In-development Neqo code in Gecko
 
-In a checked-out copy of Gecko source, set paths for the four Neqo crates to
-local versions in `netwerk/socket/neqo_glue/Cargo.toml`. For example, if Neqo
-was checked out to /home/alice/git/neqo, change:
+In a checked-out copy of Gecko source, set `[patches.*]` values for the four
+Neqo crates to local versions in the root `Cargo.toml`. For example, if Neqo
+was checked out to `/home/alice/git/neqo`, add the following lines to the root
+`Cargo.toml`.
 
 ```
-neqo-http3 = { tag = "v0.1.7", git = "https://github.com/mozilla/neqo" }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,44 +49,27 @@ Some examples:
...
-```
-
-to
+In a checked-out copy of Gecko source, set `[patches.*]` values for the four
+Neqo crates to local versions in the root `Cargo.toml`. For example, if Neqo
+was checked out to `/home/alice/git/neqo`, add the following lines to the root
+`Cargo.toml`.
 
 ```
+[patch."https://github.com/mozilla/neqo"]
```
dlrobertson, line 58: The other way works, but `./mach vendor` will complain and this is more or less what is also done in [firefox-source-docs: Patching third-party crates](https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html#patching-third-party-crates)

neqo-crypto/build.rs
```diff
@@ -371,7 +369,7 @@ fn setup_for_gecko() -> Vec<String> {
     let flags_path = TOPOBJDIR.join("netwerk/socket/neqo/extra-bindgen-flags");
 
     println!("cargo:rerun-if-changed={}", flags_path.to_str().unwrap());
-    flags = fs::read_to_string(flags_path)
+    let mut flags: Vec<String> = fs::read_to_string(flags_path)
         .expect("Failed to read extra-bindgen-flags file")
         .split_whitespace()
         .map(std::borrow::ToOwned::to_owned)
```
martinthomson, line 375: it be better to use this?

```suggestion
        .map(String::from)
```

And then (below):

```rust
   .collect::<Vec<_>>();
```

martinthomson (approved):

neqo-crypto/build.rs
dlrobertson: Yup! That works too

Issue #1364, opened 2022-10-05
WebTransport session should return the reponse headers


ddragana:
Currently when a response for a WebTransport session request is received neqo-http3 parses headers and performs some checks. The application is only informed of the response status code. neqo-http3 should return headers as well.

Current implementation returns `WebTransportEvent::Session {
        stream_id: StreamId,
        status: u16,
    },` if the response was between 200-300 and returns `WebTransportEvent::SessionClosed {
        stream_id: StreamId,
        reason: SessionCloseReason::Status(status),
    }` otherwise.  I think in both cases `WebTransportEvent::Session {...}` should be returned and should contain a response code and headers.

dlrobertson:
Started working on this and should be able to post a PR soon. When updating something like this that will require changes to `neqo_glue`, what is the preferred workflow? Just create a bug in bugzilla for the next neqo update? Or would it be worthwhile to bump the version and update the crate?

PR #1409, opened 2023-01-23
Add headers to session_start and session_end

dlrobertson:
- Add a headers argument to session_start
 - Add an optional headers argument to session_end

Fixes: #1364

### DIFF
--- a/neqo-common/src/timer.rs
+++ b/neqo-common/src/timer.rs
@@ -120,7 +120,7 @@ impl<T> Timer<T> {
             for i in &self.items {
                 debug_assert!(i.is_empty());
             }
-            self.now = time - short_span;
+            self.now = time.checked_sub(short_span).unwrap();
             self.cursor = 0;
         }
 
@@ -287,7 +287,8 @@ mod test {
         t.add(near_future, v);
         assert_eq!(near_future, t.next_time().expect("should return a value"));
         assert_eq!(
-            t.take_until(near_future - Duration::from_millis(1)).count(),
+            t.take_until(near_future.checked_sub(Duration::from_millis(1)).unwrap())
+                .count(),
             0
         );
         assert!(t
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -26,10 +26,12 @@ pub enum WebTransportEvent {
     Session {
         stream_id: StreamId,
         status: u16,
+        headers: Vec<Header>,
     },
     SessionClosed {
         stream_id: StreamId,
         reason: SessionCloseReason,
+        headers: Option<Vec<Header>>,
     },
     NewStream {
         stream_id: StreamId,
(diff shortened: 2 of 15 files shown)


dlrobertson:
Re: https://github.com/mozilla/neqo/issues/1364#issuecomment-1399107392 this requires changes to `neqo_glue`. Should a bug be created? When would the neqo_glue changes happen? I'm happy to start on this, but I'm uncertain of the typical workflow/timing for this sort of work.

neqo-http3/src/client_events.rs
```diff
@@ -181,11 +183,18 @@ impl SendStreamEvents for Http3ClientEvents {
 }
 
 impl ExtendedConnectEvents for Http3ClientEvents {
-    fn session_start(&self, connect_type: ExtendedConnectType, stream_id: StreamId, status: u16) {
+    fn session_start(
+        &self,
+        connect_type: ExtendedConnectType,
+        stream_id: StreamId,
+        status: u16,
+        headers: &Vec<Header>,
```
martinthomson, line 191: Don't you want `&[Header]` for pass by reference?  (That would means `to_vec()` below)

However, given that you immediately and always clone, I think it would be better to have the caller decide whether a clone is needed, so this can take a value.

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -266,13 +268,15 @@ impl WebTransportSession {
...
                             },
+                            Some(headers),
                         );
                         SessionState::Done
                     } else {
                         self.events.session_start(
                             ExtendedConnectType::WebTransport,
                             self.session_id,
                             status,
+                            &headers,
```
martinthomson, line 279: It looks like we don't need to clone here as this is a value that doesn't need to be used later.

martinthomson (commented):
A good start.  I'd like to see some of those modified tests include some assertions about the value of fields, so that we are more confident that there are useful values being carried, even if it is just `:authority` or something like that.

@KershawChang should look at this and work out how/when to include this.

neqo-http3/src/client_events.rs
dlrobertson: Good point! I like this much better.

dlrobertson (commented):
> A good start. I'd like to see some of those modified tests include some assertions about the value of fields, so that we are more confident that there are useful values being carried, even if it is just `:authority` or something like that.

Yeah, that definitely makes sense. I'll take a crack at this over the weekend.

neqo-common/src/timer.rs
```diff
@@ -111,6 +111,7 @@ impl<T> Timer<T> {
     /// Asserts if the time given is in the past or too far in the future.
     /// # Panics
     /// When `time` is in the past relative to previous calls.
+    #[allow(clippy::unchecked_duration_subtraction)] // guarded by assertion
```
dlrobertson, line 114: we've already checked that `time` is greater than `short_span`, so there is no need to check the duration subtraction.

dlrobertson (commented):
Added fix for clippy warning.

martinthomson:
I'm happy with how this turned out.  @KershawChang, over to you.

(For future changes like this, a separate pull request with the time changes would be fine.  Personally, I think that in tests the bare subtraction is OK, particularly since in debug mode that amounts to the same thing, but I guess we asked for pedantic mode so that's what we're getting.)

dlrobertson:
> For future changes like this, a separate pull request with the time changes would be fine.

:+1: will do

KershawChang (approved):
Looks good to me. Thanks.

dlrobertson:
Thanks!

PR #1410, opened 2023-01-25
enable settings_h3_datagram

### LINKED ISSUE #1407: Support to send SETTINGS_H3_DATAGRAM to the server
Since the [implementation](https://github.com/aiortc/aioquic/blob/main/src/aioquic/h3/connection.py#L1095) of `aioquic` requires both `SETTINGS_ENABLE_WEBTRANSPORT` and `SETTINGS_H3_DATAGRAM`, maybe we should do this.

KershawChang:
fix #1407


neqo-http3/src/settings.rs
```diff
@@ -21,6 +21,10 @@ const SETTINGS_MAX_HEADER_LIST_SIZE: SettingsType = 0x6;
 const SETTINGS_QPACK_MAX_TABLE_CAPACITY: SettingsType = 0x1;
 const SETTINGS_QPACK_BLOCKED_STREAMS: SettingsType = 0x7;
 const SETTINGS_ENABLE_WEB_TRANSPORT: SettingsType = 0x2b60_3742;
+// draft-ietf-masque-h3-datagram-04.
+// We use this old value because the current web-platform test only supports
+// this value.
+const SETTINGS_H3_DATAGRAM_DRAFT04: SettingsType = 0x00ff_d277;
```
martinthomson, line 27: RFC 9297 has been published now.  Where do we stand on that?

```suggestion
const SETTINGS_H3_DATAGRAM: SettingsType = 0x33;
```

neqo-http3/src/settings.rs
```diff
@@ -228,6 +251,13 @@ impl ZeroRttChecker for HttpZeroRttChecker {
                 let value = setting.value == 1;
                 self.settings.get_webtransport() || !value
             }
+            HSettingType::EnableH3Datagram => {
+                if setting.value > 0 {
+                    return false;
+                }
+                let value = setting.value == 1;
```
martinthomson, line 258: So this is a cargo culting of previous code, but I think that it needs to be looked at more carefully.

What we want is in this truth table.

| ... | Old Enabled | Old Disabled |
|---|---|---|
| New Enabled | Enable 0-RTT | Enable 0-RTT |
| New Disabled | Disable 0-RTT | Enable 0-RTT |

That is, it is OK to resume unless the setting was previously enabled (`setting.value == 1`) and the setting is now disabled.

This matches what the code does, mostly.  This check:

```rust
                if setting.value > 0 {
                    return false;
                }
```

Is going to disable 0-RTT if the old connection enabled the feature.  I think that there is a very simple fix:

```suggestion
                if setting.value > 1 { // <--- here!
                    return false;
                }
                let value = setting.value == 1;
```

But this highlights that there might be a problem with our 0-RTT tests in that we weren't testing these cases properly.  Can you take a look to see if there are tests for this?

martinthomson (commented):
Two questions here for you that I think are fairly important to answer.

neqo-http3/src/settings.rs
KershawChang: For sure we want to use the value defined in RFC.
I'll change this.

neqo-http3/src/settings.rs
KershawChang: Yes, we have a test [here](https://searchfox.org/mozilla-central/rev/188d0f76a73e0671d12e744a71e9f5701668cc37/third_party/rust/neqo-http3/src/features/extended_connect/tests/webtransport/negotiation.rs#84), but it doesn't really check `early_data_accepted`. I'll also try to fix this.

neqo-http3/src/settings.rs
martinthomson: Sounds reasonable.  We could also send both (as client) and just pick one (as server).

KershawChang: Thanks! I think sending both is easier.

martinthomson (approved):
We don't do HTTP/3 server stuff, do we?  Because I wanted to look for the code for that.

neqo-http3/src/settings.rs
```diff
@@ -98,6 +106,12 @@ impl HSettings {
                         enc_inner.encode_varint(SETTINGS_ENABLE_WEB_TRANSPORT);
                         enc_inner.encode_varint(iter.value);
                     }
+                    HSettingType::EnableH3Datagram => {
+                        enc_inner.encode_varint(SETTINGS_H3_DATAGRAM_DRAFT04);
+                        enc_inner.encode_varint(iter.value);
+                        enc_inner.encode_varint(SETTINGS_H3_DATAGRAM);
+                        enc_inner.encode_varint(iter.value);
```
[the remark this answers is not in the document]
KershawChang: Nice suggestion. Thanks.

neqo-http3/src/features/extended_connect/tests/webtransport/negotiation.rs
```diff
@@ -112,14 +99,8 @@ fn zero_rtt(
...
 
-    let mut client = default_http3_client(
-        Http3Parameters::default()
-            .webtransport(ClientResumedState::ClientResumed.eq(client_resumed)),
-    );
-    let mut server = default_http3_server(
-        Http3Parameters::default()
-            .webtransport(ServerResumedState::ServerResumed.eq(server_resumed)),
-    );
+    let mut client = default_http3_client(Http3Parameters::default().webtransport(client_resumed));
```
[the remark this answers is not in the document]
KershawChang: I think we can fix this in a follow up.

neqo-http3/src/settings.rs
```diff
@@ -191,7 +235,9 @@ impl HttpZeroRttChecker {
             .encode_varint(SETTINGS_QPACK_BLOCKED_STREAMS)
             .encode_varint(settings.get_max_blocked_streams())
             .encode_varint(SETTINGS_ENABLE_WEB_TRANSPORT)
-            .encode_varint(settings.get_webtransport());
+            .encode_varint(settings.get_webtransport())
+            .encode_varint(SETTINGS_H3_DATAGRAM)
+            .encode_varint(settings.get_http3_datagram());
```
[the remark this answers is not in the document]
KershawChang: Yes, it's deliberate. If I read the code correctly, `HttpZeroRttChecker::save` is only called at the server side. So, this means our server code only supports the RFC value.

KershawChang:
> We don't do HTTP/3 server stuff, do we? Because I wanted to look for the code for that.

I admit that I didn't too much attention at the server side. I'll create a follow up issue for this.

PR #1419, opened 2023-03-15
Set target to arm64

### LINKED ISSUE #1411: Failed to build neqo-crypto
I've encountered the error below:
```
ld: warning: ignoring file /Users/changkershaw/work/neqo/neqo/target/debug/build/neqo-crypto-33aaf0b813b0290d/out/dist/Debug/lib/libnspr4.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
  ld: warning: ignoring file /Users/changkershaw/work/neqo/neqo/target/debug/build/neqo-crypto-33aaf0b813b0290d/out/dist/Debug/lib/libplc4.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
  ld: warning: ignoring file /Users/changkershaw/work/neqo/neqo/target/debug/build/neqo-crypto-33aaf0b813b0290d/out/dist/Debug/lib/libplds4.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
  Undefined symbols for architecture x86_64:
    "_PL_ArenaAllocate", referenced from:
        _PORT_ArenaAlloc_Util in libnssutil.a(nssutil.secport.o)
        _PORT_ArenaMark_Util in libnssutil.a(nssutil.secport.o)
    "_PL_ArenaGrow", referenced from:
        _PORT_ArenaGrow_Util in libnssutil.a(nssutil.secport.o)
    "_PL_ArenaRelease", referenced from:
        _port_ArenaRelease in libnssutil.a(nssutil.secport.o)
```

Not sure if this is just me. I've had this issue since my macOS updated to 13.1 last week.
This can be fixed by adding the option below when building NSS.
```
--target=arm64
```

KershawChang:
Fix #1411

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -152,7 +152,8 @@ fn build_nss(dir: PathBuf) {
         build_nss.push(String::from("-j"));
         build_nss.push(d);
     }
-    if env::var("TARGET").unwrap() == "aarch64-apple-darwin" {
+    let target = env::var("TARGET").unwrap();
+    if target.strip_prefix("aarch64-").is_some() {
         build_nss.push(String::from("--target=arm64"));
     }
     let status = Command::new(get_bash())
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -152,6 +152,9 @@ fn build_nss(dir: PathBuf) {
         build_nss.push(String::from("-j"));
         build_nss.push(d);
     }
+    if env::var("TARGET").unwrap() == "aarch64-apple-darwin" {
+        build_nss.push(String::from("--target=arm64"));
+    }
     let status = Command::new(get_bash())
         .args(build_nss)
         .current_dir(dir)


neqo-crypto/build.rs
```diff
@@ -152,6 +152,9 @@ fn build_nss(dir: PathBuf) {
         build_nss.push(String::from("-j"));
         build_nss.push(d);
     }
+    if env::var("TARGET").unwrap() == "aarch64-apple-darwin" {
+        build_nss.push(String::from("--target=arm64"));
+    }
```
martinthomson, line 157: For this, we should be parsing the `$TARGET` triple to get the first token.  It might be better to do:

```suggestion
    let target = env::var("TARGET").unwrap();
    if target.strip_prefix("aarch64-").is_some() {
        build_nss.push(String::from("--target=arm64"));
    }
```

martinthomson (commented):
I don't know if there is some sort of rust API we can use to process triples, but we should at least do something more general here.

martinthomson (approved):

PR #1420, opened 2023-03-15
Switch from chrono to time 0.3

glandium:


martinthomson (approved):

martinthomson:
We can't merge this until we sort out the MSRV situation.  See #1421.

glandium:
Rebasing on top of #1423 shows that clippy is not happy with the undocumented panic that the unwrap adds.

martinthomson:
Easier to generate the comment than deal with the warning; I'll send you a PR.

glandium:
I rewrote the patch to not use unwrap, actually.

martinthomson (approved):
Perfect!

PR #1423, opened 2023-03-16
Bump minimum rust version to 1.59

glandium:


martinthomson:
@glandium if you are willing to review #1424, that might be a better way to manage this.

glandium:
Solved by #1424

PR #1424, opened 2023-03-16
Bump MSRV to 1.65.0

martinthomson:


neqo-common/src/hrtime.rs
```diff
@@ -390,7 +390,7 @@ mod test {
             let e = Instant::now();
             let actual = e - s;
             let lag = actual - d;
-            println!("sleep({:?}) \u{2192} {:?} \u{394}{:?}", d, actual, lag);
+            println!("sleep({d:?}) \u{2192} {actual:?} \u{394}{lag:?}");
```
glandium, line 393: I'm surprised. I thought this was new to 1.66...

glandium: Ah no, the clippy warning is new to 1.66, format strings capturing local variables is new to 1.58.

glandium:
I don't have a strong opinion, but it seems you could just as well go with bumping to 1.59 using the same patches.

martinthomson:
I don't know how to decide the MSRV, but if we go to 1.65 now, we might have to do this one time less in the future :)

PR #1434, opened 2023-05-05
Allow encryption to be enabled/disabled at runtime

KershawChang:
This is based on #1355 .
I also tried to address the review comments by making `Aead` a trait.


neqo-crypto/src/aead.rs
```diff
@@ -94,15 +128,26 @@ impl Aead {
...
+        if self.fuzzing_mode() {
+            write!(f, "[FUZZING AEAD]")
+        } else {
+            write!(f, "[AEAD Context]")
+        }
+    }
+}
+
+impl Aead for AeadImpl {
+    #[must_use]
```
martinthomson, line 144: I think that you can put the `#[must_use]` on the trait.

neqo-crypto/src/aead.rs
```diff
@@ -94,15 +128,26 @@ impl Aead {
...
+    #[allow(clippy::unused_self)]
```
martinthomson, line 145: Is this warning needed for a trait implementation?

neqo-crypto/src/aead.rs
```diff
@@ -157,10 +194,33 @@ impl Aead {
...
+) -> Res<Box<dyn Aead>> {
+    #[cfg(feature = "fuzzing")]
+    if fuzzing {
+        return Ok(Box::new(
+            AeadFuzzing::new(version, cipher, secret, prefix).unwrap(),
+        ));
     }
+
+    let aead = AeadImpl::new(version, cipher, secret, prefix)?;
+    Ok(Box::new(aead))
```
martinthomson, line 225: This is why I think that we should avoid the trait.  This isn't a huge deal, but the box isn't needed.

Let's say that you have fuzz::Aead with the defined interface (not a trait, just the necessary functions).  This is instantiated with `Option<real::Aead>`, such that if that is present, the various functions call back to the real function.

Then this module looks like this:

```rust
#[cfg(feature="fuzzing")]
pub use ...::fuzzing_aead::FuzzingAead;

#[cfg(not(feature="fuzzing"))]
pub use RealAead as Aead;

// ... 

impl RealAead {
pub fn new(...)
  // ...
}
}
```

The fuzzing AEAD can look like this:

```rust
pub struct FuzzingAead {
  real: Option<RealAead>;
}

pub fn new(...) {
    let real = if fuzzing {
      None
    } else {
       Some(RealAead::new(version, cipher, secret, prefix)?)
    };
    Self::new(real, version, cipher, secret, prefix)?
}
```

The advantage with that arrangement is that it ensures that you don't have a dynamic dispatch every time the AEAD is invoked (which is a fair bit).  The fuzzing feature will change what the rest of the code thinks is called "Aead", so that it ends up having different code when the feature is enabled.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -61,10 +63,8 @@ impl Aead {
             Err(Error::from(SEC_ERROR_BAD_DATA))
         }
     }
-}
 
-impl fmt::Debug for Aead {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "[FUZZING AEAD]")
+    fn fuzzing_mode(&self) -> bool {
+        true
```
martinthomson, line 68: This would be `self.real.is_none()`.

neqo-transport/src/connection/params.rs
```diff
@@ -282,6 +284,15 @@ impl ConnectionParameters {
         self
     }
 
+    pub fn get_fuzzing_mode(&self) -> bool {
```
martinthomson, line 287: is_fuzzing
```suggestion
    pub fn is_fuzzing(&self) -> bool {
```

neqo-transport/src/connection/params.rs
```diff
@@ -282,6 +284,15 @@ impl ConnectionParameters {
...
+        self.fuzzing_mode
+    }
+
+    pub fn fuzzing_mode(mut self, enable: bool) -> Self {
```
martinthomson, line 291:
```suggestion
    pub fn fuzzing(mut self, enable: bool) -> Self {
```

neqo-transport/src/packet/retry.rs
```diff
@@ -10,32 +10,33 @@ use crate::version::Version;
...
         version.label_prefix(),
     )
     .unwrap()
 }
-thread_local!(static RETRY_AEAD_29: RefCell<Aead> = RefCell::new(make_aead(Version::Draft29)));
-thread_local!(static RETRY_AEAD_V1: RefCell<Aead> = RefCell::new(make_aead(Version::Version1)));
-thread_local!(static RETRY_AEAD_V2: RefCell<Aead> = RefCell::new(make_aead(Version::Version2)));
+thread_local!(static RETRY_AEAD_29: RefCell<Box<dyn Aead>> = RefCell::new(make_aead(Version::Draft29)));
+thread_local!(static RETRY_AEAD_V1: RefCell<Box<dyn Aead>> = RefCell::new(make_aead(Version::Version1)));
+thread_local!(static RETRY_AEAD_V2: RefCell<Box<dyn Aead>> = RefCell::new(make_aead(Version::Version2)));
```
martinthomson, line 34: Even with this version of the code, you aren't providing a way to switch fuzzing on here, so you could always use the real AEAD, though I probably wouldn't bother.

neqo-transport/src/connection/params.rs
```diff
@@ -93,6 +94,7 @@ impl Default for ConnectionParameters {
             outgoing_datagram_queue: MAX_QUEUED_DATAGRAMS_DEFAULT,
             incoming_datagram_queue: MAX_QUEUED_DATAGRAMS_DEFAULT,
             fast_pto: FAST_PTO_SCALE,
+            fuzzing_mode: false,
```
martinthomson, line 97: Throughout, I have a preference for this to be called "fuzzing", rather than the more extra verbose "fuzzing_mode".

neqo-crypto/src/lib.rs
```diff
@@ -38,15 +38,9 @@ pub mod selfencrypt;
 mod ssl;
 mod time;
 
-#[cfg(not(feature = "fuzzing"))]
+pub use self::aead::create_aead;
```
martinthomson, line 41: On line 18 above, this can be:

```rust
#[cfg(feature = "fuzzing")]
mod aead_fuzzing;

**Review by martinthomson — CHANGES_REQUESTED** [review]
I like the trait-based version better, but adding a trait has costs.  Purely conditional compilation works better, I think.

File: neqo-crypto/src/aead.rs (hunk 2 above)
Comment (KershawChang) [reply neqo-crypto/src/aead.rs:145]: We'll go with non-trait version, so this comment will be ignored.

File: neqo-transport/src/packet/retry.rs (hunk 7 above)
Comment (KershawChang) [reply neqo-transport/src/packet/retry.rs:34]: According to #1355, this will be tackled in another patch, though I am not sure what do do. Anyway, I'll skip this for now.

File: neqo-crypto/src/aead.rs [hunk 10]
```diff
@@ -95,7 +101,7 @@ impl Aead {
         })
     }
 
-    /// Decrypt a plaintext.
+    /// Encrypt a plaintext.
```
martinthomson, line 104: ❤️

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -63,8 +89,12 @@ impl Aead {
...
 
-impl fmt::Debug for Aead {
+impl fmt::Debug for FuzzingAead {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "[FUZZING AEAD]")
+        if self.real.is_none() {
+            write!(f, "[FUZZING AEAD]")
+        } else {
+            write!(f, "[AEAD Context]")
+        }
```
martinthomson, line 98:
```suggestion
        if let Some(a) = self.real {
            a.fmt(f)
        } else {
            write!(f, "[FUZZING AEAD]")
        }
```

martinthomson (approved):
Nearly perfect, thanks.

If you want to take the suggestion and merge, go ahead.

neqo-transport/src/crypto.rs
```diff
@@ -94,7 +95,10 @@ impl Crypto {
             protocols,
             tls: agent,
             streams: Default::default(),
-            states: Default::default(),
+            states: CryptoStates {
+                fuzzing: fuzzing,
```
martinthomson, line 99:
```suggestion
                fuzzing,
```
Courtesy of clippy.

neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -7,29 +7,51 @@
...
+        Ok(Self { real })
     }
 
     #[must_use]
     pub fn expansion(&self) -> usize {
+        if let Some(aead) = &self.real {
+            return aead.expansion()
+        }
+
         FIXED_TAG_FUZZING.len()
```
martinthomson, line 41:
```suggestion
            aead.expansion()
        } else {
            FIXED_TAG_FUZZING.len()
        }
```

Courtesy of the formatter (which complained about the lack of semicolon).

PR #1435, opened 2023-05-09
Update CI dependencies

martinthomson:
Ubuntu version 20.04 -> 22.04
checkout action v2 -> v3
rustup 1.24.3 -> 1.26.0
LLVM 10 -> 14 (comes with the Ubuntu change)

The only one that is urgent is the action update, which will result in builds failing if we don't fix it.  The others are just so that we don't drop too far behind.

Definitely don't commit this until CI passes, which means that I need to fix the `EnumSet` thing.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,8 +37,8 @@ jobs:
             ssh \
             zlib1g-dev \
             --no-install-recommends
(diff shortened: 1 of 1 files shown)


martinthomson:
The error here in the CI build is addressed by #1436, so I'm merging anyway.

Issue #1432, opened 2023-05-02
Support SendOrder for WebTransportStream


KershawChang:
The w3c spec for `SendOrder` is [here](https://www.w3.org/TR/webtransport/#uni-stream-options).

PR #1437, opened 2023-05-18
Add support for WebTransport SendOrder to neqo (issue  #1432)

jesup:
No attempt at fairness within a sendorder (or for unordered streams) is attempted.

A few typos are corrected as well

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -232,7 +232,7 @@ possible if there is no buffered data.
 If a stream has buffered data it will be registered in the `streams_with_pending_data` queue and
 actual sending will be performed in the `process_sending` function call. (This is done in this way,
 i.e. data is buffered first and then sent, for 2 reasons: in this way, sending will happen in a
-single function,  therefore error handling and clean up is easier and the QUIIC layer may not be
+single function,  therefore error handling and clean up is easier and the QUIC layer may not be
 able to accept all data and being able to buffer data is required in any case.)
 
 The `send` and `send_data` functions may detect that the stream is closed and all outstanding data
@@ -920,7 +920,7 @@ impl Http3Connection {
         );
 
         // Call immediately send so that at least headers get sent. This will make Firefox faster, since
-        // it can send request body immediatly in most cases and does not need to do a complete process loop.
+        // it can send request body immediately in most cases and does not need to do a complete process loop.
         self.send_streams
             .get_mut(&stream_id)
             .ok_or(Error::InvalidStreamId)?
@@ -993,6 +993,17 @@ impl Http3Connection {
         // Stream may be already be closed and we may get an error here, but we do not care.
         conn.stream_stop_sending(stream_id, error)?;
         Ok(())
+    }
+
+    pub fn stream_set_sendorder(
+        &mut self,
+        conn: &mut Connection,
+        stream_id: StreamId,
+	sendorder: Option<i64>) -> Res<()> {
+        let send_stream = self.send_streams.get_mut(&stream_id)
+	                .ok_or(Error::InvalidStreamId)?;
+	send_stream.set_sendorder(conn, sendorder);
+	Ok(())
     }
 
     pub fn cancel_fetch(
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -754,6 +754,16 @@ impl Http3Client {
             - u64::try_from(Encoder::varint_len(session_id.as_u64())).unwrap())
     }
 
+    /// Sets the SendOrder for a given stream
+    /// # Errors
+    /// XXX No errors are returned
+    /// # Panics
+    /// This cannot panic.
+    pub fn webtransport_set_sendorder(&mut self, stream_id: StreamId, sendorder: i64) {
+	self.base_handler
+	    .stream_set_sendorder(&mut self.conn, stream_id, Some(sendorder)).ok();
+    }
+
     /// This function combines  `process_input` and `process_output` function.
     pub fn process(&mut self, dgram: Option<Datagram>, now: Instant) -> Output {
         qtrace!([self], "Process.");
--- a/neqo-http3/src/features/extended_connect/webtransport_session.rs
+++ b/neqo-http3/src/features/extended_connect/webtransport_session.rs
@@ -486,6 +486,10 @@ impl SendStream for Rc<RefCell<WebTransportSession>> {
         self.borrow_mut().has_data_to_send()
     }
 
+    fn set_sendorder(&mut self, _conn: &mut Connection, _sendorder: Option<i64>) {
+//	self.borrow_mut().set_sendorder(conn, sendorder);
+    }
+    
     fn stream_writable(&self) {}
 
     fn done(&self) -> bool {
--- a/neqo-http3/src/send_message.rs
+++ b/neqo-http3/src/send_message.rs
@@ -271,6 +271,10 @@ impl SendStream for SendMessage {
         self.stream.has_buffered_data()
     }
 
+    fn set_sendorder(&mut self, _conn: &mut Connection, _sendorder: Option<i64>) {
+//	self.stream.set_sendorder(conn, sendorder);
+    }
+    
     fn close(&mut self, conn: &mut Connection) -> Res<()> {
         self.state.fin()?;
         if !self.stream.has_buffered_data() {
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1911,9 +1911,13 @@ impl Connection {
             }
         }
 
+	// datagrams are best-effort and unreliable.  Let streams starve them for now
         // Check if there is a Datagram to be written
         self.quic_datagrams
             .write_frames(builder, tokens, &mut self.stats.borrow_mut());
+        if builder.is_full() {
+            return Ok(());
+        }
 
         let stats = &mut self.stats.borrow_mut().frame_tx;
 
@@ -1945,8 +1949,25 @@ impl Connection {
             return Ok(());
         }
 
-        self.streams
-            .write_frames(TransmissionPriority::Normal, builder, tokens, stats);
+        // WebTransport data (which is Normal) may have a SendOrder
+	// priority attached.  The spec states (6.3 write-chunk 6.1):
+
+        // If stream.[[SendOrder]] is null then this sending MUST NOT
+	// starve except for flow control reasons or error.  If
+	// stream.[[SendOrder]] is not null then this sending MUST starve
+	// until all bytes queued for sending on WebTransportSendStreams
+	// with a non-null and higher [[SendOrder]], that are neither
+	// errored nor blocked by flow control, have been sent.
+
+	// So data without SendOrder goes first.   Then the highest priority
+	// SendOrdered streams.   Round-robining the data at the same priority
+	// isn't required (currently) by the spec, but would be good to do in the future.
+	// "ordered()" returns a chained hash that iterates all the streams in the order
+	// described above.
+	let stream_ids = self.streams.ordered().copied().collect::<Vec<StreamId>>();
+	for stream_id in stream_ids.iter() {
+	    self.streams.get_send_stream_mut(*stream_id).unwrap().write_frames(TransmissionPriority::Normal, builder, tokens, stats);
+        }
         if builder.is_full() {
             return Ok(());
         }
@@ -2939,6 +2960,17 @@ impl Connection {
             .get_send_stream_mut(stream_id)?
             .set_priority(transmission, retransmission);
         Ok(())
+    }
+
+    /// Set the SendOrder of a stream.  Re-enqueues to keep the ordering correct
+    /// # Errors
+    /// `InvalidStreamId` the stream does not exist.
+    pub fn stream_sendorder(
+        &mut self,
+        stream_id: StreamId,
+        sendorder: Option<i64>,
+    ) {
+	self.streams.set_sendorder(stream_id, sendorder).ok();
     }
 
     /// Send data on a stream.
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -18,6 +18,7 @@ use std::{
 
 use indexmap::IndexMap;
 use smallvec::SmallVec;
+use std::hash::{Hash, Hasher};
 
 use neqo_common::{qdebug, qerror, qinfo, qtrace, Encoder, Role};
 
@@ -556,7 +557,21 @@ pub struct SendStream {
     priority: TransmissionPriority,
     retransmission_priority: RetransmissionPriority,
     retransmission_offset: u64,
+    sendorder: Option<i64>,
 }
+
+impl Hash for SendStream {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+       self.stream_id.hash(state)
+    }
+}
+
+impl PartialEq for SendStream {
+    fn eq(&self, other: &Self) -> bool {
+        self.stream_id == other.stream_id
+    }
+}
+impl Eq for SendStream {}
 
 impl SendStream {
     pub fn new(
@@ -575,11 +590,26 @@ impl SendStream {
             priority: TransmissionPriority::default(),
             retransmission_priority: RetransmissionPriority::default(),
             retransmission_offset: 0,
+	    sendorder: None,
         };
         if ss.avail() > 0 {
             ss.conn_events.send_stream_writable(stream_id);
         }
         ss
+    }
+
+    pub fn write_frames(
+        &mut self,
+        priority: TransmissionPriority,
+        builder: &mut PacketBuilder,
+        tokens: &mut Vec<RecoveryToken>,
+        stats: &mut FrameStats,
+    ) {
+        qtrace!("write STREAM frames at priority {:?}", priority);
+	if !self.write_reset_frame(priority, builder, tokens, stats) {
+          self.write_blocked_frame(priority, builder, tokens, stats);
+          self.write_stream_frame(priority, builder, tokens, stats);
+        }
     }
 
     pub fn set_priority(
@@ -589,6 +619,18 @@ impl SendStream {
     ) {
         self.priority = transmission;
         self.retransmission_priority = retransmission;
+    }
+
+    pub fn sendorder(&self) -> Option<i64> {
+       self.sendorder
+    }
+
+    pub fn set_sendorder(
+        &mut self,
+	sendorder: Option<i64>,
+    ) {
+        self.sendorder = sendorder;
+	// Caller must remove and re-insert into ordered_streams
     }
 
     /// If all data has been buffered or written, how much was sent.
@@ -1116,8 +1158,12 @@ impl SendStreams {
         self.0.clear()
     }
 
-    pub fn clear_terminal(&mut self) {
-        self.0.retain(|_, stream| !stream.is_terminal())
+    pub fn get_terminal(&self) -> Vec<StreamId> {
+	self.0.iter().filter_map(|(stream_id, stream)| stream.is_terminal().then_some(*stream_id)).collect()
+    }
+
+    pub fn remove(&mut self, stream_id: &StreamId) {
+	self.0.remove(stream_id);
     }
 
     pub(crate) fn write_frames(
--- a/neqo-transport/src/streams.rs
+++ b/neqo-transport/src/streams.rs
@@ -5,6 +5,8 @@
 // except according to those terms.
 
 // Stream management for a connection.
+use std::collections::BTreeMap;
+use std::collections::HashSet;
 
 use crate::{
     fc::{LocalStreamLimits, ReceiverFlowControl, RemoteStreamLimits, SenderFlowControl},
@@ -20,6 +22,32 @@ use crate::{
 };
 use neqo_common::{qtrace, qwarn, Role};
 use std::{cell::RefCell, rc::Rc};
+use std::cmp::Ordering;
+
+struct StreamOrder {
+    sendorder: Option<i64>,
+}
+
+impl Ord for StreamOrder {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.sendorder.cmp(&other.sendorder)
+    }
+}
+
+impl PartialOrd for StreamOrder {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for StreamOrder {
+    fn eq(&self, other: &Self) -> bool {
+        self.sendorder == other.sendorder
+    }
+}
+
+impl Eq for StreamOrder {}
+
 
 pub struct Streams {
     role: Role,
@@ -31,6 +59,21 @@ pub struct Streams {
     local_stream_limits: LocalStreamLimits,
     pub(crate) send: SendStreams,
     pub(crate) recv: RecvStreams,
+    // What we really want is a Priority Queue that we can do arbitrary removes
+    // from (so we can reprioritize). BinaryHeap doesn't work, because there's no
+    // remove().  BTreeMap doesn't work, since you can't duplicate keys.
+    // PriorityQueue does have what we need, except for an ordered iterator that doesn't consume the queue.
+
+    // So we use a HashSet for the unordered streams (that's usually all of
+    // them), and then a BTreeMap of an entry for each SendOrder value, and
+    // for each of those entries a HashSet of the stream_ids at that
+    // sendorder.  In most cases (such as stream-per-frame), there will be
+    // a single stream at a given sendorder.
+    
+    // These both store stream_ids, which need to be looked up in 'send'.  This avoids the complexity
+    // of trying to hold references to the Streams which are owned by the send IndexMap.
+    ordered_send: BTreeMap<StreamOrder, HashSet<StreamId>>,
+    unordered_send: HashSet<StreamId>, // streams with no SendOrder set
 }
 
 impl Streams {
@@ -58,6 +101,8 @@ impl Streams {
             local_stream_limits: LocalStreamLimits::new(role),
             send: SendStreams::default(),
             recv: RecvStreams::default(),
+	    ordered_send: BTreeMap::new(),
+	    unordered_send: HashSet::new(),
         }
     }
 
@@ -68,6 +113,8 @@ impl Streams {
     pub fn zero_rtt_rejected(&mut self) {
         self.send.clear();
         self.recv.clear();
+	self.ordered_send.clear();
+	self.unordered_send.clear();
         debug_assert_eq!(
             self.remote_stream_limits[StreamType::BiDi].max_active(),
             self.tps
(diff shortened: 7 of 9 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2941,6 +2962,17 @@ impl Connection {
         Ok(())
     }
 
+    /// Set the SendOrder of a stream.  Re-enqueues to keep the ordering correct
+    /// # Errors
+    /// `InvalidStreamId` the stream does not exist.
```
valenting, line 2967: This comment doesn't seem accurate. This method currently returns void

jesup:
I had added the custom Ord impl to switch the order, but didn't actually modify it to switch the order... :-)
Added logging for sendorder changes

KershawChang:
I think we should have some tests in neqo as well.
It should be easier to test this in neqo than writing a wpt test.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
     assert!(fin3);
 }
 
+#[cfg(test)]
```
martinthomson, line 116: You don't need to use `#[cfg(test)]` here.  This is in a module that already has that label.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    // Check the last bucket
+    for id in &should_recv_streams {
+	// check if this is the right id
+	assert!(recv_streams.contains(&id));
+    }
+
+    assert!(stream_ids.next().is_none());
+}
+
+macro_rules! sendorder_tests {
```
martinthomson, line 206: I have an allergy to macros that are just functions.  You should be able to do something like this:

```rust
#[test]
// TODO pick a nicer name
fn sendorder_0() {
  sendorder_test(&[None, Some(1), Some(2), Some(3)]);
}
```

The overall shape isn't that much different (other than the test function taking `&[Option<SendOrder>]` rather than a `Vec`), but you don't have a macro.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+	ordered.push(IdEntry { sendorder: *sendorder, stream_id: id });
+	client.streams.set_sendorder(id, sendorder.sendorder).ok();
+    }
+    // Write some data to all the streams
+    for stream_id in streams {
+	client.stream_send(stream_id, &[6; 100]).unwrap();
+    }
+
+    // Sending this much takes a few datagrams.
+    let mut datagrams = vec![];
```
martinthomson, line 146: Use `Vec::new()`.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    let mut out = client.process_output(now());
+    while let Some(d) = out.dgram() {
+        datagrams.push(d);
+        out = client.process_output(now());
+    }
```
martinthomson, line 151: There's a potential trap here that is worth a note.  This test uses an RTT of zero, which means that this code is able to be simpler and not have to worry about pacing.

(I would also factor this out into a separate function.)

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    assert_eq!(*client.state(), State::Confirmed);
+
+    qdebug!("---- server receives");
+    for (d_num, d) in datagrams.into_iter().enumerate() {
+        let out = server.process(Some(d), now());
+        assert_eq!(
+            out.as_dgram_ref().is_some(),
+            (d_num + 1) % usize::try_from(DEFAULT_ACK_PACKET_TOLERANCE + 1).unwrap() == 0
+        );
```
martinthomson, line 160: We're not really testing for ACKs, so maybe you don't need to worry about checking this.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    // streams should arrive in priority order, not order of creation, if sendorder prioritization
+    // is working correctly
+    let mut recv_streams = Vec::<StreamId>::new();
+    let mut should_recv_streams =  Vec::<StreamId>::new();
+    // walk the ordered list of streams, and handle random ordering within a sendorder
+    let mut last_sendorder: StreamOrder = StreamOrder{sendorder: None};
+    // BinaryHeap's sorted iterator is nightly only for now
+    // the ordering is inverted, however!
+    let mut ordered_vec = ordered.into_sorted_vec();
+    ordered_vec.reverse();
```
martinthomson, line 179: It seems like maybe you could avoid using BinaryHeap here and just rely on `<&mut []>::sort` or `<&mut []>::sort_by` instead.  (I think that you will find that `Option` implements `Ord` in a way that will fit nicely here.)

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    for identry in ordered_vec.iter() {
```
martinthomson, line 180: There might be an easier way:

Collection 1 is the declared send order, taken from the input list.  It consists of a Vec containing tuples of `(StreamId, Option<SendOrder>)`.  (Your IdEntry, though I wouldn't bother declaring a type for it.)

Collection 2 is the arrival order.  It consists of a map from `StreamId` to a simple counter that is incremented for each piece of data that arrives (without overwriting as some data might arrive twice).  (There is a clever way of constructing this.  You take your filter_map call above and then add `.enumerate().map(|(a, b)| (b, a)).collect::<HashMap<_, _>>()`)

The first collection is sorted by the SendOrder, but if two items have equal SendOrder, you use the arrival order from the second list to determine order (use `Vec::sort_unstable_by`).

Then you can check that the order of the stream identifiers in the first and second collections are equal.

neqo-transport/src/streams.rs
```diff
@@ -24,13 +24,22 @@ use neqo_common::{qtrace, qwarn, Role};
...
-    sendorder: Option<i64>,
+#[derive(Copy, Clone)]
+pub struct StreamOrder {
+    pub sendorder: Option<i64>,
 }
 
-// We want highest to lowest
+// We want highest to lowest, with None being higher than any value
 impl Ord for StreamOrder {
     fn cmp(&self, other: &Self) -> Ordering {
```
martinthomson, line 34: It should be possible to do this:

```rust
struct SendOrder(Option<i64>);

impl Ord for StreamOrder {
  fn cmp(&self, other: &Self) {
    if let (Some(a), Some(b)) = (self.0, other.0) {
      // We want reverse order (high to low) when both values are specified.
      other.0.cmp(self.0)
    } else {
      self.0.cmp(other.0)
    }
  }
}
```

see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6ab1a1e00f3baa65e47f454be320d20

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,121 @@ fn transfer() {
...
+    let mut server = default_server();
+    connect_force_idle(&mut client, &mut server);
+
+    qdebug!("---- client sends");
+    // open all streams and set the sendorders
+    let mut ordered = BinaryHeap::new();
+    let mut streams = Vec::<StreamId>::new();
+    for sendorder in order_of_sendorder {
+	streams.push(client.stream_create(StreamType::UniDi).unwrap());
+	let id = *(streams.last()).unwrap();
```
martinthomson, line 136: I think that you want this instead:

```rust
let id = client.stream_create(StreamType::UniDi).unwrap();
streams.push(id);
```

neqo-transport/src/streams.rs
```diff
@@ -58,6 +111,8 @@ impl Streams {
             local_stream_limits: LocalStreamLimits::new(role),
             send: SendStreams::default(),
             recv: RecvStreams::default(),
+	    ordered_send: BTreeMap::new(),
+	    unordered_send: HashSet::new(),
```
KershawChang, line 115: I think these two members should be in `SendStreams`.
I prefer putting all send order logic in `SendStream`.

neqo-transport/src/connection/tests/stream.rs
jesup: Cut-and-paste from the next test

neqo-transport/src/send_stream.rs
```diff
@@ -575,13 +590,28 @@ impl SendStream {
             priority: TransmissionPriority::default(),
             retransmission_priority: RetransmissionPriority::default(),
             retransmission_offset: 0,
+	    sendorder: None,
```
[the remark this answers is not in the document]
jesup: Doesn't that imply separate sendOrders for each of High/Normal/Low?   We only have one

neqo-transport/src/connection/tests/stream.rs
jesup: Yes.   I had the first, and then needed the id... fixed.

neqo-transport/src/streams.rs
jesup: With some tweaks

neqo-transport/src/connection/tests/stream.rs
jesup: Perhaps this is better; it would be shorter.   I also think (though perhaps it's just because I'm new to Rust) that this will be harder to understand or modify, if I understand correctly.
I've been trying to code this up, and I can't figure out how to do "The first collection is sorted by the SendOrder, but if two items have equal SendOrder, you use the arrival order from the second list to determine order (use Vec::sort_unstable_by)."
I see https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable_by, but I don't see how to go  from that to a working fancy sorter.

neqo-transport/src/connection/tests/stream.rs
jesup: Though if I make it a function I might want to keep it

neqo-transport/src/connection/mod.rs
```diff
@@ -1911,9 +1911,13 @@ impl Connection {
             }
         }
 
+	// datagrams are best-effort and unreliable.  Let streams starve them for now
         // Check if there is a Datagram to be written
         self.quic_datagrams
             .write_frames(builder, tokens, &mut self.stats.borrow_mut());
+        if builder.is_full() {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
jesup: I've set this up as a separate pr

neqo-http3/src/connection.rs
```diff
@@ -995,6 +995,17 @@ impl Http3Connection {
         Ok(())
     }
 
+    pub fn stream_set_sendorder(
+        &mut self,
+        conn: &mut Connection,
+        stream_id: StreamId,
+	sendorder: Option<i64>) -> Res<()> {
```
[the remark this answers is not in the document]
jesup: I want to leave it possible to set sendorder to None - we've just added the ability to set sendOrder for an existing stream, and that will let us move a stream from a sendOrder to no sendOrder (stream.sendOrder = null;)

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
jesup: We effectively have that now, since we're handling this all inside of SendStreams - we get write_frames invoked for each TransmissionPriority, and for each TransmissionPriority we look at the streams according to SendOrder

neqo-http3/src/connection_client.rs
```diff
@@ -754,6 +754,16 @@ impl Http3Client {
...
     }
 
+    /// Sets the SendOrder for a given stream
+    /// # Errors
+    /// XXX No errors are returned
+    /// # Panics
+    /// This cannot panic.
+    pub fn webtransport_set_sendorder(&mut self, stream_id: StreamId, sendorder: i64) {
+	self.base_handler
+	    .stream_set_sendorder(&mut self.conn, stream_id, Some(sendorder)).ok();
```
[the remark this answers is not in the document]
jesup: Currently it can't

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1106,174 @@ impl ::std::fmt::Display for SendStream {
...
+		self.regular.push(id);
+		return;
+	    }
+	}
+	self.insert_regular(&id);
+    }
+
+    pub fn set_sendorder(&mut self, stream_id: StreamId, sendorder: Option<SendOrder>) {
+	// don't grab stream here; causes borrow errors
+        let old_sendorder = self.map.get(&stream_id).unwrap().sendorder();
```
KershawChang, line 1165: Should the stream state between Necko and Neqo fall out of sync, this unwrap function could trigger a panic.
Could you make this function return `Res()`?
Please also check other places where `unwrap` is used.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,92 @@ fn transfer() {
...
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+struct IdEntry {
+    sendorder: StreamOrder,
+    stream_id: StreamId,
+}
+
+#[cfg(test)]
+// tests stream sendorder priorization
+fn sendorder_test(order_of_sendorder: &[Option<SendOrder>]) {
```
KershawChang, line 124: Can we have a similar test in `neqo-http3/src/features/extended_connect/tests/webtransport/streams.rs`?

KershawChang (commented):
This looks good to me, but I think this still needs @martinthomson 's approval.

neqo-transport/src/streams.rs
```diff
@@ -29,6 +63,8 @@ pub struct Streams {
     receiver_fc: Rc<RefCell<ReceiverFlowControl<()>>>,
     remote_stream_limits: RemoteStreamLimits,
     local_stream_limits: LocalStreamLimits,
+    // note: SendStreams also keeps track of SendOrder and provides an iterator
+    // based on SendOrder
```
martinthomson, line 67:
```suggestion
```

Encapsulation means that this code doesn't need to care about those details.

neqo-transport/src/streams.rs
```diff
@@ -229,7 +264,7 @@ impl Streams {
             }
         }
 
-        self.send.write_frames(priority, builder, tokens, stats);
+	self.send.write_frames(priority, builder, tokens, stats)
```
martinthomson, line 267: Add the semi-colon back.

```suggestion
	self.send.write_frames(priority, builder, tokens, stats);
```

neqo-transport/src/send_stream.rs
```diff
@@ -1128,16 +1284,41 @@ impl SendStreams {
...
+	// starve except for flow control reasons or error.  If
+	// stream.[[SendOrder]] is not null then this sending MUST starve
+	// until all bytes queued for sending on WebTransportSendStreams
+	// with a non-null and higher [[SendOrder]], that are neither
+	// errored nor blocked by flow control, have been sent.
+
+	// So data without SendOrder goes first.   Then the highest priority
+	// SendOrdered streams.   Round-robining the data at the same priority
+	// isn't required (currently) by the spec, but would be good to do in the future.
+	// "ordered_iter()" returns a chained hash that iterates all the streams in the order
```
martinthomson, line 1300: Check that your comments are up to date.

neqo-transport/src/send_stream.rs
martinthomson: Yeah, this is going to need Res, if only because the code that calls this isn't always in this crate..

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1106,174 @@ impl ::std::fmt::Display for SendStream {
...
+
+	// We could add assertions, or we can try to insert at the end and
+	// if not fall back to binary-search insertion
+	if let Some(last) = self.regular.last() {
+	    if id > *last {
+		self.regular.push(id);
+		return;
+	    }
+	}
+	self.insert_regular(&id);
```
martinthomson, line 1160:
```suggestion
	match self.regular.last() {
	  Some(last) if id > *last => self.regular.push(id);
	  _ => self.insert_regular(&id);
	}
```
There you go haxx.  You can also do

```suggestion
	if matches!(self.regular.last(), Some(last) if id > *last)) {
	  self.regular.push(id);
	} else {
	  self.insert_regular(&id);
	}
```

neqo-transport/src/streams.rs
```diff
@@ -20,6 +19,41 @@ use crate::{
...
+
+#[derive(Copy, Clone)]
+pub struct StreamOrder {
+    pub sendorder: Option<SendOrder>,
+}
+
+// We want highest to lowest, with None being higher than any value
+impl Ord for StreamOrder {
+  fn cmp(&self, other: &Self) -> Ordering {
+    if let (Some(_), Some(_)) = (self.sendorder, other.sendorder) {
```
martinthomson, line 34: Try

```suggestion
    if self.sendorder.is_some() && other.sendorder.is_some() {
```

neqo-transport/src/connection/tests/stream.rs
martinthomson: I think that transport tests will suffice.  Though a much simpler test there might test that the plumbing works.  Say make a webtransport, write to two streams, set sendorder, check that sendorder overrides stream ID order.

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -486,6 +486,10 @@ impl SendStream for Rc<RefCell<WebTransportSession>> {
         self.borrow_mut().has_data_to_send()
     }
 
+    fn set_sendorder(&mut self, _conn: &mut Connection, _sendorder: Option<SendOrder>) {
+	// Not relevant on session
+    }
```
martinthomson, line 491: I think that we should let this pass through.  We will simply not call it in gecko (who knows, maybe this makes sense at some point).

neqo-http3/src/connection_client.rs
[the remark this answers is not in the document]
martinthomson: It should: because someone could make a mistake and choose a stream that doesn't exist yet.

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1106,174 @@ impl ::std::fmt::Display for SendStream {
...
+    // Note: the BTreeMap is keyed by SendOrder, but it would normally iterate
+    // smallest-to-largest, and we want the opposite.
+    #[allow(dead_code)]
+    pub fn ordered_iter(&mut self) -> impl Iterator<Item = &StreamId> {
+	self.regular.iter().chain(self.sendordered.values().rev().flatten())
     }
 
-    pub fn clear_terminal(&mut self) {
-        self.0.retain(|_, stream| !stream.is_terminal())
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
jesup: I wanted to have the method to iterate everything available, even though I ended up having to inline it.   I can remove it, of course

neqo-http3/src/features/extended_connect/webtransport_session.rs
jesup: Pass through to where?

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1106,174 @@ impl ::std::fmt::Display for SendStream {
...
+    // them), and then a BTreeMap of an entry for each SendOrder value, and
+    // for each of those entries a HashSet of the stream_ids at that
+    // sendorder.  In most cases (such as stream-per-frame), there will be
+    // a single stream at a given sendorder.
+
+    // These both store stream_ids, which need to be looked up in 'map'.
+    // This avoids the complexity of trying to hold references to the
+    // Streams which are owned by the IndexMap.
+    sendordered: BTreeMap<SendOrder, HashSet<StreamId>>,
+    regular: Vec<StreamId>, // streams with no SendOrder set, sorted in stream_id order
```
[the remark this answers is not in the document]
jesup: Possibly, but:  In stream-per-frame (and perhaps most instances where sendorder is used), all or almost all the streams will have sendorders, and so we'll iterate over map but ignore almost or all the iterations.   That seems a higher overhead (every time we're sending packets) than the maintenance overhead.

neqo-http3/src/features/extended_connect/webtransport_session.rs
martinthomson: I am suggesting that you change the send order, rather than do nothing

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
martinthomson: Hmm. Having a count of the number of unordered streams would allow us to add `.take(count)` when iterating, which would shortcut very nicely. WDYT?

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1106,174 @@ impl ::std::fmt::Display for SendStream {
...
+    // So we use a sorted Vec<> for the regular streams (that's usually all of
+    // them), and then a BTreeMap of an entry for each SendOrder value, and
+    // for each of those entries a HashSet of the stream_ids at that
+    // sendorder.  In most cases (such as stream-per-frame), there will be
+    // a single stream at a given sendorder.
+
+    // These both store stream_ids, which need to be looked up in 'map'.
+    // This avoids the complexity of trying to hold references to the
+    // Streams which are owned by the IndexMap.
+    sendordered: BTreeMap<SendOrder, HashSet<StreamId>>,
```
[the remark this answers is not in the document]
jesup: I'll give that a try tomorrow and look at the perf

jesup:
Updated for almost all the comments; will look at using Vec instead of HashSet for each sendordered group tomorrow

neqo-transport/src/send_stream.rs
```diff
@@ -1163,23 +1163,20 @@ impl SendStreams {
...
 	    // we have to remove it from the list it was in, and reinsert it with the new
 	    // sendorder key
+	    let mut vec : &mut Vec<StreamId>;
             if let Some(old) = old_sendorder {
-		self.sendordered.get_mut(&old).unwrap().remove(&stream_id);
+		vec = self.sendordered.get_mut(&old).unwrap();
             } else {
-		Self::remove_regular(&mut self.regular, &stream_id);
+		vec = &mut self.regular;
 	    }
```
martinthomson: Better:

```rust
let v = if let Some(old) = old_sendorder {
  self.sendordered.get_mut(&old).unwrap()
} else {
  &mut self.regular
};
```

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
jesup: I'm not sure I understand your suggestion here

neqo-http3/src/features/extended_connect/webtransport_session.rs
jesup: Sessions don't have a sendorder to set, so I'm still confused by your suggestion.   Unless you mean "set the sendorder of all streams to N", which would be unhelpful and there's no reason for it.

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
martinthomson: If, rather than maintaining a collection of streams that are unordered, you merely kept a count, then in the two cases we think most likely:

1. All streams are unordered.  You iterate over the main `IndexMap` directly, but `.take(x)` where `x = self.map.len()`, so you just iterate the map.
2. All streams are ordered.  You iterate over the main `IndexMap`, but `.take(0)`, which turns into a no-op.

The advantage is that you don't have an indirection for the first iteration; you iterate the collection directly.  In the second, you short-circuit that iteration (because iterators are lazy and `.take(0)` doesn't hit the underlying iterator at all.  It's only when you have a mix that you end up paying extra, but that's not a situation we think likely.  If there are a few streams that are unordered, then you'll have a few to track, but maybe you can count the number of unordered streams with unsent data so that you have to hit the collection less often.

neqo-http3/src/features/extended_connect/webtransport_session.rs
KershawChang: Maybe add a debug assertion here?
I assume this function should not be called at all.

KershawChang (approved):

KershawChang:
@jesup please also fix all format and clippy errors. Thanks.

neqo-http3/src/send_message.rs
```diff
@@ -271,6 +271,10 @@ impl SendStream for SendMessage {
         self.stream.has_buffered_data()
     }
 
+    fn set_sendorder(&mut self, _conn: &mut Connection, _sendorder: Option<SendOrder>) {
+	// Not relevant for SendMessage
```
martinthomson, line 275: Again.  `self.stream.set_sendorder(conn, sendorder)` seems workable here, even if we don't end up using it.  (Yes, this requires an additional step to implement that on `BufferedStream`, but it's a trivial thing.)

Yes, we're building for webtransport, but this type has a direct relationship with an underlying stream and a random noop is a surprise I'd prefer to avoid.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +113,92 @@ fn transfer() {
     assert!(fin3);
 }
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+struct IdEntry {
+    sendorder: StreamOrder,
+    stream_id: StreamId,
+}
+
+#[cfg(test)]
```
martinthomson, line 122:
```suggestion
```

neqo-transport/src/streams.rs
```diff
@@ -20,6 +19,41 @@ use crate::{
...
+pub struct StreamOrder {
+    pub sendorder: Option<SendOrder>,
+}
+
+// We want highest to lowest, with None being higher than any value
+impl Ord for StreamOrder {
+  fn cmp(&self, other: &Self) -> Ordering {
+    if self.sendorder.is_some() && other.sendorder.is_some() {
+      // We want reverse order (high to low) when both values are specified.
+      other.sendorder.cmp(&self.sendorder)
```
martinthomson, line 36: The alternative (I keep coming up with these, sorry) might be better:

```suggestion
    if let (Some(s), Some(o)) = (&self.sendorder, &other.sendorder) {
      // We want reverse order (high to low) when both values are specified.
      o.cmp(s)
```

neqo-transport/src/streams.rs
```diff
@@ -386,15 +425,16 @@ impl Streams {
...
+		let stream = SendStream::new(
                         new_id,
                         send_limit,
                         Rc::clone(&self.sender_fc),
                         self.events.clone(),
-                    ),
+                    );
+                self.send.insert(
+                    new_id, stream,
                 );
```
martinthomson, line 436: Is this really what cargo fmt produces?  I'm surprised that it's breaking across lines like this.

neqo-transport/src/connection/mod.rs
```diff
@@ -2941,6 +2945,17 @@ impl Connection {
         Ok(())
     }
 
+    /// Set the SendOrder of a stream.  Re-enqueues to keep the ordering correct
+    /// # Errors
+    /// None
```
martinthomson, line 2950:
```suggestion
```

neqo-http3/src/connection.rs
```diff
@@ -995,6 +995,20 @@ impl Http3Connection {
         Ok(())
     }
 
+    /// Set the stream SendOrder.
+    /// # Errors
+    /// Returns InvalidStreamId if the stream id doesn't exist
```
martinthomson, line 1000: Clippy will complain here.

```suggestion
    /// Set the stream `SendOrder`.
    /// # Errors
    /// `InvalidStreamId` if the stream id doesn't exist.
```

PR #1440, opened 2023-05-23
chore: Cleanups for `write_appdata_frames`

### LINKED ISSUE #1439: Datagrams are sent at a higher priority than any other data
Currently WebTransport datagrams are sent before any other data, even TransmissionPriority::Critical.   Probably they should be send either before user streams (before TransmissionPriority::Normal) or after it.  The spec does not talk to relative priority of datagrams, but they're best-effort.

jesup:
Resolves #1439

We could drop it below Normal as well


neqo-transport/src/connection/mod.rs
```diff
@@ -1945,6 +1941,12 @@ impl Connection {
             return Ok(());
         }
 
+        // Check if there is a Datagram to be written
+	// Currently we're giving them priority over user streams; they could be moved
+	// to after them (after Normal)
+        self.quic_datagrams
+            .write_frames(builder, tokens, &mut self.stats.borrow_mut());
+
```
martinthomson, line 1949: I think that you want two changes here.

1. use `stats` in place of `&mut self.stats.borrow_mut()` as this is already borrowed (I'm somewhat surprised that this compiled)
2. Check `if builder.is_full()`.

Are there no tests?  Maybe there should be...  (Sorry to do that to you.)

martinthomson (changes requested):

martinthomson (commented):
Pinging @jesup 

Please also consider putting datagrams below Normal priority streams.

jesup:
> Pinging @jesup
> 
> Please also consider putting datagrams below Normal priority streams.

so... should I do that, or should I not?   Mixing datagrams and normal streams will not be the norm, I suspect.   If we want ultimate flexibility, we could allow for setting the priority of datagrams via a sendOrder value, or a before-vs-after-normal setting.  But I don't think we need that.   
A related question is the spec - should the priority of datagrams relative to normal be speced in some way?
For now I've kept it right above normal, but it's trivial to move it below.

neqo-transport/src/connection/mod.rs
```diff
@@ -1911,60 +1911,71 @@ impl Connection {
...
+            }
         }
 
-        self.streams
-            .write_frames(TransmissionPriority::Normal, builder, tokens, stats);
+        // Check if there is a Datagram to be written
+        // Currently we're giving them priority over user streams; they could be moved
+        // to after them (after Normal)
+        self.quic_datagrams
+            .write_frames(builder, tokens, &mut self.stats.borrow_mut());
```
martinthomson, line 1956: I don't understand why you needed to scope the borrow of `self.stats`.  Is there something I'm missing here?

That is, why doesn't the following work?

```rust
        self.quic_datagrams.write_frames(builder, tokens, stats);

File: neqo-transport/src/#send_stream.rs# [hunk 3]
```diff
@@ -0,0 +1,2294 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
martinthomson, line 1: You probably didn't intend to add this one.

martinthomson (changes requested):
If you are able, a test is what this change requires.

neqo-transport/src/connection/mod.rs
jesup: doesn't build: expected `&mut Stats`, found `&mut StatsCell`

jesup: stats is `let stats = &mut self.stats.borrow_mut().frame_tx`, which is not what quic_dataframes.write_frames expects (i.e. not the `frame_tx` part)

jesup: Your solution in PR #1447 is fine as an alternate way to deal with this

KershawChang:
@jesup this patch seems to have conflicts. Could you rebase again?

neqo-transport/src/connection/mod.rs
```diff
@@ -1911,60 +1911,71 @@ impl Connection {
...
-        self.crypto
-            .write_frame(PacketNumberSpace::ApplicationData, builder, tokens, stats)?;
-        if builder.is_full() {
-            return Ok(());
-        }
-        self.new_token.write_frames(builder, tokens, stats)?;
-        if builder.is_full() {
-            return Ok(());
-        }
+        {
```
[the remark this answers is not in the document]
jesup: To resolve the borrow issues for stats.   #1447 solves it in a different manner

martinthomson:
Pinging @jesup.  (I realize that we've left a few items open for probably too long and I'm checking in again.)

larseggert:
@jesup what is needed to land this?

larseggert:
Now that I looked at the code, this actually got fixed already in #1447 and the test got added in #1450. I did make some cleanups while rebasing this, so I think we should still merge it.

larseggert:
Actually, I think I misunderstood something, because we didn't merge #1447. 

But I still don't understand what the change in this PR compared to main actually is?

neqo-transport/src/connection/mod.rs
```diff
@@ -2003,6 +1996,10 @@ impl Connection {
 
         #[cfg(test)]
         {
+            // Need to check if the previous `write_frames` call filled the packet.
+            if builder.is_full() {
+                return;
+            }
```
martinthomson, line 2002: Maybe we don't need to for tests.  The test code can check for overflow... and even overflow if it wants.

martinthomson (approved):

martinthomson (approved):

PR #1441, opened 2023-06-06
Add support for configuring grease support

jesup:
This is needed since aioquic, which is used in the WPT test server (and probably other places), we need to be able to disabling greasing when needed.

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -28,6 +28,7 @@ const DEFAULT_ACK_RATIO: u8 = 4 * ACK_RATIO_SCALE;
 /// The local value for the idle timeout period.
 const DEFAULT_IDLE_TIMEOUT: Duration = Duration::from_secs(30);
 const MAX_QUEUED_DATAGRAMS_DEFAULT: usize = 10;
+const DEFAULT_GREASE: bool = true;
 
 /// What to do with preferred addresses.
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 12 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -293,6 +296,15 @@ impl ConnectionParameters {
         self
     }
 
+    pub fn get_grease(&self) -> bool {
```
martinthomson, line 299: Rust convention is to name Boolean getters `is_foo()`.  So this would be `is_greasing` or `is_greased` or `is_greasy` or something.

neqo-transport/src/connection/params.rs
```diff
@@ -28,6 +28,7 @@ const DEFAULT_ACK_RATIO: u8 = 4 * ACK_RATIO_SCALE;
 /// The local value for the idle timeout period.
 const DEFAULT_IDLE_TIMEOUT: Duration = Duration::from_secs(30);
 const MAX_QUEUED_DATAGRAMS_DEFAULT: usize = 10;
+const DEFAULT_GREASE: bool = true;
```
martinthomson, line 31: Don't bother with this.  Just turn it on by default inline.

martinthomson (commented):
I wonder whether you need a `ClientConfig` struct for this.  I don't like making generic stuff for no good reason, but the bare boolean is a bit rough on the eyes.

Please unstack the pull request, so we can decouple it.

PR #1443, opened 2023-06-08
Implement stream fairness when sending data

jesup:
This implements fairness when sending data for multiple streams; it does this by remembering the position in the iteration of each ordering group, such that on the next time we iterate that group we start at the stream that would have been next in the previous iteration.

It's not 100% strict in fairness with regards to things like added or removed streams - it keeps the position as an index, and removing or adding streams between iterations of the group does not update the index.   Also, if streams are removed and the index points past the end of the group, we start at the first stream.


neqo-transport/src/connection/tests/stream.rs
```diff
@@ -199,6 +200,90 @@ fn sendorder_4() {
...
+    let source: [u64; 1] = [0];
+    let result: [u64; 1] = [0];
+    fairness_test(&source.map(|x| StreamId(x)), 1, usize::MAX, &result.map(|x| StreamId(x)));
+}
+
+#[test]
+fn ordergroup_6() {
+    let source: [u64; 6] = [0, 1, 2, 3, 4, 5];
+    let result: [u64; 6] = [5, 0, 1, 2, 3, 4];
+    fairness_test(&source.map(|x| StreamId(x)), 5, usize::MAX, &result.map(|x| StreamId(x)));
```
martinthomson, line 277: Your first and last arguments are all the same.  Try this instead:

```rust
fn fairness_test<S, SI>(source: S, ...) where S: IntoIterator<Item = &SI>, SI: Into<StreamId> { ... }

impl From<u64> for StreamId { ... } // you'll need to do this alongside the StreamId definition.
```

neqo-transport/src/send_stream.rs
```diff
@@ -1104,6 +1104,73 @@ impl ::std::fmt::Display for SendStream {
...
+    fn next(&mut self) -> Option<Self::Item> {
+	// Stop when we would return the started_at element on the next
+	// call.  Note that this must take into account wrapping.
+	if self.started_at == Some(self.group.next) || self.group.vec.is_empty() {
+	    return None;
+	}
+	match self.started_at {
+	    None => self.started_at = Some(self.group.next),
+	    Some(_) => {},
+	}
```
martinthomson, line 1163:
```rust
self.started_at = self.started_at.or(self.group.next);
```

neqo-transport/src/send_stream.rs
```diff
@@ -1104,6 +1104,73 @@ impl ::std::fmt::Display for SendStream {
...
+        let mut next = self.group.next + 1;
+	if next >= self.group.vec.len() {
+	    next = 0;
+	}
```
martinthomson, line 1167:
```rust
let next = self.group.next();

fn next(&mut self) -> usize {
   let next = self.next;
   self.next = (self.next + 1) % self.vec.len();
   next
}
```

or

```rust
Some(self.group.next())

fn next(&mut self) -> StreamId {
   let next = self.next;
   self.next = (self.next + 1) % self.vec.len();
   self.vec[next]
}
```

neqo-transport/src/send_stream.rs
```diff
@@ -1163,19 +1230,19 @@ impl SendStreams {
...
             self.get_mut(stream_id).unwrap().set_sendorder(sendorder);
 	    if let Some(order) = sendorder {
-		vec = self.sendordered.entry(order).or_default();
+		group = self.sendordered.entry(order).or_default();
 	    } else {
-		vec = &mut self.regular;
+		group = &mut self.regular;
 	    }
-	    Self::insert_streamid(&mut vec, &stream_id);
+	    Self::insert_streamid(&mut group.stream_ids(), &stream_id);
```
martinthomson, line 1245: Please move this function to `OrderGroup` now.

neqo-transport/src/send_stream.rs
```diff
@@ -1292,10 +1344,19 @@ impl SendStreams {
 	// SendOrdered streams.   Round-robining the data at the same priority
 	// isn't required (currently) by the spec, but would be good to do in the future.
 	//
-	// iterator is in-line to avoid complaints from the compiler
-	let stream_ids = self.regular.iter().chain(self.sendordered.values().rev().flatten());
+	// Fairness is implemented by a round-robining within a single
```
martinthomson, line 1347: round-robining or "statefully iterating"

neqo-transport/src/stream_id.rs
```diff
@@ -17,7 +17,7 @@ pub enum StreamType {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy, Ord, PartialOrd, Hash)]
-pub struct StreamId(u64);
+pub struct StreamId(pub u64);
```
martinthomson, line 20: Implement `From<u64>` as above, rather than making .0 pub.

martinthomson (commented):
So this is fine apart from one thing: this is not the behaviour we would want for HTTP at all.

There, we definitely do not want streams to be send in what amounts to in parallel.  At least not over the short term.  There is a real risk that we'd make 10 requests and find that only half of each was sent, which would tank performance.

This is going to need a strategy for turning it on, perhaps on a per-stream basis, for this to avoid harming HTTP performance.

jesup:
Note, I haven't resolved the earlier comments; I'll do that soon

neqo-transport/src/connection/tests/stream.rs
jesup: From<u64> is already in StreamId
The other bit I'm not sure I exactly know how the make the compiler happy with; it has issues with lifetimes with 
```
fn fairness_test<S, SI>(source: S, number_iterates: usize, truncate_to: usize, result: S) where for<'a> S: IntoIterator<Item = &'a SI>, SI: Into<StreamId> {
```
and any variations I've tried.
What I have works, though it may not be pretty :-)

jesup: Happy to change it if you can give a further hint, but it's probably ok as-is

neqo-transport/src/send_stream.rs
jesup:
```
    fn update_next(&mut self) -> usize {
	let next = self.next;
	self.next = (self.next + 1) % self.vec.len();
	next
    }
...
	let orig = self.group.update_next();
```

neqo-transport/src/stream_id.rs
jesup: impl From<u64> is implemented for StreamId:
```
impl From<u64> for StreamId {
    fn from(val: u64) -> Self {
        Self::new(val)
    }
}
```
but without making that pub it won't compile (though note I haven't gotten the generic to work, so I'm still coercing into StreamId directly)

neqo-transport/src/send_stream.rs
```diff
@@ -1320,16 +1642,16 @@ mod tests {
         // Fill the buffer
         assert_eq!(txb.send(&[1; SEND_BUFFER_SIZE * 2]), SEND_BUFFER_SIZE);
         assert!(matches!(txb.next_bytes(),
-			 Some((0, x)) if x.len()==SEND_BUFFER_SIZE
-			 && x.iter().all(|ch| *ch == 1)));
+                         Some((0, x)) if x.len()==SEND_BUFFER_SIZE
```
KershawChang, line 1645: nit: seems like unnecessary changes?

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,283 @@ impl ::std::fmt::Display for SendStream {
...
+            Err(pos) => self.vec.insert(pos, *stream_id),
+        }
+    }
+
+    pub fn remove_streamid(&mut self, stream_id: &StreamId) -> bool {
+        match self.vec.binary_search(stream_id) {
+            Ok(pos) => { self.vec.remove(pos); },
+            Err(_) => panic!("Missing stream_id {}", stream_id), // element already in vector @ `pos`
+        }
+        true
```
KershawChang, line 1178: This return value is not used, so I think we don't need to return `bool`.

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,283 @@ impl ::std::fmt::Display for SendStream {
...
+
+    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) -> Res<()> {
+        if let Some(stream) = self.map.get_mut(&stream_id) {
+            let old_fair = stream.fair;
+            stream.set_fairness(make_fair);
+            if !old_fair && make_fair {
+                // move to the regular OrderGroup.  We must be called before SendOrder can be set
+                match stream.sendorder {
+                    None => {},
+                    Some(_) => panic!("Stream had sendOrder set before fairness!"),
```
KershawChang, line 1288: I am not a fan of strict ordering of calls like this.
Maybe add another parameter `fair` in `set_sendorder` and remove this one.
What do you think?

jesup: Modified set_sendorder to call set_fairness if it's not already fair, and removed this  match/panic

neqo-transport/src/send_stream.rs
jesup: replaces tabs with spaces (I did an untabify on the files I changed to avoid introducing tabs, and it already had some).

neqo-http3/src/connection.rs
```diff
@@ -995,6 +998,30 @@ impl Http3Connection {
...
+    /// streams of the same sendOrder group (or the unordered group).  Unfair streams
+    /// will give bandwidth preferentially to the lowest streamId with data to send.
+    /// # Errors
+    /// Returns InvalidStreamId if the stream id doesn't exist
+    pub fn stream_set_fairness(
+        &mut self,
+        conn: &mut Connection,
+        stream_id: StreamId,
+        fairness: bool) -> Res<()> {
+        Ok(conn.stream_fairness(stream_id, fairness)?)
```
martinthomson, line 1022: As conn.stream_fairness() returns Res, you can do this:

```suggestion
        conn.stream_fairness(stream_id, fairness)
```

There are a lot of other cases where you have the same redundancy.

neqo-http3/src/connection.rs
```diff
@@ -1238,6 +1265,8 @@ impl Http3Connection {
         let stream_id = conn
             .stream_create(stream_type)
             .map_err(|e| Error::map_stream_create_errors(&e))?;
+        // set outgoing WebTransport streams to be fair (share bandwidth)
+        conn.stream_fairness(stream_id, true).ok();
```
martinthomson, line 1269: please don't ignore the return value, or at least don't do it this way (use .unwrap() explicitly)

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -185,6 +185,14 @@ impl SendStream for WebTransportSendStream {
         }
     }
 
+    fn set_sendorder(&mut self, conn: &mut Connection, sendorder: Option<i64>) {
+	conn.stream_sendorder(self.stream_id, sendorder).ok();
```
martinthomson, line 189: again, you are ignoring the return value here in a very dodgy fashion; .unwrap() if you mean it, but I might prefer that you return `Res<()>`.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +114,178 @@ fn transfer() {
...
+fn sendorder_3() {
+    sendorder_test(&[Some(1), Some(2), None, Some(3)]);
+}
+#[test]
+fn sendorder_4() {
+    sendorder_test(&[Some(1), Some(2), Some(1), None, Some(3), Some(1), Some(3), None]);
+}
+
+
+#[cfg(test)]
```
martinthomson, line 205:
```suggestion
```

You need to stop doing that.

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+        }
+        OrderGroupIter {
+            started_at: None,
+            group: self,
+        }
+    }
+
+    pub fn stream_ids(&mut self) -> &mut Vec<StreamId> {
+        &mut self.vec
+    }
```
martinthomson, line 1158: Does this really need to hand out a mutable reference to its underlying store?  What isn't possible if you do this?

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+
+    fn update_next(&mut self) -> usize {
+        let next = self.next;
+        self.next = (self.next + 1) % self.vec.len();
+        next
+    }
+
+    pub fn insert_streamid(&mut self, stream_id: &StreamId) {
```
martinthomson, line 1166:
```suggestion
    pub fn insert(&mut self, stream_id: StreamId) {
```

Two things:

1. the longer name is not necessary
2. StreamId is Copy

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+        match self.vec.binary_search(stream_id) {
+            Ok(_) => panic!("Duplicate stream_id {}", stream_id), // element already in vector @ `pos`
+            Err(pos) => self.vec.insert(pos, *stream_id),
+        }
+    }
+
+    pub fn remove_streamid(&mut self, stream_id: &StreamId) {
```
martinthomson, line 1173: As above.

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
     pub fn insert(&mut self, id: StreamId, stream: SendStream) {
-        self.0.insert(id, stream);
+        self.map.insert(id, stream);
+    }
+
+    pub fn set_sendorder(&mut self, stream_id: StreamId, sendorder: Option<SendOrder>) -> Res<()> {
+	// check using non-mutable first so we can call set_fairness
+        if let Some(stream) = self.map.get(&stream_id) {
+	    // All sendordered streams are by definition fair since they're all WebTransport streams
+	    if !stream.fair {
```
martinthomson, line 1252: Don't check, just set.

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+            // don't grab stream here; causes borrow errors
+            let old_sendorder = stream.sendorder();
+            if old_sendorder != sendorder {
+                // we have to remove it from the list it was in, and reinsert it with the new
+                // sendorder key
+                let mut group = if let Some(old) = old_sendorder {
+                    self.sendordered.get_mut(&old).unwrap()
+                } else {
+                    &mut self.regular
+                };
```
martinthomson, line 1266: This might benefit from a function: `fn group_mut(&mut self, so: Option<SendOrder>) -> &mut Group { ... }`

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+            }
+            Ok(())
+        } else {
+            Err(Error::InvalidStreamId)
+        }
+    }
+
+    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) -> Res<()> {
+        if let Some(stream) = self.map.get_mut(&stream_id) {
+            let old_fair = stream.fair;
```
martinthomson, line 1288:
```suggestion
            let was_fair = stream.fair;
```

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+                );
+            }
+            Ok(())
+        } else {
+            Err(Error::InvalidStreamId)
+        }
+    }
+
+    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) -> Res<()> {
+        if let Some(stream) = self.map.get_mut(&stream_id) {
```
martinthomson, line 1287: An early return is your friend.

```rust
if let Some(stream) = self.map.get_mut(&stream_id) else { return Err(...); };

File: neqo-transport/src/send_stream.rs [hunk 21]
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+                    self.sendordered.values().collect::<Vec::<_>>()
+                );
+            }
+            Ok(())
+        } else {
+            Err(Error::InvalidStreamId)
+        }
+    }
+
+    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) -> Res<()> {
```
martinthomson, line 1286: Is this `set_fairness` or `set_fair` ?

neqo-transport/src/send_stream.rs
```diff
@@ -1063,61 +1115,290 @@ impl ::std::fmt::Display for SendStream {
...
+        self.map.clear();
+        self.sendordered.clear();
+        self.regular.stream_ids().clear();
+    }
+
+    pub fn remove_terminal(&mut self) {
+        Self::remove_terminal_internal(&mut self.map, &mut self.regular, &mut self.sendordered);
+    }
+
+    fn remove_terminal_internal(map: &mut IndexMap<StreamId, SendStream>, regular: &mut OrderGroup, sendordered: &mut BTreeMap<SendOrder, OrderGroup>) {
```
martinthomson, line 1365: You can inline this function and save the complication of the function invocation.  You will still need to grab the mutable references outside of the retain loop.

neqo-transport/src/send_stream.rs
```diff
@@ -1128,16 +1409,64 @@ impl SendStreams {
...
+        // more expensive searches for insertion and removal (since the
+        // sorted order would be lost).
+
+        // Iterate the map, but only those without fairness, then iterate
+        // OrderGroups, then iterate each group
+        qdebug!("processing streams...  unfair:");
+        for stream in self.map.values_mut() {
+            if !stream.is_fair() {
+                qdebug!("   {}", stream);
+                if !Self::write_frames_with_stream(priority, builder, tokens, stats, stream) {
```
martinthomson, line 1449: Can this be moved to be a function on the SendStream itself?

neqo-transport/src/stream_id.rs
martinthomson: Yeah, this isn't going to be good enough.  You can use the From implementation in the tests and not need this change.  But I think that making the test function better (along the lines I described) would be better yet.

neqo-http3/src/connection.rs
jesup: Aha, and here's why I used Ok(...?):
```
    --> neqo-http3/src/connection.rs:1022:9
     |
1021 |         fairness: bool) -> Res<()> {
     |                            ------- expected `Result<(), Error>` because of return type
1022 |         conn.stream_fairness(stream_id, fairness)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Error`, found `neqo_transport::Error`
     |
     = note: `neqo_transport::Error` and `Error` have similar names, but are actually distinct types
note: `neqo_transport::Error` is defined in crate `neqo_transport`
    --> /home/jesup/src/mozilla/neqo/neqo-transport/src/lib.rs:64:1
     |
64   | pub enum Error {
     | ^^^^^^^^^^^^^^
note: `Error` is defined in the current crate
    --> neqo-http3/src/lib.rs:191:1
     |
191  | pub enum Error {
     | ^^^^^^^^^^^^^^
help: use `?` to coerce and return an appropriate `Err`, and wrap the resulting value in `Ok` so the expression remains of type `Result`
     |
1022 |         Ok(conn.stream_fairness(stream_id, fairness)?)
     |         +++                                         ++
```

jesup: conn.stream_fairness(stream_id, fairness).map_err(|_| Error::InvalidStreamId) works, though likely there's a better solution

neqo-http3/src/features/extended_connect/webtransport_streams.rs
jesup: Modified to return Res<()> and in the places that implies as well.  I have to do the same map_err(|_| Error::InvalidStreamId)  as above because Error and neqo_transport::Error are different

neqo-transport/src/send_stream.rs
jesup: Since I'm allowing for changing a stream from fair to unfair, set_fairness

neqo-transport/src/send_stream.rs
jesup: We need it for 
                    self.regular.stream_ids().push(stream_id);
and
        self.regular.stream_ids().clear();

jesup: I could probably implement push() and clear() for OrderGroup().   that works.

neqo-transport/src/send_stream.rs
jesup: Changed the names for insert and remove.
Changed &StreamId to StreamId; this required changing in ~10 places that call this.

neqo-transport/src/send_stream.rs
jesup: Need to add a {} before 'else', but sure

neqo-transport/src/send_stream.rs
```diff
@@ -1401,36 +1730,36 @@ mod tests {
...
+                         && x.iter().all(|ch| *ch == 2)));
 
         txb.mark_as_sent(ten_bytes_past_end, 20);
 
         // Get bit after earlier marked range A
         assert!(matches!(txb.next_bytes(),
-			 Some((start, x)) if x.len() == 60
-			 && start == range_a_end
-			 && x.iter().all(|ch| *ch == 2)));
+                         Some((start, x)) if x.len() == 60
```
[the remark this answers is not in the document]
jesup: I untabify'd the file (there were tabs in those tests)

neqo-transport/src/connection/tests/stream.rs
jesup: Got it

neqo-transport/src/send_stream.rs
martinthomson: That's what I was working towards, yes.

neqo-transport/src/send_stream.rs
martinthomson: You shouldn't need the type declaration, but yes, this is better.

neqo-transport/src/send_stream.rs
martinthomson: set_fair(false) is quite clearly set_unfair

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
martinthomson: rustfmt adds tabs.  You might not like that, but that is how rust code is styled.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -111,6 +114,179 @@ fn transfer() {
...
+}
+
+
+// Tests stream sendorder priorization
+// Converts Vecs of u64's into StreamIds
+fn fairness_test<S, R, SI>(source: S, number_iterates: usize, truncate_to: usize, result_array: R)
+where
+    S: IntoIterator<Item = SI>,
+    R: IntoIterator<Item = SI> + std::fmt::Debug,
+    SI: Into<StreamId>, Vec<u64>: PartialEq<R>
```
martinthomson, line 210: As a general rule, you shouldn't constrain the caller by requiring the same type for items in both iterators.

That means something like this (or separate SI and RI type parameters):

```rust
where
S: IntoIterator,
S::Item: Into<StreamId>,
R: IntoIterator,
R::Item: Into<StreamId>,
Vec<u64>: PartialEq<R>,
```

However, it looks like the last constraint is only one you really need.  That is, you only use PartialEq.  R doesn't need to implement IntoIterator because you don't use that.

If I were designing an API, I probably wouldn't restrict the caller that way.  You would instead use `Iterator::eq` to compare items from the expected outcome with what really happened.  But that probably isn't worth it in this case, especially since that would change what is printed when the assertion fails, and not in a good way.

martinthomson (approved):
I'm OK with this.  As I'm on a plane soon, @KershawChang feel free to proceed once @jesup has had a chance to look at the last few comments.

martinthomson:
Oh, and of course this needs rebasing.

KershawChang:
> I'm OK with this. As I'm on a plane soon, @KershawChang feel free to proceed once @jesup has had a chance to look at the last few comments.

Thanks, I'll take a final look after this is rebased.

PR #1444, opened 2023-06-16
Implement RecvStreamStats

KershawChang:
This is for `WebTransportReceiveStreamStats`.

Note that there will be another commit for the implementation in `webtransport_stream`.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -122,6 +122,7 @@ impl RecvStreams {
 pub struct RxStreamOrderer {
     data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
     retired: u64,                        // Number of bytes the application has read
+    received: u64,                       // The number of bytes has stored in `data_ranges`
 }
 
 impl RxStreamOrderer {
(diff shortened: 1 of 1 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -510,6 +560,44 @@ impl RecvStream {
...
+                final_received,
+                final_read,
+            }
+            | RecvStreamState::ResetRecvd {
+                final_received,
+                final_read,
+            } => {
+                let received = *final_received;
+                let read = *final_read;
+                RecvStreamStats::new(received, read)
```
martinthomson, line 596: This block is identical to the one above.  Can you combine them with `|`?

neqo-transport/src/recv_stream.rs
```diff
@@ -765,19 +914,30 @@ impl RecvStream {
...
+                let read = *final_read;
+                self.set_state(RecvStreamState::ResetRecvd {
+                    final_received: received,
+                    final_read: read,
+                });
             } else {
                 let fc_copy = mem::take(fc);
                 let session_fc_copy = mem::take(session_fc);
+                let received = *final_received;
+                let read = *final_read;
```
martinthomson, line 935: You can move these assignments out of the if statement.

martinthomson (approved):

PR #1446, opened 2023-06-23
WebTransportReceiveStreamStats Impl

KershawChang:


neqo-http3/src/lib.rs
```diff
@@ -470,6 +473,11 @@ trait RecvStream: Stream {
     fn webtransport(&self) -> Option<Rc<RefCell<WebTransportSession>>> {
         None
     }
+
+    /// This function is only implemented by `WebTransportRecvStream`.
+    fn stats(&mut self, _conn: &mut Connection) -> Res<RecvStreamStats> {
+        Err(Error::Unavailable)
```
martinthomson, line 479: It seems to me like you could just make this:

```rust
Ok(conn.recv_stream_stats(self.stream_id)?)
```

It's not like the stats provided would be invalid for non-webtransport streams, only that you wouldn't be reporting adjusted values.

neqo-transport/src/lib.rs
```diff
@@ -26,7 +26,7 @@ mod path;
 mod qlog;
 mod quic_datagrams;
 mod recovery;
-mod recv_stream;
+pub mod recv_stream;
```
martinthomson, line 29: I would prefer that you re-export the stats part.  That is,

```rust
mod recv_stream;
// ...
pub use recv_stream::RecvStreamStats;
```

There is stuff in the module that we don't need to make public, such as all the RecvStream stuff.

I know that we did that for send_stream, but that was probably also a mistake.

martinthomson (approved):
Two minor comments only.  Sorry for letting this drop.

neqo-http3/src/lib.rs
KershawChang: Unfortunately, `self` doesn't have `stream_id` here.
I'll keep returning an error here.

martinthomson: Cool.  No worries.

PR #1447, opened 2023-06-28
Move datagrams below regular streams

martinthomson:
This will starve datagrams, which comes with its own risks.

Eventually, we'll need a better strategy.

@jesup is this a better starting for #1440?

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -36,6 +36,7 @@ use crate::{
         ConnectionId, ConnectionIdEntry, ConnectionIdGenerator, ConnectionIdManager,
         ConnectionIdRef, ConnectionIdStore, LOCAL_ACTIVE_CID_LIMIT,
     },
+    frame,
 };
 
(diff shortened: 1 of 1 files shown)


KershawChang (approved):

jesup:
LGTM.   Both ways work.

KershawChang:
Hi @martinthomson , I am a bit confused about this patch.
Should we land this one or #1440? @jesup thinks we should land this one, but this is still a draft.
What do you think?
Thanks.

KershawChang:
@martinthomson , could you resolve the conflicts of this patch?
I'd like to land this one, since this also includes some clean up. Thanks.

martinthomson:
The blocker here is tests.  Neither patch includes tests and I really don't like having functionality that isn't tested.

KershawChang:
> The blocker here is tests. Neither patch includes tests and I really don't like having functionality that isn't tested.

Ah, I see. I'll write a test for this.

martinthomson:
I think that we got the tests in #1440, so I'm closing this one.

PR #1450, opened 2023-07-07
add test

KershawChang:
This is the test for #1447


neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +222,48 @@ fn datagram_acked() {
     ));
 }
 
+#[test]
+fn datagram_after_stream_data() {
+    let (mut client, mut server) = connect_datagram();
+
+    // Wriate a datagram first.
+    let _dgram_sent = client.stats().frame_tx.datagram;
```
martinthomson, line 230: Do you plan to use this variable?  You could check that it increases as an alternative way to verify things (then you might not need to check receipt at the server end at all (as noted below).

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +222,48 @@ fn datagram_acked() {
     ));
 }
 
+#[test]
+fn datagram_after_stream_data() {
+    let (mut client, mut server) = connect_datagram();
+
+    // Wriate a datagram first.
```
martinthomson, line 229:
```suggestion
    // Write a datagram first.
```

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +222,48 @@ fn datagram_acked() {
...
+    let _dgram_sent = client.stats().frame_tx.datagram;
+    assert_eq!(client.send_datagram(DATA_SMALLER_THAN_MTU, Some(1)), Ok(()));
+
+    // Create a stream and send some data.
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+    client.stream_send(stream_id, &[6; 100]).unwrap();
```
martinthomson, line 235: Ok, so is the datagram too big to fit in a packet that has this stream data as well?  I'd send a little more to ensure that is the case.

If it is, then that is good.  Then you want to:

1. have the client create two packets.  the first should be smaller than the datagram size, which you can check
2. have the server process the first.  it should receive stream data, but not a datagram event
3. have the server process the second.  it should receive the datagram event then

We probably shouldn't rely on the order of events here.  That's not something we'd want to promise in the API unless we really had to (and we don't).

Then add another test that sends low priority stream data first and observe the opposite inversion.

martinthomson (changes requested):
I think we need two tests and some small changes, as noted below.

martinthomson:
Note that we're also seeing this new test fail.

neqo-transport/src/connection/tests/datagram.rs
KershawChang: I'll use it for testing. Thanks.

neqo-transport/src/connection/tests/datagram.rs
KershawChang: Will do

KershawChang:
> Note that we're also seeing this new test fail.

I'll push a new commit that includes every thing to make the test pass.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +225,76 @@ fn datagram_acked() {
...
+        .filter_map(|evt| match evt {
+            ConnectionEvent::RecvStreamReadable { .. } | ConnectionEvent::Datagram { .. } => {
+                Some(evt)
+            }
+            _ => None,
+        })
+        .collect();
+    // We should only get one event - either RecvStreamReadable or Datagram.
+    assert_eq!(events.len(), 1);
+    assert_eq!(mem::discriminant(event), mem::discriminant(&events[0]));
```
martinthomson, line 246: If you change this function to return the event, you can call it like this:

```rust
    assert!(matches!(send_packet_and_get_server_event(&mut client, &mut server), ConnectionEvent::RecvStreamReadable { stream_id: s } if s == stream_id);
```

The problem with only matching the discriminant is that you aren't testing that the stream ID is correct.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +225,76 @@ fn datagram_acked() {
...
+
+    // Create a stream with normal priority and send some data.
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+    client.stream_send(stream_id, &[6; 1200]).unwrap();
+
+    let desired_event = ConnectionEvent::RecvStreamReadable { stream_id };
+    send_packet_and_check_server_event(&mut client, &mut server, &desired_event);
+    assert_eq!(client.stats().frame_tx.datagram, dgram_sent);
+
+    let desired_event = ConnectionEvent::Datagram([].to_vec());
```
martinthomson, line 265: If you do as I suggest, you can do `matches!(..., ConnectionEvent::Datagram(_))` instead.  Or even...

```rust
   if let ConnectionEvent::Datagram(data) = send_packet_and_get_server_event(&mut client, &mut server) {
      assert_eq!(data, DATA_MTU);
   } else {
      panic!();
   }
```

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -224,6 +225,76 @@ fn datagram_acked() {
...
+            }
+            _ => None,
+        })
+        .collect();
+    // We should only get one event - either RecvStreamReadable or Datagram.
+    assert_eq!(events.len(), 1);
+    assert_eq!(mem::discriminant(event), mem::discriminant(&events[0]));
+}
+
+#[test]
```
martinthomson, line 249: Add a comment for these:

```rust
/// Write a datagram that is big enough to fill a packet, but then see that 
/// normal priority stream data is sent first.
```

martinthomson (approved):
A small suggestion for you regarding the tests, that's all.

neqo-transport/src/connection/tests/datagram.rs
KershawChang: Thanks. Will do.

PR #1459, opened 2023-08-29
General maintenance

martinthomson:
Several things:

1. https://github.com/rust-lang/log/commit/9d052b17d3ddf9b5894740c6d82899c531c789ee changed the internal-only APIs used by the `log` crate.  This was always just a roundabout way of doing `module_path!()`, so we're good.
2. The `time` crate moved to a minimum version of rust beyond the one we support, so pin its version.
3. Clippy errors are being generated as a result of new rust versions.  There are a fair number of these:
   1. Remove some `clippy::missing_panics_doc` overrides.
   2. Some functions were really missing `Panics` docs.
   3. Some pointer casts were changed from `as *mut T` to `cast_mut()` or `as *const T` to `cast()`.
   4. Some calls to `.iter()` were unnecessary.
   5. A new `clippy::single_range_in_vec_init` lint was added, but with a false positive.
 6. Clippy complained about `Header::new` not taking a reference, but we explicitly want a value, so use `Into` rather than `ToString`.

### DIFF
--- a/neqo-common/Cargo.toml
+++ b/neqo-common/Cargo.toml
@@ -12,7 +12,7 @@ log = {version = "0.4.0", default-features = false}
 env_logger = {version = "0.10", default-features = false}
 lazy_static = "1.3.0"
 qlog = "0.4.0"
-time = {version = "0.3", features = ["formatting"]}
+time = {version = "0.3.23", features = ["formatting"]}
 
 [features]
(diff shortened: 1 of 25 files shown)


martinthomson:
Comment updated to reflect the breadth of the changes involved.

KershawChang (approved):

Issue #1463, opened 2023-09-27
When CWND is low, the pacer signals its user to pause more often


KershawChang:
I've been debugging the upload performance issue, and it seems that the bottleneck is related to the pacer.
According to this equation ["(now - self.t) * PACER_SPEEDUP * cwnd / rtt"](https://searchfox.org/mozilla-central/rev/118283c2e3eb393e9b1ca00a0c328e32eeab1c40/third_party/rust/neqo-transport/src/pace.rs#83), when the `cwnd` is small, the pacer increase its capability with a smaller value.
As the following log shows, there is a `102.471µs` delay after sending a few packets. Unfortunately, the [timer resolution in necko is in milliseconds](https://searchfox.org/mozilla-central/rev/118283c2e3eb393e9b1ca00a0c328e32eeab1c40/netwerk/socket/neqo_glue/src/lib.rs#329), this delay becomes 1ms. 
```
2023-09-27 20:54:04.019175 UTC - [Parent 70649: Socket Thread]: I/neqo_transport::* [neqo_transport::pace] [Pacer id=7 728/1357] next 121081/31.174185ms wait 80.972µs = Instant { tv_sec: 3158011, tv_nsec: 382555722 }
2023-09-27 20:54:04.019178 UTC - [Parent 70649: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client a228a8261e9e8b4596] output_path send_profile SendProfile { limit: 1357, pto: None, probe: (), paced: false }
2023-09-27 20:54:04.019189 UTC - [Parent 70649: Socket Thread]: I/neqo_transport::* [neqo_transport::pace] [Pacer id=7 728/1357] spend 1357 cwnd=[121081], rtt=[31.174185ms] t=1.295291ms incr=10061
2023-09-27 20:54:04.019198 UTC - [Parent 70649: Socket Thread]: E/nsHttp Http3Session::ProcessOutput sending packet with 1357 bytes to 142.250.181.238 port=443 [this=1403e1f00].
2023-09-27 20:54:04.019214 UTC - [Parent 70649: Socket Thread]: I/neqo_transport::* [neqo_transport::pace] [Pacer id=7 2714/1357] next 121081/31.174185ms no wait = Instant { tv_sec: 3158011, tv_nsec: 383770041 }
[...]

PR #1467, opened 2023-10-13
Enable pacer tests and add a new one to demostrate the performance issue

KershawChang:
I am not sure why, but the tests in pacer.rs were disabled and broken. This patch enables those tests and add a new one to demonstrate the issue in issue #1463.

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -108,9 +108,10 @@ impl Debug for Pacer {
     }
 }
 
-#[cfg(tests)]
+#[cfg(test)]
 mod tests {
     use super::Pacer;
+    use std::time::Duration;
     use test_fixture::now;
 
     const RTT: Duration = Duration::from_millis(1000);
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
Aside from the fact that the tests were disabled (oops!), I'm not sure that I see an issue here.

From what I can see, the test below uses a very small pacing interval (20ms RTT, 100 packets per RT, 2x loading: 100μs interval).

The test spends once three times at t=0, t=1ms, t=1.03ms and t=1.06ms.  At t=0, the reserve is full, so it is OK.  At t=1ms (which is more than 100μs), the reserve has completely filled again.  At t=1.03ms, the reserve is at 130% of an MTU out of 200%, so it's good.  At t=1.06ms, there is just 60% of an MTU saved, so you are asked to wait for 40μs more (plus 14ns of rounding error).

The point of the pacer is to smooth sending out, and this would be achieving that if it weren't for the 1ms gap between each call.  Even if we manage to hit a 1ms interval consistently, which seems unlikely, a short RTT with high packet rates will not get much performance.

As I've said before, there is a need to address this in the calling code.  If that code is asked to wait for less than some threshold (let's say 2ms, to allow some slop for scheduling delays), then it should not wait.  Instead, it should send a packet and pretend to have waited. It can then get the next packet immediately, looping until the amount of time it has waited exceeds that 1ms or 2ms threshold.

The only trick there is that you can't jump ahead 2ms and have a callback in 1ms if that means that you might reverse time.

KershawChang:
> 
> As I've said before, there is a need to address this in the calling code. If that code is asked to wait for less than some threshold (let's say 2ms, to allow some slop for scheduling delays), then it should not wait. Instead, it should send a packet and pretend to have waited. It can then get the next packet immediately, looping until the amount of time it has waited exceeds that 1ms or 2ms threshold.
> 

If I understand this correctly, the calling code here is [Http3Session::ProcessOutput](https://searchfox.org/mozilla-central/rev/0ba4632ee85679a1ccaf652df79c971fa7e9b9f7/netwerk/protocol/http/Http3Session.cpp#892). In order to implement the suggested change, we need to:
1. Have neqo provide the reason of delay. Currently, it's not clear if a delay returned by neqo is due to the pacer or some other reason.
2. Pretend that necko already waited and call `ProcessOutput` again with a `future` time (now + 40μs).

IMO, this is not easy to implement. We'd need to keep track of the "pretend" wait times. It's important to ensure that the time we use to invoke [process_input](https://searchfox.org/mozilla-central/rev/0ba4632ee85679a1ccaf652df79c971fa7e9b9f7/netwerk/socket/neqo_glue/src/lib.rs#298) is always greater than the simulated future time used for the previous `ProcessOutput` call.

Can we try to address this in the pacer? Maybe we can make the `next_delay` returned by the pacer is always greater than 1ms.
Or we can just increase the `PACING_BURST_SIZE` to a larger value (maybe 10)? Note that the pacer implementation in quiche is very similiar with ours and the capacity they used is the size of [initial cwnd](https://github.com/cloudflare/quiche/blob/83d9168ab6f76302ae846cb068cc8991f2b06479/quiche/src/recovery/mod.rs#L282) (10*mtu).

KershawChang:
Close this PR, since there is no proof shows that pacing is an issue.

PR #1468, opened 2023-10-17
Only warn on warnings and don't deny compiling during development

mb:
Denying in the source makes developing neqo a bit cumbersome. It also makes it harder to compile neqo with future versions of Rust. We do have a CI pipeline for clippy denying all warnings. That should be enough to keep all the warnings out.

### DIFF
--- a/neqo-client/Cargo.toml
+++ b/neqo-client/Cargo.toml
@@ -19,5 +19,4 @@ url = "2.0"
 qlog = "0.9.0"
 
 [features]
-default = ["deny-warnings"]
 deny-warnings = []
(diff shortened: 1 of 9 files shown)


mb:
This is also how other projects like [`cargo clippy`](https://github.com/rust-lang/rust-clippy/blob/2640d5cc85ff32de901d36cebbd48ccf0b398453/Cargo.toml#L55) handle it.

martinthomson:
WFM.  Will the addition of `-D warnings` in CI still catch warnings?

mb:
Tested out now: #1469. Both rustc and clippy warnings still make the CI fail.

PR #1470, opened 2023-10-18
Introduce Output::PacedCallback to indicate if this is caused by the pacer

KershawChang:
This is for the suggested change in https://github.com/mozilla/neqo/pull/1467#pullrequestreview-1681017489.

Please also take a look at the change at calling side [here](https://phabricator.services.mozilla.com/D191325).
Thanks!

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -404,7 +404,7 @@ fn process_loop(
                         break;
                     }
                 }
-                Output::Callback(duration) => {
+                Output::Callback(duration) | Output::PacedCallback(duration) => {
                     socket.set_read_timeout(Some(duration)).unwrap();
                     break;
                 }
@@ -1095,7 +1095,7 @@ mod old {
                             break;
                         }
                     }
-                    Output::Callback(duration) => {
+                    Output::Callback(duration) | Output::PacedCallback(duration) => {
                         socket.set_read_timeout(Some(duration)).unwrap();
                         break;
                     }
--- a/neqo-interop/src/main.rs
+++ b/neqo-interop/src/main.rs
@@ -118,7 +118,7 @@ fn process_loop(
                     let dgram = handler.rewrite_out(&dgram).unwrap_or(dgram);
                     emit_datagram(&nctx.socket, dgram);
                 }
-                Output::Callback(duration) => {
+                Output::Callback(duration) | Output::PacedCallback(duration) => {
                     let delay = min(timer.check()?, duration);
                     nctx.socket.set_read_timeout(Some(delay)).unwrap();
                     break;
@@ -280,7 +280,7 @@ fn process_loop_h3(
             let output = handler.h3.conn().process_output(Instant::now());
             match output {
                 Output::Datagram(dgram) => emit_datagram(&nctx.socket, dgram),
-                Output::Callback(duration) => {
+                Output::Callback(duration) | Output::PacedCallback(duration) => {
                     let delay = min(timer.check()?, duration);
                     nctx.socket.set_read_timeout(Some(delay)).unwrap();
                     break;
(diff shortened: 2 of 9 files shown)


martinthomson:
I would rather not externalize state in this way.

The problem arises because we have precise timers in Rust, but somewhat more loose ones in reality.  This code has a very precise notion of time (nanosecond resolution), but we can't deliver that sort of precision in practice.  We're talking about how we manage that difference.

Right now, we erase any time that is less than 1ms in resolution [here](https://searchfox.org/mozilla-central/rev/c3c92500ef8ffab4fa53d4b2d5c5e2b49025e89b/netwerk/socket/neqo_glue/src/lib.rs#308).  That is partly because we know that the timer resolution on the calling side is incapable of handling that (it might not be able to handle less than 16ms in some cases, but let's at least pretend that 1ms is OK).

What I'm suggesting is that we stop [rounding the time offset](https://searchfox.org/mozilla-central/rev/c3c92500ef8ffab4fa53d4b2d5c5e2b49025e89b/netwerk/socket/neqo_glue/src/lib.rs#323):

```rust
        Output::Callback(to) => {
            *timeout = min(to, Duration::from_nanos(u64::MAX - 1)).as_nanos() as u64;
            false
        }
```

In practice, a 64-bit number will hold plenty of nanoseconds to cover most timeouts we care about, so you might get something like this then.

```c++
  uint64_t accumulatedTime = 0;
  const uint64_t maxAccumulatedTime = 2_000_000;  // in nanoseconds
  while (true) {
    // ... 
    if (!mHttp3Connection->ProcessOutput(&remoteAddrStr, &port, packetToSend,
                                         &timeout)) {
      accumulatedTime += timeout; // BUG: check for overflow
      if (accumulatedTime > maxAccumulatedTime) {
        SetupTimer(accumulatedTime / 1_000_000);  // neqo uses nanoseconds, like rust does; this uses milliseconds
        break;
      }
      // Otherwise, we're pretending that time just moved forward a little more
      // so that we can get another packet.
      continue;
    }
    // ... send the packet
  }
```

This isn't done yet, because you can't keep calling in to the stack with `Instant::now()`; you have to track the time better.  I don't know whether it is possible to pass `Instant` and `Duration` across the FFI interface.  Dragana decided not to.  But that what is really what you want here.  You want to call `Instant::now()` at the top of the sending loop and then add the accumulated time to that when invoking `mHttp3Connection->ProcessOutput()`.  For safety, you would track the largest number you had used and just exit this function if `Instant::now()` is less than the value you last used.

Leading to:

```
  Instant now = Instant::now(); // rust ???
  if (mLastTime > now) { 
    // The timer fired too early, so reschedule it.
    // The 1ms of extra delay is not ideal, but this is a fail
    SetupTimer((mLastTime - now).as_millis() + 1);
    return NS_OK;
  }
  Duration accumulatedTime = 0; // rust ???
  const Duration maxAccumulatedTime = Duration::from_millis(2);  // rust ???
  while (true) {
    // ... 
    Duration timeout = 0;  // rust ???
    mLastTime = now + accumulatedTime;
    if (!mHttp3Connection->ProcessOutput(mLastTime, &remoteAddrStr, &port,
                                  packetToSend, &timeout)) {
      accumulatedTime += timeout; // BUG: check for overflow
      if (accumulatedTime > maxAccumulatedTime) {
        SetupTimer(accumulatedTime.as_millis());
        break;
      }
      // Otherwise, we're pretending that time just moved forward a little more
      // so that we can get another packet.
      continue;
    }
    // ... send the packet
  }
```

That's more rust code in the C++ side than I'd like, but you get the idea.

The problem with your change here is that it doesn't really change the interface to the stack, except to complicate it.  We don't need to tell the caller why they are being told to wait.

KershawChang:
Thanks for the pseudo code. It's really clear!
I'll close this PR and we can move our discussion to [this patch](https://phabricator.services.mozilla.com/D191325).

PR #1478, opened 2023-10-24
Take timer accuracy into consideration when spending pacer credit

mb:
My suggestion from https://github.com/mozilla/neqo/pull/1470#issuecomment-1771201741 in code. Still draft, because it only shows the concept, but I can finalize if this solution is way we want to go. I don't like in https://phabricator.services.mozilla.com/D191325 that we pass neqo wrong timestamps, because it also has some less visible side effects (e.g. RTT calculation), but if we go into the direction of having a pacer in kernel that might(?) be the way to go.

The gist of this patch.
> when `spend` is called within timer accuracy, assume that we just want to continuously send.
> Calculate how much time we would need to fill out the capacity to its maximum and
> advance the current time by that. This time is used as next_send_time in `next`
> function

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -26,6 +26,8 @@ const PACER_SPEEDUP: usize = 2;
 
 /// A pacer that uses a leaky bucket.
 pub struct Pacer {
+    /// The timer accuracy
+    accuracy: Duration,
     /// The last update time.
     t: Instant,
     /// The maximum capacity, or burst size, in bytes.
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
This might not be necessary with the changes that @KershawChang is working on outside of this stack.  Perhaps the two of you should talk.

There is still a gap there in that timer granularity might not reach the 1ms we configured.  If we can't deliver timer reliability on that scale and delays run long, we'll have longer pauses.

KershawChang:
> This might not be necessary with the changes that @KershawChang is working on outside of this stack. Perhaps the two of you should talk.
> 
> There is still a gap there in that timer granularity might not reach the 1ms we configured. If we can't deliver timer reliability on that scale and delays run long, we'll have longer pauses.

We've decided to finish this [patch](https://phabricator.services.mozilla.com/D191325) first, since this patch is not complete.

martinthomson:
OK, great.

In thinking about this approach more, there is potentially some value in having some tracking in this code so that it can alter how it behaves when it asks for a delay of X and ends up with a delay of X+D.  If that happens consistently, then maybe we could adapt in this code, but it's a little risky to do that.

mb:
Yeah, some other approach is probably better in neqo. Having it detect the "too long delay after only small pacer delay requested" could be viable.

Closing in favor of https://phabricator.services.mozilla.com/D191325. If we implement another approach that should be done in another PR.

PR #1479, opened 2023-10-25
Count as non-applimited when at least one packet was sent during non-applimited time period

### LINKED ISSUE #1475: Receiving multiple ACK ranges makes later ACK ranges not count towards cwnd due to app_limited
Last observation from graphs in #1472. I created multiple issues because I think they can be fixed separately.

We receive ack range covering ~10 packets at a time and call `on_packets_acked` on each ack range. The first one is not `app_limited`, but because we don't send out enough packets right away (none to be specific), all consecutive ack ranges don't count towards the cwnd. In my testing during the recovery phase 7153 out of 25839 acks were app_limited (27.68%) due to this.

mb:
Fixes #1475 

Still have to fix test cases. All green dots are ack-ranges that don't count towards cwnd due to app-limit detection.
| before patch | after patch |
| --- | --- |
| before_patch | after_patch |

Notice that with the patch applied during normal operation without packet loss, all ack ranges are not app-limited. The slightly longer drops are due to pacer and disappear with also #1478 applied.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -114,6 +114,14 @@ pub struct ClassicCongestionControl<T> {
     acked_bytes: usize,
     ssthresh: usize,
     recovery_start: Option<PacketNumber>,
+    /// `first_app_limited` indicates the packet number after which the application might be
+    /// underutilizing the congestion window. When underutilizing the congestion window due to not
+    /// sending out enough data, we SHOULD NOT increase the congestion window.[1] Packets sent
+    /// before this point are deemed to fully utilize the congestion window and count towards
+    /// increasing the congestion window.
+    ///
+    /// [1]: https://datatracker.ietf.org/doc/html/rfc9002#section-7.8
+    first_app_limited: PacketNumber,
 
     qlog: NeqoQlog,
 }
@@ -153,19 +161,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
 
     // Multi-packet version of OnPacketAckedCC
     fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], min_rtt: Duration, now: Instant) {
-        // Check whether we are app limited before acked packets are removed
-        // from bytes_in_flight.
-        let is_app_limited = self.app_limited();
-        qtrace!(
-            [self],
-            "limited={}, bytes_in_flight={}, cwnd={}, state={:?} pacing_burst_size={}",
-            is_app_limited,
-            self.bytes_in_flight,
-            self.congestion_window,
-            self.state,
-            MAX_DATAGRAM_SIZE * PACING_BURST_SIZE,
-        );
-
+        let mut is_app_limited = true;
         let mut new_acked = 0;
         for pkt in acked_pkts {
             qinfo!(
@@ -178,6 +174,9 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             );
             if !pkt.cc_outstanding() {
                 continue;
+            }
+            if pkt.pn < self.first_app_limited {
+                is_app_limited = false;
             }
             assert!(self.bytes_in_flight >= pkt.size);
             self.bytes_in_flight -= pkt.size;
@@ -326,6 +325,13 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         if !pkt.cc_in_flight() {
             return;
         }
+        if !self.app_limited() {
+            // Given the current non-app-limited condition, we're fully utilizing the congestion
+            // window. Assume that all in-flight packets up to this one are NOT app-limited.
+            // However, subsequent packets might be app-limited. Set `first_app_limited` to the
+            // next packet number.
+            self.first_app_limited = pkt.pn + 1;
+        }
 
         self.bytes_in_flight += pkt.size;
         qinfo!(
@@ -357,6 +363,7 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
             ssthresh: usize::MAX,
             recovery_start: None,
             qlog: NeqoQlog::disabled(),
+            first_app_limited: 0,
         }
     }
 
@@ -532,6 +539,7 @@ mod tests {
     };
     use crate::{
         cc::{
+            classic_cc::State,
             cubic::{Cubic, CUBIC_BETA_USIZE_DIVIDEND, CUBIC_BETA_USIZE_DIVISOR},
             new_reno::NewReno,
             CongestionControl, CongestionControlAlgorithm, CWND_INITIAL_PKTS, MAX_DATAGRAM_SIZE,
@@ -539,6 +547,7 @@ mod tests {
         packet::{PacketNumber, PacketType},
         tracking::SentPacket,
     };
+    use neqo_common::qinfo;
     use std::{
         convert::TryFrom,
         time::{Duration, Instant},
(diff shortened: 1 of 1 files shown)


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -110,6 +111,7 @@ pub struct ClassicCongestionControl<T> {
     acked_bytes: usize,
     ssthresh: usize,
     recovery_start: Option<Instant>,
+    last_non_app_limited: PacketNumber,
```
martinthomson, line 114: Adding some documentation here about what this is used for and what that means would be good.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -332,6 +338,7 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
             ssthresh: usize::MAX,
             recovery_start: None,
             qlog: NeqoQlog::disabled(),
+            last_non_app_limited: 0,
```
martinthomson, line 341: Better than `Option<PacketNumber>`, I agree.

martinthomson (commented):
If I understand the reasoning here, I think that I agree that this is the right approach.  Even if I didn't, the basic idea that app limiting is affecting our ability to increase CWND in cases where we should be increasing it, means that we need to find ways to narrow its effect.  This is fine.

mb:
I had to adjust the code slightly (store first potentially app-limit packet instead of last non-app-limited) to keep a few tests running, which assumed that the first packet can be app-limited. I think the complexity isn't increased too much and I hope the explaining comments will help.

Adapted the in_applimit test cases to have bursts of packets that don't reach cwnd and check that cwnd doesn't get increased when the acks arrive.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -300,6 +310,13 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         if !pkt.cc_in_flight() {
             return;
         }
+        if !self.app_limited() {
+            // Given the current non-app-limited condition, we're fully utilizing the congestion
+            // window. We can safely consider that all in-flight packets up to the current to be not
+            // app-limiting. However, the subsequent packets might fall under app-limited
+            // condition. Hence, set the `first_app_limited` to the next packet number.
+            self.first_app_limited = pkt.pn + 1;
```
martinthomson, line 318: Maybe you should just track the last non-app-limited packet instead.

martinthomson: Hmm, I shouldn't do code review during meetings at 3am.  That's what you had before.

I don't understand why this would be different than what you had before.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1045,42 +1048,37 @@ mod tests {
...
+        assert_eq!(cc.state, State::CongestionAvoidance);
+        // simulate 10 packet bursts below app_limit
+        for k in 0..10 {
+            // always stay below app_limit during sent.
+            let mut pkts = Vec::new();
+            for i in 0..BELOW_APP_LIMIT_PKTS {
+                let p = SentPacket::new(
+                    PacketType::Short,
+                    u64::try_from(k * BELOW_APP_LIMIT_PKTS + i + 3).unwrap(), // pn
+                    now() + (k as u32 + 2) * RTT,                             // time sent
```
martinthomson: k is usize, so maybe

```suggestion
                    now() + (u32::try_from(k).unwrap() + 2) * RTT,                             // time sent
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1045,42 +1048,37 @@ mod tests {
...
+                BELOW_APP_LIMIT_PKTS * MAX_DATAGRAM_SIZE
             );
-            assert_eq!(cc.cwnd(), CWND_PKTS_CA * MAX_DATAGRAM_SIZE);
-            assert_eq!(cc.acked_bytes, MAX_DATAGRAM_SIZE * 3);
+            for (i, pkt) in pkts.into_iter().enumerate() {
+                cc.on_packets_acked(&[pkt], RTT, now() + (k as u32 + 3) * RTT);
+
+                assert_eq!(
+                    cc.bytes_in_flight(),
+                    (CWND_PKTS_CA - 2 - i - 1) * MAX_DATAGRAM_SIZE
```
martinthomson, line 1079:
```suggestion
                    (BELOW_APP_LIMIT_PKTS - i - 1) * MAX_DATAGRAM_SIZE
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1045,42 +1048,37 @@ mod tests {
...
+                );
+                cwnd_is_halved(&cc);
```
martinthomson, line 1081:
```suggestion
                cwnd_is_halved(&cc);  // CWND doesn't grow because we're app limited
```

neqo-transport/src/cc/classic_cc.rs
martinthomson: Alternatively, to save on duplication, you might want to make a variable for this, so that you can just do:

```rust
let mut now = now();
// ...
for k in 0..10 {
now += RTT;
for i in 0..BELOW_APP_LIMIT_PKTS {
}
now += RTT;
}
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -956,63 +975,48 @@ mod tests {
...
+        // simulate 10 packet bursts below app_limit
+        for k in 0..10 {
+            let k_u32 = u32::try_from(k).unwrap();
+            // always stay below app_limit during sent.
+            let mut pkts = Vec::new();
+            for i in 0..BELOW_APP_LIMIT_PKTS {
+                let p = SentPacket::new(
+                    PacketType::Short,
+                    u64::try_from(k * BELOW_APP_LIMIT_PKTS + i + 3).unwrap(), // pn
+                    now() + (k_u32 + 2) * RTT,                             // time sent
```
martinthomson, line 991: Same comment here as below.

neqo-transport/src/cc/classic_cc.rs
mb: Before packet 0 was always non-app-limited so increased the cwnd by 1MSS (if not lost). Some tests checked if only the first packet was send+received, that the cwnd stayed the same. To keep these tests running, I switched to this approach. It feels a little bit like a workaround, but I would be ok with it.

martinthomson: Oh, I see what is going on now.  That's unfortunate.  I would prefer that we increase CWND for the first packet.  We should increase CWND if we get an ACK; skipping the first packet for that makes no real sense.  Especially when you consider that for the entirety of slow start, that missed packet will double its effect in terms of lost opportunity to send each round trip.

If the scope of changes here is too large, then maybe we can make the change separately, but I think that we should change what tests expect from the CWND.

mb:
Thanks for review so far. I adjusted the test case, so that we test packet burst up to `BELOW_APP_LIMIT` and check that the congestion window is unaffected. And then added one packet burst with `ABOVE_APP_LIMIT` packets and check that each packet in the burst affects the congestion window.

neqo-transport/src/cc/classic_cc.rs
mb: The 0 packet still does count towards cwnd if we utilize the half of the cwnd during the first RTT. It only doesn't count towards cwnd if we sent out 5 or less packets. That might be fine?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -978,131 +986,184 @@ mod tests {
...
+                    MAX_DATAGRAM_SIZE, // size
+                );
+                next_pn += 1;
+                cc.on_packet_sent(&p);
+                pkts.push(p);
+            }
+            assert_eq!(cc.bytes_in_flight(), packet_burst_size * MAX_DATAGRAM_SIZE);
+            now += RTT;
+            for (i, pkt) in pkts.into_iter().enumerate() {
+                cc.on_packets_acked(&[pkt], RTT, now);
```
martinthomson, line 1016: Is there any reason you couldn't do this instead?

```suggestion
            cc.on_packets_acked(&pkts, RTT, now);
```

The assertions would get easier as well; you would just check that bytes_in_flight is 0 and the cwnd hasn't grown.

I suggest this because it's a piece of the test that you aren't really interested in; we're still in a setup phase.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -978,131 +986,184 @@ mod tests {
...
+            cc.on_packets_acked(&[pkt], RTT, now);
 
             assert_eq!(
                 cc.bytes_in_flight(),
-                (CWND_INITIAL_PKTS - i - 1) * MAX_DATAGRAM_SIZE
+                (ABOVE_APP_LIMIT_PKTS - i - 1) * MAX_DATAGRAM_SIZE
             );
-            assert_eq!(cc.cwnd(), (CWND_INITIAL_PKTS + 4) * MAX_DATAGRAM_SIZE);
+            // increase acked_bytes with each packet
+            println!("{} {}", cc.congestion_window, cwnd + i * MAX_DATAGRAM_SIZE);
```
martinthomson, line 1056: Want to change this to `qinfo!()` or remove it?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -978,131 +986,184 @@ mod tests {
...
+                assert_eq!(
+                    cc.bytes_in_flight(),
+                    (packet_burst_size - i - 1) * MAX_DATAGRAM_SIZE
+                );
+                cwnd_is_halved(&cc); // CWND doesn't grow because we're app limited
+                assert_eq!(cc.acked_bytes, 0);
+            }
+        }
+
+        // Utilize congestion window and increase it for each ACKs arriving
```
martinthomson, line 1135: I don't know what this comment means, particularly "Utilize".  Do you intend to say that you are now looking to increase the congestion window by having packets sent and then acknowledged?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -978,131 +986,184 @@ mod tests {
...
+            cc.on_packets_acked(&[pkt], RTT, now);
 
             assert_eq!(
                 cc.bytes_in_flight(),
-                (CWND_PKTS_CA - i - 1) * MAX_DATAGRAM_SIZE
+                (ABOVE_APP_LIMIT_PKTS - i - 1) * MAX_DATAGRAM_SIZE
             );
-            assert_eq!(cc.cwnd(), CWND_PKTS_CA * MAX_DATAGRAM_SIZE);
-            assert_eq!(cc.acked_bytes, MAX_DATAGRAM_SIZE * 3);
+            cwnd_is_halved(&cc);
```
martinthomson, line 1163: It might be worth noting here that the cwnd doesn't increase, but the acknowledged bytes do, which will eventually lead to an increase, once the number of bytes reaches the necessary level.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -978,131 +986,184 @@ mod tests {
...
+
+                assert_eq!(cwnd, cc.congestion_window); // CWND doesn't grow because we're app limited
+                assert_eq!(cc.acked_bytes, 0);
+            }
         }
-        assert_eq!(cc.bytes_in_flight(), CWND_INITIAL);
 
-        for i in 0..LESS_THAN_CWND_PKTS {
-            let acked = SentPacket::new(
+        // Utilize congestion window and increase it for each ACKs arriving
```
martinthomson, line 1028: See below regarding the comment.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -323,6 +322,13 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         if !pkt.cc_in_flight() {
             return;
         }
+        if !self.app_limited() {
+            // Given the current non-app-limited condition, we're fully utilizing the congestion
+            // window. We can safely consider that all in-flight packets up to the current to be not
+            // app-limiting. However, the subsequent packets might fall under app-limited
+            // condition. Hence, set the `first_app_limited` to the next packet number.
```
martinthomson, line 329:
```suggestion
            // window. Assume that all in-flight packets up to this one are NOT
            // app-limited. However, subsequent packets might be app-limited.
            // Set `first_app_limited` to the next packet number.
```

martinthomson (approved):

neqo-transport/src/cc/classic_cc.rs
mb: Utilize is meant as opposite of underutilized from the rfc: https://datatracker.ietf.org/doc/html/rfc9002#name-underutilizing-the-congesti

PR #1482, opened 2023-11-01
Use packet number instead of time sent to identify packets sent after recovery event

mb:
This doesn't change behavior, simplifies after_recovery_start check by removing the ambiguousness when the recovery packet was sent at the same time as some other packet.

Noticed while implementing #1465 that I would prefer using pn instead of time_stamp.


martinthomson (approved):
Same thing, right?

Is there any observable difference here? The packet number will be smaller, so this is an improvement, but I don’t see any real change as a result.

PR #1484, opened 2023-11-01
Add qlog traces to make debugging congestion controller easier

mb:
This are the log events I use to analyze the congestion controller. I have simply put print statements for my analysis, but I'd like to request qlogs with these events from other users reporting H3 upload speed issues. Therefore having them in tree with qlog would be good for me.

Depending how stable we need the qlog events to be, this might not be sufficient due to modifying existing qlog events to not have duplicate information. But if we're fine with changing the qlog output then this PR might already be good.


martinthomson (approved):
Please change the name of the pull request. “qlog” means something very specific and this is just improved debug logging.

mb:
The "q" prefix of the macro has been misleading me ^^

martinthomson:
I believe that our q prefix predates qlog.  We could rename it, I guess, but it seems like it wouldn't do that much good.

PR #1488, opened 2023-11-05
ACK immediately after an RTT

martinthomson:
This should improve loss recovery performance if we haven't been active for a short while.

The actual change is fairly trivial, but the tests needed to be tweaked in several ways to deal with a few places where we generated an immediate acknowledgment.

I also found a test where I couldn't work out why it ever passed before.  One of the migration tests was sending multiple packets without acknowledgment on a new path, which had the default RTT.  It was therefore pacing for the third packet.  I don't know how that ever passed.


KershawChang (approved):
Looks reasonable to me. Thanks!

PR #1491, opened 2023-11-08
A test to trigger 'earliest > now' assertion

KershawChang:
This test demonstrates one possible case that how the assertion is triggered.
Basically, this happens when the state of `IdleTimeout` is set to `IdleTimeoutState::AckElicitingPacketSent` and we haven't receive any packets after that.

Note that in [expired](https://searchfox.org/mozilla-central/rev/02841791400cf7cf5760c0cfaf31f5d772624253/third_party/rust/neqo-transport/src/connection/idle.rs#85) function, we set `keep_alive` to false, so the [expiry](https://searchfox.org/mozilla-central/rev/02841791400cf7cf5760c0cfaf31f5d772624253/third_party/rust/neqo-transport/src/connection/idle.rs#85) function returns a value larger than `now` and we don't think the idle timer is expired.
However, when `expiry` is called in `next_delay`, we set `keep_alive` to true, so the value returns from `expiry` is smaller than `now`.


neqo-transport/tests/server.rs
```diff
@@ -56,6 +56,47 @@ pub fn complete_connection(
...
+fn connect_to_nothing() {
+    // Start a server to get the token that will be used for calling
+    // `client.enable_resumption`.
+    let mut server = default_server();
+    let token = generate_ticket(&mut server);
+
+    // Discharge the old connection so that we don't have to worry about it.
+    let mut now = now();
+    let t = server.process(None, now).callback();
+    now += t;
```
martinthomson, line 69: Do you really need this given that the code below doesn't even use the server?

neqo-transport/tests/server.rs
```diff
@@ -56,6 +56,47 @@ pub fn complete_connection(
...
+        match client.process(None, now) {
+            Output::Datagram(_) | Output::None => None,
+            Output::Callback(t) => Some(t),
+        }
+    };
+
+    let mut count = 0;
+    // Run a loop to create packets repeatedly until we hit the "earliest > now"
+    // assertion.
+    loop {
```
martinthomson, line 89:
```rust
for _ in 0..100 {
    if let Some(t) = client_send(now) {
        now += t;
    }
}
```

neqo-transport/tests/server.rs
```diff
@@ -56,6 +56,47 @@ pub fn complete_connection(
...
+    let mut client = default_client();
+    // After enable_resumption, the client have the stream limit settings from
+    // the server, so we can create a stream.
+    client.enable_resumption(now, &token).unwrap();
+
+    let client_stream = client.stream_create(StreamType::BiDi).unwrap();
+    client.stream_keep_alive(client_stream, true).unwrap();
+    let mut client_send = |now| {
+        client.stream_send(client_stream, &[0x0; 500]).unwrap();
+        match client.process(None, now) {
```
martinthomson, line 80: Does this change anything?

```suggestion
        match client.process_output(now) {
```

martinthomson (commented):
Code comments only.  Once we have a fix, we should land this test with that fix.

martinthomson:
Is the fix to propagate the `keep_alive` flag through [expired](https://searchfox.org/mozilla-central/rev/02841791400cf7cf5760c0cfaf31f5d772624253/third_party/rust/neqo-transport/src/connection/idle.rs#85)?

neqo-transport/tests/server.rs
KershawChang: Yes, I think we need it because this seems to be the only to get the transport parameters from peer.

neqo-transport/tests/server.rs
KershawChang: No, same result ('earliest > now' assertion).

KershawChang:
> Is the fix to propagate the `keep_alive` flag through [expired](https://searchfox.org/mozilla-central/rev/02841791400cf7cf5760c0cfaf31f5d772624253/third_party/rust/neqo-transport/src/connection/idle.rs#85)?

Yes, this is one of the fixes in my mind. I'll update this PR soon.

neqo-transport/tests/server.rs
martinthomson: The client should be remembering transport parameters through the session ticket (token) that it gets above.

neqo-transport/tests/server.rs
martinthomson: Good.  Just confirming.  I generally prefer not to use `process(None, ...)` in place of `process_output(...)`.

martinthomson (commented):
> If I understand this fix correctly, doing this means that the connection will closed when the keep_alive timer is expired.

No, just that we would compute the idle timeout based on whether a keep alive is active.  We'd still send a ping when that shorter time elapsed, but we wouldn't have this drift issue you refer to.

One thing that is reassuring about this situation is that it only occurs when the server goes AWOL, as this test demonstrates.  That's also troubling also because it's a crash, so it's good to see that you caught this one.

KershawChang:
I've moved the test to `idle.rs`, since it seems to be a better place for testing idle timeout.
Note this PR breaks some tests, so I am not sure if this is the change we want.

martinthomson:
Hmm, yeah, this doesn't look right.  We should not be closing the connection at this point.  I think that we need to look for another approach.

martinthomson:
Thinking on this more, I think that we probably need to consider tweaks to the expiry() function and how that works.  What we want is to have this track when keep-alive packets were sent and either provide updated estimates for the next timeout based on that or simply switch to reporting the `keep_alive == false` timeout when a keep-alive has been sent.

My understanding is that we'll retransmit the keep-alive using the usual logic for lost packets (PTO, etc...) so there is no need to continue to report the keep-alive time once we've initiated the keep-alive.

So maybe the idle code needs to be told when the keep-alive is sent so that it can stop reporting the shorter timeout.

larseggert:
@KershawChang do we have a way forward for this PR?

larseggert:
Making this a draft PR until it becomes reviewable again.

KershawChang:
> Thinking on this more, I think that we probably need to consider tweaks to the expiry() function and how that works. What we want is to have this track when keep-alive packets were sent and either provide updated estimates for the next timeout based on that or simply switch to reporting the `keep_alive == false` timeout when a keep-alive has been sent.
> 
> My understanding is that we'll retransmit the keep-alive using the usual logic for lost packets (PTO, etc...) so there is no need to continue to report the keep-alive time once we've initiated the keep-alive.
> 
> So maybe the idle code needs to be told when the keep-alive is sent so that it can stop reporting the shorter timeout.

@martinthomson , hope I understand this comment correctly. I think we can make `expiry` always return the idle timeout and introduce another function `keep_alive_timeout` that is only used for sending `keep-alive`.

Note that I didn't fix test failures in this PR. I'll fix them if you think this is a right approach. Thanks.

neqo-transport/src/connection/idle.rs
```diff
@@ -46,20 +46,14 @@ impl IdleTimeout {
         self.timeout = min(self.timeout, peer_timeout);
     }
 
-    pub fn expiry(&self, now: Instant, pto: Duration, keep_alive: bool) -> Instant {
+    pub fn expiry(&self, now: Instant, pto: Duration) -> Instant {
         let start = match self.state {
             IdleTimeoutState::Init => now,
             IdleTimeoutState::PacketReceived(t) | IdleTimeoutState::AckElicitingPacketSent(t) => t,
         };
```
martinthomson, line 53: It might pay to factor this into a function now to save repetition.

neqo-transport/src/connection/idle.rs
```diff
@@ -46,20 +46,14 @@ impl IdleTimeout {
...
-        } else {
-            max(self.timeout, pto * 3)
-        };
+        let delay = max(self.timeout, pto * 3);
         qtrace!(
-            "IdleTimeout::expiry@{now:?} pto={pto:?}, ka={keep_alive} => {t:?}",
+            "IdleTimeout::expiry@{now:?} pto={pto:?} => {t:?}",
             t = start + delay
         );
         start + delay
```
martinthomson, line 59: Looking at this, perhaps we can simplify:

```rust
let t = start + delay;
qtrace!("IdleTimeout::expiry@{now:?} pto={pto:?} => {t:?}");
t
```

Or with the above factoring...

```rust
let t = self.start() + delay;
qtrace!("IdleTimeout::expiry@{now:?} pto={pto:?} => {t:?}");
t
```

martinthomson (approved):

neqo-transport/src/connection/idle.rs
KershawChang: Will do.

neqo-transport/src/connection/idle.rs
KershawChang: Will do.

KershawChang:
> I'm seeing some build failures here that are probably worth looking into.

@martinthomson , I've fixed those test failures.
Could you take a look again? Thanks.

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -709,12 +706,6 @@ fn keep_alive_with_ack_eliciting_packet_lost() {
     let retransmit = client.process_output(now).dgram();
     assert!(retransmit.is_some());
 
-    // The next callback should be for an idle PING.
-    assert_eq!(
-        client.process_output(now).callback(),
-        IDLE_TIMEOUT / 2 - pto
```
KershawChang, line 715: With this change, this timeout is actually `pto * 2`, because the idle timeout is longer.
I've removed this check since we already did the same check below.

larseggert:
@KershawChang I merged this into #1875 locally, but the `idle_timeout_crazy_rtt` still asserts on `earliest > now`. FWIW.

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -564,17 +561,18 @@ fn keep_alive_reset() {
...
     transfer_force_idle(&mut client, &mut server);
-    assert_idle(&mut client, now(), default_timeout() / 2);
+    assert_idle(&mut client, now(), default_timeout());
 
     server.stream_reset_send(stream, 0).unwrap();
     transfer_force_idle(&mut server, &mut client);
     assert_idle(&mut client, now(), default_timeout());
 
     // The client will fade away from here.
+    eprintln!("The client will fade away from here.");
```
martinthomson, line 575:
```suggestion
```

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -709,12 +706,6 @@ fn keep_alive_with_ack_eliciting_packet_lost() {
     let retransmit = client.process_output(now).dgram();
     assert!(retransmit.is_some());
 
-    // The next callback should be for an idle PING.
-    assert_eq!(
-        client.process_output(now).callback(),
-        IDLE_TIMEOUT / 2 - pto
-    );
-
     // Wait that long and the client should send a PING frame.
```
martinthomson, line 709:
```suggestion
    // Wait for half the idle timeout (less the PTO we've already waited)
    // so that we get a keep-alive. The actual timeout is the smaller of this
    // duration or PTO*2, whichever is smaller.
```

neqo-transport/src/connection/tests/idle.rs
martinthomson: Maybe we need to make these checks more robust, rather than removing them.  Would this work?

```rust
    // The timeout is the smaller of:
    // * the idle timeout, less the time we've already waited (one PTO)
    // * twice the PTO, because we've already sent one probe
    assert_eq!(
        client.process_output(now).callback(),
        min(IDLE_TIMEOUT / 2 - pto, pto * 2),
    );
```

That would mean updating the code below as well.

The alternative is to sort out the relationship between the PTO timer we're using and the idle timeout, so that we can know for sure which timer is active.

martinthomson (approved):
I'm happy with this.  Want to flip out of draft state?

Think about the comment below, but I'll defer to your judgment on the resolution once you've worked through it.

neqo-transport/src/connection/tests/idle.rs
KershawChang: Sorry about this.

neqo-transport/src/connection/tests/idle.rs
KershawChang: Done. Thanks.

neqo-transport/src/connection/tests/idle.rs
KershawChang: I think this should be:
```
assert_eq!(
    client.process_output(now).callback(),
    min(IDLE_TIMEOUT - pto, pto * 2),
);
```
Because we only waited one pto here.

martinthomson: IDLE_TIMEOUT or IDLE_TIMEOUT/2 ?

KershawChang: Should be `IDLE_TIMEOUT`.

As the test code below shows, the `start` of `IdleTimeout` is set to `now` when `_lost_packet` is sent.
We then wait for one `pto` duration to retransmit the data.
After the retransmission, `idle_timeout.expiry` returns `now + IDLE_TIMEOUT`.
Since we have only waited for one `pto` in the test, the remaining time before the idle timer expires 
is `IDLE_TIMEOUT - pto`. Therefore, in the assertion, we should use `min(IDLE_TIMEOUT - pto, pto * 2)`.


```
    // Send data on the stream that will be lost.
    _ = client.stream_send(stream, DEFAULT_STREAM_DATA).unwrap();
    let _lost_packet = client.process_output(now).dgram();

    let pto = client.process_output(now).callback();
    // Wait for packet to be marked lost.
    assert!(pto < IDLE_TIMEOUT / 2);
    now += pto;
    let retransmit = client.process_output(now).dgram();
    assert!(retransmit.is_some());
    let retransmit = client.process_output(now).dgram();
    assert!(retransmit.is_some());

    // The timeout is the smaller of:
    // * the idle timeout, less the time we've already waited (one PTO)
    // * twice the PTO, because we've already sent one probe
    assert_eq!(
        client.process_output(now).callback(),
        min(IDLE_TIMEOUT - pto, pto * 2),
    );
```

KershawChang: @martinthomson , please let me know if my understanding is correct.
If you are fine with this change, I think we can land this.

martinthomson: I think my suggested comment text was incorrect.

We have three operative timers:
* PTO (which is `pto * 2` at this stage)
* The idle timeout
* The keep alive timer, which is half of the idle timeout.

We can disregard the idle timeout, leaving the other two.

KershawChang: Thanks, I'll update the patch and only compare the PTO.

KershawChang:
Please take another look at test `keep_alive_with_ack_eliciting_packet_lost`.
Thanks.

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -709,21 +705,20 @@ fn keep_alive_with_ack_eliciting_packet_lost() {
...
-        client.process_output(now).callback(),
-        IDLE_TIMEOUT / 2 - pto
-    );
+    // The timeout is the twice the PTO, because we've already sent one probe.
+    assert_eq!(client.process_output(now).callback(), pto * 2,);
 
-    // Wait that long and the client should send a PING frame.
+    // Wait for half the idle timeout (less the PTO we've already waited)
+    // so that we get a keep-alive. The actual timeout is the smaller of this
+    // duration or PTO*2, whichever is smaller.
```
martinthomson, line 713:
```suggestion
    // so that we get a keep-alive.
```

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -709,21 +705,20 @@ fn keep_alive_with_ack_eliciting_packet_lost() {
...
     now += IDLE_TIMEOUT / 2 - pto;
     let pings_before = client.stats().frame_tx.ping;
     let ping = client.process_output(now).dgram();
     assert!(ping.is_some());
     assert_eq!(client.stats().frame_tx.ping, pings_before + 1);
 
     // The next callback is for a PTO, the PTO timer is 2 * pto now.
-    assert_eq!(client.process_output(now).callback(), pto * 2);
+    assert_eq!(client.process_output(now).callback(), pto * 2,);
```
martinthomson, line 721:
```suggestion
    assert_eq!(client.process_output(now).callback(), pto * 2);
```

martinthomson (approved):

neqo-transport/src/connection/tests/idle.rs
KershawChang: Great catch! Thanks.
I am surprised that compiler doesn't complain about this.

martinthomson: Macro formatting is still imperfect.  The compiler knows what sorts of symbols need to appear and can sort of guess, but it can't really know whether the macro cares about a trailing comma or not.

PR #1495, opened 2023-11-21
Being refactored: ECN additions

larseggert:
This PR is getting too large. I'm in the process of refactoring it into smaller chunks:

- [x] #1568
- [x] #1604 
- [ ] #1678


Cargo.toml
```diff
@@ -10,3 +10,9 @@ members = [
   "neqo-interop",
   "test-fixture",
 ]
+
+[patch.crates-io]
+#nix = { path = "./../nix" }
+#libc = { path = "./../libc" }
+libc = { git = "https://github.com/larseggert/libc.git", branch = "feat-ttl" }
+nix = { git = "https://github.com/larseggert/nix.git", branch = "feat-tos" }
```
martinthomson, line 18: We won't be able to land a change that pulls from a private branch, at least not without an extraordinary approval process.  I assume that you are in the process of landing changes upstream.

neqo-helper/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+        },
+        AddressFamily, ControlMessage, ControlMessageOwned, MsgFlags, SockaddrLike,
+        SockaddrStorage,
+    },
+};
+
+use neqo_common::Datagram;
+
+// Bind a UDP socket and set some default socket options.
+pub fn bind(local_addr: SocketAddr) -> io::Result<UdpSocket> {
```
martinthomson, line 30: The alternative to creating a new crate is to add this to a new module in neqo-common.  A non-default feature flag that makes this module conditionally available would ensure that mainline code doesn't need to pull in the nix and libc dependencies.

A new crate isn't that big of a deal if you only use it for the command-line tools, but I can see how we might want to use these facilities in Firefox proper, for which having it in neqo-common might help.

neqo-helper/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    Ok(())
+}
+
+fn to_socket_addr(addr: &SockaddrStorage) -> SocketAddr {
+    let fam = addr.family().unwrap();
+
+    let port = match fam {
+        AddressFamily::Inet => addr.as_sockaddr_in().unwrap().port(),
+        AddressFamily::Inet6 => addr.as_sockaddr_in6().unwrap().port(),
+        _ => todo!(),
```
martinthomson, line 100:
```suggestion
        _ => unreachable!(),
```

I think that we're safe with that in this case.  IPv10 is still a ways off.

neqo-helper/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    match fam {
```
martinthomson, line 103: You should be able to do just one match.  You are repeating the `addr.as_sockaddr_in[6]().unwrap()` part as it is.

neqo-common/src/datagram.rs
```diff
@@ -9,10 +9,52 @@ use std::ops::Deref;
...
+pub enum IpTosEcn {
+    EcnNotEct = 0b00, // Not-ECT (Not ECN-Capable Transport) [RFC3168]
+    EcnEct1 = 0b01,   // ECT(1) (ECN-Capable Transport(1))[1] [RFC8311][RFC Errata 5399][RFC9331]
+    EcnEct0 = 0b10,   // ECT(0) (ECN-Capable Transport(0)) [RFC3168]
+    EcnCe = 0b11,     // CE (Congestion Experienced) [RFC3168]
+}
+
+// DiffServ Codepoints, mapped to the upper six bits of the TOS field.
+// https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml
+// #[derive(Copy, Clone)]
```
martinthomson, line 25:
```suggestion
#[derive(Copy, Clone, PartialEq, Eq)]
```

neqo-common/src/datagram.rs
```diff
@@ -9,10 +9,52 @@ use std::ops::Deref;
 
 use crate::hex_with_len;
 
+// ECN (Explicit Congestion Notification) codepoints mapped to the
+// lower 2 bits of the TOS field.
+// https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml
+// #[derive(Copy, Clone)]
```
martinthomson, line 15:
```suggestion
#[derive(Copy, Clone, PartialEq, Eq)]
```

neqo-helper/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+        AddressFamily::Inet6 => SocketAddr::new(
+            std::net::IpAddr::V6(addr.as_sockaddr_in6().unwrap().ip()),
+            port,
+        ),
+        _ => todo!(),
+    }
+}
+
+pub fn recv_datagram(
+    fd: i32,
```
martinthomson, line 117: How often do we have a raw fd when calling this function?  Can this function call `to_raw_fd()` itself?

neqo-helper/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    };
+    Ok(socket)
+}
+
+fn to_sockaddr(addr: SocketAddr) -> SockaddrStorage {
+    SockaddrStorage::from(addr)
+}
+
+pub fn emit_datagram(fd: i32, d: Datagram) -> io::Result<()> {
```
martinthomson, line 68: See below regarding `fd`

martinthomson (commented):
I take it that this is groundwork for ECN

Cargo.toml
larseggert: Yep. This is just temporary so the CI runs. I will remove this once upstreams are updated.

neqo-helper/src/lib.rs
larseggert: I couldn't easily figure out how, because it's being called with a `mio::net::UdpSocket` from the server and a `std::net::UdpSocket` from the client. Will investigate.

neqo-helper/src/lib.rs
larseggert: Thanks for the feedback. Will need to read up on those concepts.

neqo-helper/src/lib.rs
martinthomson: Oh, I missed that.  Unless the two implement a shared trait, you won't be able to.  But then it is possible to have that happen.  Whether it is worthwhile, I'll leave as an exercise :)

The basic idea here would be that you have two types that can feed in, so you would have a function signature that is generic:

```rust
pub fn recv_datagram<S: SocketTrait>(socket: S, ...) {
}
```

Then, to allow you to pass in either form, you would define that new trait with the functions you need as follows:

```rust
pub trait SocketTrait { // don't use my name
  fn raw_fd(&self) -> i32;
}
```

You can then implement that trait for the different types you want to support, like so:

```rust
impl SocketTrait for std::net::UdpSocket {
  fn raw_fd(&self) -> i32 { self.get_raw_fd() }
}


impl SocketTrait for mio::net::UdpSocket {
  fn raw_fd(&self) -> i32 { self.get_raw_fd() }
}
```

This means that your code can have multiple instantiations, one for each concrete type that implements this new trait.  In practice, you want one instantiation in the client and another in the server, so there is no code bloat.

This would make the interface more ergonomic, but it's not clear that we really need that capability in this instance.  I made the comment without realizing that there were different types involved.

neqo-helper/src/lib.rs
larseggert: @martinthomson I think I made that change.

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,217 @@
...
+/// On non-POSIX platforms, TOS and TTL will be set to 0xff to indicate they were not retrieved
+/// from the packet. TODO: Figure out how to retrieve TOS and TTL at least on Windows.
+pub fn recv_datagram<S: AsRawFd>(
+    socket: &S,
+    buf: &mut [u8],
+    tos: &mut u8,
+    ttl: &mut u8,
+) -> io::Result<(usize, SocketAddr)> {
+    #[cfg(any(target_os = "macos", target_os = "linux", target_os = "android"))]
+    {
```
martinthomson, line 182: I would recommend extracting these two blocks into (private) functions.

neqo-server/src/main.rs
```diff
@@ -710,7 +711,9 @@ impl ServersRunner {
         match self.server.process(dgram, self.args.now()) {
             Output::Datagram(dgram) => {
                 let socket = self.find_socket(dgram.source());
-                emit_packet(socket, dgram);
+                if let Err(e) = emit_datagram(socket, &dgram) {
+                    eprintln!("UDP write error: {}", e);
```
martinthomson, line 715: Do you want to open an issue to track EAGAIN handling in these?

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,217 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::{
+    io::{self, Error, ErrorKind, IoSlice, IoSliceMut},
+    net::{SocketAddr, UdpSocket},
+    os::fd::{AsRawFd, FromRawFd},
```
martinthomson, line 10: `AsRawFd` = win.

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,217 @@
...
+            SocketAddr::V6(..) => [Ipv6TClass(&tos), Ipv6HopLimit(&ttl)],
+        };
+        sent = sendmsg(
+            socket.as_raw_fd(),
+            &iov,
+            &cmsgs,
+            MsgFlags::empty(),
+            Some(&SockaddrStorage::from(d.destination())),
+        )
+        .unwrap();
```
martinthomson, line 121: These uses of `unwrap()` still concern me a little; convince me that this is rather than wrapping the error in a new type.  You aren't using the `io::Result::Err` branch on this either.

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,217 @@
...
+            panic!("Unable to bind UDP socket: {}", e);
+        }
+        Ok(s) => {
+            // Don't let the host stack or network path fragment our IP packets
+            // (RFC9000, Section 14).
+            let res = match local_addr {
+                SocketAddr::V4(..) => setsockopt(&s, IpDontFrag, &true),
+                SocketAddr::V6(..) => setsockopt(&s, Ipv6DontFrag, &true),
+            };
+            assert!(res.is_ok());
```
martinthomson, line 62:
```suggestion
            debug_assert!(res.is_ok());
```

If this fails, let's not worry too much.

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,217 @@
...
+            // Request IPv4 type-of-service (TOS) and IPv6 traffic class
+            // information for all incoming packets.
+            let res = match local_addr {
+                SocketAddr::V4(..) => setsockopt(&s, IpRecvTos, &true),
+                SocketAddr::V6(..) => setsockopt(&s, Ipv6RecvTClass, &true),
+            };
+            assert!(res.is_ok());
```
martinthomson, line 69: As above, ...

martinthomson (dismissed):
I've a question for you and some suggestions, but I trust you to resolve those from here (and I don't want to block you while I'm on leave).

neqo-common/src/socket.rs
```diff
@@ -26,12 +25,31 @@ use nix::{
...
+///
+/// Panics if the UDP socket fails to bind to the specified local address or if various
+/// socket options cannot be set.
+///
+/// # Notes
+///
+/// This function binds the UDP socket to the specified local address and performs additional
+/// configuration on the socket, such as setting socket options to request TOS and TTL
+/// information for incoming packets.
+#[allow(clippy::missing_errors_doc)]
```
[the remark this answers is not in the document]
larseggert: This is probably because I am new to Rust, but I don't see _how_ this can return an error? AFAICT it can only panic?

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,249 @@
...
+    let sent = emit_datagram_posix(socket, d);
+    #[cfg(not(posix_socket))]
+    let sent = emit_datagram_generic(socket, d);
+    if sent != d.len() {
+        eprintln!("Unable to send all {} bytes of datagram", d.len());
+    }
+    Ok(())
+}
+
+// #[cfg(posix_socket)]
```
martinthomson, line 150: ???

larseggert: Leftover debug thing; am still trying to get this to build/run on Windows.

larseggert:
@martinthomson could you re-review? Made a bunch of changes that make things work on all platforms again.

larseggert:
Blocked until https://github.com/nix-rust/nix/pull/2203 has shipped.

neqo-transport/src/recovery.rs
```diff
@@ -412,7 +412,7 @@ impl LossRecoverySpace {
             .sent_packets
             .iter_mut()
             // BTreeMap iterates in order of ascending PN
-            .take_while(|(&k, _)| Some(k) < largest_acked)
+            .take_while(|(&k, _)| largest_acked.is_none() || Some(k) < largest_acked)
```
martinthomson, line 415: This sort of change really needs to be in a separate patch so that we can look at the recovery tests in more detail.

neqo-common/Cargo.toml
```diff
@@ -7,16 +7,27 @@ rust-version = "1.70.0"
...
+env_logger = { version = "0.10", default-features = false }
 lazy_static = "1.3.0"
 qlog = "0.11.0"
-time = {version = "0.3", features = ["formatting"]}
+time = { version = "=0.3.23", features = ["formatting"] }
+nix = { git = "https://github.com/larseggert/nix.git", branch = "feat-tos", features = [
+        "socket",
+        "net",
+        "uio",
+], optional = true }
```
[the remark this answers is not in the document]
larseggert: It's in progress. Depends on another PR upstream.

neqo-common/src/datagram.rs
```diff
@@ -21,6 +89,24 @@ impl Datagram {
         Self {
             src,
             dst,
+            tos: IpTosEcn::Ect0 as u8,
```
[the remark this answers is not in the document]
larseggert: I think I did this elsewhere, must have missed this one.

neqo-common/src/socket.rs
```diff
@@ -0,0 +1,429 @@
...
+        // MacOS has a kernel bug that prevents setting the TTL via CMSG, so just check that it is not 0.
+        assert_ne!(ttl, 0);
+        assert_eq!(buf, [0x42; 16]);
+
+        drop(client);
+        drop(server);
+    }
+
+    #[test]
+    fn test_bind_v4() {
```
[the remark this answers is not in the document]
larseggert: ACK

neqo-transport/src/recovery.rs
larseggert: Yeah, I realized this is getting big. I will break it into individual chunks.

mxinden:
Crossposting https://github.com/mxinden/neqo/pull/1#pullrequestreview-1892959112 here to ensure review on fork does not get lost.

larseggert:
Nothing left in this PR that would be of interest.

PR #1500, opened 2023-11-28
Expand the CI matrix and generate coverage information

larseggert:
~~TODO: Support Windows.~~

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -39,8 +39,6 @@ jobs:
       - name: Install dependencies (MacOS)
         if: runner.os == 'MacOS'
         run: |
-          python -v || true
-          brew search python || true
           brew install --force --overwrite ninja mercurial
           python3 -m pip install gyp-next
           echo "$(python3 -m site --user-base)/bin" >> "$GITHUB_PATH"
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -13,8 +13,8 @@ jobs:
         rust-toolchain: [1.65.0, stable] #, beta]
     runs-on: ${{ matrix.os }}
     defaults:
-          run:
-            shell: ${{ matrix.os == 'windows-latest' && 'msys2 {0}' || 'bash' }}
+      run:
+        shell: ${{ matrix.os == 'windows-latest' && 'msys2 {0}' || 'bash' }}
 
     steps:
       - name: Install Rust
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,41 +14,56 @@ jobs:
     runs-on: ${{ matrix.os }}
     defaults:
       run:
-        shell: ${{ matrix.os == 'windows-latest' && 'msys2 {0}' || 'bash' }}
+        shell: bash
 
     steps:
       - name: Install Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust-toolchain }}
           components: rustfmt, clippy
-
-      - name: Install Rust tools
-        uses: taiki-e/install-action@v2
-        with:
-          tool: nextest, cargo-llvm-cov
 
       - name: Install dependencies (Linux)
         if: runner.os == 'Linux'
         env:
           DEBIAN_FRONTEND: noninteractive
         run: |
-          sudo apt-get update
           sudo apt-get install -y --no-install-recommends gyp mercurial ninja-build
+          curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
 
       - name: Install dependencies (MacOS)
         if: runner.os == 'MacOS'
         run: |
-          brew install --force --overwrite ninja mercurial
-          python3 -m pip install gyp-next
-          echo "$(python3 -m site --user-base)/bin" >> "$GITHUB_PATH"
+          brew install --force --overwrite ninja mercurial cargo-binstall || brew link --overwrite python@3.11
+          # python3 -m pip install gyp-next
+          # Above does not work, since pypi only has gyp 0.15.0, which is too old
+          # for the homebrew python3. Install from source instead.
+          python3 -m pip install git+https://github.com/nodejs/gyp-next
+          python3 -m pip install packaging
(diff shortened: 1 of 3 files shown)


larseggert:
So the CI run for this fails with
```
actions-rust-lang/setup-rust-toolchain@v1, taiki-e/install-action@v2, and msys2/setup-msys2@v2 are not allowed to be used in mozilla/neqo. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, or matching the following: !/mozilla/**, !mozilla/**, ./**, 10up/wpcs-action@*, aws-actions/*, docker/*, pypa/gh-action-pypi-publish@v1.4.2, slackapi/slack-github-action@*, google-github-actions/*, erlef/setup-beam@v1, yesolutions/mirror-action@*, codecov/codecov-action@*, tj-actions/changed-files@*, tj-actions/glob@*, vmactions/freebsd-vm@v0, actions-rs/toolchain@v1, shivammathur/setup-php@*, EmbarkStudios/*, dependabot/fetch-metadata@*, ilammy/msvc-dev-cmd@v1*, canonical/actions/*, canonical/setup-lxd@*, dtolnay/rust-toolchain@*, pypa/gh-action-pypi-publish@release/v1*.
```
I guess something needs to be allowlisted at the Mozilla org?

martinthomson:
> actions-rust-lang/setup-rust-toolchain@v1, taiki-e/install-action@v2, and msys2/setup-msys2@v2 

Yeah, the mozilla org has policies about the use of actions that will need to be adjusted.

We previously manually worked around the "actions-rust-lang/setup-rust-toolchain" one, basically by re-implementing it, but that's less feasible with the full matrix.  Since then, I have had success with "dtolnay/rust-toolchain", [which is allow-listed](https://bugzilla.mozilla.org/show_bug.cgi?id=1835662). 

We can ask on bugzilla for other actions [here](https://bugzilla.mozilla.org/enter_bug.cgi?product=mozilla.org&component=Github%3A+Administration).

.github/workflows/check.yml
```diff
@@ -5,101 +5,106 @@ env:
...
-    runs-on: ubuntu-22.04
-    env:
-      RUSTFLAGS: -C link-arg=-fuse-ld=lld
+    name: Build & test
     strategy:
+      fail-fast: false
       matrix:
-        rust-toolchain: [stable, 1.65.0]
+        os: [ubuntu-latest, macos-latest] #, windows-latest]
+        rust-toolchain: [1.65.0, stable] #, beta]
```
martinthomson, line 13: @mb wanted beta as an early warning for new lints.  We get less value from having the full matrix there, but given that we've seen errors in Mac builds, it might be OK to just do the full set.

.github/workflows/check.yml
```diff
@@ -5,101 +5,106 @@ env:
...
+    runs-on: ${{ matrix.os }}
+    defaults:
+          run:
+            shell: ${{ matrix.os == 'windows-latest' && 'msys2 {0}' || 'bash' }}
```
martinthomson, line 17:
```suggestion
      run:
        shell: ${{ matrix.os == 'windows-latest' && 'msys2 {0}' || 'bash' }}
```

.github/workflows/check.yml
```diff
@@ -5,101 +5,106 @@ env:
...
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust-toolchain }}
+          components: rustfmt, clippy
+
+      - name: Install Rust tools
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest, cargo-llvm-cov
```
martinthomson, line 29: Does `cargo install X` work for these two?  It seems to.  And it might be easier than getting approval for the action.

martinthomson (commented):
With codecov, this looks pretty good.

Shame about the NSS/Windows thing, or this would be so much easier.

.github/workflows/check.yml
larseggert: I put `beta` there as a reminder how to figure out how to only make it run informationally (i.e., have fails not be critical).

.github/workflows/check.yml
larseggert: Yup, it occurred to me to that this action might not be needed. Will replace it.

larseggert:
I made it work using the already-approved actions. I'm also getting closer to having Windows working. Will make this a draft PR until then.

larseggert:
Current status:
* Linux builds and tests
* MacOS builds, but tests have failures and there are clippy issues
* Windows builds (yay), but the tests fail completely and there are clippy issues

Questions:
1. Should I include fixes to the clippy issues in this PR?
2. Does anyone have any idea why the tests fail to execute at all on Windows?

.github/workflows/check.yml
```diff
@@ -5,72 +5,66 @@ env:
...
-    runs-on: ubuntu-22.04
-    env:
-      RUSTFLAGS: -C link-arg=-fuse-ld=lld
+    name: Build & test
     strategy:
+      fail-fast: false
       matrix:
-        rust-toolchain: [stable, 1.65.0]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        rust-toolchain: [1.65.0] #, stable] #, beta]
```
martinthomson, line 13: Any reason that you dropped stable from this?

martinthomson (approved):
FixedGIF

martinthomson:
I see a few test failures on macos that seem worth looking into.  A missing or incorrect `DYLD_LIBRARY_PATH` appears to be the issue, but I don't see a problem with the code that sets the value.

The other failures are such that we can correct them in other updates (hrtime test failures, clippy warnings)

.github/workflows/check.yml
larseggert: Oops. I'll put it back.

larseggert:
> I see a few test failures on macos that seem worth looking into. A missing or incorrect `DYLD_LIBRARY_PATH` appears to be the issue, but I don't see a problem with the code that sets the value.

@martinthomson where do you see those? All I see ar `hrtimer` issues. (Or I need more coffee.)

martinthomson:
Hmm, not there now, which means...

martinthomson:
For whatever reason, in case this comes up again: https://github.com/mozilla/neqo/actions/runs/7140768584/job/19446707656#step:13:139 was the CI run for [e639e5f](https://github.com/mozilla/neqo/pull/1500/commits/e639e5fdeefd9ddfb86ee66f45cddc6d750f06c5) that had a build failure (on the test step).

PR #1502, opened 2023-11-30
Make neqo-client support upload test

KershawChang:
This PR makes neqo-client support doing upload test.
The command is `./target/release/neqo-client http://127.0.0.1:4433/  --test upload  --upload-size 104857600`.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -469,6 +469,28 @@ enum StreamHandlerType {
     Upload,
 }
 
+impl StreamHandlerType {
+    fn make_handler(
+        handler_type: &Self,
+        url: &Url,
+        args: &Args,
+        all_paths: &mut Vec<PathBuf>,
+    ) -> Box<dyn StreamHandler> {
+        match handler_type {
+            Self::Download => {
+                let out_file = get_output_file(&url, &args.output_dir, all_paths);
+                Box::new(DownloadStreamHandler { out_file })
+            }
+            Self::Upload => Box::new(UploadStreamHandler {
+                data: vec![42; args.upload_size],
+                offset: 0,
+                chunk_size: 32768,
+                start: Instant::now(),
+            }),
+        }
+    }
+}
+
 struct DownloadStreamHandler {
     out_file: Option<File>,
 }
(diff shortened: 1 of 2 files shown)


neqo-client/src/main.rs
```diff
@@ -528,29 +631,46 @@ impl<'a> Handler<'a> {
...
         true
     }
+}
+
+fn to_headers(values: &[impl AsRef<str>]) -> Vec<Header> {
+    values
+        .iter()
+        .scan(None, |state, value| {
+            if let Some(name) = state.take() {
+                *state = None;
```
martinthomson, line 641: Do you need to assign after having called `.take()`?

neqo-client/src/main.rs
```diff
@@ -528,29 +631,46 @@ impl<'a> Handler<'a> {
         }
         true
     }
+}
+
+fn to_headers(values: &[impl AsRef<str>]) -> Vec<Header> {
+    values
+        .iter()
+        .scan(None, |state, value| {
```
martinthomson, line 639: Would slice::chunks() be easier to use here?

That is, `values.chunks(2).map(|c| Header::new(c[0].as_ref(), c[1].as_ref()))`  (That would panic if the number of values is not divisible by two, but that should not be allowed by the argument parser.

neqo-client/src/main.rs
```diff
@@ -612,21 +734,105 @@ impl<'a> Handler<'a> {
...
+            if self.streams.len() >= self.args.concurrency {
+                break;
+            }
+            if !self.upload_next(client) {
+                break;
+            }
+        }
+    }
+
+    fn upload_next(&mut self, client: &mut Http3Client) -> bool {
```
martinthomson, line 768: There seems to be a lot of overlap between this and the other code.  I'd suggest a refactoring, but it is a little hard to disentangle.  Can you try to find some more commonality?

neqo-client/src/main.rs
```diff
@@ -671,6 +877,20 @@ fn client(
...
+            let mut h = UploadHandler {
+                streams: HashMap::new(),
+                upload_stats: HashMap::new(),
+                url_queue: VecDeque::from(urls.to_vec()),
+                args,
+                chunk_size: 32768,
+            };
+            process_loop(&local_addr, socket, &mut client, &mut h)?;
+            return Ok(None);
+        }
```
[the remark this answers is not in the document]
KershawChang: Will do.

neqo-client/src/main.rs
KershawChang: Pleas see [here](https://github.com/mozilla/neqo/blob/main/neqo-client/src/main.rs#L621).
This is actually the original code, but I'll think about this comment.

neqo-client/src/main.rs
KershawChang: Yeah, I'll try.

neqo-server/src/main.rs
```diff
@@ -479,6 +481,17 @@ impl HttpServer for SimpleServer {
                 } => {
                     println!("Headers (request={} fin={}): {:?}", stream, fin, headers);
 
+                    let post = if let Some(method) = headers.iter().find(|&h| h.name() == ":method")
+                    {
+                        method.value() == "POST"
+                    } else {
+                        false
+                    };
```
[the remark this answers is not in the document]
KershawChang: I didn't address this comment because bumping to Rust 1.70 is outside of this PR.

martinthomson (approved):
I'll leave the next round of refactoring up to your discretion.

neqo-server/src/main.rs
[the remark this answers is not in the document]
larseggert: We should do that bump. because some other things (like `nix`) depend on it.

KershawChang:
Given those CI failures, I cam not sure if I can merge this PR.
@martinthomson , what do you think?

martinthomson:
I'm OK with landing this one.  The errors are unrelated.

We should suppress warnings in ech.rs, fix #1519, and work out what is going on with DYLD_LIBRARY_PATH on mac CI.  But that doesn't need to block this work.

Issue #528, opened 2020-04-07
Meta: Implement all qlog probes


agrover:
Once basic qlog support lands in Neqo (#526), we still need to add probes for all the things qlog can log. These should be *relatively* straightforward, and can likely be done as a group effort.

The below list is all event types supported by the `qlog` crate, version 0.2. Future versions of the qlog specification and the qlog crate may change this list. `qlog` supports "min" versions of many event constructors, which contain only the most commonly relevant fields of the event. Either version may be used.

- [ ] congestion_state_updated(_min)
- [ ] connection_id_updated(_min)
- [x] connection_started(_min)
- [ ] connection_state_updated(_min)
- [ ] datagram_dropped(_min)
- [ ] datagrams_received(_min)
- [ ] datagrams_sent(_min)
- [ ] frames_processed
- [ ] h3_data_moved(_min)
- [ ] h3_frame_created(_min)
- [ ] h3_frame_parsed(_min)
- [ ] h3_parameters_set(_min)
- [ ] h3_push_resolved(_min)
- [ ] h3_stream_type_set(_min)
- [ ] loss_timer_set(_min)
- [ ] marked_for_retransmit
- [ ] metrics_updated(_min)
- [ ] packet_buffered
- [ ] packet_dropped(_min)
- [ ] packet_lost(_min)
- [x] packet_received(_min)
- [x] packet_sent(_min)
- [ ] qpack_dynamic_table_updated
- [ ] qpack_headers_decoded(_min)
- [ ] qpack_headers_encoded(_min)
- [x] qpack_instruction_received(_min) (partial)
- [ ] qpack_instruction_sent(_min)
- [ ] qpack_state_updated(_min)
- [ ] qpack_stream_state_updated
- [ ] recovery_parameters_set(_min)
- [ ] server_listening(_min)
- [ ] ~~spinbit_updated~~ (Neqo does not implement spin bit)
- [ ] stream_state_updated(_min)
- [x] transport_parameters_set(_min)

agrover:
Nice to have but implementing all of this is not critical.

mansf-osk:
Hey, I could help here! Are there any [events in the currently used qlog version](https://docs.rs/qlog/0.13.0/qlog/events/index.html#modules) that would be especially useful to implement?

larseggert:
Thanks for helping out! I think we first need to check if the list in the first comment is still accurate.

mansf-osk:
Makes sense, I've compiled a new list below

[...]

PR #1505, opened 2023-12-10
feat(qlog): log version_information on client

mxinden:
:wave: Hi there

This is Max, first time contributor.

Let me know if this tiny pull request is of some help. Would appreciate a review and pointers to what to tackle next.

Thank you for QUIC support in Firefox!

---

This commit adds support for the qlog [`version_information` QUIC event](https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-version_information) on the client.

Depends on https://github.com/mozilla/neqo/pull/1504

Depends on https://github.com/cloudflare/quiche/pull/1684

Meta issue: https://github.com/mozilla/neqo/issues/528

This commit does not add server-side support. I was unsure where to add the `qlog::xxx` call. With my limited knowledge of the code base, I would place it in the [`TransportParametersHandler`](https://github.com/mozilla/neqo/blob/83735a88217a6b3a6a9d3cd5d9243040c5e41319/neqo-transport/src/tparams.rs#L633-L643), though as of today, it does not have access to the connection's `NeqoQlog`. Thoughts?


Cargo.toml
```diff
@@ -10,3 +10,6 @@ members = [
   "neqo-interop",
   "test-fixture",
 ]
+
+[patch.crates-io]
+qlog = { git = "https://github.com/mxinden/quiche", branch = "type-to-importance" }
```
mxinden, line 15: To be removed once https://github.com/cloudflare/quiche/pull/1684 is merged and released.

martinthomson: We'll want to wait for that before we merge here.  For future reference, a draft PR is probably better for cases where you need to wait for upstream changes.  That might delay review, but we're generally happy to provide feedback on direction from a draft.

neqo-transport/src/qlog.rs
```diff
@@ -127,6 +128,44 @@ pub fn connection_state_updated(qlog: &mut NeqoQlog, new: &State) {
...
+}
+
+pub fn client_version_information_negotiated(
+    qlog: &mut NeqoQlog,
+    client: &[Version],
+    server: &[WireVersion],
+    chosen: Version,
+) {
+    qlog.add_event_data(|| {
+        let ev_data = EventData::VersionInformation(VersionInformation {
```
martinthomson, line 155: You can inline `ev_data` (it's less code overall).

martinthomson (approved):
For the server, look in `server.rs` for `fn accept_connection(...)`.  That is where a new connection is handled.  There shouldn't be a negotiated equivalent on the server, but the recording of failed version negotiation can be recorded.

For VN, you can make a call in `fn process_input(...)` right where it calls `PacketBuilder::version_negotiation(...)`, which is where the Version Negotiation packet is sent.  For that, you'll need to create a new QLog instance with `self.create_qlog_trace(attempt_key)`.  You will have to follow other code for that.  I probably wouldn't bother trying to reduce the number of logging events from the same remote address and DCID, because that just creates a cleanup hazard.

That won't catch all cases, but enough.

Cargo.toml
larseggert: https://github.com/cloudflare/quiche/pull/1684 is merged.

larseggert: @mxinden could you update the PR to not require the patched crate?

neqo-transport/src/qlog.rs
mxinden: Sounds good. Done in https://github.com/mozilla/neqo/pull/1505/commits/a5f7681671085ab0fa3f65bf8eb94927e6c44f4f.

Cargo.toml
mxinden: I am sorry for the delay here @larseggert.

With https://github.com/cloudflare/quiche/pull/1684 merged, I updated from `github.com/mxinden` to `github.com/cloudflare`. See https://github.com/mozilla/neqo/pull/1505/commits/5884b456e10c5e1f6a52e89af8c0fcacc86f07dd. I am assuming that we still want to block merging here until the patch in https://github.com/cloudflare/quiche/pull/1684 has been released to crates.io. See also https://crates.io/crates/qlog/versions. Let me know in case I am missing something.

larseggert: Yes, thanks for catching that. (I had assumed quiche would be uploaded to crates.io after the merge, but I guess not yet.)

mxinden: Patch section removed with https://github.com/mozilla/neqo/pull/1505/commits/4754645b1c6192fdf6a5d17a927acae2da18c17a with upgrade to `qlog` `v0.11.0` in https://github.com/mozilla/neqo/pull/1547. Resolving here.

mxinden:
With https://github.com/mozilla/neqo/pull/1547 merged, all outstanding comments are addressed. @martinthomson or @larseggert could either of you take another look?

martinthomson:
I opened #1549 for the server side.  I was waiting to see what you planned there, that's all.

mxinden:
> I opened #1549 for the server side. I was waiting to see what you planned there, that's all.

Sorry, badly communicated from my end. I prepared the server side in https://github.com/mozilla/neqo/pull/1531.

PR #1515, opened 2023-12-14
Implement the final RFC9369 version of QUIC v2

larseggert:
Wireshark can decode our test traffic. Leaving as a draft PR until I can test against someone else's public server.


larseggert:
Any idea for how to re-generate the binary samples that are part of the (failing) tests? Is there a script or do I simply copy in the values from the test failure output?

martinthomson:
I generally double-check the encodings.  For things that involve crypto, you should be able to use https://github.com/quicwg/base-drafts/pull/4997

larseggert:
Interops with www.haproxy.org.

martinthomson (approved):

PR #1512, opened 2023-12-12
Fix clippy issues on MacOS and Windows

larseggert:
I.e. the ones found via the new expanded CI matrix.

Leaving this as a draft PR until CI is green.

PR #1517, opened 2023-12-14
Fix hrtimer tests on MacOS

larseggert:
Whatever I did to this file in #1512 made the tests on MacOS fail (because timer accuracy got bad.) Restoring the code and adding clippy ignores fixes this.

### DIFF
--- a/neqo-common/src/hrtime.rs
+++ b/neqo-common/src/hrtime.rs
@@ -119,14 +119,13 @@ mod mac {
     #[repr(C)]
     #[derive(Debug, Copy, Clone, Default)]
     pub struct thread_time_constraint_policy {
-        period: f64,
-        computation: f64,
-        constraint: f64,
+        period: u32,
+        computation: u32,
+        constraint: u32,
         preemptible: boolean_t,
     }
 
     const THREAD_TIME_CONSTRAINT_POLICY: thread_policy_flavor_t = 2;
-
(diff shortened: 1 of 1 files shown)


larseggert:
Failures still happen in CI it seems, but at least locally the tests now always pass (instead of always failing.)

We do need to do something to not have CI go red for MacOS and Windows if these failures are due to the CI environment and not bugs.

neqo-common/src/hrtime.rs
```diff
@@ -119,14 +119,13 @@ mod mac {
     #[repr(C)]
     #[derive(Debug, Copy, Clone, Default)]
     pub struct thread_time_constraint_policy {
-        period: f64,
-        computation: f64,
-        constraint: f64,
+        period: u32,
+        computation: u32,
+        constraint: u32,
```
martinthomson, line 124: It will be this bit.  Sorry I missed it.  This is passed to a C interface, so we can't change the types.

martinthomson (approved):

martinthomson:
So, as far as CI failures go, we can either iterate more or loosen up the bounds.  I chose those somewhat arbitrarily, but realize that CI machines are probably doing other work that might delay timers at that sort of tight resolution.  Maybe run the timers N times and ensure that the values are consistently under X, where X is generous and sometimes under Y where Y is less generous.

PR #1520, opened 2023-12-18
Add beta toolchain to CI

larseggert:
And then let's see how to make it warn-only.


martinthomson (approved):
An early warning system is good, but we'll still be surprised by failures (esp. clippy).  I don't see a good solution though.

larseggert:
Yeah, I looked into downgrading all errors to warnings with `clippy` and I couldn't find a way.

PR #1522, opened 2023-12-19
add support to log rtt

KershawChang:
Maybe it's worth to also log RTT.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -167,7 +167,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         let mut new_acked = 0;
         for pkt in acked_pkts {
             qinfo!(
-                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}, rtts={:?}",
+                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}, rtt_est={:?}",
                 self,
                 pkt.pn,
                 pkt.size,
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -162,7 +162,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
     }
 
     // Multi-packet version of OnPacketAckedCC
-    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtts: &RttEstimate, now: Instant) {
+    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtt_est: &RttEstimate, now: Instant) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
         for pkt in acked_pkts {
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -173,7 +173,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
                 pkt.size,
                 i32::from(!pkt.cc_outstanding()),
                 i32::from(pkt.lost()),
-                rtts,
+                rtt_est,
             );
             if !pkt.cc_outstanding() {
                 continue;
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -224,7 +224,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             let bytes_for_increase = self.cc_algorithm.bytes_for_cwnd_increase(
                 self.congestion_window,
                 new_acked,
-                rtts.minimum(),
+                rtt_est.minimum(),
                 now,
             );
             debug_assert!(bytes_for_increase > 0);
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -19,6 +19,7 @@ use crate::{
     cc::MAX_DATAGRAM_SIZE,
     packet::PacketNumber,
     qlog::{self, QlogMetric},
+    rtt::RttEstimate,
     sender::PACING_BURST_SIZE,
     tracking::SentPacket,
 };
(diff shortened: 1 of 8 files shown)


martinthomson (commented):
Logging on packet send rather than when we update the value (which is when we receive an ACK) seems like we'll log more often than is necessary.

KershawChang:
> Logging on packet send rather than when we update the value (which is when we receive an ACK) seems like we'll log more often than is necessary.

Agreed. We should only log when it's changed.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -165,12 +165,13 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
             qinfo!(
-                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}",
+                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={} min_rtt={:?}",
                 self,
                 pkt.pn,
                 pkt.size,
                 i32::from(!pkt.cc_outstanding()),
-                i32::from(pkt.lost())
+                i32::from(pkt.lost()),
+                min_rtt
```
larseggert, line 174: Why are we logging `min_rtt` here, and `rtt` below in `on_packet_sent`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -315,7 +316,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         );
     }
 
-    fn on_packet_sent(&mut self, pkt: &SentPacket) {
+    fn on_packet_sent(&mut self, pkt: &SentPacket, rtt: Duration) {
```
larseggert, line 319: Is there a better way to get at the RTT information other than passing it around the stack?

martinthomson: Not without doing something gross with `Rc`.  That's a consequence of rust being rust.

KershawChang:
> Logging on packet send rather than when we update the value (which is when we receive an ACK) seems like we'll log more often than is necessary.

I've decided to log rtts in `on_packets_acked`. The main reason is that there could be multiple `ClassicCongestionControl` objects in a log and we need to link rtt information and the other information in `ClassicCongestionControl` together.
Please take another look. Thanks.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -161,17 +161,27 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
     }
 
     // Multi-packet version of OnPacketAckedCC
-    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], min_rtt: Duration, now: Instant) {
+    fn on_packets_acked(
+        &mut self,
+        acked_pkts: &[SentPacket],
+        min_rtt: Duration,
+        latest_rtt: Duration,
+        smooth_rtt: Duration,
```
martinthomson, line 169: Any reason why this can't be passed `&RttEstimate`?  The function is called from `Path::on_packets_acked()`, so I think that is much easier.  Then your logging calls might just invoke the `Debug` implementation on `RttEstimate` (which might need to be improved, as it uses the default).

martinthomson: Your tests can probably build a `const RTT_ESTIMATE`.

martinthomson (dismissed):
A simplification might be in order, but this is OK otherwise.

neqo-transport/src/cc/classic_cc.rs
KershawChang: Thanks. This is neat. 
I'll do this.

martinthomson (dismissed):

martinthomson:
@larseggert, do you want to review this?  If not, would you mind merging it?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -161,17 +162,18 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
     }
 
     // Multi-packet version of OnPacketAckedCC
-    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], min_rtt: Duration, now: Instant) {
+    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtts: &RttEstimate, now: Instant) {
```
larseggert, line 165:
```suggestion
    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtt_est: &RttEstimate, now: Instant) {
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -161,17 +162,18 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         let mut is_app_limited = true;
         let mut new_acked = 0;
         for pkt in acked_pkts {
             qinfo!(
-                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}",
+                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}, rtts={:?}",
```
larseggert, line 170:
```suggestion
                "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}, rtt_est={:?}",
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -161,17 +162,18 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
                 self,
                 pkt.pn,
                 pkt.size,
                 i32::from(!pkt.cc_outstanding()),
-                i32::from(pkt.lost())
+                i32::from(pkt.lost()),
+                rtts,
```
larseggert, line 176:
```suggestion
                rtt_est,
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -222,7 +224,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             let bytes_for_increase = self.cc_algorithm.bytes_for_cwnd_increase(
                 self.congestion_window,
                 new_acked,
-                min_rtt,
+                rtts.minimum(),
```
larseggert, line 227:
```suggestion
                rtt_est.minimum(),
```

neqo-transport/src/cc/mod.rs
```diff
@@ -42,7 +42,7 @@ pub trait CongestionControl: Display + Debug {
     #[must_use]
     fn cwnd_avail(&self) -> usize;
 
-    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], min_rtt: Duration, now: Instant);
+    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtts: &RttEstimate, now: Instant);
```
larseggert, line 45:
```suggestion
    fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], rtt_est: &RttEstimate, now: Instant);
```

neqo-transport/src/sender.rs
```diff
@@ -68,8 +69,13 @@ impl PacketSender {
         self.cc.cwnd_avail()
     }
 
-    pub fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], min_rtt: Duration, now: Instant) {
-        self.cc.on_packets_acked(acked_pkts, min_rtt, now);
+    pub fn on_packets_acked(
+        &mut self,
+        acked_pkts: &[SentPacket],
+        rtts: &RttEstimate,
```
larseggert, line 75:
```suggestion
        rtt_est: &RttEstimate,
```

neqo-transport/src/sender.rs
```diff
@@ -68,8 +69,13 @@ impl PacketSender {
...
+        now: Instant,
+    ) {
+        self.cc.on_packets_acked(acked_pkts, rtts, now);
```
larseggert, line 78:
```suggestion
        self.cc.on_packets_acked(acked_pkts, rtt_est, now);
```

larseggert (dismissed):
LGTM, but suggest renaming `rtts` -> `rtt_est` for clarity.

KershawChang:
I assume I can just merge this PR, since the change after last review was addressing some nits.

PR #1523, opened 2023-12-19
Fix hrtimer CI

larseggert:
Or try to.


martinthomson (commented):
So the theory is that over 30 runs we'll get something closer to 15 of those below the cutoff.  Worth a try.  We could also increase the cutoff.

Note that this is one of the few tests that runs in real time in our entire suite.  Running in CI this is fine, but `30\Sum_1^9{Fibonnaci}` is a large number.

larseggert:
@martinthomson Wonder if we should just look at `min`? Or do different things on different platforms? Or disable the test in CI?

martinthomson:
The minimum won't really tell us much (that could be chance).  Maybe we just need to disable in CI.

larseggert:
OK, ignored those tests in CI on MacOS and Windows.

There are some clippy failures on Windows with Rust 1.65.0. Another reason we should bump to 1.70.0 :-)

larseggert:
Someone let me know if the clippy failures on Windows with Rust 1.70.0 should be suppressed? Or can they be fixed otherwise?

KershawChang (approved):

Issue #1521, opened 2023-12-19
macos CI seems to be building the wrong feature profile.


martinthomson:
https://github.com/mozilla/neqo/actions/runs/7257208473/job/19770667986?pr=1502#step:13:52 says:

```
warning: neqo-crypto v0.6.8 (/Users/runner/work/neqo/neqo/neqo-crypto) ignoring invalid dependency `mozbuild` which is missing a lib target
```

That suggests that we are looking at an error that is the result of passing the wrong features to `build.rs`.  Maybe we should avoid ~~`--all-features`~~ `--all-targets` [updated] when it comes to this repository.

larseggert:
Did you mean `--all-targets`? (Because I don't see `--all-features` in the workflow file.)

martinthomson:
Sorry, yes, that. :)

larseggert:
I'm only getting this warning with `nextest`, not `test`. Hm.

PR #1524, opened 2023-12-19
Replace nextest with test

larseggert:
To see if this addresses #1521


larseggert:
@martinthomson this seems to remove the warning you saw. (But I don't understand why that happened in the first place, since the `mozbuild` dependency is gated by `gecko`, which is not set...)

martinthomson (approved):
Is this temporary?  Because you should avoid installing nextest if this is the change you want.

larseggert:
I'll remove it - I wanted to first have CI check that the error is actually gone.

PR #1529, opened 2023-12-20
Add some instructions for doing upload test locally

KershawChang:


performance-test/README.md
```diff
@@ -0,0 +1,23 @@
...
+RUST_LOG=neqo_transport=info ./target/release/neqo-client http://127.0.0.1:4433/ --test upload --upload-size 8388608 &>upload.log
+```
+
+## To visualize the log
+
+We have a simple Python script `neqo_cwnd.py` that can be used to visualize the log.
+To run this script, ensure you have `matplotlib` installed.
+```
+python3 neqo_cwnd.py upload.log
+```
```
larseggert, line 23: It would be nice if this were a shell script, rather than instructions.

performance-test/README.md
```diff
@@ -0,0 +1,23 @@
...
+1. Start neqo-client and specify parameters to start the upload test.
+   ` ./target/release/neqo-client http://127.0.0.1:4433/  --test upload  --upload-size ${size_in_bytes}`
+
+## To enable log messages for analyzing upload performance
+
+This can be done by setting the `RUST_LOG` environment variable to `neqo_transport=info`.
+For example, the command below starts neqo-client and uploads 8MB of content to the server.
+```
+RUST_LOG=neqo_transport=info ./target/release/neqo-client http://127.0.0.1:4433/ --test upload --upload-size 8388608 &>upload.log
+```
```
larseggert, line 15: What is the performance impact of enabling logging? Also, is a debug build OK here or should it be an optimized release build?

larseggert: Ideally, we'd want to take a packet trace and SSLKEYLOG during a performance run, decrypt the trace via Wireshark, conveft it to - say - JSON, and extract the information to plot out of *that*, rather than slowing down the endpoints by logging there.

larseggert:
Also wonder if we want to call the directory simply `test`, since we might want to add other (e.g., conformance) tests and not just performance tests.

performance-test/README.md
KershawChang: Sure, I can also submit a shell script for this.

performance-test/README.md
larseggert: Wireshark will show you what on the wire :-) What the logs show depend on in which order neqo pushes and pulls data from the sockets, which depends on its internal scheduling logic.

larseggert (approved):

PR #1531, opened 2023-12-27
feat(qlog): log version_information on server

mxinden:
This commit adds support for the qlog [`version_information` QUIC event](https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-version_information) on the server.

See https://github.com/mozilla/neqo/pull/1505 for the client side.

~~Depends on `qlog` patch release with https://github.com/cloudflare/quiche/pull/1684~~ :heavy_check_mark: 

Meta issue: https://github.com/mozilla/neqo/issues/528

Fixes https://github.com/mozilla/neqo/issues/1549

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,6 @@ members = [
   "neqo-interop",
   "test-fixture",
 ]
+
+[patch.crates-io]
+qlog = { git = "https://github.com/cloudflare/quiche", branch = "master" }
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -383,11 +383,11 @@ impl Server {
         }
     }
 
-    fn create_qlog_trace(&self, attempt_key: &AttemptKey) -> NeqoQlog {
+    fn create_qlog_trace(&self, odcid: &ConnectionIdRef<'_>) -> NeqoQlog {
         if let Some(qlog_dir) = &self.qlog_dir {
             let mut qlog_path = qlog_dir.to_path_buf();
 
-            qlog_path.push(format!("{}.qlog", attempt_key.odcid));
+            qlog_path.push(format!("{}.qlog", odcid));
 
             // The original DCID is chosen by the client. Using create_new()
             // prevents attackers from overwriting existing logs.
@@ -449,7 +449,7 @@ impl Server {
             c.set_retry_cids(odcid, initial.src_cid, initial.dst_cid);
         }
         c.set_validation(Rc::clone(&self.address_validation));
-        c.set_qlog(self.create_qlog_trace(attempt_key));
+        c.set_qlog(self.create_qlog_trace(&attempt_key.odcid.as_cid_ref()));
         if let Some(cfg) = &self.ech_config {
             if c.server_enable_ech(cfg.config, &cfg.public_name, &cfg.sk, &cfg.pk)
                 .is_err()
@@ -504,7 +504,7 @@ impl Server {
                 qwarn!([self], "Unable to create connection");
                 if e == crate::Error::VersionNegotiation {
                     crate::qlog::server_version_information_failed(
-                        &mut self.create_qlog_trace(&attempt_key),
+                        &mut self.create_qlog_trace(&attempt_key.odcid.as_cid_ref()),
                         self.conn_params.get_versions().all(),
                         initial.version.wire_version(),
                     )
@@ -585,10 +585,7 @@ impl Server {
             );
 
             crate::qlog::server_version_information_failed(
-                &mut self.create_qlog_trace(&AttemptKey {
-                    remote_address: dgram.source(),
-                    odcid: packet.dcid().into(),
-                }),
+                &mut self.create_qlog_trace(packet.dcid()),
                 self.conn_params.get_versions().all(),
                 packet.wire_version(),
             );
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -489,20 +489,30 @@ impl Server {
             params,
         );
 
-        if let Ok(mut c) = sconn {
-            self.setup_connection(&mut c, &attempt_key, initial, orig_dcid);
-            let c = Rc::new(RefCell::new(ServerConnectionState {
-                c,
-                last_timer: now,
-                active_attempt: Some(attempt_key.clone()),
-            }));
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -10,3 +10,6 @@ members = [
   "neqo-interop",
   "test-fixture",
 ]
+
+[patch.crates-io]
+qlog = { git = "https://github.com/cloudflare/quiche", branch = "master" }
```
mxinden, line 15: I will remove the patch **before** merge once `qlog` `v0.10.1` is released. See also https://github.com/mozilla/neqo/pull/1505#discussion_r1421777643.

neqo-transport/src/server.rs
```diff
@@ -578,6 +588,18 @@ impl Server {
...
+
+            crate::qlog::server_version_information_failed(
+                &mut self.create_qlog_trace(&AttemptKey {
+                    remote_address: dgram.source(),
+                    odcid: packet.dcid().into(),
+                }),
+                self.conn_params.get_versions().all(),
+                packet
+                    .version()
+                    .expect("version on `OtherVersion` and `Initial` to be set"),
```
mxinden, line 600: Same as `unwrap` in line 567 / 573. Let me know if you would prefer making this state impossible at compile time instead.

neqo-transport/src/server.rs
```diff
@@ -578,6 +588,18 @@ impl Server {
                 packet.wire_version(),
                 self.conn_params.get_versions().all(),
             );
+
+            crate::qlog::server_version_information_failed(
+                &mut self.create_qlog_trace(&AttemptKey {
+                    remote_address: dgram.source(),
```
mxinden, line 594: Note that `remote_address` is not used by `self.create_qlog_trace`. Want me to change the signature of `self.create_qlog_trace` to only take the `odcid`?

mxinden (commented):
Following up on Martin's comment in https://github.com/mozilla/neqo/pull/1505#pullrequestreview-1774142992:

> For the server, look in `server.rs` for `fn accept_connection(...)`.

Done in this pull request. Let me know if the direction is correct.

> There shouldn't be a negotiated equivalent on the server

Do I understand correctly that you don't want to log the happy path, i.e. the case where version negotiation succeeds on the server? Fine by me.

> For VN, you can make a call in `fn process_input(...)` right where it calls `PacketBuilder::version_negotiation(...)`, which is where the Version Negotiation packet is sent.

Done in this pull request. Let me know if the direction is correct.

neqo-transport/src/server.rs
mxinden: :facepalm: not the same. Will fix.

neqo-transport/src/server.rs
```diff
@@ -573,6 +583,16 @@ impl Server {
...
+                    odcid: packet.dcid().into(),
+                }),
+                self.conn_params.get_versions().all(),
+                packet.wire_version(),
```
mxinden, line 593: `packet.wire_version()` expects `packet.version` to be `Some`.

https://github.com/mozilla/neqo/blob/abf2636153d5c88feee5faf5663facc79defc858/neqo-transport/src/packet/mod.rs#L694-L697

Based on `PublicPacket::decode` this is valid for `PacketType::OtherVersion` and `PacketType::Initial`.

https://github.com/mozilla/neqo/blob/abf2636153d5c88feee5faf5663facc79defc858/neqo-transport/src/packet/mod.rs#L553-L642

Suggestions for a way to obtain the version without this implicit assumption? I am afraid it might turn into a footgun with future refactorings.

mxinden:
@martinthomson can you give this pull request a first review? Is this the direction you intended in https://github.com/mozilla/neqo/pull/1505#pullrequestreview-1774142992?

neqo-transport/src/server.rs
martinthomson: That's worth doing.  We might have needed an address once upon a time.  Or maybe all the other instances where a trace was needed involved an address.  Now that we have a case where  we make a useless `AttemptKey`, it seems worthwhile changing the signature to eliminate that extra work (and have the other call sites use `AttemptKey::odcid()`.

neqo-transport/src/server.rs
martinthomson: I'm not seeing a problem here.  We will have `Some` version in this case, so the assertion should pass.  The only case where we have a `None` version is `VersionNegotiation`, which is inherently not a version.  The effect is only felt if people get a wire version from `VersionNegotiation`, which they shouldn't do.  And the true effect, if assertions are disabled, is that you get version 0, which is correct.

martinthomson (approved):
Once you get the refactoring sorted, I'll merge this. Sorry for missing it (I blame my vacation).

mxinden (commented):
Thanks for the review. Mind taking another look @martinthomson?

neqo-transport/src/server.rs
```diff
@@ -383,11 +383,11 @@ impl Server {
         }
     }
 
-    fn create_qlog_trace(&self, attempt_key: &AttemptKey) -> NeqoQlog {
+    fn create_qlog_trace(&self, odcid: &ConnectionIdRef<'_>) -> NeqoQlog {
```
mxinden, line 386: `PublicPacket::dcid` returns a `&ConnectionIdRef<'a>`:

https://github.com/mozilla/neqo/blob/9e6cd932b140fde141e6414d53f539b6d296d4f9/neqo-transport/src/packet/mod.rs#L676-L678

Thus I opted for `create_qlog_trace` to take a `&ConnectionIdRef` instead of a `&ConnectionId` since the former can be derived from the latter for free.

Let me know if that is fine by you, or whether you consider this a premature optimization not worth the noise.

(I am a bit confused why `PublicPacket::dcid` returns a reference to a reference (i.e. `&` and `Ref<'_>`) to some data. But that might be out of scope for this pull request.)

neqo-transport/src/server.rs
mxinden: :+1: Continuing discussion in https://github.com/mozilla/neqo/pull/1531#pullrequestreview-1823048224.

neqo-transport/src/server.rs
martinthomson: Oh, the ref to a ref is probably just to avoid putting `Copy` on that struct.  Which is silly because it is just a reference itself.

martinthomson (approved):

PR #1532, opened 2023-12-29
chore: turn on clippy::pedantic

mxinden:
Enable clippy's _pedantic_ lint category.

- Adjusts CI check accordingly.
- Adds `#![deny(clippy::pedantic)]` to all (but tests) top level crate files.
- Fixes various lint failures.
- Removes module `#![deny(clippy::pedantic)]` whereever it is already enabled at the crate level.

Fixes https://github.com/mozilla/neqo/issues/553

More work needed here. _Draft_ for now.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -6,6 +6,7 @@
 
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![warn(clippy::use_self)]
+#![deny(clippy::pedantic)]
 
 use qlog::{events::EventImportance, streamer::QlogStreamer};
 
--- a/neqo-transport/src/addr_valid.rs
+++ b/neqo-transport/src/addr_valid.rs
@@ -23,15 +23,15 @@ use std::convert::TryFrom;
 use std::net::{IpAddr, SocketAddr};
 use std::time::{Duration, Instant};
 
-/// A prefix we add to Retry tokens to distinguish them from NEW_TOKEN tokens.
+/// A prefix we add to Retry tokens to distinguish them from `NEW_TOKEN` tokens.
 const TOKEN_IDENTIFIER_RETRY: &[u8] = &[0x52, 0x65, 0x74, 0x72, 0x79];
-/// A prefix on NEW_TOKEN tokens, that is maximally Hamming distant from NEW_TOKEN.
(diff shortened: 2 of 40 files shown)


neqo-client/src/main.rs
```diff
@@ -6,6 +6,7 @@
 
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![warn(clippy::use_self)]
+#![deny(clippy::pedantic)]
```
martinthomson, line 9: In other projects, we use `warn(clippy::pedantic)` in code, but then execute pre-commit hooks and CI with `-D warnings`.  I think that is a better approach overall than having this in code.

neqo-transport/src/addr_valid.rs
```diff
@@ -234,16 +234,16 @@ impl AddressValidation {
                     panic!("AddressValidation: Retry token with small CID {}", cid);
                 }
             } else if cid.is_empty() {
-                // An empty connection ID means NEW_TOKEN.
+                // An empty connection ID means `NEW_TOKEN`.
                 if self.validation == ValidateAddress::Always {
-                    qinfo!("AddressValidation: valid NEW_TOKEN token; validating again");
+                    qinfo!("AddressValidation: valid `NEW_TOKEN` token; validating again");
```
martinthomson, line 239: I don't think that we need these to be quoted like this.

neqo-transport/src/lib.rs
```diff
@@ -6,6 +6,18 @@
...
+#![allow(clippy::must_use_candidate)]
+#![allow(clippy::return_self_not_must_use)]
+#![allow(clippy::module_name_repetitions)]
+#![allow(clippy::missing_errors_doc)]
+#![allow(clippy::needless_pass_by_value)]
+#![allow(clippy::doc_markdown)]
+#![allow(clippy::too_many_lines)]
+#![allow(clippy::single_match_else)]
+#![allow(clippy::if_not_else)]
+#![allow(clippy::match_wild_err_arm)]
```
martinthomson, line 20: These are fairly broad and I'm not sure that I like a blanket allow on each of them.  (There are some that might be justified, for sure, but not all of these.)

martinthomson (commented):
Making good progress. :)

mxinden:
Thank you for the review!

Closing here. I will create individual pull requests (see e.g. linked ones above), to ease review and to reduce merge conflicts. Still lots to do.

PR #1533, opened 2023-12-30
fix(transport/qlog): map State::Closing onto ConnectionState::Closing

mxinden:
Map `State::Closing` to `ConnectionState::Closing` instead of `ConnectionState::Draining`.

Assuming that this is a typo.

CI Clippy failures are unrelated. See https://github.com/mozilla/neqo/pull/1534 for a fix.


martinthomson (approved):
This is a timely reminder that we don't have tests for qlog logging.

Issue #1538, opened 2024-01-05
Accept Datagrams by reference, not value


martinthomson:
It looks like `process()` and friends (`process_input()`, `process_path()`, etc..) will work perfectly well with datagram references, so maybe we could change to take those.

For compatibility, we might instead make these take `impl AsRef<Datagram>` instead so that we don't need to change existing uses of the API.

For the multiple inputs, we might change from `Vec<Datagram>` to something like `impl IntoIterator<Item = impl AsRef<Datagram>>`.

mxinden:
> It looks like `process()` and friends (`process_input()`, `process_path()`, etc..) will work perfectly well with datagram references, so maybe we could change to take those.

Agreed that this would be the better API, since most of `process_*` only needs a reference. That said, `save_datagram` requires an owned `Datagram` in order to store it in `Connection`'s  `saved_datagrams`.

https://github.com/mozilla/neqo/blob/28b5c9a669c5f028b425bf96fcb02374c88d454e/neqo-transport/src/connection/mod.rs#L1142-L1152

https://github.com/mozilla/neqo/blob/28b5c9a669c5f028b425bf96fcb02374c88d454e/neqo-transport/src/connection/mod.rs#L251-L255

Is `save_datagram` i.e. partial processing of a datagram, enough of an edge case to warrant a `clone()`?

> For compatibility, we might instead make these take `impl AsRef<Datagram>` instead so that we don't need to change existing uses of the API.

Is `impl AsRef<Datagram>` over `&Datagram` worth it? I see the following problems:

- Adds complexity to the function signatures.
- `AsRef` is not an auto trait, thus requires an additional trait implementation:
    ``` diff
        modified   neqo-common/src/datagram.rs
    @@ -16,6 +16,12 @@ pub struct Datagram {
         d: Vec<u8>,
     }
     
    +impl AsRef<Datagram> for Datagram {
    +    fn as_ref(&self) -> &Self {
    +        self
    +    }
    +}
    +
     impl Datagram {
    ```
- Prevents rustc from inferring the type when providing `None`
    ```
    error[E0283]: type annotations needed
        --> neqo-transport/src/connection/tests/handshake.rs:845:30
         |
    845  |     let out = client.process(None, now());
         |                      ------- ^^^^ cannot infer type of the type parameter `T` declared on the enum `Option`
         |                      |
         |                      required by a bound introduced by this call
         |
[...]

PR #1543, opened 2024-01-06
refactor: accept Datagrams by reference, not value

mxinden:
Fixes https://github.com/mozilla/neqo/issues/1538.

See discussion in https://github.com/mozilla/neqo/issues/1538#issuecomment-1879731931.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -432,7 +432,7 @@ fn process_loop(
             };
         }
         if !datagrams.is_empty() {
-            client.process_multiple_input(datagrams.iter(), Instant::now());
+            client.process_multiple_input(&datagrams, Instant::now());
             handler.maybe_key_update(client)?;
         }
 
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -432,7 +432,7 @@ fn process_loop(
             };
         }
         if !datagrams.is_empty() {
-            client.process_multiple_input(datagrams, Instant::now());
+            client.process_multiple_input(datagrams.iter(), Instant::now());
             handler.maybe_key_update(client)?;
         }
 
@@ -1364,7 +1364,7 @@ mod old {
                         }
                         if sz > 0 {
                             let d = Datagram::new(remote, *local_addr, &buf[..sz]);
-                            client.process_input(d, Instant::now());
+                            client.process_input(&d, Instant::now());
                             handler.maybe_key_update(client)?;
                         }
                     }
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -804,7 +804,7 @@ impl Http3Client {
     }
 
     /// This function combines  `process_input` and `process_output` function.
-    pub fn process(&mut self, dgram: Option<Datagram>, now: Instant) -> Output {
+    pub fn process(&mut self, dgram: Option<&Datagram>, now: Instant) -> Output {
         qtrace!([self], "Process.");
         if let Some(d) = dgram {
             self.process_input(d, now);
@@ -822,15 +822,20 @@ impl Http3Client {
     /// packets need to be sent or if a timer needs to be updated.
     ///
     /// [1]: ../neqo_transport/enum.ConnectionEvent.html
-    pub fn process_input(&mut self, dgram: Datagram, now: Instant) {
+    pub fn process_input(&mut self, dgram: &Datagram, now: Instant) {
         qtrace!([self], "Process input.");
         self.conn.process_input(dgram, now);
         self.process_http3(now);
     }
 
-    pub fn process_multiple_input(&mut self, dgrams: Vec<Datagram>, now: Instant) {
+    pub fn process_multiple_input<'a>(
+        &mut self,
+        // TODO: Need ExactSizeIterator for qtrace below. Worth the type restriction?
+        dgrams: impl ExactSizeIterator<Item = &'a Datagram>,
+        now: Instant,
+    ) {
         qtrace!([self], "Process multiple datagrams, len={}", dgrams.len());
-        if dgrams.is_empty() {
+        if dgrams.len() == 0 {
             return;
         }
         self.conn.process_multiple_input(dgrams, now);
@@ -1589,11 +1594,11 @@ mod tests {
         assert_eq!(client.state(), Http3State::Initializing);
 
         assert_eq!(*server.conn.state(), State::Init);
-        let out = server.conn.process(out.dgram(), now());
+        let out = server.conn.process(out.as_dgram_ref(), now());
         assert_eq!(*server.conn.state(), State::Handshaking);
 
-        let out = client.process(out.dgram(), now());
-        let out = server.conn.process(out.dgram(), now());
+        let out = client.process(out.as_dgram_ref(), now());
+        let out = server.conn.process(out.as_dgram_ref(), now());
         assert!(out.as_dgram_ref().is_none());
 
         let authentication_needed = |e| matches!(e, Http3ClientEvent::AuthenticationNeeded);
@@ -1606,12 +1611,14 @@ mod tests {
     fn connect_only_transport_with(client: &mut Http3Client, server: &mut TestServer) {
         let out = handshake_only(client, server);
 
-        let out = client.process(out.dgram(), now());
+        let out = client.process(out.as_dgram_ref(), now());
         let connected = |e| matches!(e, Http3ClientEvent::StateChange(Http3State::Connected));
         assert!(client.events().any(connected));
 
         assert_eq!(client.state(), Http3State::Connected);
-        server.conn.process_input(out.dgram().unwrap(), now());
+        server
+            .conn
+            .process_input(out.as_dgram_ref().unwrap(), now());
         assert!(server.conn.state().connected());
     }
 
@@ -1625,8 +1632,10 @@ mod tests {
 
     fn send_and_receive_client_settings(client: &mut Http3Client, server: &mut TestServer) {
         // send and receive client settings
-        let dgram = client.process(None, now()).dgram();
-        server.conn.process_input(dgram.unwrap(), now());
+        let out = client.process(None, now());
+        server
+            .conn
+            .process_input(out.as_dgram_ref().unwrap(), now());
         server.check_client_control_qpack_streams_no_resumption();
     }
 
@@ -1640,8 +1649,8 @@ mod tests {
 
         server.create_qpack_streams();
         // Send the server's control and qpack streams data.
-        let dgram = server.conn.process(None, now()).dgram();
-        client.process_input(dgram.unwrap(), now());
+        let out = server.conn.process(None, now());
+        client.process_input(out.as_dgram_ref().unwrap(), now());
 
         // assert no error occured.
         assert_eq!(client.state(), Http3State::Connected);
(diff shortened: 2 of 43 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1134,20 +1139,21 @@ impl Connection {
...
-        let d = if remaining < d.len() {
-            Datagram::new(d.source(), d.destination(), &d[d.len() - remaining..])
+    fn save_datagram(&mut self, cspace: CryptoSpace, d: &Datagram, remaining: usize, now: Instant) {
+        if remaining < d.len() {
+            let d = Datagram::new(d.source(), d.destination(), &d[d.len() - remaining..]);
+            self.saved_datagrams.save(cspace, d, now);
         } else {
-            d
+            // TODO: Sane?
+            self.saved_datagrams.save(cspace, d.clone(), now);
```
mxinden, line 1155: By having `process_*` take a `&Datagram` instead of `Datagram`, one requires a `clone()` here.

See discussion in https://github.com/mozilla/neqo/issues/1538#issuecomment-1879731931.

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
```diff
@@ -78,28 +78,28 @@ fn connect_with(client: &mut Http3Client, server: &mut Http3Server) {
     let out = client.process(None, now());
     assert_eq!(client.state(), Http3State::Initializing);
 
-    let out = server.process(out.dgram(), now());
```
martinthomson, line 81: It's possible that we could make `dgram()` return a reference as well now and remove the reference function.  Would that help?

martinthomson (commented):
As you have done the work for this one, I'm fairly heppy with how it turned out.  The `AsRef` idea isn't really necessary, I think.  I was mostly looking to save the cost of updating all the tests like this.

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
mxinden: It would reduce the diff of the PR. That said, `as_dgram_ref` is the more [Rust idiomatic name](https://rust-lang.github.io/api-guidelines/naming.html). Thus I suggest keeping the methods (`dgram`, `as_dgram_ref`) as is.

Not a strong opinion. Your call @martinthomson.

neqo-client/src/main.rs
```diff
@@ -432,7 +432,7 @@ fn process_loop(
             };
         }
         if !datagrams.is_empty() {
-            client.process_multiple_input(datagrams, Instant::now());
+            client.process_multiple_input(datagrams.iter(), Instant::now());
```
martinthomson, line 435: I think that this is preferable here.

```suggestion
            client.process_multiple_input(&datagrams, Instant::now());
```

neqo-http3/src/connection_client.rs
```diff
@@ -822,15 +822,20 @@ impl Http3Client {
...
         qtrace!([self], "Process input.");
         self.conn.process_input(dgram, now);
         self.process_http3(now);
     }
 
-    pub fn process_multiple_input(&mut self, dgrams: Vec<Datagram>, now: Instant) {
+    pub fn process_multiple_input<'a>(
+        &mut self,
+        // TODO: Need ExactSizeIterator for qtrace below. Worth the type restriction?
+        dgrams: impl ExactSizeIterator<Item = &'a Datagram>,
```
martinthomson, line 834: Hmm, I might prefer instead `I: IntoIterator, I::IntoIter: ExactSizeIterator, I::Item: &'a Datagram` in this case.  Then the caller can pass a slice.

neqo-transport/src/connection/mod.rs
```diff
@@ -917,15 +917,20 @@ impl Connection {
...
         self.process_saved(now);
         self.streams.cleanup_closed_streams();
     }
 
     /// Process new input datagrams on the connection.
-    pub fn process_multiple_input(&mut self, dgrams: Vec<Datagram>, now: Instant) {
-        if dgrams.is_empty() {
+    pub fn process_multiple_input<'a>(
+        &mut self,
+        dgrams: impl IntoIterator<Item = &'a Datagram>,
```
martinthomson, line 929: I don't have a problem with requiring ExactSizeIterator for this, especially to avoid having to wrap in a peekable.  See the previous comment.

neqo-transport/src/connection/mod.rs
martinthomson: This is fine.  This is rare enough.

neqo-transport/src/connection/mod.rs
```diff
@@ -1134,20 +1139,21 @@ impl Connection {
...
-    fn save_datagram(&mut self, cspace: CryptoSpace, d: Datagram, remaining: usize, now: Instant) {
-        let d = if remaining < d.len() {
-            Datagram::new(d.source(), d.destination(), &d[d.len() - remaining..])
+    fn save_datagram(&mut self, cspace: CryptoSpace, d: &Datagram, remaining: usize, now: Instant) {
+        if remaining < d.len() {
+            let d = Datagram::new(d.source(), d.destination(), &d[d.len() - remaining..]);
+            self.saved_datagrams.save(cspace, d, now);
         } else {
-            d
+            // TODO: Sane?
```
martinthomson, line 1154:
```suggestion
```

martinthomson (approved):
I'm OK with the `as_dgram_ref()` conclusion you reached.

neqo-client/src/main.rs
mxinden: See 3f76accc.

neqo-http3/src/connection_client.rs
mxinden: Good idea. Thank you @martinthomson. See 3f76accc.

neqo-transport/src/connection/mod.rs
mxinden: See 3f76accc.

mxinden:
Added two commits instead of updating the existing and force pushing in order to maintain review history. Let me know if you prefer a single commit instead.

martinthomson (approved):
@KershawChang because you will be responsible for integrating this into gecko, I wanted to highlight this.  Do you have any opinions on this change?

I think that it's a good one, but I don't want to merge until you are happy with it.

KershawChang:
> @KershawChang because you will be responsible for integrating this into gecko, I wanted to highlight this. Do you have any opinions on this change?
> 
> I think that it's a good one, but I don't want to merge until you are happy with it.

I think this change should be fine.
Looking at the code [here](https://searchfox.org/mozilla-central/rev/6f90f50b7a32cc062ab755e0653b3d3f512fe3bd/netwerk/socket/neqo_glue/src/lib.rs#322), I think it should be easy to pass the reference of datagram to `process_input`.

I'd be great if we can also avoid copying the data (`(*packet).to_vec()`), but I am not sure if this is possible.

mxinden:
> I'd be great if we can also avoid copying the data (`(*packet).to_vec()`), but I am not sure if this is possible.

How about the following:

- Have `Datagram` hold a slice instead of a `Vec`.
    ``` diff
    modified   neqo-common/src/datagram.rs
    @@ -10,10 +10,10 @@ use std::ops::Deref;
     use crate::hex_with_len;
     
     #[derive(PartialEq, Eq, Clone)]
    -pub struct Datagram {
    +pub struct Datagram<'a> {
         src: SocketAddr,
         dst: SocketAddr,
    -    d: Vec<u8>,
    +    d: &'a [u8],
     }
    ```
- Construct a `Datagram` via [`ThinVec::as_slice`](https://docs.rs/thin-vec/latest/thin_vec/struct.ThinVec.html#method.as_slice).
    ``` diff
    pub unsafe extern "C" fn neqo_http3conn_process_input(
        conn: &mut NeqoHttp3Conn,
        remote_addr: *const NetAddr,
        packet: *const ThinVec<u8>,
    ) -> nsresult {
        let remote = match netaddr_to_socket_addr(remote_addr) {
            Ok(addr) => addr,
            Err(result) => return result,
        };
        conn.conn.process_input(
    -       Datagram::new(remote, conn.local_addr, (*packet).to_vec()),
    +       Datagram::new(remote, conn.local_addr, (*packet).as_slice()),
            get_current_or_last_output_time(&conn.last_output_time),
        );
        return NS_OK;
    }
    ```

- Have `process_*` take an owned `Datagram` generic over the lifetime of the borrowed data.
    ``` diff
    modified   neqo-transport/src/connection/mod.rs
    @@ -917,7 +917,7 @@ impl Connection {
        }

        /// Process new input datagrams on the connection.
    -    pub fn process_input(&mut self, d: Datagram, now: Instant) {
    +    pub fn process_input<'a>(&mut self, d: Datagram<'a>, now: Instant) {
    ```



Intuitively, from the  `process_*` perspective, getting a reference to a `Datagram` (i.e. this pull request) or an owned instance of `Datagram` holding a reference to the data (i.e. `&'a [u8]`), should be identical. From Gecko's perspective, this saves an allocation on each `Datagram`.

(Off the top of my head, only issue I see is with `save_datagram`. Though here we might simply be able to introduce an `OwnedDatagram` type holding a `Vec<u8>`.)

martinthomson:
Those changes would be nice, but it would require more work on the sending side.

An idea that got cut early in the design of neqo was to have the caller of process_output provide the memory into which the packet would be written.  We could maybe do that.  (I would also want to change our AEAD interface so that we could in place encryption for that - in-place decryption is harder if we take non-mutable slices as input...)

There is a small difference between `&Datagram` and `Datagram<'a>` in that you are copying the addressing information in the latter.  That's not huge (maybe it is around 34 bytes right now, maybe more as we get ECN markings and whatnot), but it might be noticeable.

KershawChang:
> 
> What do you think @KershawChang? I will make a draft to see whether this is feasible.

As Martin commented, if implementing this is overly complicated, we need to evaluate whether the effort is justified.

If @martinthomson has no further concerns regarding this PR, we can merge this first.

mxinden:
Thank you for the help on this patch!

---

> As Martin commented, if implementing this is overly complicated, we need to evaluate whether the effort is justified.

Agreed. I have ~40h to spare in January and I am eager to contribute more to neqo. In case there is interest from your end, let me know whether this or any other work would be helpful for the project, @KershawChang and @martinthomson.

---

> Those changes would be nice, but it would require more work on the sending side.

The receiving and sending side could be done separately, correct? E.g. as a first iteration have neqo receive a slice but still return (i.e. send) owned memory, and then later in a second iteration also update the sending side.

> An idea that got cut early in the design of neqo was to have the caller of process_output provide the memory into which the packet would be written. We could maybe do that. (I would also want to change our AEAD interface so that we could in place encryption for that - in-place decryption is harder if we take non-mutable slices as input...)

The in-place en-/decryption sounds interesting! Have these allocations shown up in Firefox memory profiles in a significant way?

> There is a small difference between `&Datagram` and `Datagram<'a>` in that you are copying the addressing information in the latter. That's not huge (maybe it is around 34 bytes right now, maybe more as we get ECN markings and whatnot), but it might be noticeable.

Each of these are static in size, correct? Given their static nature and their size, they can be and currently are stack allocated. Thus intuitively I would expect there to be no performance impact. (To be validated by a benchmark.) Am I missing something?

martinthomson:
> Given their static nature and their size, they can be and currently are stack allocated. Thus intuitively I would expect there to be no performance impact. (To be validated by a benchmark.) Am I missing something?

No, that should work, but it is worth noting that there is a bunch of copying needed just to construct them.

PR #1544, opened 2024-01-08
Fix qlog issues to enable correct qvis rendering

larseggert:
Fixes include:
* Change `raw.payload_length` to `raw.length`  - the information logged was the raw (total) length already, but the label incorrectly stated it was the _payload_ length.
* Log transport parameters with a valid (in-sequence) timestamp.
* Log `ACTIVE_CONNECTION_ID_LIMIT` and `PREFERRED_ADDRESS` to qlog.

Also:
* Add some tests to `neqo-common::qlog`

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -29,12 +29,6 @@ pub struct NeqoQlogShared {
     streamer: QlogStreamer,
 }
 
-impl NeqoQlogShared {
-    pub fn streamer(&mut self) -> &mut QlogStreamer {
-        &mut self.streamer
-    }
-}
-
 impl NeqoQlog {
     /// Create an enabled `NeqoQlog` configuration.
     /// # Errors
@@ -159,7 +153,7 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
 #[cfg(test)]
 mod test {
     use qlog::events::Event;
-    use test_fixture::{neqo_qlog_contents, new_neqo_qlog, EXPECTED_LOG_HEADER};
+    use test_fixture::{new_neqo_qlog, EXPECTED_LOG_HEADER};
 
     const EV_DATA: qlog::events::EventData =
         qlog::events::EventData::SpinBitUpdated(qlog::events::connectivity::SpinBitUpdated {
@@ -171,7 +165,7 @@ mod test {
     #[test]
     fn test_new_neqo_qlog() {
         let (_log, contents) = new_neqo_qlog();
-        assert_eq!(neqo_qlog_contents(&contents), EXPECTED_LOG_HEADER);
+        assert_eq!(contents.to_string(), EXPECTED_LOG_HEADER);
     }
 
     #[test]
@@ -179,7 +173,7 @@ mod test {
         let (mut log, contents) = new_neqo_qlog();
         log.add_event(|| Some(Event::with_time(1.1, EV_DATA)));
         assert_eq!(
-            neqo_qlog_contents(&contents),
+            contents.to_string(),
             format!(
                 "{EXPECTED_LOG_HEADER}{}",
                 EXPECTED_LOG_EVENT.replace("\"time\":0.0,", "\"time\":1.1,")
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -50,7 +50,7 @@ impl NeqoQlog {
 
     #[must_use]
     pub fn inner(&self) -> Rc<RefCell<Option<NeqoQlogShared>>> {
-        self.inner.clone()
+        Rc::clone(&self.inner)
     }
 
     /// Create a disabled `NeqoQlog` configuration.
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -170,16 +170,16 @@ mod test {
 
     #[test]
     fn test_new_neqo_qlog() {
-        let log = new_neqo_qlog();
-        assert_eq!(neqo_qlog_contents(&log), EXPECTED_LOG_HEADER);
+        let (_log, contents) = new_neqo_qlog();
+        assert_eq!(neqo_qlog_contents(&contents), EXPECTED_LOG_HEADER);
     }
 
     #[test]
     fn test_add_event() {
-        let mut log = new_neqo_qlog();
+        let (mut log, contents) = new_neqo_qlog();
         log.add_event(|| Some(Event::with_time(1.1, EV_DATA)));
         assert_eq!(
-            neqo_qlog_contents(&log),
+            neqo_qlog_contents(&contents),
             format!(
                 "{EXPECTED_LOG_HEADER}{}",
                 EXPECTED_LOG_EVENT.replace("\"time\":0.0,", "\"time\":1.1,")
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -144,3 +144,66 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
         }),
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::{new_trace, NeqoQlog};
+    use crate::Role;
+    use qlog::{
+        events::{Event, EventImportance},
+        streamer::QlogStreamer,
+    };
+    use std::io::Cursor;
+
+    // TODO: Find event with less info.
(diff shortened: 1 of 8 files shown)


neqo-common/src/qlog.rs
```diff
@@ -144,3 +144,71 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
+        assert!(log.is_ok());
+        log.unwrap()
+    }
+
+    fn log_contents(log: &NeqoQlog) -> String {
+        // TODO: Figure out a way to make this less ugly.
+        #[allow(clippy::borrowed_box)]
+        let w: &Box<std::io::Cursor<Vec<u8>>> = unsafe {
+            #[allow(clippy::transmute_ptr_to_ptr)]
+            std::mem::transmute(log.inner.borrow_mut().as_mut().unwrap().streamer.writer())
```
martinthomson, line 191: Please try to avoid using `unsafe`.  You should be able to do that by implementing Writer on a struct that holds a shared reference to the underlying Vec.  You'll need to use Arc for the shared reference (stupid Send+Sync requirement in the qlog interface).

larseggert: I'll try to work it out. That particular piece of magic came from GitHub Copilot, because my Rust-fu isn't there yet to understand the intricacies here.

martinthomson: This might help...

```rust
#[derive(Clone)]
struct SharedVec {
  buf: Arc<Mutex<Cursor<Vec<u8>>>>,
}
impl Writer for SharedVec {
    fn write(&mut self, buf: &[u8]) -> Result<usize> {
        self.buf.lock().write(buf)
    }
    fn flush(&mut self) -> Result<()> {
        self.buf.lock().flush(buf)
    }
}
```

larseggert: OK, I am creating streamer as 
```rust
    let streamer = QlogStreamer::new(
        qlog::QLOG_VERSION.to_string(),
        None,
        None,
        None,
        std::time::Instant::now(),
        trace,
        EventImportance::Base,
        Box::new(SharedVec {
            buf: Arc::new(Mutex::new(Cursor::new(Vec::new()))),
        }),
    );
```
but am then struggling to get at the `SharedVec`:
```rust
pub fn neqo_qlog_contents(log: &NeqoQlog) -> String {
    let w = log
        .inner()
        .borrow_mut()
        .as_mut()
        .unwrap()
        .streamer()
        .writer();

    String::from_utf8(w.as_ref().get_ref().clone()).unwrap() // Not like this?
}
```

martinthomson: Create it like this instead:

```rust
let buf = Arc::new(Mutex::new(Cursor::new(Vec::new())));
let other_buf = Arc::clone(&buf);
let streamer = ... SharedVec { buf } ...;
```

You will then have to return other_buf, which you can query the contents of at will.

neqo-common/src/qlog.rs
```diff
@@ -48,6 +54,11 @@ impl NeqoQlog {
         })
     }
 
+    #[must_use]
+    pub fn inner(&self) -> Rc<RefCell<Option<NeqoQlogShared>>> {
+        Rc::clone(&self.inner)
```
mxinden, line 59:
```suggestion
        self.inner.clone()
```

Compiles for me on Rust `v0.175.0`. Maybe a bit easier to read?

mxinden (commented):
> Change `raw.payload_length` to `raw.length` - the information logged was the raw (total) length already, but the label incorrectly stated it was the _payload_ length.

Great catch!

I left a couple of suggestions below. Totally unsolicited. Hope you don't mind. Please feel free to ignore.

neqo-common/src/qlog.rs
martinthomson: The guidance for Rc::clone is to use this form: https://rust-lang.github.io/rust-clippy/master/index.html#/clone_on_ref_ptr

larseggert:
@mxinden thanks for the suggestions! I think I rolled all of them in.

neqo-common/src/qlog.rs
mxinden: Oh, I did not know. Thanks for the pointer!

neqo-common/src/qlog.rs
```diff
@@ -144,3 +149,40 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
+
+    const EV_DATA: qlog::events::EventData =
+        qlog::events::EventData::SpinBitUpdated(qlog::events::connectivity::SpinBitUpdated {
+            state: true,
+        });
+
+    const EXPECTED_LOG_EVENT: &str = concatcp!(
+        "\u{1e}",
+        r#"{"time":0.0,"name":"connectivity:spin_bit_updated","data":{"state":true}}"#,
+        "\n"
```
martinthomson, line 167: Any reason that [`concat!()`](https://doc.rust-lang.org/std/macro.concat.html) is insufficient here?

neqo-common/src/qlog.rs
```diff
@@ -144,3 +149,40 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
+
+    #[test]
+    fn test_add_event() {
+        let (mut log, contents) = new_neqo_qlog();
+        log.add_event(|| Some(Event::with_time(1.1, EV_DATA)));
+        assert_eq!(
+            contents.to_string(),
+            format!(
+                "{EXPECTED_LOG_HEADER}{}",
+                EXPECTED_LOG_EVENT.replace("\"time\":0.0,", "\"time\":1.1,")
```
martinthomson, line 184:
```suggestion
                "{EXPECTED_LOG_HEADER}{e}",
                e = EXPECTED_LOG_EVENT.replace("\"time\":0.0,", "\"time\":1.1,")
```
With inline items, I prefer to have all of them inline like this.

neqo-common/src/qlog.rs
```diff
@@ -144,3 +149,40 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
+        });
+
+    const EXPECTED_LOG_EVENT: &str = concatcp!(
+        "\u{1e}",
+        r#"{"time":0.0,"name":"connectivity:spin_bit_updated","data":{"state":true}}"#,
+        "\n"
+    );
+
+    #[test]
+    fn test_new_neqo_qlog() {
```
martinthomson, line 171:
```suggestion
    fn new_neqo_qlog() {
```

neqo-common/src/qlog.rs
```diff
@@ -144,3 +149,40 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
+        let (_log, contents) = new_neqo_qlog();
+        assert_eq!(contents.to_string(), EXPECTED_LOG_HEADER);
+    }
+
+    #[test]
+    fn test_add_event() {
```
martinthomson, line 177:
```suggestion
    fn add_event() {
```

test-fixture/src/lib.rs
```diff
@@ -323,3 +337,57 @@ pub fn split_datagram(d: &Datagram) -> (Datagram, Option<Datagram>) {
...
+        Box::new(SharedVec { buf }),
+    );
+    let log = NeqoQlog::enabled(streamer, "");
+    (
+        log.expect("to be able to write to new log"),
+        SharedVec { buf: contents },
+    )
+}
+
+pub const EXPECTED_LOG_HEADER: &str = concatcp!(
```
martinthomson, line 389: As above for [`concat!()`](https://doc.rust-lang.org/std/macro.concat.html).

test-fixture/src/lib.rs
```diff
@@ -323,3 +337,57 @@ pub fn split_datagram(d: &Datagram) -> (Datagram, Option<Datagram>) {
...
+    let contents = Arc::clone(&buf);
+    let streamer = QlogStreamer::new(
+        qlog::QLOG_VERSION.to_string(),
+        None,
+        None,
+        None,
+        std::time::Instant::now(),
+        trace,
+        EventImportance::Base,
+        Box::new(SharedVec { buf }),
```
martinthomson, line 380: You should be able to make a `SharedVec` and clone it directly to save some code, like:

```rust
let buf = SharedVec { buf: Arc::default() };
let streamer = QlogStreamer::new( ..., Box::new(buf.clone()));
...
( ..., buf)
```

martinthomson (approved):

PR #1545, opened 2024-01-10
Undo the restriction on the time dependency.

glandium:
It was done because of time's MSRV, but neqo's MSRV was raised to 1.70.0, which now more recent than what any version of time requires.


glandium:
I'll add that restricting the dependency in this way is not a nice thing to do because it blocks things that depend on the crate and that don't have its MSRV restriction to use a newer version. (Like, I found this because I tried to update time on mozilla-central)

PR #1546, opened 2024-01-10
Ensure that too much crypto data is fatal for a connection

martinthomson:
This is a denial of service opportunity, but one that only is available to an authenticated peer or while we are willing to receive Initial packets.  This fits the existing DoS threat model.


KershawChang (approved):
Looks good!

PR #1550, opened 2024-01-12
Make pacing configurable

KershawChang:
This PR enables tests in pacer.rs and also make pacing configurable.

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -26,6 +26,8 @@ const PACER_SPEEDUP: usize = 2;
 
 /// A pacer that uses a leaky bucket.
 pub struct Pacer {
+    /// Whether pacing is enabled.
+    enabled: bool,
     /// The last update time.
     t: Instant,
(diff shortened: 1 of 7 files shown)


neqo-transport/src/pace.rs
```diff
@@ -78,6 +86,12 @@ impl Pacer {
     /// an estimate of the round trip time (`rtt`), the estimated congestion
     /// window (`cwnd`), and the number of bytes that were sent (`count`).
     pub fn spend(&mut self, now: Instant, rtt: Duration, cwnd: usize, count: usize) {
+        if !self.enabled {
+            self.c = self.m;
```
martinthomson, line 90:
```suggestion
```

The value of `self.c` is assigned to `m` and doesn't change.

If you were to add a `set_enabled(bool)` function, then you would want to have that function set `self.c` like this when disabling pacing, as you do in the constructor.  But you don't need to do it every time.

martinthomson (approved):
Just a tiny nit.

Issue #1552, opened 2024-01-14
consider building and pushing Interop Runner Docker image in CI


marten-seemann:
I noticed that the neqo interop runner image is now 3 years (!) old: https://hub.docker.com/r/neqoquic/neqo-qns/tags. I assume that building / pushing was done manually using [this script](https://github.com/mozilla/neqo/blob/main/qns/update.sh) so far.

May I suggest to push a new interop runner on every push to master, or on a daily (?) cron job? GitHub Actions has some predefined workflows that should make building and pushing Docker images pretty straightforward. For inspiration, [here](https://github.com/quic-go/quic-go/blob/master/.github/workflows/build-interop-docker.yml) is quic-go's workflow, and Quinn has [one](https://github.com/quinn-rs/quinn-interop/blob/main/.github/workflows/dockerhub.yml) as well.

mxinden:
I am happy to contribute the required GitHub Action in case there is interest from the neqo maintainers.

martinthomson:
@mxinden that would be great.  I had a cron job setup, but I can supply the necessary secrets to get something running if you want to do the initial work.  I remember having a bunch of problems with the docker hub setup at the time, so it might be easier to use the GitHub image store (I've moved to that for other projects).

mxinden:
I prepared https://github.com/mozilla/neqo/pull/1554/. Let me know what you think.

PR #1554, opened 2024-01-15
ci: automate quic-network-simulator docker image build & push

mxinden:
Changes
- Publish `ghcr.io/mozilla/neqo-qns` docker image on a daily schedule and test build on related pull requests.
- Remove `qns/update.sh`. Is this still needed?
- Remove `/neqo-reference` indirection in `qns/Dockerfile`.

Follow-ups
- Update [`quic-interop/quic-interop` neqo image name](https://github.com/quic-interop/quic-interop-runner/blob/ca27dcb5272a82d994337ae3d14533c318d81b76/implementations.json#L37-L41).

Fixes https://github.com/mozilla/neqo/issues/1552


.github/workflows/qns.yml
```diff
@@ -0,0 +1,46 @@
+name: qns (quic-network-simulator)
+
+on:
+  push:
+    branches: ["main", "docker"] # TODO: update
```
mxinden, line 5:
```suggestion
    branches: ["main"]
```

To be removed before merge.

.github/workflows/qns.yml
```diff
@@ -0,0 +1,47 @@
+name: qns # (quic-network-simulator)
+
+on:
+  push:
+    branches: ["main", "docker"] # TODO: remove docker
```
mxinden, line 5:
```suggestion
    branches: ["main"]
```

To be removed before merge.

qns/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH \
-    RUST_VERSION=1.45.2
+    RUST_VERSION=1.70.0
```
mxinden, line 16: Same as version in `check.yml`:

 https://github.com/mozilla/neqo/blob/6fb7a6cf1ffcc1af67f72313b59c05e1370b99d3/.github/workflows/check.yml#L20.

larseggert (changes requested):
Do you think this could be made a ["reusable workflow"](https://docs.github.com/en/actions/using-workflows/reusing-workflows) contingent on the ubuntu-latest/1.70.0 run of the main CI succeeding?

mxinden:
Thank you for the quick review!

> Do you think this could be made a ["reusable workflow"](https://docs.github.com/en/actions/using-workflows/reusing-workflows) contingent on the ubuntu-latest/1.70.0 run of the main CI succeeding?

@larseggert can you expand on why a _reusable workflow_ would be helpful here? Where would we be reusing this logic?

For the sake of not pushing a faulty image, I have now made the `qns-docker-image` job dependent on the `check` job.

Note that it does not depend on the matrix variant `ubuntu-latest/1.70.0` in particular, but on the success of the entire matrix. In case we want to optimize the job's runtime, I suggest looking into further caching mechanisms first. The last execution of the new job took 3m 35s.

.github/workflows/check.yml
mxinden, line 1: Given that this workflow goes beyond _check_ing now, I suggest renaming the file to e.g. `main.yml`. Thoughts?

martinthomson: I would put your additions in a new file instead.  Call it qns.yml

The conditions for this should be different.  We want pushes to occur at most daily on main.

We should not build this for pull requests on the expectation that if tests pass, this build will work, so we don't need to run this process unless someone changes the qns.yml file or the Dockerfile (conditions you might add for a pull_request in that file).

.github/workflows/check.yml
```diff
@@ -127,3 +127,51 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Extract Rust version
+        id: rust-version
+        run: |
+          RUST_VERSION=$(grep -m1 'rust-version' neqo-common/Cargo.toml | cut -d '"' -f 2)
+          echo "::set-output name=version::$RUST_VERSION"
```
martinthomson, line 142: That's going to run this with our MSRV, rather than the latest.  I would just run on stable.

qns/Dockerfile
```diff
@@ -33,35 +34,22 @@ RUN set -eux; \
...
+ADD . /neqo
 
 RUN set -eux; \
     cd /neqo; \
     RUSTFLAGS="-g -C link-arg=-fuse-ld=lld" cargo build --release \
-      --bin neqo-client --bin neqo-server; \
-    cp target/release/neqo-client target; \
-    cp target/release/neqo-server target; \
-    rm -rf target/release
+      --bin neqo-client --bin neqo-server;
```
martinthomson, line 42:
```suggestion
      --bin neqo-client --bin neqo-server
```

.github/workflows/check.yml
```diff
@@ -127,3 +127,51 @@ jobs:
...
           fail_ci_if_error: false
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  qns-docker-image: # Docker image for QUIC network simulator i.e. interop runner (https://github.com/quic-interop/quic-interop-runner)
+    needs: check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
```
martinthomson, line 137:
```suggestion
```

https://github.com/docker/build-push-action#git-context

qns/Dockerfile
```diff
@@ -10,10 +10,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+ARG RUST_VERSION
+
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
-    PATH=/usr/local/cargo/bin:$PATH \
-    RUST_VERSION=1.45.2
+    PATH=/usr/local/cargo/bin:$PATH
 
 RUN set -eux; \
     curl -sSLf "https://static.rust-lang.org/rustup/archive/1.22.1/x86_64-unknown-linux-gnu/rustup-init" -o rustup-init; \
```
martinthomson, line 20: It might be a good idea to update the version of rustup-init we use.

qns/Dockerfile
```diff
@@ -10,10 +10,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
  && apt-get autoremove -y && apt-get clean -y \
  && rm -rf /var/lib/apt/lists/*
 
+ARG RUST_VERSION
```
martinthomson, line 13: I think that we can keep this change, but I would prefer if our builds use 'stable'.

martinthomson (changes requested):

.github/workflows/check.yml
mxinden: > I would put your additions in a new file instead. Call it qns.yml

:heavy_check_mark: 

I moved the job into a separate file and adjusted the triggers. It will now run on a daily schedule. In addition, pull requests touching related files will build the image but not push it. Let me know if that is good enough.

be2d4815b62ce1fbb7ada7ef5187e3d8bde798cf 598e12a1af6328b7ef0a7947661c6ebe39598450

.github/workflows/check.yml
mxinden: a98aa1edf4e72f9c7778afa66d3056ae296d1e06 updates the workflow to use `stable`.

qns/Dockerfile
mxinden: :+1: thanks for the catch! ae21e354d529448d2ccaf690005f21a5de73c531

.github/workflows/check.yml
mxinden: Cool. This was news to me.

e2af258cd7b5670357e023ce759176d55d9d68c8

qns/Dockerfile
mxinden: :+1: 5d10edeb4fb2414debe4df774f3c3fd3cba86a64

mxinden:
Thanks @martinthomson. Mind taking another look?

mxinden:
> We'll need to update the interop runner repo with the outcome.

:+1: see https://github.com/quic-interop/quic-interop-runner/pull/344.

PR #1556, opened 2024-01-15
chore: Bump various dependencies to what FF ships with

larseggert:
Also allow patch-level variations.


martinthomson (approved):
It might be enough to drop the patch level specification for some of these. The effect is similar, but if we clearly don't care about a minimum patch level for these.  Most of these changes are patch level changes, which we could probably avoid tracking by simply dropping the minimum.

mxinden (commented):
Three suggestions aiming for simplicity. As always, please feel free to ignore. Hope this is helpful.

larseggert:
@mxinden thanks for the suggestions! I have no real preference here, I just changed them with a script and hence the tilde everywhere. I somewhat like the consistency of having it everywhere, but really don't care much. Anyone else care?

martinthomson:
So there is a material difference between "A.B.C" and "~A.B.C". For former will accept "A.D.0" where D >B, the latter won't.  There are cases where we want to avoid a minor version.  We'll find out at some point what those are as long as we don't check in `Cargo.lock`, in which case we can use "~" or compound rules as necessary.  But generally, it is a nice thing to be able to pick up new minor versions without a lot of fuss.  I prefer that over having to meticulously track dependencies.

neqo-common/Cargo.toml
```diff
@@ -8,12 +8,12 @@ license = "MIT OR Apache-2.0"
...
-log = { version = "0.4.0", default-features = false }
-enum-map = "~2.7.3"
+enum-map = "2.7"
 env_logger = { version = "0.10", default-features = false }
-lazy_static = "1.3.0"
+lazy_static = "1.4"
+log = { version = "0.4", default-features = false }
 qlog = "0.11.0"
-time = { version = "0.3", features = ["formatting"] }
+time = {version = "~0.3.23", features = ["formatting"]}
```
martinthomson, line 16:
```suggestion
time = {version = "0.3.23", features = ["formatting"]}
```

neqo-http3/Cargo.toml
```diff
@@ -7,17 +7,17 @@ rust-version = "1.70.0"
...
-smallvec = "1.0.0"
+neqo-transport = { path = "./../neqo-transport" }
 qlog = "0.11.0"
-sfv = "0.9.1"
-url = "2.0"
-lazy_static = "1.3.0"
-enumset = "1.1.2"
+sfv = "~0.9.3"
+smallvec = "~1.11.1"
+url = "~2.5.0"
```
martinthomson, line 20:
```suggestion
sfv = "0.9.3"
smallvec = "1.11.1"
url = "2.5"
```

neqo-interop/Cargo.toml
```diff
@@ -7,14 +7,13 @@ rust-version = "1.70.0"
...
+lazy_static = "~1.4.0"
 neqo-common = { path="./../neqo-common" }
+neqo-crypto = { path = "./../neqo-crypto" }
 neqo-http3 = { path = "./../neqo-http3" }
 neqo-qpack = { path = "./../neqo-qpack" }
-
-structopt = "0.3.7"
-lazy_static = "1.3.0"
+neqo-transport = { path = "./../neqo-transport" }
+structopt = "~0.3"
```
martinthomson, line 16:
```suggestion
structopt = "0.3"
```

neqo-qpack/Cargo.toml
```diff
@@ -7,13 +7,13 @@ rust-version = "1.70.0"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
+lazy_static = "~1.4.0"
+log = {version = "~0.4.17", default-features = false}
```
martinthomson, line 11:
```suggestion
lazy_static = "1.4"
log = {version = "0.4.17", default-features = false}
```

neqo-qpack/Cargo.toml
```diff
@@ -7,13 +7,13 @@ rust-version = "1.70.0"
...
 neqo-common = { path = "./../neqo-common" }
-neqo-transport = { path = "./../neqo-transport" }
 neqo-crypto = { path = "./../neqo-crypto" }
-log = {version = "0.4.0", default-features = false}
-static_assertions = "1.1.0"
+neqo-transport = { path = "./../neqo-transport" }
 qlog = "0.11.0"
-lazy_static = "1.3.0"
+static_assertions = "~1.1.0"
```
martinthomson, line 16:
```suggestion
static_assertions = "1.1"
```

neqo-server/Cargo.toml
```diff
@@ -7,17 +7,17 @@ rust-version = "1.70.0"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-neqo-crypto = { path = "./../neqo-crypto" }
-neqo-transport = { path = "./../neqo-transport" }
+log = {version = "~0.4.17", default-features = false}
+mio = "~0.6.23"
+mio-extras = "~2.0.6"
```
martinthomson, line 12:
```suggestion
log = {version = "0.4.17", default-features = false}
mio = "0.6.23"
mio-extras = "2.0.6"
```

neqo-server/Cargo.toml
```diff
@@ -7,17 +7,17 @@ rust-version = "1.70.0"
...
 neqo-qpack = { path = "./../neqo-qpack" }
-structopt = "0.3.7"
-regex = "1"
-mio = "0.6.17"
-mio-extras = "2.0.5"
-log = {version = "0.4.0", default-features = false}
+neqo-transport = { path = "./../neqo-transport" }
 qlog = "0.11.0"
+regex = "~1.9"
+structopt = "~0.3"
```
martinthomson, line 20:
```suggestion
regex = "1.9"
structopt = "0.3"
```

neqo-transport/Cargo.toml
```diff
@@ -7,13 +7,13 @@ rust-version = "1.70.0"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-neqo-crypto = { path = "../neqo-crypto" }
+indexmap = "~1.9.3"
+lazy_static = "~1.4.0"
+log = {version = "~0.4.17", default-features = false}
```
martinthomson, line 12:
```suggestion
indexmap = "1.9.3"
lazy_static = "1.4"
log = {version = "0.4.17", default-features = false}
```

martinthomson (approved):

PR #1557, opened 2024-01-15
fix(pace): conditional compile for test not testS

mxinden:
Fix typo in `neqo-transport` `pace` conditional compilation of test module.

See also https://doc.rust-lang.org/reference/conditional-compilation.html


martinthomson (commented):
See #1550.

PR #1560, opened 2024-01-16
refactor: address clippy::single-match-else

mxinden:
Fixes clippy lint `single-match-else`.

https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else

(I don't find `if let Some` more intuitive than `match ...`. I find the type explicitness of the latter easier to reason about. Not a strong opinion. Happy to add `allow`s instead of fixing the lint, in case you feel the same.)

Meta issue: https://github.com/mozilla/neqo/issues/553


neqo-transport/src/connection/mod.rs
```diff
@@ -1261,31 +1261,25 @@ impl Connection {
...
+                        || packet.scid() != self.odcid().unwrap()
+                        || matches!(self.address_validation, AddressValidationInfo::Retry { .. })
+                    {
+                        // Ignore VersionNegotiation packets that contain the current version.
+                        // Or don't have the right connection ID.
+                        // Or are received after a Retry.
+                        self.stats.borrow_mut().pkt_dropped("Invalid VN");
                         return Ok(PreprocessResult::End);
                     }
+
```
martinthomson, line 1277: This block could probably be an else.  Then the return statement can be moved outside of all three blocks here.

neqo-transport/src/server.rs
```diff
@@ -625,28 +625,23 @@ impl Server {
...
-                }
-                _ => {
-                    qtrace!([self], "Go dormant");
-                    Output::None
-                }
-            },
-        }
+                })
+            })
+            .unwrap_or_else(|| {
```
martinthomson, line 641: Nice.

neqo-transport/src/send_stream.rs
```diff
@@ -1595,16 +1595,13 @@ impl SendStreams {
...
+            } else {
+                qdebug!("   None")
             }
-            if !self
-                .map
-                .get_mut(&stream_id)
-                .unwrap()
-                .write_frames_with_early_return(priority, builder, tokens, stats)
-            {
+            if !stream.write_frames_with_early_return(priority, builder, tokens, stats) {
```
martinthomson, line 1604: Nice.

martinthomson (approved):

neqo-transport/src/connection/mod.rs
mxinden: Well spotted. Thank you. Done via 2586da491114be7e43ee8060d98a695d7b336a9f.

mxinden:
Thank you for the review @martinthomson. Mind taking another look?

martinthomson (approved):

PR #1561, opened 2024-01-16
Make ConnectionIdRef Copy

martinthomson:
This is just a reference, so Copy is cheap.
This reduces the number of unnecessary references that are taken.


mxinden (approved):
Neat. Thank you for the follow-up! For what my review is worth, this looks good to me.

PR #1562, opened 2024-01-16
Add workflow_dispatch for QNS, reduce rate

martinthomson:
We don't need this to run so often.

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -2,7 +2,8 @@ name: QUIC Network Simulator
 
 on:
   schedule:
-    - cron: '42 3 * * *' # Runs at 03:42 UTC (m and h chosen arbitrarily) every day.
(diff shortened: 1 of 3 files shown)


.github/workflows/qns.yml
```diff
@@ -2,7 +2,8 @@ name: QUIC Network Simulator
 
 on:
   schedule:
-    - cron: '42 3 * * *' # Runs at 03:42 UTC (m and h chosen arbitrarily) every day.
+    - cron: '42 3 * * 2,5' # Runs at 03:42 UTC (m and h chosen arbitrarily) twice a week.
+  workflow_dispatch:
```
mxinden, line 6: :+1:

mxinden (approved):
For what my review is worth, looks good to me. Thank you for the follow-up!

PR #1563, opened 2024-01-17
fix(qns): always use Version1 on server

mxinden:
The Quic Interop Runner expects the server to use `Version::Version1` in all testcases but `versionnegotiation` and `v2`.

See discussion in [Quic Interop Runner issue](https://github.com/quic-interop/quic-interop-runner/pull/344#issuecomment-1895771528) and corresponding code in the [Quic Interop Runner](https://github.com/quic-interop/quic-interop-runner/blob/ca27dcb5272a82d994337ae3d14533c318d81b76/testcases.py#L188-L195).

While Neqo's default `Version` is `Version1`, the default `VersionConfig` includes all versions (`Version::all()`) and thus the server will currently upgrade the connection to the compatible `Version::Version2`.

With this commit, the server will always choose `Version1` when running a qns testcase. Given that Neqo's qns implementation does not support the `versionnegotiation` and `v2` testcases, this fix is applicable for all supported testcases.

@larseggert I opted for a fix in `neqo-server/src/main.rs` instead of, as you suggested, in `qns/interop.sh`. I find it much easier to maintain this logic in Rust rather than a shell script. I don't feel strongly about it. Let me know whether you prefer the `-V 1` fix in `interop.sh`.

---

While this does not fix all interop test failures, it does get us much closer.

Before:

image

After:

image

I will take a look at the remaining failures next.

### DIFF
--- a/neqo-server/src/main.rs
+++ b/neqo-server/src/main.rs
@@ -836,14 +836,14 @@ fn main() -> Result<(), io::Error> {
     init_db(args.db.clone());
 
     if let Some(testcase) = args.qns_test.as_ref() {
-        assert!(
-            args.quic_parameters.quic_version.is_empty(),
-            "-V and --qns-test are mutually exclusive. Each qns testcase itself dictates the supported versions.",
-        );
-        // Quic Interop Runner expects the server to support `Version1` only.
-        // Exceptions are testcases `versionnegotiation` and `v2`. Neither are
-        // supported by Neqo. Thus always set `Version1`.
-        args.quic_parameters.quic_version = vec![VersionArg(Version::Version1)];
+        if args.quic_parameters.quic_version.is_empty() {
+            // Quic Interop Runner expects the server to support `Version1` only.
+            // Exceptions are testcases `versionnegotiation` and `v2`. Neither are
+            // supported by Neqo. Thus always set `Version1`.
+            args.quic_parameters.quic_version = vec![VersionArg(Version::Version1)];
+        } else {
+            qwarn!("Both -V and --qns-test were set. Ignoring testcase specific versions.");
+        }
 
         match testcase.as_str() {
             "http3" => (),
--- a/neqo-server/src/main.rs
+++ b/neqo-server/src/main.rs
@@ -836,6 +836,15 @@ fn main() -> Result<(), io::Error> {
     init_db(args.db.clone());
 
     if let Some(testcase) = args.qns_test.as_ref() {
+        assert!(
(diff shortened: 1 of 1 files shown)


neqo-server/src/main.rs
```diff
@@ -836,6 +836,15 @@ fn main() -> Result<(), io::Error> {
     init_db(args.db.clone());
 
     if let Some(testcase) = args.qns_test.as_ref() {
+        assert!(
```
martinthomson, line 839: I'd rather leave the footgun in place (that what these tools are for) and only set the version if it is unset.

mxinden:
@martinthomson can you take another look? Let me know if you want me to remove the `qwarn!`.

neqo-server/src/main.rs
```diff
@@ -836,6 +836,15 @@ fn main() -> Result<(), io::Error> {
     init_db(args.db.clone());
 
     if let Some(testcase) = args.qns_test.as_ref() {
+        if args.quic_parameters.quic_version.is_empty() {
+            // Quic Interop Runner expects the server to support `Version1` only.
+            // Exceptions are testcases `versionnegotiation` and `v2`. Neither are
+            // supported by Neqo. Thus always set `Version1`.
```
larseggert, line 842: Neqo does support v2?

mxinden: As far as I can tell, it does not implement the `v2` Quic Interop Runner testcase:

https://github.com/mozilla/neqo/blob/64fb41f47cbbbb1101484dcd37533c7b295fa659/neqo-client/src/main.rs#L989-L1017

Am I missing something @larseggert?

mxinden: Happy to work on it once the current test failures (https://github.com/quic-interop/quic-interop-runner/pull/344) are resolved.

larseggert:
I'd prefer to lock the version in the shell script, but don't care strongly.

martinthomson (approved):
This is fine, though if Lars would prefer a simple version override in the interop runner script, that's also fine.

mxinden:
@larseggert note that once we implement the `versionnegotiation` and `v2` testcase, this logic will have to live in `neqo-server/src/main.rs`. We will then have to use `Version::Version1` in all but those two cases. In other words, a simpel `-V 1` override in `interop.sh` won't do.

Let me know if that changes your mind. If not, given that neither of us feels strongly about it, let's go with your approach, namely to override in `interop.sh`.

larseggert:
That's a good point. Let's do it this way.

PR #1565, opened 2024-01-18
Don't crash when timeBeginPeriod/timeEndPeriod returns an error

KershawChang:
See the crash in this [bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1874581).
The crash reason `assertion failed: (left == right) left: 0, right: 97` means that `timeEndPeriod` returns the error `TIMERR_NOCANDO`.
It's unclear to me why we got this error, but I think we should not crash in this case.

### DIFF
--- a/neqo-common/src/hrtime.rs
+++ b/neqo-common/src/hrtime.rs
@@ -301,7 +301,7 @@ impl Time {
         #[cfg(windows)]
         {
             if let Some(p) = self.active {
-                assert_eq!(0, unsafe { timeBeginPeriod(p.as_uint()) });
+                unsafe { timeBeginPeriod(p.as_uint()); };
             }
         }
     }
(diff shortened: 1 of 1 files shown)


neqo-common/src/hrtime.rs
```diff
@@ -301,7 +301,9 @@ impl Time {
         #[cfg(windows)]
         {
             if let Some(p) = self.active {
-                assert_eq!(0, unsafe { timeBeginPeriod(p.as_uint()) });
+                unsafe {
+                    timeBeginPeriod(p.as_uint());
+                };
```
martinthomson, line 306: Are we likely to get an unused return value warning?  Should this be:

```suggestion
                _ = unsafe { timeBeginPeriod(p.as_uint()) };
```

martinthomson (approved):
The alternative would be to use `debug_assert_eq!()`, but I don't think we run CI on Windows.

neqo-common/src/hrtime.rs
KershawChang: Will do.
Thanks.

PR #1567, opened 2024-01-18
feat: Add the ability to send PING frames

larseggert:


martinthomson (commented):
I mentioned an alternative approach to getting a frame through in Slack.  I'm not completely opposed to expanding the API surface, but it might be easier to use frame injection for the tests we care about.

larseggert:
OK, if we end up needing PINGs to force idle I will look into using the backdoor instead of this.

larseggert:
Making this a draft since we may not need/want this.

larseggert:
Closing this. Can reopen if we intend to support this API.

PR #1568, opened 2024-01-18
Add TOS and TTL information to Datagrams

larseggert:
This part of the refactor of #1495.

This doesn't include any ECN I/O or other logic, just adds the required fields to the datagram header.

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -170,7 +170,7 @@ fn ecn() {
 fn fmt_datagram() {
     let d = datagram([0; 1].to_vec());
     assert_eq!(
-        format!("{:?}", d),
+        format!("{d:?}"),
         "Datagram TOS Some(2) TTL Some(128) [fe80::1]:443->[fe80::1]:443: [1]: 00".to_string()
     );
 }
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -150,3 +150,27 @@ impl std::fmt::Debug for Datagram {
         )
     }
 }
+
+#[cfg(test)]
+use test_fixture::datagram;
+
+#[test]
+fn ecn() {
+    assert_eq!(IpTosEcn::from(0b00), IpTosEcn::NotEct);
+    assert_eq!(IpTosEcn::from(0b01), IpTosEcn::Ect1);
+    assert_eq!(IpTosEcn::from(0b10), IpTosEcn::Ect0);
+    assert_eq!(IpTosEcn::from(0b11), IpTosEcn::Ce);
+    assert_eq!(u8::from(IpTosEcn::NotEct), 0b00);
+    assert_eq!(u8::from(IpTosEcn::Ect1), 0b01);
+    assert_eq!(u8::from(IpTosEcn::Ect0), 0b10);
+    assert_eq!(u8::from(IpTosEcn::Ce), 0b11);
+}
+
+#[test]
+fn fmt_datagram() {
+    let d = datagram([0; 1].to_vec());
+    assert_eq!(
+        format!("{:?}", d),
+        "Datagram TOS Some(2) TTL Some(128) [fe80::1]:443->[fe80::1]:443: [1]: 00".to_string()
+    );
+}
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -7,20 +7,97 @@
 use std::net::SocketAddr;
 use std::ops::Deref;
 
+use enum_map::Enum;
+
 use crate::hex_with_len;
+
+// ECN (Explicit Congestion Notification) codepoints mapped to the
+// lower 2 bits of the TOS field.
+// https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml
+#[derive(Copy, Clone, PartialEq, Eq, Enum)]
+#[repr(u8)]
+pub enum IpTosEcn {
+    NotEct = 0b00, // Not-ECT (Not ECN-Capable Transport) [RFC3168]
+    Ect1 = 0b01,   // ECT(1) (ECN-Capable Transport(1))[1] [RFC8311][RFC Errata 5399][RFC9331]
+    Ect0 = 0b10,   // ECT(0) (ECN-Capable Transport(0)) [RFC3168]
+    Ce = 0b11,     // CE (Congestion Experienced) [RFC3168]
+}
+
+impl From<u8> for IpTosEcn {
+    fn from(v: u8) -> Self {
+        match v & 0b11 {
+            0b00 => IpTosEcn::NotEct,
+            0b01 => IpTosEcn::Ect1,
+            0b10 => IpTosEcn::Ect0,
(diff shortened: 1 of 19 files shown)


neqo-common/src/datagram.rs
```diff
@@ -7,20 +7,103 @@
...
+    Af31 = 0b0110_1000,       // [RFC2597]
+    Af32 = 0b0111_0000,       // [RFC2597]
+    Af33 = 0b0111_1000,       // [RFC2597]
+    Af41 = 0b1000_1000,       // [RFC2597]
+    Af42 = 0b1001_0000,       // [RFC2597]
+    Af43 = 0b1001_1000,       // [RFC2597]
+    Ef = 0b1011_1000,         // [RFC3246]
+    VoiceAdmit = 0b1011_0000, // [RFC5865]
+    Le = 0b0000_0100,         // [RFC8622]
+}
```
martinthomson, line 83: This one doesn't have any From implementations.

I wonder whether it would make sense to have an `impl From<u8> for (IpTosDscp, IpTosEcn)`.  You might even have a `type IpTos = (IpTosDscp, IpTosEcn)` alias or a `struct IpTos(u8)` that has appropriate `From` implementations attached.  I guess that's something we'd want if DSCP were more serious, but we need to be realistic.

neqo-common/src/datagram.rs
```diff
@@ -7,20 +7,103 @@
...
+
 #[derive(PartialEq, Eq, Clone)]
 pub struct Datagram {
     src: SocketAddr,
     dst: SocketAddr,
+    tos: Option<u8>,
+    ttl: Option<u8>,
```
martinthomson, line 90: If these arguments were enum types, we could implement `Default` and avoid the `Option` wrapper.  I know that implementing `Default` might be a bit on the nose, but I don't see a problem with doing that within our code.

martinthomson (approved):
This is looking good.

I'm semi-serious about having a wrapper type for `IpTos`.  You have a case where you set a `tos` argument to a value of the Ecn enum.  But I think that you want a wrapper type (that trivially converts to `u8`) that has simple functions for converting from `IpTosEcn` and `(IpTosDscp, IpTosEcn)`.

neqo-common/src/datagram.rs
larseggert: Would you do an enum for `ttl`? (Will we ever care about TTL, or should I just drop that?)

neqo-common/src/datagram.rs
larseggert: Something like this?
```rust
#[derive(Copy, Clone, PartialEq, Eq, Default)]
pub struct IpTos((IpTosDscp, IpTosEcn));

#[derive(Clone, PartialEq, Eq)]
pub struct Datagram {
    src: SocketAddr,
    dst: SocketAddr,
    tos: IpTos,
    ttl: Option<u8>,
    d: Vec<u8>,
}
```

Would we add a `None` default value to `IpTosDscp` and  `IpTosEcn` then?

larseggert:
@martinthomson does the current version of this PR implement what you suggested?

neqo-common/src/datagram.rs
```diff
@@ -7,20 +7,129 @@
...
+    Af43 = 0b1001_1000,       // Assured Forwarding 43, [RFC2597]
+    Ef = 0b1011_1000,         // Expedited Forwarding, [RFC3246]
+    VoiceAdmit = 0b1011_0000, // Capacity-Admitted Traffic, [RFC5865]
+    Le = 0b0000_0100,         // Lower-Effort, [RFC8622]
+}
+
+impl std::fmt::Debug for IpTosDscp {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            IpTosDscp::Cs0 => f.write_str("CS0"),
```
martinthomson, line 72: There's a little less typing if you do
```rust
let s = match self {
Self::foo -> "foo",
...
};
f.write_str(s)
```

Though I would say that it would also be OK to just `#[derive(Debug)]`.  It's just debugging.

neqo-common/src/datagram.rs
martinthomson: This works.

The alternative is to have IpTos backed by u8.  Then you would do:

```rust
impl From<IpTosEcn> for IpTos {
  fn from(v: IpTosEcn) -> Self {  Self(u8::from(v)) }
}
impl From<IpTosDscp> for IpTos {
  fn from(v: IpTosDscp) -> Self {  Self(u8::from(v)) }
}
impl From<(IpTosDscp, IpTosEcn)> for IpTos {
  fn from(v: (IpTosDscp, IpTosEcn)) -> Self {  Self(u8::from(v.0) | u8::from(v.1)) }
}
impl From<IpTos> for u8 {
  fn from(v: IpTos) -> Self { v.0 }
}
```

I believe that a `Default` implementation works here with the value equal to zero and that `#[derive(Default)]` does that.

The advantage of that is that you can invoke functions with `IpTos::from(IpTosEcn::ECT1)` and `IpTos::from(IpTosDscp::Cs1)` and `IpTos::from((IpTosDscp::Cs1,IpTosEcn::ECT1))` and `IpTos::from(0x05)` equally.

The disadvantage is that you need a `Debug` implementation that is a little more complex.

```rust
        f.debug_tuple("IpTos")
            .field(&IpTosDscp::from(self.0 & 0xfc))
            .field(&IpTosEcn::from(self.0 & 0x3))
            .finish()
```

neqo-common/src/datagram.rs
```diff
@@ -7,20 +7,129 @@
...
 use crate::hex_with_len;
 
-#[derive(PartialEq, Eq, Clone)]
+// ECN (Explicit Congestion Notification) codepoints mapped to the
+// lower 2 bits of the TOS field.
+// https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml
+#[derive(Copy, Clone, PartialEq, Eq, Enum, Default)]
+#[repr(u8)]
+pub enum IpTosEcn {
+    #[default]
```
martinthomson, line 20: I like enum-map, but it's a bit of a shame to have to pick it up for just this.  That should be a language-native feature.

neqo-common/src/datagram.rs
```diff
@@ -7,20 +7,129 @@
...
     d: Vec<u8>,
 }
 
 impl Datagram {
-    pub fn new<V: Into<Vec<u8>>>(src: SocketAddr, dst: SocketAddr, d: V) -> Self {
+    pub fn new<V: Into<Vec<u8>>>(
+        src: SocketAddr,
+        dst: SocketAddr,
+        tos: IpTos,
+        ttl: Option<u8>,
```
martinthomson, line 125: FWIW, you could also new-type `IpTtl(u8)` and `impl Default for IpTtl { fn default() -> Self { Self(128) } }` to save an `Option<>` wrapper.

I guess that the only potential problem is that if the OS has some different default for TTL, we might start overriding that unnecessarily.  I'm not sure that I care about that though.

martinthomson (approved):
As always, I have suggestions, but none of them should block landing should you need that.

PR #1569, opened 2024-01-19
fix(client): prevent out-of-range slice access

mxinden:
The Quic Network Simulator `handshakeloss` testcase instructs the `neqo-client` to download from multiple URLs in series. The client takes one URL after the next and tries to request it:

``` rust
to_request = &remaining[..1];
remaining = &remaining[1..];
```

This will panic on the last URL, trying to access the first element in the empty slice `remaining`.

This commit removes the out-of-range access. Instead of using slice operations which hide potential panics, it uses owned collections (`VecDeque` and `Vec`) and their safe methods (e.g. `pop_front`, `drain`).

---

Needed for https://github.com/quic-interop/quic-interop-runner/pull/344.
Related to https://github.com/mozilla/neqo/issues/1552.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -832,7 +832,7 @@ fn handle_test(
     local_addr: SocketAddr,
     remote_addr: SocketAddr,
     hostname: &str,
-    urls: Vec<Url>,
+    url_queue: VecDeque<Url>,
     resumption_token: Option<ResumptionToken>,
 ) -> Res<Option<ResumptionToken>> {
     let key_update = KeyUpdateState(args.key_update);
@@ -843,7 +843,7 @@ fn handle_test(
                     .expect("failed to create client");
             args.method = String::from("POST");
             let url_handler = URLHandler {
-                url_queue: VecDeque::from(urls.to_vec()),
+                url_queue,
                 stream_handlers: HashMap::new(),
                 all_paths: Vec::new(),
                 handler_type: StreamHandlerType::Upload,
@@ -912,7 +912,7 @@ fn client(
     local_addr: SocketAddr,
     remote_addr: SocketAddr,
     hostname: &str,
-    urls: Vec<Url>,
+    url_queue: VecDeque<Url>,
     resumption_token: Option<ResumptionToken>,
 ) -> Res<Option<ResumptionToken>> {
     let testcase = args.test.clone();
@@ -925,7 +925,7 @@ fn client(
             local_addr,
             remote_addr,
             hostname,
-            urls,
+            url_queue,
             resumption_token,
         );
     }
@@ -934,7 +934,7 @@ fn client(
         .expect("failed to create client");
     let key_update = KeyUpdateState(args.key_update);
     let url_handler = URLHandler {
-        url_queue: VecDeque::from(urls.to_vec()),
+        url_queue,
         stream_handlers: HashMap::new(),
         all_paths: Vec::new(),
         handler_type: StreamHandlerType::Download,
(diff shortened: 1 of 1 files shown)


neqo-client/src/main.rs
```diff
@@ -1077,23 +1078,19 @@ fn main() -> Res<()> {
...
                         "Error: resumption to {hostname} cannot work without at least 2 URLs."
                     );
                     exit(127);
                 }
+                urls.pop_front().map_or(vec![], |u| vec![u])
             } else {
-                to_request = remaining;
-                remaining = &[][..];
-            }
+                urls.drain(..).collect()
```
martinthomson, line 1092:
```suggestion
                std::mem::take(urls)
```

neqo-client/src/main.rs
```diff
@@ -832,7 +832,7 @@ fn handle_test(
     local_addr: SocketAddr,
     remote_addr: SocketAddr,
     hostname: &str,
-    urls: &[Url],
+    urls: Vec<Url>,
```
martinthomson, line 835: I would prefer if you changed this to a VecDeque.  You can see that this argument is transformed into a VecDeque for processing anyway, so you might as well save a transformation.

(At first, this change looked to be gratuitous.  The change works without it.  But I think that you are right to change to a pass-by-value in this case as the test functions all basically rely on having ownership of the list of URLs.)

neqo-client/src/main.rs
```diff
@@ -1022,13 +1022,14 @@ fn main() -> Res<()> {
         }
     }
 
-    let mut urls_by_origin: HashMap<Origin, Vec<Url>> = HashMap::new();
+    let mut urls_by_origin: HashMap<Origin, VecDeque<Url>> = HashMap::new();
     for url in &args.urls {
         let entry = urls_by_origin.entry(url.origin()).or_default();
-        entry.push(url.clone());
+        entry.push_back(url.clone());
     }
```
martinthomson, line 1029: Looking at this, I find this loop annoy, because it could produce what the next loop needs `(host, port) -> Url` more directly.

neqo-client/src/main.rs
```diff
@@ -1077,23 +1078,19 @@ fn main() -> Res<()> {
...
-        let mut remaining = &urls[..];
         let mut first = true;
         loop {
-            let to_request;
-            if (args.resume && first) || args.download_in_series {
-                to_request = &remaining[..1];
-                remaining = &remaining[1..];
-                if args.resume && first && remaining.is_empty() {
+            let to_request = if (args.resume && first) || args.download_in_series {
+                if args.resume && first && urls.len() < 2 {
```
martinthomson, line 1084: Another annoying thing.  This check could be made outside of the loop, in which case you wouldn't need `first`.

neqo-client/src/main.rs
mxinden: Oh neat!

neqo-client/src/main.rs
mxinden: Cool. Yep. Simplifies things.

25dd1a83

neqo-client/src/main.rs
mxinden: See a0975fcc.

neqo-client/src/main.rs
mxinden: I missed this. Thank you. 

ad02584d

martinthomson (approved):
Very nice.  Thanks for indulging me.  That new per-origin rollup is much cleaner.

mxinden:
Great. Any time @martinthomson!

PR #1571, opened 2024-01-21
fix: Rust intra doc links

mxinden:
- Fix minor typos in Rust intra doc links.
- Add CI check.

---

I find Rust intra doc links helpful, not so much for the links themselves, but for the static check flagging outdated documentation.


larseggert:
Thanks for this!

Are there any other rustdoc thing we should check?

Should we also build and upload the current docs for the main branch during CI to another GitHub branch for display?

PR #1574, opened 2024-01-23
ci: Add a data transfer test

larseggert:


.github/workflows/check.yml
```diff
@@ -112,6 +112,19 @@ jobs:
...
         run: cargo +${{ matrix.rust-toolchain }} llvm-cov nextest --features ci --all-targets --no-fail-fast --lcov --output-path lcov.info
 
+      - name: Run client/server transfer
+        run: |
+          cargo +${{ matrix.rust-toolchain }} build --bin neqo-client --bin neqo-server
+          cargo +${{ matrix.rust-toolchain }} run --bin neqo-server -- $HOST:4433 &
+          PID=$!
+          cargo +${{ matrix.rust-toolchain }} run --bin neqo-client -- --output-dir . https://$HOST:4433/$SIZE
+          kill $PID
+          [ "$(wc -c <"$SIZE")" -eq "$SIZE" ] || exit 1
```
martinthomson, line 122: Oh, this statement is weird...  The fact that the filename is "$SIZE" is offputting.

neqo-server/src/main.rs
```diff
@@ -318,9 +318,15 @@ impl QuicParameters {
 }
 
 fn emit_packet(socket: &mut UdpSocket, out_dgram: Datagram) {
-    let sent = socket
-        .send_to(&out_dgram, &out_dgram.destination())
-        .expect("Error sending datagram");
+    let sent = match socket.send_to(&out_dgram, &out_dgram.destination()) {
+        Err(ref err) => {
+            if err.kind() != io::ErrorKind::WouldBlock {
```
martinthomson, line 323: Do you want to avoid logging Interrupted as well?

martinthomson: A random thought:

```rust
pub trait IsRetryable {
  fn is_retryable(&self) -> bool;
}

impl IsRetryable for std::io::Error {
  fn is_retryable(&self) -> bool {
    e.kind() == ErrorKind::WouldBlock || e.kind() == ErrorKind::Interrupted
  }
}
```

martinthomson (approved):

.github/workflows/check.yml
larseggert: Yeah, there doesn't seem to be a way to choose a name for downloaded objects. I could obviously have two env variables `FILENAME` and `SIZE` that are both `54321`, but I doubt that's much clearer.

neqo-server/src/main.rs
larseggert: I might add this as part of the addition of socket I/O to neqo-common. Right now I'd need to duplicate it in the client and the server.

PR #1575, opened 2024-01-23
Fix unresolved link errors in rustdoc CI

larseggert:


mxinden:
Sorry for the trouble. Thank you for the fix!

Unless I am missing something this is due to https://github.com/mozilla/neqo/pull/1568 merging before https://github.com/mozilla/neqo/pull/1571, where https://github.com/mozilla/neqo/pull/1571 tested with a version of `main` before https://github.com/mozilla/neqo/pull/1568.

Have you considered setting the bottom check-box in the GitHub branch protection rules for `main`? That would prevent the issue going forward.

image

larseggert:
That's a great suggestion, thanks.

PR #1576, opened 2024-01-23
Use matrix toolchain when checking rustdoc

larseggert:


mxinden (commented):
:+1: Thanks! Much better.

PR #1578, opened 2024-01-23
fix(client): only take single url on (args.resume && first)

mxinden:
QUIC Interop Runner testcases `resumption` and `0-rtt` expect the client to download the first file, close the connection, and then download the remaining files on a second connection.

https://github.com/quic-interop/quic-interop-runner/tree/master#test-cases

Thus `neqo-client`, when `args.resume` is `true` must only take a single URL on the **first** loop iteration. On the second iteration it must take all remaining URLs.

Regression introduced in https://github.com/mozilla/neqo/pull/1569. My bad.

---

image

With this patch all, but the `amplification` testcase, are green. Note that the `amplification` testcase failed with the previous qns image as well. See https://interop.seemann.io/. 

```
Received a 1337 byte Handshake packet from the server. Total: 5348
Server violated the amplification limit, but stayed within 3-4x amplification. Letting it slide.
Received a 1337 byte Handshake packet from the server. Total: 6685
Server violated the amplification limit.
```

Needed for https://github.com/quic-interop/quic-interop-runner/pull/344.
Related to https://github.com/mozilla/neqo/issues/1552.

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -1086,12 +1086,15 @@ fn main() -> Res<()> {
 
         let hostname = format!("{host}");
         let mut token: Option<ResumptionToken> = None;
+        let mut first = true;
         while !urls.is_empty() {
-            let to_request = if args.resume || args.download_in_series {
+            let to_request = if (args.resume && first) || args.download_in_series {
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Oops, I missed that one.

PR #1580, opened 2024-01-24
Remove docs/linux_build.md

larseggert:
The current working way to do this is documented in .github/workflows/check.yml


valenting:
BTW, in the past I remember having to consult this readme as the other docs were missing some steps info.
It's probably fine to remove this file as long as we reference the dependencies and any additional steps in the main readme (or some other doc) - I don't expect contributors to consult .github content on their own 🙂

Issue #1487, opened 2023-11-04
investigate high CPU use in neqo_transport::send_stream::TxBuffer::next_bytes


jesup:
See https://share.firefox.dev/3tXRSIb
next_bytes uses ~24% of the CPU in this trace, almost all of it in <alloc::collections::btree::map::Iter as core::iter::traits::iterator::Iterator>::next

mb:
It seems to be this function being called repeatedly

https://github.com/mozilla/neqo/blob/2582acda8c9c2a1b00304e07f46be43246c56db7/neqo-transport/src/send_stream.rs#L168-L181

`self.used` is a [`BTreeMap`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html):

https://github.com/mozilla/neqo/blob/2582acda8c9c2a1b00304e07f46be43246c56db7/neqo-transport/src/send_stream.rs#L145-L151

`first_unmarked_range` is called repeatedly in
https://github.com/mozilla/neqo/blob/2582acda8c9c2a1b00304e07f46be43246c56db7/neqo-transport/src/send_stream.rs#L413-L414

martinthomson:
The output of `first_unmarked_range()` looks to be totally cacheable.

PR #1582, opened 2024-01-24
perf: Add a cache for first_unmarked_range()

jesup:
Cache results of RangeTracker::first_unmarked_range().
Coalesce additions to the tail of a RangeTracker tree to avoid unnecessary tree growth.

Tree length in large uploads now ~2 instead of 50-300.   85% reduction in CPU time in a trace-driven benchmark

Resolves issue #1487


neqo-transport/src/send_stream.rs
```diff
@@ -792,21 +816,23 @@ impl SendStream {
...
             SendStreamState::DataSent {
-                ref send_buf,
+                ref mut send_buf,
                 fin_sent,
                 ..
             } => {
-                let bytes = send_buf.next_bytes();
-                if bytes.is_some() {
-                    bytes
+                let used = send_buf.used(); // immutable first
```
martinthomson, line 828: I think that you just want to  add this line, not change the next two.  Your change might involve an allocation (though the compiler is probably smart enough to avoid that, no point in writing `Some((offset, slice))` twice.

martinthomson (approved):

neqo-transport/src/send_stream.rs
```diff
@@ -148,6 +148,7 @@ enum RangeState {
 struct RangeTracker {
     // offset, (len, RangeState). Use u64 for len because ranges can exceed 32bits.
     used: BTreeMap<u64, (u64, RangeState)>,
+    cached: Option<(u64, Option<u64>)>,
```
martinthomson, line 151:
```suggestion
    /// A cache for the last unmarked range, which we check an awful lot more than we update.
    last_unmarked: Option<(u64, Option<u64>)>,
```

neqo-transport/src/send_stream.rs
```diff
@@ -208,6 +215,18 @@ impl RangeTracker {
         let mut tmp_len = new_len;
         let mut v = Vec::new();
 
+        // Check for the common case of adding to the end
+        if let Some(mut last) = self.used.last_entry() {
+            let prev_off = *last.key();
+            let (prev_len, prev_state) = last.get_mut();
+            if new_off == prev_off + *prev_len && new_state == *prev_state {
```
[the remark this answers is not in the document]
jesup: Fails a test if we do that

neqo-transport/src/send_stream.rs
jesup: The compiler suggested I make this change (failed 2 checks, so I changed it to make the compiler happy via 'more cleanup').  I'll try changing it back...

jesup: And this time it passes.  Dunno, ok

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
jesup: Figured out how to do it; needs one more conditional.  Tests pass now

neqo-transport/src/send_stream.rs
```diff
@@ -208,6 +216,22 @@ impl RangeTracker {
...
 
+        // Check for the common case of adding to the end
+        if let Some(mut last) = self.used.last_entry() {
+            let prev_off = *last.key();
+            let (prev_len, prev_state) = last.get_mut();
+            // allow for overlap between new chunk and the last entry
+            if new_off >= prev_off
+                && new_off <= prev_off + *prev_len
+                && new_off + new_len > prev_off + *prev_len
+                && new_state == *prev_state
```
martinthomson, line 227: I'm sure that the compiler is capable of doing this, but maybe this check could be moved before the call to `.last_entry()`.

neqo-transport/src/send_stream.rs
martinthomson: Yeah, it's funny, but I think that you can also move the call to used back.  The only change being to the addition of 'mut'.

martinthomson (commented):
I have one final suggestion.  Move the new code here to a separate function.  Have that function `extend_final_range() -> bool`. Call that function from `mark_range` and return early if it works.  (We shouldn't need to coalesce from zero in this case either.)

neqo-transport/src/send_stream.rs
jesup: We can't, it needs prev_state which comes from last_entry()

neqo-transport/src/send_stream.rs
larseggert: @jesup any reason not to rename this as suggested?

neqo-transport/src/send_stream.rs
```diff
@@ -306,6 +336,10 @@ impl RangeTracker {
             return;
         }
 
+        self.cached = None;
+        if self.extend_final_range(off, len as u64, state) {
+            return;
+        }
         let subranges = self.chunk_range_on_edges(off, len as u64, state);
 
         for (sub_off, sub_len, sub_state) in subranges {
```
martinthomson, line 345: Another optimization, though maybe a small one:

Move `coalesce_acked_from_zero()` to `mark_as_acked()`.  Then make it faster.  (You said that you had a fix that improved that function, but I think we can do a whole lot better than that.)

neqo-transport/src/send_stream.rs
```diff
@@ -167,19 +168,46 @@ impl RangeTracker {
...
+                self.cached = Some(res);
+                return res;
             }
         }
+        self.cached = Some((prev_end, None));
         (prev_end, None)
     }
 
+    // Check for the common case of adding to the end.  If we can, do it and
+    // return true.
```
martinthomson, line 192:
```suggestion
    /// Check for the common case of adding to the end.  If we can, do it and
    /// return true.
```

neqo-transport/src/send_stream.rs
jesup: I don't like mark_as_acked(), since this isn't marking things.  cleanup_acked()?   The current name is long but descriptive

martinthomson: Not sure that I'm following you.

The problem I'm trying to address is that coalesce_acked_from_zero is called when we mark things with any state, not just when we mark things as Acked.  We could move that function call to the place where we mark things as acked.  That is all.

larseggert (approved):

PR #1585, opened 2024-01-25
Improve coalesce_acked_from_zero perf by avoiding an extra vector

jesup:
Avoiding the SmallVec<> yields about a 12% improvement in speed when we have a missing ack that shows up (lots of acked packets waiting behind it in the tree -- very rare).   For the more common case of in-order acks (tree is 0-N Acked, N+1 - N+M Sent, and we Ack N+1 to N+Q in one mark_range), we save ~6%.  And it's simpler

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -276,8 +276,6 @@ impl RangeTracker {
             .map(|(len, _)| *len);
 
         if let Some(len_from_zero) = acked_range_from_zero {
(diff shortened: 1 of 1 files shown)


larseggert (approved):

martinthomson (commented):
Though this change is clearly an improvement, I think that we can get a much bigger win by turning the acknowledged range from zero into a single u64.  Then we never have to worry about tracking that.  (It's similar to the `retired` property on the send side piece, and may indeed collapse into the same thing...)

PR #1587, opened 2024-01-26
perf: Optimize packet reception for the common case of adding to the end

jesup:
Small optimization to avoid SmallVec and range in the common case of in-order data reception.   Benchmarks on 100000 1337-byte packets shows a 4-7% improvement


neqo-transport/src/recv_stream.rs
```diff
@@ -200,41 +200,46 @@ impl RxStreamOrderer {
...
+            if *entry.key() > new_start {
+                // Now handle possible overlap with next entries
+                let mut to_remove = SmallVec::<[_; 8]>::new();
+
+                for (&next_start, next_data) in self.data_ranges.range_mut(new_start..) {
+                    let next_end = next_start + u64::try_from(next_data.len()).unwrap();
+                    let overlap = new_end.saturating_sub(next_start);
+                    if overlap == 0 {
+                        break;
+                    } else if next_end >= new_end {
```
martinthomson, line 215: No else after break...  However, the overlap == 0 block could be removed.  The next block is a noop when overlap == 0.

neqo-transport/src/recv_stream.rs
```diff
@@ -200,41 +200,46 @@ impl RxStreamOrderer {
             false
         };
 
-        // Now handle possible overlap with next entries
-        let mut to_remove = SmallVec::<[_; 8]>::new();
         let mut to_add = new_data;
+        if let Some(entry) = self.data_ranges.last_entry() {
```
martinthomson, line 204: It took me a while to see this, but your change is pretty small.

This probably needs pictures, like the ones above.

This chunk is for any ranges that might start after the one that is being proposed.

In most cases, there are no ranges after the block, so you want a quick shortcut for that case.  That is your addition:

```
//  PPPPPP
//             NNNNNNN
```

Otherwise, you then have two lots of collapsing to do:

Firstly, any ranges for which the end of the old range is inside the new range, that's no longer needed:

```
//                PPPP
//             NNNNNNN
```

That might happen multiple times AND the open iterator means that we can't mutate the collection until the loop concludes, so we have to save those items that need to get dropped.  (`BTreeMap::drain(range)` would be awesome for that, but we don't get to have nice things, so we have a `SmallVec` for that.)

Finally, the end of this might be inside the next range, so this one can be truncated before adding it:

```
//                  PPPPPPP
//             NNNNNnn
```

neqo-transport/src/recv_stream.rs
```diff
@@ -200,41 +200,46 @@ impl RxStreamOrderer {
...
+            // is this at the end (common case)?  If so, don't need this block
+            if *entry.key() > new_start {
```
martinthomson, line 206:
```suggestion
        if self.data_ranges.last_entry().map_or(false,  |e| *e.key() > new_start) {
```

martinthomson (dismissed):

jesup:
FYI, it's now about a 9% improvement on my system

jesup:
Added benchmarks in PR 1609

larseggert (approved):

PR #1589, opened 2024-01-26
chore: Make the qns image multiarch & shrink the build context further

larseggert:
Also use `git` to get nss, which is a bit faster.

CC @mxinden

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,8 +2,8 @@
 *
 # Except for the following:
 !**/*.toml
-!Cargo.lock
 !**/*.rs
 !**/*.h
 !**/*.hpp
 !qns
+!Cargo.lock
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,8 @@
-nss
-nspr
-target
+# Ignore everything:
+*
+# Except for the following:
+!**/*.toml
+!**/*.rs
+!**/*.h
+!**/*.hpp
+!qns
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 *
 # Except for the following:
 !**/*.toml
+!Cargo.lock
 !**/*.rs
 !**/*.h
 !**/*.hpp
--- a/qns/Dockerfile
+++ b/qns/Dockerfile
@@ -13,8 +13,7 @@ ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH
 
-RUN \
-    curl https://sh.rustup.rs -sSf | \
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,3 +1,8 @@
-nss
-nspr
-target
+# Ignore everything:
+*
+# Except for the following:
+!**/*.toml
```
mxinden, line 4:
```suggestion
!**/*.toml
!Cargo.lock
```

Easier for debugging, as both a local `cargo xxx` as well as the `cargo build`  in the `Dockerfile` will use the same dependencies.

mxinden (approved):
:+1:

qns/Dockerfile
```diff
@@ -14,19 +14,15 @@ ENV RUSTUP_HOME=/usr/local/rustup \
     PATH=/usr/local/cargo/bin:$PATH
 
 RUN set -eux; \
```
martinthomson, line 16: I'm less of a fan of dropping the checksum, but I can see the value.

```suggestion
RUN \
```

You don't need the set call (unless you want to call `set -o pipefail`).

martinthomson (dismissed):

qns/Dockerfile
```diff
@@ -13,20 +13,16 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
-RUN set -eux; \
-    curl -sSLf "https://static.rust-lang.org/rustup/archive/1.26.0/x86_64-unknown-linux-gnu/rustup-init" -o rustup-init; \
-    echo '0b2f6c8f85a3d02fde2efc0ced4657869d73fccfce59defb4e8d29233116e6db *rustup-init' | sha256sum -c -; \
-    chmod +x rustup-init; \
-    ./rustup-init -y -q --no-modify-path --profile minimal --default-toolchain "$RUST_VERSION"; \
-    rm -f rustup-init; \
-    chmod -R a+w "$RUSTUP_HOME" "$CARGO_HOME"
+RUN \
+    curl https://sh.rustup.rs -sSf | \
+    sh -s -- -y -q --no-modify-path --profile minimal --default-toolchain $RUST_VERSION
```
larseggert, line 18:
```suggestion
RUN curl https://sh.rustup.rs -sSf | \
    sh -s -- -y -q --no-modify-path --profile minimal --default-toolchain $RUST_VERSION
```

.dockerignore
```diff
@@ -1,3 +1,9 @@
...
+!Cargo.lock
+!**/*.rs
+!**/*.h
+!**/*.hpp
+!qns
```
larseggert, line 9:
```suggestion
!**/*.toml
!**/*.rs
!**/*.h
!**/*.hpp
!qns
!Cargo.lock
```

PR #1592, opened 2024-01-26
test: benchmarks for coalesce/mark_range

jesup:


martinthomson (commented):
See #1594 for a few updates on this benchmark that should make it a little nicer.

neqo-transport/src/lib.rs
```diff
@@ -28,7 +28,7 @@ mod quic_datagrams;
 mod recovery;
 mod recv_stream;
 mod rtt;
-mod send_stream;
+pub mod send_stream;
```
larseggert, line 31: See how @martinthomson does it here with a new feature: https://github.com/mozilla/neqo/pull/1594/files#diff-ea9df8dd0e492efc27d0690a0529971f8debe8ff01291cee81ca7db8ab1e6465R31-R33

neqo-transport/src/send_stream.rs
```diff
@@ -136,15 +136,15 @@ impl Default for RetransmissionPriority {
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
-enum RangeState {
+pub enum RangeState {
```
larseggert, line 139: Might use the same trick here and below where `pub` is added.

larseggert (changes requested):

PR #1593, opened 2024-01-26
perf: Move acked-from-zero out of the tree and directly into RangeTracker

jesup:
Simplifies coalesce, but complicates mark_range slightly because the code will try to make data as acked multiple times.

benchmarks improve (with the bench_coalesce PR) by ~65% for the common case (2 nodes), and 75% in the edge-cases

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -137,7 +137,7 @@ impl Default for RetransmissionPriority {
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
-enum RangeState {
+pub enum RangeState {
     Sent,
     Acked,
 }
@@ -145,7 +145,7 @@ enum RangeState {
 /// Track ranges in the stream as sent or acked. Acked implies sent. Not in a
 /// range implies needing-to-be-sent, either initially or as a retransmission.
 #[derive(Debug, Default, PartialEq)]
-struct RangeTracker {
+pub struct RangeTracker {
     // offset, (len, RangeState). Use u64 for len because ranges can exceed 32bits.
     used: BTreeMap<u64, (u64, RangeState)>,
 }
@@ -300,7 +300,7 @@ impl RangeTracker {
         }
     }
 
-    fn mark_range(&mut self, off: u64, len: usize, state: RangeState) {
+    pub fn mark_range(&mut self, off: u64, len: usize, state: RangeState) {
         if len == 0 {
             qinfo!("mark 0-length range at {}", off);
             return;
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -148,6 +148,7 @@ pub enum RangeState {
 pub struct RangeTracker {
     // offset, (len, RangeState). Use u64 for len because ranges can exceed 32bits.
     used: BTreeMap<u64, (u64, RangeState)>,
+    acked: u64,
 }
 
 impl RangeTracker {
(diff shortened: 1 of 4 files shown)


jesup:
Needs to be applied on top of bench_coalesce

jesup:
Sorry about the conflicts, I'm still trying to get used to the workflows for github

neqo-transport/src/send_stream.rs
```diff
@@ -269,46 +267,53 @@ impl RangeTracker {
...
             qinfo!("mark 0-length range at {}", off);
             return;
         }
 
         let subranges = self.chunk_range_on_edges(off, len as u64, state);
 
+        let mut first: bool = true;
         for (sub_off, sub_len, sub_state) in subranges {
+            // Check if the first range overlaps the acked-from-zero range and adjust
+            if first {
```
martinthomson, line 292: If you move the checks you add BEFORE `chunk_range_on_edges` you will avoid a bunch of work.

```rust
let end = (off + len).saturating_sub(self.acked);
let off = max(self.acked, off);
if sub_state == RangeState::Acked && if off == self.acked {
  self.acked += len;
  return;
}
if len == 0 {
  return;
}
let subranges = self.check_range_on_edges(off, len, sub_state);
```

neqo-transport/src/send_stream.rs
```diff
@@ -269,46 +267,53 @@ impl RangeTracker {
...
+                first = false;
+                if sub_off < self.acked {
```
martinthomson, line 294: I think that you want

```suggestion
                if sub_off <= self.acked {
```

If `self.acked` is 0, then the first range that will increase the value will start from `0..`

I included that in the above.

neqo-transport/src/send_stream.rs
```diff
@@ -269,46 +267,53 @@ impl RangeTracker {
...
+                    if sub_state != RangeState::Acked {
+                        qinfo!(
+                            "trying to downgrade acked-from-zero range {} to Sent {} {}",
+                            self.acked,
+                            sub_off,
+                            sub_len
+                        );
+                        // in this API misuse case, should we retain any part that goes
+                        // past 'acked'?   perhaps for consistency
```
martinthomson, line 303: I would strive for internal consistency here, if only because of the >= case I mention above.

neqo-transport/src/send_stream.rs
```diff
@@ -269,46 +267,53 @@ impl RangeTracker {
...
     }
 
-    fn mark_range(&mut self, off: u64, len: usize, state: RangeState) {
+    pub fn mark_range(&mut self, off: u64, len: usize, state: RangeState) {
         if len == 0 {
             qinfo!("mark 0-length range at {}", off);
             return;
         }
 
         let subranges = self.chunk_range_on_edges(off, len as u64, state);
```
martinthomson, line 287: `chunk_range_on_edges` needs a bunch of work.  It is inefficient.  It should be the one that inserts into `self.used`; a fresh allocation might be necessary because of the interaction with mutable borrows, but that could be kept local, not externalized like this.

martinthomson (commented):
First of all, I think that you probably need quite a few more specific tests for this.

That is, you have two types of tests:

1. Mark a range as acked (e.g., 0..1000) and then check what happens when various bits of that range interact with it.  Including multiple ranges that fall within it (e.g., 0..10, 20..30).  Also when a ranges overlap the end (900..1100), which needs to include both acked and sent ranges.  You probably also need a check for when there are both internal ranges (which are ignored) and end-overlapping.
2. Mark a range as acked out of order (e.g., 50..100 then 0..50 or 6..10 then 0..3 then 3..6).

It is possible that we have those tests already; I didn't check.

larseggert:
#1594 landed, hence closing this. @jesup, if there is anything left in this PR, please open a new one for any diffs?

PR #1594, opened 2024-01-29
perf: Coalesce better

martinthomson:
@jesup this is a version of your patch that looks worse on the bench, but I think that it is concretely better.

The bench you created sets up with N abutting acknowledged ranges.  The current code coalesces just one of them when it fills in the gap.  This patch coalesces all of them.  The result is a benchmark regression (and a noticeable one) when there are more entries.  But that is the risk of having microbenchmarks.  They don't measure what really matters.  I think that this change will ensure less thrashing.

<details>
<summary>Results</summary>

```
coalesce_acked_from_zero 2 entries
                        time:   [61.129 ns 61.278 ns 61.476 ns]
                        change: [-48.216% -48.086% -47.935%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  4 (4.00%) high severe

coalesce_acked_from_zero 100 entries
                        time:   [710.17 ns 710.94 ns 711.66 ns]
                        change: [+304.53% +306.46% +308.06%] (p = 0.00 < 0.05)
                        Performance has regressed.

coalesce_acked_from_zero 1000 entries
                        time:   [1.5435 µs 1.5450 µs 1.5467 µs]
                        change: [+593.07% +594.29% +595.49%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
```
</details>


martinthomson:
Looking at the build failures here, the windows ones look unrelated, but troubling.  The mac one looks transient as the latest stable toolchain couldn't be downloaded.

larseggert:
Do you maybe need to change `neqo-crypto/Cargo.toml` similarly to `neqo-transport/Cargo.toml`, i.e., add the `bench` stuff? (Just guessing.)

neqo-transport/benches/range_tracker.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut used = build_coalesce(count);
+    c.bench_function(
+        &format!("coalesce_acked_from_zero {count}+1 entries"),
+        |b| {
+            b.iter(|| {
+                let used: &mut RangeTracker = &mut used;
+                used.mark_acked(CHUNK, CHUNK as usize);
+                let tail = (count + 1) * CHUNK;
+                used.mark_sent(tail, CHUNK as usize);
+                used.mark_acked(tail, CHUNK as usize);
```
jesup, line 28: What exactly is this supposed to be doing?  
'used' is modified, and then reused in the modified state in the next iteration, etc.
After the first iteration, it keeps re-marking CHUNK to CHUNK+CHUNK-1 as acked, though they're already acked.
Unfortunately, if you restore the 'used' to the original state, or (re)clone it, that will be included in the time for the benchmark, and given cloning or restoring probably mean allocations, that may swamp the coalesce function.
I have a question in the criterion maintainers about if there's a way to have a per-iteration setup function that's not  included in the measurement

jesup: We could build an vector of N copies of the input, and then limit iterations to N.  I'm not finding any other good options

jesup: https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_batched_ref should do it

larseggert (changes requested):
@martinthomson could we separate the bench from the perf improvement into two different PRs? That would allow to see the performance improvement. (And might also give us some idea of what exactly makes the Windows linking fail.)

larseggert:
@martinthomson when you have a chance, would you resolve the conflicts and get this ready to land?

larseggert (approved):

PR #1596, opened 2024-01-29
Add tc commands for upload testing on Linux.

KershawChang:


larseggert (approved):

larseggert (approved):
You might want to run this through `shellcheck` (or install the VScode plugin.) There are some nits.

larseggert (changes requested):
Looking at this again, I would suggest taking out all the stuff about reading and using a root password. Use `sudo` and assume users have set up `/etc/sudoers` to not always prompt for passwords.

test/upload_test.sh
```diff
@@ -50,13 +52,20 @@ setup_network_conditions() {
...
+            "sudo tc qdisc add dev lo root handle 1: tbf rate $bw_in_bits_per_sec burst $bdp_bytes_rounded_up limit 30000"
+            "sudo tc qdisc add dev lo parent 1:1 handle 10: netem delay ${delay_ms}ms loss ${plr_p}%"
+        )
     fi
 
     for command in "${set_condition_commands[@]}"; do
-        echo $command
-        echo $root_password | sudo -S bash -c "$command"
+        echo "$command"
+        bash -c "$command"
```
larseggert, line 68: Shouldn't this use `sudo`? Also, couldn't you use `eval` rather than starting another shell?

test/upload_test.sh
```diff
@@ -67,20 +76,21 @@ stop_network_conditions() {
...
-        return 0
+        stop_condition_commands=(
+            "sudo tc qdisc del dev lo root"
+        )
     fi
 
-    for command in "${set_condition_commands[@]}"; do
-        echo $root_password | sudo -S bash -c "$command"
+    for command in "${stop_condition_commands[@]}"; do
+        bash -c "$command"
```
larseggert, line 85: Shouldn't this use `sudo`? Also, couldn't you use `eval` rather than starting another shell?

larseggert (changes requested):

test/upload_test.sh
KershawChang: Yes, I can use `eval $command`, but not `sudo eval $command`.
I got this warning from `shellcheck`:
```
Can't use sudo with builtins like eval. Did you want sudo sh -c .. instead?shellcheckSC2232
```

larseggert (approved):

PR #1597, opened 2024-01-29
update .gitignore for nss and nspr folders

KershawChang:
At [here](https://github.com/mozilla/neqo#faster-builds-with-separate-nssnspr) we said that one can clone `nss` and `nspr` in the same directory, so I think we should update `gitignore` to ignore those folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ Cargo.lock
 *.qlog
 *.swp
 /qns/.last-update-*
+/nss
+/nspr
+/dist


larseggert:
I think by "same directory" we mean that they should both go into some _other_ directory (that you would then set `NDD_DIR` to), not the neqo one?

KershawChang:
> I think by "same directory" we mean that they should both go into some _other_ directory (that you would then set `NDD_DIR` to), not the neqo one?

I see. Then, I think we don't need this PR.

PR #1598, opened 2024-01-29
fix: Don't send superfluous PING-only Initial packets during handshake

### LINKED ISSUE #744: Stop sending useless Initial packets on PTO
We send a PING in every PTO packet, but for an Initial packet that's useless.  If we don't have CRYPTO data to send, we shouldn't bother sending a packet.  On Initial PTO we should just send what we have and leave it at that.  I think that it should be enough to not send a PING for Initial packets.

Tests will need to be updated as quite a few assume that PTO always involves two packets.

larseggert:
This limits the Initial packet number space to sending one packet when a PTO fires (other packet number spaces will continue to send two.) This stops PING-only Initial packets during handshake.

Some tests based in the assumption that those PINGs would be sent. Fix those, too. I'd appreciate if someone could esp. double-check the test modifications, esp. to the `idle_caching` test, which is gnarly.

Fixes #744

### DIFF
--- a/neqo-transport/src/connection/tests/idle.rs
+++ b/neqo-transport/src/connection/tests/idle.rs
@@ -296,18 +296,14 @@ fn idle_caching() {
     client.process_input(&handshake.unwrap(), start);
 
     // Perform an exchange and keep the connection alive.
-    // Only allow a packet containing a PING to pass.
     let middle = start + AT_LEAST_PTO;
-    mem::drop(client.process_output(middle));
     let dgram = client.process_output(middle).dgram();
 
     // Get the server to send its first probe and throw that away.
     mem::drop(server.process_output(middle).dgram());
-    // Now let the server process the client PING.  This causes the server
+    // Now let the server process the client packet.  This causes the server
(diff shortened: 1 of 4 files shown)


martinthomson (commented):
This comes at a cost that I want to make sure is understood.

This will increase our exposure to loss if (or more likely: when, considering PQC requirements) multiple Initial packets are needed.  In that case, a single-packet PTO will allow us to detect the loss, but it will not provide repair.  This is true for large first flights and it is true for a TLS HelloRetryRequest that then results in a larger ClientHello than can fit in a single packet.

Right now, we are repairing those packets with the PING.

The PING also helps mitigate the reduction in congestion control window that comes from a loss of a packet, because its acknowledgment causes an increase.

Note that if we're using 0-RTT, the Initial packet won't take much space.  The remainder of the datagram can contain 0-RTT data.

On balance, I'm not sure that we really need to avoid sending a PING.  It's not that wasteful.

martinthomson (approved):
Oh, that's clever.  This is a nice tweak.

PR #1599, opened 2024-01-29
Add key exchange group configuration knobs

jschanck:
Currently, neqo clients list all of the key exchange groups supported by NSS in their supported_groups extension, and this is not configurable. They also send only one key share entry.

The first patch in this PR makes it possible to configure both the supported_groups list and the number of key share entries that are sent. The second patch sets defaults for both. 

According to [telemetry](https://glam.telemetry.mozilla.org/firefox/probe/ssl_kea_ecdhe_curve_full/explore), about 82% of TLS 1.3 connections over TCP negotiate X25519 and 16% negotiate P-256. The remainder use P-384 or P-521. My guess is that QUIC is even more heavily skewed towards X25519 but that it still makes sense to send a P-256 share.

Some additional motivation for this change:

I've recently [added support](https://bugzilla.mozilla.org/show_bug.cgi?id=1871629) for [xyber768d00](https://www.ietf.org/archive/id/draft-tls-westerbaan-xyber768d00-02.html) to NSS, and I'm going to start an [experiment](https://bugzilla.mozilla.org/show_bug.cgi?id=1874959) with it in Firefox. I need the new knobs here to keep xyber768d00 out of supported_groups when it is disabled by pref. Currently I'm relying on an [NSS policy kludge](https://bugzilla.mozilla.org/show_bug.cgi?id=1875158) to do this.

### DIFF
--- a/neqo-crypto/bindings/bindings.toml
+++ b/neqo-crypto/bindings/bindings.toml
@@ -49,6 +49,7 @@ functions = [
     "SSL_PeerSignedCertTimestamps",
     "SSL_PeerStapledOCSPResponses",
     "SSL_ResetHandshake",
+    "SSL_SendAdditionalKeyShares",
     "SSL_SetNextProtoNego",
     "SSL_SetURL",
     "SSL_VersionRangeSet",
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -461,6 +461,16 @@ impl SecretAgent {
         })
     }
 
+    /// Set the number of additional key shares that will be sent in the client hello
+    ///
+    /// # Errors
+    /// If the underlying API fails (which shouldn't happen).
+    pub fn send_additional_key_shares(&mut self, count: usize) -> Res<()> {
+        secstatus_to_res(unsafe {
+            ssl::SSL_SendAdditionalKeyShares(self.fd, c_uint::try_from(count)?)
+        })
+    }
+
     /// Set TLS options.
     ///
     /// # Errors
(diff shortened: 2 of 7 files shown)


martinthomson (commented):
Seems broadly reasonable.

What NSS version was ssl_grp_kem_xyber768d00 added in?  We have an [NSS version check](https://github.com/mozilla/neqo/blob/fca21791a2a0f32671959b08aec9b50c532c9980/neqo-crypto/src/lib.rs#L113) that probably needs updating.

Now, the new code will mostly be covered by virtue of you calling it, but it would be nice to include a test or two.

I suggest a couple:

1. Test that the handshake works when you set an alternative group (P-256 is fine).
2. Test that the handshake works when the client chooses one group, but the server prefers another (i.e., HelloRetryRequest is involved; this is potentially scary in that we haven't really tested it all the way).
3. Test that the handshake works when the client sends shares in two groups and the server picks either (this might be two tests).

Those should all set you up well in terms of being confident that the code is all in place.

jschanck:
Thanks for taking a look. I added some tests.

The github nss-dev/nss mirror seems to have stopped updating a few months ago, so CI is going to fail until we fix that;  ssl_grp_kem_xyber768d00 was added in 3.97.

martinthomson (approved):
OK, pending an update to CI (or the NSS mirror) and a rebase, which I hope won't take long, this LGTM.

larseggert:
If the NSS mirror is flaky, I'll do a PR that pulls the source via `hg`. (But given the recent flakiness of `hg.mozilla.com`, I had hoped we mirror everything to GitHub instead. Sigh.)

larseggert:
@jschanck who is running that NSS mirror? Would be trivial to pull regularly with a GitHub workflow. Can do a PR?

jschanck:
Ben Beurdouche owns it. Apparently the mirroring scripts were updating the `master` branch instead of `main`. He's switched the default branch to `master` for now.

PR #1600, opened 2024-01-30
Don't use lld on windows

martinthomson:
The windows build ignores this attribute, but there is no point in having a warning.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           sudo apt-get install -y --no-install-recommends gyp mercurial ninja-build lld
           curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
-          echo "RUSTFLAGS='-C link-arg=-fuse-ld=lld'" >> "$GITHUB_ENV"
+          echo "RUSTFLAGS=-C link-arg=-fuse-ld=lld" >> "$GITHUB_ENV"
 
       # In addition to installing dependencies, first make sure System Integrity Protection (SIP)
       # is disabled on this MacOS runner. This is needed to allow the NSS libraries to be loaded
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -76,7 +77,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@v1
 
       - name: Install Rust tools
-        run: cargo +${{ matrix.rust-toolchain }} binstall --no-confirm cargo-llvm-cov cargo-nextest
+        run: cargo +${{ matrix.rust-toolchain }} binstall --no-confirm bench cargo-llvm-cov cargo-nextest
```
larseggert, line 80: Did this sneak in from another PR?

.github/workflows/check.yml
```diff
@@ -114,6 +115,9 @@ jobs:
       - name: Run tests and determine coverage
         run: cargo +${{ matrix.rust-toolchain }} llvm-cov nextest --features ci --all-targets --no-fail-fast --lcov --output-path lcov.info
 
+      - name: Benches should at least build
+        run: cargo +${{ matrix.rust-toolchain }} bench --features bench --no-run
+
```
larseggert, line 120: Ditto?

larseggert (commented):
Did something leak in from another PR?

larseggert:
Also see https://www.reddit.com/r/rust/comments/nivznh/using_lld_as_the_linker_on_windows/ for how to use lld on Windows. (It should be installed in the runner somewhere: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md)

.github/workflows/check.yml
martinthomson: Yeah, I'm trying to get benches built, in anticipation of one of the benchmarks landing.  Though I can move that to the other PR.

larseggert: No, leave it in. I just wanted to check that it's deliberate.

larseggert (approved):

Issue #1548, opened 2024-01-11
Don't format packets for dumping if qlog has not been configured to log


jesup:
If the log::max_level() is Debug or Info, we'll spend the time to format packets in dump_packet() even if the qdebug!() will end up as a no-op, since we haven't configured a file for logging.

This can cost us 10% of a core when downloading over h3 on a high end machine.

The log level is a shared resource, so it can get bumped by other modules: https://bugzilla.mozilla.org/show_bug.cgi?id=1874215

mxinden:
Maybe I can help here, though I would need some pointers. Sorry in case I am missing something obvious.

As far as I can tell the `qdebug!` call in `dump_packet` boils down to a call to `init` and `do_log!`:

https://github.com/mozilla/neqo/blob/9e6cd932b140fde141e6414d53f539b6d296d4f9/neqo-common/src/log.rs#L52-L72

https://github.com/mozilla/neqo/blob/9e6cd932b140fde141e6414d53f539b6d296d4f9/neqo-common/src/log.rs#L15-L33

`init` initializes the logger. While one can provide a custom target via `Builder::target` (i.e. an `fmt::Target`, e.g. a file), this is not done here, hence the default is chosen which is logging to `stderr`.

`do_log!` then takes the above initialized logger via the global `::log::logger()` and logs to it, i.e. logs to `stderr`.

> since we haven't configured a file for logging.

How would that file be configured? Are you referring to a pipe of the processe's `stderr` to a file? If so, how could one know within `dump_packet`?

Are you potentially confusing the `qdebug!` call with the similar [`qlog.rs`](https://github.com/mozilla/neqo/blob/main/neqo-common/src/qlog.rs) infrastructure?

Thank you @jesup. Again, sorry in case I am missing something obvious.

larseggert:
I think the issue is that the condition `::log::Level::Debug > ::log::max_level()` is not sufficient to determine whether `qdebug!` will actually produce output.

For example, if I run with `RUST_LOG=info,neqo_transport::cc=debug`, the condition is not satisfied, but the `qdebug!` call still isn't logging anything.

mxinden:
Thank you @larseggert. This was helpful. I created https://github.com/mozilla/neqo/pull/1601.

PR #1601, opened 2024-01-30
fix(transport/dump): check module not max log level and move to conn

mxinden:
This commit contains two changes:

The first change. To avoid expensive packet decoding and string encoding, the `neqo-transport` `dump_packet` function first checks the current log level:

``` rust
if ::log::Level::Debug > ::log::max_level() {
    return;
}
```

This is problematic when e.g. setting `RUST_LOG=info,neqo_crypto=debug`. While `::log::max_level` will return `::log::Level::Debug`, given the `neqo_crypto=debug`, the log level of the `neqo_transport` crate and `dump` module will be `info`. Thus the packet will be de-/encoded, but never logged, given that the `qdebug!` call calls `log::debug!`, which will check the module log level and not use the maximum log level.

Instead, with this commit, `dump_packet` checks the module log level.

``` rust
if !log::log_enabled!(log::Level::Debug) {
    return;
}
```

The second change. `dump_packet` is in `neqo_transport::dump`, but only called in `neqo_transport::connection`.
`RUST_LOG=info,neqo_transport::connection=debug` will thus not dump the packet as `dump_packet` checks its own module's log level, not the log level of its call side. To remove this small footgun, move `dump_packet` to `neqo_transport::connection::dump` and thus enable logging on e.g. `RUST_LOG=info,neqo_transport::connection=debug`.

Note that the `dump` module was never exposed beyond `neqo_transport` and thus this is not a breaking change to other crates.

An alternative, arguably more complex, approach would be to write `dump_packet` as a proc macro.

//CC @jesup @larseggert 

Closes https://github.com/mozilla/neqo/issues/1548.

### DIFF
--- a/neqo-transport/src/connection/dump.rs
+++ b/neqo-transport/src/connection/dump.rs
@@ -27,7 +27,7 @@ pub fn dump_packet(
     pn: PacketNumber,
     payload: &[u8],
 ) {
-    if ::log::Level::Debug > ::log::max_level() {
+    if !log::log_enabled!(log::Level::Debug) {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


larseggert (approved):

larseggert:
Merging this now, since it LGTM. @jesup if this **doesn't** fix it for you, please reopen the issue?

PR #1603, opened 2024-01-30
Do proper congestion control before the first ACK has come in

larseggert:
The small fix here is to `neqo-transport/src/recovery.rs`, to properly declare packets lost when a PTO fires before we have gotten a first ACK.

All the rest of the PR is to fix the various tests that had baked in the assumption that the initial `cwnd` wouldn't shrink before an ACK was received. The easiest way (= requiring the fewest changes) to do this was to insert PING frames into the handshake method used for the tests, causing ACKs to be emitted and preventing the PTO from firing. As a side effect, that also leaves a connection idle after the handshake, leading to simplifications. A few tests did their own force idle and required other fixes.

The `idle_timeout_crazy_rtt` test kept failing. The reason here was that there is a bug in our PTO calculation (in `pto_period_inner`) when the PTO count became large (> 26). I have an ugly, inefficient fix for this in the PR below, but we need a proper one, hence making this a draft PR.

CC @martinthomson

### DIFF
--- a/neqo-transport/src/connection/tests/mod.rs
+++ b/neqo-transport/src/connection/tests/mod.rs
@@ -292,14 +292,13 @@ fn exchange_ticket(
 /// The `handshake` method inserts PING frames into the first application data packets,
 /// which forces each peer to ACK them. As a side effect, that causes both sides of the
 /// connection to be idle aftwerwards. This method simply verifies that this is the case.
-fn assert_idle(client: &mut Connection, server: &mut Connection, rtt: Duration, mut now: Instant) {
+fn assert_idle(client: &mut Connection, server: &mut Connection, rtt: Duration, now: Instant) {
     let idle_timeout = min(
         client.conn_params.get_idle_timeout(),
         server.conn_params.get_idle_timeout(),
     );
-    now -= rtt / 2; // Client started its idle period half an RTT before now.
-    assert_eq!(client.process_output(now), Output::Callback(idle_timeout));
-    now += rtt / 2;
+    // Client started its idle period half an RTT before now.
+    assert_eq!(client.process_output(now.checked_sub(rtt/2).unwrap()), Output::Callback(idle_timeout));
     assert_eq!(server.process_output(now), Output::Callback(idle_timeout));
 }
 
--- a/neqo-transport/src/connection/tests/mod.rs
+++ b/neqo-transport/src/connection/tests/mod.rs
@@ -34,6 +34,8 @@ use neqo_common::{event::Provider, qdebug, qtrace, Datagram, Decoder, Role};
 use neqo_crypto::{random, AllowZeroRtt, AuthenticationStatus, ResumptionToken};
 use test_fixture::{self, addr, fixture_init, new_neqo_qlog, now};
 
+use enum_map::enum_map;
+
 // All the tests.
 mod ackrate;
 mod cc;
@@ -153,11 +155,11 @@ pub fn maybe_authenticate(conn: &mut Connection) -> bool {
     false
 }
 
-/// Compute the RTT variance after `n` ACKs (or other RTT updates).
-pub fn rttvar_after_n_acks(n: usize, rtt: Duration) -> Duration {
+/// Compute the RTT variance after `n` ACKs or other RTT updates.
+pub fn rttvar_after_n_updates(n: usize, rtt: Duration) -> Duration {
     assert!(n > 0);
     let mut rttvar = rtt / 2;
-    for _ in 0..n - 1 {
+    for _ in 1..n {
         rttvar = rttvar * 3 / 4;
     }
     rttvar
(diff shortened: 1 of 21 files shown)


neqo-transport/src/connection/tests/mod.rs
```diff
@@ -151,6 +155,25 @@ pub fn maybe_authenticate(conn: &mut Connection) -> bool {
...
+    assert!(n > 0);
+    let mut rttvar = rtt / 2;
+    for _ in 1..n {
+        rttvar = rttvar * 3 / 4;
+    }
+    rttvar
+}
+
+/// This inserts a PING frame into packets.
+pub struct Ping {}
```
martinthomson, line 169: Maybe avoid collisions with other potential uses of Ping (like on the Frame enum).

```suggestion
pub struct PingWriter {}
```

neqo-transport/src/connection/tests/mod.rs
```diff
@@ -247,51 +289,26 @@ fn exchange_ticket(
...
         server.conn_params.get_idle_timeout(),
     );
-    assert_eq!(client.process_output(now), Output::Callback(idle_timeout));
-    now += rtt / 2;
+    // Client started its idle period half an RTT before now.
     assert_eq!(
-        server.process(ack.as_dgram_ref(), now),
+        client.process_output(now.checked_sub(rtt / 2).unwrap()),
         Output::Callback(idle_timeout)
     );
```
martinthomson, line 304: I think that you should subtract the half RTT from the expectation, rather than calling process_output in (what might be) the past.

There is a potential problem with rounding errors (if the RTT is an odd number of nanoseconds), but I think we're OK in that regard, because the test rounds the RTT down when it increments the time.

neqo-transport/src/connection/tests/mod.rs
```diff
@@ -170,10 +193,29 @@ fn handshake(
...
+        let should_ping = !did_ping[a.role()]
+            && (a.role() == Role::Client && a.tls_info().is_some()
+                || (a.role() == Role::Server && b.state() == &State::Connected));
+        if should_ping {
+            a.test_frame_writer = Some(Box::new(Ping {}));
+        }
         let output = a.process(input.as_ref(), now).dgram();
+        if should_ping {
+            a.test_frame_writer = None;
+            did_ping[a.role()] = true;
```
martinthomson, line 217: Let me see if I can work out your logic a bit better, because it doesn't seem like you need to ping on both peers for this to work.

I think that you can just have the server reach Confirmed and then force a PING into the packet it sends then.  The client will acknowledge that immediately, but have nothing further to send.

This arrangement has the client send a PING as well as soon as the connection is complete.  That happens when the state is Connected at the TLS layer, which corresponds with the Connected state for QUIC (I don't know why you need to check the TLS layer).  The client is already sending a TLS Finished as a Handshake packet at that point, which causes the server to send HANDSHAKE_DONE, so you don't need to do anything to get the server to respond immediately.

neqo-transport/src/recovery.rs
```diff
@@ -412,7 +412,7 @@ impl LossRecoverySpace {
             .sent_packets
             .iter_mut()
             // BTreeMap iterates in order of ascending PN
-            .take_while(|(&k, _)| Some(k) < largest_acked)
+            .take_while(|(&k, _)| largest_acked.is_none() || Some(k) < largest_acked)
```
martinthomson, line 415: This might be easier to grok.

```suggestion
            .take_while(|(&k, _)| k < largest_acked.unwrap_or(PacketNumber::MAX))
```

martinthomson (commented):
I think that #1605 will solve our problem with the PTO going to zero.  I've also suggested a pretty big simplification to the test.

Overall though, this is a nice improvement.

neqo-transport/src/connection/tests/mod.rs
larseggert: What I saw was the client sending a bunch of new connection IDs to the server which remained unACK'ed (due to ACK delay being in effect). This change forces a PING into that packet, to trigger a server ACK.

martinthomson: Ah, that's right.  I forget that I added that feature.  Won't it be enough for the client to do that then?  You can want for Confirmed and send a ping with that.

larseggert: I'll try. (I think i tried to do it client-only and there were some more test failures.)

larseggert: I don't think that works, because the `HandshakeDone` doesn't elicit an immediate ACK. So when the `handshake()` methid is done, the client still has the delayed ACK timer pending and is not idle. 

There is a case to be made to send an immediate ACK on `HandshakeDone`, which would fix this.

larseggert: (Well, it would fix it for a subset of tests, but many are still failing with a client-only PING and an immediate ACK on HandshakeDone.)

larseggert: OK to merge this as-is?

martinthomson: Ah yes.  It's a bit ugly having the check for "Connected" on the client always, but I think you have it right.  We can't rely on the server Connected signal and the server Confirmed happens after the ping.  Thanks for checking.

martinthomson (approved):

Issue #1535, opened 2024-01-03
`mio-extras` not being updated for mio>0.6


larseggert:
See https://github.com/dimbleby/mio-extras:
> This library is not compatible with mio 0.7. However, there's a similar project named [mio-misc](https://github.com/onurzdg/mio-misc) that works with mio 0.7.

I haven't checked whether and how `mio-misc` could be used to provide `Timer` functionality for `neqo-server.`

See #1511

mxinden:
@larseggert how about moving away from `mio` and instead use the higher-level [`tokio`](https://docs.rs/tokio/latest/tokio/) in `neqo-server` and `neqo-client`? Note that `tokio` is based on `mio`.

With my limited knowledge of `neqo-*`, I consider `tokio` the better fit for the problem. It has support for UDP sockets and timers and, given its higher level abstractions, is easier to use.

In case there is interest, I am happy to give it a shot and propose a pull request.

Related: https://github.com/mozilla/neqo/issues/530

larseggert:
I'm running into issues trying to use quinn-udp for ECN packet I/O with the client and server code based on that old version of mio, so I'm even more in favor of tokio now.

CC @martinthomson @KershawChang

KershawChang:
> I'm running into issues trying to use quinn-udp for ECN packet I/O with the client and server code based on that old version of mio, so I'm even more in favor of tokio now.
> 
> CC @martinthomson @KershawChang

+1 for using `tokio`.
I think the reason why we used `mio-extras` is that we want to use `mio_extras::timer`, but I assume we should be able to find an alternative for it.

mxinden:
I have a draft in https://github.com/mozilla/neqo/pull/1581 with a basic client/server interaction working. Still needs more work. Will ping folks once it is ready. Not sure I will get to it this week though.

> I'm running into issues trying to use quinn-udp for ECN packet I/O with the client and server code based on that old version of mio, so I'm even more in favor of tokio now.

Yes. I expect the move to `quinn-udp` to be easy once https://github.com/mozilla/neqo/pull/1581 is merged. Happy to do that as a follow-up.

larseggert:
I have some draft code for quinn-udp in https://github.com/larseggert/neqo/tree/feat-ecn-io, if you want to make that part of this PR.

mxinden:
> I have some draft code for quinn-udp in https://github.com/larseggert/neqo/tree/feat-ecn-io, if you want to make that part of this PR.

[...]

PR #1604, opened 2024-01-30
refactor(client,server): use quinn_udp for I/O

mxinden:
Use `quinn_udp` for socket I/O, laying the groundwork for ECN, enabling one to send/receive TOS.

---

Builds on top of

- https://github.com/mozilla/neqo/pull/1612
- https://github.com/mozilla/neqo/pull/1581
- https://github.com/larseggert/neqo/tree/feat-ecn-io

Discussed in https://github.com/mozilla/neqo/issues/1535.

//CC @larseggert as this might already unblock you for testing.


larseggert:
Let's get this merged once the deps are merged. I can resume the ECN work then.

mxinden:
Sounds good @larseggert. Currently at FOSDEM, thus limited bandwidth. Will continue https://github.com/mozilla/neqo/pull/1612 on Monday.

mxinden:
I resolved the merge conflicts due to https://github.com/mozilla/neqo/pull/1581 and https://github.com/mozilla/neqo/pull/1612.

I will refactor `neqo-client` to use `quinn-udp` next.

mxinden:
With the latest commit `neqo-server` and `neqo-client` now use `quinn-udp` to send and receive. A simple GET succeeds (`200`).

Note that there are still lots of cleanups to be done. Thus will keep this in Draft for now.

neqo-common/src/udp.rs
```diff
@@ -0,0 +1,155 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![allow(clippy::missing_errors_doc)]
+#![allow(clippy::missing_panics_doc)]
```
mxinden, line 8: I don't think these docs would be helpful, since each of the functions just delegate to `tokio` and `quinn-udp`. I doubt it would ever be complete, nor up-to-date for long. Happy to add it in case you consider it helpful.

mxinden (commented):
This is ready for a first review.

neqo-common/src/udp.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{IpTos, IpTosDscp, IpTosEcn};
+
+    #[tokio::test]
+    async fn datagram_tos() -> Result<(), io::Error> {
```
larseggert, line 124: This test doesn't seem to terminate on Windows.

larseggert (changes requested):
We need to figure out why the `datagram_tos` test isn't terminating on Windows. I have a hunch this is related to (non-)blocking sockets.

neqo-common/src/udp.rs
martinthomson: Can you put that text (or a shorter version thereof) in a comment?

neqo-common/src/udp.rs
martinthomson: You can see on lines 138 and 141 that there is a `.await`.  That means that this needs to be async.  Or rather, it is easier to use `#[tokio::test]` rather than `#[test]` so that this runs faster.

As for the termination, it's possible that port 8080 is taken.  I believe that you can bind the receiver to a random/ephemeral port instead, which might help this complete.  No idea whether that is the problem though.

mxinden:
Sorry for the silence here. Have been traveling and been sick. I will have a lot more time starting Tuesday.

Thank you for the input thus far.

larseggert:
No stress. I am on PTO the entire week.

neqo-common/src/udp.rs
mxinden: Found the problem. Setting ECN CE is not allowed on Windows.

With https://github.com/mozilla/neqo/pull/1604/commits/e9ba397891451563e4c5674a0949ee060e8a83d7 we are now using ECN ECT1 for testing.

> PASS [   0.236s] neqo-common udp::tests::datagram_tos

---



> As for the termination, it's possible that port 8080 is taken. I believe that you can bind the receiver to a random/ephemeral port instead, which might help this complete. No idea whether that is the problem though.

Thanks Martin. Wasn't the issue here. Still, good idea. Changed with https://github.com/mozilla/neqo/pull/1604/commits/8ced9d9b97e39fb7e8d5e8d01c251f54384c7eda.

mxinden:
@larseggert @martinthomson this is ready for another review. Mind taking a look (once you are back from PTO)?

larseggert (approved):

PR #1605, opened 2024-01-30
Cap the PTO count when working out the period

### LINKED ISSUE #1602: `pto_period_inner` can return `0ns`
When `pto_count` is `30` or `31`, `pto_period_inner` computes a PTO of `0ns`. That seems incorrect. (Higher and lower values produce good output.)

The original code (before #1236) did not have this issue, FWIW: https://github.com/mozilla/neqo/pull/1236/files#diff-69602dcab54959a67ff71f9fbdcd82eed8f15dee641f913ef12ffe8c59ad7f73L823-L824

(This happens during the `idle_timeout_crazy_rtt` test.)

martinthomson:
The previous code would overflow and produce a 0ns timeout when the PTO count got large (well over 24).  This is because the `fast_pto` value tends to have trailing zero bits and shift values between 32-(number of zeros) and 31 (inclusive) would still shift left.

This code instead stops increasing the PTO once it hits 24.  That should be enough for most cases.

Closes #1602.

### DIFF
--- a/neqo-transport/src/recovery.rs
+++ b/neqo-transport/src/recovery.rs
@@ -848,11 +848,7 @@ impl LossRecovery {
         // where F = fast_pto / FAST_PTO_SCALE (== 1 by default)
         let pto_count = pto_state.map_or(0, |p| u32::try_from(p.count).unwrap_or(0));
         rtt.pto(pn_space)
-            .checked_mul(
-                u32::from(fast_pto)
-                    .checked_shl(pto_count)
-                    .unwrap_or(u32::MAX),
-            )
+            .checked_mul(u32::from(fast_pto) << min(pto_count, u32::BITS - u8::BITS))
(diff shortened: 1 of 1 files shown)


larseggert (approved):

larseggert:
This change ends up not being sufficient to fix the `idle_timeout_crazy_rtt` test with #1603.

With the change in this PR, the PTO now doubles until it reaches 5033164.8s, which is ~8.32 weeks. The test runs with an RTT of 150 weeks and 10% loss rate, meaning it never terminates.

I think the way forward here is to make the RTT used for `idle_timeout_crazy_rtt` a little less crazy, so that the max. PTO is still sufficient. I'll incorporate that change into #1603 - yell if you disagree.

martinthomson:
Nah, that seems fine.  We could use u64 for this, but that would mean that `Duration::checked_mul` isn't an option.

PR #1606, opened 2024-01-31
ci: Clone NSS from hg and not the GitHub mirror

larseggert:
Because the latter can be outdated. This is why #1081 is failing CI.


mxinden (approved):
:+1: 

For the sake of consistency, should we do the same in the `qns/Dockerfile`? I.e. revert parts of https://github.com/mozilla/neqo/commit/fca21791a2a0f32671959b08aec9b50c532c9980#diff-8c454c011b5d6322625b9a9f9c00c1791ae2818fa2d07c1edb32fc1a8e6eec49L29-R24?

larseggert:
Probably. I'll add that to this PR.

PR #1615, opened 2024-02-02
rollback qlog to 0.9.0

KershawChang:
Please see https://phabricator.services.mozilla.com/D200303#inline-1111249
I think we can use `qlog 0.9.0` temporarily  until this [PR](https://github.com/cloudflare/quiche/pull/1719) to be merged.


KershawChang:
Based on this [comment](https://phabricator.services.mozilla.com/D200303#inline-1111317), we don't need this PR.

PR #1616, opened 2024-02-02
feat: Use the simulator as a benchmark

martinthomson:
This is probably a bit much for a benchmark, but it turns out to be a pretty good one.

This only measures the transfer part, not connection establishment.  So @jesup might be interested in this as a means of testing upload performance.

I don't think that it is possible to isolate just one peer in this test, but I guess we could in theory try.


larseggert:
LGTM, but I see pretty high variability run-to-run. Do I understand this correctly that you run with 0ms RTT and 10ms jitter? That's a lot (relatively).

martinthomson:
Yes, there will be high variability, but that can be cut by fixing the seed.

There is an additional delay added into the uplink and downlink components.  That is 50ms each way, I think, which makes the jitter much less proportionally.  It could probably be less, but the idea is to stress the send and receive logic more than it is to test that we can drive up throughput under "realistic" conditions.

larseggert:
For use as a test, that variability is nice. For use as a bench, it would be good if we could pin the random seed so a particular test case is generated that exhibits a bunch of "interesting" events, and use that to track performance regressions.

martinthomson:
Hmm, the question is whether averaging multiple runs with the included variance is better than picking one run and maybe missing some variance.  I guess the answer might genuinely be "why not both?"  What do you think?

larseggert:
I think Hyperion is already doing multiple runs w/averaging, but if you run it multiple times there are still big variations.

However, those also happen with @jesup's bench, so maybe we generally need to find better parameters for the Hyperion benches to make the results more stable.

larseggert (approved):

PR #1619, opened 2024-02-05
ci: Build docker image for multiple architectures

larseggert:


larseggert:
@martinthomson any reason we're not building a new image every time `main` is updated? Would make it easier to run qnd locally with up-to-date images.

mxinden:
@larseggert past comment by Martin on the topic: https://github.com/mozilla/neqo/pull/1554#discussion_r1452833839

I am in favor of a new image on each push to `main`.

larseggert:
@mxinden any idea why https://github.com/mozilla/neqo/pkgs/container/neqo-qns under "OS / Arch" currently shows an `unknown/unknown` second image? Something wrong with the meta tags maybe?

larseggert:
> @martinthomson any reason we're not building a new image every time `main` is updated? Would make it easier to run qnd locally with up-to-date images.

Ah, nevermind! We have `workflow_dispatch`, so I can just have it build the image manually.

PR #1623, opened 2024-02-06
Test benchmarks

martinthomson:
Remove separate bench build, just run the benchmarks as tests, which works, without the performance hit.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -117,9 +117,6 @@ jobs:
       - name: Run tests and determine coverage
         run: cargo +${{ matrix.rust-toolchain }} llvm-cov nextest --features ci --all-targets --no-fail-fast --lcov --output-path lcov.info
 
-      - name: Benches should at least build
-        run: cargo +${{ matrix.rust-toolchain }} build --features bench --benches
-
       - name: Run client/server transfer
         run: |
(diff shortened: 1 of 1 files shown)


larseggert (approved):

.github/workflows/check.yml
```diff
@@ -109,20 +109,17 @@ jobs:
...
-        run: cargo +${{ matrix.rust-toolchain }} llvm-cov nextest --features ci --all-targets --no-fail-fast --lcov --output-path lcov.info
-
-      - name: Benches should at least build
-        run: cargo +${{ matrix.rust-toolchain }} build --features bench --benches
+        run: cargo +${{ matrix.rust-toolchain }} llvm-cov nextest \
+           --all-targets --features ci,bench --no-fail-fast --lcov --output-path lcov.info
 
       - name: Run client/server transfer
         run: |
-          cargo +${{ matrix.rust-toolchain }} build --bin neqo-client --bin neqo-server
```
larseggert, line 125: Removing this seems to make the benchmark hang in some cases.

larseggert (approved):

.github/workflows/check.yml
martinthomson: Oh, that's because what happens is that the build needs to run for these targets.  We should probably add `--feature ci,bench` here so that it can use the previous build.

PR #1624, opened 2024-02-06
fix: Don't call `write_entry` from inside `debug_assert!`

larseggert:
And also run release tests during CI (one one platform and Rust toolchain).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,6 +23,8 @@ jobs:
           - os: ubuntu-latest
             rust-toolchain: stable
             type: release
+    env:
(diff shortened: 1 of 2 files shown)


martinthomson (dismissed):
That's a pretty stoopid thing for me to have done.  Hopefully this doesn't get tagged `#[must_use]`

larseggert:
We all have our moments :-)

I for example can't figure out why the release transfer is failing in CI...

larseggert:
Hm. It looks like neqo-crypto links dynamically if the `DEBUG` env variable is set, and statically when it's not. No idea if that matters here.

PR #1627, opened 2024-02-06
fix: More green for the neqo fields in the interop runner

larseggert:
* Only allow v1 for most interop runner runs

### DIFF
--- a/neqo-client/src/main.rs
+++ b/neqo-client/src/main.rs
@@ -1005,6 +1005,8 @@ async fn main() -> Res<()> {
     let mut args = Args::from_args();
 
     if let Some(testcase) = args.qns_test.as_ref() {
+        // Only allow v1 for QNS tests.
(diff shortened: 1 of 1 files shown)


mxinden:
Am I correctly assuming that this is needed in runs where `neqo-client` is matched with a server other than `neqo-server`?

larseggert:
Not only there. A bunch of other servers will do compatible version upgrades with a neqo client in cases where the test runner doesn't expect this. Like [picoquic](https://interop.seemann.io/logs/2024-02-05T08:08/picoquic_neqo/handshake/output.txt) or [haproxy](https://interop.seemann.io/logs/2024-02-05T08:08/haproxy_neqo/handshake/output.txt). (The giveaway is `Wrong version. Expected 0x1, got ['0x6b3343cf']` at the end of an interop runner log file.

mxinden (approved):
For what my review is worth.

martinthomson (approved):

PR #1630, opened 2024-02-07
These addresses can now be constants

martinthomson:


martinthomson:
As far as I can tell, the coverage report is wrong.  However, that const fn is only invoked by the *compiler* now, so it might be that it just isn't being captured.

larseggert (approved):

PR #1632, opened 2024-02-07
ci: Use mozilla-actions/sccache-action

larseggert:
Hm. This doesn't seem to speed things up at all...


martinthomson (approved):
If I'm reading this right, the performance gain here is pretty significant.

The last build on main was:

image

This build was:

image

Maybe that reduction isn't as much as you might have liked, but it is still a win.

larseggert:
Enabling this now - at least it doesn't seem to hurt.

PR #1633, opened 2024-02-07
chore: Add a `[workspace.package]` section and refer to it from members

larseggert:
This means we only need to change stuff in one place from now on. Also eliminates the need for the `update_version` script, so remove that.


mxinden (approved):
:+1:

clippy.toml
```diff
@@ -1 +0,0 @@
-msrv = "1.70.0"
```
martinthomson, line 1: Just confirming: clippy will look at `workspace.package.rust-version` instead?  That's great.

martinthomson (approved):

clippy.toml
larseggert: It seems to, if there is a `rust-version.workspace = true` in the `Cargo.toml` of the member package.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

larseggert: It seems to:

> [msrv](https://doc.rust-lang.org/clippy/lint_configuration.html#msrv)
> The minimum rust version that the project supports. Defaults to the rust-version field in Cargo.toml

https://doc.rust-lang.org/clippy/lint_configuration.html#msrv

larseggert:
@martinthomson related question: Is there any point in updating `authors`, e.g., to `necko@mozilla.com`?

martinthomson:
Yeah, it's probably time to do that.

larseggert:
OK, I'll roll that into this one and land it.

PR #1635, opened 2024-02-08
chore: Drop patch versions from deps

larseggert:
And more clearly mark those needing to be in sync with Firefox.


martinthomson (approved):
Did you script this one or just do it manually?

larseggert:
Manual. I was hoping someone had a tool to limit reps by some other Cargo.lock, but I wasn't able to find one.

PR #1636, opened 2024-02-09
Remove some TimeTravel error cases

martinthomson:
This might make our code more robust.


larseggert:
@martinthomson tests failing, and not because of clap, etc.

larseggert (approved):

PR #1638, opened 2024-02-09
Avoid allocation in random()

martinthomson:
This makes the function take its argument as a const generic argument, which allows the allocation to be performed on the stack.

There are a few cases where we need to do in-place randomization of a different type of object (`SmallVec` in a few places) so I've also exposed an in-place mutation function.

Next step is to add some caching or something along those lines.


larseggert (approved):

larseggert:
@martinthomson conflicts arose after landing #1640

martinthomson:
That's OK, this builds on that one.

PR #1641, opened 2024-02-09
ci: Add benchmark run

larseggert:
Work in progress


larseggert:
Merging this now, because apparently the workflow needs to exist on the `main` branch before it is triggered. So I can't debug it just inside this PR.

PR #1644, opened 2024-02-11
Add a wrapper for slice::from_raw_parts

martinthomson:
And add a lint to prevent the method from being invoked directly.

The documentation on the wrapper explains everything.

I'm a little surprised that this hasn't bitten us before.


martinthomson:
That test failure is an interesting one - it seems to have struck a seed that causes the simulation to do something bad.  But without the seed (see #1645), that will be very hard to debug.

larseggert (approved):

PR #1647, opened 2024-02-12
ci: Generate perf profile for transfer bench

larseggert:
This runs the `transfer` bench inside Linux `perf` and exports the data (raw and converted for [profiler.firefox.com](https://profiler.firefox.com/)) to enable analysis.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Perf transfer benchmark
         run: |
           nice -n -20 taskset -c 0 \
-            perf record -F997 --call-graph=lbr \
+            perf record -F997 --call-graph=lbr -o perf.data \
             cargo +$TOOLCHAIN bench --features ci,bench --bench transfer
 
       # Enable turboboost, hyperthreading and use powersave governor.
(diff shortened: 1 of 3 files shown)


.github/workflows/bench.yml
```diff
@@ -43,13 +43,20 @@ jobs:
...
           nice -n -20 taskset -c 0 \
             cargo +$TOOLCHAIN bench --features ci,bench | tee output.txt
 
-      # Enable turboboost, hyperthreading and use powersave governor.
+      # Pin the transfer benchmark to core 0 and run it at elevated priority inside perf.
+      - name: Perf transfer benchmark
+        run: |
+          nice -n -20 taskset -c 0 \
+            perf record -F997 --call-graph=lbr -o perf.data \
+            cargo +$TOOLCHAIN bench --features ci,bench --bench transfer
```
martinthomson, line 57: You should add `--exact` when matching, just so that you don't pick up other stuff.

I recommend `cargo bench -F bench --bench transfer -- --exact "Run multiple transfers with varying seeds"`

larseggert: Ah, thanks! I didn't know one could match on that long string.

PR #1648, opened 2024-02-12
ci: Benchmark without a qlog

larseggert:
To remove those overheads from the profile.

This PR also rolls in some changes to the bencher, i.e., to make it generate flamegraphs and make them and the raw perf/criterion data available as build artifacts.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -93,10 +93,6 @@ impl NeqoQlog {
         F: FnOnce(&mut QlogStreamer) -> Result<(), qlog::Error>,
     {
         if let Some(inner) = self.inner.borrow_mut().as_mut() {
-            // TODO: Hack.
-            if inner.qlog_path == Path::new("") {
-                return;
-            }
             if let Err(e) = f(&mut inner.streamer) {
                 crate::do_log!(
                     ::log::Level::Error,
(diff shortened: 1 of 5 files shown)


larseggert:
Need to check the flamegraph whether this actually makes a difference.

test-fixture/src/lib.rs
```diff
@@ -392,12 +392,16 @@ impl ToString for SharedVec {
...
 #[must_use]
 pub fn new_neqo_qlog() -> (NeqoQlog, SharedVec) {
-    let mut trace = new_trace(Role::Client);
-    // Set reference time to 0.0 for testing.
-    trace.common_fields.as_mut().unwrap().reference_time = Some(0.0);
     let buf = SharedVec {
         buf: Arc::default(),
     };
+
+    #[cfg(feature = "bench")]
```
martinthomson, line 399: Based on prior experience here, I would use this instead:

```rust
if cfg!(feature = "bench") {
    return (NeqoQlog::disabled(), buf);
}
```

Otherwise the compiler will complain unnecessarily.

martinthomson (approved):

neqo-common/src/qlog.rs
```diff
@@ -93,6 +93,10 @@ impl NeqoQlog {
         F: FnOnce(&mut QlogStreamer) -> Result<(), qlog::Error>,
     {
         if let Some(inner) = self.inner.borrow_mut().as_mut() {
+            // TODO: Hack.
+            if inner.qlog_path == Path::new("") {
+                return;
+            }
```
larseggert, line 99: @KershawChang said in Slack `self.inner` should be `None` in the case where `self` is `NeqoQlog::disabled()`, but that doesn't seem to be the case?

But we need a better check here than the current hack above.

PR #1657, opened 2024-02-15
Track sent items using SmallVec

martinthomson:
DO NOT MERGE THIS - IT MAKES THINGS SLOWER

I thought that this was worth sharing though.

I was looking at the profiles that @KershawChang shared and noticed that `write_stream_frame` was eating a decent amount of CPU because there was an allocation in there.  It took me a while to track it down, but it turns out that this is the first place that we add a `RecoveryToken` to the `Vec` of those that we hold in each `SentPacket`.

The obvious thing to test in this case is `SmallVec`.  `SmallVec` is designed for places where you mostly add a few items to a collection.  It holds a small array into which values are deposited unless more than the array size is added, when those values are allocated on the heap, just like a normal `Vec`.  This means that if you stay within the array, you avoid a heap allocation, which can save a bunch of work.  The cost is a little complexity, the extra stack allocation, and that your stuff needs to move if you ever overflow the backing array.  So if your prediction is wrong, it is more expensive.

I tried this out and it makes performance worse.  With a backing array of size 1, the performance change isn't observable.  But larger values progressively degrade performance.  Increasing the array size to 8 makes it worse than the value of 4 I started with.  My theory is that the increased size of the base object starts is the problem.  This collection is typically kept in a `SentPacket`.  Our acknowledgements code needs extra time to move the larger struct around.

The size of the performance regression suggests that there are some ways we could further improve how `SentPacket` is handled.

<details>
<summary>Benchmark results</summary>

```
Run multiple transfers with varying seeds
                        time:   [42.395 ms 42.547 ms 42.712 ms]
                        change: [+3.3754% +4.3870% +5.2776%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  1 (1.00%) high severe

Run multiple transfers with the same seed
                        time:   [42.513 ms 42.584 ms 42.657 ms]
                        change: [+3.4443% +3.7777% +4.0976%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
```
</details>


martinthomson:
@larseggert, I'm looking at the profiles that were generated here and it looks like these are off the `main` branch.  The flamegraph for transfer shows calls into `BTreeMap`, but I don't see any way that could happen.

larseggert:
I can't get to it this week. Wonder if the checkout action needs another parameter?

larseggert:
Let's see what the new flamegraphs show (also because LTO is now enabled.)

larseggert:
It might be worthwhile revisiting this now that @mxinden made some improvements related to `SentPacket`.

martinthomson:
<details>
<summary>It still looks pretty bad</summary>

```
RxStreamOrderer::inbound_frame()
                        time:   [51.698 ms 51.889 ms 52.085 ms]
                        change: [+4.3290% +5.1052% +5.9211%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild

     Running benches/sent_packets.rs (target/release/deps/sent_packets-a9a340a4dfd0be33)
SentPackets::take_ranges
                        time:   [12.966 µs 13.264 µs 13.516 µs]
                        change: [+226.83% +258.23% +293.90%] (p = 0.00 < 0.05)
                        Performance has regressed.

     Running benches/transfer.rs (target/release/deps/transfer-8555e2732eb99220)
transfer/pacing-false/varying-seeds
                        time:   [14.021 ms 14.244 ms 14.462 ms]
                        change: [+5.1022% +7.4030% +9.5750%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) low mild

transfer/pacing-true/varying-seeds
                        time:   [14.828 ms 15.058 ms 15.281 ms]
                        change: [+4.0012% +6.3270% +8.5372%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) low mild

transfer/pacing-false/same-seed
                        time:   [12.132 ms 12.391 ms 12.643 ms]
                        change: [+4.6848% +7.8215% +10.869%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild

transfer/pacing-true/same-seed
                        time:   [13.924 ms 14.178 ms 14.432 ms]
                        change: [+4.0431% +6.9063% +9.9521%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

</details>

larseggert:
Closing in favor of #2779, which was rebased.

PR #1659, opened 2024-02-15
ci: Fix cache updates

larseggert:


larseggert:
This needs to run on `main` to be effective, so can't tell whether it works yet.

PR #1662, opened 2024-02-15
Sent packets vec

martinthomson:
This is NOT faster than main and I don't quite understand why.

I'm putting this up so that I can get benchmarks (my machine doesn't produce anything useful).

If this line of investigation doesn't work out, I might try to salvage some of the cleanup that I did here.


larseggert:
I don't understand why CI isn't running here. Force push to unstick it?

larseggert:
If I had to venture a guess why it's slower, it'd be because a `BTreeMap` holds a bunch of items contiguos in memory, where as a `VecDeque` has each item in a different location in memory. So the `BTreeMap` shoudl traverse faster.

larseggert:
Argh! I followed the steps GitHub showed to merge main into this branch. Not sure what it did.

larseggert:
Reverted and re-extracted the rebased PR into #1761. Sorry for the mess.

PR #1678, opened 2024-02-26
feat: Send and process ACK-ECN

larseggert:
The remaining bits from #1495

The remaining todo item after this PR is to actually act on incoming CE marks, i.e., trigger a congestion control action.


larseggert:
Making this a draft PR again until I can add some tests.

mxinden (approved):

larseggert:
@KershawChang @martinthomson when you have a chance, please review?

neqo-transport/src/path.rs
```diff
@@ -979,6 +1007,25 @@ impl Path {
...
+            self.lost_ecn_bytes += lost_packets.iter().map(|p| p.size).sum::<usize>();
+
+            // If we lost more than 3 MTUs worth of ECN-marked bytes, then
+            // disable ECN on this path. See RFC 9000, Section 13.4.2.
+            // This doesn't quite implement the algorithm given in RFC 9000,
+            // Appendix A.4, but it should be OK. (It might be worthwhile caching
+            // destination IP addresses for paths on which we had to disable ECN,
+            // in order to not persitently delay connection establishment to
+            // those destinations.)
+            if self.lost_ecn_bytes > MAX_DATAGRAM_SIZE * 3 {
```
KershawChang, line 1023: It'd be good to have a test for this behavior.

neqo-transport/src/path.rs
```diff
@@ -573,14 +578,37 @@ impl Path {
             challenge: None,
             rtt: RttEstimate::default(),
             sender,
-            tos: IpTos::default(), // TODO: Default to Ect0 when ECN is supported.
-            ttl: 64,               // This is the default TTL on many OSes.
+            tos: IpTosEcn::Ect0.into(),
```
KershawChang, line 581: Do we need to make `ecn_enabled` configurable by a gecko pref?
I assume the risk of this should be low, but I am not super sure.

KershawChang (commented):
Overall, this looks good to me.

neqo-transport/src/path.rs
larseggert: Not sure. I assume for TCP-based networking, we are letting the OS decide what to do about ECN?

neqo-transport/src/path.rs
larseggert: Good point, will add.

neqo-common/src/datagram.rs
```diff
@@ -90,7 +94,6 @@ fn fmt_datagram() {
     let d = datagram([0; 1].to_vec());
     assert_eq!(
         format!("{d:?}"),
-        "Datagram IpTos(Cs0, NotEct) TTL Some(128) [fe80::1]:443->[fe80::1]:443: [1]: 00"
-            .to_string()
+        "Datagram IpTos(Cs0, Ect0) TTL Some(128) [fe80::1]:443->[fe80::1]:443: [1]: 00".to_string()
```
martinthomson, line 97: Rather than call to_string(), you should compare &format!(...) with the &str.  It saves an allocation.  (In tests, no big deal, but it's a good habit.)

neqo-transport/src/connection/mod.rs
```diff
@@ -1501,7 +1501,16 @@ impl Connection {
...
                         match self.process_packet(path, &payload, now) {
-                            Ok(migrate) => self.postprocess_packet(path, d, &packet, migrate, now),
+                            Ok(migrate) => {
+                                // Since we processed frames from this IP packet now,
+                                // update the ECN counts (RFC9000, Section 13.4.1).
+                                if let Some(space) = self.acks.get_mut(space) {
+                                    space.inc_ecn_count(d.tos().into());
+                                } else {
+                                    qdebug!("Not tracking ECN for dropped packet number space");
+                                }
```
martinthomson, line 1511: Is this the right place for the change?  postprocess_packet() seems better.

neqo-transport/src/connection/mod.rs
```diff
@@ -2901,6 +2919,43 @@ impl Connection {
...
+            if sum_inc < newly_acked {
+                qwarn!(
+                    "ACK had {} new marks, but acked {} packets, disabling ECN",
+                    sum_inc,
+                    newly_acked
+                );
+                path.borrow_mut().disable_ecn();
+            }
+            path.borrow_mut().set_ecn_count(space, ecn_count);
+        }
```
martinthomson, line 2957: This is a method.  Extracting this into a method would give it a name and everything.

neqo-transport/src/connection/mod.rs
```diff
@@ -2901,6 +2919,43 @@ impl Connection {
...
+            //
+            // We always mark with ECT(0) - if at all - so we only need to check for that.
+            // Also, if we sent a packet with ECT(0) and get only an ACK frame (and not an
+            // ACK-ECN frame), `ecn_counts` will be all zero and the check below will fail,
+            // so no need to explicitly check for the above.
+            //
+            // > ECN validation also fails if the sum of the increase in ECT(0) and ECN-CE counts is
+            // > less than the number of newly acknowledged packets that were originally sent with an
+            // > ECT(0) marking.
+            let newly_acked = acked_packets.len() as u64;
```
martinthomson, line 2945: Please use try_from() rather than `as` where possible.  I think that you can use `?` rather than `.unwrap()` in places that already return `Res`, but `.unwrap()` is totally fine for `u64` <-> `usize` conversions..

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,77 @@
...
+        server.process_input(&client_pkt, now());
+    }
+
+    // Client should now process ACKs with incorrect ECN counts and disable ECN.
+    while let Some(server_pkt) = server.process_output(now()).dgram() {
+        client.process_input(&server_pkt, now());
+    }
+
+    // ECN should now be disabled.
+    assert!(!client.paths.primary().borrow().is_ecn_enabled());
```
martinthomson, line 44: Rather than poke around in the internals, just get the client to `send_something()` and check the TOS on those datagrams.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[test]
+fn disables_when_bleached() {
+    let mut client = default_client();
+    let mut server = default_server();
+    connect_force_idle(&mut client, &mut server);
+    assert!(client.paths.primary().borrow().is_ecn_enabled());
+
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
+    client.stream_send(stream_id, DATA_CLIENT).unwrap();
```
martinthomson, line 30: We have a utility for this called `send_something()`.

neqo-transport/src/frame.rs
```diff
@@ -442,17 +445,25 @@ impl<'a> Frame<'a> {
                 }
 
                 // Now check for the values for ACK_ECN.
-                if t == FRAME_TYPE_ACK_ECN {
-                    dv(dec)?;
-                    dv(dec)?;
-                    dv(dec)?;
-                }
+                let ecn_count: EcnCount = match t {
```
martinthomson, line 448: There is enough code here that I want to suggest that you extract this arm of the match statement into a `decode_ack()` method.

neqo-transport/src/path.rs
martinthomson: The general sentiment from IETF folks doing this stuff is that ECN is safe to use.  I consider it low risk, particularly with the defense against drops and bleaching that is included in this patch.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,77 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
martinthomson, line 1: I'm not seeing any CE markings being applied, to test CC response.  Is that coming later?

neqo-transport/src/path.rs
```diff
@@ -573,14 +578,37 @@ impl Path {
...
+
+    /// Set the ECN count received on this path.
+    pub fn set_ecn_count(&mut self, space: PacketNumberSpace, ecn_count: EcnCount) {
+        qinfo!("Setting ECN count for {:} to {:?}", space, ecn_count);
+        self.ecn_count[space] = ecn_count;
+    }
+
+    /// Whether this path is currently marking packets with ECN.
+    pub fn is_ecn_enabled(&self) -> bool {
+        self.tos != IpTosEcn::NotEct.into()
```
martinthomson, line 604: Is this right, or should this be:

```suggestion
        self.tos.ecn() == IpTosEcn::Ect0
```

neqo-transport/src/path.rs
```diff
@@ -573,14 +578,37 @@ impl Path {
...
+    }
+
+    /// Disable ECN marking on this path.
+    pub fn disable_ecn(&mut self) {
+        self.tos = IpTosEcn::NotEct.into();
```
martinthomson, line 609: If we're going to ignore DSCP, that's fine, but it's worth noting that this will have that effect.

neqo-transport/src/path.rs
```diff
@@ -979,6 +1007,32 @@ impl Path {
...
+            // destination IP addresses for paths on which we had to disable ECN,
+            // in order to not persitently delay connection establishment to
+            // those destinations.)
+            //
+            // (Reusing `MAX_PATH_PROBES` here as a threshold is a bit of a hack,
+            // but avoids defining another constant just for this.)
+            println!(
+                "XXX Lost {} ECN-marked packets on this path",
+                self.lost_ecn_packets
+            );
```
martinthomson, line 1029:
```suggestion
```

Changing this to `qtrace!()` might also be OK.

neqo-transport/src/tracking.rs
```diff
@@ -351,6 +353,23 @@ pub struct AckToken {
     ranges: Vec<PacketRange>,
 }
 
+/// The counts for different ECN marks.
+pub type EcnCount = EnumMap<IpTosEcn, u64>;
```
martinthomson, line 357: You aren't 100% consistent about using this type.  Just check that you are using the alias for things like `EcnCount::default()` and so on.

neqo-transport/src/tracking.rs
```diff
@@ -351,6 +353,23 @@ pub struct AckToken {
...
+
+/// Whether any of the ECN counts are non-zero.
+pub fn have_ecn_count(ecn: &EcnCount) -> bool {
+    ecn[IpTosEcn::Ect0] > 0 || ecn[IpTosEcn::Ect1] > 0 || ecn[IpTosEcn::Ce] > 0
+}
+
+/// Subtract the ECN counts in `b` from `a`.
+pub fn diff_ecn_count(a: &EcnCount, b: &EcnCount) -> EcnCount {
```
martinthomson, line 365: Given that you have two utility functions attached to this, I would change from a type alias to a simple struct wrapper:

```rust
struct EcnCount(EnumMap<IpTosEcn, u64>);
```

Then, I would recommend implementing `std::ops::Sub` for this.  It's saturating, but it's still a fine use of the feature.

Are we counting non-ect?  If not, this might be a more complicated structure, but a leaner one as a result.

neqo-transport/src/tracking.rs
```diff
@@ -396,9 +417,15 @@ impl RecvdPackets {
             unacknowledged_count: 0,
             unacknowledged_tolerance: DEFAULT_ACK_PACKET_TOLERANCE,
             ignore_order: false,
+            ecn_count: EcnCount::default(),
         }
     }
 
+    /// Increase the ECN count for the mark given by `ecn` by `1`.
+    pub fn inc_ecn_count(&mut self, ecn: IpTosEcn) {
+        self.ecn_count[ecn] += 1;
```
martinthomson, line 426: If you have a new type for `EcnCount` you could implement `std::ops::Add<IpTosEcn>` for this.  Just a thought.

neqo-transport/src/tracking.rs
```diff
@@ -619,6 +651,12 @@ impl RecvdPackets {
             last = r.smallest;
         }
 
+        if have_ecn_count {
```
martinthomson, line 654: Do you want to use the wrapper?

martinthomson (commented):
Getting there for sure.  No CC response, but sending and receiving is useful.

mxinden:
> No CC response

Happening with https://github.com/mozilla/neqo/pull/1689 @martinthomson.

neqo-transport/src/connection/tests/ecn.rs
mxinden: I can extend it in https://github.com/mozilla/neqo/pull/1689.

neqo-transport/src/connection/mod.rs
larseggert: Thanks, will fix. `as` just seemed so much more readable, and I didn't know about the downsides.

neqo-transport/src/connection/mod.rs
larseggert: Good idea, moved.

neqo-transport/src/connection/mod.rs
larseggert: Moved to `Path::validate_ecn_use()`.

neqo-transport/src/path.rs
larseggert: I think `IpTosEcn::from(self.tos) != IpTosEcn::NotEct` - we could just check if it's equal to `IpTosEcn::Ect0`, but maybe we'll mark `IpTosEcn::Ect1` at some point in the future.

neqo-transport/src/path.rs
larseggert: Good catch.

neqo-transport/src/tracking.rs
larseggert: We are currently counting Not-ECT.

neqo-transport/src/tracking.rs
larseggert: I'd need to make `RecvdPackets::ecn_count` public though, no?

neqo-transport/src/tracking.rs
larseggert: Sorry, I don't follow?

larseggert:
@martinthomson thanks for the detailed review. I think I fixed most things, see the review responses on a few things where I was unclear.

We probably need to implement something like https://datatracker.ietf.org/doc/html/rfc9000#section-a.4; the current disablement based on a fixed number of lost packets on a path is obviously bad, because eventually we **will** exceed that number.

neqo-transport/src/tracking.rs
```diff
@@ -351,6 +353,58 @@ pub struct AckToken {
     ranges: Vec<PacketRange>,
 }
 
+/// The counts for different ECN marks.
+#[derive(PartialEq, Eq, Debug, Clone, Default, Copy)]
+pub struct EcnCount(EnumMap<IpTosEcn, u64>);
```
martinthomson, line 358: Do you really need a `Copy` implementation here?  This is 256 bits, which I believe is a bit larger than the usual cutoff for `Copy`.

neqo-transport/src/tracking.rs
martinthomson: Since you did implement `Add`:

```suggestion
        self.ecn_count += ecn;
```
:)

martinthomson (approved):
LGTM

neqo-transport/src/ecn.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.count[space] = ecn_count.clone();
+    }
+
+    /// The ECN mark to use for packets sent on this path.
+    pub fn ecn_mark(&self) -> IpTosEcn {
+        match self.state {
+            EcnValidationState::Testing | EcnValidationState::Capable => IpTosEcn::Ect0,
+            EcnValidationState::Failed | EcnValidationState::Unknown => IpTosEcn::NotEct,
+        }
+    }
```
mxinden, line 115: When n `EcnValidationState::Unknown`, one should still mark with `Ect0`, no?

Otherwise `sum_inc < newly_acked` check above triggers, where a returning ACK acknowledges more packets than it has marks.

```suggestion
    /// The ECN mark to use for packets sent on this path.
    pub fn ecn_mark(&self) -> IpTosEcn {
        match self.state {
            EcnValidationState::Testing | EcnValidationState::Capable | EcnValidationState::Unknown => IpTosEcn::Ect0,
            EcnValidationState::Failed => IpTosEcn::NotEct,
        }
    }
```

mxinden (commented):
Rebased https://github.com/mozilla/neqo/pull/1689 onto latest commit of this pull request.

neqo-transport/src/ecn.rs
larseggert: That's not how I understand the text in RFC9000 (but that text is far from clear). I think the assumption is that one would receive an ACK within those ten packets that would validate or invalidate the ECN capability of the path.

neqo-transport/src/ecn.rs
```diff
@@ -38,73 +27,34 @@ enum EcnValidationState {
...
-            },
-            count: EnumMap::default(),
-        }
-    }
-
+    /// Count the number of packets sent out on this path during ECN validation.
+    /// Exit ECN validation if the number of packets sent exceeds `ECN_TEST_COUNT`.
+    /// We do not implement the part of the RFC that says to exit ECN validation if the time since
+    /// the start of ECN validation exceeds 3 * PTO, since this seems to happen much too quickly.
     pub fn count_packets_out(&mut self) {
```
martinthomson, line 46: This doesn't seem to do any counting.  It adds one more to the count.  I'd use `on_packet_sent` or `track_sent_packet`.

That said, I think that your validation is going to be a bit challenging if you have multiple paths.  We don't really get clear signals about which path the ECN markings apply to, so there are a lot of scenarios to think through.

1. The existing path is happily using ECN and a new path comes online.  On that path, we start testing and send marked packets.  Those packets will be acknowledged along with all the packets on the original path.  This should work fine because both paths pass the markings correctly.
2. The new path is bleaching, but the original is fine.  Here, I see no problem either.  Correct markings on the original path will not prevent the gap from being noticed.
3. The old path is bleaching, but the new one is fine.  This is a problematic case.  We aren't able to attribute a lack of markings to either path.
4. Both are bleaching.  This is also fine, in the sense that ECN is disabled on the new path.

Maybe case 3 is just something we can't handle, but I don't think that's true.  We have a count of the probing packets we sent during the testing phase, so we should be able to look for an increase in counts that match.  What we don't have is anything that allows us to track which of the acknowledged packets were sent on the new path.  We could use ACK tokens for that, I guess.

We could also decide that we don't care about this case.  Bleaching can be infectious.  Though that's not particularly satisfactory.  We would leak that we previously used a bleaching path, which isn't great.

Either way, I'd like to see a few tests that covered these cases.

neqo-transport/src/ecn.rs
```diff
@@ -38,73 +27,34 @@ enum EcnValidationState {
...
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct EcnInfo {
     /// The current state of ECN validation on this path.
     state: EcnValidationState,
-
+    /// The number of packets sent so far on the path during the ECN validation.
+    sent: usize,
```
martinthomson, line 35: This is not a great name if the number is only the number of packets that were sent during testing.  `test_packet_count` might work.

neqo-transport/src/ecn.rs
```diff
@@ -38,73 +27,34 @@ enum EcnValidationState {
...
     /// The ECN counts received in the last ACK on this path, for each packet number space.
-    /// Won't be updated after ECN has been disabled on a path.
+    /// Won't be updated after ECN has been tested on the path.
     count: EnumMap<PacketNumberSpace, EcnCount>,
```
martinthomson, line 38: I'd suggest `acked_count` or `acknowledged_count` for this one..

neqo-transport/src/ecn.rs
```diff
@@ -0,0 +1,116 @@
...
+        acked_packets: &[SentPacket],
+        ecn_count: &EcnCount,
+    ) {
+        // RFC 9000, Appendix A.4:
+        // From the "unknown" state, successful validation of the ECN counts in an ACK frame
+        // (see Section 13.4.2.1) causes the ECN state for the path to become "capable", unless
+        // no marked packet has been acknowledged.
+        if self.state != EcnValidationState::Unknown {
+            return;
+        }
```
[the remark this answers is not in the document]
martinthomson: The problem with ECN validation is that it is hard to reset the state.  So performing this validation constantly is not going to work with the code in its current state.  The reason for that is that a single bleached marking will always manifest in the counts.

That's why this doesn't re-test or continuously check for success: this will just fail.

A more robust approach would involve re-testing a failed path, but you have to deliberately switch back to a testing state, plus snapshot the counts at that time, and account for any packets that are still in flight.  It's a bit tricky.  Better to just accept that ECN doesn't work.

neqo-transport/src/tracking.rs
```diff
@@ -349,6 +351,64 @@ pub struct AckToken {
     ranges: Vec<PacketRange>,
 }
 
+/// The counts for different ECN marks.
+#[derive(PartialEq, Eq, Debug, Clone, Default)]
+pub struct EcnCount(EnumMap<IpTosEcn, u64>);
```
martinthomson, line 356: Should this be moved to `ecn.rs`?

neqo-transport/src/ecn.rs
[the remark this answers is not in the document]
mxinden: I think there is a misunderstanding here @martinthomson.

I am suggesting to support going from `EcnValidationState::Capable` to `EcnValidationState::Failed`. In other words, once in `Capable` I am suggesting to still continuously check ECN validity, potentially going to `Failed` on failure.  Once in `EcnValidationState::Failed`, there is no going back, i.e. no re-testing.

If I understand your comment above, you are saying that re-testing, i.e. going from `EcnValidationState::Failed` eventually to `EcnValidationState::Capable` is not going to work with the current logic, but instead requires a more involved approach. I agree with that.

neqo-transport/src/ecn.rs
larseggert: > When in `EcnValidationState::Unknown`, one should still mark with `Ect0`, no?

How would we recover from the case where a path drops all packets with ECT(0)?

mxinden:
How about enabling the `ecn` QUIC Interop testcase with this pull request? https://github.com/larseggert/neqo/pull/22

mxinden:
`neqo-latest` -> `ngtcp2` ECN failure comes at a surprise to me.

`ntcp2` -> `ngtcp2` ECN succeeds on https://interop.seemann.io/.

Downloading the logs from Neqo's CI, no logs for the `ecn` testcase in `neqo-latest_ngtcp2` folder.

larseggert:
> `neqo-latest` -> `ngtcp2` ECN failure comes at a surprise to me.

Looking at the pcap, the very last packet the ngtcp2 server sends (after the neqo client sends CONNECTION_CLOSE) isn't sent with ECT(0) and also cannot be decrypted, either with its key logs or with neqo's.

larseggert:
Thanks for confirming! 

We need to find out why the second CONNECTION_CLOSE comes from. Seems to be a recent regression AFAICT.

larseggert:
@mxinden 199987445d875556feaf6985031814f7a5ed4dea made ngtcp2 happy - thanks for the tip!

larseggert:
@KershawChang @martinthomson this is - yet again - ready for review.

neqo-transport/src/ecn.rs
```diff
@@ -0,0 +1,213 @@
...
+        //
+        // > If an ACK frame newly acknowledges a packet that the endpoint sent with
+        // > either the ECT(0) or ECT(1) codepoint set, ECN validation fails if the
+        // > corresponding ECN counts are not present in the ACK frame.
+        if ack_ecn.is_none() {
+            qwarn!("ECN validation failed, no ECN counts in ACK frame");
+            self.state = EcnValidationState::Failed;
+            return;
+        }
+        let ack_ecn = ack_ecn.unwrap();
```
mxinden, line 164: Might be less error prone without the unwrap:

```suggestion
        let Some(ack_ecn) = ack_ecn else {
            qwarn!("ECN validation failed, no ECN counts in ACK frame");
            self.state = EcnValidationState::Failed;
            return;
        };
```

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+/// The function returns the TOS value of the last packet sent on the old path and the TOS value
+/// of the last packet sent on the new path to allow for verification of correct behavior.
+pub fn migration_with_modifiers(
+    mut orig_path_modifier: impl DatagramModifier,
+    mut new_path_modifier: impl DatagramModifier,
+    burst: usize,
+) -> (IpTos, IpTos) {
+    fixture_init();
+    let mut client = new_client(
+        ConnectionParameters::default().max_streams(StreamType::UniDi, 64), /* .idle_timeout(Duration::from_millis((1 << 62) - 1)), */
```
martinthomson, line 102:
```suggestion
        ConnectionParameters::default().max_streams(StreamType::UniDi, 64),
```

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+        send_something(&mut client, now);
+    }
+
+    // ECN should now be disabled.
+    let client_pkt = send_something(&mut client, now);
+    assert_ecn_disabled(client_pkt.tos());
+}
+
+/// This function performs a handshake over a path that modifies packets via `orig_path_modifier`.
+/// It then sends `path_packets` packets on that path, and then migrates to a new path that
```
martinthomson, line 91: I don't see `path_packets` here.  It looks like this is out of date.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+    server.process_input(&orig_path_modifier(client_pkt).unwrap(), now);
+
+    client
+        .migrate(Some(DEFAULT_ADDR_V4), Some(DEFAULT_ADDR_V4), false, now)
+        .unwrap();
+
+    let mut migrated = false;
+    let probe = new_path_modifier(client.process_output(now).dgram().unwrap());
+    if probe.is_some() {
+        let probe = probe.unwrap();
```
martinthomson, line 137:
```suggestion
    if let Some(probe) = probe {
```

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+    assert_eq!(
+        server.stats().frame_tx.path_challenge,
+        if migrated { 2 } else { 0 }
+    );
+    assert_eq!(
+        server.stats().frame_tx.stream,
+        if migrated { stream_before } else { 1 }
+    );
+
+    if migrated {
```
martinthomson, line 184: In looking at the tests below, you aren't really checking that the connection migrated successfully.  I know that it's more outputs on the function, but I think it would be helpful to track if migration happened.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+#[test]
+fn ecn_migration_ce_bleach_nodata() {
+    let (before, after) = migration_with_modifiers(ce(), bleach(), 0);
+    assert_ecn_enabled(before); // Too few packets sent before migration to conclude ECN validation.
+    assert_ecn_enabled(after); // Too few packets sent after migration to conclude ECN validation.
+}
+
+#[test]
+fn ecn_migration_ce_remark_nodata() {
+    let (before, after) = migration_with_modifiers(ce(), remark(), 0);
```
martinthomson, line 459: If - when the burst is 0 - all the connections succeed with ECN enabled before and after, why not make a single test that loops through all of the functions?  I know that if you are testing whether migration occurred, you will need to check for noop, but that's just one extra check.

Right here, you have a lot of copy-pasted code.

neqo-transport/src/connection/mod.rs
```diff
@@ -1417,6 +1418,13 @@ impl Connection {
         migrate: bool,
         now: Instant,
     ) {
+        let space = PacketNumberSpace::from(packet.packet_type());
+        if let Some(space) = self.acks.get_mut(space) {
+            *space.ecn_marks() += d.tos().into();
+        } else {
+            qdebug!("Not tracking ECN for dropped packet number space");
```
martinthomson, line 1425:
```suggestion
            qtrace!("Not tracking ECN for dropped packet number space");
```

neqo-transport/src/frame.rs
```diff
@@ -426,6 +428,41 @@ impl<'a> Frame<'a> {
...
+                    gap: dv(dec)?,
+                    range: dv(dec)?,
+                };
+                arr.push(ar);
+            }
+
+            // Now check for the values for ACK_ECN.
+            let ecn_count: Option<EcnCount> = match t {
+                FRAME_TYPE_ACK_ECN => Some(EcnCount::new(0, dv(dec)?, dv(dec)?, dv(dec)?)),
+                _ => None,
```
martinthomson, line 454:
```suggestion
                FRAME_TYPE_ACK => None,
                _ => unreachable!("Not a valid ACK frame type"),
```

neqo-transport/src/frame.rs
```diff
@@ -426,6 +428,41 @@ impl<'a> Frame<'a> {
...
+                let ar = AckRange {
+                    gap: dv(dec)?,
+                    range: dv(dec)?,
+                };
+                arr.push(ar);
+            }
+
+            // Now check for the values for ACK_ECN.
+            let ecn_count: Option<EcnCount> = match t {
+                FRAME_TYPE_ACK_ECN => Some(EcnCount::new(0, dv(dec)?, dv(dec)?, dv(dec)?)),
```
martinthomson, line 453: There's another way to spell this:

```rust
let ecn_count = (t == FRAME_TYPE_ACK_ECN).then(|| EcnCount::new(...));

File: neqo-transport/src/frame.rs [hunk 34]
```diff
@@ -426,6 +428,41 @@ impl<'a> Frame<'a> {
             d(dec.decode_varint())
         }
 
+        fn decode_ack<'a>(dec: &mut Decoder<'a>, t: u64) -> Res<Frame<'a>> {
```
martinthomson, line 431:
```suggestion
        fn decode_ack<'a>(dec: &mut Decoder<'a>, ecn: bool) -> Res<Frame<'a>> {
```

This might be easier for the stuff below.

neqo-transport/src/ecn.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Count the number of packets sent out on this path during ECN validation.
+    /// Exit ECN validation if the number of packets sent exceeds `ECN_TEST_COUNT`.
+    /// We do not implement the part of the RFC that says to exit ECN validation if the time since
+    /// the start of ECN validation exceeds 3 * PTO, since this seems to happen much too quickly.
+    pub fn on_packet_sent(&mut self) {
+        if self.state != EcnValidationState::Testing {
+            return;
+        }
```
martinthomson, line 114: You could move the probe count to the state:

```suggestion
        if let EcnValidationState::Testing(probes_sent) = &mut self.state else {
            return;
        }
        *probes_sent += 1;
```

That doesn't make sense for `baseline` or `largest_acked`, but it's a possibility here.

martinthomson (approved):

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -0,0 +1,512 @@
...
+        connect_force_idle, connect_force_idle_with_modifier, default_client, default_server,
+        migration::get_cid, new_client, new_server, send_something,
+    },
+    ecn::ECN_TEST_COUNT,
+    ConnectionId, ConnectionParameters, StreamType,
+};
+
+fn assert_ecn_enabled(tos: IpTos) {
+    assert_eq!(IpTosEcn::from(tos), IpTosEcn::Ect0);
+}
```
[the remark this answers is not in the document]
larseggert: For a single ECN mark, it's a bit odd to say that "ECN is enabled". Maybe `is_ecn_marked()`.

neqo-transport/src/connection/tests/ecn.rs
larseggert: But on second thought this special case makes sense to consolidate.

larseggert:
```rust
#[test]
fn ecn_migration_zero_burst_all_cases() {
    let modifiers = [noop(), bleach(), remark(), ce(), drop()];
    for orig_path_modifier in modifiers.iter() {
        for new_path_modifier in modifiers.iter() {
            let (before, after) =
                migration_with_modifiers(orig_path_modifier, new_path_modifier, 0);
            // Too few packets sent before and after migration to conclude ECN validation.
            assert_ecn_enabled(before);
            assert_ecn_enabled(after);
        }
    }
}
```
@martinthomson any idea how to work around this error?
```
   --> neqo-transport/src/connection/tests/ecn.rs:227:30
    |
38  | fn noop() -> impl DatagramModifier {
    |              --------------------- the expected opaque type
...
42  | fn bleach() -> impl DatagramModifier {
    |                --------------------- the found opaque type
...
227 |     let modifiers = [noop(), bleach(), remark(), ce(), drop()];
    |                              ^^^^^^^^ expected opaque type, found a different opaque type
    |
    = note: expected opaque type `impl connection::tests::DatagramModifier` (opaque type at <neqo-transport/src/connection/tests/ecn.rs:38:14>)
               found opaque type `impl connection::tests::DatagramModifier` (opaque type at <neqo-transport/src/connection/tests/ecn.rs:42:16>)
    = note: distinct uses of `impl Trait` result in different opaque types
```

larseggert: This works
```rust
    let modifiers = [Some, |d| Some(set_tos(d, IpTosEcn::NotEct))];
```
but it obviously duplicates what `noop()` and `bleach()`, etc., are doing...

larseggert: Nevermind, figure it out.

PR #1683, opened 2024-02-26
ci: Try and see if fp-based callgraph provides more detail

larseggert:


larseggert:
I don't think it does.

PR #1688, opened 2024-02-27
feat: disable log during benchmarks

### LINKED ISSUE #1684: Is `env_logger` useful for release builds?
We seem to spend around 8% of cycles in `neqo_common::log::init` (see `transfer.perf` in [this artifact](https://github.com/mozilla/neqo/actions/runs/8052648773/artifacts/1276017609)). That's quite a lot.

Is there even any use in having `env_logger` do something in a release build? Should we simply make the logging macros no-ops for release builds?

larseggert:
To see if this reduces CPU consumption for `neqo_common::log::init`.

Fixes #1684 (if this works)

### DIFF
--- a/neqo-common/src/log.rs
+++ b/neqo-common/src/log.rs
@@ -91,12 +91,8 @@ macro_rules! qerror {
 #[cfg(feature = "bench")]
 #[macro_export]
 macro_rules! qerror {
-    ([$ctx:expr], $($arg:tt)*) => {
-        ()
-    };
-    ($($arg:tt)*) => {
-        ()
-    };
+    ([$ctx:expr], $($arg:tt)*) => ( &($ctx, $($arg)*); );
+    ($($arg:tt)*) => ( &($($arg)*); );
 }
 
 #[cfg(not(feature = "bench"))]
@@ -109,12 +105,8 @@ macro_rules! qwarn {
 #[cfg(feature = "bench")]
 #[macro_export]
 macro_rules! qwarn {
-    ([$ctx:expr], $($arg:tt)*) => {
-        ()
-    };
-    ($($arg:tt)*) => {
-        ()
-    };
+    ([$ctx:expr], $($arg:tt)*) => ( &($ctx, $($arg)*); );
+    ($($arg:tt)*) => ( &($($arg)*); );
 }
 
 #[cfg(not(feature = "bench"))]
@@ -127,12 +119,8 @@ macro_rules! qinfo {
 #[cfg(feature = "bench")]
 #[macro_export]
 macro_rules! qinfo {
-    ([$ctx:expr], $($arg:tt)*) => {
-        ()
-    };
-    ($($arg:tt)*) => {
-        ()
-    };
+    ([$ctx:expr], $($arg:tt)*) => ( &($ctx, $($arg)*); );
+    ($($arg:tt)*) => ( &($($arg)*); );
 }
 
 #[cfg(not(feature = "bench"))]
@@ -145,12 +133,8 @@ macro_rules! qdebug {
 #[cfg(feature = "bench")]
 #[macro_export]
 macro_rules! qdebug {
-    ([$ctx:expr], $($arg:tt)*) => {
-        ()
-    };
-    ($($arg:tt)*) => {
-        ()
-    };
+    ([$ctx:expr], $($arg:tt)*) => ( &($ctx, $($arg)*); );
+    ($($arg:tt)*) => ( &($($arg)*); );
 }
 
 #[cfg(not(feature = "bench"))]
@@ -163,10 +147,6 @@ macro_rules! qtrace {
 #[cfg(feature = "bench")]
 #[macro_export]
 macro_rules! qtrace {
-    ([$ctx:expr], $($arg:tt)*) => {
-        ()
-    };
-    ($($arg:tt)*) => {
-        ()
-    };
+    ([$ctx:expr], $($arg:tt)*) => ( &($ctx, $($arg)*); );
+    ($($arg:tt)*) => ( &($($arg)*); );
 }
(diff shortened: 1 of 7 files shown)


martinthomson (dismissed):
Worth a shot

larseggert:
Doesn't seem to matter: https://share.firefox.dev/3TaGJh4

larseggert:
The current approach works. But it causes [a lot of warnings about unused variables](https://github.com/mozilla/neqo/actions/runs/8066064297/job/22034049296?pr=1688#step:11:47).

neqo-common/src/log.rs
```diff
@@ -80,28 +80,93 @@ macro_rules! log_invoke {
...
 macro_rules! qerror {
     ([$ctx:expr], $($arg:tt)*) => (::neqo_common::log_invoke!(::log::Level::Error, $ctx, $($arg)*););
     ($($arg:tt)*) => ( { ::neqo_common::log::init(); ::neqo_common::do_log!(::log::Level::Error, $($arg)*); } );
 }
+
+#[cfg(feature = "bench")]
+#[macro_export]
+macro_rules! qerror {
+    ([$ctx:expr], $($arg:tt)*) => {
+        ()
```
martinthomson, line 95: According to [this](https://stackoverflow.com/a/44722605), you could do something like this:

```rust
([$ctx:expr], $($arg:tt)*) => {
  {&($ctx, $($arg),*)}
}
```

You would have to validate that the compiler turned that into a noop.

larseggert:
I think I will make this part of a new `no-log` feature, since we probably want to turn this one for `gecko` as well.

larseggert:
@martinthomson going to this construction
```rs
#[cfg(feature = "bench")]
#[macro_export]
macro_rules! qerror {
    ([$ctx:expr], $($arg:tt)*) => { &($ctx, $($arg)*); };
    ($($arg:tt)*) => { &($($arg)*); };
}
```
omits the warnings, but results in borrow issues like this:
```
error[E0382]: use of moved value: `alpn`
   --> neqo-crypto/src/agent.rs:111:8
    |
95  |     let alpn = match (pre, alpn_state) {
    |         ---- move occurs because `alpn` has type `Option<String>`, which does not implement the `Copy` trait
...
110 |     qtrace!([format!("{fd:p}")], "got ALPN {:?}", alpn);
    |                                                   ---- value moved here
111 |     Ok(alpn)
    |        ^^^^ value used here after move
    |
help: consider cloning the value if the performance cost is acceptable
    |
110 |     qtrace!([format!("{fd:p}")], "got ALPN {:?}", alpn.clone());
    |                                                       ++++++++
```

neqo-common/src/log.rs
larseggert: @martinthomson going to this construction
```rs
#[cfg(feature = "bench")]
#[macro_export]
macro_rules! qerror {
    ([$ctx:expr], $($arg:tt)*) => { &($ctx, $($arg)*); };
    ($($arg:tt)*) => { &($($arg)*); };
}
```
omits the warnings, but results in borrow issues like this:
```
error[E0382]: use of moved value: `alpn`
   --> neqo-crypto/src/agent.rs:111:8
    |
95  |     let alpn = match (pre, alpn_state) {
    |         ---- move occurs because `alpn` has type `Option<String>`, which does not implement the `Copy` trait
...
110 |     qtrace!([format!("{fd:p}")], "got ALPN {:?}", alpn);
    |                                                   ---- value moved here
111 |     Ok(alpn)
    |        ^^^^ value used here after move
    |
help: consider cloning the value if the performance cost is acceptable
    |
110 |     qtrace!([format!("{fd:p}")], "got ALPN {:?}", alpn.clone());
    |                                                       ++++++++
```

larseggert:
With the latest changes, here are the remaining errors:
```
warning: unused variable: `fd`
   --> neqo-crypto/src/agent.rs:360:9
    |
360 |         fd: *const ssl::PRFileDesc,
    |         ^^ help: if this is intentional, prefix it with an underscore: `_fd`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `fd`
   --> neqo-crypto/src/agent.rs:875:9
    |
875 |         fd: *mut ssl::PRFileDesc,
    |         ^^ help: if this is intentional, prefix it with an underscore: `_fd`

   Compiling neqo-transport v0.7.1 (/Users/lars/Documents/Code/neqo/neqo-transport)
warning: `neqo-crypto` (lib) generated 2 warnings (run `cargo fix --lib -p neqo-crypto` to apply 2 suggestions)
warning: unused variable: `conn`
  --> neqo-transport/src/connection/dump.rs:23:5
   |
23 |     conn: &Connection,
   |     ^^^^ help: if this is intentional, prefix it with an underscore: `_conn`
   |
   = note: `#[warn(unused_variables)]` on by default

   Compiling neqo-qpack v0.7.1 (/Users/lars/Documents/Code/neqo/neqo-qpack)
warning: `neqo-transport` (lib) generated 1 warning (run `cargo fix --lib -p neqo-transport` to apply 1 suggestion)
   Compiling neqo-http3 v0.7.1 (/Users/lars/Documents/Code/neqo/neqo-http3)
warning: unused variable: `label`
  --> neqo-http3/src/buffered_send_stream.rs:65:13
   |
65 |         let label = ::neqo_common::log_subject!(::log::Level::Debug, self);
   |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_label`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `label`
   --> neqo-http3/src/recv_message.rs:257:13
    |
257 |         let label = ::neqo_common::log_subject!(::log::Level::Debug, self);
    |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_label`
```

larseggert:
#1699 has a simpler fix for this.

Issue #1267, opened 2021-11-16
ACK_FREQUENCY update


martinthomson:
This is changing, we should follow.

martinthomson:
I'm fairly sure that we need to track the latest changes. still.

PR #1689, opened 2024-02-27
feat(transport): pass ECN CE marks to CC

mxinden:
Pass ECN CE marks received through FRAME_TYPE_ACK_ECN frames to the congestion controler.


mxinden:
Rebasing onto latest https://github.com/mozilla/neqo/pull/1678/ now.

larseggert (approved):

neqo-transport/src/path.rs
```diff
@@ -978,8 +978,16 @@ impl Path {
         now: Instant,
     ) {
         debug_assert!(self.is_primary());
+
+        let ecn_ce_received = self.ecn_info.on_packets_acked(acked_pkts, ack_ecn);
+        if ecn_ce_received {
+            // TODO: Should congestion window reduction through CE marks result in
+            // an update to ACK delay? See on_packets_lost below.
```
mxinden, line 985: Remaining open question.

larseggert: Could you say a bit more about this? I don't see anything immediately relevant in `on_packets_lost`.

mxinden: In `Path::on_packets_lost` `if cwnd_reduced` it calls `update_ack_delay`:

https://github.com/mxinden/neqo/blob/705227436c4d9996defee13f2bbeef61a4cd29f3/neqo-transport/src/path.rs#L999-L1001

Should we do the same in `Path::on_packets_acked` if a CE marking was received and the CWND was reduced?

larseggert: I think I'm missing something. Why would we reduce cwnd when we receive ACKs? We should be reducing cwnd when we receive a CE mark on any inbound packet.

larseggert: We should treating the reception of an ECN mark identical to the detection of a loss. (At least for now.)

larseggert: I think what we're doing here in the loss case if part of our (outdated) implementation of draft-ietf-quic-ack-frequency. We should revisit what to do in the case of loss ECN as we update that implementation. See #1267.

mxinden: Thank you @larseggert. Updated with https://github.com/mozilla/neqo/pull/1689/commits/13fd396a565f0cce4e3d9eedd9484bf733fd4f79.

mxinden:
@larseggert I addressed the outstanding comment. Mind taking another look?

larseggert (approved):

Issue #685, opened 2020-06-02
neqo-client: use common logging macros


agrover:
Now that neqo-client is getting more attention and gaining features (and failure modes) it's probably time to move beyond our current text output to something a little better.

Right now it is only using `println` for normal output and `eprintln` for error output. We should look at these callsites and see if they should be log messages instead, using neqo-common's qerror, qinfo, qwarn, qdebug, and qtrace macros.

larseggert:
CC @mxinden since you are touching this code now.

PR #1692, opened 2024-02-28
feat(client,server): rework logging

mxinden:
- In `neqo-client` and `neqo-server` use `neqo_common::log` instead of `println!`   and `eprintln!`.
- Add `-q`, `-v`, `-vv`, `-vvv`, `-vvvv` log level flags via   `clap_verbosity_flag`.
- Set default log level to INFO. Demote many `qinfo!` to `qdebug!`.

---

Fixes https://github.com/mozilla/neqo/issues/685.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -164,7 +164,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         let mut is_app_limited = true;
         let mut new_acked = 0;
         for pkt in acked_pkts {
-            qinfo!(
+            qdebug!(
                 "packet_acked this={:p}, pn={}, ps={}, ignored={}, lost={}, rtt_est={:?}",
                 self,
                 pkt.pn,
@@ -198,7 +198,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
 
         if is_app_limited {
             self.cc_algorithm.on_app_limited();
-            qinfo!("on_packets_acked this={:p}, limited=1, bytes_in_flight={}, cwnd={}, state={:?}, new_acked={}", self, self.bytes_in_flight, self.congestion_window, self.state, new_acked);
+            qdebug!("on_packets_acked this={:p}, limited=1, bytes_in_flight={}, cwnd={}, state={:?}, new_acked={}", self, self.bytes_in_flight, self.congestion_window, self.state, new_acked);
             return;
         }
 
@@ -249,7 +249,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
                 QlogMetric::BytesInFlight(self.bytes_in_flight),
             ],
         );
-        qinfo!([self], "on_packets_acked this={:p}, limited=0, bytes_in_flight={}, cwnd={}, state={:?}, new_acked={}", self, self.bytes_in_flight, self.congestion_window, self.state, new_acked);
+        qdebug!([self], "on_packets_acked this={:p}, limited=0, bytes_in_flight={}, cwnd={}, state={:?}, new_acked={}", self, self.bytes_in_flight, self.congestion_window, self.state, new_acked);
     }
 
     /// Update congestion controller state based on lost packets.
@@ -265,7 +265,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         }
 
         for pkt in lost_packets.iter().filter(|pkt| pkt.cc_in_flight()) {
-            qinfo!(
+            qdebug!(
                 "packet_lost this={:p}, pn={}, ps={}",
                 self,
                 pkt.pn,
@@ -286,7 +286,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             pto,
             lost_packets,
         );
-        qinfo!(
+        qdebug!(
             "on_packets_lost this={:p}, bytes_in_flight={}, cwnd={}, state={:?}",
             self,
             self.bytes_in_flight,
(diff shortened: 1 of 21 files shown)


neqo-client/src/main.rs
```diff
@@ -956,6 +957,10 @@ fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<NeqoQlog> {
 async fn main() -> Res<()> {
     init();
 
+    if std::env::var("RUST_LOG").is_err() {
+        eprintln!("Set RUST_LOG=<level> for logs.")
+    }
```
mxinden, line 962: A bit odd, but probably better than surprising folks that are used to logs appearing automatically. Happy to remove.

larseggert: Maybe add a `-v` command line parameter to increase the log level?

mxinden:
> Maybe add a `-v` command line parameter to increase the log level?

Would `-v` then set the `RUST_LOG` environment variable to be later picked up by `neqo_common::log::init`? That looks like spaghetti code to me. An alternative would be to introduce a `neqo_common::log::init_with_level`. Though in case you want this, I suggest doing so in a separate pull request to keep this one focused on #685.

Exploring another alternative: I removed the `RUST_LOG` hint (https://github.com/mozilla/neqo/pull/1692/commits/514e0acc654b698ac0440c9c18abbed4307cfdf1) and instead changed `neqo-client` to continue printing request content (or content metadata) to the console (https://github.com/mozilla/neqo/pull/1692/commits/28bed420478f5afeaabf0e51530ac91e9fb8c391).

A request to google.com now looks like the below. Always printing request content is somewhat similar to curl's default behavior.

```
➜  neqo git:(common-log) cargo run --bin neqo-client -- 'https://google.com/'                   
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/neqo-client 'https://google.com/'`
READ[0]: 220 bytes
READ[0]: 0 bytes
READ[0]: 0 bytes

➜  neqo git:(common-log) cargo run --bin neqo-client -- 'https://google.com/' --output-read-data   
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/neqo-client 'https://google.com/' --output-read-data`
READ[0]: <HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>
```

@larseggert let me know what you think.

mxinden:
Friendly ping @larseggert.

I don't feel strongly about any of the proposals in this pull request. I want to start on https://github.com/mozilla/neqo/issues/1696, laying the groundwork for https://github.com/mozilla/neqo/issues/1693. For the sake of keeping merge conflicts small, shall we merge or close here?

larseggert:
I don't feel strongly either. I have a mile preference that if we're going to use the logging framework for things, we should use it as much as possible, i.e., avoid esp. `eprintln!` but probably also `println!`. Adding a `neqo_common::log::init_with_level` to set the level based on a command line options seems fine.

KershawChang:
> Friendly ping @larseggert.
> 
> I don't feel strongly about any of the proposals in this pull request. I want to start on #1696, laying the groundwork for #1693. For the sake of keeping merge conflicts small, shall we merge or close here?

Probably close?
I am fine with leaving `eprintln!` and `println!` in neqo_client and neqo_server.

larseggert:
I've changed my mind on this, we **should** use the neqo log functions, and also use [`clap_verbosity_flag`](https://docs.rs/clap-verbosity-flag/2.2.0/clap_verbosity_flag/) to control verbosity. The demo client and server can be quite noisy on the console when large amounts of data is transferred, and it would be good if the less interesting output could be suppressed.

mxinden:
> and also use [`clap_verbosity_flag`](https://docs.rs/clap-verbosity-flag/2.2.0/clap_verbosity_flag/) to control verbosity

Neat!

I will revive the pull request.

larseggert (commented):
LGTM but want to hear from the other owners.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -164,7 +164,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         let mut is_app_limited = true;
         let mut new_acked = 0;
         for pkt in acked_pkts {
-            qinfo!(
```
KershawChang, line 167: Please see this [python script](https://github.com/mozilla/neqo/blob/main/test/mozlog-neqo-cwnd.py) that we use to generate some graphs from the `qinfo` logs.
Can we not change this to `qdebug!`?

martinthomson (approved):

neqo-transport/src/cc/classic_cc.rs
mxinden: How about running [`upload_test.sh`](https://github.com/mozilla/neqo/blob/main/test/upload_test.sh) with `RUST_LOG=neqo_transport::cc=debug` instead?

That way [`mozlog-neqo-cwnd.py`](https://github.com/mozilla/neqo/blob/main/test/mozlog-neqo-cwnd.py) still functions and plain `neqo-client` and `neqo-server` invocations aren't noisy.

What do you think @KershawChang?

If that sounds good, I can extend `upload_test.sh`.

mxinden: @KershawChang is https://github.com/mozilla/neqo/pull/1692/commits/b090a931e6c2d791f3f7843efde5c48d8fa43905 an option for the above concern?

KershawChang: Yes, I think it's fine.
Thanks.

larseggert (approved):

Issue #1693, opened 2024-02-28
Explore multi-packet I/O


mxinden:
On the I/O layer send and receive batches of UDP datagrams (GSO and GRO). In the state machine(s) consume / act on and produce batches of UDP datagrams.

```[tasklist]
### Prototype using quinn-udp
- [ ] Expose `quinn-udp`s multi-packet APIs through `neqo_common::udp`.
- [ ] Pass multiple packets to and accept from `neqo-http3` / `neqo-transport` `process_*` methods.
```

Past discussions:
- https://github.com/mozilla/neqo/pull/1530/

---

Creating an issue early in order to coordinate the various efforts in a single place. 

I will work on a prototype the next couple of days.

larseggert:
@mxinden are we done here?

mxinden:
Multi-packet send and receive are in place. Thus, yes, I think this generic issue can be closed. Thanks!

PR #1695, opened 2024-02-29
fix: feature flag Datagram::into_data

mxinden:
`Datagram::into_data` is used in `neqo_common::udp` only. `neqo_common::udp` is behind the `udp` feature. The feature is not part of the crate's default features. Thus a plain `cargo check` rightly complains with:

```
warning: method `into_data` is never used
  --> neqo-common/src/datagram.rs:58:19
   |
20 | impl Datagram {
   | ------------- method in this implementation
...
58 |     pub(crate) fn into_data(self) -> Vec<u8> {
   |                   ^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

This commit hides `into_data` behind the `udp` feature as well.

Introduced in https://github.com/mozilla/neqo/pull/1604. My bad.

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -54,6 +54,7 @@ impl Datagram {
         self.ttl
     }
 
+    #[cfg(feature = "udp")]
     #[must_use]
     pub(crate) fn into_data(self) -> Vec<u8> {
         self.d
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -98,7 +98,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@v1
 
       - name: Install Rust tools
-        run: cargo +${{ matrix.rust-toolchain }} binstall --no-confirm cargo-llvm-cov cargo-nextest
+        run: cargo +${{ matrix.rust-toolchain }} binstall --no-confirm cargo-llvm-cov cargo-nextest cargo-hack
 
       # This step might be removed if the distro included a recent enough
       # version of NSS.  Ubuntu 20.04 only has 3.49, which is far too old.
@@ -158,7 +158,6 @@ jobs:
           # respective default features only. Can reveal warnings otherwise
           # hidden given that a plain cargo clippy combines all features of the
           # workspace. See e.g. https://github.com/mozilla/neqo/pull/1695.
-          cargo +${{ matrix.rust-toolchain }} install cargo-hack
           cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
         if: success() || failure()
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -153,7 +153,13 @@ jobs:
         if: success() || failure()
 
       - name: Clippy
-        run: cargo +${{ matrix.rust-toolchain }} clippy --all-targets -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
+        run: |
+          # Use cargo-hack to run clippy on each crate individually with its
+          # respective default features only. Can reveal warnings otherwise
+          # hidden given that a plain cargo clippy combines all features of the
+          # workspace. See e.g. https://github.com/mozilla/neqo/pull/1695.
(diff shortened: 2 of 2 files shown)


mxinden:
@larseggert do you run `cargo check` in the root directory or `neqo-common`? Can you try the latter?

```
➜  neqo git:(main) cd neqo-common 
➜  neqo-common git:(main) cargo check
   Compiling syn v2.0.48
   Compiling neqo-common v0.7.1 (/home/mxinden/code/github.com/mozilla/neqo/neqo-common)
   Compiling darling_core v0.20.3
   Compiling serde_derive v1.0.195
   Compiling enum-map-derive v0.17.0
    Checking enum-map v2.7.3
   Compiling darling_macro v0.20.3
   Compiling darling v0.20.3
   Compiling serde_with_macros v3.4.0
    Checking serde v1.0.195
    Checking smallvec v1.11.2
    Checking serde_json v1.0.108
    Checking serde_with v3.4.0
    Checking qlog v0.12.0
warning: method `into_data` is never used
  --> neqo-common/src/datagram.rs:58:19
   |
20 | impl Datagram {
   | ------------- method in this implementation
...
58 |     pub(crate) fn into_data(self) -> Vec<u8> {
   |                   ^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `neqo-common` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 10.90s
➜  neqo-common git:(main) cargo --version
cargo 1.76.0 (c84b36747 2024-01-18)
```

larseggert:
I ran it at the root. 

If there are issues with running it in the subdirectories, we should add that to CI.

mxinden:
A `cargo check` at the root will compile all crates of the workspace. That includes `neqo-client` and `neqo-server`, each activating the `udp` feature in `neqo-common`. With the `udp` feature the `Datagram::into_data` function is used. Thus a `cargo check` at the root does not trigger the warning.

I altered the `cargo clippy` CI step to use [`cargo-hack`'s `--workspace` feature](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--all---workspace).

``` diff
modified   .github/workflows/check.yml
@@ -153,7 +153,13 @@ jobs:
         if: success() || failure()
 
       - name: Clippy
-        run: cargo +${{ matrix.rust-toolchain }} clippy --all-targets -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
+        run: |
+          # Use cargo-hack to run clippy on each crate individually with its
+          # respective default features only. Can reveal warnings otherwise
+          # hidden given that a plain cargo clippy combines all features of the
+          # workspace. See e.g. https://github.com/mozilla/neqo/pull/1695.
+          cargo +${{ matrix.rust-toolchain }} install cargo-hack
+          cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
```

I have pushed this commit in front of the original fix. The first commit (CI change only) now fails as expected. See e.g. https://github.com/mozilla/neqo/actions/runs/8101846886/job/22143006404.

With the second commit (the actual fix) CI is green. (Well one of them failed unrelated on the `Fetch NSS and NSPR` step.)

@larseggert let me know what you think.

neqo-common/src/datagram.rs
```diff
@@ -54,6 +54,7 @@ impl Datagram {
         self.ttl
     }
 
+    #[cfg(feature = "udp")]
```
martinthomson, line 57: Would this warning be generated if we added `impl From<Datagram> for Vec<u8>` instead?

mxinden: Currently `Datagram::into_data` is `pub(crate)` as it is only used within `neqo-common`, more specifically the `neqo_common::udp` module. Replacing `Datagram::into_data` with `impl From<Datagram> for Vec<u8>` would make the functionality `pub`. All `pub` crate items are assumed to be used. Thus the warning would be gone, yes.

I suggest not exposing it as `pub`, as that exposes the internal data representation of `Datagram`. This is not a strong opinion. Happy to go with whatever you prefer.

(Tangentially, we might want to reconsider the data representation of `Datagram` as part of https://github.com/mozilla/neqo/issues/1693.)

larseggert (approved):

Issue #1697, opened 2024-02-29
Crash in [@ neqo_transport::path::Path::local_cid]


KershawChang:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1882863.

The crash is at this [line](https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-transport/src/path.rs#L664).
It seems the reason is that `local_cid` is called before `make_permanent`, but I am not exactly sure how this could happen.

larseggert:
We need to find better ways to create reproducers from these crash reports. (In general, not just for this one.) What can we do?

larseggert:
Was fixed by #1698

PR #1698, opened 2024-03-01
Return an error if the path is not permanent

KershawChang:
For issue #1697 
This is just a workaround for stop crashing.
The root cause of this crash is still not clear. Maybe necko does something wrong.

Note that the [crash signature](https://crash-stats.mozilla.org/report/index/42c84380-18ac-4d7a-b425-8947f0240224) seems wrong. The frame 9 is `neqo_transport::path::Path::local_cid()`, but the provided [link](https://hg.mozilla.org/releases/mozilla-beta/file/f52b92d785d0346848c447a502d5d50bebd2ea53/third_party/rust/neqo-transport/src/path.rs#l679) to the code indicates the function is `neqo_transport::path::Path::remote_cid()`.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1937,6 +1937,10 @@ impl Connection {
             };
 
             let path = close.path().borrow();
+            if path.is_temporary() {
+                debug_assert!(false, "Path is not permanent");
+                return Err(Error::InternalError);
+            }
             let (_, mut builder) = Self::build_packet_header(
                 &path,
                 cspace,
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1942,9 +1942,7 @@ impl Connection {
             // a packet on a new path, we avoid sending (and the privacy risk) rather
             // than reuse a connection ID.
             if path.is_temporary() {
-                if cfg!(test) {
-                    panic!("attempting to close with a temporary path");
-                }
+                assert!(!cfg!(test), "attempting to close with a temporary path");
                 return Err(Error::InternalError);
             }
             let (_, mut builder) = Self::build_packet_header(
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1942,8 +1942,9 @@ impl Connection {
             // a packet on a new path, we avoid sending (and the privacy risk) rather
             // than reuse a connection ID.
             if path.is_temporary() {
-                #[cfg(test)]
-                assert!(false, "path is temporary");
+                if cfg!(test) {
+                    panic!("attempting to close with a temporary path");
+                }
                 return Err(Error::InternalError);
             }
             let (_, mut builder) = Self::build_packet_header(
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1942,6 +1942,8 @@ impl Connection {
             // a packet on a new path, we avoid sending (and the privacy risk) rather
             // than reuse a connection ID.
             if path.is_temporary() {
+                #[cfg(test)]
+                assert!(false, "path is temporary");
                 return Err(Error::InternalError);
             }
             let (_, mut builder) = Self::build_packet_header(
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1937,8 +1937,11 @@ impl Connection {
             };
 
             let path = close.path().borrow();
+            // In some error cases, we will not be able to make a new, permanent path.
+            // For example, if we run out of connection IDs and the error results from
+            // a packet on a new path, we avoid sending (and the privacy risk) rather
+            // than reuse a connection ID.
             if path.is_temporary() {
-                debug_assert!(false, "Path is not permanent");
                 return Err(Error::InternalError);
             }
             let (_, mut builder) = Self::build_packet_header(
--- a/neqo-transport/src/connection/tests/migration.rs
+++ b/neqo-transport/src/connection/tests/migration.rs
@@ -6,6 +6,7 @@
 
 use std::{
     cell::RefCell,
+    mem,
     net::{IpAddr, Ipv6Addr, SocketAddr},
     rc::Rc,
     time::{Duration, Instant},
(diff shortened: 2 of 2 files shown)


martinthomson (commented):
Is there any way that you could generate a test case for this?

KershawChang:
> Is there any way that you could generate a test case for this?

It appears that `make_permanent` is called at the very beginning when a connection is created, which suggests that the only place setting `remote_cid` to none is [here](https://searchfox.org/mozilla-central/rev/6bc0f370cc459bf79e1330ef74b21009a9848c91/third_party/rust/neqo-transport/src/path.rs#393). Maybe this is caused by receiving an unexpected `NewConnectionId` frame?
Unfortunately, I'm uncertain about how to write a test case to confirm this hypothesis

KershawChang:
@martinthomson , I think I am out of my ideas.
I'd really appreciate if you can point me a new direction to investigate. Thanks.

martinthomson:
I spent a little time with process_input and we have several cases that are relevant here:

1. The packet is processed successfully.  No problem.
3. There was an error in packet preprocessing:
    1. The packet was garbage, which we ignore.  No problem.
    1. We received a Retry packet.  Again, no close message sent then. No problem.
    2. We received a bad Version Negotiation packet.  We go straight to closed without sending anything.  No problem.
    3. We received a Version Negotiation packet and were not able to create a new client.  PROBLEM?
5. We want to update keys, but we have run out of keys.  We deliberately do not send any close message in this case, so no problem.
6. We received a stateless reset.  We don't send anything more from this state.  No problem.
7. There was an error in the packet processing. We always setup the path from this state.  No problem?

That last one contains a problem.  If we receive a bad packet from a new path AND we don't have a connection ID AND the peer requires a connection ID, the attempt to make a new path will fail.

That seems like a reasonable test case to try out.

KershawChang:
I found a way to trigger the crash by calling `process_output` in the end of test `retire_prior_to_migration_failure`.
Note that we do call `ensure_permanent`, but we are not able to make the path permanent.
Given this situation, it seems our options are limited. One possible solution might be adding a check to see if the path is temporary before building a packet.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -897,6 +898,10 @@ fn retire_prior_to_migration_failure() {
             ..
         }
     ));
+
+    // See issue #1697. We had a crash when the client had a temporary path and
+    // process_output is called.
+    mem::drop(client.process_output(now()));
```
martinthomson, line 904: This is good.  I'd prefer to see this as a separate test case though.

Did you not have any luck with pushing a bad packet in on a new path?  You can use the [test frame writer](https://github.com/mozilla/neqo/blob/e76645f2094365f4228b8c3b6ed8cf8694cf8849/neqo-transport/src/connection/mod.rs#L283-L284) to make a garbage packet (a single varint should do the trick, as that will be interpreted as a frame type that we don't support).

neqo-transport/src/connection/mod.rs
```diff
@@ -1937,6 +1937,9 @@ impl Connection {
             };
 
             let path = close.path().borrow();
+            if path.is_temporary() {
+                return Err(Error::InternalError);
+            }
```
martinthomson, line 1942: This is a good fix.  Can you add a comment here about how we might get here?

```suggestion
            // In some error cases, we will not be able to make a new, permanent path.
            // For example, if we run out of connection IDs and the error results from
            // a packet on a new path, we avoid sending (and the privacy risk) rather
            // than reuse a connection ID.
            if path.is_temporary() {
                return Err(Error::InternalError);
            }
```

neqo-transport/src/connection/tests/migration.rs
KershawChang: Thanks. Will try to write another test case.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -950,3 +951,43 @@ fn retire_prior_to_migration_success() {
...
+/// Test the case that we run out of connection ID and receive an invalid frame
+/// from a new path.
+#[test]
+fn error_on_new_path_with_no_connection_id() {
+    let mut client = default_client();
+    let mut server = default_server();
+    connect_force_idle(&mut client, &mut server);
+
+    let dgram = send_something(&mut server, now());
+    let dgram = change_path(&dgram, DEFAULT_ADDR_V4);
```
martinthomson, line 972: Do you need to change the path for this datagram?  This could create a valid path that is later available for use when you switch for the final test.

I think that your later calls to process_input on the default path will cause a switch back to v6, which should invalidate the path that this switch creates, but it's interesting that we would forget this path.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -950,3 +951,43 @@ fn retire_prior_to_migration_success() {
...
+    client.process_input(&dgram, now());
+
+    let cid_gen: Rc<RefCell<dyn ConnectionIdGenerator>> =
+        Rc::new(RefCell::new(CountingConnectionIdGenerator::default()));
+    server.test_frame_writer = Some(Box::new(RetireAll { cid_gen }));
+    let retire_all = send_something(&mut server, now());
+    server.test_frame_writer = None;
```
martinthomson, line 979:
```suggestion
```

This is just reassigned immediately anyway.

You might also avoid the last assignment to `None` as well, but that matters less.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -950,3 +951,43 @@ fn retire_prior_to_migration_success() {
     assert_ne!(get_cid(&dgram), original_cid);
     assert_ne!(get_cid(&dgram), probe_cid);
 }
+
+struct GarbageWriter {}
+
+impl crate::connection::test_internal::FrameWriter for GarbageWriter {
+    fn write_frames(&mut self, builder: &mut PacketBuilder) {
+        builder.encode_varint(u32::MAX);
```
martinthomson, line 959:
```suggestion
        // Not a valid frame type.
        builder.encode_varint(u32::MAX);
```

martinthomson (commented):
That test is very nice.  I'm sure that it took longer to write than it looks, because it looks effortless.  Thanks Kershaw.  Good work.

neqo-transport/src/connection/tests/migration.rs
KershawChang: Yes, you are right. We don't need this.

neqo-transport/src/connection/tests/migration.rs
KershawChang: Will remove this.

neqo-transport/src/connection/mod.rs
```diff
@@ -1937,6 +1937,15 @@ impl Connection {
             };
 
             let path = close.path().borrow();
+            // In some error cases, we will not be able to make a new, permanent path.
+            // For example, if we run out of connection IDs and the error results from
+            // a packet on a new path, we avoid sending (and the privacy risk) rather
+            // than reuse a connection ID.
+            if path.is_temporary() {
+                #[cfg(test)]
+                assert!(false, "path is temporary");
```
martinthomson, line 1946:
```suggestion
                if cfg!(test) {
                    panic!("attempting to close with a temporary path");
                }
```

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -950,3 +951,39 @@ fn retire_prior_to_migration_success() {
...
+    fn write_frames(&mut self, builder: &mut PacketBuilder) {
+        // Not a valid frame type.
+        builder.encode_varint(u32::MAX);
+    }
+}
+
+/// Test the case that we run out of connection ID and receive an invalid frame
+/// from a new path.
+#[test]
+#[should_panic(expected = "path is temporary")]
```
martinthomson, line 967: Change as above, or similar.

martinthomson (approved):
Clippy complained, but sort that out and this is good to land.

neqo-transport/src/connection/mod.rs
KershawChang: Thanks!

PR #1700, opened 2024-03-01
ci: Use Swatinem/rust-cache

larseggert:
If it ends up speeding CI up; testing.

### DIFF
--- a/.github/actions/rust/action.yml
+++ b/.github/actions/rust/action.yml
@@ -27,9 +27,11 @@ runs:
 
     - name: Install cargo-binstall (Windows)
       if: runner.os != 'Windows'
+      shell: bash
       run: curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
 
     - name: Install Rust tools
+      shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/bench.yml
```diff
@@ -25,15 +25,15 @@ jobs:
...
+
       - name: Install Rust
         uses: ./.github/actions/rust
         with:
           version: $TOOLCHAIN
           components: rustfmt
 
-      - name: Configure Rust
-        run: cargo install flamegraph
-
```
martinthomson, line 36: Glad to see the ad-hoc stuff cleaned up.

.github/actions/rust/action.yml
```diff
@@ -20,11 +20,21 @@ runs:
...
+      run: curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+
+    - name: Install Rust tools
+      shell: bash
+      run: cargo +${{ inputs.version }} binstall --no-confirm cargo-llvm-cov cargo-nextest flamegraph
+
+    - name: Use Rust cache
+      uses: Swatinem/rust-cache@v2
+      with:
+        cache-all-crates: "true"
```
martinthomson, line 40: Looking at the docs, `cache-targets` seems like it is where the performance wins come from, but I'm concerned that we're signing on for very large caches.

martinthomson (approved):

PR #1703, opened 2024-03-01
ci: Try longer when downloading NSS and NSPR

larseggert:


mxinden (approved):

larseggert:
I actually wonder if we should try `hg` once and if that doesn't work, fall back to the `git` mirror instead?

larseggert:
The `git` mirror _should_ be up-to-date, but apparently there are sometimes sync issues.

martinthomson:
A failure in git synchronization could be treated like any other CI failure.  It makes us canaries for that process, but I think that is a reasonable thing for us to do.

larseggert:
OK, closing this in favor of #1710.

Issue #1690, opened 2024-02-27
Performance opportunities


larseggert:
```diff
diff --git a/neqo-common/src/udp.rs b/neqo-common/src/udp.rs
index 7ad0b976..ded473cb 100644
--- a/neqo-common/src/udp.rs
+++ b/neqo-common/src/udp.rs
@@ -71,6 +71,8 @@ impl Socket {

     /// Receive a UDP datagram on the specified socket.
     pub fn recv(&self, local_address: &SocketAddr) -> Result<Option<Datagram>, io::Error> {
+        // TODO: The next line uses about 7% of CPU time on neqo-client to zero out the buffer.
+        // Figure out a way to avoid this.
         let mut buf = [0; u16::MAX as usize];

         let mut meta = RecvMeta::default();
diff --git a/neqo-transport/src/recv_stream.rs b/neqo-transport/src/recv_stream.rs
index 5da80d60..518ee45d 100644
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -364,6 +364,8 @@ impl RxStreamOrderer {
                 keep = true;
             }
             if keep {
+                // TODO: We're spending about 4% of client CPU time in `split_off`, and almost 8%
+                // in this `read` function. Optimize.
                 let mut keep = self.data_ranges.split_off(&range_start);
                 mem::swap(&mut self.data_ranges, &mut keep);
                 return copied;
```

martinthomson:
It seems to me like we could try using a `Vec` or `VecDeque` for the receive buffer.  That would mean copying bytes into a larger buffer (with wrap-around), but it might make these trim operations easier to manage.

A simpler approach would be to try `BTreeMap::retain` as an alternative.

mxinden:
I can tackle the first one as part of the prototype proposed in https://github.com/mozilla/neqo/issues/1693.

larseggert:
Note that #1751 really changed where cycles are spent. It's not clear whether these are still bottlenecks.

larseggert:
Hence closing.

PR #1707, opened 2024-03-02
fix(udp): pre-allocate receive buffer

mxinden:
Instead of allocating a new receive buffer on each call to `neqo_common::udp::Socket::recv`, preallocate a long-lived buffer in `Socket::bind` and reuse this buffer on each read.

See flamegraphs below. With the patch `neqo_common::udp::Socket::recv` simply delegates to `tokio::net::udp::UdpSocket::try_io` instead of spending significant time in `memset`.

Quick-fix while working on https://github.com/mozilla/neqo/issues/1693.

Addresses first performance issue in https://github.com/mozilla/neqo/issues/1690.

Before:

flamegraph

After:

patch

### DIFF
--- a/neqo-common/src/udp.rs
+++ b/neqo-common/src/udp.rs
@@ -21,6 +21,7 @@ use crate::{Datagram, IpTos};
 pub struct Socket {
     socket: tokio::net::UdpSocket,
     state: UdpSocketState,
+    recv_buf: Vec<u8>,
 }
 
 impl Socket {
@@ -31,6 +32,7 @@ impl Socket {
         Ok(Self {
             state: quinn_udp::UdpSocketState::new((&socket).into())?,
             socket: tokio::net::UdpSocket::from_std(socket)?,
+            recv_buf: vec![0; u16::MAX as usize],
         })
     }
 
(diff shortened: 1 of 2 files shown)


neqo-common/src/udp.rs
```diff
@@ -101,16 +101,19 @@ impl Socket {
...
         }
 
         Ok(Some(Datagram::new(
             meta.addr,
             *local_address,
             meta.ecn.map(|n| IpTos::from(n as u8)).unwrap_or_default(),
             None, // TODO: get the real TTL https://github.com/quinn-rs/quinn/issues/1749
-            &buf[..meta.len],
+            &self.recv_buf[..meta.len],
         )))
```
mxinden, line 117: Note that each call to `Datagram::new` allocates the datagram's data in a new `Vec`.

https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-common/src/datagram.rs#L20-L35

Changing this is a larger undertaking I would like to tackle with https://github.com/mozilla/neqo/issues/1693.

larseggert (approved):

neqo-common/src/udp.rs
```diff
@@ -31,6 +32,7 @@ impl Socket {
         Ok(Self {
             state: quinn_udp::UdpSocketState::new((&socket).into())?,
             socket: tokio::net::UdpSocket::from_std(socket)?,
+            recv_buf: vec![0; u16::MAX as usize],
```
martinthomson, line 35: This is too large.  I get that we don't want to overflow, but unless we are receiving multiple datagrams, we only need 2k or so (a constant might be a good idea for that limit).

larseggert: Why is this too large? This is about ~43 packets, which is smaller than realistic WAN congestion windows. Are you worried about Fx memory usage of this is per-socket?

martinthomson: Ignore the comment.  I was reading this patch on its own and missed the follow-up that includes multiple packets.

I do think that a constant is the right thing to use for this, rather than the (rather arbitrary) `u16::MAX`.

mxinden: > I do think that a constant is the right thing to use for this, rather than the (rather arbitrary) `u16::MAX`.

:+1: I prepared https://github.com/mozilla/neqo/pull/1709.

> I was reading this patch on its own and missed the follow-up that includes multiple packets.

For others to follow, this is  https://github.com/mozilla/neqo/pull/1708.

PR #1709, opened 2024-03-05
refactor(udp): define const for receive buffer size

mxinden:
Follow up to https://github.com/mozilla/neqo/pull/1707#discussion_r1510447571.

Reading multiple datagrams in single `Socket::recv` call will happen with https://github.com/mozilla/neqo/pull/1708.

### DIFF
--- a/neqo-common/src/udp.rs
+++ b/neqo-common/src/udp.rs
@@ -18,6 +18,11 @@ use tokio::io::Interest;
 
 use crate::{Datagram, IpTos};
 
+/// Socket receive buffer size.
+///
+/// Allows reading multiple datagrams in a single [`Socket::recv`] call.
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

larseggert:
Does quinn-udp maybe already define a constant for max. GSO size?

mxinden:
It does, though that relies on the system dependent GRO _segment size_ and the `recvmmsg` batch size:

https://github.com/quinn-rs/quinn/blob/9611c5ef6932781eb84bc775e7f008474531b20e/quinn/src/endpoint.rs#L685-L690

I suggest we introduce this logic once we introduce proper GSO and GRO support (https://github.com/mozilla/neqo/issues/1693). I see this pull request and https://github.com/mozilla/neqo/pull/1708 as bug fixes only.

PR #1710, opened 2024-03-05
ci: Use the git mirrors for NSS and NSPR

larseggert:
Since the mercurial ones are often flaky.

Also build NSS as part of the `nss` action, to reduce replication.

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -1,21 +1,19 @@
-name: Checkout NSS and NSPR
+name: Fetch and build NSS
 
 runs:
   using: composite
   steps:
-    - name: Fetch NSS and NSPR
+    - name: Fetch and build NSS
       shell: bash
       run: |
-        for i in {1..$RETRIES}; do
-          hg clone https://hg.mozilla.org/projects/nspr "$NSPR_DIR" && break || sleep $DELAY && false
-        done
-        for i in {1..$RETRIES}; do
-          hg clone https://hg.mozilla.org/projects/nss "$NSS_DIR" && break || sleep $DELAY && false
-        done
+        git clone --depth=1 https://github.com/nss-dev/nspr "$NSPR_DIR"
+        git clone --depth=1 https://github.com/nss-dev/nss "$NSS_DIR"
+        $NSS_DIR/build.sh -g -Ddisable_tests=1 -o
         echo "NSS_DIR=$NSS_DIR" >> "$GITHUB_ENV"
         echo "NSPR_DIR=$NSPR_DIR" >> "$GITHUB_ENV"
+        echo "LD_LIBRARY_PATH=$NSS_DIR/../dist/Release/lib" >> "$GITHUB_ENV"
+        echo "DYLD_FALLBACK_LIBRARY_PATH=$NSS_DIR/../dist/Release/lib" >> "$GITHUB_ENV"
+        echo "$NSS_DIR/../dist/Release/lib" >> "$GITHUB_PATH"
       env:
         NSS_DIR: ${{ github.workspace }}/nss
         NSPR_DIR: ${{ github.workspace }}/nspr
-        RETRIES: 10
-        DELAY: 10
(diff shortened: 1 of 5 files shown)


.github/workflows/check.yml
```diff
@@ -106,14 +98,15 @@ jobs:
           # See https://github.com/ilammy/msvc-dev-cmd#name-conflicts-with-shell-bash
           rm /usr/bin/link.exe
 
+      # This step might be removed if the distro included a recent enough
+      # version of NSS.  Ubuntu 20.04 only has 3.49, which is far too old.
+      # (neqo-crypto/build.rs would also need to query pkg-config to get the
+      # right build flags rather than building NSS.)
```
martinthomson, line 104: Want to use an issue rather than a comment?

.github/actions/nss/action.yml
```diff
@@ -1,21 +1,19 @@
...
-          hg clone https://hg.mozilla.org/projects/nss "$NSS_DIR" && break || sleep $DELAY && false
-        done
+        git clone --depth=1 https://github.com/nss-dev/nspr "$NSPR_DIR"
+        git clone --depth=1 https://github.com/nss-dev/nss "$NSS_DIR"
+        $NSS_DIR/build.sh -g -Ddisable_tests=1 -o
         echo "NSS_DIR=$NSS_DIR" >> "$GITHUB_ENV"
         echo "NSPR_DIR=$NSPR_DIR" >> "$GITHUB_ENV"
+        echo "LD_LIBRARY_PATH=$NSS_DIR/../dist/Release/lib" >> "$GITHUB_ENV"
+        echo "DYLD_FALLBACK_LIBRARY_PATH=$NSS_DIR/../dist/Release/lib" >> "$GITHUB_ENV"
+        echo "$NSS_DIR/../dist/Release/lib" >> "$GITHUB_PATH"
```
martinthomson, line 16: I would create an intermediate variable for these like so:

```suggestion
        NSS_OUT="$NSS_DIR/../dist/$(cat "$NSS_DIR/../dist/latest")"
        echo "LD_LIBRARY_PATH=$NSS_OUT/lib" >> "$GITHUB_ENV"
        echo "DYLD_FALLBACK_LIBRARY_PATH=$NSS_OUT/lib" >> "$GITHUB_ENV"
        echo "$NSS_OUT/lib" >> "$GITHUB_PATH"
```

That has the advantage of being resilient to changes in the flags on the build.

martinthomson (approved):
I guess we'll see if approval leads to merging even when I make suggestions...

.github/workflows/check.yml
larseggert: Done: #1711 1711

.github/actions/nss/action.yml
larseggert: Thanks, I'll pick this up. If I understand `neqo-common/build.rs` correctly though, we do need to also keep `NSS_DIR` defined, so the crate build doesn't do another checkout?

larseggert:
@martinthomson see #1712

.github/actions/nss/action.yml
martinthomson: That build script can be conditioned to do whatever you choose.  My suggestion would be to attempt to use the system NSS (which means detecting that it has a sufficient version), THEN fall back to a checkout.

FWIW, you can maybe lift some of the work in https://github.com/martinthomson/ohttp/blob/978dcfbcfafeeea152e52670dff778e60a840eb4/ohttp/build.rs#L342 for this.

larseggert: Agreed, using the system NSS (if sufficiently recent) would be a great simplification to how we build neqo.

larseggert: > FWIW, you can maybe lift some of the work in https://github.com/martinthomson/ohttp/blob/978dcfbcfafeeea152e52670dff778e60a840eb4/ohttp/build.rs#L342 for this.

Aside: It might be worthwhile to factor this (= Rust NSS packaging) into its own crate...

PR #1712, opened 2024-03-06
ci: Further tweaks to building NSS

larseggert:
Picks up the comments @martinthomson left on #1710

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -37,7 +37,7 @@ runs:
           # build process doesn't provide.
           OPT="-o"
           NSS_TARGET=Release
-          [ "${{ runner.os }}" != "Windows" ] && CFLAGS="-ggdb3 -fnoomit-frame-pointer"
+          [ "${{ runner.os }}" != "Windows" ] && CFLAGS="-ggdb3 -fno-omit-frame-pointer"
         else
           NSS_TARGET=Debug
         fi
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -31,13 +31,13 @@ runs:
     - name: Build
       shell: bash
       run: |
-        if [ "${{ inputs.type }}" != "Debug" && "${{ runner.os }}" != "Windows" ]; then
+        if [ "${{ inputs.type }}" != "Debug" ]; then
           # We want to do an optimized build for accurate CPU profiling, but
           # we also want debug symbols and frame pointers for that, which the normal optimized NSS
           # build process doesn't provide.
           OPT="-o"
           NSS_TARGET=Release
-          export CFLAGS="-ggdb3 -fno-omit-frame-pointer"
+          [ "${{ runner.os }}" != "Windows" ] && export CFLAGS="-ggdb3 -fno-omit-frame-pointer"
         else
           NSS_TARGET=Debug
         fi
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -19,7 +19,7 @@ runs:
         # we also want debug symbols for that, which the normal optimized NSS
         # build process doesn't provide.
         [ "${{ runner.os }}" != "Windows" ] && export CFLAGS=-ggdb3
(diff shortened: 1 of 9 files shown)


.github/actions/nss/action.yml
```diff
@@ -1,19 +1,53 @@
 name: Fetch and build NSS
 
+inputs:
+  type:
+    description: 'Whether to do a debug or release build of NSS'
+    required: true
+    default: 'Debug'
```
martinthomson, line 7: Having both required and default seem to be incompatible.

I would also recommend a default of 'Release' on the basis that we won't be debugging NSS code in CI, so the extra performance is probably worth it.

.github/actions/nss/action.yml
```diff
@@ -1,19 +1,53 @@
...
+    - name: Build
       shell: bash
       run: |
-        git clone --depth=1 https://github.com/nss-dev/nspr "$NSPR_DIR"
-        git clone --depth=1 https://github.com/nss-dev/nss "$NSS_DIR"
-        $NSS_DIR/build.sh -g -Ddisable_tests=1 -o
+        # We want to do an optimized build for accurate CPU profiling, but
+        # we also want debug symbols for that, which the normal optimized NSS
+        # build process doesn't provide.
+        [ "${{ runner.os }}" != "Windows" ] && export CFLAGS=-ggdb3
```
martinthomson, line 37: Does this belong inside the if statement then?

.github/actions/nss/action.yml
```diff
@@ -1,19 +1,53 @@
...
+        if [ "${{ inputs.type }}" != "Debug" ]; then
+          export OPT="-o"
+          export NSS_TARGET=Release
+        else
+          export NSS_TARGET=Debug
+        fi
+        $NSS_DIR/build.sh -g -Ddisable_tests=1 $OPT --static
+        echo "NSS_TARGET=$NSS_TARGET" >> "$GITHUB_ENV"
+        export NSS_OUT="$NSS_DIR/../dist/$NSS_TARGET"
```
martinthomson, line 46:
```suggestion
        NSS_OUT="$NSS_DIR/../dist/$NSS_TARGET"
```

I'm fairly sure that you don't need `export` if you are using the string in this shell.

neqo-crypto/build.rs
```diff
@@ -147,10 +148,9 @@ fn build_nss(dir: PathBuf) {
     let mut build_nss = vec![
         String::from("./build.sh"),
         String::from("-Ddisable_tests=1"),
+        String::from("--static"),
```
martinthomson, line 151: I don't understand this change.  Can you explain?

.github/workflows/bench.yml
```diff
@@ -8,8 +8,8 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
   TOOLCHAIN: nightly
-  RUSTFLAGS: -C link-arg=-fuse-ld=lld -C link-arg=-Wl,--no-rosegment
-  PERF_CMD: record -o perf.data -F997 --call-graph lbr -g
+  RUSTFLAGS: -C link-arg=-fuse-ld=lld -C link-arg=-Wl,--no-rosegment, -Cforce-frame-pointers=yes
+  PERF_CMD: record -o perf.data -F997 --call-graph fp -g
```
martinthomson, line 12: Unrelated?

neqo-crypto/build.rs
larseggert: I had queued up a change with a comment that I forgot to push. 
```
// Generate static libraries in addition to shared libraries.
```
Basically, `--static` seems to not only build static libraries, it seems to build static libraries **in addition** to dynamic ones. So no reason to not always use it.

.github/workflows/bench.yml
larseggert: Yes. Still mucking around with trying to find the best flags for `perf`, and don't want to do a whole other parallel CI run for that.

martinthomson (approved):

PR #1713, opened 2024-03-06
ci: Delay client start during transfer test

larseggert:
Hopefully addresses spurious "An existing connection was forcibly closed by the remote host" issue on Windows CI.


mxinden (approved):
Looks reasonable to me.

PR #1715, opened 2024-03-06
refactor(.github): rename comment workflow

mxinden:
Rename `Comment on the pull request` to `QUIC Network Simulator Comment`.

Might be a bit more intuitive, when seeing it next to the other workflows:

image


mxinden:
Not sure whether the pull request comment (https://github.com/mozilla/neqo/pull/1715#issuecomment-1981594845) adds more signal than noise in the long run. Happy to remove it, in case folks find it annoying. The important part for me is the CI check itself.

image

larseggert (approved):

PR #1717, opened 2024-03-07
ci: Hopefully fix the spurious Windows transfer test failure

larseggert:
And small changes, like not trying to use `lld` anymore on Windows, which never really worked.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -80,10 +80,9 @@ jobs:
         run: |
           echo "C:\\msys64\\usr\\bin" >> "$GITHUB_PATH"
           echo "C:\\msys64\\mingw64\\bin" >> "$GITHUB_PATH"
-          /c/msys64/usr/bin/pacman -S --noconfirm nsinstall lld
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

.github/workflows/check.yml
```diff
@@ -116,22 +111,21 @@ jobs:
       - name: Run client/server transfer
         run: |
           cargo +${{ matrix.rust-toolchain }} build $BUILD_TYPE --bin neqo-client --bin neqo-server
-          cargo +${{ matrix.rust-toolchain }} run $BUILD_TYPE --bin neqo-server -- $HOST:4433 &
+          target/$BUILD_DIR/neqo-server $HOST:4433 &
           PID=$!
-          # Give the server time to start.
-          sleep 3
```
mxinden, line 122: Why remove the sleep? If I understand the failure below, the client is starting up faster than the server.

https://github.com/mozilla/neqo/actions/runs/8374577016/job/22930136500?pr=1692

PR #1719, opened 2024-03-07
ci: Disable Rust cache

larseggert:
The caching seems to not lead to measurable decreases in CI times.


larseggert:
As much as I want it to, the cache really doesn't help.

Issue #1696, opened 2024-02-29
Merge neqo-client and neqo-server


mxinden:
`neqo-client` and `neqo-server` have a lot of logic in common, yet each duplicates it.

A couple of examples

- A lot of command line arguments are shared between client and server:
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-client/src/main.rs#L123-L219
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-server/src/main.rs#L92-L151
- Each has (at least) one `process` implementation calling `client.process` and handling its return value.
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-client/src/main.rs#L860-L881
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-server/src/main.rs#L646-L665
- Each has a similar run loop.
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-client/src/main.rs#L819-L858
    - https://github.com/mozilla/neqo/blob/ad027cfe2901dc98b3dae20574349f52621c5622/neqo-server/src/main.rs#L686-L707
- ...

Before making any major changes to `neqo-client` and `neqo-server`, like e.g. https://github.com/mozilla/neqo/issues/1693, I thus suggest merging the two into one crate, and thereby deduplicating the shared logic.

Suggestion:
- Introduce `neqo-bin/` crate.
- Move `neqo-common/src/udp.rs` into `neqo-bin/src/udp.rs`.
- Consolidate shared logic between `neqo-client` and `neqo-server` in `neqo-bin/src/lib.rs` and submodules.
- Move `neqo-client/src/main.rs` into `neqo-bin/bin/client.rs` and `neqo-server/src/main.rs` into `neqo-bin/bin/server.rs`.

What do folks think? Why have they been split into two separate crates thus far?

larseggert:
Done via #1724

Issue #1403, opened 2023-01-16
Unify some of the configuration for the client and server tools


martinthomson:
As I noted in #1402, there is a lot of common stuff between the client and server command-line tools.  It would be good if we could unify some of this.

It might make sense to move this to a module under the mainline code (both neqo-transport and neqo-http3), covered by feature flags that the command-line tools enable.  Then, a standard build won't pull in a bunch of dependencies, but we have a single place for this stuff.

larseggert:
CC @mxinden, since you're touching that code now.

mxinden:
Close here in favor of the recent https://github.com/mozilla/neqo/issues/1696?

(Sorry for creating a ~duplicate in the first place.)

larseggert:
Sure.

PR #1724, opened 2024-03-08
refactor: merge neqo-client and neqo-server into neqo-bin

mxinden:
This commit merges the `neqo-client` and `neqo-server` crates into a single new `neqo-bin` crate. In addition it de-duplicates command line argument definitions by introducing a `SharedArgs` `struct`.

Future commits will consolidate more shared code into `neqo-bin/src/lib.rs` and its sub-modules.

Note that binary names are unchanged. E.g. one can still run the client via `cargo run --bin neqo-client`. See `neqo-bin/Cargo.toml` for binary declarations.

---

Part of https://github.com/mozilla/neqo/issues/1696. See also past issue https://github.com/mozilla/neqo/issues/1403.

Small step forward already: 
image


larseggert (approved):
LGTM, but would like to hear from the others.

martinthomson (approved):

PR #1725, opened 2024-03-08
Priority headers should be set by the application

valenting:
Instead of setting the priority headers in neqo, the headers should be provided by the application.

See related https://phabricator.services.mozilla.com/D201265

### DIFF
--- a/neqo-http3/tests/priority.rs
+++ b/neqo-http3/tests/priority.rs
@@ -68,7 +68,7 @@ fn priority_update() {
             Instant::now(),
             "GET",
             &("https", "something.com", "/"),
-            &[],
+            &[Header::new("priority", "u=4,i")],
             Priority::new(4, true),
         )
         .unwrap();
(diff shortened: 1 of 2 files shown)


neqo-http3/tests/priority.rs
```diff
@@ -68,7 +68,7 @@ fn priority_update() {
             Instant::now(),
             "GET",
             &("https", "something.com", "/"),
-            &[],
+            &[Header::new("priority", "u=4,i")],
```
larseggert, line 71: Is the test verifying that this gets used somewhere?

neqo-http3/tests/priority.rs
```diff
@@ -129,7 +129,7 @@ fn priority_update_dont_send_for_cancelled_stream() {
             Instant::now(),
             "GET",
             &("https", "something.com", "/"),
-            &[],
+            &[Header::new("priority", "u=5")],
```
larseggert, line 132: Ditto.

neqo-http3/tests/priority.rs
valenting: Yes, line 96 in expected_headers.

neqo-http3/tests/priority.rs
valenting: This test doesn't check the headers as far as I can tell, but I put them here for consistency.

larseggert:
LGTM, but I'd like to see another review, since the related Fx parts are new to me.

KershawChang (approved):
Looks good.

PR #1726, opened 2024-03-08
ci: Try `hg` and then try `git` to download NSPR and NSS

larseggert:
Because on the bencher, `git` now has issues. You can't win.


larseggert:
Merging this to unbreak CI.

PR #1727, opened 2024-03-10
refactor(client): de-duplicate process & run and split into h3 & h09 

mxinden:
The Neqo Client binary supports both http3 and http09 (prev. "old").

Before this commit both the http3 and the http09 implementation had their own `run` and `process` `fn`, orchestrating the interaction between handler, client and I/O. While similar, they had subtle differences e.g. when to terminate.

This commit splits the http3 and http09 specific logic into two separate modules, but extracts duplicate logic (e.g. `run` and `process`) into the shared root module.

---

**For reviewers**: I suggest reviewing the two commits individually. The first simply moves `bin/client.rs` to `bin/client/main.rs`. It is done in a separate commit in order for git to recognize the move as a move and not as a remove and an add, thus showing a proper diff.

Part of #1696.

Simplifying `process` and `run` will make https://github.com/mozilla/neqo/issues/1693 easier to implement.


neqo-bin/src/bin/client.rs
mxinden, line 1: > For reviewers: I suggest reviewing the two commits individually. The first simply moves bin/client.rs to bin/client/main.rs. It is done in a separate commit in order for git to recognize the move as a move and not as a remove and an add, thus showing a proper diff.

neqo-bin/src/bin/client/main.rs
```diff
@@ -0,0 +1,541 @@
...
+    local_addr: SocketAddr,
+    socket: &'a mut udp::Socket,
+    client: H::Client,
+    handler: H,
+    timeout: Option<Pin<Box<Sleep>>>,
+    args: &'a Args,
+}
+
+impl<'a, H: Handler> Runner<'a, H> {
+    async fn run(mut self) -> Res<Option<ResumptionToken>> {
```
mxinden, line 339: With this commit, the Neqo client implementation has a single `run` function, not one for `http3` and one for `http09`.

neqo-bin/src/bin/client/main.rs
```diff
@@ -0,0 +1,541 @@
...
+                    args: &args,
+                    client,
+                    handler,
+                    local_addr: real_local,
+                    socket: &mut socket,
+                    timeout: None,
+                }
+                .run()
+                .await?
+            };
```
mxinden, line 536: `http3` and `http09` expose the same methods (`create_client` and `Handler::new`) and can be _run_ with the same `Runner`.

larseggert (approved):

PR #1728, opened 2024-03-11
fix: Disable generation of criterion HTML graphs

larseggert:
Because we pin the benchmark runs to single cores, and criterion hence runs the report generation also on those cores, and based on `top` output it appears as this is parallelized and hence may interfere with the benchmark runs.


martinthomson (approved):
A bit gross, but solid.

PR #1729, opened 2024-03-11
ci: Safe commenting on PRs

larseggert:
This generalizes and refactors what the QNS workflow did.

### DIFF
--- a/.github/actions/quic-interop-runner/action.yml
+++ b/.github/actions/quic-interop-runner/action.yml
@@ -88,24 +88,20 @@ runs:
         name: logs
         path: quic-interop-runner/logs
 
-    # This action might be running off of a fork and would thus not have write
-    # permissions on the origin repository. In order to allow a separate
-    # priviledged action to post a comment on a pull request, upload the
-    # necessary metadata.
-    - name: store comment-data
-      shell: bash
-      if: github.event_name == 'pull_request'
-      env:
-        PULL_REQUEST_NUMBER: ${{ github.event.number }}
+    - name: Format GitHub comment
       run: |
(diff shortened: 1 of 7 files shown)


.github/actions/quic-interop-runner/action.yml
```diff
@@ -88,24 +88,20 @@ runs:
...
+        cat quic-interop-runner/summary >> comment
+        echo '```' >> comment
+        echo '' >> comment
+      shell: bash
 
-    - name: Upload comment data
-      uses: actions/upload-artifact@v4
-      if: github.event_name == 'pull_request'
+    - name: Export PR comment data
+      if: ${{ github.event_name == 'pull_request' }}
```
mxinden, line 102: Each step in `pr-comment-data-export` has this `if` already. Can we remove it either here or in the action?

mxinden (approved):
Thank you for pulling this into separate actions! Overall looks good to me. Couple of nits / questions.

.github/workflows/qns-comment.yml
```diff
@@ -17,40 +17,8 @@ jobs:
     permissions:
       pull-requests: write
     runs-on: ubuntu-latest
-    if: >
-      github.event.workflow_run.event == 'pull_request' &&
-      github.event.workflow_run.conclusion == 'failure'
+    if: github.event.workflow_run.event == 'pull_request'
```
mxinden, line 20: In the case of qns, the comment should still only be printed on `failure`, right?

larseggert: Yes, this needs to be restored - I was just trying to make it generate any output.

PR #1732, opened 2024-03-11
ci: Benchmark NewReno and Cubic

larseggert:


.github/workflows/bench.yml
```diff
@@ -80,19 +80,26 @@ jobs:
...
+              cd server; \
+              taskset -c 0 nice -n -20 \
+              cargo "+$TOOLCHAIN" flamegraph -v -c "$PERF_CMD" \
+                --bin neqo-server -- --db ../test-fixture/db --cc "$CC" "$HOST:4433" || true; } &
+            mkdir -p client; \
+              cd client; \
+              time taskset -c 1 nice -n -20 \
+              cargo "+$TOOLCHAIN" flamegraph -v -c "$PERF_CMD" \
+                --bin neqo-client -- --output-dir . --cc "$CC" "https://$HOST:4433/$SIZE"
+            killall -INT neqo-server
```
martinthomson, line 94: Can we please capture `$!` instead?  killall is a blunt instrument.

.github/workflows/bench.yml
```diff
@@ -80,19 +80,26 @@ jobs:
...
+            cd ${{ github.workspace }}
+            [ "$(wc -c < client/"$SIZE")" -eq "$SIZE" ] || exit 1
+            while [ ! -f server/flamegraph.svg ]; do sleep 1; done
+            for PEER in client server; do
+              mv "$PEER/flamegraph.svg" "$PEER-$CC.svg"
```
martinthomson, line 99: You can use `cargo flamegraph -o "$PEER-$CC.svg"` for this instead.

.github/workflows/bench.yml
```diff
@@ -80,19 +80,26 @@ jobs:
...
+              mv "$PEER/perf.data" "$PEER-$CC.perf"
```
martinthomson, line 100: You might also use `-o` with perf to set where this file goes.  Though that means moving that option out of `$PERF_CMD`

.github/workflows/bench.yml
```diff
@@ -105,19 +112,23 @@ jobs:
       - name: Convert for profiler.firefox.com
         run: |
           perf script -i perf.data -F +pid > transfer.perf &
-          perf script -i client/perf.data -F +pid > client.perf &
-          perf script -i server/perf.data -F +pid > server.perf &
-          wait
           mv flamegraph.svg transfer.svg
```
martinthomson, line 115: As above, it might be better to have this file created in the right place to start with.

.github/workflows/bench.yml
larseggert: I tried that initially. The problem is that `$!` has the pid of the outer `taskset` process, not that of the innermost `neqo-server` one. And killing `taskset` doesn't make `perf` dump its data correctly, whereas killing the innermost `neqo-server` does.

I can experiment more with this if you like.

martinthomson: Ah, I tried with taskset on my machine and it seemed to work fine if you just do `taskset ... whatever &`.  Maybe all you need to do is drop the curly braces.  The braces will mean that `$!` will be the bash wrapper process.  Swap `cd` for `pushd`/`popd` and it should work out.

larseggert: Let's try that.

larseggert: I tried, but even in the current version, which has no curly braces, etc. anymore, a `kill -INT $!` isn't working. I at least switched it to a `pkill` call that only kills one (the oldest) `neqo-server`.

larseggert:
Merging this into #1750

PR #1736, opened 2024-03-12
refactor(udp): move udp mod from neqo-common to neqo-bin

mxinden:
The `udp` module is only used in the `neqo_bin` crate, more specifically by the Neqo client and server implementation. This commit moves the `udp` module to `neqo-bin`.

---

Depends on https://github.com/mozilla/neqo/pull/1727/. Thus opening as draft for now.


larseggert:
I'll unbreak mutants after dinner

neqo-bin/src/udp.rs
```diff
@@ -61,7 +60,7 @@ impl Socket {
         let transmit = Transmit {
             destination: d.destination(),
             ecn: EcnCodepoint::from_bits(Into::<u8>::into(d.tos())),
-            contents: d.into_data().into(),
+            contents: Vec::from(d).into(),
```
larseggert, line 63: Is this part of the move or another change?

neqo-common/src/datagram.rs
```diff
@@ -83,6 +77,12 @@ impl std::fmt::Debug for Datagram {
     }
 }
 
+impl From<Datagram> for Vec<u8> {
+    fn from(datagram: Datagram) -> Self {
+        datagram.d
+    }
+}
+
```
larseggert, line 85: Is this part of the move or another change?

mxinden: The `udp` module needs to convert a `Datagram` to a `Vec`. Previously this would have been done via the `pub(crate)` `Datagram::into_data`.

Given that `udp` module is moved outside of `neqo-common` crate now, `pub(crate)` `into_data` is changed into a `pub` `From<Datagram> for Vec<u8>` implementation. See also https://github.com/mozilla/neqo/pull/1695#discussion_r1508330175.

Long story short, this just exposes `into_data` beyond the crate and implements it as a `From` instead.

neqo-bin/src/udp.rs
mxinden: `Datagram::into_data` is changed to a `From<Datagram> for Vec<u8>` (see comment below). Instead of calling `into_data`, this line first leverages the new `From` implementation and then, as before, converts it into a `Bytes` (`into` at the end).

None of these methods allocate.

larseggert (approved):

PR #1739, opened 2024-03-13
build: Use system NSS when possible

### LINKED ISSUE #1711: Use system-installed NSS dependeny
Rather than checking out and building our own copy of NSS during the build of neqo, we should detect and use a system-installed NSS. This will require some heavy surgery to `neqo-crypto/build.rs` but should really simplify things.

larseggert:
When `NSS_DIR` is **not** set, we now attempt to use the system NSS, and check whether it is recent enough for us. The current NSS minimum version is in `neqo-crypto/min_version.txt`, so it can be accessed by Rust (via `neqo-crypto/src/min_version.rs`) and by the CI scripts.

Fixes #1711


neqo-crypto/build.rs
```diff
@@ -297,37 +187,7 @@ fn build_bindings(base: &str, bindings: &Bindings, flags: &[String], gecko: bool
...
-        dynamic_link();
-    }
-
-    let mut flags: Vec<String> = Vec::new();
-    for i in includes {
-        flags.push(String::from("-I") + i.to_str().unwrap());
-    }
-
-    flags
+    pkg_config()
```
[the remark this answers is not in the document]
larseggert: We could make it contingent on whether `NSS_DIR` is defined. (If it is, build and use that one.)

neqo-crypto/build.rs
```diff
@@ -90,150 +78,52 @@ fn setup_clang() {
...
+    );
+    if let Some(minor) = v.next() {
+        let minor = minor
+            .trim_end()
+            .parse::<u32>()
+            .expect("NSS minor version is not a number");
         assert!(
-            !path.is_relative(),
-            "The NSS_DIR environment variable is expected to be an absolute path."
+            minor >= 62,
```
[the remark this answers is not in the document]
larseggert: Good suggestions, will look into it. First trying to get CI building again.

.github/actions/nss/action.yml
```diff
@@ -16,35 +16,68 @@ inputs:
...
+      run: |
+        if ! command -v pkg-config &> /dev/null; then
+          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          exit 0
+        fi
+        if ! pkg-config --exists nss; then
+          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          exit 0
+        fi
+        NSS_VERSION=$(pkg-config --modversion nss)
```
martinthomson, line 30:
```suggestion
        NSS_VERSION="$(pkg-config --modversion nss)"
```

.github/actions/nss/action.yml
```diff
@@ -16,35 +16,68 @@ inputs:
...
+        if [ "$?" -ne 0 ]; then
+          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          exit 0
+        fi
+        NSS_MAJOR=$(echo "$NSS_VERSION" | cut -d. -f1)
```
martinthomson, line 35: So there is also `read NSS_MAJOR NSS_MINOR NSS_PATCH < <(echo "${NSS_VERSION//./ }")`.  ¯\\\_(ツ)\_/¯

neqo-crypto/build.rs
```diff
@@ -295,6 +298,53 @@ fn build_bindings(base: &str, bindings: &Bindings, flags: &[String], gecko: bool
...
+    let modversion = Command::new("pkg-config")
+        .args(["--modversion", "nss"])
+        .output()
+        .expect("pkg-config reports NSS as absent")
+        .stdout;
+    let modversion = String::from_utf8(modversion).expect("non-UTF8 from pkg-config");
+    let modversion = modversion.trim().to_string();
+    // The NSS version number does not follow semver numbering, because it omits the patch version
+    // when that's 0. Deal with that.
+    let modversion_for_cmp = if modversion.split('.').count() == 2 {
```
martinthomson, line 311: Rather than split, because you don't need the slices, you can do this:

```suggestion
    let modversion_for_cmp = if modversion.chars().filter(|c| c == '.').count() == 1 {
```

neqo-crypto/build.rs
```diff
@@ -295,6 +298,53 @@ fn build_bindings(base: &str, bindings: &Bindings, flags: &[String], gecko: bool
...
+        modversion.clone() + ".0"
+    } else {
+        modversion.clone()
```
martinthomson, line 314: You shouldn't need to clone here.  You already have an owned copy (that's what the `to_string()` above did.  That said, I would not use `to_string()` above, instead making a copy here instead.

neqo-crypto/build.rs
```diff
@@ -406,6 +456,8 @@ fn setup_for_gecko() -> Vec<String> {
 fn main() {
     let flags = if cfg!(feature = "gecko") {
         setup_for_gecko()
+    } else if env::var("NSS_DIR").is_err() {
+        pkg_config()
     } else {
         setup_standalone()
```
martinthomson, line 462:
```suggestion
    } else if let Ok(nss_dir) = env::var("NSS_DIR") {
        setup_standalone(nss_dir)
    } else {
        pkg_config()
```

martinthomson (commented):
I think that you can cut most of `fn nss_dir()` now that you won't be hitting it as often.  Just pass the env variable in to `setup_standalone()`

larseggert:
> I think that you can cut most of `fn nss_dir()` now that you won't be hitting it as often. Just pass the env variable in to `setup_standalone()`

`nss_dir()` has the code that checks out NSS and NSPR if there is no copy already. Are you saying that should be cut, i.e., will we require people to do their own checkouts?

neqo-crypto/build.rs
```diff
@@ -295,7 +298,54 @@ fn build_bindings(base: &str, bindings: &Bindings, flags: &[String], gecko: bool
...
+    // The NSS version number does not follow semver numbering, because it omits the patch version
+    // when that's 0. Deal with that.
+    let modversion_for_cmp = if modversion.chars().filter(|c| *c == '.').count() == 1 {
+        modversion.to_owned() + ".0"
+    } else {
+        modversion.to_owned()
+    };
+    let modversion_for_cmp =
+        Version::parse(&modversion_for_cmp).expect("NSS version not in semver format");
+    let version_req = VersionReq::parse(&(">=".to_owned() + MINIMUM_NSS_VERSION.trim())).unwrap();
```
martinthomson, line 318:
```suggestion
    let version_req = VersionReq::parse(&format!(">={}", MINIMUM_NSS_VERSION.trim())).unwrap();
```

neqo-crypto/build.rs
```diff
@@ -12,8 +12,11 @@ use std::{
 };
 
 use bindgen::Builder;
+use semver::{Version, VersionReq};
 use serde_derive::Deserialize;
 
+include!("min_version.rs");
```
martinthomson, line 18: I think that you can use `mod min_version; use min_version::MINIMUM_NSS_VERSION;` rather than relying on `include!()`.

neqo-crypto/src/lib.rs
```diff
@@ -59,7 +59,7 @@ pub use self::{
     ssl::Opt,
 };
 
-const MINIMUM_NSS_VERSION: &str = "3.97";
+include!("../min_version.rs");
```
martinthomson, line 62: Again, I think that it is better to use:

```rust
#[path = "../min_version.rs"]
mod min_version;
use min_version::MINIMUM_NSS_VERSION;
```

neqo-crypto/min_version.rs
```diff
@@ -0,0 +1,9 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
martinthomson, line 1: See other comments, but this *might* be better off if it were moved to src/

.github/actions/nss/action.yml
```diff
@@ -16,35 +16,68 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Check system NSS version
+      shell: bash
+      run: |
+        if ! command -v pkg-config &> /dev/null; then
+          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          exit 0
+        fi
```
martinthomson, line 25: For testing the availability of commands, I use `hash pkg-config 2>/dev/null`.  YMMV.  Note that just invoking pkg-config without testing for it first will also achieve the same:

```suggestion
```

martinthomson (approved):

martinthomson:
OK, so approvals with comments get merged.  That's not ideal.

martinthomson:
@larseggert should we require that conversations are resolved before merging?

image

larseggert:
@martinthomson we can turn that on, but then we should probably turn **off** `Require approval of the most recent reviewable push`, because otherwise a re-review would be required just as it is now if one had chosen "request changes"?

neqo-crypto/src/lib.rs
larseggert: Do you have a preference in which directory `min_version.txt` and `min_version.rs` live? I wasn't sure whether to put one or both under `src`.

.github/actions/nss/action.yml
larseggert: Just invoking it if it isn't installed will fail with an error code != 0 and so stop CI.

larseggert:
> I think that you can cut most of `fn nss_dir()` now that you won't be hitting it as often. Just pass the env variable in to `setup_standalone()`
 
@martinthomson: `nss_dir()` has the code that checks out NSS and NSPR if there is no copy already. Are you saying that should be cut, i.e., will we require people to do their own checkouts?

martinthomson:
> turn off Require approval of the most recent reviewable push

Yeah, I think that we can trust people to exercise their judgment on the small stuff.

> will we require people to do their own checkouts?

Yeah.  You will observe that you can't reach that code unless you set NSS_DIR but haven't got a copy of NSS.  Given that the main reason to use NSS_DIR is to point to a copy you are working on, that seems like a good trade to me.

PR #1744, opened 2024-03-14
fix(log): pad INIT_ONCE to cache line size to prevent false sharing

mxinden:
Just a wild guess. Thus far unable to reproduce https://github.com/mozilla/neqo/issues/1684 on my local machine (12th Gen Intel(R) Core(TM) i9-12900HK). Here to trigger CI.


mxinden:
No visible impact.

Client profiles compared:

- Recent `main` https://share.firefox.dev/3Iz2eSP
- This pull request https://share.firefox.dev/3v16B65

larseggert:
I would have been surprised if this had improved things, because we are single-threaded during the benchmark run and even pin the run to a core.

PR #1743, opened 2024-03-14
feat(.github/interop): use markdown table

mxinden:
1. Provide `--markdown` flag to QUIC Interop Runner to get markdown formatted table.
2. Ignore all log output, but table.

---

Previous work: https://github.com/quic-interop/quic-interop-runner/pull/372

PR #1747, opened 2024-03-15
refactor(github/interop): use grep instead of awk

mxinden:
Follow-up to https://github.com/mozilla/neqo/pull/1743#discussion_r1525498120.

### DIFF
--- a/.github/actions/quic-interop-runner/action.yml
+++ b/.github/actions/quic-interop-runner/action.yml
@@ -92,8 +92,8 @@ runs:
       run: |
         echo '[**QUIC Interop Runner**](https://github.com/quic-interop/quic-interop-runner)' >> comment
         echo '' >> comment
-        # Ignore all, but table, which starts with "|:--".
-        cat quic-interop-runner/summary | awk '/^\|:--/{flag=1} flag' >> comment
+        # Ignore all, but table, which starts with "| :--" or "|:--".
+        grep -A 999999 -E '^\|\s?:--' quic-interop-runner/summary >> comment
         echo '' >> comment
       shell: bash
 
--- a/.github/actions/quic-interop-runner/action.yml
+++ b/.github/actions/quic-interop-runner/action.yml
@@ -92,8 +92,8 @@ runs:
       run: |
         echo '[**QUIC Interop Runner**](https://github.com/quic-interop/quic-interop-runner)' >> comment
         echo '' >> comment
-        # Ignore all, but table, which starts with "| :--" or "|:--".
-        grep -A 999999 -E '^\|\s?:--' quic-interop-runner/summary >> comment
+        # Ignore all, but table, which starts with "|".
+        grep -E '^\|' quic-interop-runner/summary >> comment
         echo '' >> comment
       shell: bash
 


.github/actions/quic-interop-runner/action.yml
```diff
@@ -92,8 +92,8 @@ runs:
       run: |
         echo '[**QUIC Interop Runner**](https://github.com/quic-interop/quic-interop-runner)' >> comment
         echo '' >> comment
-        # Ignore all, but table, which starts with "|:--".
-        cat quic-interop-runner/summary | awk '/^\|:--/{flag=1} flag' >> comment
+        # Ignore all, but table, which starts with "| :--" or "|:--".
+        grep -A 999999 -E '^\|\s?:--' quic-interop-runner/summary >> comment
```
larseggert, line 96: Why the `-A` ugliness? There should be a pattern to grep the table out that doesn't need this.

mxinden: With `-A SOME_LARGE_NUMBER` `grep` returns all lines after the first match. This is obviously a hack. As far as I know there is no other option when using `grep`.

An alternative would be to match all lines starting with a `^\|`. Not sure matching on a single char is a safe bet in the long run. Note you can't match on `^\|\s`, as prettytable formats small columns without a space in the separator line, i.e. `|:--:|` instead of `| :--: |`.

On the one hand I find the previous `awk` command less of a hack. On the other, I would expect folks to be more familiar with `grep`. 

I don't have a strong opinion here. Nor do I think the choice of `awk` or `grep` is particularly relevant. Just a follow-up to https://github.com/mozilla/neqo/pull/1743#discussion_r1525498120.

larseggert: > Note you can't match on `^\|\s`, as prettytable formats small columns without a space in the separator line, i.e. `|:--:|` instead of `| :--: |`.

`^\| ?:` https://regex101.com/

mxinden: Sorry, that sentence could have been more clear.

> `^\| ?:`

Yes, that is what the change is doing, though with `\s` instead of ` ` (see above `-E '^\|\s?:--'`).

Note however that this would only match the first row of the table, given that only the header row has a `:`.

Either we:

1. Keep the `awk` command.
2. Move to `-A 999999` hack.
3. Shorten the Regex to `^\| ?` (note the missing `:` or `:--`) in order to match all rows of the table. Though at the risk that any other line starting with a `|` is matched as well.

I am fine with either option.

larseggert: Could you post examples of the different table lines? I'll make a  regex.

larseggert: @mxinden ping

mxinden: Sorry for the delay. Two examples:

```
Saving logs to logs.
Run took 0:03:27.581731
Server: neqo-latest. Client: neqo-latest. Running test case: handshake
Server: quic-go. Client: neqo-latest. Running test case: handshake
Server: ngtcp2. Client: neqo-latest. Running test case: handshake
Server: neqo. Client: neqo-latest. Running test case: handshake
Server: msquic. Client: neqo-latest. Running test case: handshake
Server: neqo-latest. Client: quic-go. Running test case: handshake
Server: neqo-latest. Client: ngtcp2. Running test case: handshake
Server: neqo-latest. Client: neqo. Running test case: handshake
Server: neqo-latest. Client: msquic. Running test case: handshake
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
|             | neqo-latest | quic-go | ngtcp2 | neqo | msquic |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
| neqo-latest |     ✓(H)    |   ✓(H)  |  ✓(H)  | ✓(H) |  ✓(H)  |
|             |     ?()     |   ?()   |  ?()   | ?()  |  ?()   |
|             |     ✕()     |   ✕()   |  ✕()   | ✕()  |  ✕()   |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
|   quic-go   |     ✓(H)    |         |        |      |        |
|             |     ?()     |         |        |      |        |
|             |     ✕()     |         |        |      |        |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
|    ngtcp2   |     ✓(H)    |         |        |      |        |
|             |     ?()     |         |        |      |        |
|             |     ✕()     |         |        |      |        |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
|     neqo    |     ✓(H)    |         |        |      |        |
|             |     ?()     |         |        |      |        |
|             |     ✕()     |         |        |      |        |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
|    msquic   |     ✓(H)    |         |        |      |        |
|             |     ?()     |         |        |      |        |
|             |     ✕()     |         |        |      |        |
| :---------: | :---------: | :-----: | :----: | :--: | :----: |
```

https://github.com/mozilla/neqo/actions/runs/8293337157/job/22696235816?pr=1747

```
Saving logs to logs_2024-03-25T09:25:04.
Server: quic-go. Client: quic-go. Running test case: handshake
Run took 0:00:23.982204
|:-------:|:-------:|
|         | quic-go |
|:-------:|:-------:|
| quic-go |   ✓(H)  |
|         |   ?()   |
|         |   ✕()   |
|:-------:|:-------:|
```

martinthomson: Or just `^\|`

larseggert: Fine with me, too. @mxinden was concerned it might be too general, but we can always revise the pattern if there ever is an issue.

larseggert (approved):

PR #1750, opened 2024-03-15
ci: Compare performance to msquic

larseggert:
In progress


larseggert:
I looked at why we're so much slower then msquic, and the neqo server spends an insane amount of time in `neqo-common::timer`:
```
       │      /// Take the next item, unless there are no items with                                                   ▒
       │      /// a timeout in the past relative to `until`.                                                           ▒
       │      pub fn take_next(&mut self, until: Instant) -> Option<T> {                                               ▒
       │      for i in 0..self.items.len() {                                                                           ▒
       │      let idx = self.bucket(i);                                                                                ▒
       │      if !self.items[idx].is_empty() && self.items[idx][0].time <= until {                                     ▒
       │        test      %rbx,%rbx                                                                                    ▒
       │        jne       2444                                                                                         ▒
       │        jmp       23f0                                                                                         ▒
       │      neqo_common::timer::Timer<T>::bucket:                                                                    ▒
       │        nop                                                                                                    ▒
       │      (self.cursor + delta) % self.items.len()                                                                 ▒
  0.03 │2430:   mov       %ecx,%eax                                                                                    ▒
       │        xor       %edx,%edx                                                                                    ▒
 29.50 │        div       %esi                                                                                         ▒
       │      <usize as core::slice::index::SliceIndex<[T]>>::index:                                                   ▒
  1.27 │        lea       (%rdx,%rdx,2),%rax                                                                           ▒
       │      alloc::vec::Vec<T,A>::len:                                                                               ▒
 22.02 │        mov       0x10(%r8,%rax,8),%rbx                                                                        ▒
       │      neqo_common::timer::Timer<T>::take_next:                                                                 ▒
       │      if !self.items[idx].is_empty() && self.items[idx][0].time <= until {                                     ▒
  0.04 │        test      %rbx,%rbx                                                                                    ▒
  7.64 │        je        23f0                                                                                         ▒
  ```
 ```
        │      neqo_common::timer::Timer<T>::bucket:                                                                    ▒
       │      (self.cursor + delta) % self.items.len()                                                                 ▒
  0.11 │2bf0:   mov       %ecx,%eax                                                                                    ▒
  0.09 │        xor       %edx,%edx                                                                                    ▒
  8.55 │        div       %edi                                                                                         ▒
       │      <usize as core::slice::index::SliceIndex<[T]>>::index:                                                   ▒
  0.12 │        lea       (%rdx,%rdx,2),%rax                                                                           ▒
       │      core::slice::<impl [T]>::first:                                                                          ▒
  4.36 │        inc       %rcx                                                                                         ▒
  4.53 │        cmpq      $0x0,0x10(%rsi,%rax,8)                                                                       ▒
  2.20 │        jne       2c32                                                                                         ▒
  ```
(Numbers in the first column are percent of total cycles spent in that line.)

File: .github/workflows/bench.yml [hunk 1]
```diff
@@ -155,9 +155,9 @@ jobs:
           sed 's/`//g' steps.md |\
             awk '!/^\| Command/ || !c++' |\
             awk '!/^\|:/ || !c++' |\
-            sed -e 's/^\|:/\|:---\|:---\|:---\|:/g' |\
-            sed -e 's/,/ \| /g' |\
-            sed -e 's/^\| Command/\| Client \| Server \| CC \| Pacing/g' >> comparison.md
+            sed -E 's/^\|:/\|:---\|:---\|:---\|:/g' |\
+            sed -E 's/,/ \| /g' |\
+            sed -E 's/^\| Command/\| Client \| Server \| CC \| Pacing/g' >> comparison.md
```
martinthomson, line 160: Invoking sed multiple times isn't necessary.  You can use ';'.

.github/workflows/bench.yml
```diff
@@ -155,9 +155,9 @@ jobs:
           sed 's/`//g' steps.md |\
             awk '!/^\| Command/ || !c++' |\
             awk '!/^\|:/ || !c++' |\
```
martinthomson, line 157: Is this `sed -e '/^\|:/,$ { 1d; }` ?

larseggert: Possibly? I tried various things but `sed` not being really cross-platform made me switch to `awk`.

KershawChang (approved):

PR #1752, opened 2024-03-16
feat: Use `std::io::BufWriter` with a 64KB buffer for dumping data

larseggert:
Let's see if this makes things faster.

### DIFF
--- a/neqo-bin/src/bin/client/http09.rs
+++ b/neqo-bin/src/bin/client/http09.rs
@@ -10,7 +10,7 @@ use std::{
     cell::RefCell,
     collections::{HashMap, VecDeque},
     fs::File,
-    io::Write,
+    io::{BufWriter, Write},
     net::SocketAddr,
     path::PathBuf,
     rc::Rc,
@@ -26,7 +26,7 @@ use neqo_transport::{
 use url::Url;
 
 use super::{get_output_file, Args, KeyUpdateState, Res};
-use crate::qlog_new;
+use crate::{qlog_new, BUFWRITER_BUFFER_SIZE};
 
 pub struct Handler<'a> {
     streams: HashMap<StreamId, Option<File>>,
@@ -219,7 +219,7 @@ impl<'b> Handler<'b> {
         client: &mut Connection,
         stream_id: StreamId,
         output_read_data: bool,
-        maybe_out_file: &mut Option<File>,
+        maybe_out_file: &mut Option<BufWriter<File>>,
     ) -> Res<bool> {
         let mut data = vec![0; 4096];
         loop {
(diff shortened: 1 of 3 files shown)


mxinden (commented):
Missing a `<BufWriter as Write>::flush` on `fin`, right? Otherwise some data might remain in the buffer and measurements are hard to compare.

larseggert:
Probably, although the docs say it is called (just if there is an error that remains unhandled.)

neqo-bin/src/bin/client/http3.rs
```diff
@@ -300,8 +301,12 @@ impl StreamHandler for DownloadStreamHandler {
...
         }
 
-        if fin && self.out_file.is_none() {
-            println!("<FIN[{stream_id}]>");
+        if fin {
+            if self.buf_writer.is_none() {
+                println!("<FIN[{stream_id}]>");
+            } else {
+                self.buf_writer.take().unwrap().flush()?;
+            }
```
mxinden, line 309:
```suggestion
            if let Some(buf_writer) = self.buf_writer.take() {
                buf_writer.flush()?;
            } else {
                println!("<FIN[{stream_id}]>");
            }
```

mxinden (approved):

larseggert:
That whole "write to file" functionality could be factored out.

neqo-bin/src/bin/client/http09.rs
```diff
@@ -253,15 +253,21 @@ impl<'b> Handler<'b> {
                 return Ok(());
             }
             Some(maybe_out_file) => {
+                let mut buf_writer = maybe_out_file
+                    .take()
+                    .map(|file| BufWriter::with_capacity(BUFWRITER_BUFFER_SIZE, file));
```
mxinden, line 258: Why construct a new `BufWriter` on each `read` call? Shouldn't the `BufWriter` be wrapping the `File` for the lifetime of the `File`? That would be consistent with the `http3.rs` implementation.

neqo-bin/src/bin/client/http3.rs
```diff
@@ -255,8 +255,10 @@ impl StreamHandlerType {
         match handler_type {
             Self::Download => {
                 let out_file = get_output_file(url, &args.output_dir, all_paths);
+                let buf_writer =
+                    out_file.map(|file| BufWriter::with_capacity(BUFWRITER_BUFFER_SIZE, file));
```
mxinden, line 259: How about moving this into `get_output_file`. `get_output_file` is called by both `http3` and `http09`, thus no need for duplication.

mxinden (commented):
I combined the suggestions in https://github.com/larseggert/neqo/pull/19.

larseggert:
Thanks for that, it's the better approach. Can you point the PR at the main repo?

mxinden:
> Thanks for that, it's the better approach. Can you point the PR at the main repo?

:+1: see https://github.com/mozilla/neqo/pull/1756.

larseggert:
Overtaken by #1756

PR #1753, opened 2024-03-16
fix(bin): enable pacing on client & server by default

mxinden:
_Updated version:_

Before https://github.com/mozilla/neqo/pull/1724 `neqo-server` would always run with `ConnectionParameters::pacing` `true`. `neqo-client` would provide a command line flag `--pacing` which defaulted to `false`.

https://github.com/mozilla/neqo/pull/1724 consolidated the `QuicParameters` `struct` from `neqo-client` and `neqo-server`. Since both `neqo-client` and `neqo-server` default to `pacing` `false`.

This commit replaces the `--pacing` flag with `--no-pacing`. Pacing is enabled by default on both `neqo-server` and `neqo-client`.

### DIFF
--- a/neqo-bin/src/bin/client/main.rs
+++ b/neqo-bin/src/bin/client/main.rs
@@ -437,8 +437,6 @@ async fn main() -> Res<()> {
     init();
 
     let mut args = Args::parse();
-    // By default, disable pacing on client.
-    args.shared.quic_parameters.pacing.get_or_insert(false);
     args.update_for_tests();
 
     let urls_by_origin = args
--- a/neqo-bin/src/bin/client/main.rs
+++ b/neqo-bin/src/bin/client/main.rs
@@ -437,6 +437,8 @@ async fn main() -> Res<()> {
     init();
 
     let mut args = Args::parse();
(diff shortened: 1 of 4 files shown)


larseggert:
It's probably an oversight that the two have different defaults? I think the defaults should be the same, for consistency.

martinthomson (commented):
I found https://github.com/clap-rs/clap/issues/815#issuecomment-1808282085 for doing negation.  So you can have --pacing and --no-pacing options.  Not sure what this does with a default of true though...

mxinden:
> It's probably an oversight that the two have different defaults? I think the defaults should be the same, for consistency.

I suggest making `pacing=true` the default on both client and server then. This is also consistent with the default of `ConnectionParameters`:

https://github.com/mozilla/neqo/blob/ef5caeb46b510f89cd6f8668a3bf31a2d597cbd2/neqo-transport/src/connection/params.rs#L105

We can then expose a `--no-pacing` flag on both `neqo-server` and `neqo-client`.

Sounds good?

larseggert:
Yes. Thank you!

larseggert (approved):

PR #1741, opened 2024-03-13
feat(udp): use sendmmsg and recvmmsg

mxinden:
Write and read up to `BATCH_SIZE` datagrams with single `sendmmsg` and `recvmmsg` syscall.

Previously `neqo_bin::udp::Socket::send` would use `sendmmsg`, but provide a single buffer to write into only, effectively using `sendmsg` instead of `sendmmsg`. Same with `Socket::recv`.

With this commit `Socket::send` provides `BATCH_SIZE` number of buffers on each `sendmmsg` syscall, thus writing more than one datagram at a time if available. Same with `Socket::recv`.

---

Part of https://github.com/mozilla/neqo/issues/1693.

PR #1758, opened 2024-03-18
perf(bin): add criterion benchmarks

mxinden:
Wraps the `neqo-client` and `neqo-server` code, starts the server and runs various benchmarks through the client.

Benchmarks:
- single-request-1gb
- single-request-1mb
- requests-per-second
- handshakes-per-second

---

It would be great to have **end-to-end** benchmarks (in CI)  to land performance optimizations like https://github.com/mozilla/neqo/pull/1741 with confidence.

This pull request is a shot at the above using criterion.

Benefits:

- Little boilerplate, written in Rust.
- Easy to extend with various benchmarks (e.g. multiple connections, handshakes per second, ...)
- Leverage features of criterion (though arguably the wrong tool for the job)
- Hook easily into existing [`bench.yml`](https://github.com/mozilla/neqo/blob/main/.github/workflows/bench.yml) workflow including its baseline comparison

Obviously there  are a million other ways to do this. 
- Using [hyperfine](https://github.com/sharkdp/hyperfine/) like it is done in https://github.com/mozilla/neqo/pull/1750/, thus agnostic to the binary being tested
- Building on top of [`upload_test.sh`](https://github.com/mozilla/neqo/blob/main/test/upload_test.sh) simulating an actual network (latency, packet drop, ...), i.e. not just loopback.
- Build our own platform, like
    - https://microsoft.github.io/msquic/
    - https://observablehq.com/@libp2p-workspace/performance-dashboard
- ...

What are people's thoughts? Are there long-term plans for neqo? Are there prominent examples worth taking inspiration from? Does Firefox / Mozilla already have some continuous benchmarking setup worth hooking into?


larseggert:
Better performance (regression) testing is definitely something we are interested in having.

One part of this is the building-up of a suite of benchmarks that we run in CI. I envision a process where if we identify a performance issue, we first quantify it through some (micro)benchmarks that from then on become part of the suite. If, based on the benchmark results, we decide that a performance issue needs to be fixed, we do that.

A second part is macro-level benchmarks, e.g., end-to-end throughput and other tests. I haven't spend much time on how we should best do this. At the moment, I feel that some simple end-to-end tests, like you propose here or we have in `bench.yml` are already sufficient to identify performance issues that we can address using the process in the previous paragraph.

mxinden:
Making some progress. Actual numbers are still preliminary.

```
Benchmarking single-request-1gb/client: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 76.6s.
single-request-1gb/client
                        time:   [8.3150 s 9.7354 s 11.247 s]
                        thrpt:  [91.047 MiB/s 105.18 MiB/s 123.15 MiB/s]
                 change:
                        time:   [-29.289% -14.840% +0.7625%] (p = 0.10 > 0.05)
                        thrpt:  [-0.7568% +17.426% +41.421%]

Benchmarking single-request-1mb/client: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 8.8s or enable flat sampling.
single-request-1mb/client
                        time:   [159.50 ms 162.72 ms 167.38 ms]
                        thrpt:  [5.9743 MiB/s 6.1454 MiB/s 6.2697 MiB/s]
                 change:
                        time:   [+30.929% +53.168% +81.193%] (p = 0.00 < 0.05)
                        thrpt:  [-44.810% -34.712% -23.623%]
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low severe

requests-per-second(10_000)/client
                        time:   [217.67 ms 242.75 ms 269.26 ms]
                        thrpt:  [37.140 Kelem/s 41.194 Kelem/s 45.942 Kelem/s]
                 change:
                        time:   [-13.591% -0.6218% +13.808%] (p = 0.92 > 0.05)
                        thrpt:  [-12.133% +0.6256% +15.729%]

Benchmarking handshakes-per-second(100)/client: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 25.0s.
handshakes-per-second(100)/client
                        time:   [1.8661 s 1.9559 s 2.0763 s]
                        thrpt:  [48.162  elem/s 51.127  elem/s 53.587  elem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

neqo-bin/benches/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Download resources in series using separate connections.
+    download_in_series: bool,
+    sample_size: Option<usize>,
+}
+
+fn transfer(c: &mut Criterion) {
+    log::init(Some(log::LevelFilter::Off));
+    neqo_crypto::init_db(PathBuf::from_str("../test-fixture/db").unwrap());
+
+    let done_sender = spawn_server();
```
mxinden, line 26: As of today, `cargo bench` is pinned to a single core.

https://github.com/mozilla/neqo/blob/bb3ab602481c0ec61597ef76242a3008108d524b/.github/workflows/bench.yml#L55-L59

Thus `neqo-client` and `neqo-server` compete for the same CPU resource in the benchmarks below.

This could explain the significantly lower throughput on the benchmark server compared to my laptop:

Benchmark server:
```
1-conn/1-100mb-resp/client
                        time:   [12.316 s 12.378 s 12.437 s]
                        thrpt:  [8.0404 MiB/s 8.0786 MiB/s 8.1195 MiB/s]
```
https://github.com/mozilla/neqo/actions/runs/8402727182/job/23012699901?pr=1758

My laptop:
```
1-conn/1-100mb-resp/client
                        time:   [407.05 ms 421.78 ms 439.20 ms]
                        thrpt:  [227.69 MiB/s 237.09 MiB/s 245.67 MiB/s]
```

mxinden: @larseggert do you have thoughts on multi-threaded benchmarks?

mxinden (commented):

larseggert:
Is there a way to group the benchmarks, so we can pin those where it makes sense to a core, and let others such as this one be  scheduled on multiple cores?

.github/workflows/bench.yml
```diff
@@ -70,11 +70,18 @@ jobs:
...
-            cargo "+$TOOLCHAIN" bench --features bench -- --noplot | tee results.txt
+          # Pin all but neqo-bin benchmarks to CPU 0. neqo-bin benchmarks run both a client and a server, thus benefiting from multiple CPU core.
+          benchmarks=(
+            "taskset -c 0,--exclude neqo-bin"
+            ",--package neqo-bin"
+          )
+          for config in "${benchmarks[@]}"; do
+            IFS=',' read -r cpu_pinning packages <<< "$config"
+            # Run all benchmarks at elevated priority.
+            nice -n -20 "$cpu_pinning" cargo "+$TOOLCHAIN" bench "$packages" --features bench -- --noplot | tee -a results.txt
```
larseggert, line 83: @mxinden you probably need to suppress the `shellcheck` error

mxinden:
@larseggert mind taking another look?

.github/workflows/bench.yml
```diff
@@ -70,11 +70,15 @@ jobs:
...
         run: |
-          taskset -c 0 nice -n -20 \
-            cargo "+$TOOLCHAIN" bench --features bench -- --noplot | tee results.txt
+          # Pin all but neqo-bin benchmarks to CPU 0. neqo-bin benchmarks run
+          # both a client and a server, thus benefiting from multiple CPU cores.
+          #
+          # Run all benchmarks at elevated priority.
+          taskset -c 0 nice -n -20 cargo "+$TOOLCHAIN" bench --workspace --exclude neqo-bin --features bench -- --noplot | tee results.txt
+          taskset -c 0 nice -n -20 cargo "+$TOOLCHAIN" bench --package neqo-bin --features bench -- --noplot | tee -a results.txt
+
```
larseggert: You **are** pinning both of these `cargo bench` runs to core 0 - this is not what the comment says?

larseggert (changes requested):
Could you break out the renaming of `ClientError` and `ServerError` to `Error` into its own PR? Are there any other chores that should not be part of this one?

neqo-bin/src/udp.rs
```diff
@@ -23,6 +23,8 @@ use tokio::io::Interest;
 const RECV_BUF_SIZE: usize = u16::MAX as usize;
 
 pub struct Socket {
+    #[allow(unknown_lints)] // available with Rust v1.75
+    #[allow(clippy::struct_field_names)]
```
mxinden, line 27: This one is unrelated. Should I open a separate pull request for it?

larseggert: Let's leave it in.

larseggert (approved):

PR #1762, opened 2024-03-20
feat: Extend `Frame::Padding` with length

larseggert:
This is preparation of qlog supporting the logging of runs of padding frames via `payload_length`, instead of each one individually.

Needs https://github.com/cloudflare/quiche/pull/1745 (eventually).

### DIFF
--- a/neqo-transport/src/frame.rs
+++ b/neqo-transport/src/frame.rs
@@ -20,7 +20,7 @@ use crate::{
 #[allow(clippy::module_name_repetitions)]
 pub type FrameType = u64;
 
-const FRAME_TYPE_PADDING: FrameType = 0x0;
+pub const FRAME_TYPE_PADDING: FrameType = 0x0;
 pub const FRAME_TYPE_PING: FrameType = 0x1;
 pub const FRAME_TYPE_ACK: FrameType = 0x2;
 const FRAME_TYPE_ACK_ECN: FrameType = 0x3;
@@ -103,7 +103,9 @@ pub struct AckRange {
 
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum Frame<'a> {
-    Padding,
+    Padding {
+        length: u64,
+    },
     Ping,
     Ack {
         largest_acknowledged: u64,
@@ -215,7 +217,7 @@ impl<'a> Frame<'a> {
 
     pub fn get_type(&self) -> FrameType {
         match self {
-            Self::Padding => FRAME_TYPE_PADDING,
+            Self::Padding { .. } => FRAME_TYPE_PADDING,
             Self::Ping => FRAME_TYPE_PING,
             Self::Ack { .. } => FRAME_TYPE_ACK, // We don't do ACK ECN.
             Self::ResetStream { .. } => FRAME_TYPE_RESET_STREAM,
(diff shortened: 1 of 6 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -257,7 +258,8 @@ impl PacketBuilder {
     /// Returns true if padding was added.
     pub fn pad(&mut self) -> bool {
         if self.padding && !self.is_long() {
-            self.encoder.pad_to(self.limit, 0);
+            self.encoder
+                .pad_to(self.limit, FRAME_TYPE_PADDING.try_into().unwrap());
```
larseggert, line 262: I can back this out, but I found it weird that we hard-coded the frame type here.

neqo-transport/src/frame.rs
```diff
@@ -103,7 +103,9 @@ pub struct AckRange {
 
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum Frame<'a> {
-    Padding,
+    Padding {
+        length: u64,
+    },
```
martinthomson, line 108: Padding(u64) would be OK.  But I question the need for a u64 here.  Wouldn't u16 work?

neqo-transport/src/frame.rs
```diff
@@ -409,13 +409,24 @@ impl<'a> Frame<'a> {
...
-        let t = d(dec.decode_varint())?;
+        let t = dv(dec)?;
         match t {
-            FRAME_TYPE_PADDING => Ok(Self::Padding),
+            FRAME_TYPE_PADDING => {
+                let mut length = 1;
+                // TODO: It may be worthwhile to find a more efficient way to skip
+                // over runs of zero bytes.
+                while dec.remaining() > 0
+                    && u64::from(dec.peek_byte().unwrap()) == FRAME_TYPE_PADDING
```
martinthomson, line 419: Can we maybe have a peek() that returns `Option<u8>`?

neqo-transport/src/packet/mod.rs
martinthomson: Can you use `Frame::Padding(1).get_type()`?  That should optimize well enough.

neqo-transport/src/qlog.rs
```diff
@@ -395,7 +395,7 @@ pub fn metrics_updated(qlog: &mut NeqoQlog, updated_metrics: &[QlogMetric]) {
 #[allow(clippy::cast_possible_truncation, clippy::cast_precision_loss)] // No choice here.
 fn frame_to_qlogframe(frame: &Frame) -> QuicFrame {
```
martinthomson, line 396: Hmm, why are we not implementing `From<&Frame>` for `QuicFrame`?

martinthomson (approved):
Looks good.

neqo-transport/src/frame.rs
martinthomson: Ok, we have one already.

Do this instead:

```rust
while let Some(b) == dec.peek_byte() {
    if b != FRAME_TYPE_PADDING { break; }
    len += 1;
    dec.skip(1);
}
```

neqo-transport/src/packet/mod.rs
larseggert: Also returns `u64`, so would still need to convert.

martinthomson: Ah, of course.  Which is of course why I just put a raw 0 there before.  What you have is totally fine.

KershawChang (approved):
lgtm

PR #1764, opened 2024-03-21
test: Experiment with `cargo fuzz`

larseggert:
Adds four simple fuzzing targets, for packets and frames, and for fuzzed data inside client and server initials.


larseggert:
@KershawChang @martinthomson any chance for a review?

fuzz/fuzz_targets/packet.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    use neqo_transport::{packet::PublicPacket, RandomConnectionIdGenerator};
+
+    static INIT: Once = Once::new();
+    static mut DECODER: Option<neqo_transport::RandomConnectionIdGenerator> = None;
+
+    // Initialize things
+    unsafe {
+        INIT.call_once(|| {
+            neqo_crypto::init();
```
mxinden, line 18: `neqo_crytpo::init` already itself uses a `std::sync::Once`. Is the additional `Once` needed here?

https://github.com/mozilla/neqo/blob/6344d25eaebb9a873e72adb6252901dd9579a763/neqo-crypto/src/lib.rs#L114-L129

fuzz/fuzz_targets/packet.rs
```diff
@@ -0,0 +1,29 @@
...
+            DECODER = Some(RandomConnectionIdGenerator::new(20));
```
mxinden, line 19: Instead of the use of `unsafe`, how about using `lazy_static!`. (Note that we already have the dependency in our tree, thus no additional compile time or attack vector.)

``` diff
modified   fuzz/Cargo.toml
@@ -16,6 +16,7 @@ libfuzzer-sys = { version = "0.4" }
 neqo-transport = { path = "../neqo-transport", features = ["fuzzing"] }
+lazy_static = "1.4.0"
 
 [lints]
 workspace = true
modified   fuzz/fuzz_targets/packet.rs
@@ -1,22 +1,16 @@
 #![no_main]
-use std::sync::Once;
+use lazy_static::lazy_static;
 
 use libfuzzer_sys::fuzz_target;
 use neqo_transport::{packet::PublicPacket, RandomConnectionIdGenerator};
 
-static INIT: Once = Once::new();
-static mut DECODER: Option<neqo_transport::RandomConnectionIdGenerator> = None;
+lazy_static! {
+    static ref DECODER: RandomConnectionIdGenerator = RandomConnectionIdGenerator::new(20);
+}
 
 fuzz_target!(|data: &[u8]| {
-    // Initialize things
-    unsafe {
-        INIT.call_once(|| {
-            neqo_crypto::init();
-            DECODER = Some(RandomConnectionIdGenerator::new(20));
-        });
-    }
-    let decoder = unsafe { DECODER.as_ref().unwrap() };
+    neqo_crypto::init();
 
     // Run the fuzzer
-    let _ = PublicPacket::decode(data, decoder);
+    let _ = PublicPacket::decode(data, &*DECODER);
 });
```

mxinden (commented):
I don't think the use of `unsafe` is necessary here.

In case you like the suggestions, you can simply merge https://github.com/mozilla/neqo/pull/1789.

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+  "fuzz",
```
martinthomson, line 3: I think that the convention is to have the directory name match the package name.

```suggestion
  "neqo-fuzz",
```

fuzz/fuzz_targets/frame.rs
```diff
@@ -0,0 +1,17 @@
+#![cfg_attr(not(windows), no_main)]
+
+#[cfg(not(windows))]
```
martinthomson, line 3: Can we simply avoid building this package on windows?  Because having this annotation everywhere is going to get tiresome.

neqo-transport/src/packet/mod.rs
```diff
@@ -737,6 +737,7 @@ impl<'a> PublicPacket<'a> {
     }
 
     #[must_use]
+    #[allow(clippy::len_without_is_empty)]
```
martinthomson, line 740:
```suggestion
    #[allow(clippy::len_without_is_empty)] // is_empty() would always return false in this case
```

fuzz/fuzz_targets/packet.rs
```diff
@@ -0,0 +1,21 @@
+#![cfg_attr(not(windows), no_main)]
+
+#[cfg(not(windows))]
+use libfuzzer_sys::fuzz_target;
+
+#[cfg(not(windows))]
+fuzz_target!(|data: &[u8]| {
+    use neqo_transport::{packet::PublicPacket, RandomConnectionIdGenerator};
+
+    lazy_static::lazy_static! {
```
martinthomson, line 10: Rather than once lazy_static, use OnceCell: https://docs.rs/once_cell/latest/once_cell/#faq (don't use the crate, it's in the std library now)

martinthomson (commented):
This whole "not-windows" thing is a bit of a problem if we're going to do more of this.

Is it possible to structure this differently?  Let's say that you could build a single binary with multiple fuzzing targets in it, that includes all the `#[cfg]` garbage and conditionally includes a module that has all the real fuzzing code in that.  I don't know how the `fuzz_target!()` macro is supposed to work, but if that's possible it would be a lot better that way.

The best option would be to have the entire package disabled on Windows, but I haven't worked out how to do that.

On that note, the libfuzzer-sys crate (bad name that) says that it is only supported on Linux, but it looks like this is working fine on Mac.  I assume that's a documentation fail on their end.

fuzz/fuzz_targets/frame.rs
larseggert: I had looked into this, but wasn't able to figure out how to do this.

Cargo.toml
larseggert: I undid this and instead changed the directory name to `fuzz`. Otherwise one has to pass the `--fuzz-dir neqo-fuzz` argument on each invocation of the fuzzer (because it assumes the default is `fuzz`), which is annoying.

larseggert:
> Is it possible to structure this differently? Let's say that you could build a single binary with multiple fuzzing targets in it, 

I don't think this is how the fuzzer works. Each fuzz target is its own binary.

> The best option would be to have the entire package disabled on Windows, but I haven't worked out how to do that.

Me either. I looked but it's apparently not possible with cargo.

> On that note, the libfuzzer-sys crate (bad name that) says that it is only supported on Linux, but it looks like this is working fine on Mac. I assume that's a documentation fail on their end.

[The README](https://github.com/rust-fuzz/cargo-fuzz?tab=readme-ov-file#installation) says:
> Note: libFuzzer needs LLVM sanitizer support, so this only works on x86-64 Linux, x86-64 macOS and Apple-Silicon (aarch64) macOS for now. You'll also need a C++ compiler with C++11 support.

fuzz/fuzz_targets/packet.rs
larseggert: I used `OnceLock`.

test-fixture/src/sim/rng.rs
```diff
@@ -29,6 +30,12 @@ impl Random {
         }
     }
 
+    #[cfg(feature = "disable-random")]
+    pub fn random(&mut self) -> u64 {
+        0
+    }
```
martinthomson, line 36: I think that it would be better if you used a fixed seed (or one controlled by the fuzzer) instead.

fuzz/src/lib.rs
```diff
@@ -0,0 +1,134 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
martinthomson, line 1: The functions in this file seem to be copied from the neqo-transport crate.  Maybe you can just move them to the test fixture instead.

neqo-common/src/fuzz.rs
```diff
@@ -0,0 +1,45 @@
...
+pub fn write_item_to_fuzzing_corpus(target: &str, data: &[u8]) {
+    // This bakes in the assumption that we're executing in the root of the neqo workspace.
+    // Unfortuntely, `cargo fuzz` doesn't provide a way to learn the location of the corpus
+    // directory. As a safety measure, panic if the directory does not exist.
+    let corpus = Path::new("../fuzz/corpus").join(target);
+    assert!(
+        corpus.is_dir(),
+        "corpus directory {} does not exist",
+        corpus.display()
+    );
```
martinthomson, line 30: Can you make it instead?

neqo-crypto/src/p11.rs
```diff
@@ -290,15 +290,21 @@ impl Item {
...
     let m_buf = buf.as_mut();
-    let len = c_int::try_from(m_buf.len()).unwrap();
-    secstatus_to_res(unsafe { PK11_GenerateRandom(m_buf.as_mut_ptr(), len) }).unwrap();
+    #[cfg(not(feature = "disable-random"))]
+    {
+        let len = std::os::raw::c_int::try_from(m_buf.len()).unwrap();
+        secstatus_to_res(unsafe { PK11_GenerateRandom(m_buf.as_mut_ptr(), len) }).unwrap();
+    }
+    #[cfg(feature = "disable-random")]
+    m_buf.fill(0);
```
martinthomson, line 307:
```suggestion
    if cfg!(feature = "disable-random") {
        m_buf.fill(0);
    } else }
        let len = std::os::raw::c_int::try_from(m_buf.len()).unwrap();
        secstatus_to_res(unsafe { PK11_GenerateRandom(m_buf.as_mut_ptr(), len) }).unwrap();
    }
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1099,10 +1099,18 @@ impl Connection {
     #[must_use = "Output of the process function must be handled"]
     pub fn process(&mut self, dgram: Option<&Datagram>, now: Instant) -> Output {
         if let Some(d) = dgram {
+            #[cfg(feature = "build-fuzzing-corpus")]
+            neqo_common::write_item_to_fuzzing_corpus("packet", d);
             self.input(d, now, now);
             self.process_saved(now);
         }
-        self.process_output(now)
+        #[warn(clippy::let_and_return)]
```
martinthomson, line 1107:
```suggestion
        #allow(clippy::let_and_return)]
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1099,10 +1099,18 @@ impl Connection {
...
+        let output = self.process_output(now);
+        #[cfg(feature = "build-fuzzing-corpus")]
+        if let Output::Datagram(ref d) = output {
```
martinthomson, line 1110:
```suggestion
        if let Some(d) = output.dgram().as_ref() {
```

neqo-common/src/fuzz.rs
```diff
@@ -0,0 +1,45 @@
...
+
+/// Write a data item `data` for the fuzzing target `target` to the fuzzing corpus. The caller needs
+/// to make sure that `target` is the correct fuzzing target name for the data written.
+///
+/// # Panics
+///
+/// Panics if the corpus directory does not exist or if the corpus item cannot be written.
+pub fn write_item_to_fuzzing_corpus(target: &str, data: &[u8]) {
+    // This bakes in the assumption that we're executing in the root of the neqo workspace.
+    // Unfortuntely, `cargo fuzz` doesn't provide a way to learn the location of the corpus
```
martinthomson, line 23:
```suggestion
    // Unfortunately, `cargo fuzz` doesn't provide a way to learn the location of the corpus
```

test-fixture/src/sim/rng.rs
larseggert: I'm not sure if the fuzzer would be as successful in mutating the practically random (but fixed) output that would generate vs. zimple zeros?

neqo-common/src/fuzz.rs
larseggert: I can, but that would take away this being a safety measure.

fuzz/src/lib.rs
larseggert: See #1830. When that's merged, I'll rebase this PR.

test-fixture/src/sim/rng.rs
larseggert: I changed that in 376686814eda359fb17f8067a0fd744d98bd9708, please take a look?

(It does have the nice side effect that way more tests now pass, generating a much larger corpus for the fuzzer to then mutate.)

neqo-transport/src/connection/mod.rs
```diff
@@ -1101,10 +1101,18 @@ impl Connection {
     #[must_use = "Output of the process function must be handled"]
     pub fn process(&mut self, dgram: Option<&Datagram>, now: Instant) -> Output {
         if let Some(d) = dgram {
+            #[cfg(feature = "build-fuzzing-corpus")]
+            neqo_common::write_item_to_fuzzing_corpus("packet", d);
```
martinthomson, line 1105: I think that you should avoid the use of *input* datagrams unless you can show that they are valid.  I think that it is enough to check the output datagrams.

neqo-transport/src/connection/mod.rs
```diff
@@ -1101,10 +1101,18 @@ impl Connection {
...
             self.input(d, now, now);
             self.process_saved(now);
         }
-        self.process_output(now)
+        #[allow(clippy::let_and_return)]
+        let output = self.process_output(now);
+        #[cfg(feature = "build-fuzzing-corpus")]
+        if let Some(d) = output.clone().dgram() {
+            neqo_common::write_item_to_fuzzing_corpus("packet", &d);
```
martinthomson, line 1113: I would also check here whether the test frame writer was enabled and not use packets where it was on.

I know that we might want to have a corpus that also includes invalid stuff, but it probably won't help to give the fuzzer invalid garbage.

neqo-common/src/lib.rs
```diff
@@ -9,6 +9,8 @@
 mod codec;
 mod datagram;
 pub mod event;
+#[cfg(feature = "build-fuzzing-corpus")]
+pub mod fuzz;
```
martinthomson, line 13: Because you are re-exporting this...
```suggestion
mod fuzz;
```

martinthomson (commented):
https://github.com/mozilla/neqo/commit/376686814eda359fb17f8067a0fd744d98bd9708 has a couple of comments.

I'm on vacation from tomorrow, so I trust you to do the right thing with this.  This is generally OK.

(I did question whether it is the right decision to inline the write_to_corpus calls, rather than having specific code to generate that corpus, but this works.)  Do you have a script for that?

neqo-transport/src/connection/mod.rs
larseggert: Should I do the same with frames? At the moment, I only save _inbound_ frames, not generated ones.

larseggert:
> (I did question whether it is the right decision to inline the write_to_corpus calls, rather than having specific code to generate that corpus, but this works.) Do you have a script for that?

A corpus can get generated in a number of ways, it's basically just a collection of files. The fuzzer has commands to manage it (e.g., minify it, etc.), in Rust via `cargo fuzz`. For quant, I found it convenient to (re-)generate a corpus as a side effect of running tests, I just followed that approach here, too.

neqo-transport/src/connection/mod.rs
larseggert: Merging this now. We can always change things.

PR #1767, opened 2024-03-22
chore: Rename feature `fuzzing` to `disable-encryption`

larseggert:
Because `cargo fuzz` relies on being able to use `fuzzing`.

Also remove the `fuzzing` argument from the signatures of several functions and the `fuzzing` member from several structs, and instead disable encryption solely based on the `disable-encryption` feature.


neqo-crypto/src/aead_fuzzing.rs
```diff
@@ -20,20 +20,18 @@ pub struct FuzzingAead {
 }
 
 impl FuzzingAead {
+    #[cfg(feature = "disable-encryption")]
```
martinthomson, line 23: Why not disable this file completely if encryption is disabled.  That is, put this at the top:

```rust
#![cfg(feature = "disable-encryption")]
```

Then you don't have to use an `Option` on the implementation.  There is a lot that you can then remove related to the real AEAD at that point.

martinthomson (changes requested):
It looks like you decided to remove the configurability.  That's fine, but you should complete the job in `aead_fuzzing.rs`.

.github/workflows/check.yml
```diff
@@ -114,6 +114,8 @@ jobs:
         run: |
           # shellcheck disable=SC2086
           cargo +${{ matrix.rust-toolchain }} llvm-cov nextest $BUILD_TYPE --features ci --no-fail-fast --lcov --output-path lcov.info
+          # shellcheck disable=SC2086
+          cargo +${{ matrix.rust-toolchain }} llvm-cov nextest $BUILD_TYPE --features ci,disable-encryption --no-fail-fast --lcov --output-path lcov-de.info
```
martinthomson, line 118: So, looking again, what purpose does this serve?  To test that we have good coverage when encryption is disabled?  That seems like a non-goal to me.  Particularly since the run that disables encryption is merged with the regular run.

If the goal is to determine how much coverage our fuzzing tests have, then we should be running coverage for just the fuzzing runs.

martinthomson (commented):
So much cleaner.  Approved.

.github/workflows/check.yml
larseggert: Happy to take this out

PR #1768, opened 2024-03-22
ci: Run `clippy` for all features except `gecko`

larseggert:
Because it didn't get to have opinions about benches and various other code before that was hidden behind features. Avoids bitrot, too.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -148,7 +148,7 @@ jobs:
           # respective default features only. Can reveal warnings otherwise
           # hidden given that a plain cargo clippy combines all features of the
           # workspace. See e.g. https://github.com/mozilla/neqo/pull/1695.
-          cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets --each-feature --exclude-features gecko -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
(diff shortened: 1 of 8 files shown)


.github/workflows/check.yml
```diff
@@ -148,7 +148,7 @@ jobs:
           # respective default features only. Can reveal warnings otherwise
           # hidden given that a plain cargo clippy combines all features of the
           # workspace. See e.g. https://github.com/mozilla/neqo/pull/1695.
-          cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
+          cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets --each-feature --exclude-features gecko -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
```
mxinden, line 151:
```suggestion
          cargo +${{ matrix.rust-toolchain }} hack clippy --all-targets --feature-powerset --exclude-features gecko -- -D warnings || ${{ matrix.rust-toolchain == 'nightly' }}
```

In case it doesn't add to much delay, how about the powerset instead of each feature individually?

https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--feature-powerset

mxinden (approved):

.github/workflows/check.yml
larseggert: Let's try.

KershawChang (approved):

PR #1769, opened 2024-03-23
ci: Disable some caching

larseggert:
The docker and python caches put quite a bit of stress on the GHA cache, causing it - for example - to time out the benchmark results too soon. See https://github.com/mozilla/neqo/actions/caches

(Need to check how much slower this makes the build though, hence draft PR.)


larseggert:
Closed in favor of #1770

PR #1772, opened 2024-03-25
perf(bin/server): increase msg size and don't allocate msg per resp

mxinden:
Previously `neqo-server` would respond to a request by repeatedly sending a static 440 byte message (Major-General's Song). Instead of sending 440 bytes, increase the batch size to 4096 bytes. This also matches the `neqo-client` receive buffer size.

https://github.com/mozilla/neqo/blob/76630a5ebb6c6b94de6a40cf3f439b9a846f6ab7/neqo-bin/src/bin/client/http3.rs#L165

Previously `ResponseData::repeat` would convert the provided `buf: &[u8]` to ` Vec<u8>`, i.e. re-allocate the buf. Instead keep a reference to the original buf, thus removing the allocation.

---

Preliminary benchmark comparison:

`main`:

```
                        time:   [768.10 ms 1.0894 s 1.5517 s]
                        thrpt:  [64.443 MiB/s 91.794 MiB/s 130.19 MiB/s]
```

this pull request:

```
                        time:   [674.30 ms 837.63 ms 1.0564 s]
                        thrpt:  [94.661 MiB/s 119.38 MiB/s 148.30 MiB/s]
```

Also visible in qviz where client reads MANY small (440 bytes) chunks from server:

image

I suggest merging https://github.com/mozilla/neqo/pull/1758 first. We can then validate this pull request using the CI benchmark server.

I will take a look whether a long lived receive buffer on the client side has any impact next.

https://github.com/mozilla/neqo/blob/76630a5ebb6c6b94de6a40cf3f439b9a846f6ab7/neqo-bin/src/bin/client/http3.rs#L165

### DIFF
--- a/neqo-bin/src/bin/server/main.rs
+++ b/neqo-bin/src/bin/server/main.rs
@@ -212,7 +212,7 @@ impl From<Vec<u8>> for ResponseData {
     fn from(data: Vec<u8>) -> Self {
         let remaining = data.len();
         Self {
-            data: Cow::Owned(data.into()),
+            data: Cow::Owned(data),
             offset: 0,
             remaining,
         }
--- a/neqo-bin/src/bin/server/main.rs
+++ b/neqo-bin/src/bin/server/main.rs
@@ -5,6 +5,7 @@
 // except according to those terms.
 
 use std::{
+    borrow::Cow,
     cell::RefCell,
     cmp::min,
     collections::HashMap,
@@ -196,7 +197,7 @@ trait HttpServer: Display {
 }
 
 struct ResponseData {
-    data: Vec<u8>,
+    data: Cow<'static, [u8]>,
     offset: usize,
     remaining: usize,
 }
@@ -211,7 +212,7 @@ impl From<Vec<u8>> for ResponseData {
     fn from(data: Vec<u8>) -> Self {
         let remaining = data.len();
         Self {
-            data,
+            data: Cow::Owned(data.into()),
             offset: 0,
             remaining,
         }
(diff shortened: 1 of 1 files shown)


neqo-bin/src/bin/server/main.rs
```diff
@@ -260,14 +261,7 @@ struct SimpleServer {
...
 impl SimpleServer {
-    const MESSAGE: &'static [u8] = b"I am the very model of a modern Major-General,\n\
-        I've information vegetable, animal, and mineral,\n\
-        I know the kings of England, and I quote the fights historical\n\
-        From Marathon to Waterloo, in order categorical;\n\
-        I'm very well acquainted, too, with matters mathematical,\n\
-        I understand equations, both the simple and quadratical,\n\
-        About binomial theorem, I'm teeming with a lot o' news,\n\
-        With many cheerful facts about the square of the hypotenuse.\n";
+    const MESSAGE: &'static [u8] = &[0; 4096];
```
mxinden, line 264: To the Gilbert and Sullivan fans: should I try to find a larger excerpt from the opera, rather than a boring `[0; 4096]`?

https://en.wikipedia.org/wiki/Major-General%27s_Song

larseggert: `git blame` says this came from @martinthomson. I don't care :-)

larseggert (approved):
Fix `clippy`, but otherwise LGTM.

larseggert (approved):

neqo-bin/src/bin/server/main.rs
larseggert: If anyone cares, they can do a PR to propose some new text.

PR #1773, opened 2024-03-26
test: Let packets be modified with a closure during tests

larseggert:
Broken out of #1678 to reduce the size of that PR.

### DIFF
--- a/neqo-transport/src/connection/tests/mod.rs
+++ b/neqo-transport/src/connection/tests/mod.rs
@@ -170,13 +170,17 @@ impl crate::connection::test_internal::FrameWriter for PingWriter {
     }
 }
 
+trait DatagramModifier: FnMut(Datagram) -> Option<Datagram> {}
+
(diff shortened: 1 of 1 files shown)


larseggert:
is there a way to _drop_ a packet inside the closure, and not just modify it?

mxinden:
> is there a way to _drop_ a packet inside the closure, and not just modify it?

Wouldn't the below work?

``` diff
- mut modifier: impl FnMut(&mut Datagram),
+ mut modifier: impl FnMut(Datagram) -> Option<Datagram>,

larseggert:
Ah yes! I think that is what @martinthomson also suggested, but my brain is fried after the IETF.

martinthomson (approved):

martinthomson:
Merge override due to known failed test.  Separately, the qns build was taking a worrying amount of time.  I cancelled at 22 minutes.  We might want to look into that.

PR #1774, opened 2024-03-26
feat: TOS improvements

larseggert:
Broken out of #1678 to reduce the size of that PR.

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -53,6 +53,16 @@ impl Datagram {
     pub fn ttl(&self) -> Option<u8> {
         self.ttl
     }
+
+    #[cfg(feature = "udp")]
+    #[must_use]
+    pub(crate) fn into_data(self) -> Vec<u8> {
+        self.d
+    }
+
+    pub fn set_tos(&mut self, tos: IpTos) {
+        self.tos = tos;
+    }
 }
 
 impl Deref for Datagram {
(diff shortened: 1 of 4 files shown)


neqo-common/src/datagram.rs
```diff
@@ -53,6 +53,16 @@ impl Datagram {
     pub fn ttl(&self) -> Option<u8> {
         self.ttl
     }
+
+    #[cfg(feature = "udp")]
+    #[must_use]
+    pub(crate) fn into_data(self) -> Vec<u8> {
+        self.d
+    }
```
mxinden, line 61: Note that the `udp` feature no longer exists and that the `into_data` function was moved to a `From<Datagram> for Vec<u8>`. See https://github.com/mozilla/neqo/pull/1736.

mxinden (commented):
One comment. Otherwise looks good to me.

neqo-transport/src/connection/mod.rs
```diff
@@ -2255,6 +2256,7 @@ impl Connection {
                 pt,
                 pn,
                 &builder.as_ref()[payload_start..],
+                IpTos::default(), // TODO: set from path
```
martinthomson, line 2259: Want to open a bug for that?

martinthomson (approved):

neqo-transport/src/connection/mod.rs
larseggert: It's already part of #1678, not sure this would be useful?

PR #1775, opened 2024-03-26
fix: Don't panic in `neqo_crypto::init()`

### LINKED ISSUE #1675: Crash in [@ core::result::unwrap_failed | neqo_crypto::init::closure$0]
See https://bugzilla.mozilla.org/show_bug.cgi?id=1854653

Not sure why we can't init NSS, but maybe we should not crash.

larseggert:
Return `Res<()>` instead.

Fixes #1675


neqo-bin/src/bin/client/main.rs
```diff
@@ -451,7 +452,7 @@ async fn main() -> Res<()> {
     neqo_common::log::init(Some(args.verbose.log_level_filter()));
     args.update_for_tests();
 
-    init();
+    init().map_err(|_| ClientError::InternalError)?;
```
martinthomson, line 455:
```suggestion
    init()?;
```

Then, you can implement `From<neqo_crypto::Error> for ClientError`.

The same applies for the server code, though you can't get `io::Error`.

neqo-crypto/src/lib.rs
```diff
@@ -107,24 +107,27 @@ fn version_check() {
...
+        secstatus_to_res(unsafe { nss::NSS_NoDB_Init(null()) })?;
+        secstatus_to_res(unsafe { nss::NSS_SetDomesticPolicy() })?;
 
-        NssLoaded::NoDb
+        Ok(NssLoaded::NoDb)
     });
+    match res {
+        Ok(_) => Ok(()),
+        Err(e) => Err(e.clone()),
+    }
```
martinthomson, line 130: I think that you can do this instead:

```suggestion
    res.clone()
```

neqo-crypto/tests/init.rs
```diff
@@ -39,7 +39,7 @@ fn init_nodb() {
 #[cfg(not(nss_nodb))]
 #[test]
 fn init_withdb() {
-    init_db(::test_fixture::NSS_DB_PATH);
+    let _ = init_db(::test_fixture::NSS_DB_PATH);
```
martinthomson, line 42:
```suggestion
    init_db(::test_fixture::NSS_DB_PATH).unwrap();
```
The test should panic here, not later.

martinthomson (changes requested):

neqo-crypto/src/lib.rs
larseggert: That actually doesn't work:
```
   --> neqo-crypto/src/lib.rs:127:5
    |
113 | pub fn init() -> Res<()> {
    |                  ------- expected `std::result::Result<(), err::Error>` because of return type
...
127 |     res.clone()
    |     ^^^^^^^^^^^ expected `Result<(), Error>`, found `&Result<NssLoaded, Error>`
    |
    = note:   expected enum `std::result::Result<(), _>`
            found reference `&std::result::Result<NssLoaded, _>`

```

mxinden: Maybe `res.map(|_| ()).clone()` @larseggert?

larseggert: Wouldn't this map all errors to not-errors? (Am still a Rust newbie.)

mxinden: No, it will only map the `Ok` variants. The `Err` enum variants are untouched.

> Maps a Result<T, E> to Result<U, E> by applying a function to a contained [Ok](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Ok) value, leaving an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) value untouched.

https://doc.rust-lang.org/std/result/enum.Result.html#method.map

mxinden: > Wouldn't this map all errors to not-errors? (Am still a Rust newbie.)

Mapping all errors to not-errors would happen when e.g. doing:

``` rust
res.and_then(|_res| Ok(()))
```

----

In more technical terms, `Result::map` is a functor transition.

``` rust
    pub fn map<U, F: FnOnce(T) -> U>(self, op: F) -> Result<U, E> {
```

`Result::and_then` is a monad transition.

``` rust
    pub fn and_then<U, F: FnOnce(T) -> Result<U, E>>(self, op: F) -> Result<U, E> {
```

larseggert: Thanks for the explanation! That didn't quite work, but 
```rs
res.as_ref().map(|_| ()).map_err(Clone::clone)
```
did.

neqo-bin/src/bin/client/main.rs
```diff
@@ -45,6 +45,13 @@ pub enum ClientError {
     IoError(io::Error),
     QlogError,
     TransportError(neqo_transport::Error),
+    InternalError(neqo_crypto::Error),
```
martinthomson, line 48:
```suggestion
    CryptoError(neqo_crypto::Error),
```

martinthomson (approved):
Suggestions on naming is all.

PR #1777, opened 2024-03-27
fix: Address more clippy warnings

larseggert:
PR #1764 exports `frame` and `packet` if feature `fuzzing` is enabled. This apparently turns on a bunch of clippy checks that are not on by default? This PR fixes them.

One side effect is that this returns a bunch of `panic!`s into returning errors.

Made this separate from #1764 to reduce that PR's size.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -256,6 +256,10 @@ impl PacketBuilder {
     /// Maybe pad with "PADDING" frames.
     /// Only does so if padding was needed and this is a short packet.
     /// Returns true if padding was added.
+    ///
+    /// # Panics
+    ///
+    /// Cannot happen.
     pub fn pad(&mut self) -> bool {
         if self.padding && !self.is_long() {
             self.encoder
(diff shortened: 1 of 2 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -792,7 +832,10 @@ impl<'a> PublicPacket<'a> {
             // too small (which is public information).
             let (key_phase, pn, header, body) = self.decrypt_header(rx)?;
             qtrace!([rx], "decoded header: {:?}", header);
-            let rx = crypto.rx(version, cspace, key_phase).unwrap();
+            let Some(rx) = crypto.rx(version, cspace, key_phase) else {
+                return Err(Error::DecryptError);
+            };
+            // let rx = crypto.rx(version, cspace, key_phase).unwrap();
```
KershawChang, line 838: Forget to remove this?

neqo-transport/src/packet/mod.rs
```diff
@@ -290,6 +294,10 @@ impl PacketBuilder {
     /// The length is filled in after calling `build`.
     /// Does nothing if there isn't 4 bytes available other than render this builder
     /// unusable; if `remaining()` returns 0 at any point, call `abort()`.
+    ///
+    /// # Panics
+    ///
+    /// This will panic if the packet number length is too large.
```
KershawChang, line 300: I don't see where the panic could happen in this function. Am I missing something?

KershawChang (approved):

PR #1779, opened 2024-03-28
fix: Exit with `ProtocolViolation` if a packet has no frames

### LINKED ISSUE #1476: Neqo server does not close the connection when receiving a packet without any frames.
(In reply to Kian Kai from comment [bug 1860211 #0](https://bugzilla.mozilla.org/show_bug.cgi?id=1860211#c0))
> Steps to reproduce:
> 
> Packet sequence to replicate the violation:
> 1) Client sends an Initial packet carrying a CRYPTO frame with Client Hello.
> 2) Client sends a Handshake packet without any frames.
> 
> 
> Actual results:
> 
> Violation detail:
> Neqo server does not close the connection immediately after receiving a packet containing no frames.
> 
> 
> Expected results:
> 
> According to Section 12.4, RFC 9000, an endpoint MUST treat receipt of a packet containing no frames as a connection error of type PROTOCOL_VIOLATION. Violations of the protocol lead to an immediate close.

larseggert:
Fixes #1476

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1552,7 +1552,9 @@ impl Connection {
         packet: &DecryptedPacket,
         now: Instant,
     ) -> Res<bool> {
-        (!packet.is_empty()).then_some(()).ok_or(Error::ProtocolViolation)?;
+        (!packet.is_empty())
+            .then_some(())
+            .ok_or(Error::ProtocolViolation)?;
 
         // TODO(ekr@rtfm.com): Have the server blow away the initial
         // crypto state if this fails? Otherwise, we will get a panic
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1560,14 +1560,20 @@ impl Connection {
         let mut ack_eliciting = false;
         let mut probing = true;
         let mut d = Decoder::from(&packet[..]);
+        let mut decoded_frames = false;
         while d.remaining() > 0 {
             let f = Frame::decode(&mut d)?;
+            decoded_frames = true;
             ack_eliciting |= f.ack_eliciting();
             probing &= f.path_probing();
             let t = f.get_type();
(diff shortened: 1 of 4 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -764,14 +764,10 @@ impl<'a> PublicPacket<'a> {
...
 
-        qtrace!(
-            "unmask hdr={}",
-            hex(&self.data[..self.header_len + SAMPLE_OFFSET])
-        );
-
         let sample_offset = self.header_len + SAMPLE_OFFSET;
         let mask = if let Some(sample) = self.data.get(sample_offset..(sample_offset + SAMPLE_SIZE))
         {
+            qtrace!("unmask hdr={}", hex(&self.data[..sample_offset]));
```
larseggert, line 770: This crashed during debugging, when `self.header_len + SAMPLE_OFFSET` was too large.

neqo-transport/src/connection/mod.rs
```diff
@@ -1560,15 +1560,21 @@ impl Connection {
...
             probing &= f.path_probing();
             let t = f.get_type();
             if let Err(e) = self.input_frame(path, packet.version(), packet.packet_type(), f, now) {
                 self.capture_error(Some(Rc::clone(path)), now, t, Err(e))?;
             }
         }
+        if !decoded_frames {
+            qerror!([self], "Received packet with no frames");
+            return Err(Error::ProtocolViolation);
+        }
```
martinthomson, line 1577: Why not check `packet.len()`?

This is going to pass if the packet contains only padding, which I think is OK, but you can make the code simpler.

```rust
packet.is_empty().then_some(()).ok_or(Error::ProtocolViolation)?;
```

I don't think that we need to use `qerror!()` here.  I'd stop at info level for errors that come from the other end.  Errors are best saved for stuff where we are badly stuck.  The code I suggest has no logging, which I think is also fine in this case.

martinthomson (approved):

PR #1780, opened 2024-03-28
perf(timer): check subset of self.items

mxinden:
`Timer::take_next` iterates each `VecDeque` in the `Vec` `self.items`. It then checks the first item in those `VecDeque`s.

**Under the assumption that all items in `self.items[t]` are smaller than all items in `self.items[t+1]` (ignoring wrap around), only a subset of `self.items` needs to be iterated, namely only from `self.cursor` to `self.delta(until)`.**

**Is this assumption correct?**

This commit changes `take_nextl` to only check this subset.

---

Why is `Timer::take_next`'s performance relevant?

Whenever `Server::process_next_output` has no more other work to do, it checks for expired timers.

https://github.com/mozilla/neqo/blob/3151adc53e71273eed1319114380119c70e169a2/neqo-transport/src/server.rs#L650

A `Server` has at most one item per connection in `Timer`. Thus, a `Server` with a single connection has a single item in `Timer` total.

The `Timer` timer wheel has 16_384 slots.

https://github.com/mozilla/neqo/blob/3151adc53e71273eed1319114380119c70e169a2/neqo-transport/src/server.rs#L55

Thus whenever `Server::process_next_output` has no more other work to do, it iterates a `Vec` of length `16_384`, only to find at most one timer, which might or might not be expired.

This shows up in CPU profiles with up to 33%. See e.g. https://github.com/mozilla/neqo/actions/runs/8452074231/artifacts/1363138571. On my local machine, a call to `Timer::take_next` takes between 5-10 micro seconds.

Note that the profiles do not always show `take_next` as it is oftentimes inlined. Add `#[inline(never)]` to make sure it isn't.

``` diff
modified   neqo-common/src/timer.rs
@@ -193,6 +193,7 @@ impl<T> Timer<T> {

     /// Take the next item, unless there are no items with
     /// a timeout in the past relative to `until`.
+    #[inline(never)]
     pub fn take_next(&mut self, until: Instant) -> Option<T> {
```

With this patch `Timer::take_next` takes significantly less CPU time on my machine (~2.8%).

Arguably a 16_384 slot timer wheel is overkill for a single timer. Maybe, to cover the use-case of a `Server` with a small amount of connections, a hierarchical timer wheel is helpful?

### DIFF
--- a/neqo-common/src/timer.rs
+++ b/neqo-common/src/timer.rs
@@ -193,22 +193,27 @@ impl<T> Timer<T> {
 
     /// Take the next item, unless there are no items with
     /// a timeout in the past relative to `until`.
+    ///
+    /// # Panics
+    ///
+    /// Impossible, I think.
     pub fn take_next(&mut self, until: Instant) -> Option<T> {
-        fn maybe_take<T>(v: &mut VecDeque<TimerItem<T>>, until: Instant) -> Option<T> {
-            if !v.is_empty() && v[0].time <= until {
-                Some(v.pop_front().unwrap().item)
-            } else {
-                None
+        let delta = self.delta(until);
+        let range = if self.cursor < delta {
+            #[allow(clippy::range_plus_one)] // non-inclusive range to match with type below
+            (self.cursor..(self.cursor + delta + 1)).chain(0..0) // additional empty range to match
+                                                                 // with type below
+        } else {
+            (self.cursor..self.items.len()).chain(0..delta)
+        };
+
+        for i in range {
+            let slot = &mut self.items[i];
+            if slot.front().is_some_and(|t| t.time <= until) {
+                return Some(slot.pop_front().unwrap().item);
             }
         }
 
-        let items_len = self.items.len();
-        for i in (self.cursor..=(self.cursor + self.delta(until))).map(|i| i % items_len) {
-            let res = maybe_take(&mut self.items[i], until);
-            if res.is_some() {
-                return res;
-            }
-        }
         None
     }
 
--- a/neqo-common/src/timer.rs
+++ b/neqo-common/src/timer.rs
@@ -202,8 +202,8 @@ impl<T> Timer<T> {
             }
         }
 
-        for i in self.cursor..(self.cursor + self.delta(until)) {
-            let i = i % self.items.len();
+        let items_len = self.items.len();
+        for i in (self.cursor..=(self.cursor + self.delta(until))).map(|i| i % items_len) {
             let res = maybe_take(&mut self.items[i], until);
             if res.is_some() {
                 return res;
@@ -264,7 +264,7 @@ impl<T> Timer<T> {
 
 #[cfg(test)]
 mod test {
-    use std::sync::OnceLock;
+    use std::{iter, sync::OnceLock};
 
     use super::{Duration, Instant, Timer};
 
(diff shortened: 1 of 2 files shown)


neqo-common/src/timer.rs
```diff
@@ -202,15 +202,9 @@ impl<T> Timer<T> {
...
-            let res = maybe_take(&mut self.items[i], until);
-            if res.is_some() {
-                return res;
-            }
-        }
-        for i in 0..idx {
-            let res = maybe_take(&mut self.items[i], until);
+        for i in self.cursor..(self.cursor + self.delta(until)) {
+            let i = i % self.items.len();
+            let res = maybe_take(&mut self.items[i], until)?;
```
martinthomson, line 207: If you do this, then the function can be inlined.

However, this will regress one aspect of performance to save another.  The reason for two loops is to avoid the modulus operation inside the loop.  I think that what you want is more complex yet.

You want the first loop to be `self.bucket(0)..min(self.items.len(), self.bucket(0) + self.delta(until))` and the second to be `0..min(self.bucket(0), self.bucket(0) + self.delta(until) - self.items.len())`  ... I think.

martinthomson: Oh, and I should have said: this is an obvious win, so it's worth doing.  I just want to make sure that we get the whole win.

larseggert:
I used https://github.com/wahern/timeout in quant, but I'm not sure there is a Rust crate. _Someone_ must have surely implemented a timing wheel we can use?

mxinden:
I updated the pull request to do no `%` at all. Still, it regresses on the _drain timer quickly_ benchmark.

> * **drain a timer quickly   time:   [696.22 ns 702.27 ns 707.77 ns]**
>       change: [+79.201% **+81.607%** +83.963%] (p = 0.00 < 0.05)
>       💔 **Performance has regressed.**

I assume this is due to the additional division in `Timer::delta` now needed to find the subset of `self.items` to iterate.

https://github.com/mozilla/neqo/blob/3151adc53e71273eed1319114380119c70e169a2/neqo-common/src/timer.rs#L83-L89

Though note that _drain a timer quickly_ only checks existing timers, i.e. never makes use of the optimization of this pull request and that it is tiny (`100`) compared to the `Timer` instance used in `server.rs` (`16384`). Thus the additional division has an artificially large impact in the benchmark.

I added an additional benchmark _drain an empty timer_ which shows a significant improvement compared to `main`:

```
$ neqo git:(main) ✗ critcmp main timer                                                                                               
group                                          main                                     timer
-----                                          ----                                     -----
drain an empty timer                           1.57    40.6±11.14µs        ? ?/sec      1.00     25.8±9.07µs        ? ?/sec
```

mxinden:
All that said, once we optimized `Timer::take_next`, we still spend a significant portion of CPU time on `Timer::next_time`.

https://github.com/mozilla/neqo/blob/3151adc53e71273eed1319114380119c70e169a2/neqo-common/src/timer.rs#L56-L58

See e.g. this flamegraph off of `main` with 4.77% spend in `next_time` and  18.86% in `take_next`:
flamegraph

mxinden:
`Timer` is only used in `neqo_transport::server`. And if I understand correctly, [we only ever insert a single time per connection.](https://github.com/mozilla/neqo/blob/3151adc53e71273eed1319114380119c70e169a2/neqo-transport/src/server.rs#L285-L287)

Do I understand correctly that the main goal of `neqo-server` is to be able to test `neqo-client` and thus `neqo` as it is used in Firefox? If so, isn't the 16384 slot timer wheel overkill and a simple `BTreeMap` suffices? Thoughts @martinthomson and @larseggert?

---

> I used https://github.com/wahern/timeout in quant, but I'm not sure there is a Rust crate. _Someone_ must have surely implemented a timing wheel we can use?

There are timer wheel crates. Though none that stand out to me [with large adoption](https://crates.io/search?q=timer%20wheel&sort=downloads). In addition, if feasible, I would much rather like to do without the additional dependency, and instead just wrap `std::collection::BTreeMap`.

mxinden:
Closing here since `neqo_common::timer` was removed with https://github.com/mozilla/neqo/pull/1784.

Thank you for the help!

PR #1784, opened 2024-04-01
perf(transport): remove Server::timers

mxinden:
The current `neqo_transport::server::Server::timers` has a large performance overhead, especially when serving small amount of connections. See https://github.com/mozilla/neqo/pull/1780 for details.

This commit optimizes for the small-number-of-connections case, keeping a single callback timestamp only, iterating each connection when there is no other work to be done.

---

Draft for now.

I need more context to know what to optimize for.

> Do I understand correctly that the main goal of `neqo-server` is to be able to test `neqo-client` and thus `neqo` as it is used in Firefox? If so, isn't the 16384 slot timer wheel overkill and a simple `std::collections::BTreeMap` (or the likes) suffices?

See https://github.com/mozilla/neqo/pull/1780#issuecomment-2030067812.

Opening up early to get some performance stats from the CI benchmark machine.


larseggert:
I'd be OK to rip this out and limit our demo server to only being useful with a small number of active connections.

neqo-transport/src/server.rs
```diff
@@ -646,37 +630,35 @@ impl Server {
...
-        while let Some(c) = self.timers.take_next(now) {
-            if let Some(d) = self.process_connection(&c, None, now) {
+
+        qtrace!([self], "No packet to send still, check callbacks");
+        loop {
+            let connection = self
+                .connections
+                .borrow()
+                .values()
+                .find(|c| c.borrow().callback.map_or(false, |t| t <= now))
```
martinthomson, line 640: Maybe wrap the ".callback.map_or()" bit in a function on the type itself.

neqo-transport/src/server.rs
```diff
@@ -174,8 +167,8 @@ pub struct Server {
     active: HashSet<ActiveConnectionRef>,
     /// The set of connections that need immediate processing.
     waiting: VecDeque<StateRef>,
-    /// Outstanding timers for connections.
-    timers: Timer<StateRef>,
+    /// The latest [`Output::Callback`] returned from [`Server::process`].
+    callback: Option<Instant>,
```
martinthomson, line 171: Maybe "wake_at" or something like that.  "callback" implies that this is a function.

neqo-transport/src/server.rs
```diff
@@ -646,37 +630,35 @@ impl Server {
...
                 qtrace!([self], "Send packet: {:?}", d);
                 Output::Datagram(d)
             })
-            .or_else(|| {
-                self.next_time(now).map(|delay| {
-                    qtrace!([self], "Wait: {:?}", delay);
-                    Output::Callback(delay)
-                })
-            })
+            .or_else(|| self.callback.take().map(|c| Output::Callback(c - now)))
```
martinthomson, line 661: Again, this would be better as a function rather than relying on direct access to the field.

neqo-transport/src/server.rs
mxinden: :+1: done in https://github.com/mozilla/neqo/pull/1784/commits/1a3b66412e890aaece900c1825952118dc6507c9.

neqo-transport/src/server.rs
mxinden: :+1: done via https://github.com/mozilla/neqo/pull/1784/commits/32db5153ef784f46d8745a2fd91b7aa94f2f7b8c.

neqo-transport/src/server.rs
mxinden: Note that `self` here is `Server` and not `ServerConnectionState`. Is the additional function on `Server` worth the indirection @martinthomson? Not a strong opinion. Happy to add e.g. the below.

``` diff
modified   neqo-transport/src/server.rs
@@ -15,7 +15,7 @@ use std::{
     ops::{Deref, DerefMut},
     path::PathBuf,
     rc::{Rc, Weak},
-    time::Instant,
 };
 
 use neqo_common::{
@@ -661,6 +661,10 @@ impl Server {
         }
     }
 
+    fn take_wake_at(&mut self, now: Instant) -> Option<Duration> {
+        self.wake_at.take().map(|c| c - now)
+    }
+
         if self.wake_at.map_or(false, |c| c <= now) {
             self.wake_at = None;
@@ -673,7 +677,7 @@ impl Server {
                 qtrace!([self], "Send packet: {:?}", d);
                 Output::Datagram(d)
             })
-            .or_else(|| self.wake_at.take().map(|c| Output::Callback(c - now)))
+            .or_else(|| self.take_wake_at(now).map(|d| Output::Callback(d)))
             .unwrap_or_else(|| {
                 qtrace!([self], "Go dormant");
                 Output::None
```

mxinden:
Thank you for the review @martinthomson. Can you take another look?

neqo-transport/src/server.rs
martinthomson: Nah, that's fine.  It's much easier to process now that the variable is more appropriately named.

martinthomson (approved):

PR #1785, opened 2024-04-02
ci(qns): run all tests and remove msquic & quic-go

mxinden:
Run all QUIC Interop Runner testcases.

For the sake of reducing runtime, this commit also removes msquic and quic-go, leaving only ngtcp2 and neqo itself. Reason being, that ngtcp2 is the only implementation that passes all testcases, especially the ECN testcase.

Long term we can consider parallelizing the test execution as suggested by Marten in https://github.com/quic-go/quic-go/pull/4339#issuecomment-1977606709.

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -71,6 +71,5 @@ jobs:
           name: 'neqo-latest'
           image: ${{ steps.docker_build_and_push.outputs.imageID }}
           url: https://github.com/mozilla/neqo
-          test: handshake
-          client: neqo-latest,quic-go,ngtcp2,neqo,msquic
-          server: neqo-latest,quic-go,ngtcp2,neqo,msquic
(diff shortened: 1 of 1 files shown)


larseggert:
Would we not also want to remove `neqo` and only run `neqo-latest`?

mxinden:
I had in mind to catch regressions between neqo versions, i.e. between `neqo-latest` and `neqo` (last `main` commit). I don't feel strongly about it.

larseggert:
Yes, tests should be sufficient.

larseggert:
> I had in mind to catch regressions between neqo versions, i.e. between `neqo-latest` and `neqo` (last `main` commit). I don't feel strongly about it.

Yeah, that makes sense. Leave it in. Esp. if we do `-t onlyTests`.

mxinden:
Replaced by https://github.com/mozilla/neqo/pull/1847/.

PR #1786, opened 2024-04-02
fix(client): exit with non-zero on error

mxinden:
When a connection closes with an error, surface the error to the user and exit with non-zero.


larseggert (approved):

neqo-transport/src/lib.rs
```diff
@@ -70,8 +70,8 @@ const ERROR_AEAD_LIMIT_REACHED: TransportError = 15;
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
 pub enum Error {
     NoError,
-    // Each time tihe error is return a different parameter is supply.
-    // This will be use to distinguish each occurance of this error.
+    // Each time this error is returned a different parameter is supplied.
+    // This will be used to distinguish each occurance of this error.
```
mxinden, line 74: Note that I only fixed the typos in this comment. I am not sure the comment itself is still applicable.

(Also let me know if you would prefer separate pull requests for these non-code drive-by changes.)

larseggert (approved):

WesleyRosenblum:
I think this may have broken some neqo-client QUIC interop tests. 

I'm looking at the resumption (`R`) tests: https://interop.seemann.io/

Any server that does not send a connection close before `neqo-client`'s idle timer expires causes the test to fail:

```
client  | 0s967ms INFO Saving https://server4/fbpsuqvmkw to "/downloads/fbpsuqvmkw"
client  | 1s  4ms WARN Unhandled event StateChange(Confirmed)
client  | 33s178ms INFO [Client 8ae08d14ec800655] idle timeout expired
client  | Error: TransportError(IdleTimeout)
client exited with code 1
Aborting on container exit...
```

I'm guessing this PR isn't the root cause, but it is causing an underlying issue (the client not existing cleanly after the file has been transferred) to surface

mxinden:
Thank you for tracing this down @wesleyrosenblum. I will take a look.

mxinden:
Still need to take a look.

> shouldn't the QNS CI step have caught that?

We are not running the resumption test on CI yet :( I am lagging behind with https://github.com/mozilla/neqo/pull/1785.

https://github.com/mozilla/neqo/blob/d74ad5ea9b6dbed83674fc1963a6424c79c12588/.github/workflows/qns.yml#L74

PR #1788, opened 2024-04-02
fix(server): log error when failing to read file

mxinden:
As part of the QUIC Interop testcases, a server needs to be able to respond to a request with the content of a local file.

Previously, when failing to read a file, the server would simply reply with the predefined `SELF::MESSAGE`.

Now the server logs the error and moves on.

### DIFF
--- a/neqo-bin/src/server/mod.rs
+++ b/neqo-bin/src/server/mod.rs
@@ -10,8 +10,7 @@ use std::{
     cmp::min,
     collections::HashMap,
     fmt::{self, Display},
-    fs::OpenOptions,
-    io::{self, Read},
+    fs, io,
     net::{SocketAddr, ToSocketAddrs},
     path::PathBuf,
     pin::Pin,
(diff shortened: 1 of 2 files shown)


neqo-bin/src/server/mod.rs
```diff
@@ -188,28 +187,11 @@ impl Args {
...
-                    Some(data)
-                }
-                Err(e) => {
-                    qerror!("Error reading data: {e:?}");
-                    None
-                }
-            }
-        })
+fn qns_read_response(filename: &str) -> Result<Vec<u8>, io::Error> {
+    let path: PathBuf = ["/www", filename.trim_matches(|p| p == '/')]
```
martinthomson, line 191: This is not safe still.  We should be rejecting attempts to read files with a '/' or '\', but also paths that are equal to "." or ".." (though these a probably not good).  It's OK if this is not entirely robust, but still.

I would prefer to see a 404 response for files that don't exist.

Finally, "/www" should probably come in through the command line and default to "./www" (which might not exist, which is fine).

mxinden: Note that this function is used for the QUIC Interop Runner only. It is only executed when `--qns-test` is provided.

https://github.com/mxinden/neqo/blob/27ae7cbeb21d1858d3e371fa3a3ec01d029a93cd/neqo-bin/src/server/mod.rs#L343-L349

> Finally, "/www" should probably come in through the command line and default to "./www"

Given that it is only used through QUIC Interop, would this additional flag ever be used with anything other than `/www`?

> This is not safe still. We should be rejecting attempts to read files with a '/' or '', but also paths that are equal to "." or ".." (though these a probably not good). It's OK if this is not entirely robust, but still.

`..` and `.` will be sanitized by `url`, see following example:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=20a9aa366afa3df27f3f021d11e2d40d

> I would prefer to see a 404 response for files that don't exist.

:+1: added with https://github.com/mozilla/neqo/pull/1788/commits/cf0259124c6350eac3251b6f1a1c002f51000795.

mxinden:
@martinthomson would you mind taking another look?

PR #1790, opened 2024-04-03
ci(interop/action): always upload comment data

mxinden:
In https://github.com/mozilla/neqo/pull/1785 the QUIC Network Simulator workflow failed.

https://github.com/mozilla/neqo/actions/runs/8523273988/job/23345192160?pr=1785

This triggers the QUIC Network Simulator Comment workflow.

https://github.com/mozilla/neqo/actions/runs/8524906268

Though currently, the former does not upload the _comment_ artifact on failure, which is to be consumed by the latter.

This commit makes the former always upload the comment data, such that the latter can consume it on failure.

### DIFF
--- a/.github/actions/quic-interop-runner/action.yml
+++ b/.github/actions/quic-interop-runner/action.yml
@@ -89,6 +89,7 @@ runs:
         path: quic-interop-runner/logs
 
     - name: Format GitHub comment
+      if: always()
       run: |
         echo '[**QUIC Interop Runner**](https://github.com/quic-interop/quic-interop-runner)' >> comment
         echo '' >> comment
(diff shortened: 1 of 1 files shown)


larseggert:
Just to check, the other comment workflow doesn't have this issue, right?

larseggert (approved):

mxinden:
> Just to check, the other comment workflow doesn't have this issue, right?

The Bench Comment workflow only posts on success.

https://github.com/mozilla/neqo/blob/07514299a70a8aea8d28a25544e85cf96620d976/.github/workflows/bench-comment.yml#L22

Thus, in case the Bench workflow fails, the Bench Comment workflow never runs and thereby can't fail to find the artifacts.

PR #1792, opened 2024-04-04
refactor: have process_input delegate to process_multiple_input

mxinden:
The `Connection::process_input` and `Connection::process_multiple_input` functions are identical, except that the latter handles multiple `Datagram`s.

To avoid any changes to one without updating the other, have `process_input` simply delegate to `process_multiple_input`.

Commit also does the equivalent change to `neqo_http3::Http3Client`.

---

I would be surprised if `std::iter::once` adds a performance hit. Let's see what the benchmarks say.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -7,7 +7,7 @@
 use std::{
     cell::RefCell,
     fmt::{Debug, Display},
-    mem,
+    iter, mem,
     net::SocketAddr,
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
I'm fairly sure that std::iter::once optimizes out well enough that this won't matter.  We have far too much noise in benchmarks to notice, either way.

At this layer of the code, simply avoid duplication is worth the extra overhead.

PR #1795, opened 2024-04-07
fix(bin): benchmark HPS without zero RTT

mxinden:
The "100-seq-conn/1-1b-resp (aka. HPS)" benchmark supposedly measures the time for 100 uniform handshakes. Though `neqo-client` by default uses a resumption token for zero RTT if available. Thus the first is a normal and all following 99 handshakes are a zero RTT handshake.

To simplify the benchmark and to measure the same handshake type, establish a single connection within the benchmark function and have criterion do all the iterating.

---

Benchmark introduced in https://github.com/mozilla/neqo/pull/1758.

### DIFF
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -13,8 +13,6 @@ use tokio::runtime::Runtime;
 struct Benchmark {
     name: String,
     requests: Vec<u64>,
-    /// Download resources in series using separate connections.
-    download_in_series: bool,
     sample_size: Option<usize>,
 }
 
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
This looks fine, but I noticed that these benchmarks are awfully slow.  It might pay to adjust the time or the number of iterations in criterion.

mxinden:
> This looks fine, but I noticed that these benchmarks are awfully slow. It might pay to adjust the time or the number of iterations in criterion.

:+1: this pull request significantly reduces the HPS benchmark run time. Will keep this in mind for the RPS benchmark, which is still slow (~40s).

```
Benchmarking 1-conn/1-100mb-resp (aka. Download)/client: Collecting 10 samples in estimated 10.844 s (10 iterations)
Benchmarking 1-conn/10_000-1b-seq-resp (aka. RPS)/client: Collecting 100 samples in estimated 39.564 s (100 iterations)
Benchmarking 1-conn/1-1b-resp (aka. HPS)/client: Collecting 100 samples in estimated 8.5971 s (200 iterations)
```

PR #1802, opened 2024-04-09
test: Add my `test.sh` script

larseggert:
Because I keep blowing it away when I `git clean`...

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -1,0 +1,41 @@
+#! /usr/bin/env bash
+
+# This script builds the client and server binaries and runs them in a tmux
+# session side-by-side. The client connects to the server and the server
+# responds with a simple HTTP response. The client and server are run with
+# verbose logging and the qlog output is stored in a temporary directory. The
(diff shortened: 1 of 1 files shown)


test/test.sh
```diff
@@ -0,0 +1,41 @@
...
+tmux -CC \
+        set-option -g default-shell "$(which bash)" \; \
+        new-session "$client && kill -USR2 $tcpdump_pid && touch $tmp/done" \; \
+        split-window -h "$server" \; \
+        split-window -v -f "\
+                until [ -e $tmp/done ]; do sleep 1; done && \
+                tshark -r $tmp/test.pcap -o tls.keylog_file:$tmp/test.tlskey" \; \
+        set remain-on-exit on
+
+# rm -rf "$tmp"
```
martinthomson, line 41: You should probably use `trap 'rm -rf "$tmp"' EXIT` instead.  Especially with `set -e`.

larseggert: Thanks! #1810

PR #1805, opened 2024-04-09
fix: Correctly track sent `CONNECTION_CLOSE` frames

larseggert:
And as a side effect, remove `output_close()`.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1826,7 +1826,7 @@ impl Connection {
             | State::Connected
             | State::Confirmed => {
                 if let Some(path) = self.paths.select_path() {
-                    let res = self.output_path(&path, now);
+                    let res = self.output_path(&path, now, &None);
                     self.capture_error(Some(path), now, 0, res)
                 } else {
                     Ok(SendOption::default())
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1835,7 +1835,16 @@ impl Connection {
             State::Closing { .. } | State::Draining { .. } | State::Closed(_) => {
                 if let Some(details) = self.state_signaling.close_frame() {
                     let path = Rc::clone(details.path());
-                    let res = self.output_close(&details);
+                    // In some error cases, we will not be able to make a new, permanent path.
+                    // For example, if we run out of connection IDs and the error results from
+                    // a packet on a new path, we avoid sending (and the privacy risk) rather
+                    // than reuse a connection ID.
+                    let res = if details.path().borrow().is_temporary() {
```
KershawChang, line 1842: nit: Maybe use `path.path().borrow().is_temporary()` here or move the `clone` above into the `else` branch below.

KershawChang (approved):

martinthomson (commented):
I'm not really a fan of this change.  It moves a bunch of complexity to the output_path function, which is already pretty busy.

Why did you not add frame counting to the existing function if that was the problem?

larseggert:
I thought that the duplication between `output_close` and `output_path` would be good to reduce.

Issue #921, opened 2020-08-10
Server active connection list: avoid duplicates


agrover:
Follow-on to #920, the connection list is a VecDeque. This is good because things put on the list are handled in order, but bad because a single connection can be put on the list multiple times. Either `waiting` should be some type of collection that doesn't add duplicates (but still returns things in order added?) or we should take care to not add a connection to it, if it's already on the list.

larseggert:
@mxinden I think this is OBE with your recent changes?

mxinden:
Yes. Good catch. Thanks.

PR #1806, opened 2024-04-09
fix(bin): don't sleep when events available

mxinden:
A call to `process_*` can emit events as a side effect, e.g. a `ConnectionEvent::StateChange(State::Connected)`. These are not returned from the function call, but instead internally enqueued to be later on consumed through the various `Provider::next_event` implementations.

https://github.com/mozilla/neqo/blob/166b84c5a3307d678f38d9994af9b56b68c6b695/neqo-common/src/event.rs#L9-L15

In the case of `neqo-client` the events are consumed through `self.handler.handle` which internally calls `Provider::next_event`.

A client or server should not go to sleep, waiting for either a UDP datagram to arrive or a timeout to fire, when there are events available.

This commit ensures `ready().await` is only called when no events are available.

---

Extracted from https://github.com/mozilla/neqo/pull/1741.

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -166,6 +166,10 @@ impl super::Client for Connection {
     fn stats(&self) -> neqo_transport::Stats {
         self.stats()
     }
+
+    fn has_events(&self) -> bool {
+        neqo_common::event::Provider::has_events(self)
+    }
 }
 
 impl<'b> Handler<'b> {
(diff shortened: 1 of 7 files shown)


mxinden:
> Also see #921 in case this is something that might affect things here.

Given that `Server::active_connections` always takes all of `self.active` (`mem::take`), I don't think the change proposed in this pull request is affected by #921. I created https://github.com/mozilla/neqo/pull/1818 to test whether the potential duplication in `Http3Server::process` is an issue in general. Let's see what the benchmarks say.

---

> Could we add a test for `servers.rs`?

Done with the latest commit. @larseggert mind taking another look?

larseggert (approved):

PR #1807, opened 2024-04-09
ci: Delete QNS PR comment on successful run

larseggert:
~~Because we only add one when there is a failure, and when things then later get fixed we don't update it again (because we skip the update on success). So the outdated, failed state keeps being shown.~~

Because if we only update on failure, we keep showing the last-failed state after success.

### DIFF
--- a/.github/actions/pr-comment/action.yml
+++ b/.github/actions/pr-comment/action.yml
@@ -5,6 +5,9 @@ inputs:
   name:
     description: 'Artifact name to import comment data from.'
     required: true
+  mode:
+    description: 'Mode of operation (upsert/recreate/delete).'
+    required: true
   token:
     description: 'A Github PAT'
     required: true
--- a/.github/actions/pr-comment/action.yml
+++ b/.github/actions/pr-comment/action.yml
@@ -32,5 +32,6 @@ runs:
     - uses: thollander/actions-comment-pull-request@v2
       with:
         filePath: contents
+        mode: ${{ inputs.mode }}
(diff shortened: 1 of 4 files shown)


mxinden (approved):
Unfortunate to have it show up on each successful pull request then, i.e. more noise. That said, maybe better than the above mentioned confusion.

larseggert:
Do you see a way to only update an existing comment, rather than always add or updated?

mxinden:
On success, we could use `mode` `delete` and ignore a failure for the case where it doesn't exist.

https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#action-inputs

larseggert:
> On success, we could use `mode` `delete`

Oh nice!

.github/actions/pr-comment/action.yml
```diff
@@ -5,6 +5,9 @@ inputs:
   name:
     description: 'Artifact name to import comment data from.'
     required: true
+  mode:
+    description: 'Mode of operation (upsert/recreate/delete).'
+    required: true
```
mxinden, line 10: Variable needs to be forwarded to the `thollander/actions-comment-pull-request` action, no? Or is this somehow happening automatically?

mxinden, line 10:
```suggestion
    default: 'upsert'
```

Would make the change in `bench-comment` obsolete. Just an idea.

larseggert: Doh. Of course.

mxinden (approved):
Neat! Thank you!

PR #1811, opened 2024-04-09
fix: Check that frame types use minimal varint encoding

larseggert:


neqo-transport/src/frame.rs
```diff
@@ -426,8 +426,13 @@ impl<'a> Frame<'a> {
             d(dec.decode_varint())
         }
 
-        // TODO(ekr@rtfm.com): check for minimal encoding
+        // Check for minimal encoding of frame type.
+        let pos = dec.offset();
         let t = dv(dec)?;
+        if Decoder::minimal_varint_len(t).ok_or(Error::ProtocolViolation)? != dec.offset() - pos {
```
larseggert, line 432:
```suggestion
        if Decoder::minimal_varint_len(t).unwrap() != dec.offset() - pos {
```
should also be safe, given that we just decoded `t`.

neqo-transport/src/frame.rs
```diff
@@ -989,14 +994,14 @@ mod tests {
...
-        just_dec(&f, "4030010203");
+        just_dec(&f, "30010203");
 
         // With the length bit.
         let f = Frame::Datagram {
             data: &[1, 2, 3],
             fill: false,
         };
-        just_dec(&f, "403103010203");
+        just_dec(&f, "3103010203");
```
larseggert, line 1004: These were actually not minimally encoded :-)

neqo-transport/src/frame.rs
```diff
@@ -426,8 +426,13 @@ impl<'a> Frame<'a> {
             d(dec.decode_varint())
         }
 
-        // TODO(ekr@rtfm.com): check for minimal encoding
+        // Check for minimal encoding of frame type.
+        let pos = dec.offset();
         let t = dv(dec)?;
+        if Decoder::minimal_varint_len(t).unwrap() != dec.offset() - pos {
```
mxinden, line 432:
```suggestion
        // RFC 9000, Section 12.4:
        //
        // The Frame Type field uses a variable-length integer encoding [...],
        // with one exception. To ensure simple and efficient implementations of
        // frame parsing, a frame type MUST use the shortest possible encoding.
        if Decoder::minimal_varint_len(t).unwrap() != dec.offset() - pos {
```

Maybe worth documenting why a minimal encoding is a MUST. Feel free to ignore.
```

**Review by mxinden — APPROVED** [review]
For what my review is worth.

File: neqo-common/src/codec.rs [hunk 4]
```diff
@@ -15,6 +15,22 @@ pub struct Decoder<'a> {
 }
 
 impl<'a> Decoder<'a> {
+    /// Return the minimal encoded length of a varint with value `v`.
+    #[must_use]
+    pub const fn minimal_varint_len(v: u64) -> Option<usize> {
```
martinthomson, line 20: I'd return a usize and panic if this was passed anything >= 2^62.  The only place you call this does the same.

`Encoder::varint_len()` already exists and does exactly that.

martinthomson (commented):
I don't see a test that hits this new error path.  (The tests you fixed would have, but you changed them.)

neqo-common/src/codec.rs
larseggert: Good suggestion. I'll also use the `match`-pattern there.

larseggert:
> I don't see a test that hits this new error path. (The tests you fixed would have, but you changed them.)

There is `minimal_varint_len` in `codec.rs`?

neqo-common/src/codec.rs
```diff
@@ -15,6 +15,22 @@ pub struct Decoder<'a> {
...
+    pub const fn minimal_varint_len(v: u64) -> usize {
+        match () {
+            () if v < (1 << 6) => 1,
+            () if v < (1 << 14) => 2,
+            () if v < (1 << 30) => 4,
+            () if v < (1 << 62) => 8,
+            () => panic!("Varint value too large"),
+        }
+    }
+
```
martinthomson, line 33: Perhaps I wasn't clear.  This code duplicates the code in the encoder, so you can just use the version in the encoder rather than duplicating it.

```suggestion
```

I'm open to suggestions about a singular place.  You might also do this:

```suggestion
    /// Return the minimal encoded length of a varint with value `v`.
    ///
    /// # Panics
    ///
    /// When `v` is too large.
    #[must_use]
    pub const fn minimal_varint_len(v: u64) -> usize {
        Encoder::varint_len(v)
    }
```

neqo-common/src/codec.rs
```diff
@@ -842,4 +858,22 @@ mod tests {
...
+        assert_eq!(Decoder::minimal_varint_len(0x3fff_ffff), 4);
+        assert_eq!(Decoder::minimal_varint_len(0x4000_0000), 8);
+        assert_eq!(Decoder::minimal_varint_len(0x3fff_ffff_ffff_ffff), 8);
+    }
+
+    #[test]
+    #[should_panic(expected = "Varint value too large")]
+    fn invalid_varint_len() {
+        let _ = Decoder::minimal_varint_len(0x4000_0000_0000_0000);
+    }
```
martinthomson, line 878:
```suggestion
```

neqo-transport/src/frame.rs
```diff
@@ -464,8 +464,18 @@ impl<'a> Frame<'a> {
...
-        // TODO(ekr@rtfm.com): check for minimal encoding
+        // Check for minimal encoding of frame type.
+        let pos = dec.offset();
         let t = dv(dec)?;
+        // RFC 9000, Section 12.4:
+        //
+        // The Frame Type field uses a variable-length integer encoding [...],
+        // with one exception. To ensure simple and efficient implementations of
+        // frame parsing, a frame type MUST use the shortest possible encoding.
+        if Decoder::minimal_varint_len(t) != dec.offset() - pos {
```
martinthomson, line 475:
```suggestion
        if Encoder::varint_len(t) != dec.offset() - pos {
```

neqo-transport/src/frame.rs
```diff
@@ -660,8 +670,11 @@ mod tests {
 
     fn just_dec(f: &Frame, s: &str) {
         let encoded = Encoder::from_hex(s);
-        let decoded = Frame::decode(&mut encoded.as_decoder()).unwrap();
-        assert_eq!(*f, decoded);
+        if let Ok(decoded) = Frame::decode(&mut encoded.as_decoder()) {
+            assert_eq!(*f, decoded);
+        } else {
+            panic!("Failed to decode frame");
+        }
```
martinthomson, line 677:
```suggestion
        let decoded = Frame::decode(&mut encoded.as_decoder()).expect("Failed to decode frame");
        assert_eq!(*f, decoded);
```

If you need a specific message, this is neater.

neqo-common/src/codec.rs
larseggert: Ah! Yeah, sorry that I misunderstood.

martinthomson (approved):

PR #1814, opened 2024-04-10
fix: Make sure the primary path exists at various steps in the handshake

### LINKED ISSUE #1448: Server crashes when client sends a CONNECTION_CLOSE frame.
When a client sends a CONNECTION_CLOSE frame as the first packet to the Neqo server, the server tries to respond with a CONNECTION_CLOSE frame. However, the primary path of the connection has not been initialized yet, causing a null pointer dereference at "neqo-transport/src/path.rs:147".

larseggert:
Fixes #1448

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -46,7 +46,7 @@ use crate::{
     quic_datagrams::{DatagramTracking, QuicDatagrams},
     recovery::{LossRecovery, RecoveryToken, SendProfile},
     recv_stream::RecvStreamStats,
-    rtt::GRANULARITY,
+    rtt::{GRANULARITY, INITIAL_RTT},
     send_stream::SendStream,
     stats::{Stats, StatsCell},
     stream_id::StreamType,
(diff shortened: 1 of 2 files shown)


neqo-transport/tests/server.rs
```diff
@@ -477,6 +477,66 @@ fn bad_client_initial() {
...
+fn bad_client_initial_connection_close() {
+    let mut client = default_client();
+    let mut server = default_server();
+
+    let dgram = client.process(None, now()).dgram().expect("a datagram");
+    let (header, d_cid, s_cid, payload) = decode_initial_header(&dgram, Role::Client);
+    let (aead, hp) = initial_aead_and_hp(d_cid, Role::Client);
+    let (_, pn) = remove_header_protection(&hp, header, payload);
+
+    // let mut payload_enc = Encoder::from(plaintext);
```
mxinden, line 490:
```suggestion
```

Leftover?

neqo-transport/src/connection/mod.rs
```diff
@@ -2861,8 +2862,11 @@ impl Connection {
     {
         qdebug!([self], "Rx ACK space={}, ranges={:?}", space, ack_ranges);
 
+        let Some(path) = self.paths.primary_fallible() else {
```
martinthomson, line 2865: If we can eliminate .primary(), then we can use that name for this function...  Maybe we should do that now anyway.  We want to encourage people to use the fallible option anyway.

martinthomson (approved):
The coverage complaint seems like a reasonable trade.

neqo-transport/src/connection/mod.rs
larseggert: I opened #1828 for this.

PR #1818, opened 2024-04-11
refactor: deduplicate events

mxinden:
See whether https://github.com/mozilla/neqo/issues/921 has any performance impact.

Though arguably, given that it origins from a `HashSet` and given the `contains` call, there is no duplication.

(Calling `dedup` on an unsorted `Vec` is not complete.)

### DIFF
--- a/neqo-http3/src/server.rs
+++ b/neqo-http3/src/server.rs
@@ -131,28 +131,21 @@ impl Http3Server {
     fn process_http3(&mut self, now: Instant) {
         qtrace!([self], "Process http3 internal.");
         let mut active_conns = self.server.active_connections();
-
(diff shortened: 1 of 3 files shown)


martinthomson (commented):
I'm not seeing any impact.  But this seems like a decent change.

mxinden:
Under the assumption that `neqo`'s server logic is for testing purposes only, I believe one can greatly simplify the server related logic in `neqo-transport` and `neqo-http3` instead of optimizing it (this pull request). Closing here in favor of a simpler approach. More to come.

mxinden:
> Closing here in favor of a simpler approach. More to come.

For the record, this has been done in https://github.com/mozilla/neqo/pull/1929.

Issue #1819, opened 2024-04-13
`http3::SendMessage::send_data` can stall


mxinden:
#### Problem

The QUIC layer (`neqo-transport`) notifies the http3 layer (`neqo-http3`) when new data can be written to a stream via `ConnectionEvent::SendStreamWritable`. For example when the maximum stream data is increased by the remote.

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-transport/src/send_stream.rs#L1206-L1216

Note that the event is only emitted in case the stream was previously blocked on the limit, not each time the limit increases.

This event is surfaced through the http3 layer to the user as a `Http3ServerEvent::DataWritable` or `Http3ClientEvent::DataWritable`. The user can then write data to the stream, ultimately calling `neqo_http3::SendMessage::send_data`.

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-http3/src/send_message.rs#L168

`send_data` will check with the underlying `neqo_transport::SendStream` how much bytes are available to send. To make sure it can send at least a header & a 1 byte payload, it checks that the available bytes are not `<=2`.

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-http3/src/send_message.rs#L177-L182

If it is `<=2` it returns with `Ok(0)`, effectively wasting the `DataWritable` event.

In this case:

- The `neqo_transport::SendStream` has available bytes, even though small (`0< available <=2`), and thus is waiting for input from `neqo_http3::SendMessage`. It will not emit more `DataWritable` events e.g. on maximum stream data increase, given that it has data available (even though small `0< available <=2`).
- The `neqo_http3::SendMessage` is waiting for `neqo_transport::SendStream` to have more bytes available, i.e. it is waiting for a `DataWritable` event.

Thus sending of data on the stream stalls indefinitely.

#### Potential solutions

1. Emit a `DataWritable` event on each increase of `neqo_transport::SendStream::avail()` (connection limit, stream limit, `tx_buf` limit).
    - This can lead to a lot of events and thus a lot of resulting work.
[...]

PR #1821, opened 2024-04-13
fix(transport): only emit SendStreamWritable on ack if prev. blocked

mxinden:
`neqo_transport::SendStream` emits `ConnectionEvent::SendStreamWritable` to inform a user that they can write more data into a stream.

It emits this event in 3 cases:

- on `SendStream::new`
- on `SendStream::set_max_data`
- on `SendStream::mark_as_acked`

Emitting a `SendStreamWritable` event, even though writing wasn't previously blocked, is useless. That is why `SendStream::set_max_data` only emits the event if sending was previously blocked on stream flow control.

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-transport/src/send_stream.rs#L1206-L1216

This commit adds the corresponding check to `SendStream::mark_as_acked` as well. I.e. it checks whether sending was previously blocked on the tx buffer size, before emitting a `SendStreamWritable` event.

This is especially relevant for `SendStream::mark_as_acked`, as each received ack would otherwise result in a `SendStreamWritable` event. The event in itself isn't expensive. The logic it triggers is.

---

As an example, I was seeing > 200 `SendStreamWritable` events queued at once in the [`Download` benchmark](https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-bin/benches/main.rs#L30-L34).

I suggest prioritizing the related https://github.com/mozilla/neqo/issues/1819 (bug) before this pull request (optimization).

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -2514,7 +2514,7 @@ mod tests {
         conn_events.events().next().unwrap();
 
         // Fill the tx buffer.
-        let res = s.send(&[4; SEND_BUFFER_SIZE]).unwrap();
+        let res = s.send(&vec![4; SEND_BUFFER_SIZE]).unwrap();
         assert_eq!(res, SEND_BUFFER_SIZE);
         assert!(conn_events.events().next().is_none());
         // The stream is now constrained by the tx buffer.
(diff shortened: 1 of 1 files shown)


mxinden:
I expect the latest benchmark stall to be due to https://github.com/mozilla/neqo/issues/1819.

mxinden:
Note that this pull request would be made obsolete by https://github.com/mozilla/neqo/pull/1838. I prefer https://github.com/mozilla/neqo/pull/1838 over this pull request.

PR #1825, opened 2024-04-15
refactor(SendStream): remove borrow-checker work-around

mxinden:
With recent Rust versions, the Rust borrow-checker has become more capable, especially when it comes to borrowing subsets of `self`.

This commit removes the no longer necessary borrow-checker work-around.

---

Sorry for the many tiny pull requests today. Let me know if you don't think they are worth the noise.


larseggert:
> Sorry for the many tiny pull requests today. Let me know if you don't think they are worth the noise.

No worries, we'll take contributions as they come.

larseggert (approved):

PR #1827, opened 2024-04-15
chore: Bump qlog to 0.13 and deal with the fallout

### LINKED ISSUE #1826: Bump qlog to 0.13
It's out: https://github.com/cloudflare/quiche/commit/00638b167684b7631feef4cbe71dff904bb097cf

That should also enable us logging padding length.

larseggert:
Also make `log` and `qlog` workspace dependencies.

Fixes #1826


neqo-bin/Cargo.toml
```diff
@@ -34,7 +34,7 @@ neqo-crypto = { path = "./../neqo-crypto" }
 neqo-http3 = { path = "./../neqo-http3" }
 neqo-qpack = { path = "./../neqo-qpack" }
 neqo-transport = { path = "./../neqo-transport" }
-qlog = { version = "0.12", default-features = false }
+qlog = { version = "0.13", default-features = false }
```
mxinden, line 37: How about promoting `qlog` to a workspace dependency? Thus a version bump would have to touch a single file only.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

neqo-transport/src/qlog.rs
```diff
@@ -392,8 +392,14 @@ impl From<&Frame<'_>> for QuicFrame {
         match frame {
             // TODO: Add payload length to `QuicFrame::Padding` once
             // https://github.com/cloudflare/quiche/pull/1745 is available via the qlog crate.
```
mxinden, line 394:
```suggestion
```

Outdated now, right?

mxinden (approved):

KershawChang (approved):

neqo-bin/Cargo.toml
larseggert: Is there a downside if not every workspace member actually needs the dependency? (Such as `fuzz`.)

mxinden: I am not aware of one @larseggert. We have been using them quite extensively in https://github.com/libp2p/rust-libp2p/blob/bee8199c0db79e2b621e3a076ed09ece65337553/Cargo.toml#L73-L127.

mxinden (approved):
:+1:

KershawChang (approved):

PR #1833, opened 2024-04-16
fix: `Paths::primary()` is now `Paths::primary_fallible()`

### LINKED ISSUE #1813: `Paths::primary()` is a foot-gun
https://github.com/mozilla/neqo/blob/c44e53647d922359ef859d18d86dd9eb7fe4891c/neqo-transport/src/path.rs#L148-L153

In error/edge cases during handshake, such as #1448, calls to `Paths::primary()` happen when the condition mentioned in the comment is false. It would be nice if we had better protections around that crashing things.

### LINKED ISSUE #1828: If we can eliminate .primary(), then we can use that name for this function...  Maybe we should do that now anyway.  We want to encourage people to use the fallible option anyway.
If we can eliminate .primary(), then we can use that name for this function...  Maybe we should do that now anyway.  We want to encourage people to use the fallible option anyway.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1814#discussion_r1566595669_

larseggert:
And deal with the fallout.

Fixes #1828
Fixes #1813

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -291,7 +291,7 @@ impl Debug for Connection {
             "{:?} Connection: {:?} {:?}",
             self.role,
             self.state,
-            self.paths.primary_fallible()
+            self.paths.primary()
         )
     }
 }
@@ -591,7 +591,11 @@ impl Connection {
     fn make_resumption_token(&mut self) -> ResumptionToken {
         debug_assert_eq!(self.role, Role::Client);
         debug_assert!(self.crypto.has_resumption_token());
-        let rtt = self.paths.primary().borrow().rtt().estimate();
+        let rtt = self.paths.primary().map_or_else(
+            || RttEstimate::default().estimate(),
+            |p| p.borrow().rtt().estimate(),
+        );
+
         self.crypto
             .create_resumption_token(
                 self.new_token.take_token(),
@@ -610,7 +614,7 @@ impl Connection {
     /// a value of this approximate order.  Don't use this for loss recovery,
     /// only use it where a more precise value is not important.
     fn pto(&self) -> Duration {
-        self.paths.primary_fallible().map_or_else(
+        self.paths.primary().map_or_else(
             || RttEstimate::default().pto(PacketNumberSpace::ApplicationData),
             |p| p.borrow().rtt().pto(PacketNumberSpace::ApplicationData),
         )
@@ -745,7 +749,11 @@ impl Connection {
         if !init_token.is_empty() {
             self.address_validation = AddressValidationInfo::NewToken(init_token.to_vec());
         }
-        self.paths.primary().borrow_mut().rtt_mut().set_initial(rtt);
+        if let Some(path) = self.paths.primary() {
+            path.borrow_mut().rtt_mut().set_initial(rtt);
+        } else {
+            return Err(Error::InternalError);
+        }
         self.set_initial_limits();
         // Start up TLS, which has the effect of setting up all the necessary
         // state for 0-RTT.  This only stages the CRYPTO frames.
@@ -785,7 +793,7 @@ impl Connection {
         // If we are able, also send a NEW_TOKEN frame.
         // This should be recording all remote addresses that are valid,
         // but there are just 0 or 1 in the current implementation.
-        if let Some(path) = self.paths.primary_fallible() {
+        if let Some(path) = self.paths.primary() {
             if let Some(token) = self
                 .address_validation
                 .generate_new_token(path.borrow().remote_address(), now)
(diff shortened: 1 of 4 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2937,7 +2969,9 @@ impl Connection {
             // Remove the randomized client CID from the list of acceptable CIDs.
             self.cid_manager.remove_odcid();
             // Mark the path as validated, if it isn't already.
-            let path = self.paths.primary();
+            let Some(path) = self.paths.primary() else {
+                return Err(Error::NoAvailablePath);
+            };
```
martinthomson, line 2974:
```suggestion
            let path = self.paths.primary().ok_or(Error::NoAvailablePath)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2412,10 +2445,10 @@ impl Connection {
                 // That's OK, they can try guessing this.
                 ConnectionIdEntry::random_srt()
             };
-            self.paths
-                .primary()
-                .borrow_mut()
-                .set_reset_token(reset_token);
+            let Some(path) = self.paths.primary() else {
+                return Err(Error::NoAvailablePath);
+            };
```
martinthomson, line 2450:
```suggestion
            let path = self.paths.primary().ok_or(Error::NoAvailablePath)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1782,7 +1810,10 @@ impl Connection {
             // has to use the existing address.  So only pay attention to a preferred
             // address from the same family as is currently in use. More thought will
             // be needed to work out how to get addresses from a different family.
-            let prev = self.paths.primary().borrow().remote_address();
+            let Some(path) = self.paths.primary() else {
+                return Err(Error::NoAvailablePath);
+            };
```
martinthomson, line 1815:
```suggestion
            let path = self.paths.primary().ok_or(Error::NoAvailablePath)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1728,8 +1751,13 @@ impl Connection {
             // Pointless migration is pointless.
             return Err(Error::InvalidMigration);
         }
-        let local = local.unwrap_or_else(|| self.paths.primary().borrow().local_address());
-        let remote = remote.unwrap_or_else(|| self.paths.primary().borrow().remote_address());
+
+        let Some(path) = self.paths.primary() else {
+            return Err(Error::InvalidMigration);
+        };
```
martinthomson, line 1757:
```suggestion
        let path = self.paths.primary().ok_or(Error::InvalidMigration)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1229,8 +1244,11 @@ impl Connection {
             assert_ne!(self.version, version);
 
             qinfo!([self], "Version negotiation: trying {:?}", version);
-            let local_addr = self.paths.primary().borrow().local_address();
-            let remote_addr = self.paths.primary().borrow().remote_address();
+            let Some(path) = self.paths.primary() else {
+                return Err(Error::NoAvailablePath);
+            };
```
martinthomson, line 1249:
```suggestion
            let path = self.paths.primary().ok_or(Error::NoAvailablePath)?;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1125,7 +1133,13 @@ impl Connection {
         }
         // At this point, we should only have the connection ID that we generated.
         // Update to the one that the server prefers.
-        let path = self.paths.primary();
+        let Some(path) = self.paths.primary() else {
+            self.stats
+                .borrow_mut()
+                .pkt_dropped("Retry without an existing path");
+            return;
```
martinthomson, line 1140: Got a test for this?

neqo-transport/src/connection/mod.rs
```diff
@@ -746,7 +750,11 @@ impl Connection {
         if !init_token.is_empty() {
             self.address_validation = AddressValidationInfo::NewToken(init_token.to_vec());
         }
-        self.paths.primary().borrow_mut().rtt_mut().set_initial(rtt);
+        if let Some(path) = self.paths.primary() {
+            path.borrow_mut().rtt_mut().set_initial(rtt);
+        } else {
+            return Err(Error::InternalError);
+        }
```
martinthomson, line 757:
```suggestion
        self.paths.primary().ok_or(Error::InternalError)?.borrow_mut().rtt_mut().set_initial(rtt);
```

`rustfmt` might have something to say about that one though.

martinthomson (approved):
Some suggestions...

neqo-transport/src/connection/mod.rs
larseggert: I'm not sure I can come up with one. `new_client()` does `setup_handshake_path()`, which calls `make_permanent()` on that first path, which should always push it into index 0, i.e., make it primary.

martinthomson: Fair enough.  I thought I'd ask though.

PR #1834, opened 2024-04-16
ci: Build Firefox with current neqo

larseggert:


larseggert:
So it's building (yay!) but given the time it takes to do so, not sure it's productive to enable this for each PR push.

mxinden:
> So it's building (yay!) but given the time it takes to do so, not sure it's productive to enable this for each PR push.

How about treating it as an _asynchronous_ CI step, i.e.:

- Don't make it required.
- Make it post a comment on failure.

Thereby we might not catch a regression before merging, but we will know through the comment posted after merge, and can thus easily revert without having to `git bisect` to find the regression in the first place.

KershawChang (commented):
I think we probably want to setup [mozconfig](https://firefox-source-docs.mozilla.org/setup/configuring_build_options.html) file to create debug and release build.

larseggert:
> I think we probably want to setup [mozconfig](https://firefox-source-docs.mozilla.org/setup/configuring_build_options.html) file to create debug and release build.

Good idea. I'm also thinking about building MacOS and Windows versions, and making the final results available as artifacts.

larseggert:
@KershawChang do we need both debug and release builds? They really kill our action cache. Can we live with one?

Ditto for macOS, Linux and Windows - can we not do one of these?

.github/actions/maximize-build-space/action.yml
```diff
@@ -0,0 +1,206 @@
+# https://github.com/easimon/maximize-build-space/blob/fadc013e293a3453768b4ddb9db8c85104752807/action.yml
```
martinthomson, line 1: Can we not request approval to run this action at the organization level?

If there is an issue tracking that, maybe flag that in a comment here.

.github/workflows/firefox.yml
```diff
@@ -0,0 +1,136 @@
+name: Firefox
+on:
+  push:
+    branches: ["main"]
+    paths-ignore: ["*.md", "*.png", "*.svg", "LICENSE-*"]
+  pull_request:
+    branches: ["main"]
+    paths-ignore: ["*.md", "*.png", "*.svg", "LICENSE-*"]
```
martinthomson, line 8: This is going to take an awfully long time to run.  Do we really want to hold pull requests up on this completing?

.github/workflows/firefox.yml
```diff
@@ -0,0 +1,136 @@
...
+          # Get qlog version used by neqo
+          cargo generate-lockfile
+          QLOG_VERSION=$(cargo pkgid qlog | cut -d@ -f2)
+          rm Cargo.lock
+          cd mozilla-unified
+          {
+            echo '[[audits.qlog]]'
+            echo 'who = "CI"'
+            echo 'criteria = "safe-to-deploy"'
+            echo "version = \"$QLOG_VERSION\""
```
martinthomson, line 99: Is this the only dependency that we need to tweak like this?

It seems like we have a general problem here in that we might bump a dependency and have to mark it as audited.  Or do we assume that all other dependencies always match what is in gecko proper?

martinthomson (approved):
LGTM, modulo a few nits.

.github/actions/maximize-build-space/action.yml
larseggert: I can, but I get the feeling that the sec team is really antsy about allowing actions that are under the control of a single person.

.github/workflows/firefox.yml
larseggert: This was the only case so far where we were using something newer than Fx. My proposal would be that if we encounter others, we revise the workflow accordingly.

.github/workflows/firefox.yml
larseggert: Not sure. @mxinden suggested to make this an async CI step (see https://github.com/mozilla/neqo/pull/1834#issuecomment-2060532361), but I don't quite know how to do this in practice.

FWIW, merging isn't blocked by this, since we currently only require these four CI workflows to pass for a merge: https://github.com/mozilla/neqo/settings/branch_protection_rules/46171042

PR #1835, opened 2024-04-16
fix(SendMessage): always use available send space

mxinden:
Always use up send space on QUIC layer to ensure receiving `ConnectionEvent::SendStreamWritable` event when more send space is available.

See https://github.com/mozilla/neqo/issues/1819 for details. This commit implements option 2.

Fixes https://github.com/mozilla/neqo/issues/1819.

---

Not particularly happy with the fix. Though, as of today, I am considering this to be the least intrusive option. Will ponder a bit more before marking this as ready-for-review. Input welcome anyways.

### DIFF
--- a/neqo-http3/src/send_message.rs
+++ b/neqo-http3/src/send_message.rs
@@ -6,7 +6,7 @@
 
 use std::{cell::RefCell, cmp::min, fmt::Debug, rc::Rc};
 
-use neqo_common::{qdebug, qtrace, Encoder, Header, MessageType};
+use neqo_common::{qdebug, qerror, qtrace, Encoder, Header, MessageType};
 use neqo_qpack::encoder::QPackEncoder;
 use neqo_transport::{Connection, StreamId};
 
@@ -166,7 +166,12 @@ impl Stream for SendMessage {
 }
 impl SendStream for SendMessage {
     fn send_data(&mut self, conn: &mut Connection, buf: &[u8]) -> Res<usize> {
-        qtrace!([self], "send_body: len={}", buf.len());
+        qtrace!([self], "send_data: len={}", buf.len());
+
+        if buf.is_empty() {
+            qerror!([self], "zero-length send_data");
+            return Err(Error::InvalidInput);
+        }
 
         self.state.new_data()?;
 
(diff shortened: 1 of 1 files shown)


neqo-http3/src/send_message.rs
```diff
@@ -166,7 +166,12 @@ impl Stream for SendMessage {
 }
 impl SendStream for SendMessage {
     fn send_data(&mut self, conn: &mut Connection, buf: &[u8]) -> Res<usize> {
-        qtrace!([self], "send_body: len={}", buf.len());
+        qtrace!([self], "send_data: len={}", buf.len());
+
+        if buf.is_empty() {
+            qerror!([self], "zero-length send_data");
+            return Err(Error::InvalidInput);
+        }
```
mxinden, line 174: Same as in `neqo_transport::SendStream::send_internal` which is called by this function anyways. Thus just moving the `return Err` up the stack, which allows the `else if available <= 2` to depend on `buf` not to be empty.

https://github.com/mozilla/neqo/blob/3e537097cbd6e75b861d0d859f9c5eb9c233a723/neqo-transport/src/send_stream.rs#L1253-L1256.

neqo-http3/src/send_message.rs
```diff
@@ -177,8 +182,20 @@ impl SendStream for SendMessage {
...
+            // header (min 2 bytes), but no data. Don't return `Ok(0)`, but
+            // instead use up `available` anyways, with a small but non-zero
+            // data frame, buffering the remainder, thereby signaling to the
+            // QUIC layer that more is needed, and thus ensure to receive
+            // [`neqo_transport::ConnectionEvent::SendStreamWritable`] when more
+            // send space is available.
+            const MIN_DATA: usize = 1; // arbitrary small value, larger zero
+            return self
+                .send_data_atomic(conn, &buf[..MIN_DATA])
+                .map(|()| MIN_DATA);
```
martinthomson, line 198: I'd like to see a test showing that this is OK.  I'm fairly sure that it isn't.

You aren't sending a DATA frame header for this.  That header is always at least 2 bytes.

martinthomson (changes requested):

neqo-http3/src/send_message.rs
mxinden: @martinthomson the DATA frame header is sent as part of the call to `self.send_data_atomic`:

https://github.com/mozilla/neqo/blob/d74ad5ea9b6dbed83674fc1963a6424c79c12588/neqo-http3/src/send_message.rs#L295-L305

Not saying that this doesn't need a test.

martinthomson: Ah, so that will result in us sending three bytes to the stream, in a case where we only have space for 2 or less.  How does that ever work?

mxinden: > in a case where we only have space for 2 or less

We will use up the available space and buffer the remaining. E.g. say that we have 2 bytes available, we send the header (2bytes) and buffer 1 byte. 

Once more space is available, we get an event (because we previously exhausted the send space) and can thus first empty the 1byte from the buffer and then send more goodput. 

Does that make sense @martinthomson?

(Again, not particularly happy with this hack. Still thinking about alternatives.)

mxinden:
Closing here in favor of https://github.com/mozilla/neqo/pull/1838.

PR #1837, opened 2024-04-17
fix(qns): fallback if server is not sending NEW_TOKEN

mxinden:
https://github.com/mozilla/neqo/pull/1676 ensured for the client to wait for a resumption token from the server if one is needed, that is when running the `resumption` QUIC Interop testcase. #1676 accidentally removed the fallback using a session ticked if no resumption token was available.

https://github.com/mozilla/neqo/pull/1676/files#diff-64d071700a607884926de1b2caaf1588558fceda6d3a0e899cfb3c3b7e5ae41fL1353-L1359

This commit re-introduces the fallback and enables the `resumption` QUIC Interop testcase on CI.

Along the way, it simplifies `Runner::run` and reserves the `resume` command line argument for the QUIC Interop testcases only. Let me know if this adds too much noise to the actual fix. Happy to extract it into a separate pull request.

---

Test failure reported in https://github.com/mozilla/neqo/pull/1786#issuecomment-2060058643.

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -71,6 +71,6 @@ jobs:
           name: 'neqo-latest'
           image: ${{ steps.docker_build_and_push.outputs.imageID }}
           url: https://github.com/mozilla/neqo
-          test: handshake,ecn,keyupdate
+          test: handshake,ecn,keyupdate,resumption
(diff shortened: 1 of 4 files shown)


martinthomson (approved):

mxinden:
Currently debugging neqo -> msquic resumption test failure.

PR #1847, opened 2024-04-18
ci: Parallelize the QNS run

larseggert:
WIP


KershawChang (approved):

mxinden:
~10 minutes per run -- cool! Lots better than the previous sequential execution.

---

With the new QNS comment, I am afraid we will not notice regressions. It is difficult for me to see whether a failure is new, or exists on `main` as well.

Once we have https://github.com/mozilla/neqo/issues/1849, we could reduce the set of implementations to test against, to the subset that succeeds across all testcases.

Or we could show the diff of failures between `main` and a pull request in the QNS comment.

Thoughts?

larseggert:
Yeah, I need to cache the results for the main branch and highlight differences to the PR, similar how we do it for the benchmarks.

PR #1848, opened 2024-04-19
fix(bin): consider env when no verbosity flag is set

mxinden:
There are two ways to specify the maximum log level of `neqo-server` and `neqo-client`. Via the `RUST_LOG` environment variable and since https://github.com/mozilla/neqo/pull/1692 via the verbosity flags (e.g. `-v`).

Previously, if no verbosity flag was provided, `INFO` was set as the maximum log level, the `RUST_LOG` environment variable was simply ignored.

With this commit, if no flag is set, the `RUST_LOG` environment variable is considered, and only then the `env_logger` crate default value (`ERROR`) is used.

In other words, the precedence order now is:

1. command line flags (e.g. `-v`)
2. `RUST_LOG` environment variable
3. default `ERROR` level

---

Among other places, ensures the `RUST_LOG=debug` in the QUIC Interop tests isn't ignored.

https://github.com/mozilla/neqo/blob/4fc4d16744ede5b099dd5bb4bfd5922d07deb046/qns/interop.sh#L15

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -105,9 +105,6 @@ type Res<T> = Result<T, Error>;
 #[allow(clippy::struct_excessive_bools)] // Not a good use of that lint.
 pub struct Args {
     #[command(flatten)]
-    verbose: Option<clap_verbosity_flag::Verbosity>,
-
-    #[command(flatten)]
     shared: SharedArgs,
 
     urls: Vec<Url>,
@@ -180,7 +177,6 @@ impl Args {
     pub fn new(requests: &[u64]) -> Self {
         use std::str::FromStr;
         Self {
-            verbose: None,
             shared: crate::SharedArgs::default(),
             urls: requests
                 .iter()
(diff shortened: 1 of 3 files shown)


mxinden:
We can as well set the final default log level to `INFO` instead of `ERROR`. Not sure whether that interferes with Firefox though. Does Firefox explicitly set the log level?

``` diff
modified   neqo-common/src/log.rs
@@ -12,7 +12,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use env_logger::Builder;
+use env_logger::{Builder, Env};
 
 #[macro_export]
 macro_rules! do_log {
@@ -58,7 +58,7 @@ pub fn init(level_filter: Option<log::LevelFilter>) {
     }
 
     INIT_ONCE.call_once(|| {
-        let mut builder = Builder::from_env("RUST_LOG");
+        let mut builder = Builder::from_env(Env::default().default_filter_or("info"));
         if let Some(filter) = level_filter {
             builder.filter_level(filter);
         }
```

larseggert (commented):
Is there a way to (easily) move this to `neqo-bin/lib.rs` instead of doing it in both the client and server?

mxinden:
> Is there a way to (easily) move this to `neqo-bin/lib.rs` instead of doing it in both the client and server?

@larseggert I moved the `verbose` field into `SharedArgs`. Not sure how to de-deuplicate this further. Mind taking another look?

larseggert (approved):

PR #1850, opened 2024-04-22
fix: Pad Client Initial with random data

larseggert:
Instead of with zeros, which apparently makes us fingerprint-able: https://github.com/quicwg/base-drafts/issues/3333#issuecomment-580808970

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -22,9 +22,9 @@ use neqo_common::{
     qlog::NeqoQlog, qtrace, qwarn, Datagram, Decoder, Encoder, Role,
 };
 use neqo_crypto::{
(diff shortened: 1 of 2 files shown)


larseggert:
Making this a draft PR until @martinthomson has a chance to weigh in. The other option would be to consider switching to padding frames inside the QUIC envelope.

martinthomson:
I really don't think that identifying an Initial packet as neqo is something we should be trying to stop.  Yes, checking for zero padding is easier than taking a look inside the packet, but there are lots of things that point to this implementation, or Firefox, as the source of the request.

larseggert:
@martinthomson so you think "do nothing" is the way to go here? (Just to clarify.)

martinthomson:
Yeah, zeros are cheaper to produce and less complex, so nothing is best.

PR #1853, opened 2024-04-23
chore: Fix clippy nightly warnings

larseggert:


neqo-crypto/src/aead_null.rs
```diff
@@ -4,8 +4,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![cfg(feature = "disable-encryption")]
-
```
mxinden, line 8: Just to double check, this can be removed, as the module is already feature flagged in `lib.rs`, correct?

https://github.com/mozilla/neqo/blob/2e4f0271fc50a8347d8c7525c66d3237a425fcfa/neqo-crypto/src/lib.rs#L11-L12

neqo-crypto/src/constants.rs
```diff
@@ -27,7 +27,7 @@ pub const TLS_EPOCH_APPLICATION_DATA: Epoch = 3_u16;
 macro_rules! remap_enum {
     { $t:ident: $s:ty { $( $n:ident = $v:path ),+ $(,)? } } => {
         pub type $t = $s;
-        $( pub const $n: $t = $v as $t; )+
+        $(#[allow(clippy::cast_possible_truncation)] pub const $n: $t = $v as $t; )+
```
mxinden, line 30: For my understanding: why are all these `c_uint`s (effectively Rust `u32`) mapped to smaller types, and thus risking possible cast truncations?

neqo-crypto/src/aead_null.rs
larseggert: Apparently so. (The change suggestion came from clippy.)

neqo-crypto/src/constants.rs
larseggert: I don't know. Question for @martinthomson I guss. Compatibility with NSS?

martinthomson: This is a mismatch in style between Rust and C.  In NSS, we use enums (which are essentially untyped, hence `c_uint`) to define a bunch of constants.

However, the TLS protocol defines these types to have a fixed-size type, like `u16`.  If we are to present these types in our API in a way that makes sense (i.e., does not allow for values outside the permitted range) then they need to be cast down.

What I'd prefer is if we could ask the compiler to fail if a truncation occurred.  But we don't seem to have that option, or at least we didn't at the time I wrote this.

martinthomson (approved):

PR #1854, opened 2024-04-23
fix: Include an ACK with a CONNECTION_CLOSE

### LINKED ISSUE #1161: Send ACK with CONNECTION_CLOSE
Servers have observed that reducing ACK frequency increases the probability that clients that close connections without acknowledging some packets.  We could help a little here by sending acknowledgments in the CONNECTION_CLOSE.  The code should be reusable.  Tweaking tests might be a fun exercise, but there aren't that many tests for connection closing, so it might not be too bad.

larseggert:
Fixes #1161

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2189,6 +2189,28 @@ impl Connection {
         (tokens, ack_eliciting, padded)
     }
 
+    fn write_ack_frame(
+        &mut self,
+        builder: &mut PacketBuilder,
+        space: PacketNumberSpace,
+        now: Instant,
+        path: &PathRef,
+        tokens: &mut Vec<RecoveryToken>,
+    ) {
+        let limit = builder.limit();
+        builder.set_limit(limit - ClosingFrame::MIN_LENGTH);
+        self.acks.immediate_ack(now);
+        self.acks.write_frame(
+            space,
+            now,
+            path.borrow().rtt().estimate(),
+            builder,
+            tokens,
+            &mut self.stats.borrow_mut().frame_tx,
+        );
+        builder.set_limit(limit);
+    }
+
     /// Build a datagram, possibly from multiple packets (for different PN
     /// spaces) and each containing 1+ frames.
     #[allow(clippy::too_many_lines)] // Yeah, that's just the way it is.
@@ -2254,18 +2276,7 @@ impl Connection {
             if let Some(ref close) = closing_frame {
                 if builder.remaining() > ClosingFrame::MIN_LENGTH + RecvdPackets::MAX_ACK_LEN {
                     // Include an ACK frame with the CONNECTION_CLOSE.
-                    let limit = builder.limit();
-                    builder.set_limit(limit - ClosingFrame::MIN_LENGTH);
-                    self.acks.immediate_ack(now);
-                    self.acks.write_frame(
-                        *space,
-                        now,
-                        path.borrow().rtt().estimate(),
-                        &mut builder,
-                        &mut tokens,
-                        &mut self.stats.borrow_mut().frame_tx,
-                    );
-                    builder.set_limit(limit);
+                    self.write_ack_frame(&mut builder, *space, now, path, &mut tokens);
                 }
                 // ConnectionError::Application is only allowed at 1RTT.
                 let sanitized = if *space == PacketNumberSpace::ApplicationData {
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -56,7 +56,7 @@ use crate::{
         self, TransportParameter, TransportParameterId, TransportParameters,
         TransportParametersHandler,
     },
-    tracking::{AckTracker, PacketNumberSpace, SentPacket},
+    tracking::{AckTracker, PacketNumberSpace, RecvdPackets, SentPacket},
     version::{Version, WireVersion},
     AppError, ConnectionError, Error, Res, StreamId,
 };
@@ -2252,6 +2252,18 @@ impl Connection {
             let payload_start = builder.len();
             let (mut tokens, mut ack_eliciting, mut padded) = (Vec::new(), false, false);
             if let Some(ref close) = closing_frame {
+                if builder.remaining() > ClosingFrame::min_length() + RecvdPackets::max_len() {
+                    // Include an ACK frame with the CONNECTION_CLOSE.
+                    self.acks.immediate_ack(now);
+                    self.acks.write_frame(
+                        *space,
+                        now,
+                        path.borrow().rtt().estimate(),
+                        &mut builder,
+                        &mut tokens,
+                        &mut self.stats.borrow_mut().frame_tx,
+                    );
+                }
                 // ConnectionError::Application is only allowed at 1RTT.
                 let sanitized = if *space == PacketNumberSpace::ApplicationData {
                     None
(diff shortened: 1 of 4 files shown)


neqo-transport/src/connection/tests/close.rs
```diff
@@ -40,7 +40,11 @@ fn connection_close() {
 
     client.close(now, 42, "");
 
+    let stats_before = client.stats().frame_tx;
     let out = client.process(None, now);
+    let stats_after = client.stats().frame_tx;
+    assert!(stats_after.connection_close == stats_before.connection_close + 1);
```
mxinden, line 46:
```suggestion
    assert_eq!(stats_after.connection_close. stats_before.connection_close + 1);
```

Nit: would print both values on failure, making debugging easier.

neqo-transport/src/connection/tests/close.rs
```diff
@@ -40,7 +40,11 @@ fn connection_close() {
...
+    assert!(stats_after.ack == stats_before.ack + 1);
```
mxinden, line 47:
```suggestion
    assert_eq!(stats_after.ack, stats_before.ack + 1);
```

neqo-transport/src/connection/tests/close.rs
```diff
@@ -57,7 +61,11 @@ fn connection_close_with_long_reason_string() {
     let long_reason = String::from_utf8([0x61; 2048].to_vec()).unwrap();
     client.close(now, 42, long_reason);
 
+    let stats_before = client.stats().frame_tx;
     let out = client.process(None, now);
+    let stats_after = client.stats().frame_tx;
+    assert!(stats_after.connection_close == stats_before.connection_close + 1);
```
mxinden, line 67:
```suggestion
    assert_eq!(stats_after.connection_close, stats_before.connection_close + 1);
```

neqo-transport/src/connection/tests/close.rs
```diff
@@ -57,7 +61,11 @@ fn connection_close_with_long_reason_string() {
...
+    assert!(stats_after.ack == stats_before.ack + 1);
```
mxinden, line 68:
```suggestion
    assert_eq!(stats_after.ack. stats_before.ack + 1);
```

mxinden (commented):
Potentially silly question, does the order matter? In other words, do some servers ignore ack frames after a closing frame?

mxinden (commented):
~Potentially silly question, does the order matter? In other words, do some servers ignore ack frames after a closing frame?~

Edit: Spotty internet connection on train.

neqo-transport/src/connection/mod.rs
```diff
@@ -2252,6 +2252,18 @@ impl Connection {
             let payload_start = builder.len();
             let (mut tokens, mut ack_eliciting, mut padded) = (Vec::new(), false, false);
             if let Some(ref close) = closing_frame {
+                if builder.remaining() > ClosingFrame::min_length() + RecvdPackets::max_len() {
```
martinthomson, line 2255: This little chunk of code (under `if let Some(ref close) ...` is large enough now that it probably needs its own function.

As for this line of code, I think that what you want to do is ask the closing frame how much space it is going to take up, then - if there is enough space for a minimal ACK, which might assume ECN for the sake of simplicity - you would temporarily reduce "remaining" by the closing frame size, write the ACK, add back the closing frame size, and off you go.

The close frame size computation doesn't need to take "remaining" into account.  We just need to know how big it would prefer to be.

neqo-transport/src/tracking.rs
```diff
@@ -559,6 +559,12 @@ impl RecvdPackets {
         }
     }
 
+    pub const fn max_len() -> usize {
+        // The worst possible ACK frame, assuming only one range.
+        // Note that this assumes one byte for the type and count of extra ranges.
+        1 + 8 + 8 + 1 + 8
+    }
```
martinthomson, line 566: This can be a constant on the type rather than a function.

neqo-transport/src/connection/state.rs
```diff
@@ -156,10 +156,14 @@ impl ClosingFrame {
         }
     }
 
-    pub fn write_frame(&self, builder: &mut PacketBuilder) {
+    pub const fn min_length() -> usize {
```
martinthomson, line 159: A constant would be better.

neqo-transport/src/connection/tests/close.rs
```diff
@@ -57,7 +64,14 @@ fn connection_close_with_long_reason_string() {
...
     client.close(now, 42, long_reason);
 
+    let stats_before = client.stats().frame_tx;
     let out = client.process(None, now);
+    let stats_after = client.stats().frame_tx;
+    assert_eq!(
+        stats_after.connection_close,
+        stats_before.connection_close + 1
+    );
+    assert_eq!(stats_after.ack, stats_before.ack + 1);
```
martinthomson, line 74: Why are you seeing an ACK sent in this case?  I thought that the close would squeeze out the ACK.

neqo-transport/src/connection/mod.rs
```diff
@@ -2252,6 +2274,10 @@ impl Connection {
             let payload_start = builder.len();
             let (mut tokens, mut ack_eliciting, mut padded) = (Vec::new(), false, false);
             if let Some(ref close) = closing_frame {
```
martinthomson, line 2276: I was suggesting that the entirety of this block (not the inner one; down to 2265(left)/2291(right)) be moved to a `write_closing_frames` function or similar.  Maybe there are too many arguments for that to look neat, I don't know.

neqo-transport/src/tracking.rs
```diff
@@ -559,6 +559,10 @@ impl RecvdPackets {
         }
     }
 
+    /// Length of the worst possible ACK frame, assuming only one range and ECN counts.
+    /// Note that this assumes one byte for the type and count of extra ranges.
+    pub const MAX_ACK_LEN: usize = 1 + 8 + 8 + 1 + 8 + 3 * 8;
```
martinthomson, line 564: I can't think of a better name, but this isn't quite right.  What you have here is the longest length that an ACK can be if it contains a single range (and ECN counts).  A useful construct, for sure, because it is the safe minimum we'd need to be able to send a useful ACK.  So maybe:

```suggestion
    pub const USEFUL_ACK_LEN: usize = 1 + 8 + 8 + 1 + 8 + 3 * 8;
```

neqo-transport/src/tracking.rs
```diff
@@ -1158,7 +1159,7 @@ mod tests {
             .is_some());
 
         let mut builder = PacketBuilder::short(Encoder::new(), false, []);
-        builder.set_limit(32);
+        builder.set_limit(RecvdPackets::MAX_ACK_LEN + 8);
```
martinthomson, line 1162:
```suggestion
        // The code pessimistically assumes that each range needs 16 bytes to express.
        // So this won't be enough for a second range.
        builder.set_limit(RecvdPackets::MAX_ACK_LEN + 8);
```

martinthomson (approved):

neqo-transport/src/connection/tests/close.rs
larseggert: The reason phrase is limited to 256 in https://github.com/mozilla/neqo/blob/52675c358c29a3a1c434050b93719cf88c86c61a/neqo-transport/src/connection/state.rs#L176-L183, which leaves enough space for the ACK.

martinthomson: Ah.  I forgot about that.  That means that we're not really testing the case where the close squeezes the ACK out.  Should we?  (Looking at the code, I don't think that we can.  I don't think we ever have more than two spaces to send a close in, and at 256 bytes, I doubt we'd hit the datagram size limit.)

larseggert: I don't think so either.

PR #1856, opened 2024-04-25
ci: Prepare to show failure differences of QNS run relative to main

larseggert:


larseggert:
Need to commit this, so the QNS run for `main` branch generates data for diffing. Will need a follow-on PR to actually generate and show the diff afterwards.

.github/workflows/qns.yml
```diff
@@ -150,20 +148,30 @@ jobs:
...
+          restore-keys: qns-${{ runner.os }}-
+
       - uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
       - uses: actions/download-artifact@9c19ed7fe5d278cd354c7dfd5d3b88589c7e2395 # v4.1.6
         with:
           pattern: '*results'
           path: results
 
       - run: |
+          ls -l results-main || true
```
martinthomson, line 172: A suggestion only: this is now long and complex enough that it might benefit from being in a file of its own.

.github/workflows/qns.yml
```diff
@@ -193,22 +201,57 @@ jobs:
...
+              else
+                echo "None :question:"
+              fi
             done
+            echo
             echo "</details>"
           } >> comment.md
 
+      - name: Remember main-branch push URL
+        # if: github.ref == 'refs/heads/main'
```
martinthomson, line 243: Presumably you plan to uncomment this.

.github/workflows/qns.yml
```diff
@@ -193,22 +201,57 @@ jobs:
...
+        run: echo "${{ github.sha }}" > baseline-sha.txt
+
+      - name: Cache main-branch results
+        # if: github.ref == 'refs/heads/main'
+        uses: actions/cache/save@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+        with:
+          path: |
+            result.json
+            baseline-sha.txt
+          key: qns-${{ runner.os }}-${{ github.sha }}
```
martinthomson, line 253: Is this right?  Anything that reads the cache will ignore the sha.  I guess that you need unique keys and this is how you achieve that...

martinthomson (commented):
The jq stuff makes me cringe, so I'm not reviewing that.

martinthomson (approved):

.github/workflows/qns.yml
larseggert: Let's push this refactoring to the follow-on PR that actually generates the diff?

.github/workflows/qns.yml
larseggert: Yeah, the download step looks like
```
      - name: Download cached main-branch results
        uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
        with:
          path: results-main
          key: qns-${{ runner.os }}-${{ github.sha }}
          restore-keys: qns-${{ runner.os }}-
```
and will never match the `key` and always fall back to the `restore-key`.

PR #1862, opened 2024-04-30
feat: Make `reason_phrase` a String

larseggert:
To make the debug logs a bit easier to parse.

### DIFF
--- a/neqo-transport/src/frame.rs
+++ b/neqo-transport/src/frame.rs
@@ -184,7 +184,7 @@ pub enum Frame<'a> {
         frame_type: u64,
         // Not a reference as we use this to hold the value.
         // This is not used in optimized builds anyway.
-        reason_phrase: Vec<u8>,
+        reason_phrase: String,
     },
     HandshakeDone,
     AckFrequency {
@@ -618,7 +618,7 @@ impl<'a> Frame<'a> {
                 Ok(Self::ConnectionClose {
                     error_code,
                     frame_type,
-                    reason_phrase,
+                    reason_phrase: String::from_utf8_lossy(&reason_phrase).to_string(),
                 })
             }
             FRAME_TYPE_HANDSHAKE_DONE => Ok(Self::HandshakeDone),
@@ -925,7 +925,7 @@ mod tests {
         let f = Frame::ConnectionClose {
             error_code: CloseError::Transport(0x5678),
             frame_type: 0x1234,
-            reason_phrase: vec![0x01, 0x02, 0x03],
+            reason_phrase: String::from_utf8("\u{1}\u{2}\u{3}"),
         };
 
         just_dec(&f, "1c80005678523403010203");
@@ -936,7 +936,7 @@ mod tests {
         let f = Frame::ConnectionClose {
             error_code: CloseError::Application(0x5678),
             frame_type: 0,
-            reason_phrase: vec![0x01, 0x02, 0x03],
+            reason_phrase: String::from_utf8("\u{1}\u{2}\u{3}"),
         };
 
         just_dec(&f, "1d8000567803010203");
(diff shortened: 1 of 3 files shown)


KershawChang (approved):

neqo-transport/src/frame.rs
```diff
@@ -618,7 +618,7 @@ impl<'a> Frame<'a> {
                 Ok(Self::ConnectionClose {
                     error_code,
                     frame_type,
-                    reason_phrase,
+                    reason_phrase: String::from_utf8_lossy(&reason_phrase).to_string(),
```
mxinden, line 621: As far as I can tell, no need for the first `to_vec()` allocation:

``` diff
modified   neqo-transport/src/frame.rs
@@ -614,11 +614,11 @@ impl<'a> Frame<'a> {
                     0
                 };
                 // We can tolerate this copy for now.
-                let reason_phrase = d(dec.decode_vvec())?.to_vec();
+                let reason_phrase = d(dec.decode_vvec())?;
                 Ok(Self::ConnectionClose {
                     error_code,
                     frame_type,
-                    reason_phrase,
+                    reason_phrase: String::from_utf8_lossy(reason_phrase).to_string(),
                 })
             }
             FRAME_TYPE_HANDSHAKE_DONE => Ok(Self::HandshakeDone),
```

neqo-transport/src/qlog.rs
```diff
@@ -545,7 +545,7 @@ impl From<&Frame<'_>> for QuicFrame {
                 },
                 error_code: Some(error_code.code()),
                 error_code_value: Some(0),
-                reason: Some(String::from_utf8_lossy(reason_phrase).to_string()),
+                reason: Some(reason_phrase.to_string()),
```
mxinden, line 548:
```suggestion
                reason: Some(reason_phrase),
```

Isn't `reason_phrase` a string now?

neqo-transport/src/frame.rs
mxinden: Also you might prefer this shorter version with a smaller diff:

``` diff
modified   neqo-transport/src/frame.rs
@@ -614,11 +614,11 @@ impl<'a> Frame<'a> {
                     0
                 };
                 // We can tolerate this copy for now.
-                let reason_phrase = d(dec.decode_vvec())?.to_vec();
+                let reason_phrase = String::from_utf8_lossy(d(dec.decode_vvec())?).to_string();
                 Ok(Self::ConnectionClose {
                     error_code,
                     frame_type,
-                    reason_phrase: String::from_utf8_lossy(&reason_phrase).to_string(),
+                    reason_phrase,
                 })
             }
             FRAME_TYPE_HANDSHAKE_DONE => Ok(Self::HandshakeDone),
```

neqo-transport/src/frame.rs
```diff
@@ -925,7 +925,7 @@ mod tests {
         let f = Frame::ConnectionClose {
             error_code: CloseError::Transport(0x5678),
             frame_type: 0x1234,
-            reason_phrase: vec![0x01, 0x02, 0x03],
+            reason_phrase: String::from_utf8(vec![0x01, 0x02, 0x03]).unwrap(),
```
martinthomson, line 928:
```suggestion
            reason_phrase: String::from("\x01\x02\x03"),
```

I think that works.

neqo-transport/src/frame.rs
```diff
@@ -936,7 +936,7 @@ mod tests {
         let f = Frame::ConnectionClose {
             error_code: CloseError::Application(0x5678),
             frame_type: 0,
-            reason_phrase: vec![0x01, 0x02, 0x03],
+            reason_phrase: String::from_utf8(vec![0x01, 0x02, 0x03]).unwrap(),
```
martinthomson, line 939: Same as above.

martinthomson (approved):
WFM.  I think we erred on the side of making the reading path non-lossy, but I don't see much point in that.  We'd need a better way to access the information to make good on that anyway.

neqo-transport/src/qlog.rs
larseggert: Making this change required some surgery to `impl From<Frame<'_>> for QuicFrame`, please take a look.

martinthomson (approved):
The change to the From implementation looks fine.

Issue #1864, opened 2024-04-30
Client spews "Setting timeout of" messages


larseggert:
Running the `test/test.sh` script generates a ton of messages like
```
0s356ms DEBUG [Client e4cf3ef0326dc9b16338] pn=9 type=Short pri-path:1bf8e83784f7e3129ce7 0.0.0.0:53440->127.0.0.1:4433 IpTos(Cs0, NotEct)
  TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: [107, 116, 104, 120, 98, 121, 101, 33] }
0s357ms DEBUG [LossRecovery] packet ap-9 sent
0s357ms DEBUG Setting timeout of 141.61858ms
0s357ms DEBUG Setting timeout of 141.612121ms
0s357ms DEBUG Setting timeout of 141.607246ms
0s357ms DEBUG Setting timeout of 141.602788ms
0s357ms DEBUG Setting timeout of 141.598496ms
0s357ms DEBUG Setting timeout of 141.594246ms
0s357ms DEBUG Setting timeout of 141.590163ms
0s357ms DEBUG Setting timeout of 141.58608ms
0s357ms DEBUG Setting timeout of 141.581955ms
0s357ms DEBUG Setting timeout of 141.577871ms
```
after the client closes the connection.

Anyone got any idea where those come from? Maybe @mxinden?
[client.log](https://github.com/mozilla/neqo/files/15166658/client.log)

mxinden:
Both the client and the server log when setting a timeout (aka. callback):

https://github.com/mozilla/neqo/blob/dcc88e323edfb5d2c4ad483bca0a93b89c5f22c2/neqo-bin/src/client/mod.rs#L418-L422

https://github.com/mozilla/neqo/blob/dcc88e323edfb5d2c4ad483bca0a93b89c5f22c2/neqo-bin/src/server/mod.rs#L516-L520

I am not sure why the client continuously sets the same timeout. I would expect the timeout to be continuously increasing until eventually the connection attempt times out.

I can take a deeper look in the next couple of days.

martinthomson:
This is probably the PTO time for the request.  We should just bump that message down to TRACE.  If it is as commonplace as it is, then we don't need to see it in logs.

larseggert:
But why would the client busy-wait like this? There are thousands of these in the log.

martinthomson:
[...]

PR #1866, opened 2024-05-02
fix(bin/client): don't close closing connection

mxinden:
The `bin/src/client/mod.rs` `Runner::run` function continuously checks whether there is more work. In case there is none, it initiates closing of the connection (`self.client.close`) and then `continue`s to the top of the loop in order to send out a closing frame.

https://github.com/mozilla/neqo/blob/14cafbaa7fa88434def2c1d19e932c08e00173f8/neqo-bin/src/client/mod.rs#L376-L409

There is a potential busy loop when closing an already closing connection. `Runner::run` will call `self.client.close` and then continously `continue` to the top of the loop.

This commit differentiates a connection state in `NotClosing`, `Closing` and `Closed`. It only attempts to close a `NotClosing` connection and only then `continue`s to the top of the loop.

---

Fixes https://github.com/mozilla/neqo/issues/1864.

### DIFF
--- a/neqo-bin/src/client/http3.rs
+++ b/neqo-bin/src/client/http3.rs
@@ -105,23 +105,27 @@ pub(crate) fn create_client(
     Ok(client)
 }
 
-impl super::Client for Http3Client {
-    fn is_closed(&self) -> Result<CloseState, ConnectionError> {
-        let (state, error) = match self.state() {
+impl TryFrom<Http3State> for CloseState {
+    type Error = ConnectionError;
+
+    fn try_from(value: Http3State) -> Result<Self, Self::Error> {
+        let (state, error) = match value {
             Http3State::Closing(error) => (CloseState::Closing, error),
             Http3State::Closed(error) => (CloseState::Closed, error),
             _ => return Ok(CloseState::NotClosing),
         };
 
-        if matches!(
-            error,
-            ConnectionError::Transport(neqo_transport::Error::NoError)
-                | ConnectionError::Application(0),
-        ) {
-            return Ok(state);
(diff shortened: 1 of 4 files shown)


larseggert:
Any chance we can refactor this to avoid the duplication between h09 and h3?

neqo-bin/src/client/http3.rs
```diff
@@ -106,15 +106,22 @@ pub(crate) fn create_client(
 }
 
 impl super::Client for Http3Client {
-    fn is_closed(&self) -> Result<bool, ConnectionError> {
-        match self.state() {
-            Http3State::Closed(
-                ConnectionError::Transport(neqo_transport::Error::NoError)
+    fn is_closed(&self) -> Result<CloseState, ConnectionError> {
```
martinthomson, line 109: It's a shame that this is almost completely duplicated code.

martinthomson (approved):

mxinden:
@larseggert @martinthomson the latest commit does the following:

- Introduce `ConnectionError::is_error`. (Long term I think it would be great for `ConnectionError` not to contain `ApplicationError(0)` and `TransportError::NoError`, but I think that is a matter for another pull request.)
- Leverage `TryFrom`, making the conversion a bit more idiomatic.

I think it is an improvement. I don't think it is great.

Let me know what you think.

martinthomson:
@mxinden 

> Long term I think it would be great for ConnectionError not to contain ApplicationError(0) and TransportError::NoError, but I think that is a matter for another pull request.

That might be because `ConnectionError` is not a good name for the type.  A better name might be `CloseReason`.

Issue #733, opened 2020-06-16
Better algorithm for stream flow control


agrover:
What we have now is the thing we did that was fast to implement, rather than a fully thought-out implementation. We should improve our algorithm, when possible.

from https://github.com/mozilla/neqo/pull/726#discussion_r440584070:
> We have an estimate of [BDP](https://en.wikipedia.org/wiki/Bandwidth-delay_product) we might use to tune this. We should aim for BDP*2 at a minimum if we have an algorithm like this, otherwise we will need more frequent sending of updates.

mxinden:
#### Status quo neqo

Static connection flow-control limit:

https://github.com/mozilla/neqo/blob/4fc4d16744ede5b099dd5bb4bfd5922d07deb046/neqo-transport/src/connection/params.rs#L21

Static stream receive window limit:

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-transport/src/recv_stream.rs#L33

Static stream send window limit:

https://github.com/mozilla/neqo/blob/c004359a817ffdea33394e94944d2f882e7e78af/neqo-transport/src/send_stream.rs#L36

#### Other implementations

- Chromium:
    - limits: https://github.com/chromium/chromium/blob/904ed34564cd5482d88cfc96abf7187a1455fa16/net/quic/quic_context.cc#L20-L22
- Google Quiche:
    - limits: https://github.com/google/quiche/blob/792c6e7d8658d7b7b27f780efec71c67930b4c9f/quiche/quic/core/quic_constants.h#L82-L86
- quic-go:
    - auto-tuning: https://github.com/quic-go/quic-go/blob/e48e1d465dffafc1cfd55dc6a4e4642087dcd5eb/internal/flowcontrol/base_flow_controller.go#L95-L118
    - limits: https://github.com/quic-go/quic-go/blob/b096e94092e97b998e396c5c67735b97e403edd3/interface.go#L291-L294

#### Potential solution

Implement auto-tuning based on Google's QUIC design document:

> Algorithm
> - As above, the flow control window update triggered when:
>   available window < max window size / 2, 
>   where available window = max receive window offset - bytes consumed
> - to realize auto-tuning, add the following logic just before issuing a window update
>   - keep track of time interval between subsequent window updates, call this since_last_update.
>   - if ( since_last_update < RTT * trigger factor ) then max window size = MIN(max window size * increase factor, upper bound).
>     - trigger factor is 2
[...]

PR #1868, opened 2024-05-02
perf(transport): auto-tune stream receive window

mxinden:
Previously the stream send and receive window had a hard limit at 1MB. On high latency and/or high bandwidth connections (i.e. large bandwidth-delay product), 1 MB is not enough to exhaust the available bandwidth.

Sample scenario:

```
delay_s = 0.05
window_bits = 1 * 1024 * 1024 * 8
bandwidth_bits_s = window_bits / delay_s
bandwidth_mbits_s = bandwidth_bits_s / 1024 / 1024 # 160.0
```

In other words, on a 50 ms connection a 1 MB window can at most achieve 160 Mbit/s.

This commit introduces an auto-tuning algorithm for the stream receive window, increasing the window towards the bandwidth-delay product of the connection.

The maximum stream send and receive window is set to 10 MB.

---

Fixes https://github.com/mozilla/neqo/issues/733.


neqo-transport/src/send_stream.rs
```diff
@@ -34,7 +34,7 @@ use crate::{
     AppError, Error, Res,
 };
 
-pub const SEND_BUFFER_SIZE: usize = 0x10_0000; // 1 MiB
+const MAX_SEND_BUFFER_SIZE: usize = 10 * 1024 * 1024;
```
mxinden, line 37: Previously Neqo would buffer at most 1 MB of send data. Now Neqo buffers up to 10 MB. In other words, it supports an up to 10 MB large send window, depending on the receive window updates of the receiver.

Thus, while this pull request is focused on increasing receive (download) throughput, this patch might as well have an impact on send (upload) throughput on high bandwidth-delay product connections.

neqo-transport/src/fc.rs
```diff
@@ -350,10 +372,42 @@ impl ReceiverFlowControl<StreamId> {
...
+                // Note that the flow control window should grow at least as
+                // fast as the congestion control window, in order to not
+                // unnecessarily limit throughput.
+                self.max_active = min(2 * self.max_active, MAX_RECV_WINDOW_SIZE);
+                qdebug!(
+                    "Increasing max stream receive window: previous max_active: {} MiB new max_active: {} MiB last update: {:?} rtt: {rtt:?} stream_id: {}",
+                    prev_max_active / 1024 / 1024, self.max_active / 1024 / 1024,  now-self.max_allowed_sent_at.unwrap(), self.subject,
+                );
+            }
+        }
```
mxinden, line 409: Note that this is not the exact algorithm suggested by @martinthomson in https://github.com/mozilla/neqo/issues/733#issuecomment-2111492717.

The algorithm proposed in this pull request adopts Martin's trigger mechanism, namely to increase the window based on the perceived BDP.

> Therefore, I suggest that if the rate at which self.retired increases (that is, the change in that value, divided by the time elapsed) exceeds some function of self.max_active / path.rtt,

It does not adopt the increase mechanism, i.e. to increase by the amount of retired data. Instead, the window is simply doubled.

> then we can increase self.max_active by the amount that self.retired has increased.



The rational is documented above. 

``` rust
                // Try doubling the flow control window.
                //
                // Note that the flow control window should grow at least as
                // fast as the congestion control window, in order to not
                // unnecessarily limit throughput.
```

mxinden:
@martinthomson can you give this pull request a review?

Not urgent, but I believe important, as this can have a significant impact on our up- and download throughput on high-bandwidth-delay connections.

neqo-transport/src/fc.rs
```diff
@@ -347,10 +370,42 @@ impl ReceiverFlowControl<StreamId> {
...
         tokens: &mut Vec<RecoveryToken>,
         stats: &mut FrameStats,
+        now: Instant,
+        rtt: Duration,
     ) {
         if !self.frame_needed() {
             return;
         }
+
+        // Auto-tune max_active, i.e. the flow control window.
```
martinthomson, line 376: This should be in a new function.  Then you can document it more cleanly and use early returns to make it neater.

neqo-transport/src/fc.rs
```diff
@@ -363,13 +418,14 @@ impl ReceiverFlowControl<StreamId> {
                 max_data: max_allowed,
             }));
             self.frame_sent(max_allowed);
+            self.max_allowed_sent_at = Some(now);
```
martinthomson, line 417: The name of this is a bit strange.  It's just the time at which you last sent a frame.  Would `last_update` make more sense?

neqo-transport/src/fc.rs
```diff
@@ -255,6 +273,11 @@ where
         }
     }
 
+    const fn should_send_flowc_update(&self) -> bool {
+        let window_bytes_unused = self.max_allowed.saturating_sub(self.retired);
```
martinthomson, line 273: Is it ever the case that `max_allowed < retired` such that you need `saturating_sub()`?

neqo-transport/src/fc.rs
martinthomson: The reason I didn't recommend doubling is that our congestion control algorithms will increase the window at a rate greater than double at times.  Then we'll lag here if the peer uses one of those.  They might be tripling or more when they are at the steep part of a Cubic ramp.  We'll be giving them less than that.

The other problem I see with this is that it doubles when the other side is close to matching the BDP.  Fluctuations in rate (or RTT estimate) will cause the condition to be tripped (retiring 1/4 of the window in ~1/4 of the RTT is expected).  Whereas retiring 1/2 of the data in 1/4 of an RTT is cause to increase much more.

On the other hand, I see how using the newly retired amount -- as I suggested -- doesn't do better than double.  Perhaps this:

```rust
let bonus_multiplier = 4;  // same as number of updates per RTT 🤔
let proportion_of_rtt = (now - last_update) / rtt;
let excess = window_bytes_used - (max_active * proportion_of_rtt);
if excess > 0 {
  self.max_active += bonus_multiplier * excess;
}
```
(Obviously, this is sloppy.  Better code would use integer math and check for overflow and whatnot.)

For instance, if they blow through the entire window in 1/4 of an RTT, then this would add 4*3/4 of the the window, quadrupling it.  There is no upper limit to how fast we can increase here if the peer sends faster (successfully).  As our estimate becomes closer to approximately correct for the BDP, we'd correct less.  If we're short by 10 bytes, then it would add just 40.  As the RTT fluctuates, I expect that this would happen a couple of times before it settles just above the BDP.

For my next trick, I will also suggest that the per-stream limit look at the connection-level estimate (which should track BDP better) so that we start each stream at a decent fraction of that (which could be 1, but I'd suggest 1/4).

neqo-transport/src/fc.rs
```diff
@@ -255,6 +273,11 @@ where
         }
     }
 
+    const fn should_send_flowc_update(&self) -> bool {
```
martinthomson, line 272:
```suggestion
    const fn should_send_update(&self) -> bool {
```

The files is `fc.rs` and the type is `ReceiverFlowControl`

neqo-transport/src/send_stream.rs
```diff
@@ -494,10 +494,10 @@ impl TxBuffer {
 
     /// Attempt to add some or all of the passed-in buffer to the `TxBuffer`.
     pub fn send(&mut self, buf: &[u8]) -> usize {
-        let can_buffer = min(SEND_BUFFER_SIZE - self.buffered(), buf.len());
+        let can_buffer = min(MAX_SEND_BUFFER_SIZE - self.buffered(), buf.len());
```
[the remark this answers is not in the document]
martinthomson: I'm OK with that.  As you say, it either gets discarded when the short-lived stream goes away, or it is available for the next write on a long-lived stream.

Was there a particular reason you increased this here?  Was it necessary to exercise the increasing receive buffer size.

neqo-transport/src/connection/params.rs
```diff
@@ -25,7 +25,7 @@ const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
 pub const ACK_RATIO_SCALE: u8 = 10;
 /// By default, aim to have the peer acknowledge 4 times per round trip time.
 /// See `ConnectionParameters.ack_ratio` for more.
-const DEFAULT_ACK_RATIO: u8 = 4 * ACK_RATIO_SCALE;
+pub const DEFAULT_ACK_RATIO: u8 = 4 * ACK_RATIO_SCALE;
```
larseggert, line 28: Unrelated?

martinthomson (commented):
Preliminary comments, with updated thoughts on the algorithm.

neqo-transport/src/fc.rs
```diff
@@ -343,10 +366,42 @@ impl ReceiverFlowControl<StreamId> {
...
+        // max_active / rtt ), try to increase the maximum flow control window (
+        // max_active ).
+        if let Some(max_allowed_sent_at) = self.max_allowed_sent_at {
+            let elapsed = now.duration_since(max_allowed_sent_at);
+            let window_bytes_used = self.max_active - (self.max_allowed - self.retired);
+
+            // Same as `elapsed / rtt < window_bytes_used / max_active`
+            // without floating point division.
+            if elapsed.as_micros() * u128::from(self.max_active)
+                < rtt.as_micros() * u128::from(window_bytes_used)
```
larseggert, line 389: Is there a way to avoid `u128` math? It can be slow on some low-end ARM chips, and if we're executing this in the fast path we might want to avoid.

neqo-transport/benches/transfer.rs
```diff
@@ -41,7 +41,7 @@ fn benchmark_transfer(c: &mut Criterion, label: &str, seed: Option<impl AsRef<st
                             boxed![SendData::new(TRANSFER_AMOUNT)]
                         ),
                         TailDrop::dsl_uplink(),
-                        Delay::new(ZERO..JITTER),
+                        RandomDelay::new(ZERO..JITTER),
```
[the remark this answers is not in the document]
mxinden: Previously `Delay` would introduce a **random** delay to each packet.

In my benchmark, I don't need randomness, thus I added a non-random delay step.

Instead of having a `Delay` and a `NonRandomDelay`, I renamed the existing `Delay` to `RandomDelay`, and introduced a new non-random `Delay`.

Thus the above change is actually just a rename. Both before and now was this step introducing a random delay.

Sorry for the confusion. Does the above make this more clear?

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
mxinden: > Was there a particular reason you increased this here?

With "this" you are referring to `MAX_SEND_BUFFER_SIZE` (i.e. previously `SEND_BUFFER_SIZE`)?

> Was it necessary to exercise the increasing receive buffer size.

Yes. Say a Neqo receiver announces a 10 MB receive window, a Neqo sender with a maximum 1 MB send window can never make use of the 10 MB receive window.

Does that answer the question @martinthomson?

neqo-transport/src/connection/params.rs
mxinden: The `WINDOW_UPDATE_FRACTION` constant is chosen based on `DEFAULT_ACK_RATIO`:

```
// fc.rs

/// Fraction of a flow control window after which a receiver sends a window
/// update.
///
/// In steady-state and max utilization, a value of 4 leads to 4 window updates
/// per RTT.
///
/// Value aligns with [`crate::connection::params::DEFAULT_ACK_RATIO`].
pub const WINDOW_UPDATE_FRACTION: u64 = 4;
```

In order to insert a Rust doc link (checked at compile time) to `DEFAULT_ACK_RATIO` above, the constant needs to be `pub`. Say that we ever rename or remove the constant, we will get a compilation error, reminding us to update `WINDOW_UPDATE_FRACTION` documentation, and potentially its value.

I don't feel strongly about this. I think it makes maintaining this dependency a bit easier. Happy to revert it in case you feel differently @larseggert.

martinthomson: This seems OK to me.

neqo-transport/src/fc.rs
martinthomson: Given that this happens at most 4 times per RTT, I'm OK with this being a tiny bit expensive.  If this were on a hot path, that would be different.

neqo-transport/benches/transfer.rs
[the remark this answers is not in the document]
larseggert: Ah, got it.

neqo-transport/src/connection/params.rs
larseggert: Ah, I missed that it had to be public in order for the doc comment.

neqo-transport/src/fc.rs
mxinden: Extracted in f351b880382616cfbf028d9f87e69e777f042522.

neqo-transport/src/fc.rs
mxinden: Renamed in 397fdd34b83b20390d7a1fd061d70d0af005b998.

neqo-transport/src/fc.rs
mxinden: Fixed in 90a7a2039983979f81d37b84befe376add19d01c. Sorry for the workaround instead of a fix in the first place.

neqo-transport/src/fc.rs
mxinden: Renamed in 200aaefe63bf5583585be85a0240880df0d3c5ad.

neqo-transport/src/fc.rs
mxinden: With f351b880382616cfbf028d9f87e69e777f042522 algorithm uses `u64` instead of `u128`.

mxinden:
I implemented @martinthomson's algorithm v2 described in https://github.com/mozilla/neqo/pull/1868#discussion_r1948875891. Basing the increase on a multiple of the excess bytes used within an RTT is clever! My experiments thus far show fast growth at first, while still accurately approaching the bandwidth-delay product thereafter. This is much better than my previous implementation, which would oftentimes overshoot the bandwidth-delay product by 2x. The unit tests can now `assert!` the auto-tuning algorithm to hit the bandwidth-delay product down to a KiB and better. 

Thank you Martin!

> For my next trick, I will also suggest that the per-stream limit look at the connection-level estimate (which should track BDP better) so that we start each stream at a decent fraction of that (which could be 1, but I'd suggest 1/4).

https://github.com/mozilla/neqo/pull/1868#discussion_r1948875891

With "connection-level estimate" are you referring to the connection receive window or the congestion controller's `cwnd`? 

For the former (connection receive window), note that we currently don't do connection level flow control. Instead, we grant the remote an excessively large connection flow control limit. Given that we do stream level flow control, this is not problematic in itself.

https://github.com/mozilla/neqo/blob/fab0046d2556eb6603c2354552a025d8ee0b782f/neqo-transport/src/connection/params.rs#L30

Setting the maximum stream receive window locally is not enough. We need to send a `FrameType::MaxStreamData` to the remote, as otherwise the remote assumes the window size based on the `InitialMaxStreamData` transport parameter for new streams.

Say a remote opens a new stream on a connection with a BDP way larger than what the `InitialMaxStreamData` transport parameter allows using with a single stream. At first, the remote can at most use `InitialMaxStreamData`. Once the local node receives the new stream, the local node can either send a `MaxStreamData` frame based on the congestion controller's `cwnd`, or send a `MaxStreamData` frame based on the here introduced auto-tuning algorithm. Either way, the remote can only make use of an increased window after 1 RTT, i.e. once it received the `MaxStreamData` frame.

Given the 1 RTT delay, and given that your new auto-tuning algorithm is quite aggressive, especially at first, is your "next trick" needed here? If you do think it is needed, do you mind if we tackle it in a consecutive pull request?

---

@martinthomson could you give this pull request another review?

neqo-transport/src/fc.rs
```diff
@@ -904,4 +1017,215 @@ mod test {
...
+        for _ in 1..11 {
+            let consumed = fc.set_consumed(fc.next_limit())?;
+            fc.add_retired(consumed);
+            write_frames(&mut fc, rtt, now);
+        }
+        let increased_max_active = fc.max_active();
+
+        assert!(
+            initial_max_active < increased_max_active,
+            "expect receive window auto-tuning to increase max_active on 0 latency connection"
```
martinthomson, line 1072: Is this right?  I guess that the expectation is that the point of the test is that time has not advanced, therefore the RTT doesn't matter.  A comment about that seems prudent, because this is NOT a 0 latency connection, it's just that the elapsed time is zero.

neqo-transport/src/fc.rs
```diff
@@ -904,4 +1017,215 @@ mod test {
...
+        /// Allow auto-tuning algorithm to be off from actual bandwidth-delay
+        /// product by up to 1KiB.
+        const TOLERANCE: u64 = 1024;
+
+        test_fixture::fixture_init();
+
+        // Run multiple iterations with randomized bandwidth and rtt.
+        for _ in 0..1_000 {
+            // Random bandwidth between 1 Mbit/s and 1 Gbit/s.
+            let bandwidth = u64::from(u16::from_be_bytes(random::<2>()) % 1_000 + 1) * 1024 * 1024;
```
martinthomson, line 1139:
```suggestion
            let bandwidth = u64::from(u16::from_be_bytes(random::<2>()) % 1_000 + 1) * 1_000 * 1_000;
```

I'm curious as to why 1_000 is needed but 1024 is accepted.

neqo-transport/src/recv_stream.rs
```diff
@@ -120,9 +132,10 @@ impl RecvStreams {
 /// from incoming STREAM frames.
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
+    // TODO: Consider shrinking from time to time?
```
martinthomson, line 135: Maybe open an issue to track rather than a comment.

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
martinthomson: The only question really is why you put this change in this changeset rather than land it on its own.  This is fine.

test-fixture/src/sim/mtu.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn process(&mut self, d: Option<Datagram>, _now: Instant) -> Output {
+        d.filter(|dgram| {
+            let header = match dgram.destination().ip() {
+                IpAddr::V4(_) => 20 + 8,
+                IpAddr::V6(_) => 40 + 8,
+            };
+
+            header + dgram.len() <= self.mtu
+        })
+        .map_or(Output::None, Output::Datagram)
```
martinthomson, line 42: I wonder if we should instead implement `From<Option<Datagram>>` for `Output`...

martinthomson (approved):
Seems like a nice outcome. Happy that it worked out.  As for the follow-up idea, having connection-level flow control would be nice because it allows us to estimate actual BDP *usage*, not availability with overheads.  If we had that, then it might be easier to produce estimates for things like WebTransport throughput availability estimates.

neqo-transport/src/recv_stream.rs
mxinden: With the above discussion (https://github.com/mozilla/neqo/pull/1868#discussion_r1900086913) in mind, I removed the TODO. In case it ever shows up in memory profiles, we can reconsider periodic shrinking.

test-fixture/src/sim/mtu.rs
mxinden: Good idea. Done in https://github.com/mozilla/neqo/pull/1868/commits/54d9d7bc51b3994ad1867770bedb325f4e5854da.

neqo-transport/src/send_stream.rs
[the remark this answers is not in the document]
mxinden: You are right, I could have landed this change on its own. Combining the two is pure convenience. Landing both in one pull request allows me to test one with the other and vice versa.

> This is fine.

Will land as is then. Thanks. Will keep it in mind for future patches.

neqo-transport/src/fc.rs
mxinden: Good catch. Addressed in https://github.com/mozilla/neqo/pull/1868/commits/bbf09a561ef93be0c25ba0bc81008e1c92a86394.

neqo-transport/src/fc.rs
mxinden: For the record, conversation continued in https://github.com/mozilla/neqo/pull/1868#issuecomment-2764623663.

mxinden:
Thank you for the quick follow-up review Martin!

@larseggert any additional comments? Happy to wait some more before merging here.

mxinden:
> As for the follow-up idea, having connection-level flow control would be nice because it allows us to estimate actual BDP _usage_, not availability with overheads. If we had that, then it might be easier to produce estimates for things like WebTransport throughput availability estimates.

For the record, this is tracked in https://github.com/mozilla/neqo/issues/2544.

neqo-transport/tests/network.rs
```diff
@@ -68,7 +68,7 @@ simulate!(
                 Error::IdleTimeout
             )))]
         ),
-        Delay::new(weeks(6)..weeks(6)),
+        RandomDelay::new(weeks(6)..weeks(6)),
```
larseggert, line 71: Unrelated change?

test-fixture/src/lib.rs
```diff
@@ -167,7 +166,23 @@ pub fn new_client(params: ConnectionParameters) -> Connection {
         now(),
     )
     .expect("create a client");
-    client.set_qlog(log);
+
+    if let Ok(dir) = std::env::var("QLOGDIR") {
```
larseggert, line 170: Should document this `env` variable somewhere.

larseggert (approved):
LGTM. Let's land this now, can always tweak as needed later.

neqo-transport/benches/transfer.rs
```diff
@@ -12,7 +12,7 @@ use test_fixture::{
     boxed,
     sim::{
         connection::{ConnectionNode, ReachState, ReceiveData, SendData},
-        network::{Delay, TailDrop},
+        network::{RandomDelay, TailDrop},
```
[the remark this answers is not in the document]
mxinden: `Delay` is renamed to `RandomDelay`. See also https://github.com/mozilla/neqo/pull/1868#discussion_r1949635270.

neqo-transport/src/recv_stream.rs
```diff
@@ -1799,47 +1824,103 @@ mod tests {
     }
 
     /// Test that the flow controls will send updates.
+    #[expect(clippy::too_many_lines, reason = "This is test code.")]
+    #[expect(clippy::cast_possible_truncation, reason = "This is test code.")]
```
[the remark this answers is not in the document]
mxinden: Neat. Thank you!

neqo-transport/benches/transfer.rs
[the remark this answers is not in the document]
larseggert: Ah. Missed that.

test-fixture/src/lib.rs
mxinden: :+1: see https://github.com/mozilla/neqo/pull/1868/commits/875b93ee8d6fce2a80548888832757f60254e675.

neqo-transport/tests/network.rs
mxinden: Same as [above](https://github.com/mozilla/neqo/pull/1868#discussion_r2020794843).

PR #1872, opened 2024-05-03
refactor: rename ConnectionError to CloseReason

mxinden:
The `neqo_transport::ConnectionError` enum contains the two non-error variants `Error::NoError` and `CloseReason::Application(0)`. In other words, `ConnectionError` contains variants that are not errors.

This commit renames `ConnectionError` to the more descriptive name `CloseReason`.

See suggestion in https://github.com/mozilla/neqo/pull/1866#issuecomment-2091654649.

To ease the upgrade for downstream users, this commit adds a deprecated `ConnectionError`, guiding users to rename to `CloseReason` via a deprecation warning.

``` rust
#[deprecated(note = "use `CloseReason` instead")]
pub type ConnectionError = CloseReason;
```

### DIFF
--- a/neqo-bin/src/client/http3.rs
+++ b/neqo-bin/src/client/http3.rs
@@ -22,8 +22,8 @@ use neqo_common::{event::Provider, hex, qdebug, qinfo, qwarn, Datagram, Header};
 use neqo_crypto::{AuthenticationStatus, ResumptionToken};
 use neqo_http3::{Error, Http3Client, Http3ClientEvent, Http3Parameters, Http3State, Priority};
 use neqo_transport::{
-    AppError, Connection, ConnectionError, EmptyConnectionIdGenerator, Error as TransportError,
-    Output, StreamId,
+    AppError, CloseReason, Connection, EmptyConnectionIdGenerator, Error as TransportError, Output,
+    StreamId,
 };
 use url::Url;
 
@@ -106,7 +106,7 @@ pub(crate) fn create_client(
 }
 
 impl TryFrom<Http3State> for CloseState {
-    type Error = ConnectionError;
+    type Error = CloseReason;
 
     fn try_from(value: Http3State) -> Result<Self, Self::Error> {
         let (state, error) = match value {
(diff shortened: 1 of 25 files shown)


neqo-bin/src/client/http3.rs
```diff
@@ -106,7 +106,7 @@ pub(crate) fn create_client(
 }
 
 impl TryFrom<Http3State> for CloseState {
-    type Error = ConnectionError;
+    type Error = CloseReason;
```
larseggert, line 109:
```suggestion
    type Reason = CloseReason;
```
?

neqo-transport/src/lib.rs
```diff
@@ -209,13 +209,17 @@ impl ::std::fmt::Display for Error {
 
 pub type AppError = u64;
 
+#[deprecated(note = "use `CloseReason` instead")]
+pub type ConnectionError = CloseReason;
```
larseggert, line 213:
```suggestion
pub type ConnectionCloseReason = CloseReason;
```
?

larseggert (approved):
Two suggestions for your consideration (not sure if they make sense in the context of the rest of the code), otherwise LGTM.

neqo-transport/src/lib.rs
larseggert: Ah I guess we can't rename this one.

neqo-bin/src/client/http3.rs
mxinden: `TryFrom` is defined in the standard library. So is the Associated type `Error` of the `TryFrom` trait. 

https://doc.rust-lang.org/std/convert/trait.TryFrom.html

Thus one can not change the name of the associated type.

larseggert: Ah got it.

neqo-transport/src/lib.rs
mxinden: Users that upgrade to this patch might still be using the old `ConnectionError`. In order to give them a compiler warning, hinting to them that they should use `CloseReason` instead, these two lines are added.

The type alias (i.e. `pub type ConnectionError = CloseReason;`) will make sure their code still compiles, as `ConnectionError` is now simply an alias for the new `CloseReason`. The `deprecated` attribute will give them the warning that they are using a deprecated type (i.e. `ConnectionError`) and that they should switch to `CloseReason`.

Does that make sense?

mxinden: Here is a sample compiler warning that a user would get:

```
warning: use of deprecated type alias `ConnectionError`: use `CloseReason` instead
    |
119 |     error: ConnectionError,
    |            ^^^^^^^^^^^^^^^

PR #1873, opened 2024-05-03
refactor: remove unused test-fixture/src/sim/net.rs

mxinden:
The `net.rs` file is not declared as a module in `src/sim/mod.rs`, thereby unreachable, and thus never compiled.


larseggert:
@martinthomson or @KershawChang might remember if there was a plan for this file?

martinthomson (approved):
Yeah, that's a mistake.  The code that compiles is in delay.rs

Issue #1490, opened 2023-11-07
assertion failed: earliest > now


KershawChang:
I hit this assertion when trying to land [this patch](https://phabricator.services.mozilla.com/D191325).
According to the log below, this seems to be caused by the Idle/keepalive timer. It shows that we called `process_output` twice, but the instant values returned from Idle/keepalive timer were not changed.
Note that the delay returned from the first `process_output` is more than 1ms, so necko waited and didn't call `process_output` immediately. I think this means that [this patch](https://phabricator.services.mozilla.com/D191325) is not the culprit, but I am not super sure.

```
[task 2023-11-07T17:01:59.631Z] 17:01:59     INFO - GECKO(2667) | [Parent 2667: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client b14e84fff18672cc] Get callback delay Instant { tv_sec: 4466, tv_nsec: 62214290 }
[task 2023-11-07T17:01:59.632Z] 17:01:59     INFO - GECKO(2667) | [Parent 2667: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client b14e84fff18672cc] Idle/keepalive timer Instant { tv_sec: 4467, tv_nsec: 707236065 }
[task 2023-11-07T17:01:59.632Z] 17:01:59     INFO - GECKO(2667) | [Parent 2667: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client b14e84fff18672cc] Loss recovery timer Instant { tv_sec: 4478, tv_nsec: 984261778 }
[task 2023-11-07T17:02:01.277Z] 17:02:01     INFO - GECKO(2667) | [Parent 2667: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client b14e84fff18672cc] Get callback delay Instant { tv_sec: 4467, tv_nsec: 707664199 }
[task 2023-11-07T17:02:01.278Z] 17:02:01     INFO - GECKO(2667) | [Parent 2667: Socket Thread]: I/neqo_transport::* [neqo_transport::connection] [Client b14e84fff18672cc] Idle/keepalive timer Instant { tv_sec: 4467, tv_nsec: 707236065 }
[...]

PR #1875, opened 2024-05-03
fix: Make neqo pass `amplificationlimit` QNS test

### LINKED ISSUE #1183: Disable PTO at server if path is not validated
We currently send probes on PTO always.  However a server is not allowed to when the active path hasn't been validated.  This is preventing us from passing amplification limit tests.

This is not particularly urgent as we don't ship a server currently.  But it is also a relatively small amount of work.  The only challenge might be getting the path validation state to the right place(s).  We have tests for the amplification limit, but these are flawed (they pass, they should not), so the biggest part of this work is in tweaking those tests.

larseggert:
Fixes #1183

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1468,10 +1468,6 @@ impl Connection {
             // We only allow one path during setup, so apply handshake
             // path validation to this path.
             path.borrow_mut().set_valid(now);
-        } else if self.role == Role::Client && packet.packet_type() == PacketType::Initial {
-            // If we are a client and we receive an Initial packet, we can
-            // assume that the server has validated the path.
-            path.borrow_mut().set_valid(now);
         }
     }
 
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2377,7 +2377,7 @@ impl Connection {
         }
 
         if encoder.is_empty() {
-            qdebug!("TX blocked, profile={:?} ", profile);
+            qdebug!("TX blocked, profile={:?}", profile);
             Ok(SendOption::No(profile.paced()))
         } else {
             // Perform additional padding for Initial packets as necessary.
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -338,7 +338,7 @@ impl Connection {
             NeqoQlog::default(),
             now,
         );
-        c.setup_handshake_path(&Rc::new(RefCell::new(path)), now);
+        c.setup_handshake_path(&Rc::new(RefCell::new(path)));
         Ok(c)
     }
 
@@ -1456,7 +1456,7 @@ impl Connection {
         }
 
         if self.state == State::WaitInitial {
-            self.start_handshake(path, packet, now);
+            self.start_handshake(path, packet);
         }
 
         if self.state.connected() {
@@ -1467,6 +1467,10 @@ impl Connection {
         {
             // We only allow one path during setup, so apply handshake
             // path validation to this path.
+            path.borrow_mut().set_valid(now);
+        } else if self.role == Role::Client && packet.packet_type() == PacketType::Initial {
+            // If we are a client and we receive an Initial packet, we can
+            // assume that the server has validated the path.
             path.borrow_mut().set_valid(now);
         }
     }
@@ -1650,7 +1654,7 @@ impl Connection {
     /// During connection setup, the first path needs to be setup.
     /// This uses the connection IDs that were provided during the handshake
     /// to setup that path.
-    fn setup_handshake_path(&mut self, path: &PathRef, now: Instant) {
+    fn setup_handshake_path(&mut self, path: &PathRef) {
         self.paths.make_permanent(
             path,
             Some(self.local_initial_source_cid.clone()),
@@ -1664,7 +1668,6 @@ impl Connection {
                     .clone(),
             ),
         );
-        path.borrow_mut().set_valid(now);
     }
 
     /// If the path isn't permanent, assign it a connection ID to make it so.
@@ -1703,7 +1706,7 @@ impl Connection {
             // First try to fill in handshake details.
             if packet.packet_type() == PacketType::Initial {
                 self.remote_initial_source_cid = Some(ConnectionId::from(packet.scid()));
-                self.setup_handshake_path(path, now);
+                self.setup_handshake_path(path);
             } else {
                 // Otherwise try to get a usable connection ID.
                 mem::drop(self.ensure_permanent(path));
@@ -1711,7 +1714,7 @@ impl Connection {
         }
     }
 
-    fn start_handshake(&mut self, path: &PathRef, packet: &PublicPacket, now: Instant) {
+    fn start_handshake(&mut self, path: &PathRef, packet: &PublicPacket) {
         qtrace!([self], "starting handshake");
         debug_assert_eq!(packet.packet_type(), PacketType::Initial);
         self.remote_initial_source_cid = Some(ConnectionId::from(packet.scid()));
@@ -1720,7 +1723,7 @@ impl Connection {
             self.cid_manager
                 .add_odcid(self.original_destination_cid.as_ref().unwrap().clone());
             // Make a path on which to run the handshake.
-            self.setup_handshake_path(path, now);
+            self.setup_handshake_path(path);
 
             self.zero_rtt_state = match self.crypto.enable_0rtt(self.version, self.role) {
                 Ok(true) => {
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -338,7 +338,7 @@ impl Connection {
             NeqoQlog::default(),
             now,
         );
-        c.setup_handshake_path(&Rc::new(RefCell::new(path)));
+        c.setup_handshake_path(&Rc::new(RefCell::new(path)), now);
         Ok(c)
     }
 
@@ -1456,7 +1456,7 @@ impl Connection {
         }
 
         if self.state == State::WaitInitial {
-            self.start_handshake(path, packet);
+            self.start_handshake(path, packet, now);
         }
 
         if self.state.connected() {
@@ -1654,7 +1654,7 @@ impl Connection {
     /// During connection setup, the first path needs to be setup.
     /// This uses the connection IDs that were provided during the handshake
     /// to setup that path.
-    fn setup_handshake_path(&mut self, path: &PathRef) {
+    fn setup_handshake_path(&mut self, path: &PathRef, now: Instant) {
         self.paths.make_permanent(
             path,
             Some(self.local_initial_source_cid.clone()),
@@ -1668,6 +1668,9 @@ impl Connection {
                     .clone(),
             ),
         );
+        if self.role == Role::Client {
+            path.borrow_mut().set_valid(now);
+        }
     }
 
     /// If the path isn't permanent, assign it a connection ID to make it so.
@@ -1706,7 +1709,7 @@ impl Connection {
             // First try to fill in handshake details.
             if packet.packet_type() == PacketType::Initial {
                 self.remote_initial_source_cid = Some(ConnectionId::from(packet.scid()));
-                self.setup_handshake_path(path);
+                self.setup_handshake_path(path, now);
             } else {
                 // Otherwise try to get a usable connection ID.
                 mem::drop(self.ensure_permanent(path));
@@ -1714,7 +1717,7 @@ impl Connection {
         }
     }
 
-    fn start_handshake(&mut self, path: &PathRef, packet: &PublicPacket) {
+    fn start_handshake(&mut self, path: &PathRef, packet: &PublicPacket, now: Instant) {
         qtrace!([self], "starting handshake");
         debug_assert_eq!(packet.packet_type(), PacketType::Initial);
         self.remote_initial_source_cid = Some(ConnectionId::from(packet.scid()));
@@ -1723,7 +1726,7 @@ impl Connection {
             self.cid_manager
                 .add_odcid(self.original_destination_cid.as_ref().unwrap().clone());
             // Make a path on which to run the handshake.
-            self.setup_handshake_path(path);
+            self.setup_handshake_path(path, now);
 
             self.zero_rtt_state = match self.crypto.enable_0rtt(self.version, self.role) {
                 Ok(true) => {
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1040,7 +1040,7 @@ impl Connection {
             qtrace!([self], "Idle/keepalive timer {:?}", idle_time);
             delays.push(idle_time);
 
-            if let Some(lr_time) = self.loss_recovery.next_timeout(rtt) {
+            if let Some(lr_time) = self.loss_recovery.next_timeout(&path) {
                 qtrace!([self], "Loss recovery timer {:?}", lr_time);
                 delays.push(lr_time);
             }
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -791,6 +791,7 @@ fn anti_amplification() {
     let cb = server.process_output(now).callback();
     // We are blocked by the amplification limit now.
     assert_eq!(cb, DEFAULT_RTT * 3);
+    now += cb;
 
     now += DEFAULT_RTT / 2;
     client.process_input(&s_init1, now);
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -789,9 +789,9 @@ fn anti_amplification() {
     let s_init3 = server.process_output(now).dgram().unwrap();
     assert_eq!(s_init3.len(), server.plpmtu());
     let cb = server.process_output(now).callback();
+
     // We are blocked by the amplification limit now.
-    assert_eq!(cb, DEFAULT_RTT * 3);
-    now += cb;
+    assert_eq!(cb, server.conn_params.get_idle_timeout());
 
     now += DEFAULT_RTT / 2;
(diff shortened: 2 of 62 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1462,6 +1462,10 @@ impl Connection {
             // We only allow one path during setup, so apply handshake
             // path validation to this path.
             path.borrow_mut().set_valid(now);
+        } else if self.role == Role::Client && packet.packet_type() == PacketType::Initial {
+            // If we are a client and we receive an Initial packet, we can
+            // assume that the server has validated the path.
+            path.borrow_mut().set_valid(now);
```
martinthomson, line 1468: This is obviously true, but what circumstances lead to a client not having a valid path at this point?  We are calling `setup_handshake_path` at the client as soon as the path is created, so this would have to involve some sort of migration during the handshake, right?

Put differently, does the test fail if you remove this block of code?

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -768,16 +768,11 @@ fn anti_amplification() {
...
-    let s_pacing = server.process_output(now).callback();
-    assert_ne!(s_pacing, Duration::new(0, 0));
-    now += s_pacing;
-
     let s_init3 = server.process_output(now).dgram().unwrap();
     assert_eq!(s_init3.len(), PATH_MTU_V6);
     let cb = server.process_output(now).callback();
-    assert_ne!(cb, Duration::new(0, 0));
+    // We are blocked by the amplification limit now.
+    assert_eq!(cb, Duration::from_millis(300));
```
martinthomson, line 775:
```suggestion
    assert_eq!(cb, DEFAULT_RTT * 3);
```

You want this to be relative to the RTT, I think.

neqo-transport/src/recovery.rs
```diff
@@ -968,7 +968,9 @@ impl LossRecovery {
         }
         self.stats.borrow_mut().lost += lost_packets.len();
 
-        self.maybe_fire_pto(primary_path.borrow().rtt(), now, &mut lost_packets);
+        if primary_path.borrow().is_valid() {
+            self.maybe_fire_pto(primary_path.borrow().rtt(), now, &mut lost_packets);
+        }
```
martinthomson, line 973: Perhaps the right thing to do is pass the `PathRef` into `maybe_fire_pto`, so that the "maybe" can do the extra work (and so you can borrow just once).

qns/interop.sh
```diff
@@ -26,7 +26,7 @@ server)
     -name "$CERT" -passout pass: -out "$P12CERT"
   pk12util -d "sql:$DB" -i "$P12CERT" -W ''
   certutil -L -d "sql:$DB" -n "$CERT"
-  RUST_LOG=info RUST_BACKTRACE=1 neqo-server --cc cubic --qns-test "$TESTCASE" \
+  RUST_LOG=debug RUST_BACKTRACE=1 neqo-server --cc cubic --qns-test "$TESTCASE" \
```
martinthomson, line 29: :+1: 

I've been meaning to do this for testing server interop stuff.

larseggert:
One test not passing, see https://github.com/mozilla/neqo/issues/1490#issuecomment-2095562404

larseggert:
@martinthomson @KershawChang if any of you understands the timer code, I'd appreciate some help here. The `idle_timeout_crazy_rtt` persistently crashes with `assertion failed: earliest > now`, which seems to be a long-standing issue.

It's straightforward to reproduce:
```
# RUST_LOG=trace cargo nextest run --no-capture idle_timeout_crazy_rtt
...
0s 35ms DEBUG [idle_timeout_crazy_rtt] processing Server c11688b68f20d6a8b4f512e565e90a
0s 35ms TRACE [Server c11688b68f20d6a8b4f512e565e90a] process_output Handshaking Instant { tv_sec: 5648807, tv_nsec: 122490916 }
0s 35ms TRACE IdleTimeout::expiry@Instant { tv_sec: 5648807, tv_nsec: 122490916 } pto=325ms, ka=false => Instant { tv_sec: 610448805, tv_nsec: 622490916 }
0s 35ms TRACE [LossRecovery] timeout Instant { tv_sec: 5648807, tv_nsec: 122490916 }
0s 35ms TRACE detect lost ap: now=Instant { tv_sec: 5648807, tv_nsec: 122490916 } delay=112.5ms
0s 35ms TRACE detect lost hs: now=Instant { tv_sec: 5648807, tv_nsec: 122490916 } delay=112.5ms
0s 35ms TRACE lost=11, time sent Instant { tv_sec: 5648806, tv_nsec: 822490916 } is before lost_delay 112.5ms
0s 35ms DEBUG packet_lost this=0x600002f741e0, pn=11, ps=675
0s 35ms DEBUG [NewReno CongCtrl 662/2674 ssthresh 2674] Cong event -> recovery; cwnd 2674, ssthresh 2674
0s 35ms DEBUG [NewReno CongCtrl 662/2674 ssthresh 2674] state -> RecoveryStart
0s 35ms DEBUG on_packets_lost this=0x600002f741e0, bytes_in_flight=662, cwnd=2674, state=RecoveryStart
0s 35ms TRACE detect lost in: now=Instant { tv_sec: 5648807, tv_nsec: 122490916 } delay=112.5ms
0s 35ms TRACE lost=11, time sent Instant { tv_sec: 5648806, tv_nsec: 822490916 } is before lost_delay 112.5ms
0s 35ms DEBUG packet_lost this=0x600002f741e0, pn=11, ps=662
0s 35ms DEBUG on_packets_lost this=0x600002f741e0, bytes_in_flight=0, cwnd=2674, state=RecoveryStart
0s 35ms DEBUG [Server c11688b68f20d6a8b4f512e565e90a] Lost: Crypto(CryptoRecoveryToken { space: Handshake, offset: 0, length: 628 })
0s 35ms INFO Lost crypto frame space=hs offset=0 length=628
0s 35ms DEBUG [Server c11688b68f20d6a8b4f512e565e90a] Lost: Crypto(CryptoRecoveryToken { space: Initial, offset: 0, length: 90 })
0s 35ms INFO Lost crypto frame space=in offset=0 length=90
0s 35ms TRACE [Server c11688b68f20d6a8b4f512e565e90a] output Instant { tv_sec: 5648807, tv_nsec: 122490916 }
0s 35ms DEBUG [LossRecovery] get send profile Instant { tv_sec: 5648807, tv_nsec: 122490916 }
0s 35ms DEBUG [Server c11688b68f20d6a8b4f512e565e90a] output_path send_profile SendProfile { limit: 255, pto: None, probe: (), paced: false }
0s 35ms DEBUG Building Initial dcid 0900000000c1f2c64a scid 0800000000c32312
0s 35ms DEBUG Building Handshake dcid 0900000000c1f2c64a scid 0800000000c32312
0s 35ms DEBUG Building Short dcid 0900000000c1f2c64a
0s 35ms DEBUG TX blocked, profile=SendProfile { limit: 255, pto: None, probe: (), paced: false }
0s 35ms TRACE [Server c11688b68f20d6a8b4f512e565e90a] Get callback delay Instant { tv_sec: 5648807, tv_nsec: 122490916 }
0s 35ms TRACE ack_time for ap = None
0s 35ms TRACE ack_time for hs = None
0s 35ms TRACE ack_time for in = None
0s 35ms TRACE IdleTimeout::expiry@Instant { tv_sec: 5648807, tv_nsec: 122490916 } pto=325ms, ka=false => Instant { tv_sec: 610448805, tv_nsec: 622490916 }
0s 35ms TRACE [Server c11688b68f20d6a8b4f512e565e90a] Idle/keepalive timer Instant { tv_sec: 610448805, tv_nsec: 622490916 }
0s 35ms TRACE [LossRecovery] next_timeout loss=None pto=Some(Instant { tv_sec: 5648807, tv_nsec: 122490916 })
0s 35ms TRACE [Server c11688b68f20d6a8b4f512e565e90a] Loss recovery timer Instant { tv_sec: 5648807, tv_nsec: 122490916 }
thread 'idle_timeout_crazy_rtt' panicked at neqo-transport/src/connection/mod.rs:1068:9:
assertion failed: earliest > now
```

KershawChang:
> @martinthomson @KershawChang if any of you understands the timer code, I'd appreciate some help here. The `idle_timeout_crazy_rtt` persistently crashes with `assertion failed: earliest > now`, which seems to be a long-standing issue.
> 

I think we need to implement what Matine suggested to do in this [comment](https://github.com/mozilla/neqo/pull/1491#issuecomment-1813399990).
I'll try to give it a shot.
In order to unblock this, maybe we can remove that debug assertion first?

larseggert:
> In order to unblock this, maybe we can remove that debug assertion first?

The it dies here
```
thread 'idle_timeout_crazy_rtt' panicked at test-fixture/src/sim/mod.rs:202:25:
assertion `left != right` failed
  left: 0ns
 right: 0ns
 ```
 And when I take that out, too, it spins forever.

File: neqo-transport/src/recovery/mod.rs [hunk 5]
```diff
@@ -828,7 +828,12 @@ impl LossRecovery {
     /// When it has, mark a few packets as "lost" for the purposes of having frames
     /// regenerated in subsequent packets.  The packets aren't truly lost, so
     /// we have to clone the `SentPacket` instance.
-    fn maybe_fire_pto(&mut self, rtt: &RttEstimate, now: Instant, lost: &mut Vec<SentPacket>) {
+    fn maybe_fire_pto(&mut self, primary_path: &PathRef, now: Instant, lost: &mut Vec<SentPacket>) {
+        let path = primary_path.borrow();
+        if !path.is_valid() {
+            return;
```
KershawChang, line 834: Sorry, I just realized that the root cause of this PR is not the same as #1491 .
That assertion `earliest > now` is caused by this line. If I remove the early return, the test `idle_timeout_crazy_rtt` passes.

I am still looking into why this change can trigger that assertion.

larseggert: @KershawChang any insights?

larseggert: @KershawChang ping?

larseggert: I just tried this (i.e., removing the early return) and while that satisfies the `earliest > now` assertion, the `loss_time_past_largest_acked` test now fails:
```
thread 'connection::tests::recovery::loss_time_past_largest_acked' panicked at neqo-transport/src/connection/tests/recovery.rs:587:5:
assertion `left != right` failed
  left: 0ns
 right: 0ns
stack backtrace:
   0: rust_begin_unwind
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed::<core::time::Duration, core::time::Duration>
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:364:5
   4: neqo_transport::connection::tests::recovery::loss_time_past_largest_acked
             at ./src/connection/tests/recovery.rs:587:5
   5: neqo_transport::connection::tests::recovery::loss_time_past_largest_acked::{closure#0}
             at ./src/connection/tests/recovery.rs:560:34
   6: <neqo_transport::connection::tests::recovery::loss_time_past_largest_acked::{closure#0} as core::ops::function::FnOnce<()>>::call_once
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/ops/function.rs:250:5
             ```

Comment (KershawChang) [reply neqo-transport/src/recovery/mod.rs:834]: Sorry, I haven't made much progress. I'll try to look at this again this week.

Comment (larseggert) [reply neqo-transport/src/recovery/mod.rs]: I think I have a fix for the test (similar to how I needed to fix another test in this PR).

**Comment by larseggert** [comment]
This again is waiting for a fix to the `idle_timeout_crazy_rtt` test failure.

**Comment by KershawChang** [comment]
I've tried to give this another look and I found that the problem might be about when the function `discard_packet` is called.

https://github.com/mozilla/neqo/blob/e2fc9b0392972f9c37e78ea1d6ec4dd307267a58/neqo-transport/src/path.rs#L948

Before this change, the server was able to update the RTT to the value set [here](https://github.com/mozilla/neqo/blob/e2fc9b0392972f9c37e78ea1d6ec4dd307267a58/neqo-transport/tests/network.rs#L69)
```
0s326ms INFO [pri-unv-path:090000000037faf420 [fe80::1]:443->[fe80::1]:443] discarding a packet without an RTT estimate; guessing RTT=2999654.700000001s
```

With this change, the server can only update the RTT to a small value:
```
0s202ms INFO [pri-unv-path:0800000000b3b9a6 [fe80::1]:443->[fe80::1]:443] discarding a packet without an RTT estimate; guessing RTT=300.000001ms
```

Unfortunately, I still can't figure out the correct way to fix this.

**Comment by larseggert** [comment]
> With this change, the server can only update the RTT to a small value:
> 
> ```
> 0s202ms INFO [pri-unv-path:0800000000b3b9a6 [fe80::1]:443->[fe80::1]:443] discarding a packet without an RTT estimate; guessing RTT=300.000001ms
> ```

I don't see this log line when I run the test for this PR? (I do see the one for the `main` branch.)

> Unfortunately, I still can't figure out the correct way to fix this.

What seems to happen is that `set_valid()` is never called on the server-side for the initial path, because this patch moves that to after the path is actually validated. So the server stays amplification-limited. I'm not sure how that would affect the timers though.

**Comment by larseggert** [comment]
Some QNS L1/C1 tests are failing because #1981 is needed.

File: neqo-transport/src/recovery/mod.rs [hunk 6]
```diff
@@ -835,7 +840,9 @@ impl LossRecovery {
     /// When it has, mark a few packets as "lost" for the purposes of having frames
     /// regenerated in subsequent packets.  The packets aren't truly lost, so
     /// we have to clone the `SentPacket` instance.
-    fn maybe_fire_pto(&mut self, rtt: &RttEstimate, now: Instant, lost: &mut Vec<SentPacket>) {
+    fn maybe_fire_pto(&mut self, primary_path: &PathRef, now: Instant, lost: &mut Vec<SentPacket>) {
+        let path = primary_path.borrow();
+        let rtt = path.rtt();
```
mxinden, line 845: Why is this change needed? Compiles and tests succeed without it on my local machine.

mxinden (approved):
Change makes sense to me. One minor comment.

neqo-transport/src/recovery/mod.rs
larseggert: Might be a leftover from a previous change. Let me check.

larseggert:
So this seems to work, in the sense that there are no more `A` failures when we are the server.

But maybe there are also more `L1` and `C1` failures now, and possibly more frequent deadlocks (= runs that get cancelled by the timeout)?

It's difficult to say something definitive here, since we have had these occur randomly all along, and they may also not be due to bugs on our end.

neqo-transport/src/connection/mod.rs
```diff
@@ -1664,7 +1664,9 @@ impl Connection {
                     .clone(),
             ),
         );
-        path.borrow_mut().set_valid(now);
+        if self.role == Role::Client {
```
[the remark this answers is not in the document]
martinthomson: That is because this code is called by both roles, but only the client should start with the assumption that the path is OK.

neqo-transport/src/recovery/mod.rs
```diff
@@ -735,9 +735,14 @@ impl LossRecovery {
 
     /// Calculate when the next timeout is likely to be.  This is the earlier of the loss timer
     /// and the PTO timer; either or both might be disabled, so this can return `None`.
```
martinthomson, line 737: This probably needs to be updated.

neqo-transport/src/recovery/mod.rs
```diff
@@ -735,9 +735,14 @@ impl LossRecovery {
...
-    pub fn next_timeout(&self, rtt: &RttEstimate) -> Option<Instant> {
+    pub fn next_timeout(&self, path: &Path) -> Option<Instant> {
+        let rtt = path.rtt();
         let loss_time = self.earliest_loss_time(rtt);
-        let pto_time = self.earliest_pto(rtt);
+        let pto_time = if path.amplification_limit() > path.plpmtu() {
```
martinthomson, line 741: This condition seems like it should be encapsulated better.  In a is_validated() function or something.

neqo-transport/src/recovery/mod.rs
larseggert: What should it say instead? The
> either or both might be disabled, so this can return `None`

bit seems to cover what happens, even with the change?

neqo-transport/src/recovery/mod.rs
larseggert: There is `is_valid()`, but this condition isn't quite the same as `!is_valid()`, since we can also PTO on a path that is not yet validated but where we have sent less than the amplification limit.

How about this?

```suggestion
        let pto_time = if path.is_valid() || path.amplification_limit() > path.plpmtu() {
```

Makes it a bit clearer and we don't need to run through `amplification_limit()` in the common case.

neqo-transport/src/recovery/mod.rs
martinthomson: Yeah, you are right.  Nevermind.

neqo-transport/src/recovery/mod.rs
martinthomson: That would help, but it might be better to encapsulate that check.

```rust
impl Path {
/// Determine whether we should be setting a PTO for this path.
/// This is true when either the path is valid or when there is enough
/// remaining in the amplification limit to fit a full-sized path (i.e., the path MTU).
pub fn pto_enabled(&self) -> bool {
  // See the implementation `amplification_limit` for details.
  self.amplification_limit() >= self.plpmtu()
}
```

martinthomson (approved):

PR #1882, opened 2024-05-06
remove the assertion that can be triggered when a HANDSHAKE_DONE fram…

### LINKED ISSUE #1871: Assertion failure: StateSignaling must be in Idle state but is in Closing
We already have a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1879698) to track this intermittent failure.
However, the [recent neqo v0.7.6](https://treeherder.mozilla.org/jobs?repo=try&revision=9cb753242518fe85a4cfcdd9e314918244644480&selectedTaskRun=ckVHEhh2SXK_8MkoW-gSsw.0) seems to make this failure happen more often.
I am currently investigating this.

KershawChang:
Fix #1871

### DIFF
--- a/neqo-transport/src/connection/state.rs
+++ b/neqo-transport/src/connection/state.rs
@@ -212,10 +212,6 @@ pub enum StateSignaling {
 impl StateSignaling {
     pub fn handshake_done(&mut self) {
         if !matches!(self, Self::Idle) {
-            debug_assert!(
-                false,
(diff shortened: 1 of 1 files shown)


larseggert (approved):

martinthomson (commented):
As a footnote, it would be very good to add a test that replicates the problem that caused this assertion to fire.  We understand how it happens now, so we should maybe prevent it from happening again with a test.

PR #1883, opened 2024-05-06
chore: Fix new clippy lints in cargo 1.80.0-nightly

larseggert:
This should fix the CI errors that have started to appear as of today.

### DIFF
--- a/fuzz/build.rs
+++ b/fuzz/build.rs
@@ -1,0 +1,9 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(fuzzing)");
+}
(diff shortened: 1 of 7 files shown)


neqo-http3/src/server.rs
```diff
@@ -1267,6 +1267,7 @@ mod tests {
         let out = peer_conn.process(None, now());
         hconn.process(out.as_dgram_ref(), now());
 
+        #[allow(clippy::mutable_key_type)]
```
martinthomson, line 1270: This is just a test, so this is probably OK, but this could be keyed on stream ID instead.  It would be better if it was.  I think that you can just use `stream.stream_id()` in those places that you use the stream.

fuzz/build.rs
```diff
@@ -0,0 +1,9 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+fn main() {
```
martinthomson, line 7: Do you list build.rs in Cargo.toml?  Should you?

larseggert: Hm. It doesn't look like it's needed, in the sense that it takes effect. The [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-build-field) seems to imply that `build.rs` is picked up by default, even if not specified. (We can even remove the `build` lines from `neqo-common` and `neqo-crypto` and things still build fine.)

larseggert:
Merging this now to get the CI things fixed. We can always tweak it later if there is something in here that should be.

PR #1886, opened 2024-05-07
Cleanup tracking

martinthomson:
A version of #1662 but retaining the `BTreeMap` backing store.

One thing I learned from this was that creating temporary stores was probably the biggest cost.  It might be that the `Vec` could be better with that realization, but this has the advantage of being a little bit simpler.


neqo-transport/src/recovery/sent.rs
```diff
@@ -0,0 +1,382 @@
...
+        let mut packets = std::mem::take(&mut self.packets);
+        for range in acked_ranges {
+            // For each acked range, split off the acknowledged part,
+            // then split off the part that hasn't been acknowledged.
+            // This order works better when processing ranges that
+            // have already been processed, which is common.
+            let mut acked = packets.split_off(range.start());
+            let keep = acked.split_off(&(*range.end() + 1));
+            self.packets.extend(keep);
+            result.extend(acked.into_values().rev());
```
martinthomson, line 209: Most of this file is just moving stuff, but this is a new algorithm.  The old one is https://github.com/mozilla/neqo/blob/bb88aab4525ff74c42b6cc187813c0b0e2813f80/neqo-transport/src/recovery.rs#L313-L336

neqo-transport/src/recovery/sent.rs
```diff
@@ -0,0 +1,382 @@
...
+                let keep = self.packets.split_off(&first_keep);
+                std::mem::replace(&mut self.packets, keep)
+            } else {
+                // All packets are expired.
+                std::mem::take(&mut self.packets)
+            };
+            to_remove
+                .into_values()
+                .filter(SentPacket::ack_eliciting)
+                .count()
```
martinthomson, line 240: The caller only needed a count, so I decided to return a count.  That seems to have improved benchmarks a noticeable amount too.

neqo-transport/src/recovery/sent.rs
mxinden: For anyone else wondering why there is no `BTreeMap::take_range`:

https://users.rust-lang.org/t/removing-range-of-elements-from-btreemap/51582

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -341,15 +341,15 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
             // However, subsequent packets might be app-limited. Set `first_app_limited` to the
             // next packet number.
-            self.first_app_limited = pkt.pn + 1;
+            self.first_app_limited = pkt.pn() + 1;
         }
 
-        self.bytes_in_flight += pkt.size;
-        qdebug!(
+        self.bytes_in_flight += pkt.len();
+        qinfo!(
```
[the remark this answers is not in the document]
martinthomson: Yeah, it's not really necessary to record all the packets at that log level.

neqo-transport/src/connection/mod.rs
```diff
@@ -2370,7 +2370,7 @@ impl Connection {
                         packets.len(),
                         mtu
                     );
-                    initial.size += mtu - packets.len();
+                    initial.add_padding(mtu - packets.len());
                     // These zeros aren't padding frames, they are an invalid all-zero coalesced
                     // packet, which is why we don't increase `frame_tx.padding` count here.
                     packets.resize(mtu, 0);
```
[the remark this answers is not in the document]
martinthomson: This really only tracks the sent information.  We'd need a bunch more stuff passed in and it's not really the business of that code.  (I considered calling it "track_padding" instead, WDYT?)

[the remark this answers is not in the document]
larseggert: That'd make it a bit clearer. `add_padding` to me sounds this will add the padding to the packet.

neqo-transport/src/connection/mod.rs
```diff
@@ -2370,7 +2370,7 @@ impl Connection {
                         packets.len(),
                         mtu
                     );
-                    initial.size += mtu - packets.len();
+                    initial.add_padding(mtu - packets.len());
```
larseggert, line 2373:
```suggestion
                    initial.track_padding(mtu - packets.len());
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /// Clears the flag that had this packet on the primary path.
+    /// Used when migrating to clear out state.
+    pub fn clear_primary_path(&mut self) {
+        self.primary_path = false;
+    }
+
+    /// For Initial packets, it is possible that the packet builder needs to amend the length.
+    pub fn add_padding(&mut self, padding: usize) {
```
larseggert, line 110:
```suggestion
    pub fn track_padding(&mut self, padding: usize) {
```

larseggert (approved):

PR #1893, opened 2024-05-08
ci: Try and fix CI Fx build

larseggert:


larseggert:
This seems to work 🤞 and also eliminates vendoring in an action.

PR #1899, opened 2024-05-13
feat(bin): use quinn-udp crates.io release instead of git ref

mxinden:
`neqo-bin` has been importing `quinn-udp` as a git reference, in order to include https://github.com/quinn-rs/quinn/pull/1765. The quinn project has since released `quinn-udp` `v0.5.0`.

This commit upgrades `neqo-bin` to use `quinn-udp` `v0.5.0`.

`quinn-udp` now takes a data reference (`&[u8]`) instead of owned data (`bytes::Bytes`) on its send path, thus no longer requiring `neqo-bin` to convert, but simply pass a reference. See
https://github.com/quinn-rs/quinn/pull/1729#discussion_r1570214659 for details.

`quinn-udp` has dropped `sendmmsg` support in the `v0.5.0` release (https://github.com/quinn-rs/quinn/commit/ee0882657a16a8ffd7ff5844f355caca519e63ce). `neqo-bin` does not (yet) use `sendmmsg`. This might change in the future (https://github.com/mozilla/neqo/issues/1693).

### DIFF
--- a/neqo-bin/src/udp.rs
+++ b/neqo-bin/src/udp.rs
@@ -58,21 +58,18 @@ impl Socket {
     }
 
     /// Send the UDP datagram on the specified socket.
-    pub fn send(&self, d: Datagram) -> io::Result<()> {
+    pub fn send(&self, d: &Datagram) -> io::Result<()> {
         let transmit = Transmit {
             destination: d.destination(),
             ecn: EcnCodepoint::from_bits(Into::<u8>::into(d.tos())),
-            contents: Vec::from(d).into(),
+            contents: d,
             segment_size: None,
             src_ip: None,
         };
 
-        let n = self.socket.try_io(Interest::WRITABLE, || {
-            self.state
-                .send((&self.socket).into(), slice::from_ref(&transmit))
+        self.socket.try_io(Interest::WRITABLE, || {
+            self.state.send((&self.socket).into(), &transmit)
         })?;
-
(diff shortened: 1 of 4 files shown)


neqo-bin/src/udp.rs
```diff
@@ -58,22 +58,19 @@ impl Socket {
     }
 
     /// Send the UDP datagram on the specified socket.
-    pub fn send(&self, d: Datagram) -> io::Result<()> {
+    pub fn send(&self, d: &Datagram) -> io::Result<()> {
         let transmit = Transmit {
             destination: d.destination(),
             ecn: EcnCodepoint::from_bits(Into::<u8>::into(d.tos())),
-            contents: Vec::from(d).into(),
```
larseggert, line 65: I think we can then also remove `impl From<Datagram> for Vec<u8>`?

larseggert (approved):
See comment, otherwise LGTM.

larseggert:
I'm not loving the way in which `sendmmsg`/`recvmmsg` support was removed from `quinn-udp`, i.e., not really based on a performance analysis/measurement.

neqo-bin/src/udp.rs
mxinden: :+1: https://github.com/mozilla/neqo/pull/1899/commits/0f68d88a52c50d213f61d8dbf574675fb2fc1f38.

larseggert (approved):

PR #1900, opened 2024-05-13
ci: Try to use NSS package on Linux

larseggert:
This should speed up CI.


larseggert:
Runners are ready, but `actionlint` needs an update.

larseggert:
@martinthomson any idea what could be causing the two CHACHA20 tests to fail? https://github.com/mozilla/neqo/actions/runs/9562522589/job/26359069786?pr=1900#step:11:317

martinthomson:
No.  No idea at all.  That failure is pretty troubling.  If that were to happen in practice, it would mean a complete failure to encrypt, without hitting the error path.

Looking at the function that fails, we have a completely fresh `HpKey` instance that is successfully created.  If something went wrong there, that's not likely to result in that sort of error, so we'll skip that.

The failure has to be in the `hp.mask()` call.  That function takes a 16 byte buffer of zeros and encrypts it using ChaCha20.  What we're seeing here is a complete failure to do anything to those zero values, but the function is returning OK.  The only way I can see that happening is if NSS is somehow returning OK.  Maybe someone on the crypto team would be able to help there.  I can't spot the broken link.

larseggert:
@martinthomson so this is not an `ubuntu-24.04` issue. `ubuntu-22.04` aka `ubuntu-latest` aka `focal` also has nss-3.98 available via apt now and hits the same issue.

I don't understand why I cant reproduce this locally in either a VM or in docker.

mxinden (approved):
For what my review is worth on an NSS related change, this looks good to me.

PR #1903, opened 2024-05-14
feat: DPLPMTUD

### LINKED ISSUE #243: PMTU discovery

larseggert:
This implements a simplified variant of PLDPMTUD (aka RFC8899), which by default probes for increasingly larger PMTUs using an MTU table.

~~There is currently no attempt to repeat the PMTUD at intervals.~~ There is also no attempt to detect PMTUs that are in between values in the table. ~~There is no attempt to handle the case where the PMTU shrinks.~~

A lot of the existing tests (~50%) break when PMTUD is enabled, so this PR disables it by default. New tests that cover PMTUD were added to this PR.

Fixes #243


mxinden:
> (There are also a bunch of warning about unused code that **is** actually used. I don't understand why that is, since those functions mirror existing ones such as `cwnd_avail`.)

As far as I can tell the trait function `CongestionControl::cwnd_min` and its implementation `<ClassicCongestionControl<T> as CongestionControl>::cwnd_min` are only called in `PacketSender::cwnd_min`. `PacketSender::cwnd_min` is only called in testing code. Thus, cargo complains about the 3 not being used.

neqo-transport/src/path.rs
```diff
@@ -560,7 +554,8 @@ impl Path {
         qlog: NeqoQlog,
         now: Instant,
     ) -> Self {
-        let mut sender = PacketSender::new(cc, pacing, Self::mtu_by_addr(remote.ip()), now);
+        let pmtud = PmtudState::new(remote.ip());
+        let mut sender = PacketSender::new(cc, pacing, pmtud.clone(), now);
```
larseggert, line 558: I couldn't come up with a better way to make the PMTU state available to both the path and the sender. Any better ideas?

mxinden: In general `Rc<RefCell<Foo>>` (interior mutability) is used across the codebase. Thus I assume this is a good start.

**Subjective**: I have shied away from interior mutability when writing Rust in the past. I find code that can be mutated by different entities (e.g. each with a reference to the same `RefCell`) hard to reason about. In addition, one can no longer depend on the borrow checker at compile time, but instead only finds out about issues at runtime. Obviously there are benefits to it though, e.g. instead of passing updates across entities, state is simply updated. Thus far, I find neqo to strike a reasonable balance.

In case you want to dig deeper, I can recommend the Rust book:

https://doc.rust-lang.org/book/ch15-05-interior-mutability.html#refcellt-and-the-interior-mutability-pattern

Especially the section combining `Rc` and `RefCell`:

https://doc.rust-lang.org/book/ch15-05-interior-mutability.html#having-multiple-owners-of-mutable-data-by-combining-rct-and-refcellt

In this particular case, given that `PacketSender` is owned by `Path`, why not have `Path` access `PmtudState` through its `PacketSender`? In other words, why not have `Path` call `self.sender.some_pmtud_state_function`?

Hope I didn't misunderstand the question.

larseggert: > In this particular case, given that PacketSender is owned by Path, why not have Path access PmtudState through its PacketSender? In other words, why not have Path call self.sender.some_pmtud_state_function?

I thought about this more, and I don't see how I could make that work, because also `ClassicCongestionControl` and `Pacer` (which are both owned by `PacketSender`) need access to the `PmtudState`. Any ideas?

neqo-transport/src/crypto.rs
```diff
@@ -637,10 +637,11 @@ impl CryptoDxState {
...
         // The numbers in `Self::limit` assume a maximum packet size of 2^11.
-        if body.len() > 2048 {
-            debug_assert!(false);
-            return Err(Error::InternalError);
-        }
+        // FIXME: Figure out how to allow packets > 2048.
+        // if body.len() > 2048 {
+        //     debug_assert!(false);
+        //     return Err(Error::InternalError);
+        // }
```
larseggert, line 644: Ping @martinthomson

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+        stats.pmtud_ack += 1;
+        self.mtu = MTU_SIZES[self.probed_index];
+        qdebug!(
+            "PMTUD probe of size {} succeeded",
+            MTU_SIZES[self.probed_index]
+        );
+        self.start_pmtud();
+    }
+
+    pub fn on_packets_lost(&mut self, lost_packets: &[SentPacket], stats: &mut Stats) {
```
martinthomson, line 124: This is a useful function, but I think it could be more general.

If, at any time, at least N packets of size greater than a previous size threshold are lost, then we reduce the MTU to that threshold.

That threshold can be lower when probing up than when dropping down, but the basic theory should work.

neqo-transport/src/connection/mod.rs
```diff
@@ -2155,20 +2162,30 @@ impl Connection {
...
         }
 
-        if primary {
-            if space == PacketNumberSpace::ApplicationData {
-                self.write_appdata_frames(builder, &mut tokens);
-            } else {
-                let stats = &mut self.stats.borrow_mut().frame_tx;
-                self.crypto.write_frame(space, builder, &mut tokens, stats);
+        if path.borrow().pmtud().borrow().is_probe_prepared() {
+            // If this is a PMTUD probe, don't include any other frames and record it as sent.
```
martinthomson, line 2166: Does this mean that we are displacing "real" frames with a probe?

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -98,8 +98,16 @@ pub fn migration_with_modifiers(
     burst: usize,
 ) -> (IpTos, IpTos, bool) {
     fixture_init();
-    let mut client = new_client(ConnectionParameters::default().max_streams(StreamType::UniDi, 64));
-    let mut server = new_server(ConnectionParameters::default().max_streams(StreamType::UniDi, 64));
+    let mut client = new_client(
+        ConnectionParameters::default()
+            .pmtud(false)
```
martinthomson, line 103: I think that it would be better if PMTUD were disabled by default.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+const MTU_SIZES: [usize; 13] = [
+    65536, // Official maximum MTU          RFC 791
+    65535, // Hyperchannel                  RFC 1044
+    17914, // 16Mb IBM Token Ring           ref. [6]
+    16384, // macOS loopback
+    8166,  // IEEE 802.4                    RFC 1042
+    4464,  // IEEE 802.5 (4Mb max)          RFC 1042
+    4352,  // FDDI (Revised)                RFC 1188
+    2048,  // Wideband Network              RFC 907
+    2002,  // IEEE 802.5 (4Mb recommended)  RFC 1042
```
martinthomson, line 23: The sizes from here on up are pretty uninteresting.  Do we really need 2002 and 2048?  The utility gain from an increase that small probably doesn't outweigh the cost of probing.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+    1536,  // Exp. Ethernet Nets            RFC 895
+    1500,  // Ethernet Networks             RFC 894
+    1492,  // IEEE 802.3                    RFC 1042
```
martinthomson, line 26: There's a lot of clear air between 1280 and this value, but that's the area where we are most likely to see real paths sitting.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+// except according to those terms.
+
+use std::{cell::RefCell, net::IpAddr, rc::Rc};
+
+use neqo_common::{qdebug, qtrace};
+
+use crate::{recovery::SentPacket, Stats};
+
+// From https://datatracker.ietf.org/doc/html/rfc1191#section-7.1, with a few modifications.
+const MTU_SIZES: [usize; 13] = [
```
martinthomson, line 14: Overall, I think that the steps here are a bit questionable.  There are some tiny steps and some huge ones.   Is there some empirical research that might inform the steps we use?  RFC 1191 and the RFCs you cite are all ancient.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+        match remote_ip {
+            IpAddr::V4(_) => 20 + 8,
+            IpAddr::V6(_) => 40 + 8,
+        }
+    }
+
+    #[must_use]
+    pub const fn default_plpmtu(remote_ip: IpAddr) -> usize {
+        MTU_SIZES[MTU_SIZES.len() - 1] - Self::header_size(remote_ip)
+    }
```
martinthomson, line 179: I wonder if we can't keep the MTU we had previously, even as a starting point for when we probe.

We don't have a way to drop down to a smaller MTU, but we could (see above).

neqo-transport/src/path.rs
```diff
@@ -776,13 +774,25 @@ impl Path {
         &mut self,
         builder: &mut PacketBuilder,
         stats: &mut FrameStats,
-        mtu: bool, // Whether the packet we're writing into will be a full MTU.
+        mtu: bool,       // Whether the packet we're writing into will be a full MTU.
+        empty_pkt: bool, // False packet is coalesced behind another one or already has frames.
```
martinthomson, line 778: Check the comment.  I'm not following it.

martinthomson (commented):
I'm not seeing PMTUD tests, which would be necessary for this.

The big question I have is the one that Christian makes about PTMUD generally: how do you know that the bytes you use on PMTUD pay you back?

There is probably a case for sending probes when you have spare sending capacity and nothing better to send.  Indeed, successfully probing will let us push congestion windows up more and could even improve performance.

What I'm seeing here displaces other data.  I'd like to see something that doesn't do that.  There's a fundamental problem that needs analysis though.  You can't predict that a connection will be used for uploads, so you don't know when probes will really help.  I see a few cases:

1. The connection is short-lived or low volume. Probes are strictly wasteful.
2. The connection is long-lived and high volume, with ample idle time for probing.  Probes can use gaps.  This might be a video stream, where probing can fit into a warmup period.  Probes are therefore easy and super-helpful.
3. The connection exists only to support a smaller upload.  The upload is small enough that probes are wasteful.
4. The connection exists only to support a larger upload.  The upload is large enough that spending bytes on probing early on is a good investment.

Case 1 and 2 are easy to deal with.  We could probe on an idle connection and tolerate a small amount of waste for case 1 if it makes case 2 appreciably better.

The split between 3 and 4 is rough.  There is an uncertain zone between the two as well where some probing is justified, but successive rounds of probing might be wasteful as the throughput gain over the remaining time diminishes relative to the wasted effort of extra probes.

Right now, you don't send real data in probes.  You are effectively betting on the probes being lost.  But you could send data, which would reduce the harm in case 3.  It might even make the code slightly simpler.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use std::{cell::RefCell, net::IpAddr, rc::Rc};
+
+use neqo_common::{qdebug, qtrace};
+
+use crate::{recovery::SentPacket, Stats};
+
+// From https://datatracker.ietf.org/doc/html/rfc1191#section-7.1, with a few modifications.
+const MTU_SIZES: [usize; 13] = [
+    65536, // Official maximum MTU          RFC 791
```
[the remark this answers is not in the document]
larseggert: QUIC assumes an (interface) MTU of 1280, which is the IPv6 minimum? https://datatracker.ietf.org/doc/html/rfc9000#name-datagram-size

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,180 @@
...
+    16384, // macOS loopback
+    8166,  // IEEE 802.4                    RFC 1042
+    4464,  // IEEE 802.5 (4Mb max)          RFC 1042
+    4352,  // FDDI (Revised)                RFC 1188
+    2048,  // Wideband Network              RFC 907
+    2002,  // IEEE 802.5 (4Mb recommended)  RFC 1042
+    1536,  // Exp. Ethernet Nets            RFC 895
+    1500,  // Ethernet Networks             RFC 894
+    1492,  // IEEE 802.3                    RFC 1042
+    1280,  // IPv6 minimum MTU              RFC 2460
```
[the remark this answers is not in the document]
larseggert: QUIC assumes an (interface) MTU of 1280, which is the IPv6 minimum? https://datatracker.ietf.org/doc/html/rfc9000#name-datagram-size

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
larseggert: Yeah, typo.

larseggert:
> I'm not seeing PMTUD tests, which would be necessary for this.

Yeah, tests are definitely still needed.

> There is probably a case for sending probes when you have spare sending capacity and nothing better to send. Indeed, successfully probing will let us push congestion windows up more and could even improve performance.

Yep. It's on my to-do list to move the probes to the end of a flight of data, if cwnd is left. The current way leads to a more predictable send pattern, which helped as I developed this.

> Right now, you don't send real data in probes. You are effectively betting on the probes being lost. But you could send data, which would reduce the harm in case 3. It might even make the code slightly simpler.

https://datatracker.ietf.org/doc/html/rfc9000#name-sending-quic-pmtu-probes talks specifically about probing with PING and PADDING. I was actually wondering why that was the case, but decided to simply do that initially.

neqo-transport/src/pmtud.rs
larseggert: I did have a version that used binary search, maybe that is something to do here. (lsquic does, I think.)

neqo-transport/src/pace.rs
```diff
@@ -49,14 +52,17 @@ impl Pacer {
     /// The value of `p` is the packet size in bytes, which determines the minimum
     /// credit needed before a packet is sent.  This should be a substantial
     /// fraction of the maximum packet size, if not the packet size.
-    pub fn new(enabled: bool, now: Instant, m: usize, p: usize) -> Self {
-        assert!(m >= p, "maximum capacity has to be at least one packet");
+    pub fn new(enabled: bool, now: Instant, m: usize, pmtud: PmtudRef) -> Self {
```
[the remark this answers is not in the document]
larseggert: I had a discussion with @mxinden about how to best structure this; he didn't like RefCell either.

The issue is that the data structures are nested like this:
* Path
   * PacketSender
      * CongestionControlAlgorithm
      * Pacer
     
and all of them need access to one Pmtud instance. In C, I'd have made it part of Path and given the contained structs a pointer to that instance, but I wasn't able to figure out how to do something like that in Rust.

neqo-transport/src/pmtud.rs
larseggert: @gorryfair had a paper a few years back: https://ieeexplore.ieee.org/document/8506538. I wonder if we should just do a binary search (to within 4-16 bytes) on the range [1280..1500].

neqo-transport/src/pmtud.rs
larseggert: This is really only used in tests. I couldn't make it `#[cfg(test)]`, because test-fixture also needs it.

neqo-transport/src/connection/tests/ecn.rs
larseggert: I was hoping that at least some tests wouldn't fail with it enabled by default and so exercise some of the logic, but I can flip the default and do all-new PMTUD tests.

neqo-transport/src/pmtud.rs
larseggert: Screenshot 2024-05-22 at 09 34 07
Note this is TCP MSS, so add mentally headers.

larseggert:
On a positive note, look at the benchmark improvements! Looks like our crappy performance against msquic was mostly because we didn't send 64KB packets over loopback :-)

neqo-transport/src/connection/mod.rs
```diff
@@ -2286,8 +2301,15 @@ impl Connection {
...
             } else {
-                (tokens, ack_eliciting, padded) =
-                    self.write_frames(path, *space, &profile, &mut builder, now);
+                (tokens, ack_eliciting, padded) = self.write_frames(
+                    path,
+                    *space,
+                    &profile,
+                    &mut builder,
+                    header_start != 0,
+                    aead_expansion,
```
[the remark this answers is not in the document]
larseggert: Sorry, putting _what_ in the builder?

neqo-transport/src/connection/mod.rs
larseggert: Yes. I'll change it to using payload data.

neqo-transport/src/pmtud.rs
martinthomson: That's useful.  I'd start by uncompressing the 1280 barrier for v6, which leads to a very similar graph overall on both.

It suggests that there isn't a whole lot of value in probing below 1380 or so (our current value).  Given how few people are in that bucket, we might ignore anything below 1350 or so, like we have done, though if we are able to do continuous checking as I suggested, having 1200 on the list as a backstop would deal with those odd few cases where we start at 1350 and the path changes.

Beyond that, we should probe at just a few steps.  I see one at ~1420 (I'm correcting for TCP header, adjust accordingly) for v4 that looks like a good point to check, then ~1460 for v6, and finally 1500 for both.  Beyond that, I don't see much value in doing anything other than a few points (16k and 64k, and maybe add 4k).  In practice, we're not going to see that often, outside of loopback tests.  Besides, the faster we can hit the 64k number, the better.

Given these numbers, I'm not seeing much value in a binary search.  The graph is just too flat to bother.  That region of slope is narrow enough that trying to dial in an exact number there doesn't seem to be worth it.

neqo-transport/src/crypto.rs
larseggert: @martinthomson what to do about this?

martinthomson: Good question.  I missed this in my first pass.

My suggestion would be to change `Self::limit()` to `self.limit()` instead, with a variable on the object and a check here to update the value if necessary.

What we want is to halve the current limit for every bit that is set starting from bit 11 (i.e., 2048).  You can use `u64::leading_zeros` to compute that efficiently.  Once you halve the limit, you should not increase it again.

We'd want to assert (debug_assert, probably) that you don't exceed 2^16 bits of data.

An alternative approach is to shift to counting the total number of 16 byte blocks (rounded up!)  You would increase the limit by a factor of 2^7 as a result.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,462 @@
...
+        let Some(last_good) = self.loss_counts.iter().rposition(|&c| c >= MAX_PROBES) else {
+            // If not, keep going.
+            if lost > 0 {
+                // Don't stop the PMTUD process.
+                self.probe_state = Probe::Needed;
+            }
+            return;
+        };
+
+        qdebug!("Packet of size > {} lost {} times", self.mtu, MAX_PROBES);
```
mxinden, line 224:
```suggestion
        qdebug!("Packet of size > {} lost >= {} times", self.mtu, MAX_PROBES);
```
 Would be more correct, right?

mxinden, line 224: Is `self.mtu` correct here? Shouldn't it be `last_good`? Couldn't there be a scenario where `last_good < self.mtu`, i.e. in the case where the MTU on a path lowers throughout its lifetime?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -281,12 +308,22 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             &[QlogMetric::BytesInFlight(self.bytes_in_flight)],
         );
 
+        // Lost PMTUD probes do not elicit a congestion control reaction.
+        let lost_packets: Vec<SentPacket> = lost_packets
+            .iter()
+            .filter(|pkt| !self.pmtud.is_pmtud_probe(pkt))
+            .cloned()
+            .collect();
+        if lost_packets.is_empty() {
```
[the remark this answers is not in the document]
mxinden: See also https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.peekable.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// Checks whether a PMTUD probe has been lost. If it has been lost more than `MAX_PROBES`
+    /// times, the PMTUD process is stopped.
+    pub fn on_packets_lost(
+        &mut self,
+        lost_packets: &[SentPacket],
+        stats: &mut Stats,
+        now: Instant,
+    ) {
+        // Track lost probes
+        let lost = self.count_pmtud_probes(lost_packets);
```
mxinden, line 202: Is there a scenario where `lost` can be larger than `1`? In other words, is there a scenario where more than one probe is in-flight at once?

If I understand correctly `prepare_probe` is only called in `Probe::Needed`. `prepare_probe` switches to `Probe::Prepared`. The state only switches back to `Probe::Needed`, when the in-flight probe has been lost. Thus there is never more than one probe in-flight.

If the above is correct, why is there a `count_pmtud_probes` function? Am I missing something?

larseggert:
Making this a draft until the test failure is fixed.

larseggert:
Working on increasing test coverage.

neqo-transport/src/connection/mod.rs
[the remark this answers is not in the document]
larseggert: @martinthomson ping

neqo-transport/src/pmtud.rs
larseggert: @martinthomson does this implement what you suggested?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -281,12 +308,22 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+            return false;
+        }
+
         let congestion = self.on_congestion_event(lost_packets.last().unwrap());
```
mxinden, line 321:
```suggestion
        let Some(last_lost_packet) = lost_packets.last() else {
            return false;
        }

        let congestion = self.on_congestion_event(last_lost_packet);
```

Making the invalid state impossible at compile time instead of a panic at runtime.

mxinden (commented):
Will take an in-depth look tomorrow.

neqo-transport/Cargo.toml
```diff
@@ -20,6 +20,7 @@ neqo-common = { path = "../neqo-common" }
 neqo-crypto = { path = "../neqo-crypto" }
 qlog = { workspace = true }
 smallvec = { version = "1.11", default-features = false }
+static_assertions = { version = "1.1", default-features = false }
```
mxinden, line 23: This is only used in `neqo-transport/src/connection/tests/datagram.rs`, right? If so, how about moving it to `[dev-dependencies]`?

neqo-transport/src/connection/tests/priority.rs
```diff
@@ -386,7 +386,7 @@ fn low() {
     // Send a session ticket and make it big enough to require a whole packet.
     // The resulting CRYPTO frame beats out the stream data.
     let stats_before = server.stats().frame_tx;
-    server.send_ticket(now, &[0; 2048]).unwrap();
+    let _ = server.send_ticket(now, &vec![0; server.plpmtu().unwrap()]);
```
mxinden, line 389: Why can a potential `Err` be ignored here?

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// Count the PMTUD probes included in `pkts`.
+    fn count_pmtud_probes(&self, pkts: &[SentPacket]) -> usize {
+        pkts.iter().filter(|p| self.is_pmtud_probe(p)).count()
+    }
+
+    /// Checks whether a PMTUD probe has been acknowledged, and if so, updates the PMTUD state.
+    /// May also initiate a new probe process for a larger MTU.
+    pub fn on_packets_acked(&mut self, acked_pkts: &[SentPacket], stats: &mut Stats) {
+        let acked = self.count_pmtud_probes(acked_pkts);
+        if acked == 0 {
```
mxinden, line 151: Do I understand correctly that `acked` could be `>1` in the case where a probe from a previous iteration is acked?

mxinden (commented):
First high level review.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,443 @@
...
+    NotNeeded,
+    Needed,
+    Sent,
+}
+
+#[derive(Debug)]
+pub struct Pmtud {
+    search_table: &'static [usize],
+    header_size: usize,
+    mtu: usize,
```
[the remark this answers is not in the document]
larseggert: I thought it was the most appropriate Rust type to use; maybe I was wrong? I which case `u16` might be sufficient, since that is the de facto limit.

neqo-transport/src/connection/tests/priority.rs
larseggert: It's a test, so I assumed a panic is OK if there is an issue.

neqo-transport/src/pmtud.rs
larseggert: Or if the ACK got delayed sufficiently by the sender so that it includes multiple probes.

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
martinthomson: When it comes to lengths, it is a bit of nuisance to use anything other than usize in rust.  usize is the same as u32 on basically none of the platforms we use (though we seem to have 32-bit builds still, a lot of those platforms still use a 64-bit usize, apart from windows, I think).  Either one is of course overkill, but the wasted bytes really aren't worth bothering with.

neqo-transport/src/pmtud.rs
martinthomson: That's right.  We can declare something lost, probe again and then find that we have two acknowledgments to handle.  That shouldn't change anything here.

martinthomson (commented):
I've lost track of where I'm up to with reviews on this.  Take a look.

(Draft deadline on Monday and I have two big drafts to land with some people who need coaching, so limited time.)

larseggert:
This can wait until after.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -159,6 +166,25 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
 
+    #[must_use]
+    fn cwnd_min(&self) -> usize {
+        self.max_datagram_size() * 2
+    }
+
+    #[cfg(test)]
+    #[must_use]
+    fn cwnd_initial(&self) -> usize {
+        cwnd_initial(self.pmtud.plpmtu())
```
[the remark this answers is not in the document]
larseggert: We could use the minimum here, true. Would it save us much?

neqo-transport/src/connection/dump.rs
```diff
@@ -27,6 +27,7 @@ pub fn dump_packet(
     pn: PacketNumber,
     payload: &[u8],
     tos: IpTos,
+    len: usize,
```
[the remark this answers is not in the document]
larseggert: This is the total datagram length.

neqo-transport/src/crypto.rs
```diff
@@ -462,6 +464,7 @@ impl CryptoDxState {
             used_pn: 0..0,
             min_pn: 0,
             invocations: Self::limit(direction, cipher),
```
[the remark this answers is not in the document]
larseggert: Sorry, I don't follow?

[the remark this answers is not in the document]
martinthomson: Oh, that was clearly a brain fart.  This is fine.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// Returns the MTU search table for the given remote IP address family.
+    const fn search_table(remote_ip: IpAddr) -> &'static [usize] {
+        match remote_ip {
+            IpAddr::V4(_) => &MTU_SIZES_V4,
+            IpAddr::V6(_) => &MTU_SIZES_V6,
+        }
+    }
+
+    /// Size of the IPv4/IPv6 and UDP headers, in bytes.
+    const fn header_size(remote_ip: IpAddr) -> usize {
```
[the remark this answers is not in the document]
larseggert: No, this also includes the UDP header.

neqo-transport/src/pmtud.rs
larseggert: > So maybe this whole thing isn't quite right and you only need to track whether the probes worked out (which is far simpler as well).

If we did that, we need a different mechanism to detect when a path change causes the PMTU to drop.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,443 @@
...
+        lost_packets: &[SentPacket],
+        stats: &mut Stats,
+        now: Instant,
+    ) {
+        // Track lost probes
+        let lost = self.count_pmtud_probes(lost_packets);
+        stats.pmtud_lost += lost;
+
+        // Increase loss counts for all sizes included in the lost packets.
+        for (count, inc) in self.loss_counts.iter_mut().zip(
```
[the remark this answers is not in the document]
larseggert: I made a bunch of changes inspired by the sketch above, please take another look.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /// Stops the PMTUD process, setting the MTU to the largest successful probe size.
+    fn stop_pmtud(&mut self, idx: usize, now: Instant) {
+        self.probe_state = Probe::NotNeeded; // We don't need to send any more probes
+        self.probe_index = idx; // Index of the last successful probe
+        self.mtu = self.search_table[idx]; // Leading to this MTU
+        self.probe_count = 0; // Reset the count
+        self.loss_counts.fill(0); // Reset the loss counts
+        self.raise_timer = Some(now + PMTU_RAISE_TIMER);
```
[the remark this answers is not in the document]
larseggert: I think the argument for a fixed time is that PMTU changes would be due to routing changes, which are not correlated with the RTT. Whether 10 min is a good number is a valid question though.

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
mxinden: Agreed, that the wasted bytes don't matter. I was hoping to reduce the large noise, converting `usize` to `f64`, in this pull request. Not a strong opinion.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,644 @@
...
+        let Some(first_failed) = self.loss_counts.iter().position(|&c| c >= MAX_PROBES) else {
+            // If not, keep going.
+            if lost > 0 {
+                // Don't stop the PMTUD process.
+                self.probe_state = Probe::Needed;
+            }
+            return;
+        };
+
+        if first_failed > 0 {
```
mxinden, line 248: Given the `if idx > 0 &&` above, how can `first_failed` ever be `0`?

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,644 @@
...
+            .count();
+        self.loss_counts.iter_mut().take(idx).for_each(|c| *c = 0);
+
+        let acked = self.count_pmtud_probes(acked_pkts);
+        if acked == 0 {
+            return;
+        }
+
+        // A probe was ACKed, confirm the new MTU and try to probe upwards further.
+        stats.pmtud_ack += acked;
```
martinthomson, line 165: Something for later perhaps: Maybe we should be tracking stats on a per-probe-size basis rather than just the total number of successful probes.

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
martinthomson: The reason I suggest this is that we don't have any fixed time values in the stack thus far.

This is the second number that is divorced from RTT.  The other is the idle timeout.  For that, we have a connection parameter.  Perhaps this needs one too.  (I'm thinking about our crazy RTT tests, where this will definitely affect things in surprising ways.)

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,644 @@
...
+        }
+
+        let mut increase = vec![0; self.search_table.len()];
+        let mut loss_counts_updated = false;
+        for p in lost_packets {
+            let idx = self
+                .search_table
+                .iter()
+                .take_while(|&&sz| p.len() > sz - self.header_size)
+                .count();
```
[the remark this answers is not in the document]
martinthomson: There's another use of take_while above too.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,644 @@
...
+            qdebug!(
+                "Packet of size > {} lost >= {} times",
+                self.search_table[last_ok],
+                MAX_PROBES
+            );
+            if self.probe_state == Probe::NotNeeded {
+                // We saw multiple losses of packets <= the current MTU outside of PMTU discovery,
+                // so we need to probe again. To limit connectivity disruptions, we start the PMTU
+                // discovery from the smallest packet up, rather than the failed packet size down.
+                self.restart_pmtud(stats);
```
martinthomson, line 259: Want to add a TODO here:

If we are declaring losses, that means that we'll be getting packets through.  The size of those will put a floor on the MTU.  We're currently conservative and starting from scratch, but we don't strictly need to do that.

(If our loss recovery processing deals with lost packets before acked packets, maybe this ends up being a no-op, I haven't checked that.  That said, if it IS a no-op, we should explain that here and maybe write some tests in support of that.  Hence the suggestion for a TODO.)

martinthomson (approved):
I like the way this has come out.  Max had some good comments and I have a few minor things, but I think we should try this and see what the performance looks like.

neqo-transport/src/connection/mod.rs
```diff
@@ -2131,7 +2133,7 @@ impl Connection {
 
         // Avoid sending probes until the handshake completes,
         // but send them even when we don't have space.
-        let full_mtu = profile.limit() == path.borrow().mtu();
+        let full_mtu = profile.limit() == path.borrow().plpmtu();
```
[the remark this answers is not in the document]
larseggert: I think you are right.

neqo-transport/src/pmtud.rs
```diff
@@ -0,0 +1,644 @@
...
+        let search_table = Self::search_table(remote_ip);
+        let probe_index = 0;
+        Self {
+            search_table,
+            header_size: Self::header_size(remote_ip),
+            mtu: search_table[probe_index],
+            probe_index,
+            probe_count: 0,
+            probe_state: Probe::NotNeeded,
+            loss_counts: vec![0; search_table.len()],
```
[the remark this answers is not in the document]
larseggert: We could add 1420 to the v6 table? Makes them the same length and also adds an intermediate value between 1380 and 1470.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -281,12 +308,22 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             &[QlogMetric::BytesInFlight(self.bytes_in_flight)],
         );
 
-        let congestion = self.on_congestion_event(lost_packets.last().unwrap());
+        // Lost PMTUD probes do not elicit a congestion control reaction.
+        let lost_packets: Vec<SentPacket> = lost_packets
+            .iter()
+            .filter(|pkt| !self.pmtud.is_pmtud_probe(pkt))
+            .cloned()
+            .collect();
```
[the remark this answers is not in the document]
larseggert: @max made [another suggestion](https://github.com/mozilla/neqo/pull/1903#discussion_r1617011637). I'll see what I can do.

neqo-transport/src/cc/classic_cc.rs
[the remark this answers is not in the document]
larseggert: @mxinden I need some Rust help to make this fly. I changed `detect_persistent_congestion` to be
```rust
    fn detect_persistent_congestion<'a, I>(
        &mut self,
        first_rtt_sample_time: Option<Instant>,
        prev_largest_acked_sent: Option<Instant>,
        pto: Duration,
        lost_packets: I,
    ) -> bool
    where
        I: Iterator<Item = &'a SentPacket>,
```
and have this in `on_packets_lost`:
```rust
        // Lost PMTUD probes do not elicit a congestion control reaction.
        let lost_packets = lost_packets.iter().filter(|pkt| !self.pmtud.is_probe(pkt));
        let Some(last_lost_packet) = lost_packets.last() else {
            return false;
        };

        let congestion = self.on_congestion_event(&last_lost_packet);
        let persistent_congestion = self.detect_persistent_congestion(
            first_rtt_sample_time,
            prev_largest_acked_sent,
            pto,
            lost_packets,
        );
```
But I get some borrow checker errors (`cannot borrow `*self` as mutable because it is also borrowed as immutable`) that I can't work through.

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
larseggert: If we wanted to make this per-RTT, we could roughly say that 10min = 12000 "standard" RTTs @ 50ms, and just work with some sort of RTT count like that?

neqo-transport/src/cc/classic_cc.rs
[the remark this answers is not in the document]
martinthomson: If you are holding a mutable reference to `self.foo` and you call `self.bar()`, that's not going to work.  The problem here is that you are holding a reference to `self.pmtud` via the iterator. 

Options I can think of without sitting down with this:

* You can change `detect_persistent_congestion` to not take `&mut self` and instead take mutable (or non-mutable) references to the fields it cares about.
* You can move `self.pmtud` off (`std::mem::take`) before making the call, restoring it afterwards.
* You can wait for the rust language to be updated to be smarter about what members are touched in functions.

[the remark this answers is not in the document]
martinthomson: * Just ask PMTUD for the packet size that is needed to identify a PMTUD probe and do this:

```rust
let probe_size = self.pmtud.min_probe_size();
let lost_packets = lost_packets.iter().filter(|pkt| pkt.len() < probe_size).rev().peekable();
let Some(last_lost_packet) = lost_packets.peek() {
  return false;
}

let congestion = self.on_congestion_event(&last_lost_packet);
        let persistent_congestion = self.detect_persistent_congestion(
            first_rtt_sample_time,
            prev_largest_acked_sent,
            pto,
            lost_packets.rev(), // flip it back over again
        );
```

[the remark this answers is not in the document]
mxinden: https://github.com/larseggert/neqo/pull/25

> You could instead define `self.pmtud.is_not_probe_fn()` to return a function like `Fn(pkt: SentPacket) -> bool` and pass that to filter instead. That maintains encapsulation, sort of, but I'm not sure that rust is that good about how it manages captured variables.

That is a neat idea @martinthomson, especially since it keeps the logic encapsulated within `pmtud.rs`.

In case you like it @larseggert, I implemented it in https://github.com/larseggert/neqo/pull/25.

neqo-transport/src/connection/mod.rs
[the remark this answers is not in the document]
mxinden: No change. Just wanted to challenge my understanding. Thank you.

neqo-transport/src/pmtud.rs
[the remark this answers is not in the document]
larseggert: If we want to make this change, let's do it in a follow-on PR.

PR #1910, opened 2024-05-21
ci: Set `RUST_LOG` during coverage tests

larseggert:
So that logging-related code is being included in the coverage check.


KershawChang (approved):

larseggert:
Holding this back pending https://github.com/codecov/codecov-action/issues/1425

PR #1913, opened 2024-05-23
Enable Xyber768d00 on servers by default

jschanck:
This enable Xyber768d00 on servers by default. It's the easiest way to get tests passing in https://phabricator.services.mozilla.com/D208048. If you'd prefer not to accept Xyber shares by default, then I can plumb the configuration up to Http3Server.


neqo-transport/src/crypto.rs
```diff
@@ -76,20 +76,36 @@ impl Crypto {
...
+        match &mut agent {
+            Agent::Server(c) => {
+                // Clients do not send xyber shares by default, but servers should accept them.
+                c.set_groups(&[
+                    TLS_GRP_KEM_XYBER768D00,
+                    TLS_GRP_EC_X25519,
+                    TLS_GRP_EC_SECP256R1,
+                    TLS_GRP_EC_SECP384R1,
+                    TLS_GRP_EC_SECP521R1,
+                ])?
```
larseggert, line 88:
```suggestion
                ])?;
```

larseggert (approved):

neqo-transport/src/crypto.rs
```diff
@@ -76,20 +76,36 @@ impl Crypto {
...
+                    TLS_GRP_EC_SECP384R1,
+                    TLS_GRP_EC_SECP521R1,
+                ])?;
+
+                // Configure clients to send both X25519 and P256 to reduce
+                // the rate of HRRs.
+                c.send_additional_key_shares(1)?;
+
+                // Always enable 0-RTT on the client, but the server needs
+                // more configuration passed to server_enable_0rtt.
```
martinthomson, line 103: :+1:

neqo-transport/src/crypto.rs
```diff
@@ -76,20 +76,36 @@ impl Crypto {
...
+            Agent::Client(c) => {
+                c.set_groups(&[
+                    TLS_GRP_EC_X25519,
+                    TLS_GRP_EC_SECP256R1,
+                    TLS_GRP_EC_SECP384R1,
+                    TLS_GRP_EC_SECP521R1,
+                ])?;
+
+                // Configure clients to send both X25519 and P256 to reduce
+                // the rate of HRRs.
```
martinthomson, line 99: With the size increase from Xyber, we still have the extra space without adding another packet to the ClientHello, right?

martinthomson (approved):
These are fine defaults.   Our server is for testing, so having it prefer Xyber makes it easier to tweak the client end.

I made a note in the other patch regarding these as defaults.  I'm not sure how much we want to build redundancies vs. having the code in one place.  I tend to think that you could drop the else clause there.

One ask.  Can you add a connection test with a client that is configured for Xyber?  You have one in gecko, but I'd prefer it if this crate has one too.  Just the one should do.

larseggert:
FYI, I'm looking into the CI bench failures.

neqo-transport/src/crypto.rs
jschanck: Extra space to send a P256 share? Good question---I don't know the exact packet size cutoffs.

jschanck: This is worth looking into. I'm going to mark this as resolved since the client configuration in this patch does not include Xyber. We can discuss further in https://phabricator.services.mozilla.com/D208048.

PR #1914, opened 2024-05-24
fix: Use correct ECN mark when sending datagram

larseggert:
`EcnInfo::on_packet_sent` may update the ECN state and can hence change it.

Also throw in some unrelated fixes to make the ECN log output more quiet, and to test/test.sh.

### DIFF
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -706,8 +706,12 @@ impl Path {
 
     /// Make a datagram.
     pub fn datagram<V: Into<Vec<u8>>>(&mut self, payload: V) -> Datagram {
+        // Make sure to use the TOS value from before calling EcnInfo::on_packet_sent, which may
+        // update the ECN state and can hence change it - this packet should still be sent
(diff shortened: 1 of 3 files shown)


mxinden (approved):

neqo-transport/src/path.rs
```diff
@@ -706,8 +706,12 @@ impl Path {
 
     /// Make a datagram.
     pub fn datagram<V: Into<Vec<u8>>>(&mut self, payload: V) -> Datagram {
+        // Make sure to use the TOS value from before calling EcnInfo::on_packet_sent, which may
+        // update the ECN state and can hence change it - this packet should still be sent
+        // with the current value.
+        let tos = self.tos();
         self.ecn_info.on_packet_sent();
```
mxinden, line 713: As an alternative, `on_packet_sent` could be renamed to e.g. `on_sending_packet` and changed to return the _to-be-used_ ecn mark for the _to-be-send_ packet.

I don't feel strongly about it.

PR #1915, opened 2024-05-27
fix: Don't initialize regex on each call to `stream_readable`

larseggert:
Each call takes ~1ms.


KershawChang (approved):

mxinden (approved):
:+1: 

Additional background: https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop

PR #1920, opened 2024-06-07
feat(bin): enable Firefox to use quinn-udp through neqo-udp

mxinden:
- Needed for:
    - https://bugzilla.mozilla.org/show_bug.cgi?id=1901295.
    - https://phabricator.services.mozilla.com/D212959
- Moves UDP related logic into new `neqo-udp` crate. Firefox does not need `neqo-server`, `neqo-client`, nor any of its dependencies. A new crate enforces none of this leaking into Firefox.
- Allow UDP `Socket` to be used both with `tokio::net::UdpSocket` (for `neqo-client`, `neqo-server`, mozilla-central's `http3server`) *and* `std::os::fd::BorrowedFd`  (`std::os::windows::io::BorrowedSocket` on Windows). The latter allows `neqo-glue` to use `neqo-udp` with a UDP socket managed by `NSPR`.


neqo-common/src/log.rs
```diff
@@ -50,7 +50,7 @@ fn since_start() -> Duration {
     START_TIME.get_or_init(Instant::now).elapsed()
 }
 
-pub fn init(level_filter: Option<log::LevelFilter>) {
+pub fn init(level_filter: Option<::log::LevelFilter>) {
```
KershawChang, line 53: Unrelated change?

KershawChang (approved):
Looks good to me.

@martinthomson , @larseggert , could you also review this PR?
Thanks.

neqo-udp/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+workspace = true
+
+[dependencies]
+log = { workspace = true }
+neqo-common = { path = "./../neqo-common" }
+quinn-udp = { version = "0.5.0", default-features = false }
+tokio = { version = "1", default-features = false, features = ["net"], optional = true }
+
+[dev-dependencies]
+tokio = { version = "1", default-features = false, features = ["net", "time", "macros", "rt", "rt-multi-thread"] }
```
larseggert, line 21: Why is `tokio` also a `dev-dependency`?

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
 
-use neqo_common::{Datagram, IpTos};
+use neqo_common::{qdebug, Datagram, IpTos};
 use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
-use tokio::io::Interest;
 
 /// Socket receive buffer size.
 ///
 /// Allows reading multiple datagrams in a single [`Socket::recv`] call.
 const RECV_BUF_SIZE: usize = u16::MAX as usize;
```
larseggert, line 23: This is the max on all platforms?

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+        };
+
+        state.send(socket, &transmit)?;
+
+        qdebug!(
+            "sent {} bytes from {} to {}",
+            d.len(),
+            d.source(),
+            d.destination()
+        );
```
larseggert, line 57: Can this be `qtrace!` or do we maybe not need it at all?

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+
+                Ok(recv_buf[0..meta.len]
+                    .chunks(meta.stride.min(recv_buf.len()))
+                    .map(|d| {
+                        qdebug!(
+                            "received {} bytes from {} to {}",
+                            d.len(),
+                            meta.addr,
+                            local_address,
+                        );
```
larseggert, line 85: Ditto.

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+                        Datagram::new(
+                            meta.addr,
+                            *local_address,
+                            meta.ecn.map(|n| IpTos::from(n as u8)).unwrap_or_default(),
+                            None, // TODO: get the real TTL https://github.com/quinn-rs/quinn/issues/1749
```
larseggert, line 90: I wonder if we should remove the TTL stuff from neqo until quinn actually supports it...

neqo-udp/Cargo.toml
martinthomson: Presumably because the dev variant has more features enabled.

The lack of an EOL here is also worth fixing.  You can configure your editor to add these to all text files.

neqo-udp/src/lib.rs
martinthomson: That seems unlikely.  But @larseggert, does that need to be fixed here, or can we track that with another issue?

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+}
+
+pub struct Socket<S> {
+    state: UdpSocketState,
     #[allow(unknown_lints)] // available with Rust v1.75
     #[allow(clippy::struct_field_names)]
-    socket: tokio::net::UdpSocket,
-    state: UdpSocketState,
-    recv_buf: Vec<u8>,
+    socket: S,
```
martinthomson, line 33: Rather than suppress lints, maybe this can now just be:

```suggestion
    inner: S,
```

neqo-udp/src/lib.rs
martinthomson: I like `qtrace` for this one.  Sometimes you need to see every packet.

neqo-udp/src/lib.rs
martinthomson: Separate issue, but I'm not sure that we really need TTL, so I'd be happy to see it removed.

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+                            *local_address,
+                            meta.ecn.map(|n| IpTos::from(n as u8)).unwrap_or_default(),
+                            None, // TODO: get the real TTL https://github.com/quinn-rs/quinn/issues/1749
+                            d,
+                        )
+                    })
+                    .collect())
+            })?;
+
+        qdebug!(
```
martinthomson, line 97: Also `qtrace`

neqo-udp/src/lib.rs
```diff
@@ -8,37 +8,133 @@
...
+
+        Ok(())
+    }
+
+    fn recv_inner(
+        local_address: &SocketAddr,
+        state: &UdpSocketState,
+        socket: quinn_udp::UdpSockRef<'_>,
+    ) -> Result<Vec<Datagram>, io::Error> {
+        let mut meta = RecvMeta::default();
```
martinthomson, line 67: This can move inside the closure.

martinthomson (approved):
LGTM, with nits.

neqo-udp/Cargo.toml
mxinden: Correct.

Firefox does not need `tokio`. `neqo-client` and `neqo-server` in `neqo-bin` do need `tokio`, but only the `net` `feature`. Thus `tokio` is an `optional` `dependency` restricted to the `net` `feature` only. 

``` toml
[dependencies]
# [...]
tokio = { version = "1", default-features = false, features = ["net"], optional = true }
```

In addition, the `neqo-udp` tests always need `tokio` and they make use of various features beyond `net`, thus the additional import as a `dev-dependency` with more features enabled.

```
[dev-dependencies]
tokio = { version = "1", default-features = false, features = ["net", "time", "macros", "rt", "rt-multi-thread"] }
```

> The lack of an EOL here is also worth fixing. You can configure your editor to add these to all text files.

Will do!

neqo-udp/src/lib.rs
mxinden: I think finding the right value needs more experimentation.

Here is what others use:

Google's Quiche handles [1472 bytes per segment](https://github.com/google/quiche/blob/36f574836fc8e8c9003ee5b08f037f998372b909/quiche/quic/core/quic_constants.h#L52) with at most [16 segments per read](https://github.com/google/quiche/blob/36f574836fc8e8c9003ee5b08f037f998372b909/quiche/quic/core/quic_packet_reader.h#L21).

MSQuic uses [`(UINT16_MAX - CXPLAT_UDP_HEADER_SIZE)`](https://github.com/microsoft/msquic/blob/8a741dc87331a346d38123b886331b240c06c509/src/platform/datapath_winuser.c#L51-L54).

Quinn uses [a receive buffer of `64 * 1024`](https://github.com/quinn-rs/quinn/blob/c445e9b8ff4026b1b27979f4c9c21127402df12a/quinn/src/endpoint.rs#L696-L701).

Note that in Firefox this buffer is allocated once per socket thread. Given that Firefox has a single socket thread only, there is effectively a single `u16::MAX` receive buffer only.

Would you recommend a larger buffer size to start with @larseggert?

neqo-common/src/log.rs
mxinden: Thanks. Addressed in https://github.com/mozilla/neqo/pull/1920/commits/c7b446de1be992a8d1b85f029386231e2a8926f3.

neqo-udp/src/lib.rs
mxinden: :+1: https://github.com/mozilla/neqo/pull/1920/commits/9dbcbae4550f4c2d6c813fee5b60894c754ce30d

neqo-udp/src/lib.rs
mxinden: :+1: https://github.com/mozilla/neqo/pull/1920/commits/9dbcbae4550f4c2d6c813fee5b60894c754ce30d

neqo-udp/src/lib.rs
mxinden: Good idea. See https://github.com/mozilla/neqo/pull/1920/commits/8040a63b070df83191f03fd0697083ea438a1e21.

neqo-udp/src/lib.rs
mxinden: :+1: https://github.com/mozilla/neqo/pull/1920/commits/9dbcbae4550f4c2d6c813fee5b60894c754ce30d

neqo-udp/src/lib.rs
mxinden: :+1: Done in https://github.com/mozilla/neqo/pull/1920/commits/af648706268ffee40e0206d6358828a511cfe4d3.

neqo-udp/src/lib.rs
mxinden: Sounds good to me. I created https://github.com/mozilla/neqo/pull/1940.

neqo-udp/src/lib.rs
larseggert: We can start with this, but maybe add a `TODO` to capture that we should experiment with what a good size would be.

mxinden: :+1: be05257e

PR #1922, opened 2024-06-13
fix(client): remove needless process_output in process_input

mxinden:
There is no need to call `process_output` within `process_multiple_input` after each GRO datagram batch. Instead, process all available incoming datagrams in `process_multiple_input` and then move on to the top of the `loop`, calling `handler.handle` and then process_output` as usual.

---

Draft for now to see what the benchmarks say.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -399,7 +399,7 @@ impl<'a, H: Handler> Runner<'a, H> {
             }
 
             match ready(self.socket, self.timeout.as_mut()).await? {
-                Ready::Socket => self.process_multiple_input().await?,
+                Ready::Socket => self.process_multiple_input()?,
                 Ready::Timeout => {
                     self.timeout = None;
                 }
@@ -435,7 +435,7 @@ impl<'a, H: Handler> Runner<'a, H> {
         Ok(())
     }
 
-    async fn process_multiple_input(&mut self) -> Res<()> {
+    fn process_multiple_input(&mut self) -> Res<()> {
         loop {
             let dgrams = self.socket.recv(&self.local_addr)?;
             if dgrams.is_empty() {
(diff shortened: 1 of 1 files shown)


mxinden:
Let's wait for a benchmark run before merging here. Just to make sure nothing depends on the additional calls to `process_output`.

larseggert:
@mxinden the RPS bench seems to just sit there idle.

larseggert:
@mxinden anything I can do to help land this?

mxinden:
I suggest we land https://github.com/mozilla/neqo/pull/1929 first. Might resolve the RPS benchmark stall.

larseggert:
Pulled in from `main` w/#1929. Let's see.

mxinden:
> Pulled in from `main` w/#1929. Let's see.

All green. Good news 😮‍💨

mxinden:
> 1-conn/1-100mb-resp (aka. Download)/client: 💔 Performance has regressed.
> 
>        time:   [1.7535 s 1.8771 s 1.9494 s]
>        thrpt:  [51.297 MiB/s 53.273 MiB/s 57.028 MiB/s]
> change:
>        time:   [+52.443% +65.455% +73.749%] (p = 0.00 < 0.05)
>        thrpt:  [-42.446% -39.561% -34.402%]

Let's hold off merging for now.

larseggert:
OK, making this a draft PR for now.

larseggert:
Is this OBE?

mxinden:
Closing here. Benchmarks still show a significant performance impact when not interleaving `process_multiple_input` with `process_output`. While likely not relevant for non-localhost traffic, I don't think the small clean-up here is worth it.

larseggert:
It might be that the RX processing takes so long that the ACK clock stalls, because we don't schedule an ACK to TX.

We might want to schedule a TX when the next ACK is due.

Issue #1917, opened 2024-05-30
`FRAME_TYPE_STREAMS_BLOCKED` ignored?


larseggert:
This code
https://github.com/mozilla/neqo/blob/121fe683ae4b39a5b694f671abfd397cbd9b4322/neqo-transport/src/streams.rs#L203-L207

says that there is no need to trigger sending a `FRAME_TYPE_MAX_STREAMS` when we get an `FRAME_TYPE_STREAMS_BLOCKED`. I wonder if that is actually correct, if the `FRAME_TYPE_MAX_STREAMS` triggered by retiring streams was lost?

The RPS bench failure for #1903 seems to stem from the client timing out while waiting for `FRAME_TYPE_MAX_STREAMS` updates that end up in PMTUD probes that are repeatedly getting dropped. That idle timeout sometimes happens before we'd RTX the `FRAME_TYPE_MAX_STREAMS`.

larseggert:
I'm attaching a client and server log where this seems to happen, and a pcap and TLS key log file: [Archive.zip](https://github.com/user-attachments/files/15509647/Archive.zip)

The server receives a `StreamsBlocked` in pn=29 from the client. The server then sends a `MaxStreams` in pn=28 which ends up in a PMTU probe that is too big and hence is lost. The server gets another `StreamsBlocked` fromt he client in pn=31, but ignores it. There is a PING-ACK exchange, but that doesn't make the server retransmit the `MaxStreams` and hence the client hits the idle timeout.

KershawChang:
> I'm attaching a client and server log where this seems to happen, and a pcap and TLS key log file: [Archive.zip](https://github.com/user-attachments/files/15509647/Archive.zip)
> 
> The server receives a `StreamsBlocked` in pn=29 from the client. The server then sends a `MaxStreams` in pn=28 which ends up in a PMTU probe that is too big and hence is lost. The server gets another `StreamsBlocked` from the client in pn=31, but ignores it. There is a PING-ACK exchange, but that doesn't make the server retransmit the `MaxStreams` and hence the client hits the idle timeout.

I think the problem that the server doesn't retransmit the `MaxStreams` frame is that the previous call to `process` returns `Output::None`. At this point, the server has no timeout timer armed and is just waiting for the client to send something. However, the client is blocked because it's waiting for the `MaxStreams` frame. In the end, the client ends due to the idle timeout.

KershawChang:
> Why is the call to `process` returning `Output::None`?
> 
[...]

PR #1926, opened 2024-06-17
don't clear self.wake_at

KershawChang:
Fix issue #1917 
See the explanation [here](https://github.com/mozilla/neqo/issues/1917#issuecomment-2173192278).

### DIFF
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -669,7 +669,7 @@ impl Server {
                 qtrace!([self], "Send packet: {:?}", d);
                 Output::Datagram(d)
             })
-            .or_else(|| self.wake_at.take().map(|c| Output::Callback(c - now)))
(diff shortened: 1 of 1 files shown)


larseggert:
Thanks for this. I think we should probably add a test for it, too?

larseggert:
When I run a loop of
```sh
while RUST_LOG=neqo_transport=debug  cargo test  --bench main --features bench ;do :; done
```
I still see a crash. FWIW, I tried running `test/test.sh` many times and that did not crash.

```

larseggert:
> When I run a loop of
> 
> ```shell
> while RUST_LOG=neqo_transport=debug cargo test --bench main --features bench ;do :; done
> ```
> 
> I still see a crash. FWIW, I tried running `test/test.sh` many times and that did not crash.

I can't reproduce this crash anymore. FWIW.

PR #1927, opened 2024-06-17
fix(transport/recovery): remove needless `mut`

mxinden:


larseggert (approved):
Interesting that `clippy` didn't find this.

mxinden:
> Interesting that `clippy` didn't find this.

I am surprised as well. 

Tried to figure out where `process_output` is not idempotent (see https://github.com/mozilla/neqo/issues/1917#issuecomment-2173292609). Still looking.

PR #1928, opened 2024-06-17
feat: Allow running parallel transfer benchmarks

larseggert:
By randomizing the port used for a bench. There is still a small chance for a collision, but seems acceptable.

### DIFF
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -20,7 +20,8 @@ fn transfer(c: &mut Criterion) {
     neqo_common::log::init(Some(log::LevelFilter::Off));
     neqo_crypto::init_db(PathBuf::from_str("../test-fixture/db").unwrap()).unwrap();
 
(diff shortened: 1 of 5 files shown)


neqo-bin/benches/main.rs
```diff
@@ -20,7 +20,8 @@ fn transfer(c: &mut Criterion) {
     neqo_common::log::init(Some(log::LevelFilter::Off));
     neqo_crypto::init_db(PathBuf::from_str("../test-fixture/db").unwrap()).unwrap();
 
-    let done_sender = spawn_server();
+    let port = 49151 + (rand::random::<u16>() % 16384);
```
larseggert, line 23: Should we use neqo-crypto here? It doesn't seem to have a clean way to return randomized types, just `u8`?

larseggert:
Closing this. It was useful while debugging a spurious failure, but it's often more/also useful to have predictable ports for tcpdumping.

PR #1929, opened 2024-06-17
refactor(transport/server): always iterate each connection

mxinden:
### Problem

Say a `neqo_transport::Server` is managing a single `neqo_transport::Connection` in `Server::connections`. Assume the following chain of events:

1. A user (e.g. `neqo-http3`) calls `Server::process`.
    https://github.com/mozilla/neqo/blob/6664452e2ba25f028ebf07c404fff3d0193c0ef4/neqo-transport/src/server.rs#L660-L677

    2. `self.wake_at` is `None`.
    3. `dgram` is `None`, thus `self.process_input` is never called.
    4. `self.process_next_output(now)` is called.
        https://github.com/mozilla/neqo/blob/6664452e2ba25f028ebf07c404fff3d0193c0ef4/neqo-transport/src/server.rs#L636-L642
        1. It attains a reference to the one `Connection` through `self.waiting.pop_front()`.
        2. It calls self.process_connection which in turn calls
        `Connection::process`, returning a `Output::Callback(_)`, which is stored in `self.wake_at`.
    5. `self.wake_at.take()` takes the callback and returns it to the user as `Output::Callback`.
6. The user calls `Server::process` again.
    1. `self.wake_at` is `None`. 
    2. `dgram` is `None`, thus `self.process_input` isn't called.
    4. `Server::process` calls `Server::process_next_output`.
        1. `Server::process_next_output` finds no connection reference in `self.waiting` and thus returns `None`.
    4. `self.wake_at.take()` is `None`
    6. `Server::process` returns `None`

Result is that the user received an `Output::None` even though the one `Connection` managed by the `Server` is waiting for a callback. Thus the server stalls.

Shout-out to @KershawChang discovering this in https://github.com/mozilla/neqo/issues/1917#issuecomment-2173408413.

### Solution

The single source of truth of whether a `Connection` needs a callback or not is the `Connection` itself. Instead of duplicating this information in `Server::wake_at`, `Server::waiting`, `ServerConnectionState::wake_at`, always ask the single source of truth, i.e. the `Connection`.

More concretely, with this patch `Server::process` always calls `Connection::process` for each of its `Connection`s. It does not try to be smart on whether a `Connection` needs `process`ing or not.

---

Work-in-progress for now. Fix needs confirming. Also needs to run on benchmark machine to see whether this introduces any regressions.

Should fix https://github.com/mozilla/neqo/issues/1917.
Alternative to https://github.com/mozilla/neqo/pull/1926.
Needed for https://github.com/mozilla/neqo/pull/1903.
Another step after https://github.com/mozilla/neqo/pull/1784 simplifying `neqo_transport::Server`.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -136,6 +136,17 @@ impl Output {
             _ => Duration::new(0, 0),
         }
     }
+
+    #[must_use]
+    pub fn or_else<F>(self, f: F) -> Self
+    where
+        F: FnOnce() -> Self,
+    {
+        match self {
+            x @ Self::Datagram(_) | x @ Self::Callback(_) => x,
+            Self::None => f(),
+        }
+    }
 }
 
 /// Used by inner functions like `Connection::output`.
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -8,7 +8,7 @@
 
 use std::{
     cell::RefCell,
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet},
     fs::OpenOptions,
     mem,
     net::SocketAddr,
@@ -50,21 +50,6 @@ type ConnectionTableRef = Rc<RefCell<HashMap<ConnectionId, StateRef>>>;
 pub struct ServerConnectionState {
     c: Connection,
     active_attempt: Option<AttemptKey>,
-    wake_at: Option<Instant>,
-}
-
-impl ServerConnectionState {
-    fn set_wake_at(&mut self, at: Instant) {
-        self.wake_at = Some(at);
-    }
-
-    fn needs_waking(&self, now: Instant) -> bool {
-        self.wake_at.map_or(false, |t| t <= now)
-    }
-
-    fn woken(&mut self) {
-        self.wake_at = None;
-    }
 }
 
 impl Deref for ServerConnectionState {
@@ -175,10 +160,6 @@ pub struct Server {
     connections: ConnectionTableRef,
     /// The connections that have new events.
     active: HashSet<ActiveConnectionRef>,
-    /// The set of connections that need immediate processing.
-    waiting: VecDeque<StateRef>,
-    /// The latest [`Output::Callback`] returned from [`Server::process`].
-    wake_at: Option<Instant>,
     /// Address validation logic, which determines whether we send a Retry.
     address_validation: Rc<RefCell<AddressValidation>>,
     /// Directory to create qlog traces in
@@ -221,11 +202,9 @@ impl Server {
             active_attempts: HashMap::default(),
             connections: Rc::default(),
             active: HashSet::default(),
-            waiting: VecDeque::default(),
             address_validation: Rc::new(RefCell::new(validation)),
             qlog_dir: None,
             ech_config: None,
-            wake_at: None,
         })
     }
 
@@ -268,23 +247,9 @@ impl Server {
         c: &StateRef,
         dgram: Option<&Datagram>,
         now: Instant,
-    ) -> Option<Datagram> {
+    ) -> Output {
         qtrace!([self], "Process connection {:?}", c);
         let out = c.borrow_mut().process(dgram, now);
-        match out {
-            Output::Datagram(_) => {
-                qtrace!([self], "Sending packet, added to waiting connections");
-                self.waiting.push_back(Rc::clone(c));
-            }
-            Output::Callback(delay) => {
-                let next = now + delay;
-                c.borrow_mut().set_wake_at(next);
-                if self.wake_at.map_or(true, |c| c > next) {
-                    self.wake_at = Some(next);
-                }
-            }
-            Output::None => {}
-        }
         if c.borrow().has_events() {
             qtrace!([self], "Connection active: {:?}", c);
             self.active.insert(ActiveConnectionRef { c: Rc::clone(c) });
@@ -303,7 +268,7 @@ impl Server {
                 .borrow_mut()
                 .retain(|_, v| !Rc::ptr_eq(v, c));
         }
-        out.dgram()
+        out
     }
 
     fn connection(&self, cid: ConnectionIdRef) -> Option<StateRef> {
@@ -315,14 +280,14 @@ impl Server {
         initial: InitialDetails,
         dgram: &Datagram,
         now: Instant,
-    ) -> Option<Datagram> {
+    ) -> Output {
         qdebug!([self], "Handle initial");
         let res = self
             .address_validation
             .borrow()
             .validate(&initial.token, dgram.source(), now);
         match res {
-            AddressValidationResult::Invalid => None,
+            AddressValidationResult::Invalid => Output::None,
             AddressValidationResult::Pass => self.connection_attempt(initial, dgram, None, now),
             AddressValidationResult::ValidRetry(orig_dcid) => {
                 self.connection_attempt(initial, dgram, Some(orig_dcid), now)
@@ -337,7 +302,7 @@ impl Server {
                 );
                 let Ok(token) = res else {
                     qerror!([self], "unable to generate token, dropping packet");
-                    return None;
+                    return Output::None;
                 };
                 if let Some(new_dcid) = self.cid_generator.borrow_mut().generate_cid() {
                     let packet = PacketBuilder::retry(
@@ -355,14 +320,14 @@ impl Server {
                             dgram.ttl(),
                             p,
                         );
-                        Some(retry)
+                        Output::Datagram(retry)
                     } else {
                         qerror!([self], "unable to encode retry, dropping packet");
-                        None
+                        Output::None
                     }
                 } else {
                     qerror!([self], "no connection ID for retry, dropping packet");
-                    None
+                    Output::None
                 }
             }
         }
@@ -374,7 +339,7 @@ impl Server {
         dgram: &Datagram,
         orig_dcid: Option<ConnectionId>,
         now: Instant,
-    ) -> Option<Datagram> {
+    ) -> Output {
         let attempt_key = AttemptKey {
             remote_address: dgram.source(),
             odcid: orig_dcid.as_ref().unwrap_or(&initial.dst_cid).clone(),
@@ -475,7 +440,7 @@ impl Server {
         dgram: &Datagram,
         orig_dcid: Option<ConnectionId>,
         now: Instant,
-    ) -> Option<Datagram> {
+    ) -> Output {
         qinfo!([self], "Accept connection {:?}", attempt_key);
         // The internal connection ID manager that we use is not used directly.
         // Instead, wrap it so that we can save connection IDs.
@@ -501,7 +466,6 @@ impl Server {
                 self.setup_connection(&mut c, &attempt_key, initial, orig_dcid);
                 let c = Rc::new(RefCell::new(ServerConnectionState {
                     c,
-                    wake_at: None,
                     active_attempt: Some(attempt_key.clone()),
                 }));
                 cid_mgr.borrow_mut().set_connection(&c);
@@ -518,7 +482,7 @@ impl Server {
                         initial.version.wire_version(),
                     );
                 }
-                None
+                Output::None
             }
         }
     }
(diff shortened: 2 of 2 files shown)


larseggert:
FYI #1926 doesn't fix the issue in all cases. I can still make it timeout if I run `test/test.sh` many times (many more than before, so it definitely fixes some cases.) I also need to reduce the logging and run it in release mode; it seems that debug mode and extra logging create delays that make it succeed.

larseggert:
I think we might have a winner! I'm running this in a loop and I see no more client timeouts.
```shell
while RUST_LOG=neqo_transport=warn  cargo test --release  --bench main --features bench RPS; do :; done
```

larseggert:
@mxinden anything I can do to help land this?

mxinden:
I have another simplification locally that I will push later today. That will resolve the outstanding TODO, namely the allocation into a `Vec` in the hot path. I will prioritize this pull request today.

neqo-transport/src/server.rs
```diff
@@ -620,71 +561,68 @@ impl Server {
...
 
     /// Whether any connections have received new events as a result of calling
     /// `process()`.
     #[must_use]
     pub fn has_active_connections(&self) -> bool {
-        !self.active.is_empty()
-    }
-
-    pub fn add_to_waiting(&mut self, c: &ActiveConnectionRef) {
-        self.waiting.push_back(c.connection());
```
mxinden, line 687: Given that now `Server` always `process`es each `Connection`, no need for users to tell `Server` that a connection might be waiting for processing.

neqo-http3/src/server.rs
```diff
@@ -131,27 +131,14 @@ impl Http3Server {
...
-                } else {
-                    None
-                }
-            })
-            .cloned()
-            .collect();
-        // For http_active connection we need to put them in neqo-transport's server
-        // waiting queue.
-        active_conns.append(&mut http3_active);
-        active_conns.dedup();
```
mxinden, line 151: `self.server.active_connections()` returns a `HashSet`. Thus no need for this complex de-duplication within a `Vec`.

neqo-bin/src/server/http09.rs
```diff
@@ -190,7 +190,6 @@ impl HttpServer {
                         .unwrap();
                     qdebug!("Wrote {}", sent);
                     *offset += sent;
-                    self.server.add_to_waiting(conn);
```
mxinden, line 193: No need to tell `Server` that a `Connection` might be waiting to be processed. `Server` will simply always `process` each `Connection`.

neqo-transport/src/server.rs
```diff
@@ -167,18 +166,8 @@ pub struct Server {
...
     /// Connection parameters.
     conn_params: ConnectionParameters,
-    /// Active connection attempts, keyed by `AttemptKey`.  Initial packets with
-    /// the same key are routed to the connection that was first accepted.
-    /// This is cleared out when the connection is closed or established.
-    active_attempts: HashMap<AttemptKey, StateRef>,
     /// All connections, keyed by `ConnectionId`.
     connections: ConnectionTableRef,
-    /// The connections that have new events.
-    active: HashSet<ActiveConnectionRef>,
```
mxinden, line 177: Instead of keeping track which `Connection` has new events, ask the `Connection`s ad-hoc whenever the information is needed.

neqo-transport/tests/common/mod.rs
```diff
@@ -48,7 +48,6 @@ pub fn connected_server(server: &mut Server) -> ActiveConnectionRef {
         .iter()
         .filter(|c: &&ActiveConnectionRef| *c.borrow().state() == State::Confirmed);
     let c = confirmed.next().expect("one confirmed");
-    assert!(confirmed.next().is_none(), "only one confirmed");
```
mxinden, line 51: The unit test `different_initials_same_path` has two clients connect to the same server using this helper function (i.e. `connected_server`). The second connection attempt will fail, as this helper function asserts a single confirmed connection only.

neqo-transport/tests/server.rs
```diff
@@ -210,15 +210,14 @@ fn same_initial_after_connected() {
     let server_initial = server.process(client_initial.as_dgram_ref(), now()).dgram();
     assert!(server_initial.is_some());
     complete_connection(&mut client, &mut server, server_initial);
-    // This removes the connection from the active set until something happens to it.
-    assert_eq!(server.active_connections().len(), 0);
+    assert_eq!(server.active_connections().len(), 1);
```
mxinden, line 213: Previously the `Server` would keep track in `Server::active` whether any connection has events to process. On a call to `Server::active_connections` it would remove this state, whether the caller of `active_connections` actually consumed the events or not.

With this pull request, `Server` no longer keeps this additional state, i.e. `Sever::active` is removed. Instead of returning the cached information on `Server::active_connections`, the server simply asks the `Connection`s itself, whether they have events to process.

Given that this test consumes no events, the expected number of active connections is now `1` and not, as previously wrongly returned,  `0`.

neqo-transport/tests/server.rs
```diff
@@ -210,15 +210,14 @@ fn same_initial_after_connected() {
...
 
     // Now make a new connection using the exact same initial as before.
     // The server should respond to an attempt to connect with the same Initial.
     let dgram = server.process(client_initial.as_dgram_ref(), now()).dgram();
     assert!(dgram.is_some());
     // The server should make a new connection object.
-    assert_eq!(server.active_connections().len(), 1);
+    assert_eq!(server.active_connections().len(), 2);
```
mxinden, line 220: Same as above, just now, given the additional connection, `2` instead of `1`.

neqo-transport/tests/server.rs
```diff
@@ -273,7 +272,7 @@ fn zero_rtt() {
     let t = server.process(None, now).callback();
     now += t;
     assert_eq!(server.process(None, now), Output::None);
-    assert_eq!(server.active_connections().len(), 1);
+    assert_eq!(server.active_connections().len(), 0);
```
mxinden, line 275: Previously the `Server` would cache the active connections in `Server::active`. It would only purge this cache on a call to `Server::active_connections`, even though the connection in this case is already closed in `generate_ticket`.

Given that this pull request removes the state tracking (i.e. `Server::active`), `server.active_connections` now correctly returns `0`.

neqo-transport/tests/server.rs
```diff
@@ -310,7 +309,17 @@ fn zero_rtt() {
...
+        active
+            .iter()
+            .next()
+            .unwrap()
+            .borrow()
+            .stats()
+            .frame_rx
+            .stream,
+        2
+    );
```
mxinden, line 322: `active_connections()` now returns a `HashSet` instead of a `Vec`. Thus `active.iter().first().unwrap()` instead of `active[0]`.

neqo-transport/src/connection/mod.rs
```diff
@@ -136,6 +136,17 @@ impl Output {
...
+    #[must_use]
+    pub fn or_else<F>(self, f: F) -> Self
+    where
+        F: FnOnce() -> Self,
+    {
+        match self {
+            x @ (Self::Datagram(_) | Self::Callback(_)) => x,
+            Self::None => f(),
+        }
+    }
```
mxinden, line 149: Port of `Option::or_else` for `Output`.

mxinden:
Thus far I am unable to reproduce the `HPS` (handshakes per second) benchmark regression locally:

```
➜  neqo-bin git:(refactor-callback) ✗ critcmp main pr  -f "HPS"
group                                 pr                                    main
-----                                 ------                                 ----
1-conn/1-1b-resp (aka. HPS)/client    1.00     26.5±2.99ms    37 Elem/sec    1.04     27.6±3.53ms    36 Elem/sec

File: neqo-transport/src/server.rs [hunk 11]
```diff
@@ -620,71 +561,68 @@ impl Server {
...
-                Output::None
-            })
+        out
     }
 
     /// This lists the connections that have received new events
     /// as a result of calling `process()`.
-    pub fn active_connections(&mut self) -> Vec<ActiveConnectionRef> {
-        mem::take(&mut self.active).into_iter().collect()
+    pub fn active_connections(&mut self) -> HashSet<ActiveConnectionRef> {
```
martinthomson, line 609: Should this return `impl Iterator<Item = ActiveConnectionRef>` instead?  That way we save on the allocation.  If code is doing - as you do in tests - very little with the iterator, then this won't be a problem.  But they can call `.collect()` if they are looking to compare notes or anything like that.

neqo-transport/src/server.rs
```diff
@@ -620,71 +561,68 @@ impl Server {
...
-        }
+        let out = dgram
+            .map_or(Output::None, |d| self.process_input(d, now))
+            .or_else(|| self.process_next_output(now));
+
+        // Clean-up closed connections.
+        self.connections.borrow_mut().retain(|_, c| {
+            if matches!(c.borrow().state(), State::Closed(_)) {
+                // TODO: Is this still needed?
+                c.borrow_mut().set_qlog(NeqoQlog::disabled());
```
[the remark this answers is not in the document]
martinthomson: This would only be necessary if the connection is somehow kept alive through a hanging reference somewhere.  (This being the essential problem with using ref counting.)  All this does is propagate a new qlog reference through the various components, so as long as those are dropped, forcing the reference to be dropped isn't necessary.  You can probably go with `.retain(|_, c| !matches!(c.borrow().state(), State::Closed(_)))`

martinthomson (approved):
Lots of red.  That's good.

larseggert:
> [neqo-transport/src/server.rs](https://app.codecov.io/gh/mozilla/neqo/pull/1929?src=pr&el=tree&filepath=neqo-transport%2Fsrc%2Fserver.rs&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mozilla#diff-bmVxby10cmFuc3BvcnQvc3JjL3NlcnZlci5ycw==) 	91.13% 	[7 Missing ⚠️](https://app.codecov.io/gh/mozilla/neqo/pull/1929?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mozilla)

Could we bump the coverage up?

larseggert (approved):

neqo-transport/src/server.rs
mxinden: `self.connections` can contain multiple references to the same `Connection`. `ActiveConnectionRef` `Hash` implementation makes sure two references to the same `Connection` have the same hash:

https://github.com/mozilla/neqo/blob/ea542735006a1075f97a6984d86029d9b5126d6a/neqo-transport/src/server.rs#L712-L723

The `collect()` into a `HashSet` serves the purpose of de-duplication here.

Note that this pull request does not introduce a new allocation per `active_connections` call. The previous implementation also allocated into a `HashSet`, see `self.active` being set back to `HashSet::default` in the above `mem::take`.

We could still return an `impl Iterator` from the `HashSet`. That said, `neqo-http3` calling `active_connections` again needs a `HashSet`.

Does that make sense @martinthomson?

martinthomson: Ah, I see.  That's fine.  Maybe a comment to that effect would help (I wouldn't want to duplicate that comment though, so ignore at your discretion).

neqo-transport/src/server.rs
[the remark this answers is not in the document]
mxinden: :+1: thank you! 76227d3f

mxinden:
> > [neqo-transport/src/server.rs](https://app.codecov.io/gh/mozilla/neqo/pull/1929?src=pr&el=tree&filepath=neqo-transport%2Fsrc%2Fserver.rs&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mozilla#diff-bmVxby10cmFuc3BvcnQvc3JjL3NlcnZlci5ycw==) 	91.13% 	[7 Missing ⚠️](https://app.codecov.io/gh/mozilla/neqo/pull/1929?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mozilla)
> 
> Could we bump the coverage up?

Note that this pull request does not introduce any untested lines. In other words, each untested line in this patch has been untested before. See coverage report.

That said, the more tests the better! Though I suggest doing so in a separate pull request. Any specific code-paths you have in mind @larseggert?

larseggert:
> Note that this pull request does not introduce any untested lines. In other words, each untested line in this patch has been untested before. See coverage report.

Hm. Maybe I misread the coverage report, or lines were moved around. Anyway, WFM. Good to go.

PR #1953, opened 2024-07-02
ci: Fix bench comment formatting regex

larseggert:
So the `<details>` stuff works properly. Hopefully.


KershawChang (approved):
Does this PR work?

larseggert:
I tested it locally in a shell, and there it did.

Issue #1975, opened 2024-07-11
`BorrowMutError` in `neqo-transport/src/server.rs`


larseggert:
[edit] This seems to be the cause of quite a few interop failures.

In https://interop.seemann.io/logs/2024-07-11T01:40/neqo_picoquic/resumption/output.txt
```
server  | thread 'main' panicked at neqo-transport/src/server.rs:685:26:
server  | already borrowed: BorrowMutError
server  | stack backtrace:
server  |    0: rust_begin_unwind
server  |    1: core::panicking::panic_fmt
server  |    2: core::cell::panic_already_borrowed
server  |    3: neqo_transport::server::ServerConnectionIdGenerator::insert_cid
server  |    4: <neqo_transport::server::ServerConnectionIdGenerator as neqo_transport::cid::ConnectionIdGenerator>::generate_cid
server  |    5: neqo_transport::connection::Connection::output_path
server  |    6: neqo_transport::connection::Connection::process_output
server  |    7: neqo_transport::server::ServerConnectionState::process
server  |    8: neqo_transport::server::Server::process
server  |    9: <neqo_bin::server::http09::HttpServer as neqo_bin::server::HttpServer>::process
server  |   10: neqo_bin::server::ServerRunner::process::{{closure}}
server  |   11: neqo_server::main
```

larseggert:
Same in https://interop.seemann.io/logs/2024-07-11T01:40/neqo_quic-go/handshakeloss/output.txt
```
server  | already borrowed: BorrowMutError
server  | stack backtrace:
server  |    0: rust_begin_unwind
server  |    1: core::panicking::panic_fmt
server  |    2: core::cell::panic_already_borrowed
server  |    3: neqo_transport::server::ServerConnectionIdGenerator::insert_cid
server  |    4: <neqo_transport::server::ServerConnectionIdGenerator as neqo_transport::cid::ConnectionIdGenerator>::generate_cid
server  |    5: neqo_transport::connection::Connection::output_path
server  |    6: neqo_transport::connection::Connection::process_output
server  |    7: neqo_transport::server::ServerConnectionState::process
server  |    8: neqo_transport::server::Server::process
server  |    9: <neqo_bin::server::http09::HttpServer as neqo_bin::server::HttpServer>::process
server  |   10: neqo_bin::server::ServerRunner::process::{{closure}}
server  |   11: neqo_server::main
```

larseggert:
[...]

PR #1976, opened 2024-07-11
ci: Fail QNS run on Rust panic

larseggert:
So we hopefully catch interop panics in CI earlier.


KershawChang (approved):

larseggert:
This needs to wait until #1975 is fixed, otherwise all CI will fail.

.github/actions/quic-interop-runner/action.yml
```diff
@@ -72,6 +72,15 @@ runs:
         # Don't fail CI if the interop test fails
```
mxinden, line 72: Do I understand correctly that long-term we would want to fail CI?

mxinden (approved):
Thanks :+1:

.github/actions/quic-interop-runner/action.yml
larseggert: Yes, once we get to the point that we consistently pass all tests (and all other stacks do to, when interop'ing with us).

mxinden: > (and all other stacks do to, when interop'ing with us).

We might want to pin their Docker images, such that a bug on their end doesn't break our CI. Not critical, thus I suggest keeping as is until this happens for the first time.

larseggert:
The single remaining failure is [Run QNS (neqo-latest vs. neqo)](https://github.com/mozilla/neqo/actions/runs/9979399509/job/27578579516#logs), which is because mainline neqo doesn't have #1981 yet.

PR #1977, opened 2024-07-11
ci: Try to use NSS package on Windows

larseggert:


larseggert:
This doesn't work. The only prebuilt NSS library is available from MSYS2, which uses the non-default GNU toolchain which has flaky Rust support. And I can't figure out how to link an MSYS2 library into an MSVC Rust build.

PR #1978, opened 2024-07-12
Use a stack allocation for header protection

martinthomson:
The use of `Vec` here is unnecessary.  We can use a fixed sized array instead.  It means that we'll not be generating 64 bytes of data for ChaCha20 header protection, but that's also a net win.

The performance gain here is negligible, but the code becomes cleaner, so I consider that a win.


larseggert (approved):
@KershawChang we should make this part of 0.8.0.

mxinden (approved):

mxinden:
> @KershawChang we should make this part of 0.8.0.

`v0.8.0` has [already been tagged](https://github.com/mozilla/neqo/releases/tag/v0.8.0). I would prefer treating git tags as immutable, i.e. not alter them once pushed.

larseggert:
OK, 0.8.1 then :-) And let's not vendor in 0.8.0.

KershawChang:
> OK, 0.8.1 then :-) And let's not vendor in 0.8.0.

Is this urgent? Can we wait until issue #1975 is fixed?

larseggert:
> Is this urgent? Can we wait until issue #1975 is fixed?

We can wait until #1975.

PR #1981, opened 2024-07-12
refactor(server): remove ServerConnectionIdGenerator

mxinden:
A `neqo_transport::Server` manages multiple `neqo_transport::Connection`s. A `Server` keeps a `HashMap` of all `Connection`s, indexed by `ConnectionId`.

Indexing by `ConnectionId` is difficult as:

- The `ConnectionId` is not known before constructing the connection.
- The `ConnectionId` might change throughout the lifetime of a connection.

Previously the index was kept up-to-date through a wrapper around the `ConnectionIdGenerator` provided to a `Connection` by a `Server`. This wrapper would be provided a shared reference to the `Server`s `HashMap` of `Connection`s. Whenever the `Server` would `process` a `Connection` and that `Connection` would generate a new `ConnectionId` via `ConnectionIdGenerator::generate_cid`, the wrapped `ConnectionIdGenerator` would also insert the `Connection` keyed by the new `ConnectionId` into the `Server`'s `HashMap` of `Connection`s.

This has two problems:

- Complexity through indirection where a `Connection` owned by a `Server` can indirectly mutate the `Server` through the provided wrapped `ConnectionIdGenerator` having access to the `Server`'s set of `Connection`s.

- Double `RefCell` borrow e.g. when a `Server` would iterate its `Connection`s, process each, where one of them might generate a new `ConnectionId` via the provided `ConnectionIdGenerator`, which in turn would insert the `Connection` into the currently iterated set of `Connection`s of the `Server`.

This commit replaces the `HashMap<ConnectionId, Rc<RefCell<Connection>>>` of the `Server` with a simple `Vec<Connection>`. Given the removal of the index, the `ConnectionIdGenerator` wrapper (i.e. `ServerConnectionIdGenerator`) is no longer needed and removed and thus the shared reference to the `Server`s `Connection` `HashMap` is no longer needed and thus the above mentioned double borrow is made impossible.

Downside to the removal of the index by `ConnectionId` is a potential performance hit. When the `Server` manages a large set of `Connection`s, finding the `Connection` corresponding to a `ConnectionId` (e.g. from an incoming `Datagram`) is `O(n)` (`n` equal the number of connections).

---

Fixes https://github.com/mozilla/neqo/issues/1975.

Simplifies `neqo_transport::Server` once more.
image

Draft for now to test for performance regressions first.

### DIFF
--- a/neqo-bin/src/server/http09.rs
+++ b/neqo-bin/src/server/http09.rs
@@ -10,7 +10,7 @@ use neqo_common::{event::Provider, hex, qdebug, qerror, qinfo, qwarn, Datagram};
 use neqo_crypto::{generate_ech_keys, random, AllowZeroRtt, AntiReplay};
 use neqo_http3::Error;
 use neqo_transport::{
-    server::{ActiveConnectionRef, Server, ValidateAddress},
+    server::{ConnectionRef, Server, ValidateAddress},
     ConnectionEvent, ConnectionIdGenerator, Output, State, StreamId,
 };
 use regex::Regex;
@@ -75,7 +75,7 @@ impl HttpServer {
         })
     }
 
-    fn save_partial(&mut self, stream_id: StreamId, partial: Vec<u8>, conn: &ActiveConnectionRef) {
+    fn save_partial(&mut self, stream_id: StreamId, partial: Vec<u8>, conn: &ConnectionRef) {
         let url_dbg = String::from_utf8(partial.clone())
             .unwrap_or_else(|_| format!("<invalid UTF-8: {}>", hex(&partial)));
         if partial.len() < 4096 {
@@ -87,7 +87,7 @@ impl HttpServer {
         }
     }
 
-    fn write(&mut self, stream_id: StreamId, data: Option<Vec<u8>>, conn: &ActiveConnectionRef) {
+    fn write(&mut self, stream_id: StreamId, data: Option<Vec<u8>>, conn: &ConnectionRef) {
         let resp = data.unwrap_or_else(|| Vec::from(&b"404 That request was nonsense\r\n"[..]));
         if let Some(stream_state) = self.write_state.get_mut(&stream_id) {
             match stream_state.data_to_send {
@@ -110,7 +110,7 @@ impl HttpServer {
         }
     }
 
-    fn stream_readable(&mut self, stream_id: StreamId, conn: &ActiveConnectionRef) {
+    fn stream_readable(&mut self, stream_id: StreamId, conn: &ConnectionRef) {
         if !stream_id.is_client_initiated() || !stream_id.is_bidi() {
             qdebug!("Stream {} not client-initiated bidi, ignoring", stream_id);
             return;
@@ -166,7 +166,7 @@ impl HttpServer {
         self.write(stream_id, resp, conn);
     }
 
-    fn stream_writable(&mut self, stream_id: StreamId, conn: &ActiveConnectionRef) {
+    fn stream_writable(&mut self, stream_id: StreamId, conn: &ConnectionRef) {
         match self.write_state.get_mut(&stream_id) {
             None => {
                 qwarn!("Unknown stream {stream_id}, ignoring event");
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -987,6 +987,12 @@ impl Connection {
         }
     }
 
+    /// Whether the given [`ConnectionIdRef`] is a valid local [`ConnectionId`].
+    #[must_use]
+    pub fn is_valid_local_cid(&self, cid: ConnectionIdRef) -> bool {
+        self.cid_manager.is_valid(cid)
+    }
+
     /// Process new input datagrams on the connection.
     pub fn process_input(&mut self, d: &Datagram, now: Instant) {
         self.process_multiple_input(iter::once(d), now);
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -11,7 +11,6 @@ use std::{
     cmp::min,
     collections::HashSet,
     fs::OpenOptions,
-    net::SocketAddr,
     ops::{Deref, DerefMut},
     path::PathBuf,
     rc::Rc,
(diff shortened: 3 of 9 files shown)


neqo-bin/src/server/http09.rs
```diff
@@ -10,7 +10,7 @@ use neqo_common::{event::Provider, hex, qdebug, qerror, qinfo, qwarn, Datagram};
 use neqo_crypto::{generate_ech_keys, random, AllowZeroRtt, AntiReplay};
 use neqo_http3::Error;
 use neqo_transport::{
-    server::{ActiveConnectionRef, Server, ValidateAddress},
+    server::{ConnectionRef, Server, ValidateAddress},
```
mxinden, line 13: As an aside, this pull request renames `ActiveConnectionRef` to `ConnectionRef`. I don't see why the `Active` prefix is needed. In other words, why would the connection not be active? Not a strong opinion. Happy to revert.

neqo-transport/src/connection/mod.rs
```diff
@@ -987,6 +987,12 @@ impl Connection {
         }
     }
 
+    /// Whether the given [`ConnectionIdRef`] is a valid local [`ConnectionId`].
+    #[must_use]
+    pub fn is_valid_local_cid(&self, cid: ConnectionIdRef) -> bool {
+        self.cid_manager.is_valid(cid)
+    }
```
mxinden, line 994: Instead of the indirection, sniffing new `ConnectionId`s through the provided `ConnectionIdGenerator` wrapper, the `Server` now explicitly asks the `Connection` whether a `ConnectionId` is a valid id of the connection.

neqo-transport/tests/retry.rs
```diff
@@ -31,7 +31,7 @@ use test_fixture::{
 
 #[test]
 fn retry_basic() {
-    let server = default_server();
+    let mut server = default_server();
```
mxinden, line 34: `Server` is now the sole owner of its `Server::connections` collection. Just a plain `Vec`, no more interior mutability via `Rc<RefCell<_>>`. Thus methods of `Server` need mutable access to `self` in order to mutate `Server::connections`, i.e. in order to process the `Connection`s.

larseggert:
You might be able to mitigate some of the performance hit by using an LRU queue for this, under the assumption that the hot set of connections is a subset of the total.

neqo-transport/tests/server.rs
```diff
@@ -160,20 +161,20 @@ fn duplicate_initial_new_path() {
...
 
-    assert_eq!(server.active_connections().len(), 2);
-    complete_connection(&mut client, &server, server_initial);
+    // The server should ignore a packet with the same destination connection ID.
+    let dgram = server.process(Some(&other), now()).dgram();
+    assert!(dgram.is_none());
+
+    assert_eq!(server.active_connections().len(), 1);
+    complete_connection(&mut client, &mut server, server_initial);
 }
```
mxinden, line 173: I believe a server SHOULD  ignore the second packet, given that it carries the same destination connection ID and packet number. In other words, I don't think the previous assertion of the test is correct.

The previous assertion (i.e. _should not ignore_) now fails with this pull request, as connections are no longer lazily indexed through the `ConnectionIdGenerator` wrapper, but instead looked-up ad-hoc on each packet.

Calling this out explicitly to make sure I am not missing something.

larseggert:
> You might be able to mitigate some of the performance hit by using an LRU queue for this, under the assumption that the hot set of connections is a subset of the total.

Or you could swap an element in the Vec with its left neighbor on every access, bubbling frequently-accessed elements to the front over time.

mxinden:
> > You might be able to mitigate some of the performance hit by using an LRU queue for this, under the assumption that the hot set of connections is a subset of the total.
> 
> Or you could swap an element in the Vec with its left neighbor on every access, bubbling frequently-accessed elements to the front over time.

According to the latest benchmark run, this pull request does not have a performance impact. Given that the Neqo server implementation is for testing and benchmarking only, do we need any of the above optimizations? Given that they come with (minimal) complexity, I would prefer introducing them only once needed, i.e. when we have a benchmark showing their impact. Thoughts @larseggert?

neqo-transport/src/server.rs
```diff
@@ -317,20 +301,22 @@ impl Server {
...
+            return self.accept_connection(&attempt_key, initial, dgram, orig_dcid, now);
+        };
+
+        let out = connection.process(Some(dgram), now);
+        qdebug!(
+            [self],
+            "Handled Initial for existing connection attempt {:?}",
+            attempt_key
+        );
+        out
```
larseggert, line 319: We might want to add a test for this error case.

neqo-transport/src/server.rs
```diff
@@ -317,20 +301,22 @@ impl Server {
...
-                );
-                c.borrow_mut().process(Some(dgram), now)
-            },
-        )
+
+        let Some(connection) = self
+            .connections
+            .iter_mut()
+            .find(|c| c.active_attempt.as_ref() == Some(&attempt_key))
+        else {
```
larseggert, line 309: This pattern also happens below; could use a `fn`?

neqo-transport/src/server.rs
```diff
@@ -500,8 +480,12 @@ impl Server {
         };
 
         // Finding an existing connection. Should be the most common case.
-        if let Some(c) = self.connection(packet.dcid()) {
-            return c.borrow_mut().process(Some(dgram), now);
+        if let Some(c) = self
+            .connections
+            .iter_mut()
+            .find(|c| c.borrow().is_valid_local_cid(packet.dcid()))
+        {
```
larseggert, line 487: Pattern mentioned above.

neqo-transport/tests/server.rs
larseggert: Yeah, I don't understand either why the server should respond to both. https://datatracker.ietf.org/doc/html/rfc9000#section-21.2 says
> Addresses cannot change during the handshake, so endpoints can discard packets that are received on a different network path.

@martinthomson may have an opinion.

larseggert (approved):
LGTM with nits.

neqo-transport/src/server.rs
mxinden: I looked into this. With the above change, [namely that two connection attempts with the same original destination connection id but different paths, result in a single connection](https://github.com/mozilla/neqo/pull/1981#discussion_r1676370725), I believe this code path is dead.

A second initial (i.e. an initial for an existing connection attempt) will be passed to the existing connection in `Server::process_input` right away:

https://github.com/mozilla/neqo/blob/8b81a090384076333f4c5dd54b2983d5da292daa/neqo-transport/src/server.rs#L502-L505

I believe this makes the entire concept of `AttemptKey` redundant. Am I missing something?

https://github.com/mozilla/neqo/blob/8b81a090384076333f4c5dd54b2983d5da292daa/neqo-transport/src/server.rs#L82-L90

Thanks for raising this @larseggert!

larseggert: If it's dead, remove and maybe add a `debug_assert`?

mxinden: :+1: done in https://github.com/mozilla/neqo/pull/1981/commits/05442f58e046edffaa8f5f8a999570cf969ca0af.

larseggert:
@KershawChang or @martinthomson, could you take a look? This would unblock vendoring in a new neqo release into Gecko.

KershawChang (approved):
Looks good.
I've also pushed this to try server.
https://treeherder.mozilla.org/jobs?repo=try&revision=1ef0d1a461cb9a381b8279da79b8e08dea64ab08&selectedTaskRun=VOdG5FRbQ-28ReByKNjk4w.0

mxinden:
> I've also pushed this to try server.
> https://treeherder.mozilla.org/jobs?repo=try&revision=1ef0d1a461cb9a381b8279da79b8e08dea64ab08&selectedTaskRun=VOdG5FRbQ-28ReByKNjk4w.0

Thank you! I see at least one potentially related test failure:

```
[task 2024-07-17T14:36:20.766Z] 14:36:20  WARNING -  TEST-UNEXPECTED-TIMEOUT | netwerk/test/unit/test_altsvc_http3.js | Test timed out
```

neqo-transport/tests/server.rs
larseggert: Marking this as resolved. We can revisit this in case @martinthomson thinks the old behavior was correct.

PR #1984, opened 2024-07-16
ci: Trigger secondary checks off `check.yml`

larseggert:
Might save some trees.


larseggert:
Triggers only work when they are on the default branch. Sigh. Merging now. YOLO!

PR #1992, opened 2024-07-18
docs(README): explain connecting with Firefox to local neqo-server

mxinden:
~~Draft for now. Depends on upcoming change in mozilla-central, having Firefox not fall-back to h2 when h3 fails due to self-signed certificate.~~

Edit: https://phabricator.services.mozilla.com/D217798 landed.


mxinden:
Depends on https://phabricator.services.mozilla.com/D217798.

larseggert (approved):

PR #1996, opened 2024-07-20
fix: Handle H3 grease frame types

### LINKED ISSUE #1991: `http3` QNS test failing with quiche
For example, https://github.com/mozilla/neqo/actions/runs/9984006511/job/27593040941#step:6:708
```
server  | 4s622ms TRACE Decoded uint 4012674743313439673
server  | 4s622ms INFO [Http3 server connection] Connection error: HTTP/3 error: HttpFrame.
```
The first line comes from https://github.com/mozilla/neqo/blob/9e3ff0d95dd6ee06b70c45511015c59a4e3e47cb/neqo-http3/src/stream_type_reader.rs#L131

The error comes from https://github.com/mozilla/neqo/blob/9e3ff0d95dd6ee06b70c45511015c59a4e3e47cb/neqo-http3/src/stream_type_reader.rs#L144

I.e., from https://github.com/mozilla/neqo/blob/9e3ff0d95dd6ee06b70c45511015c59a4e3e47cb/neqo-http3/src/stream_type_reader.rs#L55

Based on the value `4012674743313439673`, it really does look though like the `self.read` call in https://github.com/mozilla/neqo/blob/9e3ff0d95dd6ee06b70c45511015c59a4e3e47cb/neqo-http3/src/stream_type_reader.rs#L122
returns garbage?

larseggert:
The old code had the baked-in assumption that an H3 stream would always start with a HEADERS frame. That breaks when GREASE and other unknown frames precede the HEADERS frame.

Due to the odd way we determine the type of a stream, this PR remembers what the first frame type was that we parsed for an HTTP stream, rather than assuming it must have been HEADERS.

Fixes #1991


neqo-http3/src/stream_type_reader.rs
```diff
@@ -52,7 +53,16 @@ impl NewStreamType {
...
-            (_, StreamType::BiDi, Role::Server) => Err(Error::HttpFrame),
+            (stream_type, StreamType::BiDi, Role::Server) => {
+                let Some(st) = stream_type.checked_sub(0x21) else {
+                    return Err(Error::HttpFrame);
+                };
+                if st % 0x1f == 0 {
+                    Ok(Some(Self::Grease(stream_type)))
+                } else {
+                    Err(Error::HttpFrame)
+                }
```
martinthomson, line 64: As we discussed, this is not correct.

For all BiDi streams, as long as the type is not WEBTRANSPORT_STREAM, then we should be treating this as OK.  ... unless the frame type is one we understand other than HEADERS.

This will unnecessarily catch a new extension frame type that is added ahead of HEADERS.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -28,6 +28,7 @@ pub enum NewStreamType {
     WebTransportStream(u64),
     Http,
     Unknown,
+    Grease(u64),
```
martinthomson, line 31: We don't need to identify GREASE on the way in, so I don't think that we need this.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -504,6 +514,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn decode_grease_stream() {
```
martinthomson, line 518: I think that it would be better to have a higher-level test here.  Is it possible to add the ability to grease request streams?  That shouldn't be particularly hard.  I took a quick look and `SendMessage::encode()` seems like it could be tweaked to send a grease frame or two.

martinthomson (changes requested):

neqo-http3/src/stream_type_reader.rs
```diff
@@ -49,8 +49,17 @@ impl NewStreamType {
             (HTTP3_UNI_STREAM_TYPE_PUSH, StreamType::UniDi, Role::Client)
             | (WEBTRANSPORT_UNI_STREAM, StreamType::UniDi, _)
             | (WEBTRANSPORT_STREAM, StreamType::BiDi, _) => Ok(None),
-            (H3_FRAME_TYPE_HEADERS, StreamType::BiDi, Role::Server) => Ok(Some(Self::Http)),
-            (_, StreamType::BiDi, Role::Server) => Err(Error::HttpFrame),
+            (_, StreamType::BiDi, Role::Server) => {
+                if stream_type == HTTP3_UNI_STREAM_TYPE_CONTROL
+                    || stream_type == QPACK_UNI_STREAM_TYPE_ENCODER
+                    || stream_type == QPACK_UNI_STREAM_TYPE_DECODER
+                    || stream_type == WEBTRANSPORT_UNI_STREAM
```
martinthomson, line 56: I think that this is still incorrect, because it's a very tricky piece of code to get right.

In all of these cases, we are reading the first varint from the stream.  For a bidirectional stream, that varint is not a stream type, but a *frame* type.  So the test here is whether the frame is permitted.

For that, we probably want an `HFrame::is_known(u64) -> bool` function so that this can look more or less like this:

```suggestion
                // The "stream_type" for a bidirectional stream is a frame type.
                // We accept WEBTRANSPORT_STREAM (above), and HEADERS, 
                // and we have to ignore unknown types, but any other frame type
                // is bad if we know about it.
                if HFrame::is_known(stream_type) && stream_type != HEADER_FRAME
```

martinthomson (changes requested):

neqo-http3/src/stream_type_reader.rs
larseggert: Thanks for this. I made the change; it also required changing a test.

neqo-http3/src/frames/hframe.rs
```diff
@@ -78,6 +78,22 @@ impl HFrame {
...
+                | H3_FRAME_TYPE_CANCEL_PUSH
+                | H3_FRAME_TYPE_SETTINGS
+                | H3_FRAME_TYPE_PUSH_PROMISE
+                | H3_FRAME_TYPE_GOAWAY
+                | H3_FRAME_TYPE_MAX_PUSH_ID
+                | H3_FRAME_TYPE_PRIORITY_UPDATE_REQUEST
+                | H3_FRAME_TYPE_PRIORITY_UPDATE_PUSH
+        )
+    }
+
```
mxinden, line 96: How does this differ from:

https://github.com/mozilla/neqo/blob/ca0080e41ec3b06f1813da1c774ca321231e7eec/neqo-http3/src/frames/hframe.rs#L210-L224

mxinden: Related: I find matching non exhaustively on a `u64` (or new type `HFrameType`) error prone, e.g. the compiler does not prevent me from adding a `HFrameType` without updating `is_known_type`. Can one instead match exhaustively on the `enum` `HFrame`?

larseggert: I don't see how, but that might be due to my limited Rust-fu. the `enum` doesn't have any information about what byte value the encoding has.

martinthomson: If you are interested in ensuring that there is an exhaustive match, maybe write a function:
```rust
impl TryFrom<u64> for HFrameType {
  fn try_from(v: u64) -> Self {
  }
}
```

Then this could be implemented as:

```rust
fn is_known_type(v: u64) -> bool {
    HFrameType::try_from(v).is_some()
}
```
```

Comment (larseggert) [reply neqo-http3/src/frames/hframe.rs]: `error[E0117]: only traits defined in the current crate can be implemented for primitive types`
when I try

Comment (martinthomson) [reply neqo-http3/src/frames/hframe.rs]: You might need to move that to where HFrameType is defined.  Oh, and change it from `type HFrameType = u64` to `struct HFrameType(u64)`.

**Review by martinthomson — APPROVED** [review]

**Comment by martinthomson** [comment]
FWIW, it would be good to have a test for a new stream that started with something invalid.  That code isn't covered.

File: neqo-http3/src/frames/hframe.rs (hunk 5 above)
Comment (larseggert) [reply neqo-http3/src/frames/hframe.rs]: @mxinden even with the latest change, I don't see a way to do what you want. Any pointers?

Comment (mxinden) [reply neqo-http3/src/frames/hframe.rs]: I will take another look. No need to block. I can follow-up with a pull request.

Comment (mxinden) [reply neqo-http3/src/frames/hframe.rs]: I refactored this code locally using an `enum` with [explicit discriminants](https://doc.rust-lang.org/reference/items/enumerations.html#explicit-discriminants):

``` rust
#[repr(u64)]
pub enum HFrameTypeKnown {
    Data = 0x0,
    Headers = 0x1,
    CancelPush = 0x3,
    Settings = 0x4,
    PushPromise = 0x5,
    Goaway = 0x7,
    MaxPushId = 0xd,
    PriorityUpdateRequest = 0xf0700,
    PriorityUpdatePush = 0xf0701,
}
```

While it does group the various known types, it complicates the uses outside of `hframe.rs`.

Long story short, I don't think it is worth it. Let's keep it as is. Sorry for the noise.

PR #1998, opened 2024-07-21
fix: No time-based loss detection before `largest_acked` & fixes

### LINKED ISSUE #1874: `zerortt` QNS interop failures
Our client randomly fails `zerortt` interop tests with various servers. This seems to happen, because while the client correctly sends a lot of stream data in 0-RTT packets, it declares them all as lost and retransmits everything in 1-RTT, causing the interop runner to (correctly) declare a failure.

### LINKED ISSUE #2088: Huge timeout in `L1` QNS test
We're failing [this `L1` QNS test against the kwik server](https://github.com/mozilla/neqo/actions/runs/10718166512/artifacts/1895573675) due to the test taking longer than 300s.

That seems to be because we are setting **huge timeouts of ~90 seconds** during the connection-close process, and we don't continue on to the next connection in a reasonable time.
```
client  | 93s 96ms DEBUG Setting timeout of 19.971247ms
client  | 93s 96ms DEBUG stream 0 complete
client  | 93s 96ms DEBUG stream BiDi creatable
client  | 93s 96ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 93s 96ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 93s 96ms DEBUG TX blocked, profile=SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 93s 96ms DEBUG [Client 2cc1e081e1fe5325] delay duration 19.923328ms
client  | 93s 96ms DEBUG Setting timeout of 19.923328ms
client  | 93s 96ms DEBUG [Client 2cc1e081e1fe5325] State change from Confirmed -> Closing { error: Application(0), timeout: Instant { tv_sec: 667, tv_nsec: 656536104 } }
client  | 93s 96ms WARN Unhandled event StateChange(Closing { error: Application(0), timeout: Instant { tv_sec: 667, tv_nsec: 656536104 } })
client  | 93s 96ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 93s 96ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 93s 96ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 573, tv_nsec: 125867796 }
client  | 93s 96ms DEBUG [Client 2cc1e081e1fe5325] pn=3 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 42
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 13, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 93s 96ms DEBUG Setting timeout of 94.530598598s
sim     | Forwarding 82 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 48/149 (32.2%)
sim     | Forwarding 1352 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 48/150 (32.0%)
sim     | Forwarding 31 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 48/151 (31.8%)
sim     | Dropping 42 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 49/152 (32.2%)
sim     | Forwarding 24 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 34/135 (25.2%)
client  | 93s146ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 93s146ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 93s146ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 573, tv_nsec: 176033952 }
client  | 93s146ms DEBUG [Client 2cc1e081e1fe5325] pn=4 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 43
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 6283, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 93s147ms DEBUG Setting timeout of 94.48035613s
client  | 93s147ms DEBUG Setting timeout of 94.480342525s
sim     | Forwarding 43 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 49/153 (32.0%)
sim     | Forwarding 29 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 34/136 (25.0%)
client  | 93s178ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 93s178ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 93s178ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 573, tv_nsec: 207809755 }
client  | 93s178ms DEBUG [Client 2cc1e081e1fe5325] pn=5 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 43
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 10255, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 93s178ms DEBUG Setting timeout of 94.448622416s
client  | 93s178ms DEBUG Setting timeout of 94.44861391s
sim     | Dropping 43 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 50/154 (32.5%)
sim     | Forwarding 21 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 34/137 (24.8%)
client  | 108s111ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 108s111ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 108s111ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 588, tv_nsec: 140480140 }
client  | 108s111ms DEBUG [Client 2cc1e081e1fe5325] pn=6 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 45
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 1876839, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 108s111ms DEBUG Setting timeout of 79.515901617s
client  | 108s111ms DEBUG Setting timeout of 79.515885948s
sim     | Dropping 21 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 35/138 (25.4%)
sim     | Dropping 45 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 51/155 (32.9%)
sim     | Forwarding 204 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 35/139 (25.2%)
client  | 108s773ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 108s773ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 108s773ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 588, tv_nsec: 803084953 }
client  | 108s773ms DEBUG [Client 2cc1e081e1fe5325] pn=7 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 45
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 1959665, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 108s774ms DEBUG Setting timeout of 78.853327351s
client  | 108s774ms DEBUG Setting timeout of 78.853304208s
sim     | Forwarding 204 bytes 193.167.100.100:443 -> 193.167.0.100:42465, dropped 35/140 (25.0%)
client  | 108s774ms DEBUG [Client 2cc1e081e1fe5325] output_path send_profile SendProfile { limit: 1152, pto: None, probe: (), paced: false }
client  | 108s774ms DEBUG Building Short dcid Some(CID [8]: f734cc7d2e99a52a)
client  | 108s774ms DEBUG [Recvd-ap] immediate_ack at Instant { tv_sec: 588, tv_nsec: 804030443 }
client  | 108s774ms DEBUG [Client 2cc1e081e1fe5325] pn=8 type=Short pri-path:f734cc7d2e99a52a 0.0.0.0:42465->193.167.100.100:443 IpTos(Cs0, NotEct) len 45
client  |   TX -> Ack { largest_acknowledged: 2, ack_delay: 1959783, first_ack_range: 2, ack_ranges: [], ecn_count: None }
client  |   TX -> ConnectionClose { error_code: Application(0), frame_type: 0, reason_phrase: "kthxbye!" }
client  | 108s774ms DEBUG Setting timeout of 78.852422597s
client  | 108s774ms DEBUG Setting timeout of 78.852411897s
sim     | Forwarding 45 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 51/156 (32.7%)
sim     | Forwarding 45 bytes 193.167.0.100:42465 -> 193.167.100.100:443, dropped 51/157 (32.5%)
server  | QLog: done with 9840ca8b42f950a3.qlog
server  | QLog: done with d58f290ebd9bcfa9d5.qlog
server  | QLog: done with 2cc1e081e1fe5325.qlog
client  | 187s628ms DEBUG [Client 2cc1e081e1fe5325] State change from Closing { error: Application(0), timeout: Instant { tv_sec: 667, tv_nsec: 656536104 } } -> Closed(Application(0))
client  | 187s628ms INFO Closing timer expired
client  | 187s628ms DEBUG Output::None
client  | 187s628ms WARN Unhandled event StateChange(Closed(Application(0)))
client  | 187s628ms DEBUG Timer fired while closed
client  | 187s628ms DEBUG Output::None
```

larseggert:
**I will refactor this PR into smaller standalone ones.**

#1603 broke this, by falsely enabling time-based loss detection before we have received a valid `largest_acked`.

> [6.1.2. ](https://datatracker.ietf.org/doc/html/rfc9002#section-6.1.2)[Time Threshold](https://datatracker.ietf.org/doc/html/rfc9002#name-time-threshold)
>
> Once a later packet within the same packet number space has been acknowledged, an endpoint SHOULD declare an earlier packet lost if it was sent a threshold amount of time in the past.

This makes it clear that time-based LD should be disabled before an ACK is received. 

> [A.10. ](https://datatracker.ietf.org/doc/html/rfc9002#section-a.10)[Detecting Lost Packets](https://datatracker.ietf.org/doc/html/rfc9002#name-detecting-lost-packets)
>
>`assert(largest_acked_packet[pn_space] != infinite)`

The `DetectAndRemoveLostPackets` pseudo code also has the `assert` above.

When a PTO fires, we now explicitly also do a congestion reaction. 

Fixes #1874

Also includes some fixes to hopefully make retransmissions during the handshake more robust, by timing out all packets in all packet number spaces affected by a PTO (making their contents eligible for RTX) and by not applying `ack_delay` to the application packet number space PTO during the handshake.


larseggert:
Kwik failure: https://github.com/ptrd/kwik/issues/46

larseggert:
Quinn failure: https://github.com/quinn-rs/quinn/issues/1929

neqo-transport/src/recovery/mod.rs
```diff
@@ -327,7 +327,7 @@ impl LossRecoverySpace {
             .sent_packets
             .iter_mut()
             // BTreeMap iterates in order of ascending PN
-            .take_while(|p| p.pn() < largest_acked.unwrap_or(PacketNumber::MAX))
+            .take_while(|p| p.pn() < largest_acked.unwrap_or(0))
```
mxinden, line 330: Given that `Option` implements `Ord` when the inner type implements `Ord`, this could as well revert to the previous implementation, namely:

``` diff
-            .take_while(|p| p.pn() < largest_acked.unwrap_or(0))
+            .take_while(|p| Some(p.pn()) < largest_acked)
```



https://github.com/mozilla/neqo/commit/5d54da7e22f187e4d60daad486830f9d25bb4d83#diff-69602dcab54959a67ff71f9fbdcd82eed8f15dee641f913ef12ffe8c59ad7f73

I am fine either way.

test-fixture/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub fn fixture_init() {
 
 // This needs to be > 2ms to avoid it being rounded to zero.
 // NSS operates in milliseconds and halves any value it is provided.
-pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_millis(10);
+pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_millis(1000);
```
mxinden, line 55: Why is this change needed?

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -294,21 +294,16 @@ fn pto_handshake_complete() {
     // Check that the other packets (pkt2, pkt3) are Handshake packets.
     // The server discarded the Handshake keys already, therefore they are dropped.
     // Note that these don't include 1-RTT packets, because 1-RTT isn't send on PTO.
-    let (pkt2_hs, pkt2_1rtt) = split_datagram(&pkt2.unwrap());
-    assert_handshake(&pkt2_hs);
-    assert!(pkt2_1rtt.is_some());
```
mxinden, line 299: If this is deliberate, the comment above needs updating, right?

> are Handshake packets.

mxinden (commented):
Good catch! Patch looks good to me.

test-fixture/src/lib.rs
larseggert: We had an earlier attempt to fix this that required a new test. That test exposed a bug in the current test fixtures that broke 0-RTT. This change fixes that issue.

neqo-transport/src/recovery/mod.rs
```diff
@@ -973,8 +973,9 @@ mod tests {
 
     const ON_SENT_SIZE: usize = 100;
     /// An initial RTT for using with `setup_lr`.
-    const TEST_RTT: Duration = ms(80);
-    const TEST_RTTVAR: Duration = ms(40);
+    const TEST_RTT: Duration = INITIAL_RTT;
+    #[allow(clippy::cast_possible_truncation)]
+    const TEST_RTTVAR: Duration = Duration::from_millis(INITIAL_RTT.as_millis() as u64 / 2);
```
martinthomson, line 978: I would change these to ms(7000) and ms(3500) instead so that we don't engage the defaults.

martinthomson (commented):
Seems good.

neqo-transport/src/connection/tests/recovery.rs
larseggert: This comment remained unchanged since before #1603. So either it was already incorrect back then and we need to change it, or it should stay as is.

neqo-transport/src/recovery/mod.rs
larseggert: That change breaks tests:
```
        FAIL [   0.034s] neqo-transport recovery::tests::ack_delay_adjusted
        FAIL [   0.034s] neqo-transport recovery::tests::ack_delay_ignored
        FAIL [   0.036s] neqo-transport recovery::tests::big_gap_loss
        FAIL [   0.035s] neqo-transport recovery::tests::no_new_acks
        FAIL [   0.034s] neqo-transport recovery::tests::reduce_min_rtt
        FAIL [   0.033s] neqo-transport recovery::tests::time_loss_detection_timeout
```

neqo-transport/src/recovery/mod.rs
```diff
@@ -863,6 +863,19 @@ impl LossRecovery {
         // pto_time to increase which might cause PTO for later packet number spaces to not fire.
         if let Some(pn_space) = pto_space {
             qtrace!([self], "PTO {}, probing {:?}", pn_space, allow_probes);
+            // If we hit a PTO while we don't have a largest_acked yet, also do a congestion control
+            // reaction (because otherwise none would happen).
+            if let Some(space) = self.spaces.get(pn_space) {
+                if space.largest_acked.is_none() {
```
martinthomson, line 869:
```rust
if let Some(space) = self.spaces.get(pn_space).map_or_default(|sp| sp.largest_acked.is_none().then_some(sp)) {
```

Hmm, that doesn't look better.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -347,8 +347,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
 
     fn discard(&mut self, pkt: &SentPacket) {
         if pkt.cc_outstanding() {
-            assert!(self.bytes_in_flight >= pkt.len());
-            self.bytes_in_flight -= pkt.len();
+            self.bytes_in_flight = self.bytes_in_flight.saturating_sub(pkt.len());
```
martinthomson, line 350: Under what conditions does this saturate?  Is this only in the case where you PTO?

neqo-transport/src/recovery/mod.rs
```diff
@@ -863,6 +863,19 @@ impl LossRecovery {
...
+                    path.borrow_mut().on_packets_lost(
```
martinthomson, line 870: Do we really want to mark specific packets lost or do we want to just tell the path to reduce CWND directly?  That is, via something like `on_congestion_event()`.

neqo-transport/src/recovery/mod.rs
```diff
@@ -835,14 +835,14 @@ impl LossRecovery {
...
+    fn maybe_fire_pto(&mut self, path: &PathRef, now: Instant, lost: &mut Vec<SentPacket>) {
         let mut pto_space = None;
         // The spaces in which we will allow probing.
         let mut allow_probes = PacketNumberSpaceSet::default();
         for pn_space in PacketNumberSpace::iter() {
-            if let Some(t) = self.pto_time(rtt, *pn_space) {
-                allow_probes[*pn_space] = true;
+            if let Some(t) = self.pto_time(path.borrow().rtt(), *pn_space) {
                 if t <= now {
+                    allow_probes[*pn_space] = true;
```
martinthomson, line 845: Is this the real bug?

martinthomson (commented):
I'm not 100% on the loss stuff.  I'm clearly not smort enuff.

neqo-transport/src/recovery/mod.rs
larseggert: No, this is just a minor optimization.

neqo-transport/src/cc/classic_cc.rs
larseggert: Yes, the extra call to `on_packets_lost` takes packets out of `bytes_in_flight`.

neqo-transport/src/recovery/mod.rs
larseggert: I tried that, but it seemed to need a bunch of additional code changes, since that is not currently exposed. Can give it another go.

larseggert: @martinthomson take another look; I now expose `on_congestion_event`.

larseggert:
Is there any point in undoing some more of the changes from #1603?

neqo-transport/src/recovery/mod.rs
```diff
@@ -327,7 +327,7 @@ impl LossRecoverySpace {
             .sent_packets
             .iter_mut()
             // BTreeMap iterates in order of ascending PN
-            .take_while(|p| p.pn() < largest_acked.unwrap_or(PacketNumber::MAX))
+            .take_while(|p| Some(p.pn()) < largest_acked)
```
mxinden, line 330: Simplifying even further, how about not handling `Option`s in the first place:

``` diff
modified   neqo-transport/src/recovery/mod.rs
@@ -321,13 +321,15 @@ impl LossRecoverySpace {
         );
         self.first_ooo_time = None;
 
-        let largest_acked = self.largest_acked;
+        let Some(largest_acked) = self.largest_acked else {
+            return;
+        };
 
         for packet in self
             .sent_packets
             .iter_mut()
-            .take_while(|p| Some(p.pn()) < largest_acked)
+            .take_while(|p| p.pn() < largest_acked)
         {
             // Packets sent before now - loss_delay are deemed lost.
             if packet.time_sent() + loss_delay <= now {
@@ -337,7 +339,7 @@ impl LossRecoverySpace {
                     packet.time_sent(),
                     loss_delay
                 );
-            } else if largest_acked >= Some(packet.pn() + PACKET_THRESHOLD) {
+            } else if largest_acked >= packet.pn() + PACKET_THRESHOLD {
                 qtrace!(
                     "lost={}, is >= {} from largest acked {:?}",
                     packet.pn(),
@@ -345,9 +347,7 @@ impl LossRecoverySpace {
                     largest_acked
                 );
             } else {
-                if largest_acked.is_some() {
-                    self.first_ooo_time = Some(packet.time_sent());
-                }
+                self.first_ooo_time = Some(packet.time_sent());
                 // No more packets can be declared lost after this one.
                 break;
             };
```

neqo-transport/src/recovery/mod.rs
```diff
@@ -863,6 +863,15 @@ impl LossRecovery {
         // pto_time to increase which might cause PTO for later packet number spaces to not fire.
         if let Some(pn_space) = pto_space {
             qtrace!([self], "PTO {}, probing {:?}", pn_space, allow_probes);
+            // If we hit a PTO while we don't have a largest_acked yet, also do a congestion control
+            // reaction (because otherwise none would happen).
```
mxinden, line 867:
```suggestion
            // Packets are only declared as lost, relative to the
            // `largest_acked`. If we hit a PTO while we don't have a
            // largest_acked yet, also do a congestion control reaction (because
            // otherwise none would happen).
```

Is this correct? If so, change above would have made it easier for me to understand.

larseggert:
Still looking into the performance drop in "transfer/Run multiple transfers with varying seeds".

larseggert:
We should also look into why pacing is making things slower:
```
neqo 	neqo 	reno 	on 	415.4 ± 39.1 	375.7 	480.0 	1.00
neqo 	neqo 	reno 		206.9 ± 21.3 	181.8 	242.1 	1.00
neqo 	neqo 	cubic 	on 	450.3 ± 57.6 	385.6 	575.1 	1.00
neqo 	neqo 	cubic 		285.0 ± 168.1 	168.5 	610.7 	1.00
```

mxinden:
Maybe saving others some clicks, here are the numbers of a recent bench run on `main`:
```
neqo 	neqo 	reno 	on 	196.1 ± 19.4 	172.1 	236.5 	1.00
neqo 	neqo 	reno 		174.4 ± 10.3 	153.0 	195.1 	1.00
neqo 	neqo 	cubic 	on 	194.6 ± 27.0 	173.3 	275.2 	1.00
neqo 	neqo 	cubic 		175.7 ± 5.9 	168.8 	188.3 	1.00
```

https://github.com/mozilla/neqo/actions/runs/10051108975

martinthomson:
Pacing will make things slower, but might make things nicer on the network.  It should be a small delta, not a big one.

larseggert:
So I bisected this, and b6e05e304b1ef582bc232710d3e1177fb5b371fe is the first bad commit (where performance of that benchmark drops).

There are really only two changes in there, one being [the removal or the `max_ack_delay` addition from the `loss_delay`](https://github.com/mozilla/neqo/pull/1998/commits/b6e05e304b1ef582bc232710d3e1177fb5b371fe#diff-d13ee7b6b9fa303518ce615e67358f0ed764953da4b84861d288ed9391020a22), the other being the [fix to not do time-based loss recovery before having a `largest_acked`](https://github.com/mozilla/neqo/pull/1998/commits/b6e05e304b1ef582bc232710d3e1177fb5b371fe#diff-fb7d3f0ed8644dbebd84b7de9017aef7453e0ce65777b3beb8e2e9bdba39099f). I fail to see how either of these could cause this, however. Hm.

larseggert:
I validated that the key fix here (not doing time-based loss recovery while we don't have a `largest_acked`) causes the drop in benchmark performance. I can't quite figure out **why** that would be, given that we're now **not** over-aggressively retransmitting packets.

mxinden:
A guess @larseggert:

- The previous version, time-based loss recovery while we don't have a `largest_acked`, leads to more packets being declared lost.
- More packets being declared lost reduces the congestion window.
- A small congestion window requires the `Pacer` to pace the packets over a smaller time window. (See [correlation of pacer and congestion window](https://github.com/mozilla/neqo/blob/608c0a72851a55f73dcfc9f206143748b36fd1bb/neqo-transport/src/pace.rs#L17-L24).)
- Thus smaller spaces between packets.
- Thus higher throughput on a low-latency (localhost) connection.

larseggert:
>     * A small congestion window requires the `Pacer` to pace the packets over a smaller time window. (See [correlation of pacer and congestion window](https://github.com/mozilla/neqo/blob/608c0a72851a55f73dcfc9f206143748b36fd1bb/neqo-transport/src/pace.rs#L17-L24).)

Sorry for having been unclear. The performance reduction I was looking into was the drop in "transfer/Run multiple transfers with varying seeds" bench,  not the drop related to pacing visible in the comparison table.

mxinden:
Both might have the same root cause @larseggert, no? "transfer/Run multiple transfers with varying seeds" uses the default `ConnectionParameters`. Thus, pacing is enabled.

https://github.com/mozilla/neqo/blob/608c0a72851a55f73dcfc9f206143748b36fd1bb/neqo-transport/src/connection/params.rs#L105

Running this benchmark without pacing (e.g. via https://github.com/mozilla/neqo/pull/2006), increases performance, potentially back to the baseline.

larseggert:
You're right, it might be the same cause.

mxinden:
Looking at the first 50 `pace.rs` logs of the ""transfer/Run multiple transfers with varying seeds" on `main` and on `0rtt-tmp`, more specifically at the following lines.

https://github.com/mozilla/neqo/blob/88d3ed105db1b1a805a7b8c3b64078008069a403/neqo-transport/src/pace.rs#L77

https://github.com/mozilla/neqo/blob/88d3ed105db1b1a805a7b8c3b64078008069a403/neqo-transport/src/pace.rs#L87

- `main` reduces its `cwnd` quickly to `6160`, while `0rtt-tmp` stays at `12320`.
- `main` does not pace (i.e. `no wait`) while `0rtt-tmp` paces within the first 50 logs (i.e. `wait`).

#### `main`

```
➜  neqo git:(main) ✗ RUST_LOG=neqo_transport::pace=trace cargo test --bench transfer --features bench varying 2>&1 | head -n 50
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running benches/transfer.rs (target/debug/deps/transfer-7ba07c4aa3c126f9)
Testing transfer/Run multiple transfers with varying seeds
0s  0ms TRACE [Pacer 2464/1232] spend 1232 over 12320, 100ms
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 561055565 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 561055565 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 561055565 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 561055565 }
0s  2ms TRACE [Pacer 2464/1232] spend 671 over 12320, 100ms
0s  2ms TRACE [Pacer 1793/1232] spend 561 over 12320, 100ms
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 561055565 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  2ms TRACE [Pacer 1232/1232] spend 1232 over 6160, 100ms
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  2ms TRACE [Pacer 2464/1232] next 6160/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 675862863 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  2ms TRACE [Pacer 2464/1232] next 6160/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 675862863 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  4ms TRACE [Pacer 2464/1232] spend 50 over 6160, 124.38313ms
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  4ms TRACE [Pacer 2464/1232] spend 81 over 6160, 124.38313ms
0s  4ms TRACE [Pacer 2383/1232] spend 234 over 6160, 124.38313ms
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  5ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  5ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  5ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  5ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  5ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  5ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  5ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  5ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 624566863 }
0s  5ms TRACE [Pacer 1232/1232] spend 1232 over 12320, 100ms
0s  5ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  5ms TRACE [Pacer 2464/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 733268221 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2464/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 733268221 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2464/1232] next 12320/100ms no wait = Instant { tv_sec: 12085, tv_nsec: 733268221 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  6ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  7ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  7ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  7ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
0s  7ms TRACE [Pacer 2464/1232] spend 50 over 12320, 119.752079ms
0s  8ms TRACE [Pacer 2464/1232] spend 1282 over 12320, 119.752079ms
0s  8ms TRACE [Pacer 2149/1232] next 6160/124.38313ms no wait = Instant { tv_sec: 12085, tv_nsec: 685438695 }
```


#### `0rtt-tmp`
```
➜  neqo git:(0rtt-tmp) ✗ RUST_LOG=neqo_transport::pace=trace cargo test --bench transfer --features bench varying 2>&1 | head -n 50
   Compiling neqo-transport v0.8.1 (/home/mxinden/code/github.com/mozilla/neqo/neqo-transport)
   Compiling neqo-qpack v0.8.1 (/home/mxinden/code/github.com/mozilla/neqo/neqo-qpack)
   Compiling neqo-http3 v0.8.1 (/home/mxinden/code/github.com/mozilla/neqo/neqo-http3)
   Compiling test-fixture v0.8.1 (/home/mxinden/code/github.com/mozilla/neqo/test-fixture)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.85s
     Running benches/transfer.rs (target/debug/deps/transfer-7ba07c4aa3c126f9)
Testing transfer/Run multiple transfers with varying seeds
0s  0ms TRACE [Pacer 2464/1232] spend 1232 over 12320, 100ms
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  0ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  2ms TRACE [Pacer 2464/1232] spend 664 over 12320, 100ms
0s  2ms TRACE [Pacer 1800/1232] spend 568 over 12320, 100ms
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 788940966 }
0s  2ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  3ms TRACE [Pacer 1232/1232] spend 51 over 12320, 118.702751ms
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  4ms TRACE [Pacer 2464/1232] spend 82 over 12320, 118.702751ms
0s  4ms TRACE [Pacer 2382/1232] spend 256 over 12320, 118.702751ms
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  4ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  4ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  4ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  4ms TRACE [Pacer 1232/1232] next 12320/100ms no wait = Instant { tv_sec: 12154, tv_nsec: 854132492 }
0s  4ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  5ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  5ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  5ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  5ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  5ms TRACE [Pacer 1232/1232] spend 51 over 12320, 112.760801ms
0s  6ms TRACE [Pacer 2464/1232] spend 1281 over 12320, 112.760801ms
0s  6ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  6ms TRACE [Pacer 1183/1232] next 12320/112.760801ms wait 224.24µs = Instant { tv_sec: 12155, tv_nsec: 19325906 }
0s  6ms TRACE [Pacer 1183/1232] next 12320/112.760801ms wait 224.24µs = Instant { tv_sec: 12155, tv_nsec: 19325906 }
0s  6ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  6ms TRACE [Pacer 1183/1232] next 12320/112.760801ms wait 224.24µs = Instant { tv_sec: 12155, tv_nsec: 19325906 }
0s  7ms TRACE [Pacer 1183/1232] spend 1332 over 12320, 112.760801ms
0s  7ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }
0s  7ms TRACE [Pacer 0/1232] next 12320/112.760801ms wait 5.63804ms = Instant { tv_sec: 12155, tv_nsec: 24963946 }
0s  7ms TRACE [Pacer 0/1232] next 12320/112.760801ms wait 5.63804ms = Instant { tv_sec: 12155, tv_nsec: 24963946 }
0s  7ms TRACE [Pacer 2126/1232] next 12320/118.702751ms no wait = Instant { tv_sec: 12154, tv_nsec: 907643717 }

```

larseggert:
> * `main` reduces its `cwnd` quickly to `6160`, while `0rtt-tmp` stays at `12320`.

That is probably because `main` erroneously marks 0-RTT as lost, and does a congestion reaction.

> * `main` does not pace (i.e. `no wait`) while `0rtt-tmp` paces within the first 50 logs (i.e. `wait`).

It might be that because `main` is in the recovery state, it can immediately (= not paced) retransmit a single packet.

neqo-transport/src/recovery/mod.rs
```diff
@@ -321,13 +321,15 @@ impl LossRecoverySpace {
         );
         self.first_ooo_time = None;
 
-        let largest_acked = self.largest_acked;
+        let Some(largest_acked) = self.largest_acked else {
```
KershawChang, line 324: It'd be good to add some comments here to explain this.

KershawChang (approved):
Looks reasonable to me.
I only have one minor comment.

mxinden:
> > * `main` does not pace (i.e. `no wait`) while `0rtt-tmp` paces within the first 50 logs (i.e. `wait`).
> 
> It might be that because `main` is in the recovery state, it can immediately (= not paced) retransmit a single packet.

Only when we send a packet `< mtu` and we are in `RecoveryStart` do we ignore the `Pacer`:

https://github.com/mozilla/neqo/blob/b63e0ff567342a220a9633d1c3ff2b15706bf5bb/neqo-transport/src/recovery/mod.rs#L914-L930

Would that explain the near 100% performance regression on the client/server transfers with pacing enabled?

---

> We should also look into why pacing is making things slower:
> 
> ```
> neqo 	neqo 	reno 	on 	415.4 ± 39.1 	375.7 	480.0 	1.00
> neqo 	neqo 	reno 		206.9 ± 21.3 	181.8 	242.1 	1.00
> neqo 	neqo 	cubic 	on 	450.3 ± 57.6 	385.6 	575.1 	1.00
> neqo 	neqo 	cubic 		285.0 ± 168.1 	168.5 	610.7 	1.00
> ```

I am still puzzled about these benchmark regressions. I looked into it some more, but without success. I don't think my above guesses are valid.

@larseggert do you have further ideas worth investigating?

I am fine merging here even with the regression. I think bugfixes should take precedence over performance. That said, I would like to investigate further and ideally not uplift to Firefox before we found the reason.

I will take a look at qlogs next. I can reproduce the regression with the Simulator, which should allow me to generate qlogs without impacting run time, given that time is simulated.

neqo-transport/src/recovery/mod.rs
larseggert: This got me thinking about whether the logic here got broken, causing the performance issue. I have a possible fix in 76c32f3101e61b6c8864a44a70413321a15d1475; let's say what CI says.

mxinden: The `transfer.rs` benchmarks are back to normal.

> transfer/pacing-false/varying-seeds: No change in performance detected.
> transfer/pacing-true/varying-seeds: No change in performance detected.

But the client/server transfer benchmarks are still elevated.

```
neqo 	neqo 	reno 	on 	456.7 ± 51.6 	375.9 	567.0 	1.00
neqo 	neqo 	reno 		262.9 ± 126.9 	178.7 	620.9 	1.00
neqo 	neqo 	cubic 	on 	450.3 ± 58.1 	370.8 	580.7 	1.00
neqo 	neqo 	cubic 		214.4 ± 28.5 	182.3 	280.3 	1.00
```

larseggert:
I wet back to fbc5d62697c968dec968941d4c495f4fa6869c9d, which was the state of the code before #1603 was merged. I needed to add the ability to turn off pacing on the server, and comment out some noisy `eprintln!` stuff. I then did a hyperfine run, comparing our client and server with Reno, with and without pacing.

Without pacing:
```
  Time (mean ± σ):     471.5 ms ±  16.8 ms    [User: 305.3 ms, System: 114.1 ms]
  Range (min … max):   430.3 ms … 485.9 ms    10 runs
```

With pacing:
```
  Time (mean ± σ):     523.7 ms ±  14.0 ms    [User: 299.6 ms, System: 120.2 ms]
  Range (min … max):   495.8 ms … 546.4 ms    10 runs
```

In other words, we did see a similar performance difference before #1603 as we are seeing now.

However, we **also** see this performance difference **with** #1603. I'm now bisecting to see when that changed.

larseggert:
I did some comparison runs between this PR and `HEAD`. It seems that the performance difference between pacing and non-pacing is only present if the loopback MTU is 64KB.

This PR, MTU 1500:
```
Benchmark 1: target/release/neqo-client --cc reno  --output-dir /tmp/tmp.w6tjsiaAdZ -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     552.8 ms ±  67.7 ms    [User: 290.3 ms, System: 184.4 ms]
  Range (min … max):   495.9 ms … 958.6 ms    100 runs

Benchmark 1: target/release/neqo-client --cc reno --no-pacing --output-dir /tmp/tmp.w6tjsiaAdZ -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     552.9 ms ±  47.8 ms    [User: 285.3 ms, System: 182.3 ms]
  Range (min … max):   490.8 ms … 772.2 ms    100 runs
```

This PR, MTU 64KB:
```
Benchmark 1: target/release/neqo-client --cc reno  --output-dir /tmp/tmp.gKIfqAvJbs -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     397.0 ms ± 121.4 ms    [User: 149.0 ms, System: 57.5 ms]
  Range (min … max):   227.4 ms … 692.2 ms    100 runs

Benchmark 1: target/release/neqo-client --cc reno --no-pacing --output-dir /tmp/tmp.gKIfqAvJbs -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     262.3 ms ±  18.2 ms    [User: 148.3 ms, System: 56.8 ms]
  Range (min … max):   232.2 ms … 314.3 ms    100 runs
```

`HEAD`, MTU 1500:
```
Benchmark 1: target/release/neqo-client --cc reno  --output-dir /tmp/tmp.GHDuGPjylE -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     578.8 ms ±  48.4 ms    [User: 311.1 ms, System: 211.9 ms]
  Range (min … max):   520.0 ms … 766.4 ms    100 runs

Benchmark 1: target/release/neqo-client --cc reno --no-pacing --output-dir /tmp/tmp.GHDuGPjylE -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     577.3 ms ±  39.5 ms    [User: 327.2 ms, System: 213.6 ms]
  Range (min … max):   540.2 ms … 789.1 ms    100 runs
```

`HEAD`, MTU 64KB:
```
Benchmark 1: target/release/neqo-client --cc reno  --output-dir /tmp/tmp.6wBXumkvZX -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     256.8 ms ±  34.3 ms    [User: 144.6 ms, System: 52.7 ms]
  Range (min … max):   221.5 ms … 501.1 ms    100 runs

Benchmark 1: target/release/neqo-client --cc reno --no-pacing --output-dir /tmp/tmp.6wBXumkvZX -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     245.1 ms ±  23.4 ms    [User: 142.5 ms, System: 52.7 ms]
  Range (min … max):   220.5 ms … 432.7 ms    100 runs
```

martinthomson:
I wouldn't read too much into those differences.  A larger MTU is going to mean that overheads are more pronounced.

Also, pacing is always going to be slower in *this* sort of test.  Pacing is designed to be gentle on queues in the network at the cost of extra CPU and delay at the endpoint that paces.  As long as the cost is small, we're good.

larseggert:
On further investigation, this seems to be related to what `PACER_SPEEDUP` is set to. When I double it to `4`, the difference for this PR with a 64KB goes away:
```
Benchmark 1: target/release/neqo-client --cc reno  --output-dir /tmp/tmp.AP82nSCWV4 -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     268.4 ms ±  36.8 ms    [User: 155.3 ms, System: 53.1 ms]
  Range (min … max):   230.9 ms … 514.0 ms    100 runs

Benchmark 1: target/release/neqo-client --cc reno --no-pacing --output-dir /tmp/tmp.AP82nSCWV4 -o -a hq-interop https://127.0.0.1:4434/134217728
  Time (mean ± σ):     264.8 ms ±  36.1 ms    [User: 158.5 ms, System: 58.4 ms]
  Range (min … max):   228.5 ms … 503.1 ms    100 runs
```

Given all this, I think we can merge this.

mxinden (approved):
Thanks for all the measurement work!

neqo-transport/src/recovery/mod.rs
```diff
@@ -840,34 +840,49 @@ impl LossRecovery {
...
-
-                    pto_space = pto_space.or(Some(*pn_space));
-                }
+            if self
+                .pto_time(path.borrow().rtt(), *pn_space)
+                .map_or(true, |t| now < t)
+            {
+                continue;
             }
+            allow_probes[*pn_space] = true;
```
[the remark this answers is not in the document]
larseggert: I think so, but let me think about this again.

larseggert:
This isn't quite ready yet, because in the process of fixing the test failure after merging `HEAD`, I realized we have another bug: When there is packet loss during the handshake, we do not always ACK when we should be ACK'ing. That's because when we send an ACK, we reset the ACK timer (even during the handshake) and that causes ACKs to be not sent when they should be (on retransmit).

I'm working on a fix, and want to see that merged before continuing on this one, because that fix might impact what needs to change in this PR.

larseggert:
That other PR is #2027.

Issue #2001, opened 2024-07-22
WouldBlock, message: "Resource temporarily unavailable"


larseggert:
I recently started seeing these test failures. They are not repeatable, but occur occasionally.
```
0s  0ms DEBUG Logging initialized
thread 'tests::datagram_tos' panicked at neqo-udp/src/lib.rs:150:14:
receive to succeed: Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:1654:5
   3: core::result::Result<T,E>::expect
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:1034:23
   4: neqo_udp::tests::datagram_tos
             at ./src/lib.rs:148:33
   5: neqo_udp::tests::datagram_tos::{{closure}}
             at ./src/lib.rs:134:26
   6: core::ops::function::FnOnce::call_once
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

mxinden:
Thanks for reporting @larseggert. I expect this to be fixed with https://github.com/mozilla/neqo/pull/2002.

PR #2002, opened 2024-07-22
fix(udp): test blocking Socket

mxinden:
`quinn_udp::UdpSocketState` sets the the OS socket to non-blocking in order to be used by poll based runtimes like `tokio`.

For basic unit tests, this might result in race conditions, e.g. where one socket tries to receive before the other is done sending, thus returning `WouldBlock`.

This commit sets the OS sockets back to blocking on `neqo-udp` unit tests.

---

Fixes https://github.com/mozilla/neqo/issues/2001.

Taken from [`quinn-udp`](https://github.com/quinn-rs/quinn/blob/ff91ffe8154e2ff1137ff8d29cd7ca5fd1030452/quinn-udp/tests/tests.rs#L187-L188).

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -130,10 +130,17 @@ mod tests {
 
     use super::*;
 
+    fn socket() -> Result<Socket<std::net::UdpSocket>, io::Error> {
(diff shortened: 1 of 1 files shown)


mxinden:
Windows CI failure would be fixed by https://github.com/mozilla/neqo/pull/2003.

larseggert (approved):

PR #2004, opened 2024-07-23
deps(quinn-udp): enable `direct-log` feature to log via `log`

mxinden:
This pull request contains 3 commits, where the first two build up to the second:

- deps(quinn-udp): move to workspace dependency
- deps(quinn-udp): bump to latest version
- deps(quinn-udp): enable direct-log feature
     https://github.com/quinn-rs/quinn/pull/1923 made the `tracing` dependency optional. In addition, when `direct-log` is enabled, it allows using `log` instead of `tracing` for logging.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ rust-version = "1.76.0"
 [workspace.dependencies]
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
-quinn-udp = { version = "0.5.0", default-features = false }
+quinn-udp = { version = "0.5.4", default-features = false }
 
 [workspace.lints.clippy]
 cargo = { level = "warn", priority = -1 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ rust-version = "1.76.0"
 [workspace.dependencies]
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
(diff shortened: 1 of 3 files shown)


larseggert:
Could you say some words about what that change would give us? Specifically, making quinn-udp a workspace dep?

Cargo.toml
```diff
@@ -32,6 +32,7 @@ rust-version = "1.76.0"
 [workspace.dependencies]
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
+quinn-udp = { version = "0.5.4", default-features = false, features = ["direct-log"] }
```
larseggert, line 35:
```suggestion
quinn-udp = { version = "0.5", default-features = false, features = ["direct-log"] }
```
to be consistent with how we not specify minor versions for other deps? Or do we need this here?

larseggert (approved):
LGTM, modulo a question

Cargo.toml
mxinden: Sorry for not specifically calling this out. We need `v0.5.4` or above for the following reasons:

- `direct-log` feature is only available with `v0.5.3`, see https://github.com/quinn-rs/quinn/pull/1923
- `v0.5.3` had a bug, fixed in `v0.5.4` https://github.com/quinn-rs/quinn/pull/1932

The above, i.e. `v0.5.4` allows cargo to choose any version `>=0.5.4`.

PR #2007, opened 2024-07-26
fix(benches/transfer): remove throughput

mxinden:
The `neqo-transport/benches/transfer.rs` benchmarks use the `test-fixture/src/sim` simulator. The simulator can travel in time, i.e. it simulates time.

The _wall-clock time_ of a single benchmark run is not the amount of time it took to transfer `TRANSFER_AMOUNT`. The _simulated time_ is the amount of time it took to transfer `TRANSFER_AMOUNT`.

`criterion` will use the _wall-clock time_, not the _simulated time_ to calculate the throughput based on `TRANSFER_AMOUNT`. The resulting throughput number is not meaningful.

This commit removes the call to `group.throughput`, thus removing the misleading `criterion` throughput reporting.

---

Related to https://github.com/mozilla/neqo/pull/1998.

### DIFF
--- a/neqo-transport/benches/transfer.rs
+++ b/neqo-transport/benches/transfer.rs
@@ -6,7 +6,7 @@
 
 use std::time::Duration;
 
-use criterion::{criterion_group, criterion_main, BatchSize::SmallInput, Criterion, Throughput};
+use criterion::{criterion_group, criterion_main, BatchSize::SmallInput, Criterion};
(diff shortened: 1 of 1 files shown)


larseggert (approved):
LGTM, but maybe add a comment where the removed line was, so we know why we removed it?

larseggert (approved):

PR #2009, opened 2024-07-26
neqo-crypto: use SSL_PeerCertificateChainDER

jschanck:
The SSL_PeerCertificateChainDER function was added in NSS 3.103 to allow an application to retrieve the peer's presented certificate chain without constructing CERTCertificates. This is expected to improve performance, as constructing a CERTCertificate will typically involve querying the NSS certificate database.

We switched PSM over to SSL_PeerCertificateChainDER in [Bug 1899431](https://bugzilla.mozilla.org/show_bug.cgi?id=1899431).

Assuming you land this, you'll need to apply the following patch to neqo-glue when you uplift to M-C:
```diff
diff --git netwerk/socket/neqo_glue/src/lib.rs netwerk/socket/neqo_glue/src/lib.rs
--- netwerk/socket/neqo_glue/src/lib.rs
+++ netwerk/socket/neqo_glue/src/lib.rs
@@ -1218,14 +1218,11 @@ pub extern "C" fn neqo_http3conn_peer_ce
     conn: &mut NeqoHttp3Conn,
     neqo_certs_info: &mut NeqoCertificateInfo,
 ) -> nsresult {
-    let mut certs_info = match conn.conn.peer_certificate() {
-        Some(certs) => certs,
-        None => return NS_ERROR_NOT_AVAILABLE,
+    let Some(certs_info) = conn.conn.peer_certificate() else {
+        return NS_ERROR_NOT_AVAILABLE;
     };

-    neqo_certs_info.certs = certs_info
-        .map(|cert| cert.iter().cloned().collect())
-        .collect();
+    neqo_certs_info.certs = certs_info.iter().map(ThinVec::from).collect();

     match &mut certs_info.stapled_ocsp_responses() {
         Some(ocsp_val) => {
```


larseggert:
@jschanck I'll need to fix CI Monday, then will rebase.

neqo-crypto/src/p11.rs
```diff
@@ -287,6 +287,46 @@ impl Item {
...
+    type IntoIter = ItemArrayIterator<'a>;
+    fn into_iter(self) -> Self::IntoIter {
+        if self.ptr.is_null() || unsafe { *self.ptr }.len == 0 {
+            Self::IntoIter {
+                iter: Iter::default(),
+            }
+        } else {
+            // (*self.ptr).items points to (*self.ptr).len > 0 consecutive elements of type SECItem
+            let items: &[SECItem] = unsafe {
+                #[allow(clippy::disallowed_methods)]
```
martinthomson, line 306: It seems fairly trivial to change `null_safe_slice` to make it usable here...

```rust
unsafe fn null_safe_slice<'a, T, L>(data: *const T, len: L) -> &'a [T]
where
    usize: TryFrom<L>,
{
    if data.is_null() {
        &[]
    } else if let Ok(len) = usize::try_from(len) {
        if len == 0 {
            &[]
        } else {
            #[allow(clippy::disallowed_methods)]
            std::slice::from_raw_parts(data, len)
        }
    } else {
        panic!("null_safe_slice: size overflow");
    }
}
```

neqo-crypto/src/cert.rs
```diff
@@ -72,32 +74,26 @@ fn signed_cert_timestamp(fd: *mut PRFileDesc) -> Option<Vec<u8>> {
...
         })
     }
+}
 
-    fn head(certs: &CertList) -> *const CERTCertListNode {
-        // Three stars: one for the reference, one for the wrapper, one to deference the pointer.
-        unsafe { addr_of!((***certs).list).cast() }
+impl CertificateInfo {
+    #[must_use]
+    pub fn iter(&self) -> ItemArrayIterator<'_> {
```
martinthomson, line 87: I *think* that you can avoid a whole lot of work by using `null_safe_slice` here, using the `IntoIterator` implementation on the slice directly.

martinthomson (commented):
Regarding the patch to the glue, consider using `let Some(info) = ...  else { return NS_ERROR_FAILURE; };` instead.

neqo-crypto/src/p11.rs
jschanck: done!

jschanck: Oh, I missed the `len == 0` check. I'll add it.

neqo-crypto/src/cert.rs
jschanck: I don't see it. I think we need the ItemArrayIterator wrapper to get the lifetimes right.

larseggert:
@jschanck I rebased, now the NSS build is breaking due to https://bugzilla.mozilla.org/show_bug.cgi?id=1902078#c13

neqo-crypto/src/p11.rs
```diff
@@ -15,6 +15,7 @@ use std::{
     ops::{Deref, DerefMut},
     os::raw::c_uint,
     ptr::null_mut,
+    slice::Iter,
```
martinthomson, line 18:
```suggestion
    slice::Iter as SliceIter,
```

martinthomson (approved):
Seems legit.

larseggert (approved):
@jschanck any chance we could add some test to improve coverage? See https://app.codecov.io/gh/mozilla/neqo/pull/2009/blob/neqo-crypto/src/cert.rs

martinthomson (approved):
I'm going to merge this now.  Thanks for the exemplary change, John.

PR #2019, opened 2024-07-29
bench(.github): don't print capture group $4 twice 

mxinden:
In the following regex there are 6 capture groups.

```
perl -p -0777 -e 's/(.*?)\n(.*?)(((No change|Change within|Performance has).*?)(\nFound .*?)?)?\n\n/<details><summary>$1: $4<\/summary><pre>\n$2$4$6<\/pre><\/details>\n/gs' |\
```

Capture group 5 is nested within capture group 4:

```
((No change|Change within|Performance has).*?)
```

Instead of printing capture group 5, print capture group 4. As an example instead of printing

```
Performance has
```

Print

```
Performance has improved.
````

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -200,7 +200,7 @@ jobs:
           fi
           sed -E -e 's/^                 //gi' \
                  -e 's/((change|time|thrpt):[^%]*% )([^%]*%)(.*)/\1<b>\3<\/b>\4/gi' results.txt |\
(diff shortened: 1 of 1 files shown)


larseggert:
I assume this is correct, but would you have a bench comment that illustrates the bug?

larseggert:
Ah. I think we should just remove that line from the collapsible, since it's shown in the first line.

larseggert (approved):

PR #2025, opened 2024-07-31
Add @mxinden to CODEOWNERS

larseggert:
Discussed yesterday with @KershawChang. @martinthomson, are you OK with this? @mxinden is pretty actively reviewing, so it might unblock getting PRs merged when both of you are busy. 

(We also need to give @mxinden repo write access if we do this.)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*    @KershawChang @martinthomson @larseggert
+*    @KershawChang @martinthomson @larseggert @mxinden


mxinden (approved):
:pray:

martinthomson (commented):
I don't think we need to make Max a code owner in that sense, though I have no objection to giving him write access, which is what you really need.

martinthomson (approved):
Oh, you are talking about unblocking review.  Yes, no problem.  I just read my email out of order and missed the necessary context.

KershawChang (approved):

PR #2027, opened 2024-08-01
fix: Single-packet RTX in Initial/Handshake + RTX handshake ACKs

larseggert:
With this change, we're now only sending one packet on PTO in the Initial and Handshake packet number spaces, for both server and client. Before, the server was allowed to send two and the client was allowed to send two during Handshake, which meant that if one packet was lost, we would RTX the lost packet and a (useless) empty PING. The server would then hit the amplification limit. With this change, we only RTX the lost packet, making the Handshake more robust under loss.

A second change here is to immediately force an ACK during the Handshake when a previous ACK was lost. Before this, the original Initial form the server contained an ACK, but any RTX did not, because we stopped the ACK timer when generating the first ACK and never restarting it.

Both these changes reduce `L1` and `C1` interop failures.


neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1194,3 +1195,63 @@ fn emit_authentication_needed_once() {
...
+        assert_eq!(
+            server.stats().frame_tx,
+            FrameStats {
+                crypto: i * 2,
+                ack: i,
+                all: i * 3,
+                ..Default::default()
+            }
+        );
+        ci = None;
```
mxinden, line 1245:
```suggestion
        let si = server.process(&ci.take(), now).dgram().unwrap();
        assert_eq!(si.len(), server.plpmtu());
        assert_eq!(
            server.stats().frame_tx,
            FrameStats {
                crypto: i * 2,
                ack: i,
                all: i * 3,
                ..Default::default()
            }
        );
```

Using `take()` is a bit simpler and concise. Haven't tried to compile it. Feel free to ignore.

neqo-transport/src/tracking.rs
```diff
@@ -593,11 +603,9 @@ impl AckTracker {
             .ack_freq(seqno, tolerance, delay, ignore_order);
     }
 
-    // Force an ACK to be generated immediately (a PING was received).
-    pub fn immediate_ack(&mut self, now: Instant) {
-        self.get_mut(PacketNumberSpace::ApplicationData)
-            .unwrap()
-            .immediate_ack(now);
+    // Force an ACK to be generated immediately.
```
mxinden, line 606: I believe the doc comment below also needs updating:

https://github.com/mozilla/neqo/blob/0eb9174d7a67b250607f0c3ea056fe056bcf91f5/neqo-transport/src/tracking.rs#L420-L424

mxinden (commented):
For what my review is worth, this makes sense to me. That said, I suggest waiting for another review, given the subtle nature of the change.

larseggert:
That last change seems to really have helped with `L1` and `C1`. I think this is ready for review.

neqo-transport/src/recovery/mod.rs
```diff
@@ -439,32 +439,33 @@ struct PtoState {
...
-            // For the Initial space, we only send one packet on PTO if we have not received any
-            // packets from the peer yet. This avoids sending useless PING-only packets
-            // when the Client Initial is deemed lost.
-            1
-        } else {
+    fn pto_packet_count(space: PacketNumberSpace) -> usize {
+        if space == PacketNumberSpace::ApplicationData {
             MAX_PTO_PACKET_COUNT
+        } else {
+            // For the Initial ahd Handshake spaces, we only send one packet on PTO. This avoids
```
larseggert, line 446:
```suggestion
            // For the Initial and Handshake spaces, we only send one packet on PTO. This avoids
```

neqo-transport/src/tracking.rs
```diff
@@ -290,7 +297,11 @@ impl RecvdPackets {
             ack_frequency_seqno: 0,
             ack_delay: DEFAULT_ACK_DELAY,
             unacknowledged_count: 0,
-            unacknowledged_tolerance: DEFAULT_ACK_PACKET_TOLERANCE,
+            unacknowledged_tolerance: if space == PacketNumberSpace::ApplicationData {
+                DEFAULT_ACK_PACKET_TOLERANCE
+            } else {
```
larseggert, line 302:
```suggestion
            } else {
                // ACK more aggressively
```

mxinden (approved):

Issue #1363, opened 2022-10-04
Crash in neqo_transport::connection::Connection::input_path


ddragana:
See https://crash-stats.mozilla.org/search/?release_channel=release&signature=%3Dneqo_transport%3A%3Aconnection%3A%3AConnection%3A%3Ainput_path&product=Firefox&date=%3E%3D2022-09-02T00%3A00%3A00.000Z&date=%3C2022-10-04T14%3A46%3A00.000Z&_facets=install_time&_facets=version&_facets=address&_facets=moz_crash_reason&_facets=reason&_facets=build_id&_facets=platform_pretty_version&_facets=signature&_facets=useragent_locale&_sort=-date&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#crash-reports

See bug [1750978](https://bugzilla.mozilla.org/show_bug.cgi?id=1750978)

PR #2029, opened 2024-08-02
fix: Check for error when looking up packet number space

larseggert:
The crash signatures don't really give much of an indication of where exactly the issue is. Am kinda guessing it's this `unwrap` of the packet number space which was based on unverified network input (via `PacketNumberSpace::from`).

Possible fix for #1363.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1542,7 +1542,12 @@ impl Connection {
 
                     qlog::packet_received(&self.qlog, &packet, &payload);
                     let space = PacketNumberSpace::from(payload.packet_type());
-                    if self.acks.get_mut(space).unwrap().is_duplicate(payload.pn()) {
+                    if self
+                        .acks
+                        .get_mut(space)
+                        .ok_or(Error::ProtocolViolation)?
+                        .is_duplicate(payload.pn())
+                    {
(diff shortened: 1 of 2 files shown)


mxinden (approved):
What do you think of adding a `debug!` in case `get_mut` returns `None`? Otherwise, a `Error::ProtocolViolation` is very difficult to trace back to its origin as a user.

Either way, looks good to me.

(On the matter of `unwrap` in general, since we have been turning on more and more static checks, we could take the next step and prohibit calls to `unwrap` across all non-test code via [Clippy's disallowed methods](https://rust-lang.github.io/rust-clippy/master/#/disallowed_method). In cases where we do need an `unwrap`, we can use `expect` with verbal proof why the case is impossible. Worked well in previous codebases I worked on.)

martinthomson (commented):
Test?

larseggert:
Will get to both of your suggestions next week.

@mxinden I like the idea of getting rid of `unwrap` in that way. It seems that `unwrap` is behind many of the crashes that get reported on the neqo code.

larseggert:
> Test?

So I'm struggling to write a test for this, which makes me wonder if the error is actually somewhere else...

larseggert:
@martinthomson still no luck at getting a test that hits this. But I think we should merge this if only in the interest of code hygiene.

PR #2031, opened 2024-08-02
refactor: consolidate NeqoQlog instantiation

mxinden:
Previously `neqo_transport::Server`, `neqo_bin::Client` and [`neqo_glue`](https://searchfox.org/mozilla-central/rev/7279a1df13a819be254fd4649e07c4ff93e4bd45/netwerk/socket/neqo_glue/src/lib.rs#283-315) had each their own but very similar way of instantiating `NeqoQlog`.

This commit consolidates the instantiation logic into a single function in `neqo_common`.


neqo-common/src/qlog.rs
```diff
@@ -29,21 +30,53 @@ pub struct NeqoQlogShared {
...
-        qlog_path: impl AsRef<Path>,
+    /// Will return `qlog::Error` if it cannot write to the new file.
+    pub fn enabled_with_file(
+        mut qlog_path: PathBuf,
+        role: Role,
+        title: Option<String>,
+        description: Option<String>,
+        file_prefix: impl Display,
     ) -> Result<Self, qlog::Error> {
+        qlog_path.push(format!("{file_prefix}.sqlog"));
```
mxinden, line 45: Previously we would use `.qlog` on the server and `.sqlog` on the client. As far as I can tell, the `s` stands for _streaming_. Thus I argue both should use `.sqlog`. Am I missing something?

https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#name-qlog-to-json-text-sequences

neqo-common/src/qlog.rs
```diff
@@ -29,21 +30,53 @@ pub struct NeqoQlogShared {
...
+
+        let file = OpenOptions::new()
+            .write(true)
+            // As a server, the original DCID is chosen by the client. Using
+            // create_new() prevents attackers from overwriting existing logs.
+            .create_new(true)
```
mxinden, line 51: Previously client would use `create`, thus potentially overriding an existing file. I can not think of a scenario where a client would want to override an existing qlog keyed by a cid. Thus I opted for using `create_new` for both client and server. Am I missing something?

https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create

https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new

neqo-common/src/qlog.rs
larseggert: No, you are correct.

larseggert (approved):

larseggert:
This has a test failure.

mxinden:
Debugging Windows failure right now.

larseggert:
Is there anything we can do to improve coverage?

mxinden:
It complains about the below not being covered:

``` rust
                .unwrap_or_else(|e| {
                    qerror!("failed to create NeqoQlog: {}", e);
                    NeqoQlog::disabled()
                })
```

https://github.com/mxinden/neqo/blob/35e00211272f5154bc794f8623c692c83b52239a/neqo-transport/src/server.rs#L265-L268

Given that our server is a test server only, I am inclined to say that this does not need to be tested. Not a strong opinion @larseggert. Also note that this wasn't covered before.

PR #2034, opened 2024-08-02
fix: Check whether CIDs are empty

### LINKED ISSUE #1429: Crash in [@ neqo_transport::connection::Connection::build_packet_header]
See https://bugzilla.mozilla.org/show_bug.cgi?id=1829989.

The crash reason is `called `Option::unwrap()` on a `None` value`.
Crash stack:

8 | libxul.so | core::panicking::panic | library/core/src/panicking.rs:114 | cfi
-- | -- | -- | -- | --
9 | libxul.so | neqo_transport::connection::Connection::build_packet_header | third_party/rust/neqo-transport/src/packet/mod.rs:0 | cfi
10 | libxul.so | neqo_transport::connection::Connection::output_close | third_party/rust/neqo-transport/src/connection/mod.rs:1850 | inlined
10 | libxul.so | neqo_transport::connection::Connection::output | third_party/rust/neqo-transport/src/connection/mod.rs:1760 | inlined
10 | libxul.so | neqo_transport::connection::Connection::process_output | third_party/rust/neqo-transport/src/connection/mod.rs:1004 | cfi
11 | libxul.so | neqo_http3::connection_client::Http3Client::process_output | third_party/rust/neqo-http3/src/connection_client.rs:850

It's not clear to me where `unwrap` is called inside function `build_packet_header`. Maybe @martinthomson knows?

larseggert:
Avoid `unwrap` calls when accessing connection IDs.

Fixes #1429

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1679,7 +1679,11 @@ impl Connection {
                 self.paths.make_permanent(path, None, cid);
                 Ok(())
             } else if let Some(primary) = self.paths.primary() {
-                if primary.borrow().remote_cid().is_none() {
+                if primary
+                    .borrow()
+                    .remote_cid()
+                    .map_or(true, |id| id.is_empty())
+                {
                     self.paths
                         .make_permanent(path, None, ConnectionIdEntry::empty_remote());
                     Ok(())
@@ -1931,18 +1935,17 @@ impl Connection {
         grease_quic_bit: bool,
     ) -> (PacketType, PacketBuilder) {
         let pt = PacketType::from(cspace);
-        let dcid = path
-            .remote_cid()
-            .map_or(Vec::new(), |id| id.as_ref().to_vec());
         let mut builder = if pt == PacketType::Short {
-            qdebug!("Building Short dcid {:?}", dcid);
-            PacketBuilder::short(encoder, tx.key_phase(), dcid)
+            qdebug!("Building Short dcid {:?}", path.remote_cid());
+            PacketBuilder::short(encoder, tx.key_phase(), &path.remote_cid())
         } else {
-            let scid = path
-                .local_cid()
-                .map_or(Vec::new(), |id| id.as_ref().to_vec());
-            qdebug!("Building {:?} dcid {:?} scid {:?}", pt, dcid, scid,);
-            PacketBuilder::long(encoder, pt, version, dcid, scid)
+            qdebug!(
+                "Building {:?} dcid {:?} scid {:?}",
+                pt,
+                path.remote_cid(),
+                path.local_cid(),
+            );
+            PacketBuilder::long(encoder, pt, version, &path.remote_cid(), &path.local_cid())
         };
         if builder.remaining() > 0 {
             builder.scramble(grease_quic_bit);
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1679,7 +1679,7 @@ impl Connection {
                 self.paths.make_permanent(path, None, cid);
                 Ok(())
             } else if let Some(primary) = self.paths.primary() {
-                if primary.borrow().remote_cid().is_empty() {
+                if primary.borrow().remote_cid().is_none() {
                     self.paths
                         .make_permanent(path, None, ConnectionIdEntry::empty_remote());
                     Ok(())
@@ -1908,7 +1908,7 @@ impl Connection {
                         // a packet on a new path, we avoid sending (and the privacy risk) rather
                         // than reuse a connection ID.
                         let res = if path.borrow().is_temporary() {
-                            assert!(!cfg!(test), "attempting to close with a temporary path");
+                            debug_assert!(!cfg!(test), "attempting to close with a temporary path");
                             Err(Error::InternalError)
                         } else {
                             self.output_path(&path, now, &Some(details))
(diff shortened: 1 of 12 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1931,18 +1931,18 @@ impl Connection {
         grease_quic_bit: bool,
     ) -> (PacketType, PacketBuilder) {
         let pt = PacketType::from(cspace);
+        let dcid = path
+            .remote_cid()
+            .map_or(Vec::new(), |id| id.as_ref().to_vec());
```
martinthomson, line 1936: You shouldn't need to allocate a new `Vec` here.  I would suggest changing the signature of `PacketBuilder::short`/`PacketBuilder::long` instead so that they take `Option<impl AsRef<[u8]>`.  You can call on that with `if let Some(...)` or `dcid.map_or(&[], AsRef::as_ref)` or `dcid.map_or(0, |d| d.as_ref().len())` as necessary.

martinthomson (changes requested):

KershawChang (approved):

neqo-transport/src/qlog.rs
```diff
@@ -104,8 +104,8 @@ fn connection_started(qlog: &NeqoQlog, path: &PathRef) {
             protocol: Some("QUIC".into()),
             src_port: p.local_address().port().into(),
             dst_port: p.remote_address().port().into(),
-            src_cid: Some(format!("{}", p.local_cid())),
-            dst_cid: Some(format!("{}", p.remote_cid())),
+            src_cid: p.local_cid().map(|id| format!("{id}")),
+            dst_cid: p.remote_cid().map(|id| format!("{id}")),
```
mxinden, line 108: Nit, only related to this pull request:

`ConnectionId` implements `Display`. Instead of going through the full string interpolation macro `format!`, one might as well just call `ToString::to_string`:



```suggestion
            src_cid: p.local_cid().map(ToString::to_string),
            dst_cid: p.remote_cid().map(ToString::to_string),
```

See also https://doc.rust-lang.org/std/string/trait.ToString.html.

Intuitively, I would expect `ToString` to be faster than `format!`. Though that would need a benchmark proof.

Feel free to ignore.

neqo-transport/src/packet/mod.rs
```diff
@@ -149,15 +149,19 @@ impl PacketBuilder {
     ///
     /// If, after calling this method, `remaining()` returns 0, then call `abort()` to get
     /// the encoder back.
-    pub fn short(mut encoder: Encoder, key_phase: bool, dcid: impl AsRef<[u8]>) -> Self {
+    pub fn short(mut encoder: Encoder, key_phase: bool, dcid: &Option<impl AsRef<[u8]>>) -> Self {
```
mxinden, line 152:
```suggestion
    pub fn short(mut encoder: Encoder, key_phase: bool, dcid: Option<impl AsRef<[u8]>>) -> Self {
```

A reference to an option containing a reference seems unnecessary. Would the above work?

neqo-transport/src/packet/mod.rs
```diff
@@ -185,20 +189,23 @@ impl PacketBuilder {
         mut encoder: Encoder,
         pt: PacketType,
         version: Version,
-        dcid: impl AsRef<[u8]>,
-        scid: impl AsRef<[u8]>,
+        dcid: &Option<impl AsRef<[u8]>>,
+        scid: &Option<impl AsRef<[u8]>>,
```
mxinden, line 193: Same as above.

```suggestion
        dcid: Option<impl AsRef<[u8]>>,
        scid: Option<impl AsRef<[u8]>>,
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1908,7 +1912,7 @@ impl Connection {
                         // a packet on a new path, we avoid sending (and the privacy risk) rather
                         // than reuse a connection ID.
                         let res = if path.borrow().is_temporary() {
-                            assert!(!cfg!(test), "attempting to close with a temporary path");
+                            debug_assert!(!cfg!(test), "attempting to close with a temporary path");
                             Err(Error::InternalError)
```
mxinden, line 1916: Bubbling  `Error::InternalError` up multiple levels will have it loose semantic context, in other words, we won't know what exactly failed from the outside.

I suggest either logging here, or, long term, we might want to consider embracing [Rust's Error backtraces](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.provide) via `thiserror` or [`snafu`](https://docs.rs/snafu/latest/snafu/). Though at least the former still depends on Rust Nightly.

mxinden (approved):
Couple of smaller comments. Overall this looks good to me.

neqo-transport/src/packet/mod.rs
larseggert: I tried that first, but then clippy told me `this argument is passed by value, but not consumed in the function body`.

larseggert: @mxinden any idea for how to fix this? (My Rust-fu isn't sufficient.)

mxinden: I am having difficulties reproducing the clippy failure on https://github.com/larseggert/neqo/pull/28.

@larseggert can you share your clippy version?

```
➜  cargo clippy --version
clippy 0.1.80 (0514789 2024-07-21)
```

larseggert: Same.

mxinden: Let's see what CI says to https://github.com/mozilla/neqo/pull/2034/commits/3903db63fecab90228656bee50e27ca3d68c8708.

larseggert: I think I worked around this with `take()`; please take a look?

mxinden (approved):

PR #2044, opened 2024-08-06
fix(udp): ignore empty datagram

mxinden:
When receiving an emtpy datagram `meta.len` and `meta.stride` would be `0`. Chunking the receive buffer via `.chunks(0)` would panic with:

```
chunk size must be non-zero
```

See also panic docs on `slice::chunks`:

https://doc.rust-lang.org/std/primitive.slice.html#method.chunks

With this commit `recv_inner` ignores the empty datagram.

In addition, under the assumption that an empty datagram is not a faulty event, `recv_inner` attempts another receive call on the socket. This ensures that the socket eventually returns `WouldBlock` and is thus registered for the next wake-up with the corresponding event loop (e.g. tokio).

---

Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1911658.

Instead of ignoring empty datagrams, should they be passed to `neqo_transport`, e.g. in order to be accounted for in `neqo_transport::ecn`?

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -13,7 +13,7 @@ use std::{
     slice,
 };
 
-use neqo_common::{qtrace, Datagram, IpTos};
+use neqo_common::{qdebug, qtrace, Datagram, IpTos};
 use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
 
 /// Socket receive buffer size.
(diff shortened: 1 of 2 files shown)


neqo-udp/src/lib.rs
```diff
@@ -52,22 +52,44 @@ pub fn send_inner(
...
+#[cfg(unix)]
+use std::os::fd::AsFd as SocketRef;
+#[cfg(windows)]
+use std::os::windows::io::AsSocket as SocketRef;
+
 pub fn recv_inner(
     local_address: &SocketAddr,
     state: &UdpSocketState,
-    socket: quinn_udp::UdpSockRef<'_>,
+    socket: impl SocketRef,
```
mxinden, line 63: When reading an empty datagram `recv_inner` attempts another receive from the socket. `state.recv` would previously consume the `quinn_udp::UdpSocketRef`. Passing an `impl SocketRef` here allows to construct a `quinn_udp::UdpSocketRef` before each call to `state.recv`.

larseggert:
> Instead of ignoring empty datagrams, should they be passed to `neqo_transport`, e.g. in order to be accounted for in `neqo_transport::ecn`?

When you say "empty datagram", do you mean a UDP packet without a payload? Or is it just that the receive call returns nothing? And if that, does it even return ancillary data?

mxinden:
> > Instead of ignoring empty datagrams, should they be passed to `neqo_transport`, e.g. in order to be accounted for in `neqo_transport::ecn`?
> 
> When you say "empty datagram", do you mean a UDP packet without a payload? Or is it just that the receive call returns nothing? And if that, does it even return ancillary data?

In this pull request I want to address the case of an "empty datagram" as in a UDP datagram without a payload, assuming that the ancillary data is present and valid @larseggert.

This should solve the issue in https://bugzilla.mozilla.org/show_bug.cgi?id=1911658. That still leaves us with the question why the sender is sending valid, but empty UDP datagrams.

Now there might be a second case, where the empty datagrams we are seeing in https://bugzilla.mozilla.org/show_bug.cgi?id=1911658 are not due to the sender sending empty datagrams, but due to an error somewhere within the receivers stack. A possibility, but no data to back it up today.

larseggert:
> In this pull request I want to address the case of an "empty datagram" as in a UDP datagram without a payload, assuming that the ancillary data is present and valid @larseggert.

Got it. If there is no QUIC header, it's not a QUIC packet and we should IMO not act on any ECN counts.

larseggert (approved):

Issue #2040, opened 2024-08-05
Release `neqo` `v0.8.2`


mxinden:
- Fix empty UDP datagram handling https://github.com/mozilla/neqo/pull/2044.
- I would like to enable [ECN support on Firefox Nightly](https://bugzilla.mozilla.org/show_bug.cgi?id=1902065). For that I need https://github.com/mozilla/neqo/pull/1964.

Any objections to me "uplifting" a new `neqo` version into mozilla-central? Since https://github.com/mozilla/neqo/pull/2025 I should have permissions to do so myself.

KershawChang:
> I would like to enable [ECN support on Firefox Nightly](https://bugzilla.mozilla.org/show_bug.cgi?id=1902065). For that I need #1964.
> 
> Any objections to me "uplifting" a new `neqo` version into mozilla-central? Since #2025 I should have permissions to do so myself.

No objections. Thanks for doing this!

mxinden:
I assume it is as easy as:

1. Bump the workspace `version` in the root `./Cargo.toml` through a pull request.
2. Once merged, tag the merge (or squash) commit and push the tag.
3. Vendor in mozilla-central as https://phabricator.services.mozilla.com/D217271 did via version bump and `./mach vendor rust`.

Am I missing something @KershawChang?

KershawChang:
> I assume it is as easy as:
> 
>     1. Bump the workspace `version` in the root `./Cargo.toml` through a pull request.
> 
>     2. Once merged, tag the merge (or squash) commit and push the tag.
> 

Note that when [creating a new tag](https://github.com/mozilla/neqo/releases/new), I always select `Set as a pre-release `, but I think it doesn't matter. 

>     3. Vendor in mozilla-central as https://phabricator.services.mozilla.com/D217271 did via version bump and `./mach vendor rust`.
> 
> 
> Am I missing something @KershawChang?

Yeah, that's all the steps.
Thanks.

mxinden:
For the record, mozilla-central _uplift_: https://phabricator.services.mozilla.com/D218637

mxinden:
> Note that when [creating a new tag](https://github.com/mozilla/neqo/releases/new), I always select `Set as a pre-release `, but I think it doesn't matter.

I tagged as pre-release. Though I don't understand why. Given that releases are used in production, i.e. Firefox, why tag them as pre-release? I suggest not doing so for future releases. Objections @KershawChang?

KershawChang:
[...]

PR #2045, opened 2024-08-06
chore: prepare v0.8.2 release

mxinden:
Fixes https://github.com/mozilla/neqo/issues/2040.

Depends on https://github.com/mozilla/neqo/pull/2044.


larseggert:
Do we want to land any other PRs?

mxinden:
No objections to including other PRs.

That said, given that _uplifting_ into mozilla-central is relatively easy, and given [Bug 1911658](https://bugzilla.mozilla.org/show_bug.cgi?id=1911658), I would suggest releasing as is, with the option of doing another release in the next couple of days.

larseggert:
WFM too

larseggert (approved):

PR #2046, opened 2024-08-09
test: Prep env for custom NSS build

larseggert:


test/test.sh
```diff
@@ -23,6 +23,11 @@ else
         iface=lo
 fi
 
+if [ "$NSS_DIR" ] && [ "$NSS_TARGET" ]; then
+        export LD_LIBRARY_PATH=$NSS_DIR/../dist/$NSS_TARGET/lib
+        export DYLD_FALLBACK_LIBRARY_PATH=$LD_LIBRARY_PATH
```
martinthomson, line 28:
```suggestion
        export LD_LIBRARY_PATH="$NSS_DIR/../dist/$NSS_TARGET/lib"
        export DYLD_FALLBACK_LIBRARY_PATH="$LD_LIBRARY_PATH"
```

test/test.sh
```diff
@@ -23,6 +23,11 @@ else
         iface=lo
 fi
 
+if [ "$NSS_DIR" ] && [ "$NSS_TARGET" ]; then
```
martinthomson, line 26: Huh, TIL.

 I would have used `[[ -n "$NSS_DIR" -a -n "$NSS_TARGET" ]]`.  I didn't even know that you could do `[ "$foo" ]`.  Is that even robust against the variable having a value of something like "!"?

martinthomson (approved):

test/test.sh
larseggert: https://www.shellcheck.net/wiki/SC2108

larseggert: I love shellcheck

larseggert: `[` is `test`.

martinthomson: Yeah, bash manual recommends `[[` over `[`/`test`.

The nice thing about shellcheck is that it is rarely wrong.  That's something to be treasured.

PR #2047, opened 2024-08-09
feat: Use `EnumMap` for `AckTracker`

larseggert:
Hopefully addresses @martinthomson's `TODO` about this.

### DIFF
--- a/neqo-transport/src/tracking.rs
+++ b/neqo-transport/src/tracking.rs
@@ -13,10 +13,9 @@ use std::{
     time::{Duration, Instant},
 };
 
-use enum_map::Enum;
+use enum_map::{enum_map, Enum, EnumMap};
 use neqo_common::{qdebug, qinfo, qtrace, qwarn, IpTosEcn};
 use neqo_crypto::{Epoch, TLS_EPOCH_HANDSHAKE, TLS_EPOCH_INITIAL};
-use smallvec::{smallvec, SmallVec};
 
 use crate::{
     ecn::EcnCount,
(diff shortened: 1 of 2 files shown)


neqo-transport/src/tracking.rs
```diff
@@ -550,34 +540,22 @@ impl ::std::fmt::Display for RecvdPackets {
...
-            PacketNumberSpace::Handshake => {
-                let sp = self.spaces.pop();
-                self.spaces.shrink_to_fit();
-                sp
-            }
-            PacketNumberSpace::ApplicationData => panic!("discarding application space"),
-        };
-        assert_eq!(sp.unwrap().space, space, "dropping spaces out of order");
+        self.spaces[space].take();
+        assert_ne!(
```
martinthomson, line 550: It's probably best to put this assertion before the destructive change.

There is another assertion:

```rust
if space == PacketNumberSpace::Handshake {
  assert!(self.spaces[PacketNumberSpace::Initial].is_none());
}
```

neqo-transport/src/tracking.rs
```diff
@@ -602,23 +580,34 @@ impl AckTracker {
...
         }
+
+        // Ignore any time that is in the past relative to `now`.
+        // That is something of a hack, but there are cases where we can't send ACK
+        // frames for all spaces, which can mean that one space is stuck in the past.
+        // That isn't a problem because we guarantee that earlier spaces will always
+        // be able to send ACK frames.
+        self.spaces
+            .iter()
+            .filter_map(|(_, recvd)| {
```
martinthomson, line 605: Try `self.spaces.values().flatten()` for the spaces.  The condition is probably simpler in that case.

martinthomson (approved):
My only reservation here is that the AckSpaces might end up taking more space than is idea.  The reason we chose to use SmallVec is that we'd be spilling onto the heap for the handshake, but it would be an inline allocation once the handshake is done.  This version looks like it will have a permanent memory overhead.  That overhead is probably trivial, but it might be good to confirm the actual cost.

PR #2050, opened 2024-08-11
Refactor NSS initialization

martinthomson:


neqo-crypto/src/lib.rs
```diff
@@ -175,23 +136,48 @@ pub fn init_db<P: Into<PathBuf>>(dir: P) -> Res<()> {
...
+    #[cfg(debug_assertions)]
+    enable_ssl_trace()?;
+
+    Ok(state)
+}
 
-        #[cfg(debug_assertions)]
-        enable_ssl_trace()?;
+/// Initialize NSS.  This only executes the initialization routines once, so if there is any chance
+/// that
```
mxinden, line 164: Only related: Is this unfinished sentence intentional?

neqo-crypto/src/lib.rs
```diff
@@ -175,23 +136,48 @@ pub fn init_db<P: Into<PathBuf>>(dir: P) -> Res<()> {
...
+    secstatus_to_res(unsafe {
+        p11::NSS_SetAlgorithmPolicy(
+            p11::SECOidTag::SEC_OID_XYBER768D00,
+            p11::NSS_USE_ALG_IN_SSL_KX,
+            0,
+        )
+    })?;
+
+    #[cfg(debug_assertions)]
+    enable_ssl_trace()?;
```
mxinden, line 158: I assume it is fine that we do this for `init` and `init_db` now, whereas previously we only did it for `init_db`.

mxinden (approved):
Thank you for the deduplication!

neqo-crypto/src/lib.rs
martinthomson: Not really.  I've fixed it though.

neqo-crypto/src/lib.rs
martinthomson: Yeah, it didn't seem like this was a useful difference, so I nuked it.  Obviously, the session cache thing doesn't work that way, so that is only available if we have a database.

PR #2057, opened 2024-08-19
test: Add a test case for 0-RTT loss

larseggert:
This currently fails, because we have a bug where the server does not accept retransmitted 0-RTT packets when it should.

### DIFF
--- a/neqo-transport/src/connection/tests/zerortt.rs
+++ b/neqo-transport/src/connection/tests/zerortt.rs
@@ -312,7 +312,7 @@ fn zero_rtt_loss_accepted() {
         assert_eq!(client_stream_id, server_stream_id.as_u64());
 
         // 0-RTT should be accepted
-        _ = client.process(si.as_dgram_ref(), now);
+        client.process_input(si.as_dgram_ref().unwrap(), now);
         let recvd_0rtt_reject = |e| e == ConnectionEvent::ZeroRttRejected;
         assert!(
             !client.events().any(recvd_0rtt_reject),
--- a/neqo-transport/src/connection/tests/zerortt.rs
+++ b/neqo-transport/src/connection/tests/zerortt.rs
@@ -279,6 +279,7 @@ fn zero_rtt_loss_accepted() {
         client.stream_send(client_stream_id, &[1, 2, 3]).unwrap();
         let mut ci = client.process(None, now);
         assert!(ci.as_dgram_ref().is_some());
+        assertions::assert_coalesced_0rtt(&ci.as_dgram_ref().unwrap()[..]);
 
         // Drop CI/0-RTT a number of times
         qdebug!("Drop CI/0-RTT {} extra times", i);
@@ -291,6 +292,15 @@ fn zero_rtt_loss_accepted() {
         // Process CI/0-RTT
         let si = server.process(ci.as_dgram_ref(), now);
         assert!(si.as_dgram_ref().is_some());
+
+        let server_stream_id = server
+            .events()
+            .find_map(|evt| match evt {
+                ConnectionEvent::NewStream { stream_id } => Some(stream_id),
+                _ => None,
+            })
+            .expect("should have received a new stream event");
+        assert_eq!(client_stream_id, server_stream_id.as_u64());
 
         // 0-RTT should be accepted
         _ = client.process(si.as_dgram_ref(), now);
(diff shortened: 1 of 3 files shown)


martinthomson:
It's not a bug, but a test configuration issue.  See larseggert/neqo#29.

larseggert:
So this also happens in the interop runner though - that inspired the test...

martinthomson:
That probably means that we need to be more generous with our tolerance to delay.  The default we have in our server is 10 seconds, which seems fine.  Are those tests that fail doing so because of being outside that window?

larseggert:
I'm attaching a log (taken locally) where this happens. The `ZeroRtt` packet is generated at `10s778ms` and `16s905ms` the server fails to accept it (after saying `AllowZeroRtt accepting 0-RTT`). 

The real bug here may be that the client never seems to retransmit the frame data in the 0-RTT packet (after printing `WARN Unhandled event ZeroRttRejected`). Is this something that moght have crept in during the move to `tokio`, @mxinden ?

(Even if this is the real bug, I think the server should still have accepted the 0-RTT...)

[output.txt](https://github.com/user-attachments/files/16669668/output.txt)

test-fixture/src/lib.rs
```diff
@@ -53,7 +53,7 @@ pub fn fixture_init() {
 
 // This needs to be > 2ms to avoid it being rounded to zero.
 // NSS operates in milliseconds and halves any value it is provided.
-pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_millis(10);
+pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_millis(10000);
```
martinthomson, line 56: I think that I would prefer if you changed the specific test to use a different anti-replay window instead.

This value is deliberately very tight so that tests fail when there are significant changes in RTT.  You can change the server to have a different window by setting the value explicitly.  I'll send you a PR.

martinthomson (changes requested):

test-fixture/src/lib.rs
martinthomson: https://github.com/larseggert/neqo/pull/30

neqo-transport/src/connection/tests/zerortt.rs
```diff
@@ -258,3 +258,65 @@ fn zero_rtt_update_flow_control() {
...
+        let mut server = resumed_server(&client);
+        let anti_replay = AntiReplay::new(earlier, WINDOW, 1, 3).unwrap();
+        server
+            .server_enable_0rtt(&anti_replay, AllowZeroRtt {})
+            .unwrap();
+
+        // Make CI/0-RTT
+        let client_stream_id = client.stream_create(StreamType::UniDi).unwrap();
+        client.stream_send(client_stream_id, &[1, 2, 3]).unwrap();
+        let mut ci = client.process(None, now);
```
martinthomson, line 289:
```suggestion
        let mut ci = client.process_output(now);
```

If you are only reading or writing (not both), this might be cleaner.

neqo-transport/src/connection/tests/zerortt.rs
```diff
@@ -258,3 +258,65 @@ fn zero_rtt_update_flow_control() {
...
+        assert!(ci.as_dgram_ref().is_some());
+        assertions::assert_coalesced_0rtt(&ci.as_dgram_ref().unwrap()[..]);
+
+        // Drop CI/0-RTT a number of times
+        qdebug!("Drop CI/0-RTT {i} extra times");
+        for _ in 0..i {
+            now += client.process(None, now).callback();
+            ci = client.process(None, now);
```
martinthomson, line 297:
```suggestion
            now += client.process_output(now).callback();
            ci = client.process_output(now);
```

Note that the `callback()` call could return a packet, which will also be dropped.

neqo-transport/src/connection/tests/zerortt.rs
```diff
@@ -258,3 +258,65 @@ fn zero_rtt_update_flow_control() {
...
+            .events()
+            .find_map(|evt| match evt {
+                ConnectionEvent::NewStream { stream_id } => Some(stream_id),
+                _ => None,
+            })
+            .expect("should have received a new stream event");
+        assert_eq!(client_stream_id, server_stream_id.as_u64());
+
+        // 0-RTT should be accepted
+        _ = client.process(si.as_dgram_ref(), now);
```
martinthomson, line 315:
```suggestion
        client.process_input(si.as_dgram_ref().unwrap(), now);
```

martinthomson (approved):
Nits only.

mxinden:
> Is this something that might have crept in during the move to `tokio`, @mxinden ?

Off the top of my head, I am not sure how, given that the move to `tokio` didn't touch the state machines itself, but only the IO paths on the outside. That said, it might have introduced a timing issue?

I can take a look. With which server does `neqo-latest` fail most reliably?

larseggert:
@mxinden I don't think it's tokio-related. See #2062.

Issue #2028, opened 2024-08-01
ci: test on Android


mxinden:
https://github.com/quinn-rs/quinn/pull/1950 adds a GitHub job, testing `quinn` on Android.

Neqo doesn't have much platform specific code other than:

https://github.com/mozilla/neqo/blob/c80630baa4c9d9a5fcab157efe8c1394f5204971/neqo-crypto/build.rs#L226-L229

That said, Neqo's dependencies have platform dependent code and Android is a [tier 1 Firefox build target](https://firefox-source-docs.mozilla.org/build/buildsystem/supported-configurations.html#supported-build-targets).

Objections to adding the same CI job to Neqo?

larseggert:
No objections. Is it much slower than our other CI tests?

mxinden:
Running all `quinn` unit tests takes about 5 minutes, including dependency installation, build and emulator startup.

larseggert:
Let's do a PR and see how long it takes.

Are there other targets we should be testing on?

PR #2058, opened 2024-08-19
ci: run tests on Android emulator

mxinden:
Pendant to https://github.com/quinn-rs/quinn/pull/1950.

Fixes https://github.com/mozilla/neqo/issues/2028.


larseggert:
https://github.com/mozilla/neqo/settings/actions under "General" has the list. Let me know if that is not visible to you.

larseggert:
@mxinden did you ask for the actions to be allowlisted? Should I?

larseggert:
@mxinden these actions should now be available: https://bugzilla.mozilla.org/show_bug.cgi?id=1931529

mxinden:
Cross-compilation of NSS for Android on Linux is failing. I will look into this.

larseggert:
It's probably because it is grabbing the precompiled Linux build for NSS from the action cache. We need to differentiate between Linux and Android when building/caching NSS.

larseggert:
We need to discuss with the NSS team how they build for Android. I have the suspicion that there are again no instructions, and just the way that the Firefox build does it is working. Topic for Thursday.

larseggert:
I'm giving up. The build system for NSS and esp. NSPR is just too horrible.

jschanck:
I'm not sure if this will be useful, but it looks like application-services builds NSS on android: https://github.com/mozilla/application-services/blob/main/libs/build-nss-android.sh.

larseggert:
OK, this finally builds NSS. It's now dying on MTU not supporting `android` yet, which is my next fix.

larseggert:
This is now at the point where the simulator finally executes the tests. There are **many** failures.

neqo-common/src/codec.rs
```diff
@@ -223,8 +223,8 @@ impl Encoder {
     ///
     /// When `len` doesn't fit in a `u64`.
     #[must_use]
-    pub fn vvec_len(len: usize) -> usize {
-        Self::varint_len(u64::try_from(len).expect("usize should fit into u64")) + len
+    pub fn vvec_len(len: u64) -> usize {
```
larseggert, line 226: `usize` is a `u32` on `i686-linux-android`. That required a few changes.

neqo-crypto/src/lib.rs
```diff
@@ -167,6 +167,9 @@ pub fn init() -> Res<()> {
 ///
 /// If NSS cannot be initialized.
 pub fn init_db<P: Into<PathBuf>>(dir: P) -> Res<()> {
+    // Allow overriding the NSS database path with one set via an environment variable.
+    let dir = env::var("NSS_DB_PATH")
+        .unwrap_or(dir.into().to_str().ok_or(Error::InternalError)?.to_string());
```
larseggert, line 172: I don't think there is a security issue with this, but paging @martinthomson just in case.

neqo-udp/src/lib.rs
```diff
@@ -268,6 +268,8 @@ mod tests {
             .expect("receive to succeed");
 
         // Assert that the ECN is correct.
+        // FIXME: This test currently fails in the Android emulator.
+        #[cfg(not(target_os = "android"))]
         assert_eq!(
             IpTosEcn::from(datagram.tos()),
             IpTosEcn::from(received_datagrams.next().unwrap().tos())
```
larseggert, line 275: @mxinden do we know if ECN w/`quinn-udp` is supposed to work on Android?

mxinden: Unfortunately, it depends on the Android API Level:

https://github.com/quinn-rs/quinn/pull/2079

larseggert: Good idea. Let me add that.

.github/actions/nss/action.yml
```diff
@@ -211,5 +216,40 @@ runs:
           OPT="-o"
           [ "${{ runner.os }}" != "Windows" ] && export CFLAGS="-ggdb3 -fno-omit-frame-pointer"
         fi
-        [ "$SCCACHE_CC" ] && [ "$SCCACHE_CXX" ] && export CC="$SCCACHE_CC" CXX="$SCCACHE_CXX"
-        $NSS_DIR/build.sh -g -Ddisable_tests=1 $OPT --static
+        if [[ $TARGET_PLATFORM == *-android* ]]; then
+          for file in build-nss-android.sh build-android-common.sh; do
+            curl -o "$file" -sSf "https://raw.githubusercontent.com/mozilla/application-services/refs/heads/main/libs/$file"
```
mxinden, line 221: Can you pin this to a recent `main` commit? I am afraid they might accidentally break our CI in the future.

.github/actions/rust/action.yml
```diff
@@ -52,11 +52,13 @@ runs:
...
     - name: Set up build environment
       shell: bash
+      env:
+        TARGETS: ${{ inputs.targets }}
       run: |
         {
           echo "CARGO_PROFILE_RELEASE_LTO=true"
           echo "CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1"
-          if [[ "${{ runner.os }}" == "Linux" && "${{ inputs.targets }}" == "" && "$(command -v mold)" ]]; then
+          if [[ "${{ runner.os }}" == "Linux" && "$TARGETS" == "" && "$(command -v mold)" ]]; then
```
mxinden, line 61: For my own understanding, why is `TARGET` better than `inputs.targets`?

neqo-common/src/codec.rs
mxinden: `vvec_len(x)` is always `> x` for all `x`, right?

Thus if 32 bits are not enough for `len`, they are also not enough for the function return value.

All that said, do we need to support such long `vvec`s?

neqo-crypto/src/lib.rs
```diff
@@ -167,6 +167,9 @@ pub fn init() -> Res<()> {
 ///
 /// If NSS cannot be initialized.
 pub fn init_db<P: Into<PathBuf>>(dir: P) -> Res<()> {
+    // Allow overriding the NSS database path with one set via an environment variable.
```
mxinden, line 170:
```suggestion
    // Allow overriding the NSS database path with an environment variable.
```

.github/actions/rust/action.yml
larseggert: `zizmor` thinks it allows code injection.

neqo-common/src/codec.rs
larseggert: > `vvec_len(x)` is always `> x` for all `x`, right?

Yes.
 
> Thus if 32 bits are not enough for `len`, they are also not enough for the function return value.
> 
> All that said, do we need to support such long `vvec`s?

I don't think so.

Is there a change you are suggesting?

mxinden: > I don't think so.

On a 32-bit platform, what happens if I call `vvec_len(u32::MAX as u64 + 1)`? The return value does not fit into a `usize`, i.e. 32bit, yet the function returns a `usize`. Am I missing something?

> Is there a change you are suggesting?

How about keeping `len: usize`, but handling exceeding `usize` on 32bit platforms at the caller side?

.github/actions/rust/action.yml
mxinden: Weird. Ok. Thanks for clarifying.

mxinden (commented):
Other than the change to `vvec_len`, this looks good to me.

Can't properly "Approve" as I initially opened this pull request (with a very minimal draft only).

neqo-common/src/codec.rs
larseggert: Can do.

Related: I was wondering if clippy had some things we could enable for making sure our code was 32bit-clean?

larseggert: @mxinden I reverted the changes I made to deal with 32bit platforms. Maybe those should be a separate PR.

mxinden: Only lint I can think of is `clippy::cast_possible_truncation`.

https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation

I think this pull request is a big step, finally running a 32bit platform on CI.

larseggert:
@mxinden take another look, I backed out the `vvec` changes. I can force merge if you think this is ready.

neqo-crypto/src/lib.rs
```diff
@@ -167,6 +167,9 @@ pub fn init() -> Res<()> {
 ///
 /// If NSS cannot be initialized.
 pub fn init_db<P: Into<PathBuf>>(dir: P) -> Res<()> {
+    // Allow overriding the NSS database path with an environment variable.
+    let dir = env::var("NSS_DB_PATH")
+        .unwrap_or(dir.into().to_str().ok_or(Error::InternalError)?.to_string());
```
mxinden, line 172: This is not used by Firefox, right?

I just want to make sure that third parties like some Antivirus vendors don't have yet another way to mess with our stack.

mxinden (commented):
Approved. Thanks.

neqo-crypto/src/lib.rs
larseggert: In Firefox, it's only used for tests: https://searchfox.org/mozilla-central/search?q=init_db%28&path=&case=false&regexp=false

neqo-crypto/build.rs
```diff
@@ -185,7 +185,11 @@ fn static_link() {
     // Dynamic libs that aren't transitively included by NSS libs.
     let mut other_libs = Vec::new();
     if env::consts::OS != "windows" {
-        other_libs.extend_from_slice(&["pthread", "dl", "c", "z"]);
+        other_libs.extend_from_slice(&["dl", "c", "z"]);
+        if env::var("CARGO_CFG_TARGET_OS").unwrap_or_default() != "android" {
```
martinthomson, line 189: Ordering matters sometimes.  Should this be moved up a line?

martinthomson (commented):
The NSS database thing doesn't bother me.  We don't run that code.

larseggert (approved):

PR #2059, opened 2024-08-19
refactor: drop time crate in favor of std::time::SystemTime

mxinden:
Follow-up to https://github.com/mozilla/neqo/pull/2053.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -11,6 +11,7 @@ use std::{
     io::BufWriter,
     path::PathBuf,
     rc::Rc,
+    time::SystemTime,
 };
 
 use qlog::{
@@ -152,6 +153,7 @@ impl Drop for NeqoQlogShared {
     }
 }
 
+#[allow(clippy::missing_panics_doc)]
 #[must_use]
 pub fn new_trace(role: Role) -> qlog::TraceSeq {
     TraceSeq {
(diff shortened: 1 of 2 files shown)


neqo-common/src/qlog.rs
```diff
@@ -173,11 +175,16 @@ pub fn new_trace(role: Role) -> qlog::TraceSeq {
...
                 // We can't do the obvious two-step conversion with f64::from(i32::try_from(...)),
                 // because that overflows earlier than is ideal.  This should be fine for a while.
                 #[allow(clippy::cast_precision_loss)]
-                Some((time::OffsetDateTime::now_utc().unix_timestamp() * 1000) as f64)
+                Some(
+                    SystemTime::now()
+                        .duration_since(SystemTime::UNIX_EPOCH)
+                        .expect("expect UNIX_EPOCH to always be earlier than now")
+                        .as_millis() as f64,
+                )
```
mxinden, line 187: I assume this is a valid assumption. Or, viewed from a different perspective, the best thing Neqo can do on a system running before `SystemTime::UNIX_EPOCH` is panic.

Happy to change to e.g. the below if you prefer.

```suggestion
                    SystemTime::now()
                        .duration_since(SystemTime::UNIX_EPOCH)
                        .map(|d| d.as_millis() as f64),
                        .ok()

```

martinthomson: I prefer the alternative. I might prefer that you use [this](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64) instead. `as_millis_f64` would be better, but it isn’t ready [yet](https://github.com/rust-lang/rust/issues/122451).

martinthomson (approved):
This is good, but we can avoid the cast.

larseggert (approved):

PR #2061, opened 2024-08-20
fix: Don't process timers during Closing or Draining

larseggert:
The current code doesn't seem to cause any packets to go out when closing, but is a bit wasteful (and also confused me when I saw timers fire in the logs...)


martinthomson (approved):
It might be more correct to let the idle timeout run.  That is, check for `self.state.error().closing()` on line ~968 and exit before doing the other maintenance stuff.

...
```rust
pub const fn closing(&self) -> bool {
  matches!(self, Self::Closing { .. } | Self::Draining { .. })
}
```

PR #2062, opened 2024-08-20
fix: Resend stream data on `ZeroRttRejected`

larseggert:
Make the neqo client re-initiate 0-RTT requests that were rejected by the server.

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -39,9 +39,7 @@ pub struct Handler<'a> {
     read_buffer: Vec<u8>,
 }
 
-impl<'a> super::Handler for Handler<'a> {
-    type Client = Connection;
-
+impl<'a> Handler<'a> {
     fn reinit(&mut self) {
         for url in self.handled_urls.drain(..) {
             self.url_queue.push_front(url);
@@ -49,6 +47,10 @@ impl<'a> super::Handler for Handler<'a> {
         self.streams.clear();
         self.all_paths.clear();
     }
+}
+
+impl<'a> super::Handler for Handler<'a> {
+    type Client = Connection;
 
     fn handle(&mut self, client: &mut Self::Client) -> Res<bool> {
         while let Some(event) = client.next_event() {
@@ -91,6 +93,7 @@ impl<'a> super::Handler for Handler<'a> {
                 }
                 ConnectionEvent::ZeroRttRejected => {
                     qdebug!("{event:?}");
+                    // All 0-RTT data was rejected. We need to retransmit it.
                     self.reinit();
                     self.download_urls(client);
                 }
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -76,7 +76,8 @@ impl<'a> super::Handler for Handler<'a> {
                 }
                 ConnectionEvent::StateChange(
                     State::WaitInitial | State::Handshaking | State::Connected,
-                ) => {
+                )
+                | ConnectionEvent::ZeroRttRejected => {
                     qdebug!("{event:?}");
                     self.download_urls(client);
                 }
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -31,6 +31,7 @@ use crate::STREAM_IO_BUFFER_SIZE;
 pub struct Handler<'a> {
     streams: HashMap<StreamId, Option<BufWriter<File>>>,
     url_queue: VecDeque<Url>,
+    handled_urls: Vec<Url>,
     all_paths: Vec<PathBuf>,
     args: &'a Args,
     token: Option<ResumptionToken>,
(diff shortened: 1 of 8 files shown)


larseggert:
This seems to be due to this line
https://github.com/mozilla/neqo/blob/e3d8be487989f8fd53a04cc25162bffff63c91ba/neqo-transport/src/streams.rs#L103-L104
which causes all streams to be blown away on the client when 0-RTT is rejected by the server.

Removing this line causes two `zero_rtt_send_reject` tests to fail, so this looks like blowing the streams (and their data) away is expected behavior? I'm kinda at a loss here why this is the case; any insights @martinthomson?

mxinden:
As far as I can tell, this is required by RFC 9001 @larseggert:

> When 0-RTT is rejected, all connection characteristics that the client assumed might be incorrect. This includes the choice of application protocol, transport parameters, and any application configuration. The client therefore MUST reset the state of all streams, including application state bound to those streams.

https://www.rfc-editor.org/rfc/rfc9001.html#section-4.6.2

larseggert:
Right. But at least I understood "reset the state of all streams" to mean "reset what you think has been sent/acked on each stream" and not "make it so the streams never existed".

**If** what we do is correct, we need the client to correctly handle this, i.e., manually RTX that data.

neqo-bin/src/client/http3.rs
```diff
@@ -222,7 +222,9 @@ impl<'a> super::Handler for Handler<'a> {
                     }
                 }
                 Http3ClientEvent::StateChange(Http3State::Connected)
-                | Http3ClientEvent::RequestsCreatable => {
+                | Http3ClientEvent::RequestsCreatable
+                | Http3ClientEvent::ZeroRttRejected => {
+                    qinfo!("{event:?}");
```
mxinden, line 227: :+1: good idea!

larseggert: I think for the interop runner, only h09 matters.

We should also check that Fx is doing the right thing.

KershawChang: In Fx, we rewind the stream to the state it was in before attempting 0RTT. The code that handles the ZeroRttRejected event is [here](https://searchfox.org/mozilla-central/rev/527d691a542ccc0f333e36689bd665cb000360b2/netwerk/protocol/http/Http3Session.cpp#608). Essentially, we reset the [offset of the request stream](https://searchfox.org/mozilla-central/rev/527d691a542ccc0f333e36689bd665cb000360b2/netwerk/protocol/http/nsHttpTransaction.cpp#3079) and [all internal states of HTTP3Stream](https://searchfox.org/mozilla-central/rev/527d691a542ccc0f333e36689bd665cb000360b2/netwerk/protocol/http/Http3Stream.cpp#465-479).

I think this behavior is correct.

mxinden: > I think for the interop runner, only h09 matters.

Same on h09, here I would expect the client to eventually call `self.download_urls(client)` given the match on:

``` rust
 ConnectionEvent::StateChange(
    State::WaitInitial | State::Handshaking | State::Connected,
 ) => {
```

martinthomson (commented):
Keep in mind that a rejection of 0-RTT could be due to a reduction in the number of streams that you are allowed to create.  You can't just rewind, you have to start over completely.

neqo-bin/src/client/http09.rs
```diff
@@ -76,7 +76,8 @@ impl<'a> super::Handler for Handler<'a> {
                 }
                 ConnectionEvent::StateChange(
                     State::WaitInitial | State::Handshaking | State::Connected,
-                ) => {
+                )
+                | ConnectionEvent::ZeroRttRejected => {
                     qdebug!("{event:?}");
                     self.download_urls(client);
```
martinthomson, line 82: I don't think that this is correct.  You will need to reset the state of each of the streams as well.

My understanding is that if we have sent a request, we track that.  So if a request is partially sent, we need to resend.

(Maybe h09 doesn't do that, but h3 might have to.)

neqo-bin/src/client/http3.rs
larseggert: I think the issue (for both h3 and h09) is that already-tried URLs are popped off the list and are not retried. I have a fix for this now in the PR; CI is running.

neqo-bin/src/client/http09.rs
```diff
@@ -80,6 +89,11 @@ impl<'a> super::Handler for Handler<'a> {
...
                 }
+                ConnectionEvent::ZeroRttRejected => {
```
mxinden, line 92: Maybe add a comment here why all pending requests need to be retried?

neqo-bin/src/client/mod.rs
```diff
@@ -351,6 +351,7 @@ trait Handler {
 
     fn handle(&mut self, client: &mut Self::Client) -> Res<bool>;
     fn take_token(&mut self) -> Option<ResumptionToken>;
+    fn reinit(&mut self);
```
mxinden, line 354: `reinit` is always called on the concrete h09 or h3 Handler, never on the `Handler` trait, right? If so, I suggest removing the trait extension here and instead implement `reinit` on the concrete Handler types (e.g. `impl <'a> Handler<'a> { fn reinit(&mut self) {`) instead.

Not a strong opinion.

mxinden (approved):
Change looks good to me. Thank you for the debugging work!

mxinden (approved):

PR #2064, opened 2024-08-20
fix: Turn off PMTUD during QNS `zerortt` test

larseggert:
PMTUD probes inflate what we sent in 1-RTT, causing QNS to fail the test.

See https://github.com/ptrd/kwik/issues/46#issuecomment-2299061674

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -252,6 +252,8 @@ impl Args {
                 }
                 self.shared.use_old_http = true;
                 self.resume = true;
+                // PMTUD probes inflate what we sent in 1-RTT, causing QNS to fail the test.
+                self.shared.quic_parameters.no_pmtud = true;
             }
             "multiconnect" => {
                 self.shared.use_old_http = true;


mxinden (approved):
Do I understand correctly that it is these lines that fail?

https://github.com/quic-interop/quic-interop-runner/blob/b15dfbba908e5f2962715c3c3edaa40c1e79d6b8/testcases.py#L698-L700

I am fine with the fix here. In case the required changes in quic-interop-runner would be small, I would prefer the calculation in quic-interop-runner to be updated instead.

martinthomson (approved):

larseggert:
@mxinden yes, that is the failing calculation. What would the fix for QNS be? AFAIK the calculation is some heuristic that @marten-seemann came up with.

marten-seemann:
Any ideas how we can improve the heuristics? The problem is, retransmitting already transmitted STREAM data in a PMTU packet is a totally reasonable strategy...

larseggert:
I'll merge this now. I suggest opening a new issue for QNS to come up with a better heuristic.

PR #2066, opened 2024-08-21
chore: Fix various issues that markdownlint found

larseggert:


docs/linux_build.md
mxinden, line 1: Can this file be removed? All of it is either mentioned in the root README.md, or very outdated.

mxinden (approved):

docs/linux_build.md
larseggert: I asked the same before, but I think I remember that @KershawChang wanted to keep it.

larseggert: I opened 2069. Will merge this one now; we can discuss there what to do with that file.

PR #2070, opened 2024-08-22
fix: Reduce log level for QNS `transfer` and `blackhole`

larseggert:
These are just way too noisy otherwise, making artifact download slow.


larseggert:
Closed in favor of #2071.

PR #2071, opened 2024-08-22
ci: Empty all QNS logs >5MB before artifact upload

larseggert:
Because these artifacts are huge and those logs are not useful for debugging.


mxinden (approved):
Works for me.

Preference for replacing the content of these files with something like "deleted as it is larger than 5MB" for future us, forgetting about the delete step. Feel free to ignore.

Issue #2068, opened 2024-08-21
expose ECN counts in connection `Stats`


mxinden:
I suggest we add ECN counts to the `Connection`'s `Stats` `struct`.

https://github.com/mozilla/neqo/blob/13237c3f5a4740b81ff98094fbdf663253122c3a/neqo-transport/src/stats.rs#L113-L115

We can then expose these as Firefox Telemetry metrics through `neqo_glue` `Http3Stats`:

https://searchfox.org/mozilla-central/rev/e942f7bc56cd103bba86b396ddeba5b1ab04f1a4/netwerk/socket/neqo_glue/src/lib.rs#1548-1550

I was thinking of simply adding:

``` rust
/// Count derived from QUIC ACKs.
ecn_tx: EcnCount,
/// Count derived from IP's TOS.
ecn_rx: EcnCount,
```

Would you prefer exposing other metrics or alternative representations? Can you think of better names?

On the Firefox side we can then track metrics like ECT-0 / ECN-CE ratio in a Histogram on connection closing.

Before I tackle this, was it deliberately not done in https://github.com/mozilla/neqo/pull/1678 @larseggert? Any objections?

Would allow us to measure https://bugzilla.mozilla.org/show_bug.cgi?id=1902065.

larseggert:
We can, but wouldn't we (maybe in addition) like to add telemetry about whether ECN was validated (or not) for the paths used during the lifetime of a connection? Aggregated mark counts over all paths are unlikely to let us derive this based on the telemetry.

> Before I tackle this, was it deliberately not done in #1678 @larseggert? Any objections?

I didn't think about what kind of telemetry we wanted when I did the original PR, so no, the omission wasn't deliberate.

PR #2072, opened 2024-08-22
feat(transport): add ecn metrics to Stats

mxinden:
Adding 4 ECN metrics to `neqo-transport` `Stats`:

``` rust
    /// Number of paths known to be ECN capable.
    pub ecn_paths_capable: usize,
    /// Number of paths known to be ECN incapable.
    pub ecn_paths_not_capable: usize,
    /// ECN counts for outgoing UDP datagrams, returned by remote through QUIC ACKs.
    pub ecn_tx: EcnCount,
    /// ECN counts for incoming UDP datagrams, read from IP TOS header.
    pub ecn_rx: EcnCount,
```

---

Fixes https://github.com/mozilla/neqo/issues/2068.

Corresponding Phabricator Differential: https://phabricator.services.mozilla.com/D220062

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1455,7 +1455,9 @@ impl Connection {
     ) {
         let space = PacketNumberSpace::from(packet.packet_type());
         if let Some(space) = self.acks.get_mut(space) {
-            *space.ecn_marks() += d.tos().into();
+            let space_ecn_marks = space.ecn_marks();
+            *space_ecn_marks += d.tos().into();
+            self.stats.borrow_mut().ecn_rx = *space_ecn_marks;
         } else {
             qtrace!("Not tracking ECN for dropped packet number space");
         }
@@ -2422,7 +2424,10 @@ impl Connection {
                 self.loss_recovery.on_packet_sent(path, initial);
             }
             path.borrow_mut().add_sent(packets.len());
-            Ok(SendOption::Yes(path.borrow_mut().datagram(packets)))
+            Ok(SendOption::Yes(
+                path.borrow_mut()
+                    .datagram(packets, &mut self.stats.borrow_mut()),
+            ))
         }
     }
 
--- a/neqo-transport/src/connection/tests/ecn.rs
+++ b/neqo-transport/src/connection/tests/ecn.rs
@@ -12,11 +12,11 @@ use test_fixture::{
     fixture_init, now, DEFAULT_ADDR_V4,
 };
 
-use super::{send_something_with_modifier, DEFAULT_RTT};
 use crate::{
     connection::tests::{
         connect_force_idle, connect_force_idle_with_modifier, default_client, default_server,
-        handshake_with_modifier, migration::get_cid, new_client, new_server, send_something,
+        handshake_with_modifier, migration::get_cid, new_client, new_server, send_and_receive,
+        send_something, send_something_with_modifier, DEFAULT_RTT,
     },
     ecn::ECN_TEST_COUNT,
     ConnectionId, ConnectionParameters, StreamType,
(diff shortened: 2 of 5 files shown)


mxinden:

Below is a screenshot of a local Firefox `about:telemetry` page, showing the two new metrics:

- A Scalar `ecn_paths_capability` keyed by `capable` and `not-capable`.
- A Histogram `HTTP3_ECN_CE_ECT0_RATIO` keyed by `tx` and `rx`, multiplied by a factor of `10_000`.

Screenshot 2024-08-24 at 15-48-29 Telemetry Data

larseggert (commented):
One thing we need to check is whether `EcnCount[NotEct]` is correctly set by the current code in all cases, because that is not included in a received ACK-ECN frame.

mxinden:
> One thing we need to check is whether `EcnCount[NotEct]` is correctly set by the current code in all cases

`EcnCount[NotEct]` is **not** correctly set for outgoing packets where marks are received through a QUIC ACK-ECN frame.

Would it be valid to derive `NotEct` as `largest_acked - ECT0 - ECT1 - ECN_CE` @larseggert?

Note that I am not making use of `EcnCount[NotEct]` in Firefox. That said, I agree that we should not expose an invalid `NotEct` value, that is 0.

larseggert:
I'm wondering if we can omit `EcnCount[NotEct]` when we report, or even completely. `enum_map` can't handle that though AFAICT.

larseggert:
> In case this derivative is valid, I would prefer, for the sake of completeness, exposing all 4. In case it is not, I prefer going with the first alternative proposed above.

I'm not sure that it is. Can we actually count them?

mxinden:
> > In case this derivative is valid, I would prefer, for the sake of completeness, exposing all 4. In case it is not, I prefer going with the first alternative proposed above.
> 
> I'm not sure that it is. Can we actually count them?

As in _count how many `NotEct` datagrams we send_? I am not sure combining two measurement sources, i.e. (1) marks **received** through QUIC ACK-ECN and (2) the datagrams **send** is a good idea.

Say that a middlebox converts every `Ect0` to `NotEct`. Within the ECN validation timeframe, the local node would only send `Ect0` datagrams, while the remote node would only receive `NotEct`. The local node would then report an ECN count of 0 `Ect0` (derived from QUIC ACK-ECN) and 0 `NotEct` (derived from counting sent datagrams).

larseggert:
Ah, sorry. This is all about reporting received marks. For received marks, the `NotEct` counts should always be correct, no?

mxinden:
This pull request handles both:

1. ECN counts for outgoing UDP datagrams, returned by remote through QUIC ACKs.
2. ECN counts for incoming UDP datagrams, read from IP TOS header.

Our current discussion is about (1), i.e. the `EcnCounts` derived from QUIC ACK-ECN. Given that a QUIC ACK-ECN does not carry `NotEct`, the current `EcnCount` for (1) as implemented here is not correct, that is, it will always report 0 for `NotEct`.

larseggert:
I think we should make it clear that the count is unknown then. Maybe with an `Option` or is this too unwieldy?

mxinden:
I see 3 ways forward here:

1. Derive `EcnCount[NotEct]` based on the largest acknowledged packets.
2. Encode the fact that `EcnCount[NotEct]` is `0` at the type system level, e.g. via a custom `EcnCount` struct, or an `Option`.
3. Document the fact that `EcnCount[NotEct]` is `0` for outgoing UDP datagrams.

As discussed above (1) might not always be accurate. After some experimentation in my eyes (2) isn't worth the additional complexity. With https://github.com/mozilla/neqo/pull/2072/commits/512debae64018cd5ee29853315fe8c240aeabe64 I opted for (3), especially since `EcnCount[NotEct]` is never used at the consumer side, i.e. Firefox.

@larseggert is (3), i.e. https://github.com/mozilla/neqo/pull/2072/commits/512debae64018cd5ee29853315fe8c240aeabe64, a valid option for you?

larseggert:
I think 3 is fine for this PR. We can revisit if we need to.

larseggert (approved):

larseggert:
@mxinden fix the conflict, please?

mxinden:
Will fix the Rust doc failures in a bit.

PR #2076, opened 2024-08-27
perf: don't allocate in UDP recv path

mxinden:
Previously `neqo-udp` would have one long-lived receive buffer, but after reading into the buffer from the socket, it would allocate a new `Vec` for each UDP segment.

This commit does not allocate each UDP segment in a new `Vec`, but instead passes the single re-used receive buffer to
`neqo_transport::Connection::process_input` directly.

---

Draft for now. Want to see benchmark results before investing further.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -136,5 +136,6 @@ jobs:
 
   bench:
     name: "Benchmark"
-    needs: [check]
+    # TODO
+    # needs: [check]
     uses: ./.github/workflows/bench.yml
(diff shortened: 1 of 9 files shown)


martinthomson (commented):
Is it possible to change the code so that input datagrams are taken with a reference to the underlying buffer instead?

You might need to split Datagram in two parts to do that, one with an owned buffer and one with a borrowed one (maybe; for output).  Ideally though we never have to own the buffer.

This change seems like it would be harder to undo that doing this right from the outset.

mxinden:
Thank you for taking a look @martinthomson.

> Is it possible to change the code so that input datagrams are taken with a reference to the underlying buffer instead?

Yes. I created https://github.com/mozilla/neqo/pull/2093 implementing your suggestion above. It is a _Draft_ for now, but I would argue goes beyond a proof-of-concept.

Closing here in favor of https://github.com/mozilla/neqo/pull/2093.

PR #1964, opened 2024-07-06
feat(transport): Use a shorter ECN probe threshold initially

mxinden:
This commit adds a test where a client connects to a server over a connection dropping all ECN marked datagrams (ECN blackhole) in both directions, asserting 43 RTT to detect ECN blackhole, disable ECN and eventually establish connection. Test assumes 20 RTT for client to detect blackhole, 20 RTT for server to detect blackhole and 3 RTT for handshake to be confirmed.

---

Test for https://github.com/mozilla/neqo/issues/1925.

How likely are ECN blackholes on today's Internet? Should this block [ECN roll-out on Firefox](https://phabricator.services.mozilla.com/D212959)? //CC @larseggert 

[Bauer et. alia. in "Measuring the State of ECN Readiness in Servers, Clients, and Routers"](https://conferences.sigcomm.org/imc/2011/docs/p171.pdf) suggests that ECN blackholes are rare.

<details>
  <summary>Test output</summary>
  
  
  ```
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Client at 0ms
0s  0ms DEBUG ECN probing: sent 1 probes
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Server at 50ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = WaitInitial
Processing Client at 100ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Server at 150ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = WaitInitial
Processing Client at 200ms
0s  0ms DEBUG ECN probing: sent 2 probes
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Server at 250ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = WaitInitial
Processing Client at 300ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Server at 350ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = WaitInitial
Processing Client at 400ms
0s  0ms DEBUG ECN probing: sent 3 probes
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = Init
Processing Server at 450ms
[neqo-transport/src/connection/tests/mod.rs:190:13] c.state() = WaitInitial
Processing Client at 500ms
[...]

PR #2081, opened 2024-08-30
fix: Handle migrating to a path that is an ECN blackhole

larseggert:
This will send `MAX_PATH_PROBES` probes with ECN and then another `MAX_PATH_PROBES` probes without ECN, so with `MAX_PATH_PROBES = 3`, a total of 6 for a non-functional path. Is that OK?

### DIFF
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -35,6 +35,9 @@ use crate::{
 };
 
 /// The number of times that a path will be probed before it is considered failed.
+///
+/// Note that with [`crate::ecn`], a path is probed [`MAX_PATH_PROBES`] with ECN
+/// marks and [`MAX_PATH_PROBES`] without.
 pub const MAX_PATH_PROBES: usize = 3;
 /// The maximum number of paths that `Paths` will track.
 const MAX_PATHS: usize = 15;
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -15,7 +15,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use neqo_common::{hex, qdebug, qinfo, qlog::NeqoQlog, qtrace, Datagram, Encoder, IpTos};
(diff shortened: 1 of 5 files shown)


mxinden (approved):
This looks good to me. Thank you for the patch.

As discussed before, we could as well do ECN validation after connection establishment and path validation. Assuming that ECN blackholes are rare, the only upside would be simpler code. Downside would be delay in ECN signals. I assume ECN signals early on is still worth the additional code complexity, correct?

A couple of minor suggestions below.

neqo-transport/src/path.rs
```diff
@@ -749,15 +764,25 @@ impl Path {
...
+                qinfo!(
+                    [self],
+                    "Possible ECN blackhole, disabling ECN and re-probing path"
+                );
+                self.ecn_info.disable_ecn(stats);
+                ProbeState::ProbeNeeded { probe_count: 0 }
+            } else {
+                qinfo!([self], "Probing failed");
+                ProbeState::Failed
+            }
```
mxinden, line 785: Question for my own understanding:

- Ideally we would bundle all ECN related logic in `neqo_transport::ecn`.
- `neqo_transport::ecn` already has the notion of failing ECN validation early, i.e. on connection establishment. See `ECN_TEST_COUNT_INITIAL_PHASE` introduced in https://github.com/mozilla/neqo/pull/1964.
- It would be nice if we wouldn't have two cases of early ECN validation failure logic, (1) for connection establishment and (2) for path validation. But instead for both to be handled by the same logic.
- (1) is triggered by a loss-recovery timeout, determining the ECN marked packets as lost (see `EcnInfo::on_packets_lost`).
- (2) can not be triggered by loss-recovery, given that there is no timeout mechanism for non-primary paths, thus probe packets are never determined as lost and thus `EcnInfo::on_packets_lost` is never called.

Is that correct @larseggert?

larseggert: Yes. I agree that a consolidated approach would be nicer.

mxinden (approved):
Thanks for the follow-ups!

PR #2085, opened 2024-09-02
ci: Only build Firefox for non-draft PRs

larseggert:
Save some trees


KershawChang (approved):

martinthomson:
Want to add the QNS runs to that as well?

larseggert:
Can do, but I actually look at those even for my draft PRs.

PR #2089, opened 2024-09-06
ci: Retry the Rust cache

larseggert:
And maybe buildjet.


larseggert:
Still slower than not caching :-(

PR #2090, opened 2024-09-06
chore: Trim unused features

larseggert:
Courtesy of https://github.com/ToBinio/cargo-features-manager.

### DIFF
--- a/neqo-bin/Cargo.toml
+++ b/neqo-bin/Cargo.toml
@@ -44,7 +44,7 @@ tokio = { version = "1", default-features = false, features = ["net", "time", "m
 url = { version = "2.5", default-features = false }
 
 [dev-dependencies]
-criterion = { version = "0.5", default-features = false }
+criterion = { version = "0.5", default-features = false, features = ["async_tokio"] }
 tokio = { version = "1", default-features = false, features = ["sync"] }
 
 [features]
--- a/neqo-bin/Cargo.toml
+++ b/neqo-bin/Cargo.toml
@@ -27,7 +27,7 @@ workspace = true
 
 [dependencies]
 # neqo-bin is not used in Firefox, so we can be liberal with dependency versions
-clap = { version = "4.4", default-features = false, features = ["std", "help", "usage", "error-context", "suggestions"] }
+clap = { version = "4.4", default-features = false, features = ["std", "help", "usage", "error-context", "suggestions", "derive"] }
 clap-verbosity-flag = { version = "2.2", default-features = false }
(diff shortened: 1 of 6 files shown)


neqo-bin/Cargo.toml
```diff
@@ -27,7 +27,7 @@ workspace = true
 
 [dependencies]
 # neqo-bin is not used in Firefox, so we can be liberal with dependency versions
-clap = { version = "4.4", default-features = false, features = ["std", "help", "usage", "error-context", "suggestions", "derive"] }
+clap = { version = "4.4", default-features = false, features = ["std", "help", "usage", "error-context", "suggestions"] }
```
mxinden, line 30: Isn't this used here?

https://github.com/mozilla/neqo/blob/28f60bd0ba3209ecba4102eec123859a3a8afd45/neqo-bin/src/lib.rs#L32

larseggert: Weird. It seems to compile anyway.

mxinden: `cargo` does _feature unification_. In other words, if two crates import `clap` and one enables the `derive`, the `derive` feature is enabled for both.

https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

Looking at the crates in our tree importing `clap` and `clap_derive` we see e.g. `clap-verbosity-flag`:

```
➜  neqo git:(main) ✗ cargo tree -e features --invert clap_derive
clap_derive v4.5.4 (proc-macro)
└── clap_derive feature "default"
    └── clap v4.5.4
        ├── clap feature "derive"
        │   └── clap-verbosity-flag v2.2.0
        │       └── neqo-bin v0.9.0 (/home/mxinden/code/github.com/mozilla/neqo/neqo-bin)
        │           └── neqo-bin feature "default" (command-line)
        ├── clap feature "error-context"
        │   └── neqo-bin v0.9.0 (/home/mxinden/code/github.com/mozilla/neqo/neqo-bin) (*)
```

Looking at `clap-verbosity-flag`'s `Cargo.toml`, it does import `clap` with the `derive` feature:

https://github.com/clap-rs/clap-verbosity-flag/blob/c621a6a8a7c0b6df8f1464a985a5d076b4915693/Cargo.toml#L117

Thus the `clap` `derive` feature is enabled for both `clap-verbosity-flag` and `neqo-bin`, even when we remove the `derive` feature in `neqo-bin` as above.

Thereby I don't think we should remove the `derive` feature above.

mxinden (approved):

PR #2093, opened 2024-09-08
perf: don't allocate in UDP send & recv path

mxinden:
### Description

This change is best summarized by the `process` function signature.

On `main` branch the `process` function looks as such:

```rust
```

- It takes as **input** an optional reference to a `Datagram`. That `Datagram` owns an allocation of the UDP payload, i.e. a `Vec<u8>`. Thus for each incoming UDP datagram, its payload is allocated in a new `Vec`.
- It returns as **output** an owned `Output`. Most relevantly the `Output` variant `Output::Datagram(Datagram)` contains a `Datagram` that again owns an allocation of the UDP payload, i.e. a `Vec<u8>`. Thus for each outgoing UDP datagram too, its payload is allocated in a new `Vec`.

This commit changes the `process` function to:

```rust
pub fn process_into_buffer<'a>(
    &mut self,
    input: Option<Datagram<&[u8]>>,
    now: Instant,
    write_buffer: &'a mut Vec<u8>,
) -> Output<&'a [u8]> {
```

- It takes as **input** an optional `Datagram<&[u8]>`. But contrary to before, `Datagram<&[u8]>` does not own an allocation of the UDP payload, but represents a view into a long-lived receive buffer containing the UDP payload.
- It returns as **output** an `Output<&'a [u8]>` where the `Output::Datagram(Datagram<&'a [u8]>)` variant does not own an allocation of the UDP payload, but here as well represents a view into a long-lived write buffer the payload is written into. That write buffer lives outside of `neqo_transport::Connection` and is provided to `process_into_buffer` as `write_buffer: &'a mut Vec<u8>`. Note that both `write_buffer` and `Output` use the lifetime `'a`, i.e. the latter is a view into the former.

This change to the `process` function enables the following:

1. A user of `neqo_transport` (e.g. `neqo_bin`) has the OS write incoming UDP datagrams into a long-lived receive buffer (via e.g. `recvmmsg`).
2. They pass that receive buffer to `neqo_transport::Connection::process_into_buffer` along with a long-lived write buffer.
3. `process_into_buffer` reads the UDP datagram from the long-lived receive buffer through the `Datagram<&[u8]>` view and writes outgoing datagrams into the provided long-lived `write_buffer`, returning a view into said buffer via a `Datagram<&'a [u8]>`.
4. The user, after having called `process_into_buffer` can then pass the write buffer to the OS (e.g. via `sendmsg`).

To summarize a user can receive and send UDP datagrams, without allocation in the UDP IO path.

As an aside, the above is compatible with GSO and GRO, where a send and receive buffer contains a consecutive number of UDP datagram segments.

### Performance impact

Early benchmarks are promising, showing e.g. a 10% improvement in the _Download_ benchmark, and up to 40% improvement in the _neqo-neqo-reno-pacing_ benchmark.

#### Current `main` for comparison

> <details><summary>1-conn/1-100mb-resp (aka. Download)/client: :green_heart: <b>Performance has improved.</b></summary><pre>
>        time:   [112.72 ms 113.13 ms 113.51 ms]
>        thrpt:  [880.95 MiB/s 883.97 MiB/s 887.14 MiB/s]
> change:
>        time:   [-2.1601% <b>-1.6758%</b> -1.1570%] (p = 0.00 < 0.05)
>        thrpt:  [+1.1705% <b>+1.7044%</b> +2.2078%]
>        
>   2 (2.00%) low mild</pre></details>
> 
> ##### Client/server transfer results
> Transfer of 33554432 bytes over loopback.
> | Client | Server | CC | Pacing | Mean [ms] | Min [ms] | Max [ms] | Relative |
> |:---|:---|:---|:---|---:|---:|---:|---:|
> | neqo | neqo | reno | on | 260.8 ± 159.4 | 127.4 | 691.6 | 1.00 |
> | neqo | neqo | reno |  | 221.2 ± 90.7 | 139.6 | 432.0 | 1.00 |
> | neqo | neqo | cubic | on | 214.5 ± 87.0 | 125.0 | 375.0 | 1.00 |
> | neqo | neqo | cubic |  | 236.2 ± 118.1 | 136.8 | 540.0 | 1.00 |

https://github.com/mozilla/neqo/actions/runs/10850817785

### Pull request status

This pull request is ready for review.

Replaces https://github.com/mozilla/neqo/pull/2076.
Closes https://github.com/mozilla/neqo/issues/1922.


neqo-common/src/codec.rs
```diff
@@ -193,13 +193,20 @@ impl<'a, 'b> PartialEq<Decoder<'b>> for Decoder<'a> {
...
 }
 
-impl Encoder {
+impl<'a> Encoder<'a> {
+    pub fn new_with_buffer(buf: &'a mut Vec<u8>) -> Self {
+        Self { buf }
+    }
+}
+
+impl<'a> Encoder<'a> {
```
martinthomson, line 209:
```suggestion
```

neqo-common/src/datagram.rs
```diff
@@ -22,10 +26,43 @@ impl Datagram {
...
+            src,
+            dst,
+            tos,
+            segment_size: None,
+            d,
+        }
+    }
+
+    // TODO: Rename
+    pub fn new_2_with_segment_size(
```
martinthomson, line 48: This bit doesn't quite match your PR comments.

neqo-transport/src/connection/mod.rs
```diff
@@ -1137,6 +1173,37 @@ impl Connection {
         output
     }
 
+    // TODO: For the lack of a better name, `process_into` for now.
+    /// Process input and generate output.
+    #[must_use = "Output of the process function must be handled"]
+    pub fn process_into<'a>(
```
martinthomson, line 1179: This doesn't match your commentary either.  Did you forget to push something?

neqo-common/src/datagram.rs
mxinden: After a `recvmmsg` into a long-lived `recv_buf: Vec<u8>`, a user can instantiate a `Datagram<&'a [u8]>` via `new_2_with_segment_size`, where `d: &'a [u8]` is a view into the `recv_buf` and `segment_size: usize` is the GRO segment size returned by `recvmmsg`.

``` rust
    pub fn new_2_with_segment_size(
        src: SocketAddr,
        dst: SocketAddr,
        tos: IpTos,
        segment_size: usize,
        d: &'a [u8],
    ) -> Self {
        Self {
            src,
            dst,
            tos,
            segment_size: Some(segment_size),
            d,
        }
    }
```

https://github.com/mxinden/neqo/blob/5d847eed43007e7a68cf66fa76c9728b8a6525e4/neqo-common/src/datagram.rs#L47-L62

``` rust
// TODO: replace recv_inner in favor of this one.
pub fn recv_inner_2<'a>(
    local_address: &SocketAddr,
    state: &UdpSocketState,
    socket: impl SocketRef,
    recv_buf: &'a mut Vec<u8>,
) -> Result<Datagram<&'a [u8]>, io::Error> {

    // ...

    Ok(Datagram::new_2_with_segment_size(
        meta.addr,
        *local_address,
        meta.ecn.map(|n| IpTos::from(n as u8)).unwrap_or_default(),
        meta.stride,
        recv_buf,
    ))
}
```

https://github.com/mxinden/neqo/blob/5d847eed43007e7a68cf66fa76c9728b8a6525e4/neqo-udp/src/lib.rs#L170-L176



To avoid any confusion, the `new_2_with_segment_size` is a work-in-progress name for now.

Does this make sense?

neqo-transport/src/connection/mod.rs
mxinden: In the PR description I stripped away the `TODO` comment. Otherwise it is the same:

```diff
     pub fn process_into<'a>(
         &mut self,
         input: Option<Datagram<&[u8]>>,
         now: Instant,
-        // TODO: Could this as well be an &mut [u8]?
         write_buffer: &'a mut Vec<u8>,
     ) -> Output<&'a [u8]> {
```

https://github.com/mxinden/neqo/blob/5d847eed43007e7a68cf66fa76c9728b8a6525e4/neqo-transport/src/connection/mod.rs#L1179-L1185

Note the rename to `process_into` is temporary. This is a large change. In order to incrementally compile I introduced some duplication.

mxinden:
> 1-conn/1-100mb-resp (aka. Download)/client: 💚 Performance has improved.
> 
>        time:   [98.680 ms 98.918 ms 99.154 ms]
>        thrpt:  [1008.5 MiB/s 1010.9 MiB/s 1013.4 MiB/s]
> change:
>        time:   [-12.936% -12.560% -12.190%] (p = 0.00 < 0.05)
>        thrpt:  [+13.882% +14.364% +14.858%]

This looks promising.

mxinden:

Due to its substantial size, I recommend beginning with the (updated) description in the pull request.

neqo-common/src/datagram.rs
mxinden: I significantly cleaned-up the pull request, thus hopefully avoiding the confusion above.

I am closing here. I am sorry for not making the best use of your time @martinthomson.

Please take another look whenever you can. There is no rush.

neqo-transport/src/connection/mod.rs
```diff
@@ -1119,18 +1129,36 @@ impl Connection {
...
+    /// TODO
     #[must_use = "Output of the process function must be handled"]
-    pub fn process(&mut self, dgram: Option<&Datagram>, now: Instant) -> Output {
-        if let Some(d) = dgram {
+    pub fn process<'a>(
+        &mut self,
+        input: Option<Datagram<&[u8]>>,
+        now: Instant,
+        write_buffer: &'a mut Vec<u8>,
+    ) -> Output<&'a [u8]> {
```
mxinden, line 1151: This is the core of this pull request. See pull request description.

fuzz/fuzz_targets/client_initial.rs
```diff
@@ -17,18 +17,23 @@ fuzz_target!(|data: &[u8]| {
     };
 
     let mut client = default_client();
-    let ci = client.process(None, now()).dgram().expect("a datagram");
+    let ci = client
+        .process_alloc(None, now())
+        .dgram()
+        .expect("a datagram");
```
mxinden, line 23: Note `process_alloc` is simply a convenience function for tests, allocating into an ad-hoc `write_buffer`.

fuzz/fuzz_targets/client_initial.rs
```diff
@@ -17,18 +17,23 @@ fuzz_target!(|data: &[u8]| {
...
     let Some((header, d_cid, s_cid, payload)) = decode_initial_header(&ci, Role::Client) else {
         return;
     };
     let (aead, hp) = initial_aead_and_hp(d_cid, Role::Client);
     let (_, pn) = remove_header_protection(&hp, header, payload);
 
-    let mut payload_enc = Encoder::with_capacity(MIN_INITIAL_PACKET_SIZE);
+    let mut write_buffer = Vec::with_capacity(MIN_INITIAL_PACKET_SIZE);
+    let mut payload_enc = Encoder::new(&mut write_buffer);
```
mxinden, line 31: `Encoder` no longer owns its `write_buffer`, thus allowing us to have `Encoder` encode into the long-lived `write_buffer` instead.

Downside, slightly more verbose test code.

larseggert:
> Long-lived send and receive buffers come with a bit of complexity. Do you think that complexity is worth the benefit? In other words, are you in favor of this change in general?

I think it is. In order to increase performance, we need to eliminate memory allocations and copies.

> If yes, given that this is a large pull request, do you want me to split this pull request up into smaller pull requests, or do you prefer one atomic pull request?

I don't mind the large PR. Is there a way so the tests can continue to use a fn name `process` and not `process_alloc`? Many lines in the PR are only due to that renaming. 

(Nit: I also find `write_buffer` lengthy. Maybe simply `out`?)

mxinden:
> > Long-lived send and receive buffers come with a bit of complexity. Do you think that complexity is worth the benefit? In other words, are you in favor of this change in general?
> 
> I think it is. In order to increase performance, we need to eliminate memory allocations and copies.
> 
> > If yes, given that this is a large pull request, do you want me to split this pull request up into smaller pull requests, or do you prefer one atomic pull request?
> 
> I don't mind the large PR.

Great. Then, unless someone else objects, this is ready for more detailed reviews.

> > If yes, given that this is a large pull request, do you want me to split this pull request up into smaller pull requests, or do you prefer one atomic pull request?
> 
> I don't mind the large PR. Is there a way so the tests can continue to use a fn name `process` and not `process_alloc`? Many lines in the PR are only due to that renaming.

Unfortunately I don't see a way around having two functions (e.g. `process` and `process_alloc`). 

Reasons:
- `process` needs a 4th argument, namely `write_buffer`.
- `process` returns a borrowed `Output<&'a [u8]>`, `process_alloc` returns an owned `Output`.

(Obviously we can name the two functions however we like.)

> (Nit: I also find `write_buffer` lengthy. Maybe simply `out`?)

Fine by me.

larseggert:
Can we leave `process` as is and name the four-arg version something else? `process_with_buffer` or something better?

mxinden:
@larseggert I did the renames suggested above.

neqo-common/src/codec.rs
```diff
@@ -470,15 +457,17 @@ mod tests {
 
     #[test]
     fn decode() {
-        let enc = Encoder::from_hex("012345");
+        let mut out = vec![];
+        let enc = Encoder::new(&mut out).from_hex("012345");
```
larseggert, line 461: Might be able to make the diff shorter by doing something like
```
let enc = Encoder::new(&mut vec![]).from_hex("012345");
```
here and elsewhere where you don't actually need to refer to `out` otherwise? (Don't feel strongly about this; this **is** clearer.)

neqo-common/src/datagram.rs
```diff
@@ -9,23 +9,20 @@ use std::{net::SocketAddr, ops::Deref};
...
 
 #[derive(Clone, PartialEq, Eq)]
-pub struct Datagram {
+pub struct Datagram<D = Vec<u8>> {
     src: SocketAddr,
     dst: SocketAddr,
+    /// The size of each segment within the [`Datagram`]. All segments, but the
+    /// last, have the same size. The last segment can be shorter than
+    /// [`Datagram::segment_size`].
+    segment_size: usize,
```
larseggert, line 18: Is `Datagram` still the best descriptive name for this, given that it's now a sequence of datagrams of the same size? It sounds odd to me to see a datagram described as something that has segments in it?

Maybe `DatagramVec` or something like this would be better?

(Also wonder if `segment_size` -> `len` would then make sense. It's shorter.)

neqo-http3/src/connection_client.rs
```diff
@@ -854,13 +854,38 @@ impl Http3Client {
...
+    ) -> Output<&'a [u8]> {
         qtrace!([self], "Process.");
-        if let Some(d) = dgram {
+        if let Some(d) = input {
             self.process_input(d, now);
+        } else {
+            // TODO: Optimization worth it? Given that process_input calls
+            // process_http3, moving the line below outside of conditional makes
+            // it run twice when input is some.
+            self.process_http3(now);
```
larseggert, line 870: Is there still anything left to be done about this `TODO`? Because if it works, why not just do it?

neqo-http3/src/frames/hframe.rs
```diff
@@ -142,7 +142,9 @@ impl HFrame {
                 element_id,
                 priority,
             } => {
-                let mut update_frame = Encoder::new();
+                let mut out = vec![];
+                // TODO: separate encoder needed?
+                let mut update_frame = Encoder::new(&mut out);
```
larseggert, line 147: Anything left to be done about this `TODO`?

neqo-http3/src/server.rs
```diff
@@ -113,18 +113,46 @@ impl Http3Server {
...
+        out: &'a mut Vec<u8>,
+    ) -> Output<&'a [u8]> {
         qtrace!([self], "Process.");
-        let out = self.server.process(dgram, now);
+        let mut output = self
+            .server
+            // TODO: NLL borrow issue. See https://github.com/rust-lang/rust/issues/54663
+            //
+            // Find alternative.
+            .process_into_buffer(dgram, now, unsafe { &mut *std::ptr::from_mut(out) });
```
larseggert, line 128: Anything left to do here?

neqo-transport/src/connection/mod.rs
```diff
@@ -1135,18 +1151,37 @@ impl Connection {
...
+        let mut out = vec![];
+        self.process_into_buffer(dgram.map(Into::into), now, &mut out)
+            .map_datagram(Into::into)
+    }
+
+    /// Process input and generate output.
+    ///
+    /// # Panics
+    ///
+    /// TODO
```
larseggert, line 1167: `TODO` what?

neqo-transport/src/connection/mod.rs
```diff
@@ -1520,119 +1558,123 @@ impl Connection {
...
-                            // This was a valid-appearing Initial packet: maybe probe with
-                            // a Handshake packet to keep the handshake moving.
-                            self.received_untracked |=
-                                self.role == Role::Client && cspace == CryptoSpace::Initial;
+                    Err(e) => {
+                        match e {
+                            Error::KeysPending(cspace) => {
+                                // This packet can't be decrypted because we don't have the keys
+                                // yet. Don't check this packet for
+                                // a stateless reset, just return.
```
larseggert, line 1648:
```suggestion
                                // yet. Don't check this packet for a stateless reset, just return.
```

neqo-transport/src/packet/mod.rs
```diff
@@ -128,13 +128,15 @@ pub struct PacketBuilder {
...
+impl<'a> PacketBuilder<'a> {
     /// The minimum useful frame size.  If space is less than this, we will claim to be full.
     pub const MINIMUM_FRAME_SIZE: usize = 2;
 
-    fn infer_limit(encoder: &Encoder) -> usize {
-        if encoder.capacity() > 64 {
-            encoder.capacity()
+    const fn infer_limit(limit: usize) -> usize {
+        // TODO: I don't know what the 64 is all about. Thus leaving the infer_limit function intact
+        // for now.
```
larseggert, line 137: Huh. @martinthomson might know.

neqo-transport/src/packet/mod.rs
```diff
@@ -1234,29 +1276,32 @@ mod tests {
...
-        assert_eq!(builder.abort(), encoder_copy);
+        // let builder = PacketBuilder::long(
+        //     encoder,
+        //     PacketType::Initial,
+        //     Version::default(),
+        //     Some(ConnectionId::from(SERVER_CID)),
+        //     Some(ConnectionId::from(SERVER_CID)),
+        // );
+        // assert_eq!(builder.remaining(), 0);
+        // assert_eq!(builder.abort(), encoder_copy);
```
larseggert, line 1304:
```suggestion
```

larseggert (commented):
LGTM overall, but it needs a review from someone with more Rust-fu like @martinthomson.

I think we should come up with and then use consistent terminology for what we call these new datagrams that have segments in them. `Datagram` to me sounds too much like it's a single packet (payload). `DatagramVec`? `PacketVec`? We call the individual payloads "segments", which as a TCP person I find confusing. We also talk about "segment length" and "stride length".

There are a bunch of `TODO`s (beyond those I marked) that probably need addressing.

neqo-http3/src/server.rs
mxinden: I have not been able to come up with a proper solution (i.e. no `unsafe`). Before merging here, I want to spend more time on it, unless of course, we go in a direction that no longer requires the trick in the first place.

mxinden:
> There are a bunch of `TODO`s (beyond those I marked) that probably need addressing.

Sorry for the noise. Before solving some of the trickier TODOs, I wanted to make sure the direction taken here, requiring the TODOs in the first place, is the direction we want to take.

Best example is the false-positive double-borrow in https://github.com/mozilla/neqo/pull/2093#discussion_r1768003763.

I plan to solve all TODOs before merging.

mxinden:
### Naming

> We call the individual payloads "segments", which as a TCP person I find confusing.

While it can be easily confused with the TCP usage of the same term, I do believe "segments" is the right word to refer to the parts of a GSO/GRO UDP datagram. See for example UDP [GSO](https://github.com/torvalds/linux/commit/bec1f6f697362c5bc635dacd7ac8499d0a10a4e7) and [GRO](https://github.com/torvalds/linux/commit/e20cf8d3f1f763ad28a9cb3b41305b8a8a42653e) Linux Kernel patches and the [documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-segmentation-offload-uso-) of the Windows equivalent.

@larseggert are you aware of any other word used to refer to the parts of a GSO/GRO UDP datagram?

> We also talk about "segment length" and "stride length".

"stride" is only used when interacting with the `quinn-udp` API. `quinn-udp` uses "stride" and "segment" interchangeably. I don't know the origin of using "stride". I am not aware of any other system using the term for this meaning. Happy to start a discussion upstream if you think it is helpful.

 As said above, I believe "segment" is the conventional term. 

### Datastructure for a collection of UDP datagram segments

> I think we should come up with and then use consistent terminology for what we call these new datagrams that have segments in them. [...] `DatagramVec`? `PacketVec`?

I prefer the current approach, namely to extend the notion of a `Datagram` to contain one or more _segments_. Each _segment_ has the same 4-tuple (i.e. `(src_ip, src_port, dst_ip, dst_port)`), TOS, ... . Thus logic that only handles the metadata of a `Datagram` doesn't care whether it handles the metadata of one `Datagram`, or the metadata of a collection of `Datagram`s each with the same metadata. (Only exception being accounting logic like `neqo_transport::Stats`.)

I don't feel strongly about it. In case it doesn't add too much noise, I am fine introducing a new datastructure, representing a collection of UDP datagram segments. `SegmentedDatagram` maybe?

### Future with `recvmmsg`

The matter gets a bit more complicated in the future. Currently we only pass a single `iovec` to `recvmmsg`, thus only leveraging `GRO`, not actually `recvmmsg`. Once we pass multiple `iovec`s to `recvmmsg`, the UDP datagrams across `iovec`s might differ in their metadata (e.g. 4-tuple) (`recvmmsg` style). The the UDP datagrams within an `iovec` will have equal metadata (`GRO` style).

With the current approach, each `iovec` would result in a single `Datagram` with one or more segments, and the whole `recv` system call with multiple `iovec`s would result in a collection of `Datagram`s each with one or more segments. Thus the `Socket::recv` signature would change as following:

``` diff
     pub fn recv<'a>(
         &self,
         local_address: &SocketAddr,
         recv_buf: &'a mut Vec<u8>,
-    ) -> Result<Option<Datagram<&'a [u8]>>, io::Error> {
+    ) -> Result<impl Iterator<Item = Datagram<&'a [u8]>>, io::Error> {
```

larseggert:
Thanks for the detailed explanation. I'm OK with this.

larseggert:
@mxinden I guess this needs more work before it can be merged? Should we make this a draft until then?

neqo-http3/src/connection_client.rs
mxinden: Addressed in https://github.com/mozilla/neqo/pull/2093/commits/15eb2f8abbab2e4d6536973c64ffedc1f8c1b612.

neqo-common/src/datagram.rs
mxinden: Addressed in https://github.com/mozilla/neqo/pull/2093#issuecomment-2372100294.

neqo-common/src/codec.rs
mxinden: Unfortunately this is not possible. The anonymous temporary `vec![]` would life within the scope of the `Encoder::new` invocation only, i.e. dropped after the call, but it would be borrowed for the entire lifetime of `enc`. In other words `vec![]` would be dropped while still being borrowed by `enc`.

```
error[E0716]: temporary value dropped while borrowed
   --> neqo-common/src/codec.rs:459:37
    |
459 |         let enc = Encoder::new(&mut vec![]).from_hex("012345");
    |                                     ^^^^^^                    - temporary value is freed at the end of this statement
    |                                     |
    |                                     creates a temporary value which is freed while still in use
460 |         let mut dec = enc.as_decoder();
    |                       --- borrow later used here
    |
    = note: this error originates in the macro `vec` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider using a `let` binding to create a longer lived value
    |
459 ~         let binding = vec![];
460 ~         let enc = Encoder::new(&mut binding).from_hex("012345");
    |

For more information about this error, try `rustc --explain E0716`.
```

neqo-http3/src/frames/hframe.rs
mxinden: Cleaned up in https://github.com/mozilla/neqo/pull/2093/commits/ffc3708dbd9426946caa90bd0183c17d38515a79.

neqo-transport/src/packet/mod.rs
mxinden: For what it is worth, the value `64` was introduced in https://github.com/mozilla/neqo/pull/998.

Also note that no unit test depends on the value `64`, in other words, removing it does not cause any test to fail.

.github/workflows/polonius.diff
mxinden, line 1: See `.github/workflows/polonius.yml` right below.

.github/workflows/polonius.yml
```diff
@@ -0,0 +1,61 @@
+# Rustc by default uses the NLL borrow-checker. There are multiple cases where
+# NLL is too restrictive. An upcoming improvement to Rust's borrow-checker,
+# adressing these shortcomings, is Polonius. It ships with Nightly behind a
+# flag.
+#
+# This workflow first removes the workarounds necessary to please NLL and then
+# runs with Polonius to ensure each workaround only fixes the false-positive of
+# NLL and doesn't mask an actually undefined behavior.
```
mxinden, line 8: Somewhat unconventional. Still, I think statically proving `unsafe` code to be safe is worth it. As always, open for alternative suggestions.

neqo-transport/src/packet/mod.rs
```diff
@@ -118,26 +118,32 @@ struct PacketBuilderOffsets {
...
-    encoder: Encoder,
+pub struct PacketBuilder<'a> {
+    encoder: Encoder<'a>,
     pn: PacketNumber,
     header: Range<usize>,
     offsets: PacketBuilderOffsets,
+    /// The size limit of [`Self::encoder`], i.e. the maximum number of bytes to
+    /// be encoded into [`Self::encoder`]. Note that [`Self::encoder`] might
+    /// contain one or more packets already.
     limit: usize,
```
mxinden, line 129: Previously we would allocate a new `Vec` for each UDP datagram, with the capacity of the `Vec` set to the limit.

Now we have one long-lived receive buffer, with a fixed capacity. In order to enforce the limit `PacketBuilder` now owns the limit in its `limit: usize`, i.e. is explicit about it, instead of implicitly depending on the capacity of the `Vec` of `PacketBuilder::encoder`.

.github/workflows/polonius.diff
mxinden, line 1: See `.github/workflows/polonius.yml` right below.

neqo-transport/src/packet/mod.rs
larseggert: Ping @martinthomson 

We should also maybe make `64` and `2048` appropriately-named `const`s, if we determine what they do & that we want to keep them.

neqo-transport/src/packet/mod.rs
```diff
@@ -997,12 +1012,14 @@ mod tests {
         let burn = prot.encrypt(0, &[], &[]).expect("burn OK");
         assert_eq!(burn.len(), prot.expansion());
```
larseggert, line 1013: Unrelated to this PR, but these should probably return errors and not panic.

neqo-transport/src/server.rs
```diff
@@ -192,22 +192,25 @@ impl Server {
...
+    fn handle_initial<'a>(
         &mut self,
         initial: InitialDetails,
-        dgram: &Datagram,
+        dgram: Datagram<&[u8]>,
         now: Instant,
-    ) -> Output {
+        out: &'a mut Vec<u8>,
+    ) -> Output<&'a [u8]> {
+        assert!(out.is_empty());
```
larseggert, line 202: `debug_assert` or return error?

neqo-transport/src/server.rs
```diff
@@ -339,7 +345,13 @@ impl Server {
...
     }
 
-    fn process_input(&mut self, dgram: &Datagram, now: Instant) -> Output {
+    fn process_input<'a>(
+        &mut self,
+        dgram: Datagram<&[u8]>,
+        now: Instant,
+        out: &'a mut Vec<u8>,
+    ) -> Output<&'a [u8]> {
+        assert!(out.is_empty());
```
larseggert, line 354: `debug_assert` or return error?

neqo-transport/src/server.rs
```diff
@@ -426,11 +440,17 @@ impl Server {
 
     /// Iterate through the pending connections looking for any that might want
     /// to send a datagram.  Stop at the first one that does.
-    fn process_next_output(&mut self, now: Instant) -> Output {
+    fn process_next_output<'a>(&mut self, now: Instant, out: &'a mut Vec<u8>) -> Output<&'a [u8]> {
+        assert!(out.is_empty());
```
larseggert, line 444: `debug_assert` or return error?

neqo-transport/src/packet/mod.rs
martinthomson: The value exists to ensure that we respect the capacity when it is already set.  I forget the details.

neqo-transport/src/connection/mod.rs
```diff
@@ -1113,12 +1119,22 @@ impl Connection {
...
     /// Get output packets, as a result of receiving packets, or actions taken
     /// by the application.
     /// Returns datagrams to send, and how long to wait before calling again
     /// even if no incoming packets.
     #[must_use = "Output of the process_output function must be handled"]
-    pub fn process_output(&mut self, now: Instant) -> Output {
+    fn process_output_into_buffer<'a>(
+        &mut self,
+        now: Instant,
+        out: &'a mut Vec<u8>,
```
martinthomson, line 1136: Why `&mut Vec<u8>` and not `&mut [u8]`?  I don't think that -- as long as we are accepting someone else's memory -- we should be reallocating it.

neqo-transport/src/connection/mod.rs
```diff
@@ -1146,18 +1162,37 @@ impl Connection {
         }
     }
 
-    /// Process input and generate output.
+    /// Same as [`Connection::process_into_buffer`] but allocating output into
+    /// new [`Vec`].
     #[must_use = "Output of the process function must be handled"]
     pub fn process(&mut self, dgram: Option<&Datagram>, now: Instant) -> Output {
```
martinthomson, line 1168: Why not `Option<impl Into<Datagram<&'a [u8]>>>`?

neqo-transport/src/connection/mod.rs
```diff
@@ -1146,18 +1162,37 @@ impl Connection {
...
+
+    /// Process input and generate output.
+    ///
+    /// # Panics
+    ///
+    /// Panics when `out` is not empty.
+    #[must_use = "Output of the process function must be handled"]
+    pub fn process_into_buffer<'a>(
+        &mut self,
+        input: Option<Datagram<&[u8]>>,
```
martinthomson, line 1182: As above.

neqo-transport/src/connection/mod.rs
```diff
@@ -1267,24 +1302,27 @@ impl Connection {
...
+        remaining: usize,
+        now: Instant,
+    ) {
+        let d = Datagram::new(
+            d.source(),
+            d.destination(),
+            d.tos(),
+            d[d.len() - remaining..].to_vec(),
+            Some(d.segment_size()),
+        );
```
martinthomson, line 1325: Maybe you could implement a `to_owned()` for `Datagram` rather than do it this way.

neqo-transport/src/connection/mod.rs
```diff
@@ -1531,119 +1569,121 @@ impl Connection {
         self.capture_error(Some(path), now, 0, res).ok();
     }
 
-    fn input_path(&mut self, path: &PathRef, d: &Datagram, now: Instant) -> Res<()> {
-        let mut slc = &d[..];
-        let mut dcid = None;
+    fn input_path(&mut self, path: &PathRef, d: Datagram<&[u8]>, now: Instant) -> Res<()> {
+        for mut slc in d.iter_segments() {
```
martinthomson, line 1573: I have to say, this is too much for me.  The idea that a datagram is actually multiple datagrams is not something I'm comfortable with.

Making this function more complicated is less than ideal as well.

.github/workflows/polonius.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - id: nss-version
+        run: echo "minimum=$(cat neqo-crypto/min_version.txt)" >> "$GITHUB_OUTPUT"
+
+      - uses: ./.github/actions/nss
+        with:
+          minimum-version: ${{ steps.nss-version.outputs.minimum }}
+
+      - name: Apply patch, removing `unsafe` workarounds.
+        run: git apply .github/workflows/polonius.diff
```
martinthomson, line 59: This isn't really the best place to keep this diff.  Put it in the tree.

neqo-common/src/datagram.rs
martinthomson: I'm not happy with that choice.  This is not a single Datagram any more.  It is a collection of datagrams with shared addressing (and TOS) information.  `DatagramVec` isn't entirely right (as that implies ownership and other `Vec` capabilities), but `DatagramList` might be OK.  

It's OK to use segments for the individual datagrams, but I would instead use `slice_size` for that, because it is the size of each slice.  The implication here is that the data payload is sliced up to get individual datagrams.

I also find this API to be pretty suboptimal for other reasons.  Why, when you receive multiple datagrams, would they all be the same size?

neqo-common/src/datagram.rs
```diff
@@ -9,23 +9,20 @@ use std::{net::SocketAddr, ops::Deref};
...
-impl Datagram {
-    pub fn new<V: Into<Vec<u8>>>(src: SocketAddr, dst: SocketAddr, tos: IpTos, d: V) -> Self {
-        Self {
-            src,
-            dst,
-            tos,
-            d: d.into(),
-        }
-    }
+impl Copy for Datagram<&[u8]> {}
```
martinthomson, line 23: Really?  That's a pretty large thing to throw around like that.  I might prefer not to have it implement Copy.  I understand why this might be convenient, but that convenience hides actual work.  A Datagram will be at least 59 bytes in size, plus adjustments for alignment.

neqo-common/src/datagram.rs
```diff
@@ -44,17 +41,85 @@ impl Datagram {
...
+        dst: SocketAddr,
+        tos: IpTos,
+        d: D,
+        segment_size: Option<usize>,
+    ) -> Self {
+        Self {
+            src,
+            dst,
+            tos,
+            segment_size: segment_size.unwrap_or_else(|| d.as_ref().len()),
```
martinthomson, line 63:
```suggestion
            segment_size: segment_size.unwrap_or(usize::MAX),
```

You always iterate and truncate the last chunk, so this is cheaper.

neqo-common/src/datagram.rs
```diff
@@ -44,17 +41,85 @@ impl Datagram {
...
+            src: *src,
+            dst: *dst,
+            tos: *tos,
+            segment_size: *segment_size,
+            d,
+        }
+    }
+}
+
+impl From<Datagram<&[u8]>> for Datagram {
```
martinthomson, line 103: Should this be a ToOwned implementation instead?

neqo-common/src/datagram.rs
```diff
@@ -44,17 +41,85 @@ impl Datagram {
...
+impl<D: Deref<Target = [u8]>> Deref for Datagram<D> {
+    type Target = [u8];
     #[must_use]
     fn deref(&self) -> &Self::Target {
         &self.d
     }
 }
 
-impl std::fmt::Debug for Datagram {
+impl<'a> From<&'a Datagram> for Datagram<&'a [u8]> {
```
martinthomson, line 84: Should this be a Borrow implementation instead?

martinthomson (commented):
It seems to me like you could do this in pieces, starting with the encoder/decoder changes.

mxinden:
Thank you for the reviews @larseggert and @martinthomson.

> It seems to me like you could do this in pieces, starting with the encoder/decoder changes.

Sounds good. I will address the comments here and then break this large pull request into smaller ones

Back to draft for now.

neqo-transport/src/tparams.rs
```diff
@@ -688,7 +688,8 @@ impl ExtensionHandler for TransportParametersHandler {
         qdebug!("Writing transport parameters, msg={:?}", msg);
 
         // TODO(ekr@rtfm.com): Modify to avoid a copy.
```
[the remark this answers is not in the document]
mxinden: Unfortunately not. `self.local` is still first encoded into an intermediary buffer (i.e. `out`, previously `enc`) and only then copied into `d`.

I suggest addressing this `TODO` in a separate pull request.

neqo-transport/src/packet/mod.rs
mxinden: Note that this particular line is part of a unit test. What would be the benefit of returning an error over right away panicking @larseggert?

(Agreed on your comments above on non-unit-test lines, using `debug_assert` and returning an error in release mode.)

neqo-common/src/datagram.rs
mxinden: I don't think implementing `Borrow` is possible.

Here is the trait definition:

``` rust
pub trait Borrow<Borrowed: ?Sized> {
    fn borrow(&self) -> &Borrowed;
}
```

https://doc.rust-lang.org/alloc/borrow/trait.Borrow.html

The problematic part is the return value, i.e. `&Borrowed`. It would need to return a `&Datagram<&'a [u8]>`. Given that the `Datagram<&'a [u8]>` would need to be instantiated within the `borrow` implementation, the reference would point to a temporary value.

I have created the following simplified playground as a showcase.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=177f0e0f97ebbce7376092c924c72c55

neqo-common/src/datagram.rs
mxinden: I don't think implementing `ToOwned` is possible.

``` rust
pub trait ToOwned {
    type Owned: Borrow<Self>;

    fn to_owned(&self) -> Self::Owned;
}

```

https://doc.rust-lang.org/alloc/borrow/trait.ToOwned.html

`ToOwned` requires the associated type `Owned` to implement `Borrow<Self>`, in other words for `Datagram<Vec<u8>>` to implement `Borrow<Datagram<&'a [u8]>`. I don't think that is possible, see https://github.com/mozilla/neqo/pull/2093#discussion_r1788539953.

mxinden:
**Receive path** optimization has been merged through https://github.com/mozilla/neqo/pull/2184.

While this pull request can serve as guidance for a future **send path** patch, I assume it is better to start from scratch.

Thus closing here.

PR #2097, opened 2024-09-10
build: Optimize dev dependencies

larseggert:
Noticed this in https://searchfox.org/mozilla-central/source/Cargo.toml

It speeds up my local build by ~20%:

```
Benchmark 1: cargo build
  Time (mean ± σ):     22.947 s ±  0.251 s    [User: 165.047 s, System: 9.151 s]
  Range (min … max):   22.350 s … 23.231 s    10 runs

Benchmark 1: cargo build
  Time (mean ± σ):     16.868 s ±  0.091 s    [User: 62.789 s, System: 7.239 s]
  Range (min … max):   16.721 s … 16.982 s    10 runs
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,11 @@ nursery = { level = "warn", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
 multiple_crate_versions = "allow"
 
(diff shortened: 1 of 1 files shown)


larseggert:
Hm. This change is **not** why some builds are faster than others...

PR #2100, opened 2024-09-12
fix: Allow turning off greasing and connection migration

larseggert:
There was some plumbing missing. CC @mxinden


martinthomson (approved):
This looks fine.

One thing I noticed is that we have a public method called "migrate" on the connection.  This method can be called when the server has set `disable_active_migration`.  That's not ideal.  That's worth another change, I think.

mxinden:
Ah, sorry, merged `main` too early. https://github.com/mozilla/neqo/pull/2105 is still pending.

PR #2101, opened 2024-09-12
test(transport): check local and remote GREASE_QUIC_BIT TP

mxinden:
Follow-up to https://github.com/mozilla/neqo/pull/2100.

Test succeeds when run on top of https://github.com/mozilla/neqo/pull/2100.

@larseggert feel free to cherry-pick in case you find this useful.


larseggert:
Thanks, we can land this after #2100.

PR #2102, opened 2024-09-12
Expose an SSLNamedGroup for mlkem768x25519

jschanck:
NSS 3.105 will support ML-KEM-768 (the FIPS-203 version of Kyber768). An alpha release of 3.105 will land in M-C in [D222012](https://phabricator.services.mozilla.com/D222012). I'm switching PSM from xyber768d00 to mlkem768x25519 in [D222019](https://phabricator.services.mozilla.com/D222019). This patch exposes an SSLNamedGroup variant for `ssl_grp_kem_mlkem768x25519` so that I can do the same for [neqo_glue](https://searchfox.org/mozilla-central/rev/4a8bd8efdfaa43dd14a16d3cb15bf86796fd1def/netwerk/socket/neqo_glue/src/lib.rs#263-280).

This patch also changes the behavior of `init_once()` so that xyber768d00 is disabled by policy (it's disabled by default in NSS and was being enabled in `init_once()`). This will prevent servers from doing an HRR for xyber768d00 when they are presented with, say, {mlkem768x25519, x25519}. I hear that most servers that currently support xyber768d00 will support mlkem768x25519 by mid-October, slightly before the Firefox 132 release. So disabling xyber768d00 should not have a significant impact on the fraction of Firefox-initiated handshakes that use a post-quantum key exchange.

If you're not comfortable merging a change that depends on pre-release NSS code, the plan is to cut NSS_3_105_BETA on Sept 19 and NSS_3_105_RTM on Sept 26. I'll have plenty of time to make the neqo_glue changes even if you wait until the 26th.


larseggert (approved):
LGTM modulo the formatting nit

martinthomson (approved):
There's a difference here between what it takes to write the code and what it takes to land it.  This patch is great and we should be taking it.  The challenge is in managing the release arrangements.

It isn't great having this repository reliant on prerelease NSS.  But we're going to end up in that situation at times.

Is this just a case where we have to quickly slot all the pieces in place in order (NSS, neqo, gecko)?  Or can we make that a little less difficult with something like a feature flag?

jschanck:
> Is this just a case where we have to quickly slot all the pieces in place in order (NSS, neqo, gecko)?

Not really. This isn't blocking anything except for enabling mlkem768x25519 in h3 ([D222041](https://phabricator.services.mozilla.com/D222041)). The NSS uplift ([D222012](https://phabricator.services.mozilla.com/D222012)) and the PSM integration ([D222019](https://phabricator.services.mozilla.com/D222019)) are both in autoland with clean try runs, so they'll be in Nightly shortly.

The NSS release happens on the same day as the Firefox soft freeze, so waiting for the release just means that Firefox 132 will use mlkem768x25519 for TLS but not for h3 (which is fine).

> Or can we make that a little less difficult with something like a feature flag?

Thought about this for a bit---probably more hassle than it's worth.

martinthomson:
I thought that I'd ask, but I reached the same conclusion.  Thanks!

larseggert:
Landing this will need to wait until we can back out https://github.com/mozilla/neqo/pull/2106, i.e., unpin NSS to a version that has the MLKEM stuff.

PR #2110, opened 2024-09-13
ci: unpin nss

jschanck:
test to see if https://hg.mozilla.org/projects/nss/rev/18b287c5128e489b483cc156cdc7a51e41185c65 fixes the issue.


larseggert:
Thanks! I prefer to go with #2108, since this PR reverts some changes I'd like to keep (like using the GitHub action to check out the source.)

PR #2117, opened 2024-09-16
fix: Do CC reaction before `largest_acked`

larseggert:
Packets are only declared as lost relative to `largest_acked`. If we hit a PTO while we don't have a largest_acked yet, also do a congestion control reaction (because otherwise none would happen).

Broken out of #1998


larseggert:

neqo-transport/src/recovery/mod.rs
```diff
@@ -851,6 +851,17 @@ impl LossRecovery {
         // pto_time to increase which might cause PTO for later packet number spaces to not fire.
         if let Some(pn_space) = pto_space {
             qtrace!([self], "PTO {}, probing {:?}", pn_space, allow_probes);
+            // Packets are only declared as lost relative to `largest_acked`. If we hit a PTO while
+            // we don't have a largest_acked yet, also do a congestion control reaction (because
+            // otherwise none would happen).
```
martinthomson, line 856: Are we sure that we want to slow down?  If we haven't received an ACK, we should be at a low send rate.  Do we really want to slow down further?

One thing that concerns me here is that a fast PTO (if we ever enable that) will hit this condition more often and that might not be good for performance.  The same goes for long RTT connections, which might be OK while we are still retransmitting within the RTT, but once we get an RTT estimate that is long, we'll slow our send rate by a huge amount for the false PTOs we've hit.

martinthomson (commented):
The structure looks fine, but this seems overly conservative to me.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -337,6 +337,41 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         congestion || persistent_congestion
     }
 
+    /// Handle a congestion event.
```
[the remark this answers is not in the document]
larseggert: Yes.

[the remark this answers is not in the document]
larseggert: Yes, but from `ClassicCongestionControl` to `CongestionControl`.

neqo-transport/src/recovery/mod.rs
larseggert: I think the argument is that the combination of initial cwnd and initial RTT (= PTO) were made with some sort of safety criteria in mind. A PTO is an indication that either the RTT is much longer than the default assumption, or there is quite a bit of loss. In either of these two cases, we probably want to slow down?

martinthomson: I guess my main concern is that if we want to be more aggressive about PTO, such as sending another Initial we create a situation where slowing down is not the result of a misunderstanding of the RTT, but a deliberate choice to send more to start with.

mxinden:
This patch is in conflict with RFC 9002, right?

> A PTO timer expiration event does not indicate packet loss and MUST NOT cause prior unacknowledged packets to be marked as lost.

https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2

larseggert:
IIRC (can't verify right now) the issue I was trying to fix is that if a client is trying to start a connection to an unresponsive server, we would exponentially back off the retransmission, but not halve the cwnd.

The question is if we want to halve the cwnd. Proper CC would want is to (those losses should be taken as signs of congestion),  but it will of course lower performance, esp. if there are PMTUD or ECN issues.

The reason we are marking the CI as lost is that we'd otherwise never retransmit (IIRC).

mxinden:
> The reason we are marking the CI as lost is that we'd otherwise never retransmit (IIRC).

https://github.com/mozilla/neqo/pull/2129 should fix this, right?

larseggert:
@mxinden is there still anything left here to do now that #2492 is in?

mxinden:
@larseggert I believe this can be closed.

Note that I did not land the change of this pull request, i.e. `neqo-transport` still does no congestion control reaction before `largest_acked`. Similar to Martin's concerns above (https://github.com/mozilla/neqo/pull/2117#discussion_r1784493779), I don't see the need for a congestion control reaction. If I read RFC 9002 correctly, our current approach is inline.

PR #2118, opened 2024-09-16
fix: Drop of Initial space only after sending 1st Handshake

larseggert:
This is what RFC9000 says, and it helps with servers that expect ACKs in the Initial packet number space to happen. It also eliminates a difference in behavior we had when a client received an Initial packet containing an ACK and a CRYPTO frame vs. a CRYPTO frame and an ACK, where that CRYPTO frame caused the Initial packet number space to be dropped.

Broken out of #1998


mxinden (approved):
This makes sense to me.

Can you link to the concrete section of RFC 9000 you are referring to?

I can only find the following in RFC 9001 4.9.1:

> The successful use of Handshake packets indicates that no more Initial packets need to be exchanged, as these keys can only be produced after receiving all CRYPTO frames from Initial packets. Thus, a client MUST discard Initial keys when it first sends a Handshake packet and a server MUST discard Initial keys when it first successfully processes a Handshake packet. Endpoints MUST NOT send Initial packets after this point.

https://www.rfc-editor.org/rfc/rfc9001#section-4.9.1

larseggert:
That is the section in question.

larseggert:

mxinden:
@larseggert what are the next steps here?

larseggert:
I want to double-check that everything is working as intended here.

PR #2119, opened 2024-09-16
fix: RTX more data during handshake

larseggert:
We don't track which packets are coalesced with others, so when we detect a loss in one packet number space, we cannot resend any coalesced packets in other packet number space. This PR tries to approximate this behavior by scheduling un-ACKed Handshake and 0-RTT packets for RTX when packets are lost in the Initial space.

Broken out of #1998


mxinden (approved):
This makes sense to me.

> We don't track which packets are coalesced with others

Would there be any other benefits to tracking this?

larseggert:
> Would there be any other benefits to tracking this?

Well, it would make it possible to RTX the exact information we deem lost, rather than guessing. I don't know if it's worth it.

PR #2120, opened 2024-09-16
perf(transport): have add_datagram take Vec<u8>

mxinden:
`add_datagram` takes a Quic datagram and adds it to the queue of datagrams to be sent out. Previously it would take a reference (i.e. `&[u8]`) and would allocate it into a new `Vec<u8>` before enqueuing. At the call-site the original allocation (referenced by the `&[u8]`) would go out-of-scope and thus be de-allocated. This is a wasted allocation for each Quic datagram to be send.

This commit has the call-site pass the owned `Vec<u8>` down right away.

### DIFF
--- a/neqo-http3/src/features/extended_connect/webtransport_session.rs
+++ b/neqo-http3/src/features/extended_connect/webtransport_session.rs
@@ -415,7 +415,7 @@ impl WebTransportSession {
             let mut dgram_data = Encoder::default();
             dgram_data.encode_varint(self.session_id.as_u64() / 4);
             dgram_data.encode(buf);
-            conn.send_datagram(dgram_data.as_ref(), id)?;
+            conn.send_datagram(dgram_data.into(), id)?;
         } else {
             debug_assert!(false);
             return Err(Error::Unavailable);
--- a/neqo-transport/src/connection/tests/datagram.rs
+++ b/neqo-transport/src/connection/tests/datagram.rs
@@ -58,12 +58,12 @@ fn datagram_disabled_both() {
     assert_eq!(client.max_datagram_size(), Err(Error::NotAvailable));
     assert_eq!(server.max_datagram_size(), Err(Error::NotAvailable));
     assert_eq!(
-        client.send_datagram(DATA_SMALLER_THAN_MTU, None),
+        client.send_datagram(DATA_SMALLER_THAN_MTU.to_vec(), None),
         Err(Error::TooMuchData)
     );
     assert_eq!(server.stats().frame_tx.datagram, 0);
     assert_eq!(
-        server.send_datagram(DATA_SMALLER_THAN_MTU, None),
+        server.send_datagram(DATA_SMALLER_THAN_MTU.to_vec(), None),
         Err(Error::TooMuchData)
     );
     assert_eq!(server.stats().frame_tx.datagram, 0);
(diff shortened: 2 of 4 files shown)


neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -415,7 +415,7 @@ impl WebTransportSession {
             let mut dgram_data = Encoder::default();
             dgram_data.encode_varint(self.session_id.as_u64() / 4);
             dgram_data.encode(buf);
-            conn.send_datagram(dgram_data.as_ref(), id)?;
+            conn.send_datagram(dgram_data.into(), id)?;
```
mxinden, line 418: `dgram_data` is an `Encoder` wrapping a `Vec<u8>`. Instead of passing a reference (i.e. `dgram_data.as_ref()`) pass the inner `Vec<u8>` (i.e. `dgram_data.into()`) directly

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -415,7 +415,7 @@ impl WebTransportSession {
             let mut dgram_data = Encoder::default();
             dgram_data.encode_varint(self.session_id.as_u64() / 4);
             dgram_data.encode(buf);
```
mxinden, line 417: A second re-allocation of the entire Quic Datagram is happening here, just to prefix it with a varint.

I would assume there is a better way to do that, though I suggest not in this pull request.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -58,12 +58,12 @@ fn datagram_disabled_both() {
     assert_eq!(client.max_datagram_size(), Err(Error::NotAvailable));
     assert_eq!(server.max_datagram_size(), Err(Error::NotAvailable));
     assert_eq!(
-        client.send_datagram(DATA_SMALLER_THAN_MTU, None),
+        client.send_datagram(DATA_SMALLER_THAN_MTU.to_vec(), None),
```
larseggert, line 61: if this is possible with Rust, can we add this `to_vec` call to the definition of the `const`, rather than to each use? (Ditto for the other `to_vec` calls.)

mxinden: Long story short, I don't see an easy way around the `.to_vec()` calls.

Long story:

`Vec` is heap allocated, and thus can not be instantiated as a `const`, in other words, it can only be instantiated at runtime.

We could use a `lazy_static`, but then we would need to call `.clone()` everywhere, which I think adds the same amount of noise as calling `.to_vec()` everywhere.

We could change `send_datagram`:

``` diff
-    pub fn send_datagram(&mut self, buf: Vec<u8>, id: impl Into<DatagramTracking>) -> Res<()> {
+    pub fn send_datagram(&mut self, buf: impl Into<Vec<u8>>, id: impl Into<DatagramTracking>) -> Res<()> {
```

But this makes the allocation within `send_datagram` implicit to the call-site, thus once again hiding the fact that we would allocate on each datagram when called with a `&[u8]`.

larseggert: Thanks for the explanation.

larseggert (approved):

PR #2127, opened 2024-09-18
fix: Add `confirmed` parameter to PTO calculation

larseggert:
Rather than having the caller determine for which space a PTO should be calculated for.

Also arm PTOs in the application packet number space before the handshake is complete, so 0-RTT gets RTX'ed.

Broken out of #1998

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -629,13 +629,17 @@ impl Connection {
             .unwrap()
     }
 
+    fn confirmed(&self) -> bool {
+        self.state == State::Confirmed
+    }
+
     /// Get the simplest PTO calculation for all those cases where we need
     /// a value of this approximate order.  Don't use this for loss recovery,
     /// only use it where a more precise value is not important.
     fn pto(&self) -> Duration {
         self.paths.primary().map_or_else(
-            || RttEstimate::default().pto(PacketNumberSpace::ApplicationData),
-            |p| p.borrow().rtt().pto(PacketNumberSpace::ApplicationData),
+            || RttEstimate::default().pto(self.confirmed()),
+            |p| p.borrow().rtt().pto(self.confirmed()),
         )
     }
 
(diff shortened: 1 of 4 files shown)


neqo-transport/src/recovery/mod.rs
```diff
@@ -961,8 +964,8 @@ mod tests {
 
     const ON_SENT_SIZE: usize = 100;
     /// An initial RTT for using with `setup_lr`.
-    const TEST_RTT: Duration = ms(80);
-    const TEST_RTTVAR: Duration = ms(40);
+    const TEST_RTT: Duration = ms(7000);
+    const TEST_RTTVAR: Duration = ms(3500);
```
mxinden, line 968: Why these particular values? And why do they need to be that high?

mxinden (approved):
Changes make sense to me. Only one question related to tests.

neqo-transport/src/recovery/mod.rs
larseggert: They came out of a discussion with @martinthomson at the Vancouver IETF

larseggert:

neqo-transport/src/connection/mod.rs
```diff
@@ -629,13 +629,17 @@ impl Connection {
             .unwrap()
     }
 
+    fn confirmed(&self) -> bool {
```
martinthomson, line 632: This can't be `const` as well?

martinthomson (approved):
I like this one.  It's a simplification for the most part.

(There's some duplication of state information between Connection and the loss recovery stuff, but that seems necessary.  The meaning is consistent.)

neqo-transport/src/connection/mod.rs
larseggert: No, because the derived equality function isn't `const`. There are some Nightly-only features one can set, but they aren't stable.

mxinden:
Merge failure due to https://github.com/mozilla/neqo/issues/2141.

PR #2128, opened 2024-09-18
test: More Handshake tests

larseggert:
These should pass once the PR series replacing #1998 has landed.

Broken out of #1998


larseggert:
OBE

PR #2129, opened 2024-09-19
fix: Mark all packets TX'ed before PTO as lost

larseggert:
We'd previously only mark 1 one or two packets as lost when a PTO fired. That meant that we potentially didn't RTX all data that we could have (i.e., that was in lost packets that we didn't mark lost).

This also changes the probing code to suppress redundant keep-alives, i.e., PINGs that we sent for other reasons, which could double as keep-alives but did not.

Broken out of #1998


larseggert:

mxinden (approved):
This makes sense to me. Thanks for extracting it into a smaller pull request.

I am in favor of waiting for Martin's review.

neqo-transport/src/recovery/mod.rs
```diff
@@ -836,11 +833,7 @@ impl LossRecovery {
                 if t <= now {
                     qdebug!([self], "PTO timer fired for {}", pn_space);
                     let space = self.spaces.get_mut(*pn_space).unwrap();
-                    lost.extend(
-                        space
-                            .pto_packets(PtoState::pto_packet_count(*pn_space))
-                            .cloned(),
-                    );
+                    lost.extend(space.pto_packets().cloned());
```
martinthomson, line 836: Do we still need pto_packet_count if this is the decision?

The other question I have is whether this is necessary.  We're cloning all of the information so that we can process the loss, which means more work on a PTO.  Maybe PTO is rare enough that this doesn't matter, but one of the reasons for the limit on number was to avoid the extra work.

martinthomson (commented):
Do we not have tests for this?  Should we?

larseggert:
> Do we not have tests for this? Should we?

There are tests in #2128, but this PR alone doesn't make them succeed yet.

neqo-transport/src/recovery/mod.rs
larseggert: > Do we still need pto_packet_count if this is the decision?

We do still need it to limit the number of packets we send on PTO.

> The other question I have is whether this is necessary. We're cloning all of the information so that we can process the loss, which means more work on a PTO. Maybe PTO is rare enough that this doesn't matter, but one of the reasons for the limit on number was to avoid the extra work.

I've been wondering if it would be sufficient to mark _n_ packets per space as lost, instead of all.

mxinden:
> We'd previously only mark 1 one or two packets as lost when a PTO fired. That meant that we potentially didn't RTX all data that we could have (i.e., that was in lost packets that we didn't mark lost).

Will work on a unit test for this in a bit.

> This also changes the probing code to suppress redundant keep-alives, i.e., PINGs that we sent for other reasons, which could double as keep-alives but did not.

I suggest  merging this fix through https://github.com/mozilla/neqo/pull/2509.

mxinden:
> > We'd previously only mark 1 one or two packets as lost when a PTO fired. That meant that we potentially didn't RTX all data that we could have (i.e., that was in lost packets that we didn't mark lost).
> 
> Will work on a unit test for this in a bit.

Unit test added through https://github.com/mozilla/neqo/pull/2129/commits/954dc50941f863e0e72e3494339615e461bf0e55.

> > This also changes the probing code to suppress redundant keep-alives, i.e., PINGs that we sent for other reasons, which could double as keep-alives but did not.
> 
> I suggest merging this fix through #2509.

Let's wait for #2509 to merge. Once merged, I will update this pull request, thus reducing the diff.

neqo-transport/src/recovery/mod.rs
larseggert: Isn't this what we are doing now? IIRC one issue with that was that with coalesced packets during the handshake, taking two wasn't enough, because they might have been coalesced (and will be again on TX) and so we sent too few packets.

Maybe we can mark all lost until the handshake completes, and take two afterwards?

mxinden: > Isn't this what we are doing now?

Sorry for the confusion. Let me rephrase my suggestion above.

Currently we take up to 2 previously sent packets. I suggest to take as many previously sent packets, as would fit in the (up to) two upcoming PTO packets.

E.g. following the scenario of the unit test added in this pull request, all 10 previously sent packets, each very small in size, would fit in the to-be-sent PTO packets. Thus each of them would be marked as "lost" and thus retransmitted in the PTO packets.

mxinden: Here is an implementation of the suggestion above: https://github.com/larseggert/neqo/pull/35

larseggert: Ah, I get it now. Yes, this is nice!

larseggert:
@mxinden anything left to be done for this one?

mxinden (approved):
Thanks for the review of the latest additions. This is ready to merge from my end.

I will create a follow-up pull request to reduce the indentation in `maybe_fire_pto`.

PR #2131, opened 2024-09-20
ci: Use Rust stable to benchmark

larseggert:
Nightly might have performance regressions.


mxinden (approved):
Fine by me in general.

I assume you don't have a particular recent slow-down in nightly in mind?

larseggert:
I do; https://github.com/mozilla/neqo/pull/2130#issuecomment-2363502868

PR #2133, opened 2024-09-24
ci: Use `vmactions` for checks on other platforms

larseggert:
Because building NSS and NSPR from source on these tier-2/3 platforms is a pain, the workflow relies on the packaged system NSS version. If that is too old for the current neqo code, checks for that OS are skipped.


larseggert:
@mxinden is the test failure legit? https://github.com/mozilla/neqo/actions/runs/12704862858/job/35414911552?pr=2133#step:5:3588

mxinden:
Thanks for the ping @larseggert. Failure should be fixed with https://github.com/mozilla/neqo/pull/2339.

larseggert:
FreeBSD now works, OpenBSD and Solaris have too old NSS packages.

On NetBSD, `test_fixture::fixture_init()` fails with `The certificate/key database is in an old, unsupported format.`, which I don't quite know what to do about. CC @martinthomson for when you're back.

mxinden (approved):
Overall looks good to me. I am fine merging early here to make sure we don't regress on the working platforms in the meantime.

I assume there is no easy way to reduce some of the repetition?

On a more general note, do I understand correctly that we are supporting these platforms on a best effort basis only?

larseggert:
> I assume there is no easy way to reduce some of the repetition?

We could move the scripts out of the `.yml` into a shared `.sh`, but that would then need to differentiate between the platforms. Or maybe we can dynamically build the script with workflow steps before passing it to the `vmaction` step for execution?

 > On a more general note, do I understand correctly that we are supporting these platforms on a best effort basis only?

Totally. This is mostly to see if we can easily get things working for them.

Issue #2132, opened 2024-09-23
Crash in [@ neqo_common::codec::Encoder::encode_uint<T>]


mxinden:
See [bugzilla#1919678](https://bugzilla.mozilla.org/show_bug.cgi?id=1919678).

Investigating ...

mxinden:
Investigation thus far.

Panic occurs here:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-common/src/codec.rs#L309

Within `Encoder::encode_uint`:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-common/src/codec.rs#L302-L314

Called from `PacketBuilder::pn`:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/packet/mod.rs#L321-L354

Called by `Connection::add_packet_number`:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/connection/mod.rs#L1991-L2007

`n` can not be `>8` due to the line below where `MAX_PACKET_NUMBER_LEN` is `4`:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/packet/mod.rs#L343

In debug mode `n` can also not be `0` due to the line right after:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/packet/mod.rs#L344

But I assume in release mode it can.

Thus this `pn_len` calculation must somehow return `0`:

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/connection/mod.rs#L2017-L2020

For that, `unacked_range` must be `0`.

In order for `unacked_range` to be `0`, `largest_acknowledged` must be equal to `pn`.

In other words, my suspicion is that we are trying to write the packet number of the currently largest acknowledged packet.

mxinden:
Provoking the panic itself at the level of `Connection::add_packet_number()` by providing an equal `largest_acknowledged` and `next_pn` is easy:

``` rust
#[test]
fn add_packet_number() {
    test_fixture::fixture_init();
[...]

PR #2134, opened 2024-09-24
Extend panic when pn_len is 0 with metadata

mxinden:
`Connection::add_packet_number -> PacketBuilder::pn -> Encoder::encode_uint` panics when `pn_len` is `0`. These panics are seen in Firefox crash reports.

To be able to find the root cause of the panic, add additional metadata.

See https://github.com/mozilla/neqo/issues/2132 for details.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2000,7 +2000,10 @@ impl Connection {
         // Count how many bytes in this range are non-zero.
         let pn_len = mem::size_of::<PacketNumber>()
             - usize::try_from(unacked_range.leading_zeros() / 8).unwrap();
-        // pn_len can't be zero (unacked_range is > 0)
+        assert!(
+            pn_len > 0,
+            "pn_len can't be zero as unacked_range should be > 0, pn {pn}, largest_acknowledged {largest_acknowledged:?}, tx {tx}"
+        );
         // TODO(mt) also use `4*path CWND/path MTU` to set a minimum length.
         builder.pn(pn, pn_len);
         pn


neqo-transport/src/connection/mod.rs
```diff
@@ -2000,7 +2000,10 @@ impl Connection {
         // Count how many bytes in this range are non-zero.
         let pn_len = mem::size_of::<PacketNumber>()
             - usize::try_from(unacked_range.leading_zeros() / 8).unwrap();
-        // pn_len can't be zero (unacked_range is > 0)
+        assert!(
+            pn_len > 0,
+            "pn_len can't be zero as unacked_range should be > 0, pn {pn}, largest_acknowledged {largest_acknowledged:?}, tx {tx}"
+        );
```
mxinden, line 2006: Instead of panicing lower down in the call stack in `Encoder::encode_uint`, panic here already, given that more metadata is available.

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-common/src/codec.rs#L309

neqo-transport/src/connection/mod.rs
```diff
@@ -2000,7 +2000,10 @@ impl Connection {
         // Count how many bytes in this range are non-zero.
         let pn_len = mem::size_of::<PacketNumber>()
             - usize::try_from(unacked_range.leading_zeros() / 8).unwrap();
-        // pn_len can't be zero (unacked_range is > 0)
+        assert!(
+            pn_len > 0,
+            "pn_len can't be zero as unacked_range should be > 0, pn {pn}, largest_acknowledged {largest_acknowledged:?}, tx {tx}"
```
mxinden, line 2005: I assume it is fine to use `tx` (instance of `CryptoDxState`) in a `assert!` output. In other words, I assume neither `CryptoDxState`'s `epoch` nor `direction` are sensitive information.

https://github.com/mozilla/neqo/blob/d513712bc9d06e297af5cf317a9f6dd70e69a88b/neqo-transport/src/crypto.rs#L716-L720

martinthomson (approved):
I'm sure that this is fine.

mxinden:
>  codecov/patch 66.66% of diff hit (target 95.35%) 

Oh, what is happening here? Taking a look now.

neqo-transport/src/connection/mod.rs
mxinden, line 2005: Seems like codecov is failing, given that this particular line, i.e. the formatting in the case of a panic, is not covered by tests.

https://app.codecov.io/gh/mozilla/neqo/commit/f4af74c4a9fcbeb1b95092f180ea07bd7f110611

I don't think it makes sense to write tests for each of our assertions. If you think differently, I can add a test for this particular one.

PR #2135, opened 2024-09-26
feat: Stop the PMTUD search at the interface MTU

larseggert:
Should we also optimistically *start* the search at the interface MTU, and only start from 1280 when that fails?


mxinden:
> Should we also optimistically start the search at the interface MTU

Are there other projects using this optimistic approach?

If I understand RFC 8899 correctly the local interface MTU is the end value, not the start value.

> The MAX_PLPMTU is the largest size of PLPMTU. This has to be less than or equal to the maximum size of the PL packet that can be sent on the outgoing interface (constrained by the local interface MTU).

https://www.rfc-editor.org/rfc/rfc8899.html#section-5.1.2

larseggert:
This PR exposed a bug in `mtu` 🫤 https://github.com/mozilla/mtu/pull/26

mxinden:
>> > Should we also optimistically start the search at the interface MTU
>> 
>> Are there other projects using this optimistic approach?
>> 
>> If I understand RFC 8899 correctly the local interface MTU is the end value, not the start value.
>> 
>> > The MAX_PLPMTU is the largest size of PLPMTU. This has to be less than or equal to the maximum size of the PL packet that can be sent on the outgoing interface (constrained by the local interface MTU).
>> 
>> https://www.rfc-editor.org/rfc/rfc8899.html#section-5.1.2
>
> All true, but in practice, the local interface is most often the limiting hop.

Let me make sure I understand the implications here correctly. Sorry for any potential mistakes.

We only start probing once the connection is confirmed.

https://github.com/mozilla/neqo/blob/55e3a9363c28632dfb29ce91c7712cab1f6a58da/neqo-transport/src/connection/mod.rs#L2794-L2802

Say that a client's path MTU is smaller than their local interface MTU. Given that probing only starts once confirmed, i.e. after receiving `HANDSHAKE_DONE` from the server, initial HTTP requests would not be delayed, but only one subsequent flight of requests would be delayed by up to one PTO. Is that correct?

Thus this optimization, and really all of PMTUD probing, assumes that the potential delay of one subsequent flight of HTTP requests by up to one PTO is worth the trade off of potentially increasing the overall connection throughput.

Is that correct?

larseggert:
> > > > Should we also optimistically start the search at the interface MTU
> 
> Let me make sure I understand the implications here correctly. Sorry for any potential mistakes.
> We only start probing once the connection is confirmed.

This would need to change. What I think we should do is roughly this:

* Start sending at the local interface MTU when the connection starts (i.e., for the Initial).
* If we detect a loss *n* times, we revert to pobing up from 1280.

*n* should probably be something like 2, so we don't cause undue delay.

mxinden:
In the case where a client's path MTU is smaller than their local interface MTU, this would add a delay of 2*PTO to every connection establishment, right? If so, isn't that a high cost for the potential increase in throughput? Or is such scenario just very rare?

larseggert:
Yes. I _think_ this is a rare case, but maybe we add some telemetry first to confirm?

mxinden:
> Yes. I _think_ this is a rare case, but maybe we add some telemetry first to confirm?

How about we:

1. Role out PMTUD on Firefox Nightly without the optimization (i.e. starting at the local interface MTU).
2. Enable the optimization, monitoring whether connection establishment times stay stable.

larseggert:
I was thinking we just log the log the local interface MTU together with the discovered PMTUD, and check for differences.

neqo-transport/Cargo.toml
```diff
@@ -22,6 +22,7 @@ indexmap = { version = "2.2", default-features = false } # See https://github.co
 log = { workspace = true }
 neqo-common = { path = "../neqo-common" }
 neqo-crypto = { path = "../neqo-crypto" }
+mtu = { path = "../../mtu", default-features = false }
```
mxinden, line 25: Referencing your local `mtu` repository?

In case you haven't seen it, for local development I find `cargo`'s [dependency override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) feature handy.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -614,6 +614,7 @@ mod tests {
     };
 
     const IP_ADDR: IpAddr = IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0));
+    const MTU: usize = 1_500;
```
mxinden, line 617: Given that each `mod tests` in `neqo-transport/src/cc/` needs `MTU`, how about defining it only once at a higher level?

neqo-transport/src/path.rs
```diff
@@ -569,7 +571,20 @@ impl Path {
         qlog: NeqoQlog,
         now: Instant,
     ) -> Self {
-        let mut sender = PacketSender::new(cc, pacing, Pmtud::new(remote.ip()), now);
+        // We're passing `None` into `mtu::interface_and_mtu` here, which force the lookup to only
```
mxinden, line 574:
```suggestion
        // We're passing `None` into `mtu::interface_and_mtu` here, which forces the lookup to only
```

neqo-transport/src/path.rs
```diff
@@ -569,7 +571,20 @@ impl Path {
...
+        // take the local address into account. Taking the remote address into account would require
+        // a call to `connect`, which Firefox blocklists for Rust dependencies.
```
mxinden, line 576: Uff :slightly_frowning_face:

neqo-transport/src/path.rs
```diff
@@ -569,7 +571,20 @@ impl Path {
...
+        let iface_mtu = match mtu::interface_and_mtu(&(local.ip(), remote)) {
```
mxinden, line 577: Is the above comment still valid? Doesn't look like you are passing `None`.

mxinden (commented):
Generally looking good to me. Minor comments.

larseggert:
We need to bump the [`rayon` crate in mozilla-central](https://searchfox.org/mozilla-central/source/Cargo.lock#5149-5157) to `1.7`, which is what `windows-bindgen` requires, before we can use land this PR and bring in `mtu` as a dependency.

@mxinden @valenting @KershawChang any tips on how to do this?

mxinden:
We will need to update the following crates:

```
➜  cargo tree --invert=rayon
rayon v1.6.1
├── style v0.0.1
│   ├── geckoservo v0.0.1
│   │   └── gkrust-shared v0.1.0
│   │       ├── gkrust v0.1.0
│   │       └── gkrust-gtest v0.1.0
│   └── stylo_tests v0.0.1
│       [dev-dependencies]
│       └── gkrust v0.1.0
├── webrender v0.62.0
│   └── webrender_bindings v0.1.0
│       └── gkrust-shared v0.1.0
├── webrender_bindings v0.1.0
└── wr_glyph_rasterizer v0.1.0
    └── webrender v0.62.0

```

For the record, `windows-bindgen` introduced `rayon` `v1.7` dependency in [`windows-bindgen` `v0.49.0`](https://crates.io/crates/windows-bindgen/0.49.0). If it would have been a more recent version, it might have been worth exploring supporting both the most recent and the one without `rayon`, similar to what we did in https://github.com/quinn-rs/quinn/actions/runs/11553182968/job/32153809202?pr=2021.

mxinden:
```
➜  mozilla-central git:(bookmarks/central) cargo update -p rayon  --precise 1.7.0
    Updating crates.io index
    Updating rayon v1.6.1 -> v1.7.0
```

Neat. It seems to apply without issues.

@larseggert I can create a mozilla-central Phabricator patch if you want. In case there aren't any hidden issues, the only real work will be auditing `rayon` `v1.6.1 -> v1.7.0` code changes.

larseggert:
> @larseggert I can create a mozilla-central Phabricator patch if you want. In case there aren't any hidden issues, the only real work will be auditing `rayon` `v1.6.1 -> v1.7.0` code changes.

That would be great, thanks. (I'm surprised mozilla-central doesn't have some sort of auto-upgrade of dependencies...)

mxinden:
For the record, updating mozilla-central to a recent `rayon` version is happening through https://bugzilla.mozilla.org/show_bug.cgi?id=1933199.

mxinden:
> For the record, updating mozilla-central to a recent `rayon` version is happening through https://bugzilla.mozilla.org/show_bug.cgi?id=1933199.

`rayon` `v1.10.0` landed in mozilla-central last night.

larseggert:
Thanks, saw it. Trying to make `mtu` importable.

larseggert:
[Next blocker:](https://github.com/larseggert/neqo/actions/runs/12081485157/job/33690659252#step:8:98)
```
Missing audit for windows-bindgen:0.58.0 (requires ['safe-to-deploy']).
```

larseggert:
@mxinden should we update more deps on m-c before merging this, or do we want to handle that when we vendor neqo into m-c the next time?

mxinden:
@larseggert other than the bump of `rayon` which is in mozilla-central now, what else is needed?

larseggert:
> Other than the `rayon` (:heavy_check_mark:) update and the `windows-bindgen` audit, what else is missing @larseggert?

I think `windows-metadata`, but that is in the same category as `windows-bindgen`.

mxinden:
If it is audits only, I suggest to "handle that when we vendor neqo into m-c the next time".

Though note, auditing `windows-bindgen` from scratch might be a significant undertaking.

mxinden (commented):
Do any of the existing unit tests cover the case where `Pmtud` stops at the interface MTU?

larseggert:
> Though note, auditing `windows-bindgen` from scratch might be a significant undertaking.

I don't think we need to - the author is trusted.

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -22,9 +20,7 @@ use crate::{
     rtt::RttEstimate,
 };
 
-const IP_ADDR: IpAddr = IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0));
-const PTO: Duration = Duration::from_millis(100);
-const RTT: Duration = Duration::from_millis(98);
+const PTO: Duration = RTT;
```
mxinden, line 23: I assume it is fine that `PTO` is now equal to `RTT`, while previously it was `RTT - 2`.

mxinden (approved):
> Do any of the existing unit tests cover the case where Pmtud stops at the interface MTU?

Only remaining question. Otherwise this is ready to merge from my end.

neqo-transport/src/cc/tests/new_reno.rs
larseggert: It doesn't seem to cause a test failure 🤷

PR #2137, opened 2024-09-28
perf: use recvmmsg in addition to GRO 

mxinden:
Previously we would only do GRO.

Depends on https://github.com/mozilla/neqo/pull/2093.

**Draft only. Just testing on the benchmark runner for now.**


mxinden:
`recvmmsg` is still worth doing, at least on MacOS with https://github.com/quinn-rs/quinn/pull/1993. That said, likely easiest to start from scratch. 

Work tracked in https://github.com/mozilla/neqo/issues/1693. Closing here.

PR #2138, opened 2024-09-29
perf: use GSO 

mxinden:
Depends on https://github.com/mozilla/neqo/pull/2093. All, but last commit, part of #2093.

**Proof of concept for now. Thus Draft.**

Compiles. _Download_ benchmark works. Without PMTUD I see up to 64 UDP segments on a GSO `sendmsg` call.


mxinden:
If we do want to tackle GSO in the future, this can serve as guidance. That said, best to start fresh.

Work tracked in https://github.com/mozilla/neqo/issues/1693. Closing here.

Issue #2141, opened 2024-10-01
ci: macos failing with bindings/nspr_io.h:7:10: fatal error: 'prio.h' file not found


mxinden:
CI on MacOS is currently failing with:

> bindings/nspr_io.h:7:10: fatal error: 'prio.h' file not found

See e.g. https://github.com/mozilla/neqo/actions/runs/11112117096/job/30873528067?pr=2136.

I am looking into it now.

Happening on e.g. #2136, https://github.com/mozilla/neqo/pull/2093

mxinden:
Maybe related to the fact that we are now using the homebrew NSS version?

Previously (succeeding):

> System NSS is too old: 3.104

https://github.com/mozilla/neqo/actions/runs/11029596600/job/30632387433

Now (failing):

> System NSS is suitable: 3.105

https://github.com/mozilla/neqo/actions/runs/11112117096/job/30873528067?pr=2136

mxinden:
Unable to reproduce on an ARM Mac locally thus far.

larseggert:
This is an NSPR header.

mxinden:
As far as I can tell, this is what is happening:

1. On MacOS we install NSS via brew. NSPR is installed as a dependency of NSS.
2. We check the NSS version. Recently the brew version updated to 3.105, and is now considered valid.
3. We would skip all the local build steps given that `BUILD_NSS` would be `0`.
4. We would still set `NSS_DIR` to a local NSS installation. Note that `NSS_DIR` indirectly also determines the NSPR installation location.
5. Building `neqo-crypto` would fail, given that `NSS_DIR` would (indirectly) point to a non-existing local NSPR installation and thus imports like the one below wouldn't resolve.
  https://github.com/mozilla/neqo/blob/55e3a9363c28632dfb29ce91c7712cab1f6a58da/neqo-crypto/bindings/nspr_io.h#L7

https://github.com/mozilla/neqo/pull/2142 fixes the problem.

larseggert:
I think we need a new env variable to indicate that a pkf-config NSS should be used.

PR #2142, opened 2024-10-01
ci(nss): handle system-, cached-, and to-be-build-NSS

mxinden:
When the system NSS version is sufficient:

https://github.com/mozilla/neqo/blob/55e3a9363c28632dfb29ce91c7712cab1f6a58da/.github/actions/nss/action.yml#L31

Then one must not set the environment variables pointing to a local copy:

https://github.com/mozilla/neqo/blob/55e3a9363c28632dfb29ce91c7712cab1f6a58da/.github/actions/nss/action.yml#L177-L190

Fixes https://github.com/mozilla/neqo/issues/2141.

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -176,6 +176,7 @@ runs:
 
     - name: Set up environment
       shell: bash
+      if: env.BUILD_NSS == '1'
       run: |
         NSS_TARGET="${{ inputs.type }}"
         echo "NSS_TARGET=$NSS_TARGET" >> "$GITHUB_ENV"
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -33,18 +33,18 @@ runs:
       run: |
         if ! command -v pkg-config &> /dev/null; then
           echo "pkg-config: not found"
-          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          echo "USE_SYSTEM_NSS=0" >> "$GITHUB_ENV"
           exit 0
         fi
         if ! pkg-config --exists nss; then
           echo "pkg-config: NSS not found"
-          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          echo "USE_SYSTEM_NSS=0" >> "$GITHUB_ENV"
           exit 0
         fi
         NSS_VERSION="$(pkg-config --modversion nss)"
         if [ "$?" -ne 0 ]; then
           echo "pkg-config: failed to determine NSS version"
-          echo "BUILD_NSS=1" >> "$GITHUB_ENV"
+          echo "USE_SYSTEM_NSS=0" >> "$GITHUB_ENV"
           exit 0
         fi
         NSS_MAJOR=$(echo "$NSS_VERSION" | cut -d. -f1)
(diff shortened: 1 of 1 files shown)


larseggert:
Is this fixing a bug?

When a prebuilt NSS is downloaded from the GitHub cache, `BUILD_NSS` is set to 0 but the environment still needs to be set up so it is found by the neqo build process.

mxinden:
> coalesce_acked_from_zero 10+1 entries: 💚 Performance has improved.
> 
>        time:   [116.48 ns 116.99 ns 117.58 ns]
>        change: [-39.843% -35.572% -33.050%] (p = 0.00 < 0.05)
> 
> Found 15 outliers among 100 measurements (15.00%)

Ufff, a 35% performance improvement on an unrelated change :thinking:

mxinden:
> Is this fixing a bug?

Yes, this fixes https://github.com/mozilla/neqo/issues/2141.

> When a prebuilt NSS is downloaded from the GitHub cache, `BUILD_NSS` is set to 0 but the environment still needs to be set up so it is found by the neqo build process.

When `NSS_DIR` isn't set, `pkg-config` is used to find the system-wide installation of NSS:

https://github.com/mozilla/neqo/blob/55e3a9363c28632dfb29ce91c7712cab1f6a58da/neqo-crypto/build.rs#L428-L434

Thus I don't think the environment variables need to be set.

@larseggert am I missing something?

larseggert:
You are right that if `NSS_DIR` isn't set, we try and use the system NSS.

But if we need to use an NSS that was build from source, we need to set the environment variables, so that neqo can link against the NSS that was build by (or cache-restored by) the GitHub action.

With your change, is it still using the prebuilt-from-source NSS, or is it building it again as part of neqo?

mxinden:
I think I understand know. Thanks @larseggert. I missed that `BUILD_NSS` is overwritten on cache hit:

https://github.com/mozilla/neqo/blob/a484f64f64da94376f05f11cf68b8894cab18bd0/.github/actions/nss/action.yml#L124-L125

Let me take another look.

.github/actions/nss/action.yml
```diff
@@ -177,17 +178,25 @@ runs:
...
+          NSS_OUT="$NSS_DIR/../dist/$NSS_TARGET"
+          echo "LD_LIBRARY_PATH=$NSS_OUT/lib" >> "$GITHUB_ENV"
+          echo "DYLD_FALLBACK_LIBRARY_PATH=$NSS_OUT/lib" >> "$GITHUB_ENV"
+          echo "$NSS_OUT/lib" >> "$GITHUB_PATH"
+          echo "NSS_DIR=$NSS_DIR" >> "$GITHUB_ENV"
+          echo "NSS_PREBUILT=1" >> "$GITHUB_ENV"
+        fi
       env:
         NSS_DIR: ${{ github.workspace }}/nss
-        NSPR_DIR: ${{ github.workspace }}/nspr
```
mxinden, line 190: As far as I can tell, this environment variable is never used.

.github/actions/nss/action.yml
```diff
@@ -86,21 +86,21 @@ runs:
         fi
 
     - name: Checkout NSS
-      if: env.BUILD_NSS == '1'
+      if: env.USE_SYSTEM == '0'
```
martinthomson, line 89: USE_SYSTEM_NSS !

mxinden: Uff, good catch. Thank you @martinthomson! Will be fixed in https://github.com/mozilla/neqo/pull/2145.

PR #2147, opened 2024-10-02
ci(bench): don't run on draft PRs

mxinden:


.github/workflows/check.yml
```diff
@@ -99,4 +99,5 @@ jobs:
 
   bench:
     needs: [check]
+    if: ${{ github.event.pull_request.draft == false }}
```
martinthomson, line 102: It would be good if this were possible with workflow_dispatch.  Maybe I can suggest an argument to workflow_dispatch that defaults to false for draft pull requests...

mxinden: I pushed https://github.com/mozilla/neqo/pull/2147/commits/27d2c0454473626667e2b4592433faae93a5a52a. Is that what you had in mind @martinthomson?

martinthomson (approved):

PR #2150, opened 2024-10-07
fix: Check that the largest_acked was sent

larseggert:
This is a test and fix for the issue we're discussing with Avast.

It also includes an improvement to let `handle_ack` derive `largest_acked` on the fly, rather than needing to pass it in as an argument.

CC @mxinden


neqo-transport/src/connection/mod.rs
```diff
@@ -3047,23 +3067,34 @@ impl Connection {
...
+            return Ok(());
         };
+
+        // Ensure that the largest acknowledged packet number was actually sent.
+        // (If we ever start using non-contigous packet numbers, we need to check all the packet
+        // numbers in the ACKed ranges.)
+        if largest_acknowledged >= next_pn {
+            qwarn!("Largest ACKed {} was never sent", largest_acknowledged,);
+            return Err(Error::AckedUnsentPacket);
+        }
```
mxinden, line 3096: Instead of validating the untrusted value `largest_acknowledged`, can we instead derive it ourself, looking at the largest packet in `acked_packets` below, after calling `self.loss_recovery.on_ack_received`?

`acked_packets` should only contain packets we have sent before and thus a `largest_acknowledged` derived from `acked_packets` doesn't need the validation above.

mxinden (commented):
Great catch! Thank you.

neqo-transport/src/connection/mod.rs
mxinden: https://github.com/larseggert/neqo/pull/31 implements this proposal @larseggert. Note the PR description:

> Note that this pull request does not consider the remote behavior as malicious. It simply ignores the largest ACK, if it doesn't correspond to a packet previously sent.

neqo-transport/src/connection/mod.rs
```diff
@@ -2853,16 +2872,17 @@ impl Connection {
                 ack_ranges,
                 ecn_count,
             } => {
+                // Ensure that the largest acknowledged packet number was actually sent.
+                // (If we ever start using non-contiguous packet numbers, we need to check all the
+                // packet numbers in the ACKed ranges.)
+                if largest_acknowledged >= next_pn {
+                    qwarn!("Largest ACKed {} was never sent", largest_acknowledged,);
```
mxinden, line 2879:
```suggestion
                    qwarn!("Largest ACKed {} was never sent", largest_acknowledged);
```

neqo-transport/tests/connection.rs
```diff
@@ -132,6 +132,70 @@ fn reorder_server_initial() {
...
+        .decrypt(pn, &header, &payload[pn_len..], &mut buf)
+        .unwrap()
+        .to_owned();
+
+    // Now we need to find the frames.  Make some really strong assumptions.
+    let mut dec = Decoder::new(&plaintext[..]);
+    assert_eq!(dec.decode(ACK_FRAME.len()), Some(ACK_FRAME));
+    assert_eq!(dec.decode_varint(), Some(0x06)); // CRYPTO
+    assert_eq!(dec.decode_varint(), Some(0x00)); // offset
+    dec.skip_vvec(); // Skip over the payload.
```
mxinden, line 172: Given that `Decoder` takes in immutable reference to `plaintext`, in other words, does no side-effects, why are these and especially the last line needed?

larseggert: Lines were copied from existing tests without further reflection :-)

neqo-transport/tests/connection.rs
```diff
@@ -132,6 +132,70 @@ fn reorder_server_initial() {
...
+    let spoofed = Datagram::new(
+        server_initial.source(),
+        server_initial.destination(),
+        server_initial.tos(),
+        packet,
+    );
+
+    // Now deliver the packet with the spoofed ACK frame
+    client.process_input(&spoofed, now());
+    client.process_input(&server_hs.unwrap(), now());
```
mxinden, line 192:
```suggestion
```

I don't think this is necessary. As one would expect `client.state()` is already `State::Closed` after the line before.

neqo-transport/tests/connection.rs
```diff
@@ -132,6 +132,70 @@ fn reorder_server_initial() {
...
+
+    // Now we need to find the frames.  Make some really strong assumptions.
+    let mut dec = Decoder::new(&plaintext[..]);
+    assert_eq!(dec.decode(ACK_FRAME.len()), Some(ACK_FRAME));
+    assert_eq!(dec.decode_varint(), Some(0x06)); // CRYPTO
+    assert_eq!(dec.decode_varint(), Some(0x00)); // offset
+    dec.skip_vvec(); // Skip over the payload.
+
+    // Overwrite largest_acked in the ACK frame with 3 (a packet that was never sent).
+    plaintext[1] = 0x3;
```
mxinden, line 175:
```suggestion
    assert_eq!(plaintext[1], 0x0);
    plaintext[1] = 0x3;
```

Feel free to ignore. Encoding the "really strong assumptions" above might make debugging this test easier in the future.

mxinden (approved):
Apart from the outstanding comments on the added unit test, this pull request looks good to me.

neqo-transport/tests/connection.rs
larseggert: If we want to remove those, we should probably do a separate PR and remove them from all similar tests?

mxinden (approved):

neqo-transport/tests/connection.rs
```diff
@@ -132,6 +132,70 @@ fn reorder_server_initial() {
     assert_eq!(*client.state(), State::Confirmed);
 }
 
+/// Receiving an ACK frame for a packet number that was never sent is an error.
+#[test]
+fn ack_for_unsent() {
```
martinthomson, line 137: A driveby for this one...  Did you consider just injecting an ACK frame into the packet?  That might have been easier than this, which is pretty significant in size.  It's only ACK + LA=100000 + 0 + 0.

larseggert: You mean via frame_writer?

martinthomson: Right.  `test_frame_writer`.

PR #2159, opened 2024-10-08
wrappers for common usage of test_frame_writer

martinthomson:
This reduces boilerplate and encapsulates a good pattern of usage. I wasn't able to make `test_frame_writer` private, but this is close enough.


mxinden (approved):
Neat.

More generally, I wasn't aware of `Connection::test_frame_writer`. Great for https://github.com/mozilla/neqo/pull/2150!

PR #2160, opened 2024-10-08
fix(transport): remove unused Error::HandshakeFailed

mxinden:


larseggert:
Was that ever used? Or was the code that used it maybe inadvertently removed, like for `AckedUnsentPacket`?

mxinden:
Instantiation of `Error::HandshakeFailed` was removed in https://github.com/mozilla/neqo/commit/03238d471118402b1b15d3b723144c7b5e91d9b5, replaced by `Error::CryptoAlert`.

larseggert:
@mxinden it seems to be used in the glue: https://github.com/mozilla/neqo/actions/runs/11251536635/job/31282794009?pr=2163#step:9:1163

mxinden:
Yes, `neqo_glue` handles `TransportError::HandshakeFailed`, but never instantiates it.

``` rust
            TransportError::HandshakeFailed => CloseError::TransportInternalErrorOther(6),
```
https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/netwerk/socket/neqo_glue/src/lib.rs#1004

larseggert:
Ok, so we just need to remove it there.

Really wish we could compile the glue together with neqo somehow in CI (without the full Fx).

mxinden:
I guess we could run `./mach build netwerk` in `firefox.yml`:

https://github.com/mozilla/neqo/blob/f39e12ef5c738e7a79e30448c964293adec112c7/.github/workflows/firefox.yml#L122

That would still suffice as a smoke test, but won't give us a full Firefox build.

larseggert:
I was hoping we could just do a sparse checkout and build the `neqo_glue` crate...

PR #2161, opened 2024-10-08
ci: Don't run QNS when a PR is merged

larseggert:
Also don't run it on a schedule anymore.

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -1,16 +1,10 @@
 name: QNS
-
 on:
-  push:
-    branches: ["main"]
-    paths-ignore: ["*.md", "*.png", "*.svg", "LICENSE-*"]
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/qns.yml
```diff
@@ -1,16 +1,10 @@
 name: QNS
-
 on:
-  push:
-    branches: ["main"]
-    paths-ignore: ["*.md", "*.png", "*.svg", "LICENSE-*"]
```
mxinden, line 6: Say we merge multiple pull requests in a row. After the merges, a QNS test fails. How do we quickly find out which merge introduced the regression?

mxinden (commented):
Fine with not running on closed pr and scheduled, but wonder whether we should keep `push`.

mxinden (approved):

PR #2164, opened 2024-10-09
ci: Build only `netwerk` component on CI

larseggert:
Save some trees by only building the `netwerk` bits of Firefox in CI. To build the entire browser, do a manual `workload_dispatch`.


larseggert:
@mxinden the `./mach build netwerk` bit you suggested is failing. Any ideas?

mxinden:
Sorry for leading you astray here. As discussed on Slack, I guess there is no way to build `netwerk` only.

larseggert:
This seems to be impossible.

PR #2165, opened 2024-10-10
test: Dump simulator RNG seeds

### LINKED ISSUE #1645: Output seed to stdout for tests
This is harder than it sounds.

The "simulator" we use in some of our randomized tests uses a random seed.  The idea is that if a random test fails, we could take that seed and reproduce the exact failure conditions again to debug them.  Originally, this seed was printed to stdout, which would be captured by the test runner, then later validated.  Unfortunately, when I used the same code for benchmarking, stdout was not captured.  I had to suppress this message.

Ideally, if we are benchmarking, we would suppress this output, but in regular test runs we wouldn't.  It is possible that we could test `debug_assertions` for this on that basis that benchmarks run in release mode and tests run in debug mode, but I'm not 100% confident that this is the right way to do that.

larseggert:
Into the directory given in the `DUMP_SIMULATION_SEEDS` environment variable. Also, export them as artifacts from the CI runs.

Fixes #1645

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -129,6 +129,7 @@ jobs:
         run: ls -1 ${{ env.DUMP_SIMULATION_SEEDS }}
 
       - name: Save simulation seeds artifact
+        if: env.DUMP_SIMULATION_SEEDS
         uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
         with:
           name: simulation-seeds-${{ matrix.os }}-${{ matrix.rust-toolchain }}-${{ matrix.type }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -111,6 +111,9 @@ jobs:
         env:
           RUST_LOG: trace
         run: |
+          if [ "${{ env.DUMP_SIMULATION_SEEDS }}" ]; then
+            export DUMP_SIMULATION_SEEDS=${{ env.DUMP_SIMULATION_SEEDS }}/sanitizer
+          fi
           if [ "${{ matrix.os }}" = "ubuntu-latest" ]; then
             TARGET="x86_64-unknown-linux-gnu"
             SANITIZERS="address thread leak"
@@ -126,14 +129,14 @@ jobs:
 
       - name: Print simulation seeds
         if: env.DUMP_SIMULATION_SEEDS
-        run: ls -1 ${{ env.DUMP_SIMULATION_SEEDS }}
+        run: find ${{ env.DUMP_SIMULATION_SEEDS }}
 
       - name: Save simulation seeds artifact
         if: env.DUMP_SIMULATION_SEEDS
         uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
         with:
           name: simulation-seeds-${{ matrix.os }}-${{ matrix.rust-toolchain }}-${{ matrix.type }}
-          path: /tmp/simulation-seeds
+          path: ${{ env.DUMP_SIMULATION_SEEDS }}
           compression-level: 9
 
   bench:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -111,9 +111,6 @@ jobs:
         env:
           RUST_LOG: trace
         run: |
-          if [ "${{ env.DUMP_SIMULATION_SEEDS }}" ]; then
-            export DUMP_SIMULATION_SEEDS=${{ env.DUMP_SIMULATION_SEEDS }}/sanitizer
-          fi
           if [ "${{ matrix.os }}" = "ubuntu-latest" ]; then
             TARGET="x86_64-unknown-linux-gnu"
             SANITIZERS="address thread leak"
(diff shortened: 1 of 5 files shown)


test-fixture/src/sim/mod.rs
```diff
@@ -151,23 +148,37 @@ impl Simulator {
...
-    pub fn seed(&mut self, seed: [u8; 32]) {
-        self.rng = Rc::new(RefCell::new(Random::new(&seed)));
+        // Dump the seed to a file to the directory in the `DUMP_SIMULATION_SEEDS` environment
+        // variable, if set.
+        if let Ok(dir) = std::env::var("DUMP_SIMULATION_SEEDS") {
+            if create_dir_all(&dir).is_err() {
+                qerror!("Failed to create directory {dir}");
+            } else {
+                let seed_str = sim.rng.borrow().seed_str();
+                let path = format!("{dir}/{seed_str}");
```
mxinden, line 167: Instead of manual path formatting, I prefer to use [`PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html), taking care of all the idiosyncrasies of the various platforms.

mxinden (approved):
Just

.github/workflows/check.yml
```diff
@@ -118,9 +121,25 @@ jobs:
...
+            if [ "${{ env.DUMP_SIMULATION_SEEDS }}" ]; then
+              export DUMP_SIMULATION_SEEDS="${{ env.DUMP_SIMULATION_SEEDS }}/sanitizer/$sanitizer"
+            fi
             RUSTFLAGS="-Z sanitizer=$sanitizer" RUSTDOCFLAGS="-Z sanitizer=$sanitizer" cargo +nightly nextest run -Z build-std --features ci --target "$TARGET"
           done
 
+      - name: Print simulation seeds
+        if: env.DUMP_SIMULATION_SEEDS
+        run: find ${{ env.DUMP_SIMULATION_SEEDS }}
+
```
martinthomson, line 134: Presumably you don't need this...

```suggestion
```

.github/workflows/check.yml
```diff
@@ -68,12 +69,14 @@ jobs:
           cargo +${{ matrix.rust-toolchain }} check $BUILD_TYPE --all-targets --features ci
 
       - name: Run tests and determine coverage
+        env:
+          RUST_LOG: trace
```
martinthomson, line 73: Do you need this?

test-fixture/src/sim/rng.rs
```diff
@@ -69,8 +69,9 @@ impl Random {
     /// Get the seed necessary to continue from the current state of the RNG.
     #[must_use]
     pub fn seed_str(&self) -> String {
+        // Make sure to print leading zeros.
```
martinthomson, line 72:
```suggestion
```

test-fixture/src/sim/mod.rs
```diff
@@ -151,23 +148,37 @@ impl Simulator {
...
+                if write(&path, seed_str).is_err() {
+                    qerror!("Failed to write seed to {path}");
```
martinthomson, line 169:
```suggestion
                    qerror!("Simulation seed: {seed_str}");
```
Might as well just print the value out here.

test-fixture/src/sim/mod.rs
```diff
@@ -151,23 +148,37 @@ impl Simulator {
...
-        self.rng = Rc::new(RefCell::new(Random::new(&seed)));
+        // Dump the seed to a file to the directory in the `DUMP_SIMULATION_SEEDS` environment
+        // variable, if set.
+        if let Ok(dir) = std::env::var("DUMP_SIMULATION_SEEDS") {
+            if create_dir_all(&dir).is_err() {
+                qerror!("Failed to create directory {dir}");
+            } else {
+                let seed_str = sim.rng.borrow().seed_str();
+                let path = format!("{dir}/{seed_str}");
+                if write(&path, seed_str).is_err() {
```
martinthomson, line 168: If the filename is the seed, why bother with putting anything in it?

```suggestion
                if File::create(&path).is_err() {
```

.github/workflows/check.yml
larseggert: Not strictly, since we also provide them as artifacts.

test-fixture/src/sim/mod.rs
larseggert: It's being printed elsewhere already.

test-fixture/src/sim/mod.rs
larseggert: I find it sometimes easier to `cat` than to `ls` when scripting.

.github/workflows/check.yml
larseggert: We do need to set the env variable, so debug-output-producing code is getting coverage-tested. But we can also set that on each run like before, in case that is what you meant?

martinthomson (approved):

mxinden:
@larseggert do you recall why this is dumping to a file, instead of printing to stdout? We would thus not need an artifact CI step, but can just read the seed from the logs. Debugging #2499 I noticed we don't dump nor artifact the seed for the benchmarks.

larseggert:
I can't recall. Am OK with changing to a debug output.

PR #2166, opened 2024-10-10
ci: Run the sanitizers in parallel

larseggert:
This speeds up the matrix.

Also grab MSRV from `Cargo.toml` instead of hardcoding it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -38,6 +38,11 @@ jobs:
           - os: ubuntu-latest
             rust-toolchain: stable
             type: release
+          - os: [ubuntu-latest, macos-latest]
+            rust-toolchain: nightly
+            type: debug
+            sanitizer: [address thread leak]
+
     env:
       BUILD_TYPE: ${{ matrix.type == 'release' && '--release' || '' }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -25,34 +25,77 @@ concurrency:
...
     steps:
       - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # v4.2.0
 
       - uses: ./.github/actions/rust
         with:
-          version: ${{ matrix.rust-toolchain }}
-          components: ${{ matrix.rust-toolchain == 'stable' && 'llvm-tools-preview' || matrix.rust-toolchain == 'nightly' && 'rust-src' || '' }}
-          tools: ${{ matrix.rust-toolchain == 'stable' && 'cargo-llvm-cov, ' || '' }} cargo-nextest
+          version: ${{ matrix.toolchain }}
+          components: ${{ matrix.toolchain == 'stable' && 'llvm-tools-preview' || matrix.toolchain == 'nightly' && 'rust-src' || '' }}
```
mxinden, line 97:
```suggestion
          components: ${{ matrix.toolchain == 'stable' && 'llvm-tools-preview' || matrix.sanitizer != null && 'rust-src' || '' }}
```

Would be more accurate, right?

mxinden (approved):
Definitely in favor of the `toolchains` job.

In case the gain from parallelizing the sanitizers is worth the complexity, :+1:.

larseggert:
Do you think I should just do a separate workflow for the sanitizer runs, like we do for clippy? May be more maintainable.

larseggert:
@mxinden any ideas on what's causing the two failures?

mxinden (approved):

PR #2169, opened 2024-10-11
refactor(udp): pass SocketAddr instead of &SocketAddr

mxinden:
`SocketAddr` implements `Copy`.

https://doc.rust-lang.org/std/net/enum.SocketAddr.html#impl-Copy-for-SocketAddr

Thus there is no need to pass by reference instead of by value.

Broken out of https://github.com/mozilla/neqo/pull/2093.


larseggert:
Is the size of the object a concern?

mxinden:
> Is the size of the object a concern?

The size is 32 bytes. I would be surprised if this has a performance impact, especially in the context of making a syscall.

Am I missing something?

martinthomson (approved):

PR #2170, opened 2024-10-11
test: More migration testing

larseggert:
This expands the migration test quite a bit. It now tests alll the way until retirement of the original path, for both port-only (`rebinding_port`) and address-and-port (`rebinding_address_and_port`) changes.

`rebinding_address_and_port` is succeeding, but `rebinding_port` is currently failing. That's because we treat it differently for some reason. If we replace `Paths::find_path_with_rebinding` with `Paths::find_path`, i.e., do proper path validation when only the port changes, the test succeeds.

Leaving this out in case I'm missing something about the intent of the difference.

### DIFF
--- a/neqo-transport/src/connection/tests/migration.rs
+++ b/neqo-transport/src/connection/tests/migration.rs
@@ -7,7 +7,7 @@
 use std::{
     cell::RefCell,
     mem,
-    net::{IpAddr, Ipv6Addr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     rc::Rc,
     time::Duration,
 };
@@ -30,9 +30,10 @@ use crate::{
     packet::PacketBuilder,
     path::MAX_PATH_PROBES,
     pmtud::Pmtud,
+    stats::FrameStats,
     tparams::{self, PreferredAddress, TransportParameter},
     CloseReason, ConnectionId, ConnectionIdDecoder, ConnectionIdGenerator, ConnectionIdRef,
-    ConnectionParameters, EmptyConnectionIdGenerator, Error,
+    ConnectionParameters, EmptyConnectionIdGenerator, Error, MIN_INITIAL_PACKET_SIZE,
 };
 
 /// This should be a valid-seeming transport parameter.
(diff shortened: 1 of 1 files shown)


larseggert:
In a related QNS test, we have the problem that the client does **not** issue a new connection ID when the server retires the path, and then we time out because the client wont accept any more packets from the server. I was hoping to reproduce this with this test, but it's behaving differently. Sigh.

neqo-transport/src/cid.rs
```diff
@@ -405,7 +405,7 @@ impl ConnectionIdStore<[u8; 16]> {
     pub fn retire_prior_to(&mut self, retire_prior: u64) -> Vec<u64> {
         let mut retired = Vec::new();
         self.cids.retain(|e| {
```
mxinden, line 407: Nit: could as well use `SmallVec::drain_filter`. Feel free to ignore.

https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html#method.drain_filter

neqo-transport/src/cid.rs
```diff
@@ -510,8 +510,18 @@ impl ConnectionIdManager {
...
         // TODO(mt) - consider keeping connection IDs around for a short while.
 
-        self.connection_ids.retire(seqno);
-        self.lost_new_connection_id.retain(|cid| cid.seqno != seqno);
+        let zero_len_cid = seqno == CONNECTION_ID_SEQNO_EMPTY
+            || self
+                .connection_ids
+                .cids
+                .iter()
+                .any(|c| c.seqno == seqno && c.cid.len() == 0);
```
mxinden, line 518: This pull request adds to checks for zero-len-cid, here and in `ConnectionIdStore::retire_prior_to`. How about making this a method on `ConnectionIdEntry` instead, which is then called here and in `ConnectionIdStore::retire_prior_to`?

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -62,27 +64,223 @@ const fn new_port(a: SocketAddr) -> SocketAddr {
...
+    }
+}
 
-    server.process_input(&dgram, now());
-    // Have the server send something so that it generates a packet.
-    let stream_id = server.stream_create(StreamType::UniDi).unwrap();
-    server.stream_close_send(stream_id).unwrap();
+const fn id(a: SocketAddr) -> SocketAddr {
+    a
+}
```
mxinden, line 108: Where is this used? Could you use the `std::convert::identity` function instead?

https://doc.rust-lang.org/std/convert/fn.identity.html

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -62,27 +64,223 @@ const fn new_port(a: SocketAddr) -> SocketAddr {
...
+        .primary()
+        .unwrap()
+        .borrow()
+        .local_cid()
+        .unwrap()
+        .len()
+        == 0;
+    let mut now = now();
+    let mut total_delay = Duration::new(0, 0);
+    qdebug!("XXXXXXXXX");
```
mxinden, line 194:
```suggestion
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1532,7 +1532,7 @@ impl Connection {
     /// This takes two times: when the datagram was received, and the current time.
     fn input(&mut self, d: &Datagram, received: Instant, now: Instant) {
         // First determine the path.
-        let path = self.paths.find_path_with_rebinding(
+        let path = self.paths.find_path(
```
mxinden, line 1535: For others, relevant discussion happening on https://github.com/mozilla/neqo/pull/2176#pullrequestreview-2378912679.

mxinden (commented):
Thank you for expanding the test matrix!

neqo-transport/src/cid.rs
larseggert: That seems to be behind a feature and is unstable? https://docs.rs/smallvec/latest/smallvec/#drain_filter

Can obviously add that feature. Are there others we should add? `union`, `const_generics` and `const_new` seem useful?

neqo-transport/src/path.rs
```diff
@@ -389,7 +348,15 @@ impl Paths {
...
         self.paths.retain(|p| {
             let current = p.borrow().remote_cid.as_ref().unwrap().sequence_number();
-            if current < retire_prior {
+            if current < retire_prior
+                && p.borrow()
+                    .remote_cid
+                    .as_ref()
+                    .unwrap()
+                    .connection_id()
+                    .len()
```
mxinden, line 357: Are the two calls to `p.borrow()` necessary? Or would something along the lines of the below work as well?

``` rust
let current_remote_cid = p.borrow().remote_cid.as_ref.unwrap();
if current_remote_cid .sequence_number() < retire_prior && current_remote_cid.connection_id().len() != 0 {
```

neqo-transport/src/path.rs
```diff
@@ -152,13 +110,14 @@ impl Paths {
...
-        let seqno = retired
-            .borrow()
-            .remote_cid
-            .as_ref()
-            .unwrap()
-            .sequence_number();
-        to_retire.push(seqno);
+        if let Some(cid) = &retired.borrow().remote_cid {
+            let seqno = cid.sequence_number();
+            if cid.connection_id().len() == 0 {
```
mxinden, line 115:
```suggestion
            if cid.connection_id().is_empty() {
```

Would the new `ConnectionId::is_empty()` function work here as well?

neqo-transport/src/cid.rs
```diff
@@ -314,6 +314,10 @@ impl ConnectionIdEntry<[u8; 16]> {
         stats.new_connection_id += 1;
         true
     }
+
+    pub fn is_empty(&self) -> bool {
```
mxinden, line 318:
```suggestion
    pub fn is_zero_len(&self) -> bool {
```

RFC 9000 refers to this as "zero-length Connection ID", right? How about adopting that wording here?

mxinden (commented):
A couple of clarifying questions. Sorry in case I am missing something obvious. None of them are blocking.

Other than the discussion in https://github.com/mozilla/neqo/pull/2176#pullrequestreview-2378912679, this looks good to me.

neqo-transport/src/cid.rs
```diff
@@ -510,8 +514,18 @@ impl ConnectionIdManager {
...
         // TODO(mt) - consider keeping connection IDs around for a short while.
 
-        self.connection_ids.retire(seqno);
-        self.lost_new_connection_id.retain(|cid| cid.seqno != seqno);
+        let empty_cid = seqno == CONNECTION_ID_SEQNO_EMPTY
+            || self
+                .connection_ids
+                .cids
+                .iter()
+                .any(|c| c.seqno == seqno && c.cid.is_empty());
```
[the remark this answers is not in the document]
larseggert: Yes. (I'm not in love with the `CONNECTION_ID_SEQNO_EMPTY` hack, esp. since we only seem to use the magic value on one end of the connection.)

neqo-transport/src/cid.rs
larseggert: We seem to calling these "empty", e.g., `CONNECTION_ID_SEQNO_EMPTY` so I stuck with that. (You're right that that's different from the RFC.)

neqo-transport/src/cid.rs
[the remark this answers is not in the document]
mxinden: Given that `self.connection_ids` is likely very small, is the optimization needed? In other words, do you expect a performance impact when accepting the suggestion above?

mxinden (approved):

neqo-transport/src/cid.rs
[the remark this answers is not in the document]
larseggert: Not sure. I mostly kept it in so `CONNECTION_ID_SEQNO_EMPTY` is more visible :-)

PR #2171, opened 2024-10-14
feat: Option for no zero-len CIDs for client

larseggert:


martinthomson:
Can this be conditional behaviour?  That is, can we add a flag to enable/disable, defaulting to zero length, but turning it on for the migration test?

larseggert:
I can make this a flag, good suggestion.

neqo-bin/src/client/mod.rs
```diff
@@ -245,13 +251,13 @@ impl Args {
...
-            | "ecn"
-            | "rebind-port"
-            | "rebind-addr"
-            | "connectionmigration" => {}
+            "handshake" | "transfer" | "retry" | "ecn" => {}
+            "rebind-port" | "rebind-addr" | "connectionmigration" => {
+                if self.cid_len == 0 {
+                    qinfo!("Rebind/migration test won't work with len-0 CID; overwriting to 8");
+                    self.cid_len = 8;
+                }
```
martinthomson, line 259: We should be able to do client-side address rebinding if the client uses a zero-length CID.  Which side is doing the rebinding again?

larseggert: I need to check tomorrow, but I think the issue is that we don't issue a new connection ID on the client when the server retires a path and the connection ID used for it.

larseggert: Yes, what's happening is that with a client using a len-0 CID, when the server has validated the new path after a rebinding event, it will retire the old path and hence send a `RETIRE_CONNECTION_ID` for CID number 0. After the client receives and processes that frame, it will no longer accept any packets from the server (`Dropped received packet: Invalid DCID CID [0]: ;`).

larseggert: See #2182.

larseggert:
It looks like we don't need this PR, but the QNS `connectionmigration` test is failing on #2170 due to https://gitlab.com/wireshark/wireshark/-/issues/20165. Waiting for this to resolve.

larseggert:
We don't need this for migration. Left in the option, but not turning it on by default for certain tests.

PR #2176, opened 2024-10-14
fix: Replace `find_path_with_rebinding` with `find_path`

larseggert:
We need to do a path challenge even if only the remote port changes.

This makes us pass the `rebind-port` QNS test.

### DIFF
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -83,7 +83,7 @@ impl Paths {
         self.paths
             .iter()
             .find_map(|p| {
-                if p.borrow().received_on(local, remote, false) {
+                if p.borrow().received_on(local, remote) {
                     Some(Rc::clone(p))
                 } else {
                     None
@@ -580,12 +580,8 @@ impl Path {
     }
 
     /// Determine if this path was the one that the provided datagram was received on.
-    /// This uses the full local socket address, but ignores the port number on the peer
-    /// if `flexible` is true, allowing for NAT rebinding that retains the same IP.
-    fn received_on(&self, local: SocketAddr, remote: SocketAddr, flexible: bool) -> bool {
-        self.local == local
-            && self.remote.ip() == remote.ip()
-            && (flexible || self.remote.port() == remote.port())
+    fn received_on(&self, local: SocketAddr, remote: SocketAddr) -> bool {
+        self.local == local && self.remote == remote
     }
 
     /// Update the remote port number.  Any flexibility we allow in `received_on`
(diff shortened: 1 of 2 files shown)


neqo-transport/src/path.rs
```diff
@@ -98,48 +98,6 @@ impl Paths {
...
-            .find_map(|p| {
-                if p.borrow().received_on(local, remote, false) {
-                    Some(Rc::clone(p))
-                } else {
-                    None
-                }
-            })
-            .or_else(|| {
-                self.paths.iter().find_map(|p| {
-                    if p.borrow().received_on(local, remote, true) {
```
mxinden, line 124: `find_path_with_rebinding` is the only function calling `received_on` with the `flexible` parameter (i.e. last one) set to `true`. With `find_path_with_rebinding` gone, do we still need `flexible` on `received_on`?

neqo-transport/src/path.rs
```diff
@@ -98,48 +98,6 @@ impl Paths {
             })
     }
 
-    /// Find the path, but allow for rebinding.  That matches the pair of addresses
-    /// to paths that match the remote address only based on IP addres, not port.
-    /// We use this when the other side migrates to skip address validation and
-    /// creating a new path.
```
mxinden, line 104: Why did we previously skip address validation on port change?

If I understand RFC 9000 correctly, the port is always part of the validation process:

> In path validation, endpoints test reachability between a specific local address and a specific peer address, where an address is the 2-tuple of IP address and port.

https://www.rfc-editor.org/rfc/rfc9000.html#section-8.2

neqo-transport/src/path.rs
larseggert: Good catch.

neqo-transport/src/path.rs
larseggert: @martinthomson might know?

martinthomson: We did that deliberately, because if a peer changes just the port, it's probably just a NAT rebinding.  No point in wasting the effort.

martinthomson (commented):
If this results in full address validation for a port change, that's probably not OK.  I mean, we don't really implement the server end, but it's a regression in capabilities for the server.

neqo-transport/src/path.rs
larseggert: Do we care that that will make us violate the spec and fail a QNS test?

larseggert:
Now part of #2170.

PR #2180, opened 2024-10-15
feat: Support QNS `connectionmigration` test

larseggert:


neqo-bin/src/client/mod.rs
```diff
@@ -493,6 +493,14 @@ fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<NeqoQlog> {
     .map_err(Error::QlogError)
 }
 
+const fn unspecified_addr(addr: &SocketAddr) -> SocketAddr {
```
martinthomson, line 496: It seems like you could include the port number in the signature.

I'm also unsure about the name.  This might be `local_addr_for(remote_addr, local_port)`.  The usage below can use `local_port: 0`.

neqo-bin/src/client/mod.rs
```diff
@@ -554,6 +557,18 @@ pub async fn client(mut args: Args) -> Res<()> {
             remote_addr,
         );
 
+        let migration = if args.shared.qns_test == Some("connectionmigration".to_owned()) {
+            remote_addrs.next().map_or_else(
+                || {
+                    qerror!("No migration address found for {host}");
+                    exit(127);
+                },
```
martinthomson, line 565: Clippy suggested map_or_else, didn't it?  It's unfortunate, because this is probably better as:

```rust
if Some(addr) = remote_addrs.next() {
    Some((real_local.port(), addr))
} else {
    qerror!("Cannot migrate from {host} when there is no address that follows");
    exit(127);
}
```

neqo-bin/src/client/mod.rs
```diff
@@ -554,6 +557,18 @@ pub async fn client(mut args: Args) -> Res<()> {
             remote_addr,
         );
 
+        let migration = if args.shared.qns_test == Some("connectionmigration".to_owned()) {
```
martinthomson, line 560: Why is this not possible?

```suggestion
        let migration = if args.shared.qns_test == Some("connectionmigration") {
```

I know that `qns_test` is a `String`, but you should be able to `PartialEq` it with `&'static str` perfectly well.

neqo-bin/src/client/http09.rs
```diff
@@ -86,10 +87,33 @@ impl super::Handler for Handler<'_> {
...
+                                    Instant::now(),
+                                )
+                                .map(|()| {
+                                    qinfo!(
+                                        "Connection migrated to {:?} -> {:?}",
+                                        local_addr,
+                                        migration_addr
+                                    );
+                                    // Don't do another migration.
+                                    self.migration = None;
```
martinthomson, line 113: So you are only doing this change if the migration works.  What errors are you trying to ignore?  Would this be easier if the migration attempt only occurred for `State::Confirmed`?

I was going to suggest that you do `self.migration.take()` above instead, before I realized that migration could be attempted multiple times.

neqo-bin/src/client/mod.rs
```diff
@@ -493,6 +493,14 @@ fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<NeqoQlog> {
...
 }
 
+const fn unspecified_addr(addr: &SocketAddr) -> SocketAddr {
+    match addr {
+        SocketAddr::V4(..) => SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
+        SocketAddr::V6(..) => SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
+    }
+}
+
+#[allow(clippy::too_many_lines)]
```
martinthomson, line 503: This is usually a good sign that some work is needed.  Offhand, I see a nice candidate on line 517 (to 532).

neqo-bin/src/client/http09.rs
```diff
@@ -86,10 +87,33 @@ impl super::Handler for Handler<'_> {
                     }
                 }
                 ConnectionEvent::StateChange(
-                    State::WaitInitial | State::Handshaking | State::Connected,
+                    State::WaitInitial | State::Handshaking | State::Connected | State::Confirmed,
                 ) => {
                     qdebug!("{event:?}");
                     self.download_urls(client);
```
mxinden, line 93: Why do we need to execute `self.download_urls` on `State::Confirmed`? In other words, why is the migration logic below not in its own `ConnectionEvent::StateChange(State::Confirmed)` match arm?

martinthomson: I had a [similar comment](https://github.com/mozilla/neqo/pull/2180#discussion_r1802166587).

larseggert: That makes more sense, thanks.

neqo-bin/src/client/mod.rs
martinthomson: Maybe then...

```suggestion
        let migration = if args.shared.qns_test.as_deref() == Some("connectionmigration") {
```

neqo-bin/src/client/http09.rs
martinthomson: You shouldn't need to call download_urls() when you hit Confirmed.  You always hit Connected first.

neqo-bin/src/client/http09.rs
```diff
@@ -85,6 +86,28 @@ impl super::Handler for Handler<'_> {
                         self.download_urls(client);
                     }
                 }
+                ConnectionEvent::StateChange(State::Confirmed) => {
+                    if let Some((local_port, migration_addr)) = &self.migration {
```
martinthomson, line 90: Is there any reason why we would want to keep trying?  It seems like you could use this with no loss of generality, because we only ever generate ONE State::Confirmed event for a connection.

```suggestion
                    if let Some((local_port, migration_addr)) = self.migration.take() {
```

That leads to a few minor changes below...

neqo-bin/src/client/http09.rs
```diff
@@ -85,6 +86,28 @@ impl super::Handler for Handler<'_> {
...
+                        let local_addr = local_addr_for(migration_addr, *local_port);
```
martinthomson, line 91:
```suggestion
                        let local_addr = local_addr_for(migration_addr, local_port);
```

neqo-bin/src/client/http09.rs
```diff
@@ -85,6 +86,28 @@ impl super::Handler for Handler<'_> {
...
+                        qdebug!("Migrating path to {:?} -> {:?}", local_addr, migration_addr);
+                        client
+                            .migrate(
+                                Some(local_addr),
+                                Some(*migration_addr),
```
martinthomson, line 96:
```suggestion
                                Some(migration_addr),
```

neqo-bin/src/client/http09.rs
```diff
@@ -85,6 +86,28 @@ impl super::Handler for Handler<'_> {
...
+                                Instant::now(),
+                            )
+                            .map(|()| {
+                                qinfo!(
+                                    "Connection migrated to {:?} -> {:?}",
+                                    local_addr,
+                                    migration_addr
+                                );
+                                // Don't do another migration.
+                                self.migration = None;
```
martinthomson, line 107:
```suggestion
```

martinthomson (approved):

PR #2182, opened 2024-10-15
fix: Don't retire zero-length connection IDs

larseggert:
This makes us pass the `rebind-addr` QNS test.


martinthomson (commented):
wheee!

Test case?

larseggert:
Working on test cases in #2170. Should I merge all the feature fixes there?

martinthomson:
Moving the code seems fine to me.

larseggert:
Now part of #2170.

PR #2183, opened 2024-10-15
ci: Run benchmarks with Ethernet and max MTUs

larseggert:


neqo-transport/benches/transfer.rs
mxinden, line 1: Unrelated change? Feel free to leave it here, given that it is a benchmark only.

mxinden (approved):
apart from msquic failures, looks good to me.

neqo-transport/benches/transfer.rs
larseggert: Ah yes, that fixes a new clippy warning.

.github/workflows/bench.yml
```diff
@@ -132,50 +134,57 @@ jobs:
...
-                  taskset -c 1 nice -n -20 \
-                    perf $PERF_OPT -o "$client-$server$EXT.client.perf" \
-                      hyperfine -N --output null -w 1 -s "sleep 1" -n "$TAG" -u millisecond --export-markdown step.md "$CMD" |
-                      tee -a comparison.txt
-                  echo >> comparison.txt
-                  kill $PID
-                  cat step.md >> steps.md
-                  # Sanity check the size of the last retrieved file.
-                  [ "$(wc -c <"$SIZE")" -eq "$SIZE" ] || exit 1
+          # See https://github.com/microsoft/msquic/issues/4618#issuecomment-2422611592
```
martinthomson, line 137: Why is this giant thing not a shell script?

neqo-transport/benches/transfer.rs
```diff
@@ -63,14 +63,18 @@ fn benchmark_transfer(c: &mut Criterion, label: &str, seed: &Option<impl AsRef<s
...
+        std::env::var("SIMULATION_SEED").ok().as_ref(),
+    );
 }
 
 fn benchmark_transfer_fixed(c: &mut Criterion) {
     benchmark_transfer(
         c,
         "same-seed",
-        &Some("62df6933ba1f543cece01db8f27fb2025529b27f93df39e19f006e1db3b8c843"),
+        Some(&"62df6933ba1f543cece01db8f27fb2025529b27f93df39e19f006e1db3b8c843"),
```
martinthomson, line 77: So you are passing `&&str` and changing the signature to `&impl AsRef`?  That seems a bit "wat" to me.

`impl AsRef<str>` matches `String` and `&str`.  Why does it need an extra `&`?  That only reduces ergonomics.

larseggert: Clippy wants it

PR #2184, opened 2024-10-17
feat: don't allocate in UDP recv path

mxinden:
This is the final stacked pull request combining the following commits:

**feat(common): make Datagram generic over payload type**

Previously the payload type of `Datagram` was `Vec<u8>`, thus the `Datagram` always owned the UDP datagram payload.

This change enables `Datagram` to represent both (a) an owned payload allocation (i.e. `Vec<u8>`), but also represent (b) a view into an existing payload (e.g. a long lived receive buffer via `&[u8]`).

The default payload type stays `Vec<u8>`, thus not breaking existing usage of `Datagram`.

**feat(transport): accept borrowed instead of owned input datagram**

Previously `process_input` (and the like) took a `Datagram<Vec<u8>>` as input. In other words, it required allocating each UDP datagram payload in a dedicated `Vec<u8>` before passing it to `process_input`.

With this patch, `process_input` accepts a `Datagram<&[u8]>`. In other words, it accepts a `Datagram` with a borrowed view into an existing buffer (`&[u8]`), e.g. a long lived receive buffer.

**feat(udp): return borrowed Datagram on receive**

Previously `recv_inner` would return `Datagram<Vec<u8>>`. In other words, it would allocate a new `Vec<u8>` for each UDP datagram payload.

Now `recv_inner` reads into a provided buffer and returns `Datagram<&[u8]>`, i.e. it returns a view into the provided buffer without allocating.

**feat(bin): don't allocate in UDP recv path**

Pass a long lived receive buffer to `neqo_udp::recv_inner`, receiving an iterator of `Datagram<&[u8]>`s pointing into that buffer, thus no longer allocating in UDP receive path.

---

Extracted out of https://github.com/mozilla/neqo/pull/2093.

1.  (this one) https://github.com/mozilla/neqo/pull/2184
2. https://github.com/mozilla/neqo/pull/2187
4. https://github.com/mozilla/neqo/pull/2189

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -9,23 +9,14 @@ use std::{net::SocketAddr, ops::Deref};
 use crate::{hex_with_len, IpTos};
 
 #[derive(Clone, PartialEq, Eq)]
-pub struct Datagram {
+pub struct Datagram<D = Vec<u8>> {
     src: SocketAddr,
     dst: SocketAddr,
     tos: IpTos,
-    d: Vec<u8>,
+    d: D,
 }
 
-impl Datagram {
-    pub fn new<V: Into<Vec<u8>>>(src: SocketAddr, dst: SocketAddr, tos: IpTos, d: V) -> Self {
-        Self {
-            src,
-            dst,
-            tos,
-            d: d.into(),
-        }
-    }
-
+impl<D> Datagram<D> {
     #[must_use]
     pub const fn source(&self) -> SocketAddr {
         self.src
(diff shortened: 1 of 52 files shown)


mxinden:
This is the final pull request in the stack. All commits, but the first (https://github.com/mozilla/neqo/pull/2184/commits/08a0003b2a5deb3bdf4c787d874d27d8f9e79275), have been reviewed and merged through the 3 previous stacked pull requests.

@martinthomson and @larseggert can you do one more review?

neqo-common/src/datagram.rs
```diff
@@ -46,15 +37,43 @@ impl Datagram {
...
+    pub fn len(&self) -> usize {
+        self.d.as_ref().len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+// TODO: Needed?
```
martinthomson, line 50: It's harmless if not, though I think it might make some of our tests easier.

neqo-udp/src/lib.rs
```diff
@@ -57,63 +52,89 @@ use std::os::fd::AsFd as SocketRef;
...
+    })
+}
+
+pub struct DatagramIter<'a> {
+    meta: RecvMeta,
+    datagrams: Chunks<'a, u8>,
+    local_address: SocketAddr,
+}
+
+impl std::fmt::Debug for DatagramIter<'_> {
```
martinthomson, line 107: Code coverage suggests that this might be something worth testing.

martinthomson (approved):
LGTM.

neqo-common/src/datagram.rs
mxinden: Removed `TODO` and moved behing `#[cfg(test)]`. In case we ever need it in non-tests, we can remove the `cfg`.

neqo-udp/src/lib.rs
mxinden: :+1: Added a basic test.

larseggert:
@mxinden is this good to go? Nothing more pending?

mxinden:
Nothing else pending from my end @larseggert. Please merge in case you have no further comments.

I will propose a new release once merged. Excited to get this into Firefox.

mxinden:
Thank you for the help here!

For early CPU profiles of this patch in Firefox see https://github.com/mozilla/neqo/pull/2198#issue-2611368183

Issue #2185, opened 2024-10-17
ci: failed to run custom build command for `neqo-crypto v0.9.2


mxinden:
CI is currently failing with:

```
2024-10-17T12:43:37.9020958Z [1m[31merror[0m[1m:[0m failed to run custom build command for `neqo-crypto v0.9.2 (/home/runner/work/neqo/neqo/neqo-crypto)`
2024-10-17T12:43:37.9023657Z note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.
2024-10-17T12:43:37.9025414Z 
2024-10-17T12:43:37.9025675Z Caused by:
2024-10-17T12:43:37.9028539Z   process didn't exit successfully: `/home/runner/work/neqo/neqo/target/debug/build/neqo-crypto-9c0c0cd99d65ea5d/build-script-build` (exit status: 101)
2024-10-17T12:43:37.9030242Z   --- stdout
2024-10-17T12:43:37.9030776Z   cargo:rustc-check-cfg=cfg(nss_nodb)
2024-10-17T12:43:37.9031454Z   cargo:rerun-if-env-changed=NSS_DIR
2024-10-17T12:43:37.9032147Z   cargo:rerun-if-env-changed=NSS_TARGET
2024-10-17T12:43:37.9033040Z   cargo:include=/home/runner/work/neqo/neqo/dist/Release/include/nspr
2024-10-17T12:43:37.9033997Z   cargo:include=/home/runner/work/neqo/neqo/dist/public/nss
2024-10-17T12:43:37.9035137Z   cargo:rustc-link-search=native=/home/runner/work/neqo/neqo/dist/Release/lib
2024-10-17T12:43:37.9036165Z   cargo:rustc-link-lib=static=certdb
2024-10-17T12:43:37.9036816Z   cargo:rustc-link-lib=static=certhi
2024-10-17T12:43:37.9037494Z   cargo:rustc-link-lib=static=cryptohi
2024-10-17T12:43:37.9038174Z   cargo:rustc-link-lib=static=freebl
2024-10-17T12:43:37.9038862Z   cargo:rustc-link-lib=static=nss_static
2024-10-17T12:43:37.9039512Z   cargo:rustc-link-lib=static=nssb
2024-10-17T12:43:37.9040136Z   cargo:rustc-link-lib=static=nssdev
[...]

PR #2186, opened 2024-10-17
ci: fix cache-hit check

mxinden:
According to the actions/cache docs an empty string represents a cache miss:

> cache-hit - A string value to indicate an exact match was found for the key.
>  - If there's a cache hit, this will be 'true' or 'false' to indicate if there's an exact match for key.
>  - If there's a cache miss, this will be an empty string.

https://github.com/actions/cache?tab=readme-ov-file#outputs

Previously Neqo's CI would check for `"false"`, now it checks for `""`.

Fixes https://github.com/mozilla/neqo/issues/2185.

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -120,7 +120,7 @@ runs:
       if: env.USE_SYSTEM_NSS == '0'
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


mxinden:
See https://github.com/actions/cache/pull/1467 for more details on the regression.

larseggert (approved):

PR #2187, opened 2024-10-17
feat(transport): accept borrowed instead of owned input datagram

mxinden:
_[Stacked](https://github.com/marketplace/stacked-pull-requests) on top of https://github.com/mozilla/neqo/pull/2184._

Previously `process_input` (and the like) took a `Datagram<Vec<u8>>` as input. In other words, it required allocating each UDP datagram payload in a dedicated `Vec<u8>` before passing it to `process_input`.

With this patch, `process_input` accepts a `Datagram<&[u8]>`. In other words, it accepts a `Datagram` with a borrowed view into an existing buffer (`&[u8]`), e.g. a long lived receive buffer.

---

Extracted out of https://github.com/mozilla/neqo/pull/2093.

1. https://github.com/mozilla/neqo/pull/2184
2. (this one) https://github.com/mozilla/neqo/pull/2187
4. https://github.com/mozilla/neqo/pull/2189

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -374,9 +374,11 @@ enum CloseState {
 /// Network client, e.g. [`neqo_transport::Connection`] or [`neqo_http3::Http3Client`].
 trait Client {
     fn process_output(&mut self, now: Instant) -> Output;
-    fn process_multiple_input<'a, I>(&mut self, dgrams: I, now: Instant)
-    where
-        I: IntoIterator<Item = Datagram<&'a [u8]>>;
+    fn process_multiple_input<'a>(
+        &mut self,
+        dgrams: impl IntoIterator<Item = Datagram<&'a [u8]>>,
+        now: Instant,
+    );
     fn has_events(&self) -> bool;
     fn close<S>(&mut self, now: Instant, app_error: AppError, msg: S)
     where
@@ -392,9 +394,28 @@ struct Runner<'a, H: Handler> {
     handler: H,
     timeout: Option<Pin<Box<Sleep>>>,
     args: &'a Args,
+    recv_buf: Vec<u8>,
 }
 
 impl<'a, H: Handler> Runner<'a, H> {
+    fn new(
+        local_addr: SocketAddr,
+        socket: &'a mut crate::udp::Socket,
+        client: H::Client,
+        handler: H,
+        args: &'a Args,
+    ) -> Self {
+        Self {
+            local_addr,
+            socket,
+            client,
+            handler,
+            args,
+            timeout: None,
+            recv_buf: vec![0; neqo_udp::RECV_BUF_SIZE],
+        }
+    }
+
     async fn run(mut self) -> Res<Option<ResumptionToken>> {
         loop {
             let handler_done = self.handler.handle(&mut self.client)?;
@@ -457,12 +478,13 @@ impl<'a, H: Handler> Runner<'a, H> {
 
     async fn process_multiple_input(&mut self) -> Res<()> {
         loop {
-            let dgrams = self.socket.recv(&self.local_addr)?;
-            if dgrams.is_empty() {
+            let Some(dgrams) = self.socket.recv(&self.local_addr, &mut self.recv_buf)? else {
+                break;
+            };
+            if dgrams.len() == 0 {
                 break;
             }
-            self.client
-                .process_multiple_input(dgrams.iter().map(Datagram::borrow), Instant::now());
+            self.client.process_multiple_input(dgrams, Instant::now());
             self.process_output().await?;
         }
 
@@ -573,32 +595,18 @@ pub async fn client(mut args: Args) -> Res<()> {
 
                 let handler = http09::Handler::new(to_request, &args);
 
-                Runner {
-                    args: &args,
-                    client,
-                    handler,
-                    local_addr: real_local,
-                    socket: &mut socket,
-                    timeout: None,
-                }
-                .run()
-                .await?
+                Runner::new(real_local, &mut socket, client, handler, &args)
+                    .run()
+                    .await?
             } else {
                 let client = http3::create_client(&args, real_local, remote_addr, &hostname, token)
                     .expect("failed to create client");
 
                 let handler = http3::Handler::new(to_request, &args);
 
-                Runner {
-                    args: &args,
-                    client,
-                    handler,
-                    local_addr: real_local,
-                    socket: &mut socket,
-                    timeout: None,
-                }
-                .run()
-                .await?
+                Runner::new(real_local, &mut socket, client, handler, &args)
+                    .run()
+                    .await?
             };
         }
     }
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -376,7 +376,7 @@ trait Client {
     fn process_output(&mut self, now: Instant) -> Output;
     fn process_multiple_input<'a, I>(&mut self, dgrams: I, now: Instant)
     where
-        I: IntoIterator<Item = &'a Datagram>;
+        I: IntoIterator<Item = Datagram<&'a [u8]>>;
     fn has_events(&self) -> bool;
     fn close<S>(&mut self, now: Instant, app_error: AppError, msg: S)
     where
@@ -462,7 +462,7 @@ impl<'a, H: Handler> Runner<'a, H> {
                 break;
             }
             self.client
-                .process_multiple_input(dgrams.iter(), Instant::now());
+                .process_multiple_input(dgrams.iter().map(Datagram::borrow), Instant::now());
             self.process_output().await?;
         }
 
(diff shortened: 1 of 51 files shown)


neqo-bin/src/client/mod.rs
```diff
@@ -376,7 +376,7 @@ trait Client {
     fn process_output(&mut self, now: Instant) -> Output;
     fn process_multiple_input<'a, I>(&mut self, dgrams: I, now: Instant)
     where
-        I: IntoIterator<Item = &'a Datagram>;
+        I: IntoIterator<Item = Datagram<&'a [u8]>>;
```
martinthomson, line 379: Is it possible to make this `IntoIterator<Item = Datagram<impl AsRef<&'a [u8]>>>` instead?  (Or `Borrow`, as below.)   Then you could maybe drop the borrow calls in a few places. (I get that the multi interfaces are probably not that well suited to calling with `Vec<u8>`, but might as well be consistent throughout.

neqo-http3/src/connection_client.rs
```diff
@@ -873,13 +873,13 @@ impl Http3Client {
     /// packets need to be sent or if a timer needs to be updated.
     ///
     /// [1]: ../neqo_transport/enum.ConnectionEvent.html
-    pub fn process_input(&mut self, dgram: &Datagram, now: Instant) {
-        self.process_multiple_input(iter::once(dgram), now);
+    pub fn process_input<'a>(&mut self, dgram: impl Into<Datagram<&'a [u8]>>, now: Instant) {
```
martinthomson, line 876: Should this be `AsRef` rather than `Into`?  Or maybe `Borrow`.  We're not looking for a conversion.

neqo-transport/src/server.rs
```diff
@@ -430,7 +430,7 @@ impl Server {
         let mut callback = None;
 
         for connection in &mut self.connections {
-            match connection.borrow_mut().process(None, now) {
+            match connection.borrow_mut().process(None::<&Datagram>, now) {
```
martinthomson, line 433:
```suggestion
            match connection.borrow_mut().process_output(now) {
```

neqo-http3/src/connection_client.rs
mxinden: Unfortunately `Datagram<Vec<u8>>` can not implement `Borrow<Datagram<&[u8]>>` nor `AsRef<Datagram<&[u8]>`. 

See https://github.com/mozilla/neqo/pull/2093#discussion_r1788539953 for details and playground example.

Same for the opposite direction, i.e. `ToOwned`:

https://github.com/mozilla/neqo/pull/2093#discussion_r1788548422

neqo-bin/src/client/mod.rs
mxinden: > Is it possible to make this `IntoIterator<Item = Datagram<impl AsRef<&'a [u8]>>>` instead?

Interesting idea. It is possible. I sketched it in https://github.com/mxinden/neqo/commit/febd6f3082303530a562d3cfdddf1a0f48be6da1. 

That said, is this additional indirection worth the complexity it adds to the signature?

In my eyes, given that it is not strictly needed (e.g. no performance or functionality gain), and given that there are few `.borrow()` calls only, I don't think it is worth it.

> (Or `Borrow`, as below.)

Unfortunately, one can not use `Borrow`, given the returned reference to a temporary object. See comment below (https://github.com/mozilla/neqo/pull/2187#discussion_r1806044720).

neqo-http3/src/connection_client.rs
martinthomson:
```rust
impl Borrow<[u8]> for Datagram<Vec<u8>> {
    fn borrow(&self) -> &[u8] {
        &self.d
    }
}
impl Borrow<[u8]> for Datagram<&'_ [u8]> {
    fn borrow(&self) -> &[u8] {
        self.d
    }
}

#[allow(warnings)]
fn check(_d: Datagram<impl Borrow<[u8]>>) {}
```
Seems to compile just fine.

martinthomson: Or better...

```rust
impl<D: AsRef<[u8]>> Borrow<[u8]> for Datagram<D> {
    fn borrow(&self) -> &[u8] {
        self.d.as_ref()
    }
}
impl<D: AsMut<[u8]> + AsRef<[u8]>> BorrowMut<[u8]> for Datagram<D> {
    fn borrow_mut(&mut self) -> &mut [u8] {
        self.d.as_mut()
    }
}

impl Deref for Datagram
where
    Self: Borrow<[u8]>,
{
    type Target = [u8];
    #[must_use]
    fn deref(&self) -> &Self::Target {
        Borrow::<[u8]>::borrow(self)
    }
}

#[allow(warnings)]
fn test(
    _d: Datagram<impl Borrow<[u8]>>,
    _i: impl IntoIterator<Item = Datagram<impl Borrow<[u8]>>>,
) {
}
```

martinthomson: Though both violate the `Borrow` contract in that you aren't getting the whole thing, just a piece of it.  `AsRef` might be a better medium for this one...

```rust
impl<D: AsRef<[u8]>> AsRef<[u8]> for Datagram<D> {
    fn as_ref(&self) -> &[u8] {
        self.d.as_ref()
    }
}
impl<D: AsMut<[u8]> + AsRef<[u8]>> AsMut<[u8]> for Datagram<D> {
    fn as_mut(&mut self) -> &mut [u8] {
        self.d.as_mut()
    }
}

impl Deref for Datagram
where
    Self: AsRef<[u8]>,
{
    type Target = [u8];
    #[must_use]
    fn deref(&self) -> &Self::Target {
        AsRef::<[u8]>::as_ref(self)
    }
}

#[allow(warnings)]
fn test(_d: Datagram<impl AsRef<[u8]>>, _i: impl IntoIterator<Item = Datagram<impl AsRef<[u8]>>>) {}
```

mxinden:
Thanks for the input @martinthomson.

I applied your suggestion using `d: Datagram<impl AsRef<[u8]>>`.

Given that there is no more need to pass a `Datagram` by reference, as one would instead pass `Datagram<&[u8]>`,  this effectively reverts https://github.com/mozilla/neqo/pull/1543.

I am sorry for the large diff. @martinthomson let me know whether this is the direction you were suggesting above.

Commits are still split into changes to `neqo-common`, `neqo-transport` (and `neqo-http3`), `neqo-udp` and `neqo-bin`.

neqo-transport/src/connection/mod.rs
```diff
@@ -1287,17 +1288,19 @@ impl Connection {
...
+        d: &Datagram<impl AsRef<[u8]>>,
+        remaining: usize,
+        now: Instant,
+    ) {
+        let d = Datagram::new(
+            d.source(),
+            d.destination(),
+            d.tos(),
+            d[d.len() - remaining..].to_vec(),
+        );
```
mxinden, line 1303: Now always calls `to_vec`, even if the full datagram payload should be saved. `Datagram<impl AsRef<[u8]>>` can not differentiate in a `&[u8]` payload and a `Vec<u8>`, i.e. does not know when to allocate (`&[u8]`) and when to return itself (`Vec<u8>`).

Maybe the compiler optimizes this away.

I have to give this more thought.

martinthomson: If you remove the reference in the signature, that would help, I think.  Otherwise, the compiler won't be able to observe that the owned datagram is really owned.  The problem is that there is a slice involved, which might not easily map to just taking the owned datagram and transferring it (rather than allocating).  That is a common case, but maybe not something that the compiler can detect reliably.

I'm not concerned too much about the extra allocation either way. It only happens during the handshake and not that often. From memory, it's not exclusively the case that loss or reordering lead to the need to save datagrams, but those are the most obvious reasons for needing to save. Both are rare enough that we shouldn't worry about an allocation or two, especially when those allocations are pretty cheap.

In short, don't sweat it.  If it bothers you, open an issue so that we can track the problem and come back to it.

neqo-http3/src/connection_client.rs
mxinden: Thank you. Implemented with recent commits. See https://github.com/mozilla/neqo/pull/2187#issuecomment-2426040259.

neqo-bin/src/client/mod.rs
mxinden: Recent commits implement the above sketch in full. See also https://github.com/mozilla/neqo/pull/2187#issuecomment-2426040259.

martinthomson (approved):
This looks right.  There's a lot of noise, but that just makes me want to land it faster.

I'm not 100% happy with the size of Datagrams being paired with having them passed by value, but I think that is something the compiler will manage for us mostly.  (That is, SocketAddr is pretty big, and we have two of those, plus a byte, plus a slice.  Copying that isn't super expensive, but marking that Copy means that the compiler won't help us avoid work that isn't constructive.)  That said, I don't see any patterns here that are particularly problematic.

PR #2188, opened 2024-10-17
feat(udp): return borrowed Datagram on receive

mxinden:
_[Stacked](https://github.com/marketplace/stacked-pull-requests) on top of https://github.com/mozilla/neqo/pull/2187 which is stacked on top of https://github.com/mozilla/neqo/pull/2184._

Previously `recv_inner` would return `Datagram<Vec<u8>>`. In other words, it would allocate a new `Vec<u8>` for each UDP datagram payload.

Now `recv_inner` reads into a provided buffer and returns `Datagram<&[u8]>`, i.e. it returns a view into the provided buffer without allocating.

---

Extracted out of https://github.com/mozilla/neqo/pull/2093.

Part of https://github.com/mozilla/neqo/issues/1693.

[Pull request stack](https://github.com/marketplace/stacked-pull-requests):

1. https://github.com/mozilla/neqo/pull/2184
2. https://github.com/mozilla/neqo/pull/2187
3. (this one) https://github.com/mozilla/neqo/pull/2188
4. https://github.com/mozilla/neqo/pull/2189

PR #2189, opened 2024-10-17
feat(bin): don't allocate in UDP recv path

mxinden:
_Final pull request [stacked](https://github.com/marketplace/stacked-pull-requests) on top of https://github.com/mozilla/neqo/pull/2188 which is stacked on top of #2187 which is stacked on top of #2184._

Pass a long lived receive buffer to `neqo_udp::recv_inner`, receiving an iterator of `Datagram<&[u8]>`s pointing into that buffer, thus no longer allocating in UDP receive path.

---

Extracted out of #2093.
 
Part of #1693.

1. https://github.com/mozilla/neqo/pull/2184
2. https://github.com/mozilla/neqo/pull/2187
4. (this one) https://github.com/mozilla/neqo/pull/2189

### DIFF
--- a/neqo-bin/src/server/mod.rs
+++ b/neqo-bin/src/server/mod.rs
@@ -205,6 +205,7 @@ pub struct ServerRunner {
     server: Box<dyn HttpServer>,
     timeout: Option<Pin<Box<Sleep>>>,
     sockets: Vec<(SocketAddr, crate::udp::Socket)>,
+    recv_buf: Vec<u8>,
 }
 
 impl ServerRunner {
@@ -219,6 +220,7 @@ impl ServerRunner {
             server,
             timeout: None,
             sockets,
+            recv_buf: vec![0; neqo_udp::RECV_BUF_SIZE],
         }
     }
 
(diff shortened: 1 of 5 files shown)


neqo-bin/src/server/mod.rs
```diff
@@ -289,10 +291,13 @@ impl ServerRunner {
...
                     let (host, socket) = self.sockets.get_mut(inx).unwrap();
-                    let dgrams = socket.recv(host)?;
-                    if dgrams.is_empty() {
+                    let Some(dgrams) = socket.recv(host, &mut self.recv_buf)? else {
+                        break;
+                    };
+                    if dgrams.len() == 0 {
                         break;
                     }
+                    let dgrams: Vec<Datagram> = dgrams.map(|d| d.to_owned()).collect();
```
mxinden, line 300: Server allocates for now. In other words, only client UDP receive path doesn't allocate. Given that it is a test server only, I suggest we fix in a follow-up pull request.

martinthomson: SGTM, open an issue?

martinthomson (approved):

neqo-bin/src/server/mod.rs
larseggert: See #2190

Issue #2190, opened 2024-10-18
Make server UDP receive path not allocate


larseggert:
Server allocates for now. In other words, only client UDP receive path doesn't allocate. Given that it is a test server only, I suggest we fix in a follow-up pull request.

_Originally posted by @mxinden in https://github.com/mozilla/neqo/pull/2189#discussion_r1805021560_

PR #2199, opened 2024-10-25
bench(bin): add 100MB upload benchmark

mxinden:
Add 100 MiB upload benchmark to `neqo-bin`.

Might help answer questions like:

- Is Neqo upload as fast as download?
- Comparing Firefox and Chromium h3 upload speed, is Neqo the bottleneck, ...

PR #2202, opened 2024-10-26
feat(bin): don't allocate in server UDP recv path

mxinden:
Previously the `neqo-bin` server would read a set of datagrams from the socket and allocate them:

``` rust
let dgrams: Vec<Datagram> = dgrams.map(|d| d.to_owned()).collect();
```

This was done out of convenience, as handling `Datagram<&[u8]>`s, each borrowing from `self.recv_buf`, is hard to get right across multiple `&mut self` functions, that is here `self.run`, `self.process` and `self.find_socket`.

This commit combines `self.process` and `self.find_socket` and passes a socket index, instead of the read `Datagram`s from `self.run` to `self.process`, thus making the Rust borrow checker happy to handle borrowing `Datagram<&[u8]>`s
instead of owning `Datagram`s.

---

Follow-up to https://github.com/mozilla/neqo/pull/2184.
Fixes https://github.com/mozilla/neqo/issues/2190.
Hopefully speeds up https://github.com/mozilla/neqo/pull/2199.

### DIFF
--- a/neqo-bin/src/server/mod.rs
+++ b/neqo-bin/src/server/mod.rs
@@ -224,6 +224,84 @@ impl ServerRunner {
         }
     }
 
+    /// Tries to find a socket, but then just falls back to sending from the first.
+    fn find_socket(
+        sockets: &mut [(SocketAddr, crate::udp::Socket)],
+        addr: SocketAddr,
+    ) -> &mut crate::udp::Socket {
+        let ((_host, first_socket), rest) = sockets.split_first_mut().unwrap();
+        rest.iter_mut()
+            .map(|(_host, socket)| socket)
+            .find(|socket| {
+                socket
+                    .local_addr()
+                    .ok()
+                    .map_or(false, |socket_addr| socket_addr == addr)
+            })
+            .unwrap_or(first_socket)
+    }
+
+    async fn read_and_process(&mut self, sockets_index: usize) -> Result<(), io::Error> {
+        loop {
+            let (host, socket) = self.sockets.get_mut(sockets_index).unwrap();
+            let Some(input_dgrams) = socket.recv(*host, &mut self.recv_buf)? else {
+                break;
+            };
+
+            for input_dgram in input_dgrams {
+                Self::process_inner(
+                    &mut self.server,
+                    &mut self.timeout,
+                    &mut self.sockets,
+                    &self.now,
+                    Some(input_dgram),
+                )
+                .await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn process(&mut self) -> Result<(), io::Error> {
+        Self::process_inner(
+            &mut self.server,
+            &mut self.timeout,
+            &mut self.sockets,
+            &self.now,
+            None,
+        )
+        .await
+    }
+
+    // Free function (i.e. not taking `&mut self: ServerRunner`) to be callable by
+    // `ServerRunner::read_and_process` while holding a reference to
+    // `ServerRunner::recv_buf`.
+    async fn process_inner(
+        server: &mut Box<dyn HttpServer>,
+        timeout: &mut Option<Pin<Box<Sleep>>>,
+        sockets: &mut [(SocketAddr, crate::udp::Socket)],
+        now: &dyn Fn() -> Instant,
+        mut input_dgram: Option<Datagram<&[u8]>>,
+    ) -> Result<(), io::Error> {
+        loop {
+            match server.process(input_dgram.take(), now()) {
+                Output::Datagram(dgram) => {
+                    let socket = Self::find_socket(sockets, dgram.source());
+                    socket.writable().await?;
+                    socket.send(&dgram)?;
+                }
+                Output::Callback(new_timeout) => {
+                    qdebug!("Setting timeout of {:?}", new_timeout);
+                    *timeout = Some(Box::pin(tokio::time::sleep(new_timeout)));
+                    break;
+                }
+                Output::None => break,
+            }
+        }
+        Ok(())
+    }
+
     // Wait for any of the sockets to be readable or the timeout to fire.
     async fn ready(&mut self) -> Result<Ready, io::Error> {
         let sockets_ready = select_all(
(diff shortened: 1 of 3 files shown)


neqo-bin/src/server/mod.rs
```diff
@@ -224,36 +225,57 @@ impl ServerRunner {
...
+                // Then take the first datagram, if any.
+                input_dgrams.iter_mut().flatten().next().map_or_else(
+                    || {
+                        // Reading from the socket returned no datagrams. Don't try again.
+                        ready_socket_index = None;
+                        input_dgrams = None;
+                        None
+                    },
+                    Some,
+                )
```
martinthomson, line 249: I'm surprised that clippy didn't pick this one up.

```suggestion
                input_dgrams.iter_mut().flatten().next().or_else(
                    || {
                        // Reading from the socket returned no datagrams. Don't try again.
                        ready_socket_index = None;
                        input_dgrams = None;
                        None
                    },
                )
```

neqo-bin/src/server/mod.rs
```diff
@@ -224,36 +225,57 @@ impl ServerRunner {
...
+                    let socket = rest
+                        .iter_mut()
+                        .map(|(_host, socket)| socket)
+                        .find(|socket| {
+                            socket
+                                .local_addr()
+                                .ok()
+                                .map_or(false, |socket_addr| socket_addr == dgram.source())
+                        })
+                        .unwrap_or(first_socket);
```
martinthomson, line 265: You inlined this, presumably to avoid having to pass `&mut self` to it.  It's still a useful thing to have broken out.  You can make a new function that takes `&mut self.sockets` and returns `&mut Socket`.  Either that or you could go with `Self::send(&mut self.sockets, &dgram).await?` and have that function do the sending part as well.

neqo-bin/src/server/mod.rs
```diff
@@ -224,36 +225,57 @@ impl ServerRunner {
...
-            })
-            .unwrap_or(first_socket)
-    }
-
-    async fn process(&mut self, mut dgram: Option<Datagram>) -> Result<(), io::Error> {
+    async fn process(&mut self, mut ready_socket_index: Option<usize>) -> Result<(), io::Error> {
+        let mut input_dgrams: Option<DatagramIter> = None;
         loop {
-            match self.server.process(dgram.take(), (self.now)()) {
+            let input_dgram = if let Some(d) = input_dgrams.iter_mut().flatten().next() {
```
martinthomson, line 231: I found this code to be a little unintuitive.  You are taking a mutable iterator over the option, then flattening it.  It's not clear that you are mutating the underlying iterator as a result of calling `next()`.

Taking a step back, I think that this code is fairly simple:

You read from the indicated socket, process every datagram that it produces, and stop when the socket stops producing datagrams.

Would this structure work?

```rust
fn whatever(ready_socket_index: Option<usize>) -> Res<()> {
    let Some(inx) = ready_socket_index else {
        return Ok(());
    };

    let (host, socket) = self.sockets.get_mut(inx).unwrap();
    while let Some(input_dgrams) = socket.recv(*host, &mut self.recv_buf)? {
        for input in input_dgrams {
            match self.server.process(input, (self.now)()) {
                // see below for a note about sending.
                Output::Datagram(output) => Self::send(&mut self.sockets, &output).await?,
                Output::Callback(t) => {
                    self.timeout = Some(Box::pin(tokio::time::sleep(new_timeout)))
                }
                Output::None => {}
            }
        }
    }
    Ok(())
}
```

I think that's functionally equivalent to what you have, but a lot easier to read, at least for me.

I'm not a borrow checker, so I couldn't say if this works.  There are a lot of references being held here.  I can't see any obvious overlap.

mxinden: > I found this code to be a little unintuitive.

Agreed. The complexity stems from the fact that `neqo_transport::Server` does not have a `process_multiple_input` function, thus having to handle individual output `Datagram`s while still buffering input `Datagram`s.

`neqo_transport::Server` does not have a `process_multiple_input` function, because the set of input `Datagram`s provided might each be for a different `neqo_transport::Connection`, thus each result in a `Output::Datagram` and thus `process_multiple_input` would need to return a set of output `Datagram`s, not just one `Datagram`.

If you think it is worth it, I can explore this pathway, i.e. adding `process_multiple_input` to `neqo_transport::Server`. Preferably in a follow-up pull request.

---

Addressing the concrete suggestion above:

``` rust
    let Some(inx) = ready_socket_index else {
        return Ok(());
    };
```

`process` (or `whatever` above) might be called with `ready_socket_index` `None`, in which case it is expected to drive the output path only, i.e. not just `return` early.

```
    while let Some(input_dgrams) = socket.recv(*host, &mut self.recv_buf)? {
        for input in input_dgrams {
            match self.server.process(input, (self.now)()) {
                // see below for a note about sending.
                Output::Datagram(output) => Self::send(&mut self.sockets, &output).await?,
                Output::Callback(t) => {
                    self.timeout = Some(Box::pin(tokio::time::sleep(new_timeout)))
                }
                Output::None => {}
            }
        }
    }
```

If `self.server.process` returns `Output::Datagram`, one has to call it again, until it returns `Output::Callback` or `Output::None`. In the above suggestion, `self.server.process` is only called, if more input datagrams are available.

neqo-bin/src/server/mod.rs
mxinden: > I'm surprised that clippy didn't pick this one up.

And I am surprised that I didn't see this. :facepalm: Thank you Martin!

d23aa741

neqo-bin/src/server/mod.rs
mxinden: You are right. Keeping this logic separate is simpler. I re-introduced `find_socket` in 786c6166.

neqo-bin/src/server/mod.rs
```diff
@@ -224,36 +225,44 @@ impl ServerRunner {
...
-        rest.iter_mut()
-            .map(|(_host, socket)| socket)
-            .find(|socket| {
-                socket
-                    .local_addr()
-                    .ok()
-                    .map_or(false, |socket_addr| socket_addr == addr)
-            })
-            .unwrap_or(first_socket)
-    }
```
mxinden, line 239: Moved outside of `impl ServerRunner {}` as it no longer takes `&mut self` but instead `socket: &mut [(SocketAddr, crate::udp::Socket)]`.

neqo-bin/src/server/mod.rs
martinthomson: It seems like there are two functions here then:

1. When `ready_socket_index` is `None`, where you just call `process_output` and send out what comes.
2. When `ready_socket_index` is `Some`, where you do what I outlined.

That leads to:

```rust
    /// Returns `true` if it wants to be called again.
    fn drive_sending(&mut self, mut input: Option<Datagram>) {
        loop {
            match self.server.process(input.take(), (self.now)()) {
                // see below for a note about sending.
                Output::Datagram(output) => Self::send(&mut self.sockets, &output).await?,
                Output::Callback(t) => {
                    self.timeout = Some(Box::pin(tokio::time::sleep(new_timeout)));
                    return;
                }
                Output::None => return,
            }
            true
        }
    }

    let Some(inx) = ready_socket_index else {
        self.drive_sending(None);
        return Ok(());
    };

    while let Some(input_dgrams) = socket.recv(*host, &mut self.recv_buf)? {
        for input in input_dgrams {
            drive_sending(input);
        }
    }
```

mxinden: If I correctly understand the above, https://github.com/mozilla/neqo/pull/2202/commits/91c7920fd282dbcf12839f2d8b2c1051e6c78956 implements your suggestion. Please take a look @martinthomson.

> ``` rust
> fn drive_sending(&mut self, mut input: Option<Datagram>) {
> ```

Note that when calling `drive_sending` with an input datagram, one can not pass `&mut self` to `drive_sending`, as `self.recv_buf` is borrowed already.

neqo-bin/src/server/mod.rs
```diff
@@ -224,41 +224,6 @@ impl ServerRunner {
...
-                    self.timeout = Some(Box::pin(tokio::time::sleep(new_timeout)));
-                    break;
-                }
-                Output::None => {
-                    break;
-                }
-            }
-        }
-        Ok(())
-    }
```
mxinden, line 260: As with `find_socket` moved below into free function.

neqo-bin/src/server/mod.rs
martinthomson: You can move that back into `ServerRunner`.  Because it is dealing with internals of that struct, it might make sense to do that.  The only cost is that you invoke with `Self::process` rather than just `process`.

martinthomson (approved):
Yep, this looks much better and is more or less what I suggested.  Nice.  Consider the free function choice as I suggest though: I tend to prefer some amount of encapsulation for these sorts of functions, particularly when they are touching all of the internal pieces of the struct.

neqo-bin/src/server/mod.rs
mxinden: Sounds good. Done in https://github.com/mozilla/neqo/pull/2202/commits/1b7ce8fb01cf67b6a65e74d8e654bf5aea02a028.

mxinden:
@larseggert want me to hold off merging, or fine if I add it to the merge queue?

larseggert:
Feel free to merge. No reason to let it bitrot.

PR #2208, opened 2024-10-28
chore: check-in Cargo.lock

mxinden:
This commit checks the `Cargo.lock` file into git.

Version controlling `Cargo.lock` makes e.g. our CI builds more reproducible, where two consecutive CI runs on the same commit use the same set of dependencies, even if a compatible update of a dependency was published in between the two runs.

This is also helpful when cutting patch releases of old Neqo versions, where dependencies since shipped a breaking change in a patch version, e.g. a MSRV update. See for example pinned dependencies in a recent Neqo patch release to the Neqo v0.6 family.

https://github.com/mxinden/neqo/commit/66e60f360e9ce186210c50f00932872e2c2be550

While previously the recommendation by the cargo team was for libraries to not check in their `Cargo.lock`, this recommendation has since been replaced by "do what is best for the project".

https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

---

I generated a fresh `Cargo.lock`, i.e. ran `rm Cargo.lock && cargo check && git add Cargo.lock`.

We could as well base our `Cargo.lock` on mozilla-central's `Cargo.lock`. Though I can not think of a simple way to then keep the two in sync going forward.

Dependabot will attempt to keep the `Cargo.lock` up-to-date, in other words will open many pull requests updating patch versions. I don't know how to prevent this. Unfortunately, the [Dependabot `version-strategy` parameter](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) is not supported for the cargo ecosystem.

What do folks think?


larseggert:
Don't have a strong opinion here due to limited experience maintaining Rust libs. **Do** we think this is "best for the project"?

mxinden:
@martinthomson objections?

Would catch things like https://github.com/mozilla/neqo/pull/2219 early by making the `url`, `idna` and `arbitrary` version upgrades explicit, breaking the upgrade pull request only, not our entire CI.

mxinden:
Friendly ping @martinthomson.

In case there are no objections, I will update the pull request before merge.

martinthomson:
No objections, though dependabot spam seems like the real cost.  If it gets annoying, we can back it out.

mxinden:
> though dependabot spam seems like the real cost

I did some more digging. Setting the `open-pull-requests-limit` to `0` disables all updates **excluding** security updates. In my eyes, that is exactly what we want.

---

I updated the pull request:

- Setting the above `open-pull-requests-limit` to `0`, thus disabling all non-security updates.
- Synced the `Cargo.lock` with https://searchfox.org/mozilla-central/source/Cargo.lock.

larseggert (approved):

mxinden:
Once https://phabricator.services.mozilla.com/D229591 and https://phabricator.services.mozilla.com/D229592 landed, we can merge here.

mxinden:
Need to wait for https://bugzilla.mozilla.org/show_bug.cgi?id=1932635.

PR #2209, opened 2024-10-30
chore: update to quinn-udp v0.5.6

mxinden:
Currently we use `quinn-udp` `v0.5.4`.

`quinn-udp` `v0.5.5` fixes [`recvmmsg` calls on Android x86](https://github.com/quinn-rs/quinn/pull/1966).

`quinn-udp` `v0.5.6` adds [experimental multi-message support on Apple platforms](https://github.com/quinn-rs/quinn/pull/1993) and [fixes an unnecessary `windows-sys` version restriction](https://github.com/quinn-rs/quinn/pull/2021).

While not strictly necessary, given that our current version specification (i.e. `version = "0.5.4"`) already allows users to use Neqo with `quinn-udp` `v0.5.6`, this commit updates to `quinn-udp` `v0.5.6` anyways, thus making sure CI tests with latest version.

In case https://github.com/mozilla/neqo/pull/2208 lands, future compatible version updates would touch the `Cargo.lock` file, not `Cargo.toml`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ rust-version = "1.76.0"
 [workspace.dependencies]
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
-quinn-udp = { version = "0.5.4", default-features = false, features = ["direct-log"] }
+quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log"] }
 
 [workspace.lints.clippy]
 cargo = { level = "warn", priority = -1 }


Cargo.toml
```diff
@@ -31,7 +31,7 @@ rust-version = "1.76.0"
 [workspace.dependencies]
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
-quinn-udp = { version = "0.5.4", default-features = false, features = ["direct-log"] }
+quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log"] }
```
larseggert, line 34: Don't you need to add `fast-Apple-datapath` or whatever I called the feature?

mxinden: `neqo-udp` does not do GSO or `recvmmsg` today. Thus, even if I enable the `fast-apple-datapath` feature, none of the optimizations introduced in https://github.com/quinn-rs/quinn/pull/1993 would be used.

Adding `recvmmsg` shouldn't be difficult. That said, I would prefer doing so in a separate pull request.

larseggert (approved):

larseggert:
Merge queue CI might be failing because there is no 0.5.6? https://github.com/quinn-rs/quinn/releases

mxinden:
It is green now:

https://github.com/mozilla/neqo/actions/runs/11591731031/job/32320460781?pr=2209

@larseggert I assume you took some action to fix it?

I will attempt another merge through the queue.

> Merge queue CI might be failing because there is no 0.5.6? https://github.com/quinn-rs/quinn/releases (Ah, but there is a tag.)

I commented on https://github.com/quinn-rs/quinn/pull/2022. That said, agreed, a tag should suffice. If a tag isn't enough, we will have to adjust our tooling, as we can't require all upstream dependencies to maintain GitHub releases.

Issue #2211, opened 2024-10-31
test: use provided (simulated) time in qlogs


mxinden:
When logging qlog events, we currently use `qlog`'s [`add_event_data_now`](https://docs.rs/qlog/0.13.0/qlog/streamer/struct.QlogStreamer.html#method.add_event_data_now) which internally calls `std::time::Instant::now()`.

https://github.com/mozilla/neqo/blob/e08c14f5b26dfb9190caedf793ce59acf9634773/neqo-common/src/qlog.rs#L117

This is fine when capturing qlogs in real-time.

Problem arises when used with the [`text_fixtures::Simulator`](https://github.com/mozilla/neqo/blob/main/test-fixture/src/sim/mod.rs), which simulates time, i.e. where `std::time::Instant::now()` has no meaning.

I suggest we use `qlog`s [`add_event_data_with_instant`](https://docs.rs/qlog/0.13.0/qlog/streamer/struct.QlogStreamer.html#method.add_event_data_with_instant), using `now: Instant` available e.g. through `now` parameter on `Connection::process`.

https://github.com/mozilla/neqo/blob/e08c14f5b26dfb9190caedf793ce59acf9634773/neqo-transport/src/connection/mod.rs#L1165

Status: preparing a draft. Thus far, this is very noisy, given that `now` needs to be passed across many layers.

larseggert:
Sound good.

Related, it would be nice to see virtual time in the debug output for tests, and not clock time.

mxinden:
I sketched a fix for this issue in https://github.com/mozilla/neqo/pull/2212. Long story short, I don't think it is worth it. See PR description in https://github.com/mozilla/neqo/pull/2212. 

Let me know if you think otherwise, or have a better solution.

PR #2212, opened 2024-11-01
refactor: pass `now` to qlog

mxinden:
Instead of using `QLogStream::add_event_data_now`, which internally calls `std::time::Instant::now()`, pass `now` to
`QLogStream::add_event_data_with_instant`.

This **would** close https://github.com/mozilla/neqo/issues/2211. That said, given that this change is very noisy, passing `now` across all layers and adding `now` to many  of our public APIs (see `neqo-bin/src` for examples), **I don't think it is worth it**.

Let me know if you think otherwise.

Alternative: We could only do this change in `neqo-transport`, having all `neqo-http3` functions calling `neqo-transport` functions with a fresh `std::time::Instant::now()`. That would give us proper qlogs when using `test_fixtures::Simulator` as that is only concerned with `neqo-transport` and that would reduce the noise. Downside, additional complexity, as the problem itself is only half fixed, i.e. in `neqo-transport` but not `neqo-http3`.

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -53,7 +53,7 @@ impl Handler<'_> {
 impl super::Handler for Handler<'_> {
     type Client = Connection;
 
-    fn handle(&mut self, client: &mut Self::Client) -> Res<bool> {
+    fn handle(&mut self, client: &mut Self::Client, now: Instant) -> Res<bool> {
         while let Some(event) = client.next_event() {
             if self.needs_key_update {
                 match client.initiate_key_update() {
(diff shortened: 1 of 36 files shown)


larseggert:
Probably a stupid idea, but can we make `now` a thread-local atomic that is written to when time is updated via our public API and otherwise only read everywhere?

mxinden:
> Probably a stupid idea, but can we make `now` a thread-local atomic that is written to when time is updated via our public API and otherwise only read everywhere?

We could use a thread-local atomic.

That said, I think the current approach on `main`, carrying `now` explicitly from function to function, is better. In my eyes, the fact that Neqo does have little to no side-effects (e.g. not accessing "global" thread local variables) is a strong feature, enabling easy debugging. In addition, while Neqo is generally single threaded, even here, reasoning about thread local state adds a lot of complexity.

I am reluctant to give up the above properties for https://github.com/mozilla/neqo/issues/2211 alone. Instead of this pull request, or a thread local variable, I suggest we go with https://github.com/mozilla/neqo/pull/2216.

mxinden:
Closing here in favor of https://github.com/mozilla/neqo/pull/2216.

PR #2216, opened 2024-11-04
refactor(transport): reuse `now` in qlog wherever available

mxinden:
Instead of using `QLogStream::add_event_data_now`, which internally calls `std::time::Instant::now()`, pass `now` to
`QLogStream::add_event_data_with_instant` wherever `now` is available.

Less intrusive alternative to https://github.com/mozilla/neqo/pull/2212.

Fixes https://github.com/mozilla/neqo/issues/2211.


mxinden:
```
error[E0658]: use of unstable library feature 'error_in_core'
  --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/idna-1.0.3/src/lib.rs:78:6
   |
78 | impl core::error::Error for Errors {}
   |      ^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #103765 <https://github.com/rust-lang/rust/issues/103765> for more information

```
https://github.com/mozilla/neqo/actions/runs/11666011987/job/32480021975?pr=2216

I don't think this failure is related. Would be prevented by https://github.com/mozilla/neqo/pull/2208.

mxinden:
For the record, this pull request enables running the `test_fixtures::Simulator` with simulated time. Below, as a sample, a 100MB download on a 1 GBit/s 100ms simulated link.

image

neqo-transport/src/recovery/mod.rs
```diff
@@ -1526,15 +1562,18 @@ mod tests {
...
+            SentPacket::new(
+                PacketType::Initial,
+                0,
+                IpTosEcn::default(),
+                now(),
+                true,
+                Vec::new(),
+                ON_SENT_SIZE,
+            ),
+            Instant::now(),
```
larseggert, line 1575: Is there a way for `on_packet_sent` to take `now` out of the passed `SentPacket`, rather than passing it twice?

mxinden: We could, but would that be intuitive? The `SentPacket` might be sent before `now`. Given that it is an internal API and given that `now` is only an additional 4th argument, I would prefer not extracting current time from a related datastructure.

Do you feel strongly about it @larseggert?

larseggert: Don't feel strongly. Was just thinking about how to minimize the diff.

larseggert:
Can we add a comment to the remaining uses of `add_event_data_now` and explain why we didn't change them, and what the consequences are (= I assume incorrect timestamps during testing, if the test code hits those calls)?

neqo-common/Cargo.toml
```diff
@@ -29,6 +29,7 @@ windows = { version = "0.58", default-features = false, features = ["Win32_Media
 
 [dev-dependencies]
 test-fixture = { path = "../test-fixture" }
+regex = { version = "1.9", default-features = false, features = ["unicode-perl"] }
```
larseggert, line 32: If we need this, make it a workspace dep, since it's also used in `neqo-bin`?

neqo-common/src/qlog.rs
```diff
@@ -205,15 +221,14 @@ mod test {
...
         assert_eq!(
-            contents.to_string(),
-            format!(
-                "{EXPECTED_LOG_HEADER}{e}",
-                e = EXPECTED_LOG_EVENT.replace("\"time\":0.0,", "\"time\":1.1,")
-            )
+            Regex::new("\"time\":[0-9].[0-9]*,")
+                .unwrap()
+                .replace(&contents.to_string(), "\"time\":0.0,"),
+            format!("{EXPECTED_LOG_HEADER}{EXPECTED_LOG_EVENT}"),
```
larseggert, line 231: No way to avoid using `regex`?

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -112,13 +113,13 @@ fn issue_876() {
...
     for p in &sent_packets[..6] {
-        cc.on_packet_sent(p);
+        cc.on_packet_sent(p, time_now);
     }
     assert_eq!(cc.acked_bytes(), 0);
     cwnd_is_default(&cc);
     assert_eq!(cc.bytes_in_flight(), 6 * cc.max_datagram_size() - 3);
 
-    cc.on_packets_lost(Some(time_now), None, PTO, &sent_packets[0..1]);
+    cc.on_packets_lost(Some(time_now), None, PTO, &sent_packets[0..1], time_now);
```
larseggert, line 122: Might as well rename to `now`, to be in line with the rest of the code?

larseggert (approved):

neqo-common/src/qlog.rs
mxinden: Yes, but with additional complexity, basically reinventing the wheel.

Given that we already use `regex` elsewhere and given that this is a test, thus requiring `regex` as a dev-dependency only, I consider this the cleaner option.

Happy to implement an alternative if you feel strongly about it @larseggert.

mxinden:
> Can we add a comment to the remaining uses of `add_event_data_now` and explain why we didn't change them, and what the consequences are (= I assume incorrect timestamps during testing, if the test code hits those calls)?

Sorry, missed this earlier. Will address now.

PR #2217, opened 2024-11-05
feat: Enable `SSL_ENABLE_CH_EXTENSION_PERMUTATION`

larseggert:
This enables the NSS `SSL_ENABLE_CH_EXTENSION_PERMUTATION` option by default.

CC @martinthomson @dennisjackson


martinthomson (approved):
This looks good.  I don't think that we have to test this one.  Plumbing it in is sufficient.

PR #2218, opened 2024-11-05
ci: Benchmark against google-quiche

larseggert:


.github/workflows/bench.yml
```diff
@@ -132,15 +156,21 @@ jobs:
               CMD=${CMD//_pacing/--no-pacing}
               EXT="$EXT-nopacing"
             fi
+            CMD=${CMD//_flags/$flags}
           }
 
           # See https://github.com/microsoft/msquic/issues/4618#issuecomment-2422611592
-          for mtu in 1504 65536; do
+          for mtu in 1504; do # We used to measure 65536 here, too, but that's somewhat unrealistic.
```
mxinden, line 163: How about removing the loop, given that it iterates at most once. Would reduce the level of indentation.

mxinden (approved):
`neqo` to `neqo` with MTU of `1504` seems to have increased in this pull request. Is this noise or a regression?

Compare e.g. with https://github.com/mozilla/neqo/pull/2216 or https://github.com/mozilla/neqo/pull/2225.

Otherwise looks good to me. Ready to merge.

.github/workflows/bench.yml
larseggert: Can do. Was wondering if we wanted to put it back at some point.

larseggert:
> `neqo` to `neqo` with MTU of `1504` client/server transfer seems to have increased in this pull request. Is this noise or a regression?

I don't know.

.github/workflows/bench.yml
larseggert: Should I remove the 65536 MTU everywhere? Would cut down the time on benches.

mxinden: I assume we will focus on Internet paths for now, thus I suggest removing 64k MTU entirely.

mxinden:
> [ja4 workflow](https://github.com/mozilla/neqo/pull/2218/commits/648a380933ac83a48ba45ecf0ad581dc6e8b6eb4)

Bad push @larseggert?

PR #2219, opened 2024-11-05
deps: enable std feature of url crate

mxinden:
`url` `v0.5.3` and `idna` `v1.0.3` added no-std support: https://github.com/servo/rust-url/pull/831

Since Neqo sets `default-features = false`, the above would break Neqo.

Related: https://github.com/servo/rust-url/pull/831

Sample CI failure: https://github.com/mozilla/neqo/actions/runs/11683021003/job/32531372545


larseggert (approved):

mxinden:
`arbitrary` failure should be due to https://github.com/rust-fuzz/arbitrary/issues/202. Given that it is a transitive dependency, not sure how to set version in a non-hacky way. Give me a couple of minutes.

PR #2220, opened 2024-11-06
ci: Show QNS results relative to `main`

larseggert:


mxinden (approved):
Looks good to me. I assume this needs at least one run on `main` to function.

PR #2222, opened 2024-11-06
ci: Only build Firefox on manual trigger

larseggert:
Building Firefox in CI for every push is very slow and often fails when the glue hasn't been updated yet for a pending change.


mxinden (approved):
> and often fails when the glue hasn't been updated yet for a pending change.

:+1: 

In an ideal future, I would see us having a `neqo-next` branch on mozilla-central, but until then, I am in favor of disabling.

PR #2225, opened 2024-11-11
chore: Bump some dependency versions that changed upstream

larseggert:
And also convert some that are used in multiple crates to workspace dependencies.

### DIFF
--- a/neqo-transport/Cargo.toml
+++ b/neqo-transport/Cargo.toml
@@ -16,15 +16,15 @@ readme.workspace = true
 workspace = true
 
 [dependencies]
-# Sync with https://searchfox.org/mozilla-central/source/Cargo.lock 2024-02-08
-enum-map = { version = "2.7", default-features = false }
+# Checked against https://searchfox.org/mozilla-central/source/Cargo.lock 2024-11-11
+enum-map = { workspace = true }
 indexmap = { version = "2.2", default-features = false } # See https://github.com/mozilla/neqo/issues/1858
 log = { workspace = true }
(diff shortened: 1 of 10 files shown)


neqo-transport/Cargo.toml
```diff
@@ -16,15 +16,15 @@ readme.workspace = true
...
+# Checked against https://searchfox.org/mozilla-central/source/Cargo.lock 2024-11-11
+enum-map = { workspace = true }
 indexmap = { version = "2.2", default-features = false } # See https://github.com/mozilla/neqo/issues/1858
 log = { workspace = true }
 neqo-common = { path = "../neqo-common" }
 neqo-crypto = { path = "../neqo-crypto" }
 qlog = { workspace = true }
-smallvec = { version = "1.11", default-features = false }
-static_assertions = { version = "1.1", default-features = false }
+smallvec = { version = "1.13", default-features = false }
```
mxinden, line 26: Why this minor version bump? Is there a bug in `smallvec` `1.11` that is fixed in `1.13`. Unless there is a good reason for it, I don't think we should bump non-breaking versions, in this case the minor bump.

Don't feel strongly. In my eyes it adds noise. In addition, a consumer of `neqo-transport` might prefer using `neqo-transport` with `smallvec` `1.11`, e.g. as they have other dependencies using `smallvec` `1.11`.

mxinden (approved):

neqo-transport/Cargo.toml
larseggert: Gecko did the same bump

PR #2228, opened 2024-11-13
feat: Mix up the Initial crypto data a bit

larseggert:
Look for ranges of `N` or more bytes of graphical ASCII data in `data`. Create at least one split point for each range, multiple ones each `N` bytes if the range is long enough. Create data chunks based on those split points. Shuffle the chunks and return them.

CC @martinthomson @dennisjackson

### DIFF
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -16,12 +16,12 @@ use std::{
 
 use neqo_common::{hex, hex_snip_middle, qdebug, qinfo, qtrace, Encoder, Role};
 use neqo_crypto::{
-    hkdf, hp::HpKey, Aead, Agent, AntiReplay, Cipher, Epoch, Error as CryptoError, HandshakeState,
-    PrivateKey, PublicKey, Record, RecordList, ResumptionToken, SymKey, ZeroRttChecker,
-    TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, TLS_CT_HANDSHAKE,
-    TLS_EPOCH_APPLICATION_DATA, TLS_EPOCH_HANDSHAKE, TLS_EPOCH_INITIAL, TLS_EPOCH_ZERO_RTT,
-    TLS_GRP_EC_SECP256R1, TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1, TLS_GRP_EC_X25519,
-    TLS_GRP_KEM_MLKEM768X25519, TLS_VERSION_1_3,
+    hkdf, hp::HpKey, random, Aead, Agent, AntiReplay, Cipher, Epoch, Error as CryptoError,
+    HandshakeState, PrivateKey, PublicKey, Record, RecordList, ResumptionToken, SymKey,
+    ZeroRttChecker, TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256,
+    TLS_CT_HANDSHAKE, TLS_EPOCH_APPLICATION_DATA, TLS_EPOCH_HANDSHAKE, TLS_EPOCH_INITIAL,
+    TLS_EPOCH_ZERO_RTT, TLS_GRP_EC_SECP256R1, TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1,
+    TLS_GRP_EC_X25519, TLS_GRP_KEM_MLKEM768X25519, TLS_VERSION_1_3,
 };
 
 use crate::{
@@ -1358,6 +1358,68 @@ pub enum CryptoStreams {
     },
 }
 
+/// [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Modern_method)
+fn shuffle<T>(a: &mut [T]) {
+    // To shuffle an array a of n elements (indices 0..n-1)
+    let n: usize = a.len();
+    for i in 0..(n - 1) {
+        // j ← random integer such that i ≤ j < n
+        let j = (random::<8>()[0] as usize) % (n - i) + i;
+        // Exchange a[i] and a[j]
+        a.swap(i, j);
+    }
+}
+
+/// Find the start and end indices of all sequences of `n` or more graphic ASCII bytes in `data`.
+fn ascii_sequences(data: &[u8], len: usize) -> Vec<(usize, usize)> {
+    let mut sequences = vec![];
+    let mut start = None;
+    for (i, &b) in data.iter().enumerate() {
+        if b.is_ascii_graphic() {
+            if start.is_none() {
+                start = Some(i);
+            }
+        } else if let Some(s) = start {
+            if i - s >= len {
+                sequences.push((s, i));
+            }
+            start = None;
+        }
+    }
+    if let Some(s) = start {
+        if data.len() - s >= len {
+            sequences.push((s, data.len()));
+        }
+    }
+    sequences
+}
+
+/// Look for ranges of `N` or more bytes of graphical ASCII data in `data`. Create at least one
+/// split point for each range, multiple ones each `N` bytes if the range is long enough. Create
+/// data chunks based on those split points. Shuffle the chunks and return them.
+fn reorder_chunks(offset: usize, data: &[u8]) -> Vec<(u64, &[u8])> {
+    const N: usize = 3;
+    let seq = ascii_sequences(data, N);
+    // For each sequence, split it into chunks of three bytes.
+    let mut splits = vec![];
+    for (mut start, end) in seq {
+        while start + N <= end {
+            splits.push(start + N / 2);
+            start += N;
+        }
+    }
+    let mut chunks = vec![];
+    let mut start = offset;
+    for split in splits {
+        let chunk = &data[start..split];
+        chunks.push((start as u64, chunk));
+        start = split;
+    }
+    chunks.push((start as u64, &data[start..]));
+    shuffle(&mut chunks);
+    chunks
+}
+
 impl CryptoStreams {
     /// Keep around 64k if a server wants to push excess data at us.
     const BUFFER_LIMIT: u64 = 65536;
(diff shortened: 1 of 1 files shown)


neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
...
+fn shuffle<T>(a: &mut [T]) {
+    // To shuffle an array a of n elements (indices 0..n-1)
+    let n: usize = a.len();
+    if n < 2 {
+        return;
+    }
+    for i in 0..(n - 1) {
+        // j ← random integer such that i ≤ j < n
+        let j = loop {
+            let j = (random::<8>()[0] as usize) % (n - i) + i;
```
mxinden, line 1377: Is there a modulo bias here when `n - i` does not evenly divide a `256`?

larseggert: Maybe? Not sure if it matters in practice, since we just try to get some order out that isn't sequential.

martinthomson: Bias won't matter here.  This doesn't need to be cryptographically secure.

neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
...
+            if j != i {
+                break j;
```
martinthomson, line 1379: Why are you looping?

If you want a number in the range `(i+1..n)`, then you can sample modulo `n - i - 1` instead and add `i+1`.

neqo-transport/src/crypto.rs
martinthomson, line 1377: Calling `random::<8>` samples 8 bytes, but you are only using one.

This is going to more efficient if you sample once.  I suggest that - with the removal of the rejection sampling loop - you do this:

```rust
let mut randomness = [0; 32]; // set the value to what you prefer
let count = min(randomness.len(), n - 1);
randomize(&mut randomness[..count]);
for (i, r) in randomness.iter().enumerate().take(count) {
   let j = usize::from(r) % (n - i - 1) + i + 1;
   a.swap(i, j);
}
```

This will also limit the number of cuts that you swap over.  That makes it less random overall, but that's fine.

neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
...
+            }
+        };
+        // Exchange a[i] and a[j]
+        a.swap(i, j);
+    }
+}
+
+/// Find the start and end indices of all sequences of `len` or more graphic ASCII bytes in `data`.
+fn ascii_sequences(data: &[u8], len: usize) -> Vec<(usize, usize)> {
```
martinthomson, line 1388:
```suggestion
fn ascii_sequences(data: &[u8], len: usize) -> Vec<Range<usize>> {
```

Then you can use `start..end` for each value.

neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
...
+/// return them.
+fn reorder_chunks(data: &[u8]) -> Vec<(u64, &[u8])> {
+    const N: usize = 3;
+    let mut splits = vec![];
+    // For each sequence, split it into chunks of `N` bytes.
+    for (mut start, end) in ascii_sequences(data, N) {
+        while start + N <= end {
+            splits.push(start + N / 2);
+            start += N;
+        }
```
martinthomson, line 1423: Splitting a string multiple times is probably unnecessary.

This might be easier as:

```rust
let mut used = 0;
let ranges = ascii_sequences(data, N);
let mut chunks = Vec::with_capacity(ranges.len() + 1);
for (start, end) = ranges {
  debug_assert!(start > used);
  let (left, right) = data.split_at(start + N / 2 - used);
  chunks.push((u64::try_from(used).unwrap(), left));
  used += left;
  data = right;
}
shuffle(&mut chunks);
chunks
```

neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
     },
 }
 
+/// [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Modern_method)
```
martinthomson, line 1363: You've added enough stuff there that I think it might be worth creating a new module for the shuffling code.

neqo-transport/src/crypto.rs
```diff
@@ -1358,6 +1360,80 @@ pub enum CryptoStreams {
...
+/// Modified to make sure no element stays in place if `a` has more than one element.
+///
+/// TODO: We could use <https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.shuffle>
+/// instead, but I assume we want to only use the NSS RNG in neqo?
+fn shuffle<T>(a: &mut [T]) {
+    // To shuffle an array a of n elements (indices 0..n-1)
+    let n: usize = a.len();
+    if n < 2 {
```
martinthomson, line 1371: Given that each sampling draws from 0..256, it might be worth limiting the number of cuts we take.  I seriously doubt that we'll have to worry too much if we exceed 256 slices, but your current code has the potential to exceed that limit if a handshake message is >1k.  That seems like it might be a limit we hit often enough for this to be a problem.

I'm tempted to suggest that you increase the length of the LDH character span to compensate for that a little.  If you increase the number to 5, then you get to 1.5k.  That's still on the edge for me, but the chance that you'll get that in a typical ClientHello is small.

In addition, I think that you should limit the number of values that you return from `ascii_sequences()`.  For certificates, for example, the number of splits is going to be high just because the messages are long.  Limiting the number of cuts is going to work out fine, I think.

In terms of implementing that, I would suggest that rather than returning `Vec<(usize, usize)>` (or `Vec<Range<usize>>` as I suggest) from `ascii_sequences()` you instead return `impl Iterator<Item = Range<usize>>` and then you call `.take(256)` at the call site.

neqo-transport/src/crypto.rs
larseggert: This fails a bunch of tests. But also, why not split longer ASCII ranges more than once? See above for performance.

neqo-transport/src/crypto.rs
larseggert: > Calling random::<8> samples 8 bytes, but you are only using one.

Good catch. I meant to use them all.

neqo-transport/src/crypto.rs
larseggert: I fixed the `256` bug.

The remaining question is whether we want to limit the number of chunks in some way.

larseggert: @martinthomson AFAIK the only thing left is to limit the number of generated chunks. I'm wondering if we should look for the _n_ largest ASCII sequences (which _should_ include the SNI) and then use their midpoints as split points? Any suggestion for a good _n_?

martinthomson: Searching for the largest will involve unnecessarily expensive sorting (far better to parse the ClientHello in that case) or multiple passes.  I honestly think that n=32 for 4 or 5 byte sequences of LDH+. is more than enough.  (Run a test with ML-KEM enabled a few hundred times at 4 and if you hit the limit too often, go with 5.)

larseggert: Didn't see this until now. Let me know what you think of cb63418e54903517dbfb3b8c556cb70b21911099, which does what I outlined above. Happy to go a different way, if you prefer.

neqo-transport/src/shuffle.rs
```diff
@@ -0,0 +1,246 @@
...
+        let j = usize::from_ne_bytes(random::<USIZE>()) % (n - i - 1) + i + 1;
+        // Exchange a[i] and a[j]
+        debug_assert!(i != j);
+        a.swap(i, j);
+    }
+}
+
+/// Find the ranges of all sequences of two or more ASCII "LDHD" (letters, digits, hyphens, dots)
+/// bytes in `data`, and return the `n` longest ones in ascending order by start index.
+fn ascii_sequences(data: &[u8], n: usize) -> impl Iterator<Item = Range<usize>> {
```
martinthomson, line 35: I'm still not that enthusiastic about this approach, but here's a way to improve it some.

Change the signature to `fn ascii_sequences<const N: usize>(data: &[u8]) -> impl Interator...`.

Then collect the value into an array (`[_;  N]`) for sorting rather than a Vec.  You might have to unwrap, but that's not so bad.

martinthomson (commented):
I spent a few minutes writing some code that might be a good replacement for this if you only care about splitting at a useful point.  This would be used by drawing a random value in the range `(r.start + 1..end - 1)` to split on.  It's a lot less code...

```rust
/// Finds the range where the SNI extension lives.
/// If this isn't a ClientHello, then return the whole message.
fn find_sni_split(mut buf: &[u8]) -> Range<usize> {
    fn read_len(buf: &[u8]) -> usize {
        let mut len = 0;
        for v in buf {
            len = len << 8 & usize::from(v);
        }
        len
    }
    fn skip_vec<const N: usize>(i: usize, buf: &[u8]) -> usize {
        i + N + read_len(&buf[i..i + N])
    }

    // ClientHello == 1
    if buf[0] == 1 {
        let mut i = 1 + 3 + 2 + 32; // msg_type, length, version, random
        i = skip_vec::<1>(i, buf); // session_id
        i = skip_vec::<2>(i, buf); // cipher_suites
        i = skip_vec::<1>(i, buf); // compression_methods
        i += 2; // extensions length
        while i < buf.len() {
            if buf[i] == 0 || buf[i + 1] == 0 {
                // SNI!
                return i..i + 4 + read_len(buf[i + 2..i + 4]);
            }
        }
    }
    0..buf.len()
}
```

neqo-transport/src/shuffle.rs
larseggert: I did this in e9f3b58597d6dd202e7a7cb7ea7100f1ec51b3cf.

larseggert:
> I spent a few minutes writing some code that might be a good replacement for this if you only care about splitting at a useful point. 

Thanks for this, but I think it needs more work? I don't think this handles invalid crypto data well, `skip_vec` seems to easily read beyond the end of the buffer.

neqo-transport/src/shuffle.rs
larseggert: But I don't think I can collect into an array, because there might be fewer than `N` ranges. I *could* collect into an array with `Option<Range<usize>>` and have some of them be `None` in that case - is that worth it?

martinthomson:
Fair.  It was only a few minutes of work.  Adding bounds checks will complicate it a little, but not overmuch.  Of course, you could take the attitude I did, which is that the TLS stack should not be producing invalid messages, so panicking in that case is justified.

larseggert:
I am nervous about making assumptions about the input. But let me play with it and see how much complexity bounds checking adds.

larseggert:
@martinthomson is your code taking `SSL_ENABLE_CH_EXTENSION_PERMUTATION` into account? Because the range it returns isn't the SNI...

neqo-transport/src/shuffle.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    while i + 4 < buf.len() {
+        if buf[i] == 0 && buf[i + 1] == 0 {
+            // SNI!
+            i += 2;
+            let len = read_len(&buf[i..i + 2]);
+            if i + len > buf.len() {
+                break;
+            }
+            return i + 2..i + len;
```
martinthomson, line 76: You are choosing to split the name here.  I should then point out that the sni extension includes an inner TLV as well, so you should probably skip another three bytes more if you want to really split the name.

neqo-transport/src/shuffle.rs
```diff
@@ -0,0 +1,189 @@
...
+    if buf.len() < i || buf[0] != 1 {
+        return 0..buf.len();
+    }
+
+    // Skip session_id
+    i = if let Some(i) = skip_vec::<1>(i, buf) {
+        i
+    } else {
+        return 0..buf.len();
+    };
```
martinthomson, line 50: I wonder if we can't instead do something like this:

```rust
i += skip_vec::<1>(i, buf)?;
```
With the function returning `Option<Range<usize>>` to allow for that.  The caller would need to generate the default, but that seems like a good trade.

neqo-transport/src/crypto.rs
```diff
@@ -1490,38 +1493,60 @@ impl CryptoStreams {
...
+        }
 
-            // Don't bother if there isn't room for the header and some data.
-            if builder.remaining() < header_len + 1 {
-                return;
+        let cs: &mut CryptoStream = self.get_mut(space).unwrap();
+        if let Some((offset, data)) = cs.tx.next_bytes() {
+            let chunks = if shuffle {
+                // Mix up the crypto data a bit. Create five chunks that split ASCII LDHD sequences,
+                // and reorder them.
```
martinthomson, line 1532: Five is no longer the case

```suggestion
                // Cut the crypto data in two at random and swap the chunks.
```

neqo-transport/src/crypto.rs
```diff
@@ -1490,38 +1493,60 @@ impl CryptoStreams {
...
+                reorder_chunks(data)
+                    .into_iter()
+                    .map(|(off, d)| (offset + off, d))
+                    .collect()
```
martinthomson, line 1536: Do you need to do this now?  Or can you find a simpler way to manage this?

neqo-transport/src/connection/mod.rs
```diff
@@ -2309,7 +2310,13 @@ impl Connection {
                 self.write_appdata_frames(builder, &mut tokens);
             } else {
                 let stats = &mut self.stats.borrow_mut().frame_tx;
-                self.crypto.write_frame(space, builder, &mut tokens, stats);
+                self.crypto.write_frame(
+                    space,
+                    builder,
+                    space == PacketNumberSpace::Initial && self.role == Role::Client,
```
martinthomson, line 2316: I wonder whether we shouldn't just shuffle all the CRYPTO frames.

OR..  the thought just occurred  ... we now know if we have a ClientHello, so you could drop this entirely.  Have the split function return None if it isn't ClientHello and then not shuffle unless it is a ClientHello.  Then you don't need to pass down the parameter at all.

neqo-transport/src/shuffle.rs
larseggert: Is it common that there are multiple names in the extension?

neqo-transport/src/shuffle.rs
larseggert: I was thinking to check if we can use/extend `neqo-common/src/codec.rs` here.

neqo-transport/src/crypto.rs
```diff
@@ -1512,16 +1513,33 @@ impl CryptoStreams {
...
-            qdebug!("CRYPTO for {} offset={}, len={}", space, offset, length);
-            tokens.push(RecoveryToken::Crypto(CryptoRecoveryToken {
-                space,
-                offset,
-                length,
-            }));
-            stats.crypto += 1;
+        let cs = self.get_mut(space).unwrap();
+        if let Some((offset, data)) = cs.tx.next_bytes() {
+            let written = if let Some(sni) = find_sni(data) {
```
martinthomson, line 1521: I just realized that maybe you should check whether offset is zero.  Or maybe just comment that this will sometimes split something that isn't an SNI.

neqo-transport/src/crypto.rs
```diff
@@ -1493,13 +1494,13 @@ impl CryptoStreams {
         tokens: &mut Vec<RecoveryToken>,
         stats: &mut FrameStats,
     ) {
-        let cs = self.get_mut(space).unwrap();
-        if let Some((offset, data)) = cs.tx.next_bytes() {
+        fn write_chunk(chunk: (u64, &[u8]), builder: &mut PacketBuilder) -> Option<(u64, usize)> {
```
martinthomson, line 1497: I think that you want to make offset and data separate arguments to this.

The calls below all have separate values.

```suggestion
        fn write_chunk(offset: u64, data: &[u8], builder: &mut PacketBuilder) -> Option<(u64, usize)> {
```

(Eat that GitHub UI, I can type a suggestion even if you don't show the button!)

martinthomson (approved):

PR #2230, opened 2024-11-15
fix: Don't allow migration if the peer doesn't allow it

### LINKED ISSUE #2103: Make `Connection::migrate` fail when server doesn't support  active migration
One thing I noticed is that we have a public method called "migrate" on the connection.  This method can be called when the server has set `disable_active_migration`.  That's not ideal.  That's worth another change, I think.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2100#pullrequestreview-2301593264_

larseggert:
Fixes #2103

### DIFF
--- a/neqo-transport/src/connection/tests/migration.rs
+++ b/neqo-transport/src/connection/tests/migration.rs
@@ -976,8 +976,6 @@ fn migration_disabled() {
     let mut client = default_client();
     let mut server = new_server(ConnectionParameters::default().disable_migration(true));
     connect(&mut client, &mut server);
-    assert_eq!(client.state(), &State::Confirmed);
-    assert_eq!(server.state(), &State::Confirmed);
     assert_eq!(
         client
             .migrate(Some(DEFAULT_ADDR), Some(DEFAULT_ADDR), true, now())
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/tests/migration.rs
```diff
@@ -969,6 +971,21 @@ fn migration_invalid_state() {
         .is_err());
 }
 
+#[test]
+fn migration_disabled() {
+    let mut client = default_client();
+    let mut server = new_server(ConnectionParameters::default().disable_migration(true));
+    connect(&mut client, &mut server);
+    assert_eq!(client.state(), &State::Confirmed);
+    assert_eq!(server.state(), &State::Confirmed);
```
mxinden, line 980: Isn't this already checked by `connect` above?

https://github.com/mozilla/neqo/blob/a4692b57374b46969714269b1ae1d216903b507a/neqo-transport/src/connection/tests/mod.rs#L275-L276

mxinden (approved):

neqo-transport/src/connection/tests/migration.rs
larseggert: Possibly. Copied it from another test.

PR #2232, opened 2024-11-18
chore: Fix clippy `this map_or is redundant`

larseggert:


neqo-bin/src/server/mod.rs
```diff
@@ -229,12 +229,7 @@ impl ServerRunner {
         let ((_host, first_socket), rest) = sockets.split_first_mut().unwrap();
         rest.iter_mut()
             .map(|(_host, socket)| socket)
-            .find(|socket| {
-                socket
-                    .local_addr()
-                    .ok()
-                    .map_or(false, |socket_addr| socket_addr == addr)
-            })
+            .find(|socket| socket.local_addr().ok() == Some(addr))
```
martinthomson, line 232:
```suggestion
            .find(|socket| socket.local_addr().is_ok_and(|a| a == addr))
```

I *think* that's what they intended for us to use here.  Maybe better than `socket.local_addr() == Ok(addr)`.

martinthomson (approved):

PR #2237, opened 2024-11-18
Use try_send, so the caller can handle the IO error

### LINKED ISSUE #2236: `send()` returns `Ok(())` on IO errors
https://github.com/quinn-rs/quinn/blob/3a9d176a7a131a1f6d9472c1a23fccdcb1275b52/quinn-udp/src/unix.rs#L215

If I understand the code correctly, `quinn-udp` returns `Ok(())` for any IO error except `WouldBlock`. This behavior causes issues within Firefox, as certain parts of the code rely on the underlying error codes (e.g., `NS_ERROR_CONNECTION_REFUSED`) to handle specific scenarios appropriately.

KershawChang:
Fix #2236

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -35,7 +35,7 @@ pub fn send_inner(
         src_ip: None,
     };
 
-    state.send(socket, &transmit)?;
+    state.try_send(socket, &transmit)?;
 
     qtrace!(
         "sent {} bytes from {} to {}",
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -256,7 +256,7 @@ mod tests {
             segment_size: Some(SEGMENT_SIZE),
             src_ip: None,
         };
-        sender.state.send((&sender.inner).into(), &transmit)?;
+        sender.state.try_send((&sender.inner).into(), &transmit)?;
 
         // Allow for one GSO sendmmsg to result in multiple GRO recvmmsg.
         let mut num_received = 0;


mxinden (approved):
Thank you for the catch and patch!

For completeness sake, `try_send` API was introduced recently in https://github.com/quinn-rs/quinn/pull/2017.

Also for the record, corresponding NSPR implementation:

https://searchfox.org/mozilla-central/rev/b477cd37e845005dac8881427fa06a2771d993db/netwerk/base/nsUDPSocket.cpp#1180-1185

larseggert (approved):
I think some tests also use `send`; should we change those as well?

mxinden:
> I think some tests also use `send`; should we change those as well?

Sounds good. @KershawChang would you mind changing the line below to use `try_send` as well?

https://github.com/mozilla/neqo/blob/7a25525b3d368d99f19a1aa1829a5fa21c019ad9/neqo-udp/src/lib.rs#L259

mxinden:
@KershawChang we don't currently propagate the concrete IO error, but instead a generic error (i.e. `NS_ERROR_FAILURE`) only.

https://searchfox.org/mozilla-central/rev/b477cd37e845005dac8881427fa06a2771d993db/netwerk/socket/neqo_glue/src/lib.rs#764-777

Should we convert the concrete IO errors into their corresponding `NS_*` errors in `neqo_glue/src/lib.rs`? I am happy to provide a patch.

KershawChang:
> > I think some tests also use `send`; should we change those as well?
> 
> Sounds good. @KershawChang would you mind changing the line below to use `try_send` as well?
> 
> https://github.com/mozilla/neqo/blob/7a25525b3d368d99f19a1aa1829a5fa21c019ad9/neqo-udp/src/lib.rs#L259

Will do.

mxinden:
> > @KershawChang we don't currently propagate the concrete IO error, but instead a generic error (i.e. `NS_ERROR_FAILURE`) only.
> > https://searchfox.org/mozilla-central/rev/b477cd37e845005dac8881427fa06a2771d993db/netwerk/socket/neqo_glue/src/lib.rs#764-777
> > Should we convert the concrete IO errors into their corresponding `NS_*` errors in `neqo_glue/src/lib.rs`? I am happy to provide a patch.
> 
> Yes, I think we should. Once this PR merged and a new neqo released, we should prepare a patch for `NS_*` errors. Thanks.

:+1: I will prepare a new Neqo release and look into the relevant changes in mozilla-central to map the errors.

PR #2241, opened 2024-11-20
ci: Try out `super-linter` action

larseggert:
Pretty cool, but would require many changes to get it green. Unsure whether it's worth it.


larseggert:
Not useful without significant configuration and busywork.

Issue #2242, opened 2024-11-21
Optimize `SentPackets::take_ranges`


mxinden:
When CPU profiling a `test_fixtures::Simulator` run transferring 10 MiB from a server to a client, I see the following flamegraph:

 
![flamegraph2](https://github.com/user-attachments/assets/940a7bf3-9ddd-4318-bfe1-92a9f0b417d9)

The majority of CPU time is spent in `SentPackets::take_ranges`:

https://github.com/mozilla/neqo/blob/f3d01912b9a2d5962fffae1a192a3f62e738f745/neqo-transport/src/recovery/sent.rs#L189-L213

More specifically in the two calls to `BTreeMap::extend`:

https://github.com/mozilla/neqo/blob/f3d01912b9a2d5962fffae1a192a3f62e738f745/neqo-transport/src/recovery/sent.rs#L200-L211

Adding some logging, the following seems to be the case:

- Assume we have two nodes, A and B, where A sends 10 MiB to B.
- A has 100 packets in flight.
- A receives an ACK from B, acknowledging the first 2 packets.
- At the end of the first loop iteration:
  - `packets.len()` will be `0`, given the ACK from B acknowledges the first 2 packets.
  - `acked.len()` will be `2`, containing the two new acked `SentPacket`s.
  - `keep.len()` will be `98`, containing all remaining packets
- We then execute `self.packets.extend(keep)`, e.g. in the above scenario, adding all remaining 98 packets to the now empty `self.packets`.
- In other words, we re-insert all remaining packets on each ACK.

Unfortunately `BTreeMap` does not allow splitting out a full range. The closest to that is [`BTreeMap::extract_if`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.extract_if) which is currently Nightly only.

That said, I think the following change, optimizing for the scenario above, would get us a long way already:

``` diff
modified   neqo-transport/src/recovery/sent.rs
@@ -197,18 +197,17 @@ impl SentPackets {
     {
         let mut result = Vec::new();
         // Remove all packets. We will add them back as we don't need them.
-        let mut packets = std::mem::take(&mut self.packets);
         for range in acked_ranges {
-            // For each acked range, split off the acknowledged part,
-            // then split off the part that hasn't been acknowledged.
-            // This order works better when processing ranges that
-            // have already been processed, which is common.
-            let mut acked = packets.split_off(range.start());
-            let keep = acked.split_off(&(*range.end() + 1));
[...]

PR #2239, opened 2024-11-19
chore: prepare v0.11.0

mxinden:
New release to get https://github.com/mozilla/neqo/pull/2237 into mozilla-central.

Minor bump (`v0.10.0` -> `v0.11.0`) due to https://github.com/mozilla/neqo/pull/2216.

PR #2245, opened 2024-11-24
perf(transport/recovery): optimize `SentPackets::take_ranges`

mxinden:
`SentPackets` keep track of all inflight packets. On receiving an ACK, the acked
packet ranges are removed via `SentPackets:take_ranges`.

In normal network scenarios one can assume that the amount of packets before a
range is small (e.g. due to reordering) or zero, while the amount of packets
after a range is large, i.e. all remaining in-flight packets.

The previous implementation assumed the opposite, leading to an overhead linear
to the amount of packets after the range. This commit changes the algorithm such
that it is linear to the amount of packets before the range, which again, is
assumed to be smaller than the amount of packets after the acked range.

Fixes https://github.com/mozilla/neqo/issues/2242.


neqo-transport/benches/sent_packets.rs
mxinden, line 1: Once this benchmark is approved, I will split out the benchmark into a separate pull request. Merging that second pull request first will give us a baseline (i.e. `main` branch) before merging this pull request.

neqo-transport/benches/sent_packets.rs
```diff
@@ -0,0 +1,46 @@
...
+            now,
+            true,
+            Vec::new(),
+            100,
+        ))
+    }
+    pkts
+}
+
+fn take_ranges(c: &mut Criterion) {
```
martinthomson, line 34: It's probably worth putting in some more commentary.

```suggestion
/// Confirm that taking a small number of ranges from the front of
/// a large span of sent packets is performant.
/// This is the most common pattern when sending at a higher rate.
/// New acknowledgments will include low-numbered packets,
/// while the acknowledgment rate will ensure that most of the
/// outstanding packets remain in flight.
fn take_ranges(c: &mut Criterion) {
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -196,19 +196,50 @@ impl SentPackets {
         R::IntoIter: ExactSizeIterator,
     {
         let mut result = Vec::new();
-        // Remove all packets. We will add them back as we don't need them.
+
+        // Start with all packets. We will add them back as we don't need them.
```
martinthomson, line 200:
```suggestion
        // Start with all packets. We will add unacknowledged packets back.
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -196,19 +196,50 @@ impl SentPackets {
...
+
+            // According to RFC 9000 19.3.1 ACK ranges are in descending order:
+            //
+            // > Each ACK Range consists of alternating Gap and ACK Range Length
+            // > values in **descending packet number order**.
+            //
+            // <https://www.rfc-editor.org/rfc/rfc9000.html#section-19.3.1>
+            //
+            // Thus none of the following ACK ranges will acknowledge packets in
+            // `after_acked_range`. Let's put those back early.
```
martinthomson, line 223: Because you are relying on this, you should therefore make some assertions to guarantee that it is true.  `debug_assert!()` is fine.

neqo-transport/src/recovery/sent.rs
```diff
@@ -196,19 +196,50 @@ impl SentPackets {
...
+                // `after_acked_range` should be small.
+                self.packets.extend(after_acked_range);
+            }
+
+            // Take the acked packets.
+            result.extend(acked_range.into_values().rev());
         }
+
+        // Put remaining non-acked packets back.
         self.packets.extend(packets);
```
martinthomson, line 241: Does this need to be changed to the following as well?

```rust
            if self.packets.is_empty() {
                self.packets = after_acked_range;
            } else {
                self.packets.extend(after_acked_range);
            }
```

If the first range is contiguous with the end, then this might be faster.  It's not clear to me whether the additional condition is worth it though.

martinthomson (approved):

neqo-transport/src/recovery/sent.rs
mxinden: Can you think of a scenario where:
1. the range is contiguous with the end
2. AND the number of packets before the range  (here `packets`) is high?

I can only think of the following contrived scenario. Say that we have 1_000 packets in flight. Say that we receive an ACK range for packet 990 - 1_000, thus `packets` would be very large and `self.packets.extend(packets)` costly. But I would argue that in such case, `self.packets.extend(packets)` is the least important problem we have to worry about.

The above aside, I assume `packets` will always be small, thus `self.packets = packets` would not be much faster than `self.packets.extend(packets)`, and thus I don't think the above conditional on `self.packets.is_empty()` is worth it.

Agreed @martinthomson?

neqo-transport/src/recovery/sent.rs
mxinden: Good point. Done in https://github.com/mozilla/neqo/pull/2245/commits/d7454d009346cafb18b65dfd3b0d0f4f1a255d67.

neqo-transport/src/recovery/sent.rs
```diff
@@ -196,19 +196,60 @@ impl SentPackets {
...
+
+            // According to RFC 9000 19.3.1 ACK ranges are in descending order:
+            //
+            // > Each ACK Range consists of alternating Gap and ACK Range Length
+            // > values in **descending packet number order**.
+            //
+            // <https://www.rfc-editor.org/rfc/rfc9000.html#section-19.3.1>
+            if let Some(prev_start) = previous_range_start {
+                debug_assert!(
+                    *range.end() < prev_start,
```
mxinden, line 225: Given that ACK ranges are encoded on the wire relative to the previous range they can not overlap.

Given that `range` is inclusive, its length can not be zero.

Thus we need a `<` and not `<=`.

Please correct me if I am missing something.

larseggert: Not saying your analysis is false, but shouldn't we've hit the assert in that case?

neqo-transport/benches/sent_packets.rs
mxinden: Pull request merging benchmark only: https://github.com/mozilla/neqo/pull/2253

neqo-transport/src/recovery/sent.rs
martinthomson: Would `debug_assert!(prev_start.is_none_or(|s| s > *range.end()))` work better?

I don't think that you need a big print statement: it will never happen, except in a test.

neqo-transport/src/recovery/sent.rs
```diff
@@ -196,19 +196,60 @@ impl SentPackets {
...
+                // Need to extend existing one. Not the first iteration, thus
+                // `after_acked_range` should be small.
+                self.packets.extend(after_acked_range);
+            }
+
+            // Take the acked packets.
+            result.extend(acked_range.into_values().rev());
         }
+
+        // Put remaining non-acked packets back.
```
martinthomson, line 250:
```suggestion
        // Put remaining non-acked packets back.
        // This is inefficient if there the acknowledged packets include the last
        // sent packet AND there is a large unacknowledged span of packets.
        // That's rare enough that we won't do anything special for that case.
```

martinthomson (approved):
Land the other first, but this is good to go.

neqo-transport/src/recovery/sent.rs
mxinden: Committed through 31f82f4e.

mxinden:
For the record, highlighting the performance improvement of this pull request compared to baseline `main`.

> SentPackets::take_ranges: :green_heart: Performance has improved.
> 
>        time:   [5.5127 µs 5.5878 µs 5.6587 µs]
>        change: [-96.565% -96.506% -96.434%] (p = 0.00 < 0.05)
> 
> Found 4 outliers among 100 measurements (4.00%)

mxinden:
> We got to figure out these issues with `idle_timeout_crazy_rtt`; it's becoming annoying.

Agreed. @larseggert I will look into it once I am done releasing Neqo v0.11.0 (https://github.com/mozilla/neqo/pull/2239).

PR #2253, opened 2024-11-25
bench(transport/recovery): benchmark `SentPackets::take_ranges`

mxinden:
This pull request extracts the benchmark from https://github.com/mozilla/neqo/pull/2245.

I suggest we merge here first. We can then compare https://github.com/mozilla/neqo/pull/2245 against its baseline on `main`.

Note the second separate commit, addressing clippy lints only (82ef97e24c5af3d5098f5b73bcd23c72a7b365b3).

### DIFF
--- a/neqo-transport/src/recovery/mod.rs
+++ b/neqo-transport/src/recovery/mod.rs
@@ -6,6 +6,9 @@
 
 // Tracking of sent packets and detecting their loss.
 
+#[cfg(feature = "bench")]
+pub mod sent;
+#[cfg(not(feature = "bench"))]
 mod sent;
 mod token;
 
--- a/neqo-transport/src/recovery/mod.rs
+++ b/neqo-transport/src/recovery/mod.rs
@@ -526,9 +526,9 @@ impl LossRecovery {
     ///
     /// # Panics
     ///
-    /// Panics when the largest acknowledged or `loss_time` is already set. In
-    /// addition, the client should not have received any ACK frames when it
-    /// drops 0-RTT.
+    /// Panics when the largest acknowledged or `loss_time` is already set.
+    /// The client should not have received any ACK frames in the
+    /// application data packet number space when it drops 0-RTT.
     pub fn drop_0rtt(&mut self, primary_path: &PathRef, now: Instant) -> Vec<SentPacket> {
         assert!(self
             .spaces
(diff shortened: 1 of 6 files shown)


neqo-transport/src/recovery/mod.rs
```diff
@@ -505,9 +522,14 @@ impl LossRecovery {
         self.qlog = qlog;
     }
 
+    /// Drop all 0rtt packets.
+    ///
+    /// # Panics
+    ///
+    /// Panics when the largest acknowledged or `loss_time` is already set. In
+    /// addition, the client should not have received any ACK frames when it
+    /// drops 0-RTT.
```
martinthomson, line 531:
```suggestion
    /// Panics when the largest acknowledged or `loss_time` is already set.
    /// The client should not have received any ACK frames in the
    /// application data packet number space when it drops 0-RTT.
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -61,41 +62,54 @@ impl SentPacket {
...
     /// The length of the packet that was sent.
+    #[must_use]
     pub const fn len(&self) -> usize {
         self.len
     }
 
+    #[must_use]
+    pub const fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
```
martinthomson, line 109: Why did you add this?  Packets cannot be empty, so it doesn't make sense to have this.  You should suppress the warning instead.

neqo-transport/src/recovery/sent.rs
```diff
@@ -174,6 +194,8 @@ pub struct SentPackets {
 }
 
 impl SentPackets {
+    #[allow(clippy::len_without_is_empty)]
```
martinthomson, line 197: Is this the type you wanted to add an `is_empty()` function for?

martinthomson (approved):
It's possible that I'm confused.  I'm sure that you'll work it out.

neqo-transport/src/recovery/sent.rs
mxinden: Yes. Sorry about that. Fixed with https://github.com/mozilla/neqo/pull/2253/commits/1ed1ca09e6dccb7685c3b88db2afde943247d379.

neqo-transport/src/recovery/sent.rs
mxinden: The lint only complains, because `SentPackets` is now `pub` when `bench` feature is enabled.

Given that we don't need `SentPackets::is_empty` anywhere within the `neqo_transport::recovery` `mod` (i.e. parent of `sent` `mod`), I think suppressing is the better option here.

PR #2254, opened 2024-11-26
ci: Fix QNS diff to main

larseggert:
It seems like the cache is not resroring; checking why.

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -253,9 +253,12 @@ jobs:
           } >> comment.md
 
       - if: github.ref == 'refs/heads/main'
+        run: mv results results-main
+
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Ah, otherwise the previous restore (copied below) would not find `result-main` correct?

```
      - uses: actions/cache/restore@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
        with:
          path: results-main
          key: qns-${{ github.sha }}
          restore-keys: qns-
```

larseggert:
Yeah, I thought it would restore the cache into a new directory named `results-main`, but now I believe it is expecting a directory of that name _in_ the cache, and it's not finding one.

PR #2257, opened 2024-11-28
ci: Try a different formatting for the qns diff

larseggert:
CC @martinthomson


martinthomson (approved):
This seems to be working well enough now.

I wonder whether it might be better to mark new or removed items with an extra character.

New failures get `⚠️**X**` and removed failures get `🎉~~X~~` (pick your own emoji).

PR #2258, opened 2024-11-28
Add a benchmark for the decoder

martinthomson:
This adds a fairly simple varint decoding benchmark to neqo-common.  The next patch will include some improvements to the decoder.

### DIFF
--- a/neqo-common/Cargo.toml
+++ b/neqo-common/Cargo.toml
@@ -28,6 +28,8 @@ qlog = { workspace = true }
 windows = { version = "0.58", default-features = false, features = ["Win32_Media"] }
 
 [dev-dependencies]
+criterion = { version = "0.5", default-features = false }
+neqo-crypto = { path = "../neqo-crypto" }
 test-fixture = { path = "../test-fixture" }
 regex = { workspace = true }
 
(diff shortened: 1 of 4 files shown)


neqo-common/Cargo.toml
```diff
@@ -28,6 +28,8 @@ qlog = { workspace = true }
 windows = { version = "0.58", default-features = false, features = ["Win32_Media"] }
 
 [dev-dependencies]
+criterion = { version = "0.5", default-features = false }
+neqo-crypto = { path = "../neqo-crypto" }
```
larseggert, line 32: I think these should be regular dependencies gated by a `bench` feature, like we do in `neqo-transport/Cargo.toml`. (I'm surprised this compiled...)

martinthomson: It's a dev-dependency, which can't be feature-gated.

neqo-common/Cargo.toml
```diff
@@ -38,3 +40,7 @@ build-fuzzing-corpus = ["hex"]
 [lib]
 # See https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
 bench = false
+
+[[bench]]
+name = "decoder"
+harness = false
```
larseggert, line 46:
```suggestion
harness = false
required-features = ["bench"]
```

larseggert: Also needs `bench = []` under `[features]`.

larseggert (approved):

mxinden (approved):

PR #2259, opened 2024-11-29
Decoder improvements

martinthomson:
This started out as an attempt to make the decoder API slightly easier to use.

The `decode_uint()` function now takes a type as an argument, so if you want 1 byte, you say `let x: u8 = dec.decode_uint()` rather than `let x = dec.decode_unit(1)`.  This is nice because it eliminates code like this:  `let version = WireVersion::try_from(dec.decode_uint(4)?)?` (or with `.unwrap()` calls).  Now those are often just `let version = dec.decode_uint()`, or at least `let version: WireVersion = dec.decode_uint()`.

To that end, I've eliminated `decode_byte()` as well.

The big news is that the benchmark shows a massive improvement for varint decoding performance.  I was surprised at how big of an improvement it was.

<details>
<summary>Performance improvements</summary>

```
decode 1024 bytes, mask ff
                        time:   [1.3609 µs 1.3636 µs 1.3666 µs]
                        change: [-23.308% -23.055% -22.813%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  1 (1.00%) high severe

decode 32768 bytes, mask ff
                        time:   [43.383 µs 43.487 µs 43.609 µs]
                        change: [-23.465% -23.173% -22.899%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe

decode 1048576 bytes, mask ff
                        time:   [1.3416 ms 1.3448 ms 1.3483 ms]
                        change: [-24.853% -24.593% -24.334%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

decode 1024 bytes, mask 7f
                        time:   [2.4318 µs 2.4380 µs 2.4451 µs]
                        change: [-26.561% -26.237% -25.855%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe

decode 32768 bytes, mask 7f
                        time:   [77.761 µs 77.907 µs 78.068 µs]
                        change: [-26.971% -26.604% -26.251%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low severe
  4 (4.00%) high mild

decode 1048576 bytes, mask 7f
                        time:   [2.3282 ms 2.3358 ms 2.3438 ms]
                        change: [-28.019% -27.685% -27.356%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild

decode 1024 bytes, mask 3f
                        time:   [574.52 ns 578.26 ns 582.69 ns]
                        change: [-66.618% -66.344% -66.053%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

decode 32768 bytes, mask 3f
                        time:   [18.192 µs 18.257 µs 18.329 µs]
                        change: [-66.335% -66.097% -65.844%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe

decode 1048576 bytes, mask 3f
                        time:   [579.71 µs 581.40 µs 583.33 µs]
                        change: [-60.449% -60.182% -59.911%] (p = 0.00 < 0.05)
                        Performance has improved.
  7 (7.00%) high mild
  4 (4.00%) high severe
```

</details>

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -102,13 +102,10 @@ impl<'a> Decoder<'a> {
         if self.remaining() < n {
             return None;
         }
-        let mut v = 0_u64;
-        for i in 0..n {
-            let b = self.buf[self.offset + i];
-            v = v << 8 | u64::from(b);
-        }
+        let mut buf = [0; 8];
+        buf[8 - n..].copy_from_slice(&self.buf[self.offset..self.offset + n]);
         self.offset += n;
-        Some(v)
+        Some(u64::from_be_bytes(buf))
     }
 
     /// Decodes a QUIC varint.
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -4,7 +4,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::fmt::Debug;
+use std::{fmt::Debug, mem::size_of};
 
 use crate::hex_with_len;
 
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -44,12 +44,15 @@ impl<'a> Decoder<'a> {
     }
 
     /// Skip helper that panics if `n` is `None` or not able to fit in `usize`.
+    /// Only use this for tests because we panic rather than reporting a result.
+    #[cfg(any(test, feature = "test-fixture"))]
     fn skip_inner(&mut self, n: Option<u64>) {
         self.skip(usize::try_from(n.expect("invalid length")).unwrap());
     }
 
     /// Skip a vector.  Panics if there isn't enough space.
     /// Only use this for tests because we panic rather than reporting a result.
+    #[cfg(any(test, feature = "test-fixture"))]
     pub fn skip_vec(&mut self, n: usize) {
         let len = self.decode_n(n);
         self.skip_inner(len);
@@ -57,6 +60,7 @@ impl<'a> Decoder<'a> {
 
     /// Skip a variable length vector.  Panics if there isn't enough space.
     /// Only use this for tests because we panic rather than reporting a result.
+    #[cfg(any(test, feature = "test-fixture"))]
     pub fn skip_vvec(&mut self) {
         let len = self.decode_varint();
         self.skip_inner(len);
@@ -272,6 +276,7 @@ impl Encoder {
     /// # Panics
     ///
     /// When `s` contains non-hex values or an odd number of values.
+    #[cfg(any(test, feature = "test-fixture"))]
     #[must_use]
     pub fn from_hex(s: impl AsRef<str>) -> Self {
         let s = s.as_ref();
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -51,7 +51,7 @@ impl<'a> Decoder<'a> {
     /// Skip a vector.  Panics if there isn't enough space.
     /// Only use this for tests because we panic rather than reporting a result.
     pub fn skip_vec(&mut self, n: usize) {
-        let len = self.decode_uint(n);
+        let len = self.decode_n(n);
         self.skip_inner(len);
     }
 
@@ -60,16 +60,6 @@ impl<'a> Decoder<'a> {
     pub fn skip_vvec(&mut self) {
         let len = self.decode_varint();
         self.skip_inner(len);
-    }
-
-    /// Decodes (reads) a single byte.
-    pub fn decode_byte(&mut self) -> Option<u8> {
-        if self.remaining() < 1 {
-            return None;
-        }
-        let b = self.buf[self.offset];
-        self.offset += 1;
-        Some(b)
     }
 
     /// Provides the next byte without moving the read position.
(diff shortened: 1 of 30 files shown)


neqo-common/src/codec.rs
```diff
@@ -92,33 +82,43 @@ impl<'a> Decoder<'a> {
...
             return None;
         }
-        let mut v = 0_u64;
-        for i in 0..n {
-            let b = self.buf[self.offset + i];
-            v = v << 8 | u64::from(b);
-        }
-        self.offset += n;
-        Some(v)
+        Some(if n == 1 {
```
martinthomson, line 91: This looks pretty bad, but it's the reason for the largest part of the performance wins.  The other change is a 5% improvement and maybe even a regression in some cases, but this if statement leads to huge gains.

larseggert:
I had been meaning to take a look at what s2n-quic is doing; I somehow remember them having a pretty optimized implementation for this: https://github.com/aws/s2n-quic/blob/main/quic/s2n-quic-core/src/varint/mod.rs

neqo-common/src/codec.rs
```diff
@@ -92,33 +82,43 @@ impl<'a> Decoder<'a> {
...
 
-    /// Decodes an unsigned integer of length 1..=8.
-    ///
-    /// # Panics
-    ///
-    /// This panics if `n` is not in the range `1..=8`.
-    pub fn decode_uint(&mut self, n: usize) -> Option<u64> {
-        assert!(n > 0 && n <= 8);
+    #[inline]
+    pub(crate) fn decode_n(&mut self, n: usize) -> Option<u64> {
```
larseggert, line 86: Would there be any further performance benefit in splitting this out into seven functions, e.g., 
```suggestion
    pub(crate) fn decode_1(&mut self, n: usize) -> Option<u64> {...}
    ...
    pub(crate) fn decode_8(&mut self, n: usize) -> Option<u64> {...}
```
each with optimized code for that length? (Is there a way to do this with a generic parameter more concisely?)

larseggert (approved):

neqo-common/src/codec.rs
mxinden, line 86: Given that `n` is known at compile time, i.e. derived from `size_of::<T>()`, one could leverage const generics here:

```suggestion
    pub(crate) fn decode_n<const N: usize>(&mut self) -> Option<u64> {
```

That said, I don't think it has a performance gain:

```
➜  critcmp baseline const                                                                                                  
group                            baseline                                const
-----                            --------                                -----
decode 1024 bytes, mask 3f       1.00  1066.4±116.06ns        ? ?/sec    1.02   1092.5±2.61ns        ? ?/sec
decode 1024 bytes, mask 7f       1.00      3.5±0.15µs        ? ?/sec     1.01      3.5±0.00µs        ? ?/sec
decode 1024 bytes, mask ff       1.00      2.3±0.00µs        ? ?/sec     1.00      2.3±0.00µs        ? ?/sec
decode 1048576 bytes, mask 3f    1.00   1112.8±2.42µs        ? ?/sec     1.00  1116.0±11.03µs        ? ?/sec
decode 1048576 bytes, mask 7f    1.00      3.3±0.40ms        ? ?/sec     1.10      3.6±0.01ms        ? ?/sec
decode 1048576 bytes, mask ff    1.00      2.3±0.01ms        ? ?/sec     1.00      2.3±0.00ms        ? ?/sec
decode 32768 bytes, mask 3f      1.00     34.8±0.20µs        ? ?/sec     1.00     34.8±0.44µs        ? ?/sec
decode 32768 bytes, mask 7f      1.00    112.1±0.18µs        ? ?/sec     1.00    112.0±0.14µs        ? ?/sec
decode 32768 bytes, mask ff      1.00     73.7±0.12µs        ? ?/sec     1.00     73.8±0.11µs        ? ?/sec
```

test-fixture/src/header_protection.rs
```diff
@@ -120,11 +120,9 @@ pub fn remove_header_protection(hp: &HpKey, header: &[u8], payload: &[u8]) -> (V
     // Trim down to size.
     fixed_header.truncate(pn_offset + pn_len);
     // The packet number should be 1.
-    let pn = Decoder::new(&fixed_header[pn_offset..])
-        .decode_uint(pn_len)
-        .unwrap();
-
-    (fixed_header, pn)
+    let mut pn = [0; 8];
+    pn[8 - pn_len..].copy_from_slice(&fixed_header[pn_offset..pn_offset + pn_len]);
```
mxinden, line 124: What is the reason for no longer going through `Decoder` here?

neqo-common/src/codec.rs
mxinden: Oh, I would have expected replacing the `for` loop with the `copy_from_slice` to have the highest impact.

mxinden (commented):
Change looks good to me!

neqo-common/src/codec.rs
larseggert: Just the generic won't lead to improvements (the compiler should be smart enough), but I think optimized code per length might help.

test-fixture/src/header_protection.rs
martinthomson: I didn't want to expose the three-byte variant.  It seemed like an OK compromise given that we aren't using `Decoder` for its `Cursor`-like capabilities, just the integer decoding.

neqo-common/src/codec.rs
martinthomson: Yeah, I tried the const generic :)  The n==1 optimization below is part of that.

martinthomson:
I took at look at the s2n implementation, which peeks a byte, then reads a full 1, 2, 4, or 8 bytes.  For some reason, the following code is much slower (this PR is 19% faster for an 0xff mask, 27% faster for 0x7f, and 65% faster for 0x3f).

```diff
     pub fn decode_varint(&mut self) -> Option<u64> {
-        let b1 = self.decode_n(1)?;
+        let b1 = self.peek_byte()?;
         match b1 >> 6 {
-            0 => Some(b1),
-            1 => Some((b1 & 0x3f) << 8 | self.decode_n(1)?),
-            2 => Some((b1 & 0x3f) << 24 | self.decode_n(3)?),
-            3 => Some((b1 & 0x3f) << 56 | self.decode_n(7)?),
+            0 => {
+                self.offset += 1;
+                Some(u64::from(b1))
+            }
+            1 => Some(u64::from(self.decode_uint::<u16>()? & 0x3fff)),
+            2 => Some(u64::from(self.decode_uint::<u32>()? & 0x3fff_ffff)),
+            3 => Some(self.decode_uint::<u64>()? & 0x3fff_ffff_ffff_ffff),
             _ => unreachable!(),
         }
     }
```

larseggert:
Also wonder if https://docs.rs/zerocopy/latest/zerocopy/ would further help here, but probably not, given that we only ever read a single value.

neqo-transport/src/tracking.rs
```diff
@@ -1062,7 +1062,7 @@ mod tests {
         assert_eq!(stats.ack, 1);
 
         let mut dec = builder.as_decoder();
-        _ = dec.decode_byte().unwrap(); // Skip the short header.
+        _ = dec.decode_uint::<u8>().unwrap(); // Skip the short header.
```
larseggert, line 1065:
```suggestion
        dec.skip(1); // Skip the short header.
```

test-fixture/src/header_protection.rs
larseggert: Maybe make that reasoning a comment?

larseggert (approved):

Issue #2162, opened 2024-10-08
`idle_timeout_crazy_rtt` hangs more often now


larseggert:
Something recently changed that makes the idle_timeout_crazy_rtt test not terminate pretty frequently, e.g., https://github.com/mozilla/neqo/actions/runs/11161582095/job/31024435251#step:7:1334

martinthomson:
Your changes to reduce send rate when packets are lost will probably extend the time needed.

larseggert:
But not by hours of real time?

martinthomson:
No, not that.  I'm not seeing what conditions resulted in the stall though; the output of the run seems to have been lost.  If we have a seed for the test, it could be replicated locally.

larseggert:
This one, for example: https://github.com/mozilla/neqo/actions/runs/11236671253/job/31237337009

larseggert:
#2165 may help, once it's live in CI.

larseggert:
This should not have closed, since #2165 wasn't a fix.

But it just found this seed, which caused a hang: `1856037d53afc27408ba15ff7b433b729f16e1ca087403a3e3eb9f2ed82856b5`.

larseggert:
e2bca5650d71761da5ef7eac741b3880dbe8846bc7f8572a40371a065585605a failed here: https://github.com/mozilla/neqo/actions/runs/11888582176/job/33123413099?pr=2234

larseggert:
I think this may be fixed? Have not seen this issue in a while.

larseggert:
Closing this, since we haven't seen any recent occurrences.

PR #2203, opened 2024-10-26
test(transport): assert maximum bandwidth on gbit link

mxinden:
This commit adds a basic smoke test using the `test-fixture` simulator, asserting the expected bandwidth on a 1 gbit link.

Given https://github.com/mozilla/neqo/issues/733, the current expected bandwidth is limited by the fixed sized stream receive buffer (1MiB), not by the bandwidth of the link.

While a bit unconventional, I think it is worth having this smoke test, to make sure we don't regress. What do folks think?

PR #2263, opened 2024-12-02
fix(sim): correct `Waiting` state comparison in `NodeHolder::ready()`

mxinden:
A `Node` (e.g. a `Client`, `Server` or `TailDrop` router) can be in 3 states:

``` rust
enum NodeState {
    /// The node just produced a datagram.  It should be activated again as soon as possible.
    Active,
    /// The node is waiting.
    Waiting(Instant),
    /// The node became idle.
    Idle,
}
```

`NodeHolder::ready()` determines whether a `Node` is ready to be processed again. When `NodeState::Waiting`, it should only be ready when `t <= now`, i.e. the waiting time has passed, not `t >= now`.

``` rust
impl NodeHolder {
    fn ready(&self, now: Instant) -> bool {
        match self.state {
            Active => true,
            Waiting(t) => t <= now, // not >=
            Idle => false,
        }
    }
}
```

The previous behavior lead to wastefull non-ready `Node`s being processed and thus a large test runtime when e.g. simulating a gbit connection (https://github.com/mozilla/neqo/pull/2203).

---

Time to simulate a 100 MB transfer on a 1 Gbit/s 100ms connection on the `Simulator` on my `12th Gen Intel(R) Core(TM) i9-12900HK`:

- Before 40.51s
- After 17.84s

### DIFF
--- a/test-fixture/src/sim/mod.rs
+++ b/test-fixture/src/sim/mod.rs
@@ -110,7 +110,7 @@ impl NodeHolder {
     fn ready(&self, now: Instant) -> bool {
         match self.state {
             Active => true,
-            Waiting(t) => t >= now,
+            Waiting(t) => t <= now,
             Idle => false,
         }
     }


larseggert (approved):
Your explanation make sense, but suggest for @martinthomson to take a look, as he is more familiar with the simulator.

larseggert:
Is there any chance this might fix #2162?

mxinden:
Looking at the logs of a recent failure, it seems like in the latter half of the test execution, only the `Drop` node is being skipped.

```
2024-11-18T08:42:02.5636603Z 7s500ms DEBUG [idle_timeout_crazy_rtt] skipping drop
```

Thus, I don't think this patch will fix #2162 unfortunately.

martinthomson (approved):

PR #2264, opened 2024-12-03
chore: Guard testing functions properly

larseggert:
We have a bunch of functions that in their doc comments say that they are for testing purposes only. This commit adds guards to those functions to make sure they are not used in production code.

### DIFF
--- a/neqo-common/Cargo.toml
+++ b/neqo-common/Cargo.toml
@@ -37,6 +37,7 @@ regex = { workspace = true }
 bench = ["test-fixture/bench"]
 build-fuzzing-corpus = ["hex"]
 ci = []
+test-fixture = []
 
 [lib]
 # See https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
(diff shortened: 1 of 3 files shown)


neqo-common/Cargo.toml
```diff
@@ -37,6 +37,7 @@ regex = { workspace = true }
 bench = ["test-fixture/bench"]
 build-fuzzing-corpus = ["hex"]
 ci = []
+test-fixture = []
```
mxinden, line 40: Wouldn't defining `test-fixture` as `optional = true` be enough?

https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

`cargo` should implicitly add the `test-fixture` feature.

mxinden (approved):
Very much in favor of not accidentally using test functions in production code.

neqo-common/Cargo.toml
larseggert: dev-dependencies are not allowed to be optional, I think?

mxinden: Ah, I didn't know.

https://github.com/rust-lang/cargo/issues/1596

Please ignore my comment then.

PR #2266, opened 2024-12-04
fix: Send Retry and VN packets with default TOS

larseggert:
Rather than using the TOS value from the incoming packet, which might have gotten ECN marked on the way in.

Also create a debug log entry for those packets that resembles that created by `dump_packet`.

### DIFF
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -238,7 +238,7 @@ impl Server {
                         |p| {
                             qdebug!(
                                 [self],
-                                "type={:?} path:{} {}->{} {:?} len {}",
+                                "type={:?} path:{} {}->{} len {}",
                                 PacketType::Retry,
                                 initial.dst_cid,
                                 dgram.destination(),
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -399,7 +399,7 @@ impl Server {
             );
             qdebug!(
                 [self],
-                "type={:?} path:{} {}->{} {:?} len {}",
+                "type={:?} path:{} {}->{} len {}",
                 PacketType::VersionNegotiation,
                 packet.dcid(),
                 dgram.destination(),
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -238,7 +238,7 @@ impl Server {
                         |p| {
                             qdebug!(
                                 [self],
-                                "type={:?} path:{} {}->{} len {}",
+                                "type={:?} path:{} {}->{} {:?} len {}",
                                 PacketType::Retry,
                                 initial.dst_cid,
                                 dgram.destination(),
@@ -399,7 +399,7 @@ impl Server {
             );
             qdebug!(
                 [self],
-                "type={:?} path:{} {}->{} len {}",
+                "type={:?} path:{} {}->{} {:?} len {}",
                 PacketType::VersionNegotiation,
                 packet.dcid(),
                 dgram.destination(),
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -17,7 +17,8 @@ use std::{
 };
 
 use neqo_common::{
-    event::Provider, hex, qdebug, qerror, qinfo, qlog::NeqoQlog, qtrace, qwarn, Datagram, Role,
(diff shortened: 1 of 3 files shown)


mxinden (approved):

neqo-transport/src/server.rs
```diff
@@ -235,10 +236,20 @@ impl Server {
...
                         },
                         |p| {
+                            qdebug!(
+                                [self],
+                                "type={:?} path:{} {}->{} {:?} len {}",
+                                PacketType::Retry,
+                                initial.dst_cid,
+                                dgram.destination(),
+                                dgram.source(),
+                                IpTos::default(),
```
martinthomson, line 246:
```suggestion
```
No need to log a constant.

neqo-transport/src/server.rs
```diff
@@ -235,10 +236,20 @@ impl Server {
                             Output::None
                         },
                         |p| {
+                            qdebug!(
+                                [self],
+                                "type={:?} path:{} {}->{} {:?} len {}",
```
martinthomson, line 241:
```suggestion
                                "type={:?} path:{} {}->{} len {}",
```

neqo-transport/src/server.rs
```diff
@@ -386,6 +397,16 @@ impl Server {
...
                 self.conn_params.get_versions().all(),
             );
+            qdebug!(
+                [self],
+                "type={:?} path:{} {}->{} {:?} len {}",
+                PacketType::VersionNegotiation,
+                packet.dcid(),
+                dgram.destination(),
+                dgram.source(),
+                IpTos::default(),
```
martinthomson, line 407:
```suggestion
```

neqo-transport/src/server.rs
```diff
@@ -386,6 +397,16 @@ impl Server {
                 packet.wire_version(),
                 self.conn_params.get_versions().all(),
             );
+            qdebug!(
+                [self],
+                "type={:?} path:{} {}->{} {:?} len {}",
```
martinthomson, line 402:
```suggestion
                "type={:?} path:{} {}->{} len {}",
```

martinthomson (changes requested):
Presumably we have tests for Retry that can be modified to check that the TOS is correct.

neqo-transport/src/server.rs
larseggert: Logging it, because it's also part of the `dump_packet` output.

neqo-transport/src/server.rs
larseggert: Ditto.

test-fixture/src/assertions.rs
```diff
@@ -50,8 +50,9 @@ pub fn assert_version(payload: &[u8], v: u32) {
 /// # Panics
 ///
 /// If this is clearly not a Version Negotiation packet.
-pub fn assert_vn(payload: &[u8]) {
-    let mut dec = Decoder::from(payload);
+pub fn assert_vn(dgram: &Datagram) {
+    assert_eq!(dgram.tos(), IpTos::default());
```
martinthomson, line 54:
```suggestion
    assert_eq!(dgram.tos(), IpTos::default());  // VN always uses default IP TOS.
```

test-fixture/src/assertions.rs
```diff
@@ -86,8 +87,9 @@ pub fn assert_coalesced_0rtt(payload: &[u8]) {
 /// # Panics
 ///
 /// If the tests fail.
-pub fn assert_retry(payload: &[u8]) {
-    let mut dec = Decoder::from(payload);
+pub fn assert_retry(dgram: &Datagram) {
+    assert_eq!(dgram.tos(), IpTos::default());
```
martinthomson, line 91:
```suggestion
    assert_eq!(dgram.tos(), IpTos::default());  // Retry always uses default IP TOS.
```

neqo-transport/tests/retry.rs
```diff
@@ -37,12 +37,10 @@ fn retry_basic() {
 
     let dgram = client.process_output(now()).dgram(); // Initial
     assert!(dgram.is_some());
```
martinthomson, line 39: Just confirming that we *don't* use the default IP TOS for a typical Initial, because we set ECT(0).  In that case, this is good.

martinthomson (approved):

neqo-transport/tests/retry.rs
larseggert: Correct.

mxinden (approved):

PR #2267, opened 2024-12-04
chore: Simplify some webtransport checks

### LINKED ISSUE #1255: I wonder if we can't make a `PartialEq<(&str, &str)>` implementation for headers to make this more ergonomic.  `PartialEq<(impl AsRef<str>, impl AsRef<str>>` might be even better, but I don't know if we would be allowed to do that.
I wonder if we can't make a `PartialEq<(&str, &str)>` implementation for headers to make this more ergonomic.  `PartialEq<(impl AsRef<str>, impl AsRef<str>>` might be even better, but I don't know if we would be allowed to do that.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1252#discussion_r725738458_

larseggert:
So this just uses it.

While I'm here, eliminate some code redundancies.

Fixes #1255

### DIFF
--- a/neqo-http3/src/recv_message.rs
+++ b/neqo-http3/src/recv_message.rs
@@ -6,7 +6,7 @@
 
 use std::{cell::RefCell, cmp::min, collections::VecDeque, fmt::Debug, rc::Rc};
 
-use neqo_common::{qdebug, qinfo, qtrace, Header};
+use neqo_common::{header::HeadersExt, qdebug, qinfo, qtrace, Header};
 use neqo_qpack::decoder::QPackDecoder;
 use neqo_transport::{Connection, StreamId};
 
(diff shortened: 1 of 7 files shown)


neqo-http3/src/recv_message.rs
```diff
@@ -167,12 +167,8 @@ impl RecvMessage {
         }
 
         let is_web_transport = self.message_type == MessageType::Request
-            && headers
-                .iter()
-                .any(|h| h.name() == ":method" && h.value() == "CONNECT")
-            && headers
-                .iter()
-                .any(|h| h.name() == ":protocol" && h.value() == "webtransport");
+            && headers.contains(&Header::new(":method", "CONNECT"))
```
mxinden, line 170: `Header::new` allocates both key and value into a `String`. Is that an issue here?

https://github.com/mozilla/neqo/blob/a7581770ff138c16727e92ecb7b47d9ceb95425e/neqo-common/src/header.rs#L13-L23

If yes, we could have `Header` wrap `Cow<'static, str>`, or introduce a `BorrowedHeader`.

martinthomson: Either works and might save a tiny amount of work.  Though we are avoiding most of the work on the client side here, so it's not that important.

martinthomson: Reflecting on this, perhaps we could implement `PartialEq<(&str, &str)> for Header` and avoid all of that complication.  Then we'd only have to delete the string `Header::new` from these...

martinthomson: https://github.com/larseggert/neqo/pull/32 is where I got.  It turns out that `slice::contains` doesn't have the right signature.

mxinden (approved):
Neat! Thank you for the follow-up.

PR #2268, opened 2024-12-04
chore: Check request headers for `:status`

### LINKED ISSUE #1274: check that - for a request, we DON'T get :status?
I know that this is old code, but shouldn't we be checking that - for a request, we DON'T get :status?

That means we need a set of bits that should be on and a set of bits that should be off.  Or:

```rust
        const PSEUDO_HEADER_ALL: u8 = PSEUDO_HEADER_STATUS | PSEUDO_HEADER_METHOD | PSEUDO_HEADER_SCHEME | PSEUDO_HEADER_AUTHORITY | PSEUDO_HEADER_PATH;
        let (ph_required, ph_mask) = match message_type {
             // Responses contain only :status
             MessageType::Response => (PSEUDO_HEADER_STATUS, PSEUDO_HEADER_ALL),
            MessageType::Request => {
                if method_value == Some(&"CONNECT".to_string()) {
                    // CONNECT requests contain only :method and :authority
                    // This won't work once we get to :protocol...
                    (PSEUDO_HEADER_METHOD | PSEUDO_HEADER_AUTHORITY, PSEUDO_HEADER_ALL)
                } else {
                    // Other requests contain :method, :scheme, and :path, and may contain :authority.
                    (PSEUDO_HEADER_METHOD | PSEUDO_HEADER_SCHEME | PSEUDO_HEADER_PATH, PSEUDO_HEADER_ALL & !PSEUDO_HEADER_AUTHORITY)
                }
            }
        };
        if pseudo_state & ph_mask != ph_required {
            return Err(Error::InvalidHeader);
        }
```

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1270#discussion_r755542245_

larseggert:
This tries to implement the suggestion @martinthomson made in #1274. Needs a review from someone who knows the code better than me.

Fixes #1274

### DIFF
--- a/neqo-http3/src/headers_checks.rs
+++ b/neqo-http3/src/headers_checks.rs
@@ -130,29 +130,27 @@ pub fn headers_valid(headers: &[Header], message_type: MessageType) -> Res<()> {
         // Responses contain only :status
         MessageType::Response => (enum_set!(PseudoHeaderState::Status), PSEUDO_HEADER_ALL),
         MessageType::Request if method_value == Some("CONNECT") => {
-                const CONNECT_MASK: EnumSet<PseudoHeaderState> =
-                    enum_set!(PseudoHeaderState::Method | PseudoHeaderState::Authority);
-                if let Some(protocol) = protocol_value {
-                    // For a webtransport CONNECT, the :scheme field must be set to https.
-                    if protocol == "webtransport" && scheme_value != Some("https") {
-                        return Err(Error::InvalidHeader);
-                    }
-                    // The CONNECT request for with :protocol included must have the scheme,
-                    // authority, and path set.
-                    (
-                        CONNECT_MASK | PseudoHeaderState::Scheme | PseudoHeaderState::Path,
-                        PSEUDO_HEADER_ALL,
-                    )
-                } else {
-                    (CONNECT_MASK, PSEUDO_HEADER_ALL)
+            const CONNECT_MASK: EnumSet<PseudoHeaderState> =
+                enum_set!(PseudoHeaderState::Method | PseudoHeaderState::Authority);
+            if let Some(protocol) = protocol_value {
+                // For a webtransport CONNECT, the :scheme field must be set to https.
+                if protocol == "webtransport" && scheme_value != Some("https") {
+                    return Err(Error::InvalidHeader);
                 }
-            } else {
+                // The CONNECT request for with :protocol included must have the scheme,
+                // authority, and path set.
                 (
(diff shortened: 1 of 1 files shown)


neqo-http3/src/headers_checks.rs
```diff
@@ -119,24 +126,32 @@ pub fn headers_valid(headers: &[Header], message_type: MessageType) -> Res<()> {
     }
     // Clear the regular header bit, since we only check pseudo headers below.
     pseudo_state.remove(PseudoHeaderState::Regular);
-    let pseudo_header_mask = match message_type {
-        MessageType::Response => enum_set!(PseudoHeaderState::Status),
+    let (pseudo_header_required, pseudo_header_mask) = match message_type {
+        // Responses contain only :status
+        MessageType::Response => (enum_set!(PseudoHeaderState::Status), PSEUDO_HEADER_ALL),
         MessageType::Request => {
             if method_value == Some("CONNECT") {
```
mxinden, line 133: Feel free to ignore.

```suggestion
        MessageType::Request if method_value == Some("CONNECT") => {
```

Given that the `if` `else` block spans the entire `match` arm, the above would safe a level of indentation.

I can also propose this as a follow-up.

martinthomson (changes requested):
This looks fine, but it really needs a test to prove that it has addressed the issue.

neqo-http3/src/headers_checks.rs
larseggert: `rustfmt` seems to want to add the braces back around the `if`.

larseggert:
So I added a test, and it passes even without this PR. I think this is because @KershawChang already fixed this back in #1022?

neqo-http3/src/headers_checks.rs
martinthomson: So you can't do:

```rust
match whatever {
  MessageType::Request if method_value == Some("CONNECT") => { code }
  MessageType::Request => { else_clause }
}
```
?

larseggert:
@KershawChang do you recall fixing this? If yes, I will happily close this PR.

KershawChang:
> @KershawChang do you recall fixing this? If yes, I will happily close this PR.

Sorry for the delayed reply.
Based on the code below, yes, I think this was fixed.
https://github.com/mozilla/neqo/blob/ad8f3909d9ae96219e56a7c31e6822d6e64a6dc9/neqo-http3/src/headers_checks.rs#L138

I think we can close this PR. Thanks.

PR #2269, opened 2024-12-04
fix: Abort on RX of illegal stream control/data frame

### LINKED ISSUE #1457: Reject RESET_STREAM and STREAM_DATA_BLOCKED on non-existent streams
The specification doesn't mention what to do when you receive a RESET_STREAM or STREAM_DATA_BLOCKED frame on a bidirectional stream that you are responsible for creating, but you haven't created it yet.

We should kill the connection if this happens.  Otherwise, we are acknowledging frames that make changes to the state of streams. We need to stop ignoring those frames.  However, the frame could be a late-arriving frame for a stream that has been discarded, so a little care is needed to distinguish between streams that haven't been created yet from streams that have been closed already.

It also looks like we ignore STREAM frames in the same way, despite the spec being [clear on that point](https://quicwg.org/base-drafts/rfc9000.html#section-19.8-3).  That's at least as bad as ignoring RESET_STREAM.

We might also ignore these frames when they are sent in the wrong direction for unidirectional streams.  That would be bad too.

As the specification is not clear about what error code to use, but STREAM_STATE_ERROR seems most appropriate.

larseggert:
Abort the connection if we receive a stream control or data frame for a stream that is ours to initiate but we haven't yet. Don't abort if it is such a frame that arrives very late for a stream that we did use.

Fixes #1457

### DIFF
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -18,7 +18,7 @@ use crate::{
     events::ConnectionEvent,
     frame::{
         FRAME_TYPE_MAX_STREAM_DATA, FRAME_TYPE_RESET_STREAM, FRAME_TYPE_STOP_SENDING,
-        FRAME_TYPE_STREAM_DATA_BLOCKED,
+        FRAME_TYPE_STREAM_CLIENT_INI_BIDI, FRAME_TYPE_STREAM_DATA_BLOCKED,
     },
     packet::PacketBuilder,
     recv_stream::RECV_BUFFER_SIZE,
@@ -539,38 +539,39 @@ fn do_not_accept_data_after_stop_sending() {
     );
 }
 
-struct IllegalWriter(Vec<u64>);
-
-impl crate::connection::test_internal::FrameWriter for IllegalWriter {
-    fn write_frames(&mut self, builder: &mut PacketBuilder) {
-        builder.write_varint_frame(&self.0);
-    }
-}
-
-// Server sends a stream-related frame for client-initiated stream that is not yet created.
-fn illegal_frame_test(frame: &[u64]) {
-    let mut client = default_client();
-    let mut server = default_server();
-    connect(&mut client, &mut server);
-    let dgram = send_with_extra(&mut server, IllegalWriter(frame.to_vec()), now());
-    client.process_input(dgram, now());
-    assert!(client.state().closed());
-}
-
 #[test]
+/// Server sends a number of stream-related frames for a client-initiated stream that is not yet
+/// created. This should cause the client to close the connection.
 fn illegal_stream_frames() {
+    struct IllegalWriter(Vec<u64>);
+
+    impl crate::connection::test_internal::FrameWriter for IllegalWriter {
+        fn write_frames(&mut self, builder: &mut PacketBuilder) {
+            builder.write_varint_frame(&self.0);
+        }
+    }
+
+    fn test_with_illegal_frame(frame: &[u64]) {
+        let mut client = default_client();
+        let mut server = default_server();
+        connect(&mut client, &mut server);
+        let dgram = send_with_extra(&mut server, IllegalWriter(frame.to_vec()), now());
+        client.process_input(dgram, now());
+        assert!(client.state().closed());
+    }
+
     // 0 = Client-Initiated, Bidirectional; 2 = Client-Initiated, Unidirectional
     for stream_id in [0, 2] {
         // Illegal RESET_STREAM frame
-        illegal_frame_test(&[FRAME_TYPE_RESET_STREAM, stream_id, 0, 0]);
+        test_with_illegal_frame(&[FRAME_TYPE_RESET_STREAM, stream_id, 0, 0]);
         // Illegal STOP_SENDING frame
-        illegal_frame_test(&[FRAME_TYPE_STOP_SENDING, stream_id, 0]);
+        test_with_illegal_frame(&[FRAME_TYPE_STOP_SENDING, stream_id, 0]);
         // Illegal MAX_STREAM_DATA frame
-        illegal_frame_test(&[FRAME_TYPE_MAX_STREAM_DATA, stream_id, 0]);
+        test_with_illegal_frame(&[FRAME_TYPE_MAX_STREAM_DATA, stream_id, 0]);
         // Illegal STREAM_DATA_BLOCKED frame
-        illegal_frame_test(&[FRAME_TYPE_STREAM_DATA_BLOCKED, stream_id, 0]);
+        test_with_illegal_frame(&[FRAME_TYPE_STREAM_DATA_BLOCKED, stream_id, 0]);
         // Illegal STREAM frame
-        illegal_frame_test(&[0x08, stream_id, 0]);
+        test_with_illegal_frame(&[FRAME_TYPE_STREAM_CLIENT_INI_BIDI, stream_id, 0]);
     }
 }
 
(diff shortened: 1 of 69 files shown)


neqo-transport/src/connection/tests/stream.rs
```diff
@@ -539,6 +539,62 @@ fn do_not_accept_data_after_stop_sending() {
...
+        illegal_frame_test(&[FRAME_TYPE_STREAM_DATA_BLOCKED, stream_id, 0]);
+    }
+}
+
+#[test]
+fn illegal_stream_frame() {
+    for stream_id in [0, 2] {
+        illegal_frame_test(&[0x08, stream_id, 0]);
+    }
+}
```
mxinden, line 596: I would be fine combining this into a single `#[test]` function, thus being able to de-duplicate the `for stream_id in [0, 2]`.

Feel free to ignore.

neqo-transport/src/streams.rs
```diff
@@ -131,6 +131,8 @@ impl Streams {
                 stats.reset_stream += 1;
                 if let (_, Some(rs)) = self.obtain_stream(*stream_id)? {
                     rs.reset(*application_error_code, *final_size)?;
+                } else {
+                    return Err(Error::StreamStateError);
```
mxinden, line 135: What if `self.obtain_stream(*stream_id)?` is `(None, None)`, where `stream_id` is a receive stream that has already been reset?

larseggert: I _think_ `obtain_stream` will still return the stream even is it is in state `ResetRecvd`. `RecvStream::reset()` won't do anything with it though.

But good question if there is any case where `obtain_stream` would return `(None, None)` where we shouldn't close the connection.

larseggert: @mxinden see 795933b87f381a4ef6cae78008dbb4341cd6fe12

mxinden: Thanks for the additional test.

> I _think_ `obtain_stream` will still return the stream even is it is in state `ResetRecvd`.

Yes. But what if `cleanup_closed_streams` is called in between receiving the first and the second reset?

https://github.com/mozilla/neqo/blob/6575c551131b167a632be6b8db45b16fd16d50c7/neqo-transport/src/streams.rs#L333-L343

In that case `obtain_stream` should return `(None, None)`, right?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -582,6 +617,31 @@ fn simultaneous_stop_sending_and_reset() {
     );
 }
 
+#[test]
+//The client sends a duplicate stream reset.
```
mxinden, line 621:
```suggestion
//The client sends a duplicate stream reset.
#[test]
```

Seems to be the convention in the codebase.

See also https://github.com/rust-lang/style-team/issues/72. I couldn't find a clippy lint.

martinthomson: I use doc comments for tests.

Which would make this `///` instead.

neqo-transport/src/streams.rs
```diff
@@ -154,6 +158,8 @@ impl Streams {
                 stats.stream += 1;
                 if let (_, Some(rs)) = self.obtain_stream(*stream_id)? {
                     rs.inbound_stream_frame(*fin, *offset, data)?;
+                } else {
+                    return Err(Error::StreamStateError);
                 }
```
martinthomson, line 163: Are these better as let/else?

```suggestion
                let (_, Some(rs)) = self.obtain_stream(*stream_id)? else {
                    return Err(Error::StreamStateError);                    
                };
                rs.inbound_stream_frame(*fin, *offset, data)?;
```

martinthomson: Or...

```suggestion
                let rs = self.obtain_stream(*stream_id)?.1.ok_or(Error::StreamStateError)?;
                rs.inbound_stream_frame(*fin, *offset, data)?;
```

martinthomson (changes requested):
Looking at this more, I'm really not sure that this is safe.

There are cases where the frame is outright illegal.  But the code you routinely use here looks like this:

```rust
if let (_, Some(rs)) = self.obtain_stream(...)? {
  // happy path
} else {
  return Err(Error::WHATEVER);
}
```

The problem is that valid frames can arrive late.  As @mxinden points out, this results in the stream being removed.  The return value would then be `(None, None)`, which will hit the error case.  We'll kill connections that are completely valid in that case.

I think that you want a different test for whether the frame is valid than what you have described here.  My expectation is that obtain_stream() will be able to catch some cases (where flow control doesn't allow the stream to be created), but not others (like whether you want the receive stream where there shouldn't be one).

A more targeted function (or pair of functions) might do better because it could know that you want send or receive and abort if that part of the stream cannot exist.

That you have not hit this case suggests to me that we need a test that uses and retires streams fully, while sending a bunch of frames that are valid, but very late.

neqo-transport/src/streams.rs
larseggert: Need to check more, but inserting `server.streams.cleanup_closed_streams();` in between processing the two reset frames in `double_reset_stream` does **not** make the test fail.

larseggert:
@mxinden @martinthomson agree. Have been trying to craft a test that hits those error cases, but no luck so far.

neqo-transport/src/streams.rs
martinthomson: I believe that we won't drop a stream until we've had the RESET_STREAM frame acknowledged.  You have to do a back-and-forth to get that to happen.

larseggert: @martinthomson isn't the logic in question about processing at the side that _receives_ these stream control frames?

larseggert:
There now is a `double_stream_related_frame` test that at least hits the case of interest for stream reset. I need to make it hit the case for some of the other frames of interest.

larseggert:
OK, I think I have tests now that hits the new code paths for all frame types.

neqo-transport/src/streams.rs
```diff
@@ -131,6 +131,8 @@ impl Streams {
                 stats.reset_stream += 1;
                 if let (_, Some(rs)) = self.obtain_stream(*stream_id)? {
                     rs.reset(*application_error_code, *final_size)?;
+                } else if !self.ensure_existed_if_local(*stream_id) {
+                    return Err(Error::StreamStateError);
                 }
```
mxinden, line 136: The negation and the implicit `false` on remote streams, plus the repeated `return` is hard for me to parse.

What do you think of https://github.com/larseggert/neqo/pull/33?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -582,6 +618,103 @@ fn simultaneous_stop_sending_and_reset() {
...
 
+/// Make a stream data or control frame arrive after the stream has been used and cleared.
+fn late_stream_related_frame(frame_type: u64) {
+    let mut client = default_client();
+    let mut server = default_server();
+    connect(&mut client, &mut server);
+
+    // Client creates two streams and sends some data on the second.
+    _ = client.stream_create(StreamType::BiDi).unwrap();
+    let stream_id = client.stream_create(StreamType::BiDi).unwrap();
```
mxinden, line 629: Why two streams?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -582,6 +618,103 @@ fn simultaneous_stop_sending_and_reset() {
...
+            assert_eq!(after.max_stream_data, before.max_stream_data + 1);
+        }
+        FRAME_TYPE_STREAM_DATA_BLOCKED => {
+            assert_eq!(after.stream_data_blocked, before.stream_data_blocked + 1);
+        }
+        _ => panic!("unexpected frame type"),
+    }
+
+    // Now clear the streams on the client, and then deliver the test frame.
+    client.streams.clear_streams();
```
mxinden, line 686:
```suggestion
    client.streams.clear_streams();
    assert!(client.obtain_stream(stream_id).unwrap().is_none());
```

Maybe worth checking that the stream is really gone?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -582,6 +618,103 @@ fn simultaneous_stop_sending_and_reset() {
...
+
+#[test]
+fn late_max_stream_data_frame() {
+    late_stream_related_frame(FRAME_TYPE_MAX_STREAM_DATA);
+}
+
+#[test]
+fn late_stream_data_blocked_frame() {
+    late_stream_related_frame(FRAME_TYPE_STREAM_DATA_BLOCKED);
+}
```
mxinden, line 716: I would once again combine these into a single test function. Feel free to ignore.

``` rust
#[test]
fn xxx() {
  for frame in [FRAME_TYPE_RESET_STREAM, ...] {
    late_stream_related_frame(frame)
  }
}

**Review by mxinden — COMMENTED** [review]
Smaller comments. Overall  this looks good to me. Thanks for the test!

File: neqo-transport/src/streams.rs (hunk 5 above)
Comment (larseggert) [reply neqo-transport/src/streams.rs]: Like it! Merged.

File: neqo-transport/src/connection/tests/stream.rs (hunk 6 above)
Comment (larseggert) [reply neqo-transport/src/connection/tests/stream.rs:629]: Leftover from when I was debugging whether the`index` addition did the right thing. Can remove now.

File: neqo-transport/src/connection/tests/stream.rs (hunk 8 above)
Comment (larseggert) [reply neqo-transport/src/connection/tests/stream.rs:717]: Can do. I found separate tests easier to manage while I was debugging `late_stream_related_frame`.

File: neqo-transport/src/connection/tests/stream.rs [hunk 9]
```diff
@@ -539,6 +539,42 @@ fn do_not_accept_data_after_stop_sending() {
...
+        // Illegal RESET_STREAM frame
+        test_with_illegal_frame(&[FRAME_TYPE_RESET_STREAM, stream_id, 0, 0]);
+        // Illegal STOP_SENDING frame
+        test_with_illegal_frame(&[FRAME_TYPE_STOP_SENDING, stream_id, 0]);
+        // Illegal MAX_STREAM_DATA frame
+        test_with_illegal_frame(&[FRAME_TYPE_MAX_STREAM_DATA, stream_id, 0]);
+        // Illegal STREAM_DATA_BLOCKED frame
+        test_with_illegal_frame(&[FRAME_TYPE_STREAM_DATA_BLOCKED, stream_id, 0]);
+        // Illegal STREAM frame
+        test_with_illegal_frame(&[FRAME_TYPE_STREAM_CLIENT_INI_BIDI, stream_id, 0]);
```
martinthomson, line 574: HAXX:

```suggestion
        for frame_type in [FRAME_TYPE_RESET_STREAM, ...] {
            test_with_illegal_frame(&[frame_type, stream_id, 0, 0]);
        }
```

Observe how only `RESET_STREAM` differs from the others.  But the extra 0 byte on the others will be treated as a `PADDING` frame.

neqo-transport/src/streams.rs
larseggert: Except it doesn't work :-(

neqo-transport/src/connection/tests/stream.rs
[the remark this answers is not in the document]
larseggert: I had something like that before (there were some slight differences for some of the stream types). Then I figures that simply blowing them away via `clear_streams()` made things simpler.

neqo-transport/src/streams.rs
larseggert: The `!self.ensure_existed_if_local` condition is only sufficient _iff_ `obtain_stream` didn't match.

martinthomson (commented):
The checks are solid.  I'm not sure about the test.

neqo-transport/src/connection/tests/stream.rs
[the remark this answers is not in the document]
martinthomson: Ah, clear streams just removes them all.  I can handle that.  Make a comment to that effect - this is an internal API that I'd normally avoid in tests, but it makes sense here.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -539,6 +539,44 @@ fn do_not_accept_data_after_stop_sending() {
...
+        let mut server = default_server();
+        connect(&mut client, &mut server);
+        let dgram = send_with_extra(&mut server, IllegalWriter(frame.to_vec()), now());
+        client.process_input(dgram, now());
+        assert!(client.state().closed());
+    }
+
+    // 0 = Client-Initiated, Bidirectional; 2 = Client-Initiated, Unidirectional
+    for stream_id in [0, 2] {
+        // Illegal RESET_STREAM frame
```
martinthomson, line 565:
```suggestion
```

martinthomson (approved):

neqo-transport/src/streams.rs
larseggert: @mxinden see 19e7ed59d28341c50351a37534d33d7c01d8f7a8 for a different approach.

mxinden: Oh sorry for the trouble. New version looks good.

neqo-transport/src/stream_id.rs
```diff
@@ -104,6 +104,12 @@ impl StreamId {
         self.0 += 4;
     }
 
+    #[must_use]
+    /// Return the stream index for this stream ID.
```
mxinden, line 108:
```suggestion
    /// Return the stream index for this stream ID.
    #[must_use]
```

mxinden (approved):

Issue #2256, opened 2024-11-26
Expose ECN path validation failure reason in `neqo_transport::Stats`


mxinden:
Neqo now supports sending and acking ECN marks (main PR: https://github.com/mozilla/neqo/pull/1678/).

This is enabled in Firefox Nightly:

https://searchfox.org/mozilla-central/rev/dfaf02d68a7cb018b6cad7e189f450352e2cde04/modules/libpref/init/StaticPrefList.yaml#14072-14083

Paths are being validated for ECN capability:

https://github.com/mozilla/neqo/blob/5fe0c89dbe25b86b7c6aae279b181cfde63540ef/neqo-transport/src/ecn.rs#L210-L287

This is exposed as a metric:

https://github.com/mozilla/neqo/blob/5fe0c89dbe25b86b7c6aae279b181cfde63540ef/neqo-transport/src/stats.rs#L197-L201

https://searchfox.org/mozilla-central/rev/dfaf02d68a7cb018b6cad7e189f450352e2cde04/netwerk/socket/neqo_glue/src/lib.rs#386-404

https://glam.telemetry.mozilla.org/fog/probe/networking_http_3_ecn_path_capability/explore?aggKey=capable&aggType=avg&ping_type=*

Instead of only exposing whether a path is valid or not, I suggest we additionally expose the reason why a path failed ECN validation via `neqo_transport::Stats` and then in Firefox via a [Glean labeled counter](https://mozilla.github.io/glean/book/reference/metrics/labeled_counters.html).

PR #2270, opened 2024-12-07
feat(ecn): expose ECN path validation outcome in Stats

mxinden:
Expose the ECN path validation outcome via `neqo_transport::Stats`.

Closes https://github.com/mozilla/neqo/issues/2256.

### DIFF
--- a/neqo-transport/src/ecn.rs
+++ b/neqo-transport/src/ecn.rs
@@ -6,7 +6,7 @@
 
 use std::ops::{AddAssign, Deref, DerefMut, Sub};
 
-use enum_map::EnumMap;
+use enum_map::{Enum, EnumMap};
 use neqo_common::{qdebug, qinfo, qwarn, IpTosEcn};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


neqo-transport/src/ecn.rs
```diff
@@ -125,6 +127,36 @@ impl AddAssign<IpTosEcn> for EcnCount {
     }
 }
 
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Default)]
+pub struct EcnValidationCount(EnumMap<EcnValidationOutcome, u64>);
```
mxinden, line 131: Related: Following convention within this `mod` by prefixing every `pub` type with the module name, i.e. `EcnXXX`.

I prefer not doing this, but instead follow:

https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions

Which we currently disable:

https://github.com/mozilla/neqo/blob/a7581770ff138c16727e92ecb7b47d9ceb95425e/neqo-transport/src/lib.rs#L7

In case there is consensus, happy to do the rename in a follow-up pull request.

larseggert: I'm fine with removing the `allow` and do the required renames. (I'm in general for reducing clippy `allow`s.)

larseggert (approved):

PR #2280, opened 2024-12-11
ci: Enable MC/DC coverage

larseggert:
https://en.wikipedia.org/wiki/Code_coverage#Modified_condition/decision_coverage

The point of the PR is to make it easier to spot lines with partial coverage, which AFAICT currently show up as green.

MC/DC is finding some partially covered lines, so coverage drops. This is expected.


martinthomson (approved):
Do we need to set some new targets, or just accept a new lower threshold?

PR #2281, opened 2024-12-11
chore(qlog): Add meaningful qlog titles and descriptions.

mansf-osk:
Hey Hey! 👋 

First time neqo contributor here!
I'm Oskar, working student on the necko team, for those that don't already know me 😄 

@mxinden was showing me qvis the other day and we noticed that description and title were just showing something the likes of `Example qlog [...]`. With this change they now show `Neqo [client|server] qlog`. 

I also added a `server-` prefix to the filename of the server qlogs as I found it hard to understand which was which at first glance, but happy to take that out again if it's not wanted.

I hope these small change are somewhat useful and would appreciate a review and maybe a pointer as to what I could work on next. I'm definitely very interested to work more on neqo! 🙂 

Cheers
Oskar

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -508,8 +508,8 @@ fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<NeqoQlog> {
     NeqoQlog::enabled_with_file(
         qlog_dir,
         Role::Client,
-        Some("Example qlog".to_string()),
-        Some("Example qlog description".to_string()),
+        Some("Neqo client qlog".to_string()),
+        Some("Neqo client qlog".to_string()),
         format!("{hostname}-{cid}"),
     )
     .map_err(Error::QlogError)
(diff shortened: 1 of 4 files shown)


neqo-bin/src/client/mod.rs
```diff
@@ -508,8 +508,8 @@ fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<NeqoQlog> {
     NeqoQlog::enabled_with_file(
         qlog_dir,
         Role::Client,
-        Some("Example qlog".to_string()),
-        Some("Example qlog description".to_string()),
+        Some("Neqo client qlog".to_string()),
+        Some("Neqo client qlog".to_string()),
         format!("{hostname}-{cid}"),
```
mxinden, line 513:
```suggestion
        format!("client-{hostname}-{cid}"),
```

I suggest doing the same for the client, i.e. prefixing it with `client-`.

mxinden (approved):
Thank you!

Apart from the one suggestion, this looks good to me.

Let's give others a bit of time to review as well before merging.

neqo-bin/src/client/mod.rs
mansf-osk: Makes sense, I added it in 56cd6cf!

larseggert (approved):

mxinden:
@mansf-osk unit tests are failing. It should be as simple as updating the constant below:

https://github.com/mozilla/neqo/blob/fd42bed9206a0cd980be8d30b7c1ac2797acffa0/test-fixture/src/lib.rs#L419-L423

You can run them locally via `cargo test`.

mxinden:
Failure on Windows installing Rust seems unrelated. Retriggered.

PR #2282, opened 2024-12-12
ci: Use Codecov JSON format and include FFI

larseggert:
Documentation isn't super clear, but the JSON may contain more detailed info vs. the LCOV format.


mxinden (approved):
Not familiar with it. Seems low risk.

PR #2286, opened 2024-12-17
feat: Create `PushId` type

### LINKED ISSUE #785: Create PushId type

larseggert:
Fixes #785

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -16,7 +16,7 @@ use crate::{
     connection::Http3State,
     features::extended_connect::{ExtendedConnectEvents, ExtendedConnectType, SessionCloseReason},
     settings::HSettingType,
-    CloseType, Http3StreamInfo, HttpRecvStreamEvents, RecvStreamEvents, SendStreamEvents,
+    CloseType, Http3StreamInfo, HttpRecvStreamEvents, PushId, RecvStreamEvents, SendStreamEvents,
 };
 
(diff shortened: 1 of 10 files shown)


neqo-http3/src/client_events.rs
```diff
@@ -240,20 +240,20 @@ impl ExtendedConnectEvents for Http3ClientEvents {
 }
 
 impl Http3ClientEvents {
-    pub fn push_promise(&self, push_id: u64, request_stream_id: StreamId, headers: Vec<Header>) {
+    pub fn push_promise(&self, push_id: PushId, request_stream_id: StreamId, headers: Vec<Header>) {
```
mxinden, line 243: For completeness sake, we could use `push_id: impl Into<PushId>` everywhere. That said, I prefer the current approach, making the conversion explicit at the caller side.

mxinden (approved):

neqo-http3/src/client_events.rs
larseggert: Ah, I didn't think of this. Anyone got any preference? @martinthomson?

martinthomson: I agree with Max.  We can use `Into<Foo>` for public interfaces or where the conversion cost is starting to feel like a burden.  AND where the type conversion is not part of the design.  Here, there are almost no reasons to provide a `PushId` as a raw number because you get the values you use from the stack, so in most cases you will be feeding a `PushId` that you received from an event, not doing arithmetic to make one.  So an opaque type everywhere makes the most sense.

PR #2288, opened 2024-12-17
chore: Remove draft QUIC versions

larseggert:
For discussion.

-29 is used in a bunch of tests, so kept it in for now, but put it behind a feature gate.


martinthomson (approved):

mxinden (approved):
Fine by me.

I assume the fraction of QUIC < V1 is small? As far as I know, we don't have any telemetry on it.

For my understanding, the only reason the QUIC interop runner is still depending on draft-29 is that we have to put in the work to transition the tests and the various test implementations, is that correct?

PR #2293, opened 2024-12-19
fix: Avoid spurious PMTUD resets

larseggert:
Previously, after PMTUD had completed, we could
end up restarting PMTUD when packet loss counters
for packets larger than the current PMTU exceeded
the limit. We're now making sure to not do that.


neqo-transport/src/pmtud.rs
```diff
@@ -661,30 +711,46 @@ mod tests {
...
-        let expected_lc = search_table_zero(&pmtud, &pmtud.loss_counts, 8000);
+        expected_lc = search_table_zero(&pmtud, &pmtud.loss_counts, 8000);
         assert_eq!(expected_lc, pmtud.loss_counts);
 
         // Now, one more packets of size 9000 was lost, which should increase loss counts >= 9000
-        // by one. There have now been MAX_PROBES losses of packets >= 8191, so the PMTUD process
-        // should have restarted.
+        // by one. There have now been MAX_PROBES losses of packets >= 8191, but that is larger than
+        // the current MTU, so nothing will happen.
         pmtud.on_packets_lost(&[make_sentpacket(0, now, 9000)], &mut stats, now);
```
martinthomson, line 737: Isn't this pure contrivance?  We don't probe in this odd pattern.

neqo-transport/src/pmtud.rs
```diff
@@ -604,31 +623,50 @@ mod tests {
...
+        expected_lc = search_table_inc(&pmtud, &expected_lc, 5000);
         pmtud.on_packets_lost(&[make_sentpacket(0, now, 5000)], &mut stats, now);
-        assert_pmtud_restarted(&pmtud);
-        expected_lc.fill(0); // Reset the expected loss counts.
+        assert_eq!(expected_lc, pmtud.loss_counts);
 
-        // Two packets of size 4000 were lost, which should increase loss counts >= 4000 by two.
-        let expected_lc = search_table_inc(&pmtud, &expected_lc, 4000);
-        let expected_lc = search_table_inc(&pmtud, &expected_lc, 4000);
+        // Two packets of size 4000 were lost, which should increase loss counts >= 4000 by one.
```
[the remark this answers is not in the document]
larseggert: No, the comment is correct. We do only increase by one of several packets of the same size are declared lost on the same ACK. Otherwise each cwnd overshoot would make us reprobe the MTU.

neqo-transport/src/pmtud.rs
larseggert: Yes, this is made up to test corner cases.

neqo-transport/src/pmtud.rs
```diff
@@ -267,27 +266,35 @@ impl Pmtud {
...
+                    // conservative and start from scratch, but we don't strictly need to do that.
+                    self.reset(stats);
+                    qdebug!("PMTUD reset and restarting, PLPMTU is now {}", self.mtu);
+                    self.start(now, stats);
+                }
+            }
+            Probe::Needed | Probe::Sent => {
+                // We saw multiple losses of packets > the current MTU during PMTU discovery, so
+                // we're done.
+                if largest_ok_mtu > self.mtu {
```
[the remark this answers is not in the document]
larseggert: You're right, and the tests were actually failing.

neqo-transport/src/path.rs
```diff
@@ -524,12 +524,18 @@ impl Path {
...
+                    ip = remote.ip()
+                );
                 stats.pmtud_iface_mtu = mtu;
                 Some(mtu)
             }
             Err(e) => {
-                qwarn!("Failed to determine outbound interface: {e}");
+                qwarn!(
+                    "Failed to determine outbound interface for destination {}: {e}",
+                    remote.ip()
```
mxinden, line 537:
```suggestion
                    "Failed to determine outbound interface for destination {ip}: {e}",
                    ip = remote.ip()
```

For the sake of consistency with the `qwarn` above. Or remove the reassignment above.

larseggert:
@codecov-ai-reviewer review

neqo-transport/src/pmtud.rs
```diff
@@ -267,27 +266,35 @@ impl Pmtud {
             return;
         };
 
-        let last_ok = first_failed - 1;
-        qdebug!(
-            "Packet of size > {} lost >= {} times",
-            self.search_table[last_ok],
-            MAX_PROBES
+        let largest_ok_idx = first_failed - 1;
+        let largest_ok_mtu = self.search_table[largest_ok_idx];
```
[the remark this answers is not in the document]
larseggert: I've added a test for this.

neqo-transport/src/pmtud.rs
```diff
@@ -406,103 +417,55 @@ mod tests {
...
-        find_pmtu(V4, u16::MAX.into());
-    }
-
-    #[test]
-    fn pmtud_v6_max() {
-        find_pmtu(V6, u16::MAX.into());
-    }
+        let final_mtu = iface_mtu.map_or(mtu, |iface_mtu| min(mtu, iface_mtu));
+        assert_mtu(&pmtud, final_mtu);
+        // assert_mtu(&pmtud, final_mtu);
```
mxinden, line 445: Can be removed?

neqo-transport/src/pmtud.rs
```diff
@@ -660,30 +672,46 @@ mod tests {
...
+
+        // One packet of size 1400 was lost, which should increase loss counts >= 1400 by one.
+        expected_lc = search_table_inc(&pmtud, &pmtud.loss_counts, 1400);
+        pmtud.on_packets_lost(&[make_sentpacket(0, now, 1400)], &mut stats, now);
+        assert_eq!(expected_lc, pmtud.loss_counts);
+
+        // One packet of size 1400 was lost, which should increase loss counts >= 1400 by one.
+        expected_lc = search_table_inc(&pmtud, &pmtud.loss_counts, 1400);
+        pmtud.on_packets_lost(&[make_sentpacket(0, now, 1400)], &mut stats, now);
+        assert_eq!(expected_lc, pmtud.loss_counts);
```
mxinden, line 708:
```suggestion
        for _ in 0..2 {
          // One packet of size 1400 was lost, which should increase loss counts >= 1400 by one.
          expected_lc = search_table_inc(&pmtud, &pmtud.loss_counts, 1400);
          pmtud.on_packets_lost(&[make_sentpacket(0, now, 1400)], &mut stats, now);
          assert_eq!(expected_lc, pmtud.loss_counts);
        }
```

Might make it a bit easier to maintain.

mxinden (approved):

PR #2294, opened 2024-12-19
chore: Remove some `unwrap`s and replace others by `expect`

larseggert:
In functions that return `Result` or `Option` or where it is easy to make them do so w/o too many other code changes.

This lets us enable the clippy `unwrap_used` check, which is a nice reminder not to be casual with `unwrap`.


neqo-transport/src/ackrate.rs
```diff
@@ -45,8 +45,8 @@ impl AckRate {
         builder.write_varint_frame(&[
             FRAME_TYPE_ACK_FREQUENCY,
             seqno,
-            u64::try_from(self.packets + 1).unwrap(),
-            u64::try_from(self.delay.as_micros()).unwrap(),
+            u64::try_from(self.packets + 1).expect("usize fits in u64"),
+            u64::try_from(self.delay.as_micros()).expect("delay as usec fits in u64"),
```
martinthomson, line 49:
```suggestion
            u64::try_from(self.delay.as_micros()).unwrap_or(u64::MAX),
```

neqo-transport/src/ackrate.rs
```diff
@@ -45,8 +45,8 @@ impl AckRate {
         builder.write_varint_frame(&[
             FRAME_TYPE_ACK_FREQUENCY,
             seqno,
-            u64::try_from(self.packets + 1).unwrap(),
-            u64::try_from(self.delay.as_micros()).unwrap(),
+            u64::try_from(self.packets + 1).expect("usize fits in u64"),
```
martinthomson, line 48: This is a bit silly.  We can't send that many packets.

neqo-transport/src/connection/mod.rs
```diff
@@ -627,7 +627,7 @@ impl Connection {
                 self.version,
                 u64::try_from(rtt.as_millis()).unwrap_or(0),
             )
-            .unwrap()
+            .expect("can create resumption token")
```
martinthomson, line 630:
```suggestion
            .expect("caller checked if a resumption token existed")
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2655,7 +2659,7 @@ impl Connection {
             let reset_token = remote
                 .get_bytes(tparams::STATELESS_RESET_TOKEN)
                 .map_or_else(ConnectionIdEntry::random_srt, |token| {
-                    <[u8; 16]>::try_from(token).unwrap()
+                    <[u8; 16]>::try_from(token).unwrap_or_else(|_| ConnectionIdEntry::random_srt())
                 });
```
martinthomson, line 2663: Isn't this a case for:

```rust
            let reset_token = remote
                .get_bytes(tparams::STATELESS_RESET_TOKEN)
                .and_then(|token| <[u8; 16]>::try_from(token))
                .unwrap_or_else(ConnectionIdEntry::random_srt)
```

neqo-transport/src/connection/mod.rs
```diff
@@ -3124,7 +3128,8 @@ impl Connection {
         self.tps.borrow().remote.as_ref().map_or_else(
             || Duration::new(0, 0),
```
martinthomson, line 3129:
```suggestion
        self.tps.borrow().remote.as_ref().map_or_default(
```

neqo-transport/src/connection/mod.rs
```diff
@@ -3124,7 +3128,8 @@ impl Connection {
...
             |r| {
-                let exponent = u32::try_from(r.get_integer(tparams::ACK_DELAY_EXPONENT)).unwrap();
+                let exponent = u32::try_from(r.get_integer(tparams::ACK_DELAY_EXPONENT))
+                    .expect("ACK_DELAY_EXPONENT > 20 is invalid per RFC9000");
                 Duration::from_micros(v.checked_shl(exponent).unwrap_or(u64::MAX))
```
martinthomson, line 3133: Note that this `checked_shl` is not right.  The checking only ensures that the shift value is within range (i.e., between 0 and 64).  It doesn't catch overflow.  Given that `v` could be up to 2<sup>62</sup>-1, there's a real overflow risk.

Perhaps the right answer here is:

```rust
let exponent = u32::try_from(r.get_integer(tparams::ACK_DELAY_EXPONENT)).expect("your message");
let corrected = if v.leading_zeros() >= exponent {
  v << exponent
} else {
  u64::MAX
};
Duration::from_micros(corrected)
```

neqo-transport/src/crypto.rs
```diff
@@ -1284,7 +1285,8 @@ impl CryptoStates {
         }
     }
 
-    #[cfg(all(not(feature = "disable-encryption"), test))]
+    #[cfg(not(feature = "disable-encryption"))]
+    #[cfg(test)]
```
martinthomson, line 1289: Any reason for this change?

neqo-transport/src/crypto.rs
```diff
@@ -1520,34 +1522,36 @@ impl CryptoStreams {
...
-            let written = if offset == 0 {
-                if let Some(sni) = find_sni(data) {
-                    // Cut the crypto data in two at the midpoint of the SNI and swap the chunks.
-                    let mid = sni.start + (sni.end - sni.start) / 2;
-                    let (left, right) = data.split_at(mid);
-                    [
-                        write_chunk(offset + mid as u64, right, builder),
-                        write_chunk(offset, left, builder),
-                    ]
+        if let Some(cs) = self.get_mut(space) {
```
martinthomson, line 1525: I think that this would be better with:

```rust
let Some(cs) = self.get_mut(space) else { return; }
```

neqo-transport/src/path.rs
```diff
@@ -331,8 +330,10 @@ impl Paths {
                     .as_ref()
                     .is_some_and(|target| Rc::ptr_eq(target, p))
                 {
-                    let primary = self.migration_target.take();
-                    mem::drop(self.select_primary(&primary.unwrap(), now));
+                    let Some(primary) = self.migration_target.take() else {
+                        break;
```
martinthomson, line 334: I think that you can restructure this little piece of code to avoid the extra break.

```rust
let Some(primary) = self.migration_target.as_mut() else { break; }
if Rc::ptr_eq(target, primary) {
  let p = primary.take();
  mem::drop(self.select_primary(&p, now));
  return true;
}
break;
```

Once we're on 1.80 MSRV, we can use `Option::take_if` and simplify further.

```rust
if let Some(primary) = self.migration_target.take_if(|p| Rc::ptr_eq(target, p)) 
  mem::drop(self.select_primary(&primary, now));
  return true;
}
break;
```

neqo-transport/src/path.rs
```diff
@@ -852,10 +851,11 @@ impl Path {
             true
         } else if matches!(self.state, ProbeState::Valid) {
             // Retire validated, non-primary paths.
-            // Allow more than 2* `MAX_PATH_PROBES` times the PTO so that an old
+            // Allow more than 2 * `MAX_PATH_PROBES` times the PTO so that an old
```
martinthomson, line 854:
```suggestion
            // Allow more than `2 * MAX_PATH_PROBES` times the PTO so that an old
```

neqo-transport/src/recovery/mod.rs
```diff
@@ -658,41 +642,40 @@ impl LossRecovery {
...
         let cleanup_delay = self.pto_period(primary_path.borrow().rtt());
-        let mut lost = Vec::new();
-        self.spaces.get_mut(pn_space).unwrap().detect_lost_packets(
-            now,
-            loss_delay,
-            cleanup_delay,
-            &mut lost,
-        );
-        self.stats.borrow_mut().lost += lost.len();
+        if let Some(sp) = self.spaces.get_mut(pn_space) {
```
martinthomson, line 646: This is another great candidate for `let Some(foo) = expr else { return ...; };`

neqo-transport/src/tparams.rs
```diff
@@ -457,7 +457,8 @@ impl TransportParameters {
         let rbuf = random::<4>();
         let mut other = Vec::with_capacity(versions.all().len() + 1);
         let mut dec = Decoder::new(&rbuf);
-        let grease = dec.decode_uint::<u32>().unwrap() & 0xf0f0_f0f0 | 0x0a0a_0a0a;
+        let grease =
+            dec.decode_uint::<u32>().expect("dec has u32 left") & 0xf0f0_f0f0 | 0x0a0a_0a0a;
```
martinthomson, line 461: This is a good place to go with `u32::from_ne_bytes(random::<4>)`, like above.

martinthomson (commented):
Lots of improvements here, but there are a few places where some of that could be improved.

(Non-blocking comment as I'm going on leave.)

neqo-crypto/tests/selfencrypt.rs
```diff
@@ -22,6 +22,7 @@ fn se_create() {
 const PLAINTEXT: &[u8] = b"PLAINTEXT";
 const AAD: &[u8] = b"AAD";
 
+#[cfg(test)]
```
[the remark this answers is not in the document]
larseggert: I don't know, but it does. Otherwise clippers complains.

neqo-http3/tests/httpconn.rs
```diff
@@ -45,6 +45,7 @@ fn receive_request(server: &Http3Server) -> Option<Http3OrWebTransportStream> {
     None
 }
 
+#[cfg(test)]
```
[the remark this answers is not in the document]
larseggert: See above.

neqo-transport/src/connection/mod.rs
larseggert: There is no `map_or_default`...

neqo-transport/src/crypto.rs
larseggert: Same as above, we need an explicit `#[cfg(test)]` here for clippy.

neqo-transport/src/ackrate.rs
larseggert: This code will be touched again when we update to the latest ACK-frequency draft. Leaving as-is for now.

neqo-transport/src/path.rs
larseggert: Couldn't easily get this to work, opened #2309.

larseggert:
@mxinden when you have a minute, please re-review? (Not urgent.)

neqo-crypto/src/agent.rs
```diff
@@ -158,12 +160,11 @@ impl SecretAgentPreInfo {
...
 
-    /// # Panics
+    /// # Errors
     ///
     /// If `usize` is less than 32 bits and the value is too large.
-    #[must_use]
-    pub fn max_early_data(&self) -> usize {
-        usize::try_from(self.info.maxEarlyDataSize).unwrap()
+    pub fn max_early_data(&self) -> Res<usize> {
+        usize::try_from(self.info.maxEarlyDataSize).map_err(|_| Error::InternalError)
```
martinthomson, line 167: I thought that we had a From impl for this error type.

neqo-crypto/src/agent.rs
```diff
@@ -893,8 +892,12 @@ impl Client {
             // Ignore the token.
             return ssl::SECSuccess;
         }
-        let resumption = arg.cast::<Vec<ResumptionToken>>().as_mut().unwrap();
-        let len = usize::try_from(len).unwrap();
+        let Some(resumption) = arg.cast::<Vec<ResumptionToken>>().as_mut() else {
+            return ssl::SECFailure;
+        };
+        let Ok(len) = usize::try_from(len) else {
+            return ssl::SECFailure;
```
martinthomson, line 899: When can we starting using Try?

neqo-http3/tests/webtransport.rs
```diff
@@ -82,6 +82,7 @@ fn exchange_packets(client: &mut Http3Client, server: &mut Http3Server) {
     }
 }
 
+#[cfg(test)]
```
martinthomson, line 85: If you put `#![cfg(test)]` at the top of the file, does it stop complaining?

neqo-transport/src/connection/mod.rs
```diff
@@ -2625,7 +2629,7 @@ impl Connection {
             let reset_token = remote
                 .get_bytes(tparams::STATELESS_RESET_TOKEN)
                 .map_or_else(ConnectionIdEntry::random_srt, |token| {
-                    <[u8; 16]>::try_from(token).unwrap()
+                    <[u8; 16]>::try_from(token).unwrap_or_else(|_| ConnectionIdEntry::random_srt())
```
martinthomson, line 2632: This seems like a mistake.  Of anything, I would return `Err(Error::TransportErrorParameter)`.

neqo-transport/src/connection/mod.rs
```diff
@@ -3074,14 +3078,23 @@ impl Connection {
...
         // If we have remote transport parameters, use them.
         // Otherwise, ack delay should be zero (because it's the handshake).
         self.tps.borrow().remote.as_ref().map_or_else(
-            || Duration::new(0, 0),
+            || Ok(Duration::default()),
             |r| {
-                let exponent = u32::try_from(r.get_integer(tparams::ACK_DELAY_EXPONENT)).unwrap();
-                Duration::from_micros(v.checked_shl(exponent).unwrap_or(u64::MAX))
+                let exponent = u32::try_from(r.get_integer(tparams::ACK_DELAY_EXPONENT))?;
+                if exponent > 20 {
```
martinthomson, line 3088: Might need a test for this.  Or, given that we have already checked that this is not > 20, an unwrap is OK.

neqo-transport/src/connection/mod.rs
```diff
@@ -3074,14 +3078,23 @@ impl Connection {
...
+                    // ACK_DELAY_EXPONENT > 20 is invalid per RFC9000
+                    return Err(Error::TransportParameterError);
+                }
+                let corrected = if v.leading_zeros() >= exponent {
+                    v << exponent
+                } else {
+                    u64::MAX
+                };
```
martinthomson, line 3096: This seems good.  Should we have a saturating shift function?  This isn't the first time we've needed this sort of thing.

martinthomson (approved):

neqo-crypto/src/agent.rs
larseggert: You mean https://doc.rust-lang.org/std/ops/trait.Try.html?

mxinden: Alternative:

```
let len = usize_try_from(len).map_err(|_| ssl::SECFailure)?;
```

Not sure what clippy prefers.

neqo-transport/src/connection/mod.rs
mxinden: Maybe long-term we can use:

https://doc.rust-lang.org/beta/std/num/struct.Saturating.html#method.rotate_left

larseggert: New issue: https://github.com/mozilla/neqo/issues/2402

neqo-crypto/src/agent.rs
```diff
@@ -122,7 +124,7 @@ macro_rules! preinfo_arg {
         pub fn $v(&self) -> Option<$t> {
             match self.info.valuesSet & ssl::$m {
                 0 => None,
-                _ => Some(<$t>::try_from(self.info.$f).unwrap()),
+                _ => Some(<$t>::try_from(self.info.$f).ok()?),
```
mxinden, line 127: Not familiar with the code. Silently drops the `None` case. How about adding a `qdebug!` via [`Result::inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err)?

neqo-transport/src/recovery/mod.rs
```diff
@@ -515,30 +515,21 @@ impl LossRecovery {
...
+                return Vec::new();
+            }
+            let mut dropped = sp.remove_ignored().collect::<Vec<_>>();
+            let mut path = primary_path.borrow_mut();
+            for p in &mut dropped {
+                path.discard_packet(p, now, &mut self.stats.borrow_mut());
+            }
+            dropped
+        } else {
+            Vec::new()
```
mxinden, line 531: How about an early `return Vec::new()`, thus flattening this function a bit.

neqo-transport/src/send_stream.rs
```diff
@@ -741,7 +744,7 @@ impl SendStream {
             sendorder: None,
             bytes_sent: 0,
             fair: false,
-            writable_event_low_watermark: 1.try_into().unwrap(),
+            writable_event_low_watermark: NonZeroUsize::new(1).expect("1 != 0"),
```
mxinden, line 747:
```suggestion
            writable_event_low_watermark: NonZeroUsize::MIN,
```

Feel free to ignore.

https://doc.rust-lang.org/std/num/struct.NonZero.html#associatedconstant.MIN

mxinden (approved):

neqo-transport/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#![allow(clippy::module_name_repetitions)] // This lint doesn't work here. Change to `#![expect]` when MSRV is >= 1.81.
```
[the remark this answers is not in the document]
larseggert: This came in as part of https://github.com/mozilla/neqo/commit/808b01831902a8fad4576518f5c0f2e05fbe9680. I just added the comment (because we say the same elsewhere and so I probably had it in my clipboard buffer.)

Issue #1912, opened 2024-05-23
Update CUBIC to RFC 9438


larseggert:
We are currently implementing RFC 8312.

larseggert:
As part of this, we should also look into implementing [HyStart](https://datatracker.ietf.org/doc/html/rfc9406).

larseggert:
This one might be an alternative to HyStart: https://datatracker.ietf.org/doc/draft-chung-ccwg-search/

omansfeld:
Scoped the update in #2535 and will keep incrementing on the update in that draft PR.

PR #2295, opened 2024-12-21
feat: use Cubic CC by default

mxinden:
Firefox uses Cubic by default:

``` yaml
- name: network.http.http3.cc_algorithm
  type: RelaxedAtomicUint32
  value: 1
  mirror: always
  rust: true
```

https://searchfox.org/mozilla-central/rev/f9517009d8a4946dbdd3acd72a31dc34fca79586/modules/libpref/init/StaticPrefList.yaml

This commit updates Neqo to use Cubic instead of New Reno by default.


mxinden:
Cubic does not seem to grow the cwnd as fast as New Reno in congestion avoidance phase. More particularly, while our New Reno implementation increases its cwnd in each iteration, our Cubic implementation only does so in every second iteration.

https://github.com/mozilla/neqo/blob/922d266bba926532982ee3fa91574ae131d55248/neqo-transport/src/connection/tests/cc.rs#L286-L287

That is surprising to me. I expected Cubic to grow faster than New Reno in the beginning, given its concave increase early in congestion avoidance.

I will give this more thought.

larseggert:
> Cubic does not seem to grow the cwnd as fast as New Reno in congestion avoidance phase.

That seems like a bug: https://datatracker.ietf.org/doc/html/rfc9438#section-1-5

> Specifically, CUBIC may increase the congestion window more aggressively than Reno during the congestion avoidance phase. According to [[RFC5681](https://datatracker.ietf.org/doc/html/rfc5681)], during congestion avoidance, the sender must not increment cwnd by more than Sender Maximum Segment Size (SMSS) bytes once per round-trip time (RTT), whereas CUBIC may increase cwnd much more aggressively.

mxinden:
My assessment above is wrong for the following reason:

> Cubic does not seem to grow the cwnd as fast as New Reno in congestion avoidance phase.

Wrong.

> More particularly, while our New Reno implementation increases its cwnd in each iteration by the SMSS, our Cubic implementation only does so in every second iteration.

Correct.

BUT, while New Reno halves its congestion window on a congestion event:

https://github.com/mozilla/neqo/blob/9e5a622941452115a001ad4faac365f929280a97/neqo-transport/src/cc/new_reno.rs#L45

Cubic will reduce it by 30% only:

https://github.com/mozilla/neqo/blob/9e5a622941452115a001ad4faac365f929280a97/neqo-transport/src/cc/cubic.rs#L23-L25

https://github.com/mozilla/neqo/blob/9e5a622941452115a001ad4faac365f929280a97/neqo-transport/src/cc/cubic.rs#L206-L207

Thus, within the 5 iterations of the test, Cubic does not grow its congestion window as fast as New Reno, because it starts off with a larger congestion window after the congestion event.

mxinden:
> Aside: Is there any point at having `CUBIC_BETA_USIZE_DIVIDEND` and `CUBIC_BETA_USIZE_DIVISOR` vs. just `CUBIC_BETA = 0.7`? Are we ever using just one of them without the division?

I assume the split is due to its usage with the `usize`s `curr_cwnd` and `acked_bytes`. The split in dividend and divisor allows for integer multiplication of a `usize` with a fraction without a conversion from/to a floating point.

https://github.com/mozilla/neqo/blob/9e5a622941452115a001ad4faac365f929280a97/neqo-transport/src/cc/cubic.rs#L206-L207

I don't have an intuition whether it is worth the complexity it introduces. Given our work ahead (https://github.com/mozilla/neqo/issues/1912) I suggest we keep it as is for now.

neqo-transport/src/connection/tests/cc.rs
```diff
@@ -211,12 +210,11 @@ fn single_packet_on_recovery() {
     assert!(dgram.is_some());
 }
 
-#[test]
 /// Verify that CC moves out of recovery period when packet sent after start
 /// of recovery period is acked.
-fn cc_cong_avoidance_recovery_period_to_cong_avoidance() {
-    let mut client = default_client();
-    let mut server = default_server();
+fn cc_cong_avoidance_recovery_period_to_cong_avoidance(cc_algorithm: CongestionControlAlgorithm) {
```
mxinden, line 215: This is the only test in `cc.rs` that specifically depends on the underlying `CongestionControlAlgorithm` used. The other tests focus on the high level `ClassicCongestionControl`.

For the above reason, I think it is worth running this unit test with both `NewReno` and `Cubic`.

Previously this test would assert on the concrete individual `NewReno` window increases. That is difficult to do when supporting two `CongestionControlAlgorithm`s.

Instead I suggest simply asserting that after x rounds, the congestion controller reaches the congestion window before the loss event again. Thus it switched from `Recovery` to `CongestionAvoidance`.

larseggert (approved):

mxinden:
Curious what the benchmark results will be.

Issue #2279, opened 2024-12-10
perf(udp): enable multi-message IO on apple


mxinden:
We currently pass a single large read buffer to the OS across all platforms:

https://github.com/mozilla/neqo/blob/3e6526177e69ae9ef00fd3af465563d4f8a40e02/neqo-udp/src/lib.rs#L66-L70

Linux, Android and Windows can leverage GRO (and URO) to read multiple datagrams into the buffer.

Apple platforms do not support GRO. But instead, we can fall-back to the older `recvmmsg` style syscalls.

@larseggert  added `recvmmsg` style reads on apple to quinn-udp: https://github.com/quinn-rs/quinn/pull/1993

In order to make use of this optimization on apple platforms, we should be passing multiple smaller read buffers to the apple OS, instead of one large buffer (see code above).

I suggest the following:

1. Introduce a `struct ReceiveBuffer(Vec<u8>)` wrapper type with a `new` function using [`RECV_BUF_SIZE`](https://github.com/mozilla/neqo/blob/3e6526177e69ae9ef00fd3af465563d4f8a40e02/neqo-udp/src/lib.rs#L23).
2. Have the user pass a long-lived instance of `ReceiveBuffer` to [`recv_inner`](https://github.com/mozilla/neqo/blob/3e6526177e69ae9ef00fd3af465563d4f8a40e02/neqo-udp/src/lib.rs#L55).
3. Within `recv_inner` we can assert the size of `receive_buffer.0`, given that a user can construct it via `new` only. On Linux, Android, Windows, ... proceed as before, passing the single large buffer to the OS. On Apple, partition the buffer and pass multiple slices of it to the OS.
4. On apple platforms, enable the `fast_apple` `quinn-udp` feature.

Alternatively on Apple platforms, we could change `ReceiveBuffer(Vec<u8>)` to `ReceiveBuffer(Vec<Vec<u8>>)`.

Overarching tracking issue: https://github.com/mozilla/neqo/issues/1693

PR #2302, opened 2024-12-23
perf(udp): multi-message receive on apple

mxinden:
Do multi-message receive via https://github.com/quinn-rs/quinn/pull/1993 on apple platforms.

---

Localhost benchmark (_Download_) with PMTUD disabled shows ~11% performance improvement.

```
$ critcmp main mmsg                                              
group                                         main                                   mmsg
-----                                         ----                                   ----
1-conn/1-100mb-resp (aka. Download)/client    1.11   790.5±12.51ms   126.5 MB/sec    1.00   711.5±31.65ms   140.5 MB/sec
```

Fixes https://github.com/mozilla/neqo/issues/2279.

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -7,20 +7,66 @@
 #![allow(clippy::missing_errors_doc)] // Functions simply delegate to tokio and quinn-udp.
 
 use std::{
+    array,
     io::{self, IoSliceMut},
+    iter,
     net::SocketAddr,
     slice::{self, Chunks},
 };
 
+use log::{log_enabled, Level};
 use neqo_common::{qdebug, qtrace, Datagram, IpTos};
 use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
 
-/// Socket receive buffer size.
+/// Receive buffer size
 ///
-/// Allows reading multiple datagrams in a single [`Socket::recv`] call.
-//
-// TODO: Experiment with different values across platforms.
-pub const RECV_BUF_SIZE: usize = u16::MAX as usize;
+/// Fits a maximum size UDP datagram, or, on platforms with segmentation
+/// offloading, multiple smaller datagrams.
+const RECV_BUF_SIZE: usize = u16::MAX as usize;
+
+/// The number of buffers to pass to the OS on [`Socket::recv`].
+///
+/// Platforms without segmentation offloading, i.e. platforms not able to read
+/// multiple datagrams into a single buffer, can benefit from using multiple
+/// buffers instead.
+///
+/// Platforms with segmentation offloading don't benefit from additionally using
+/// multiple buffers.
+///
+/// - Linux/Android: use segmentation offloading via GRO
+/// - Windows: use segmentation offloading via URO (caveat see <https://github.com/quinn-rs/quinn/issues/2041>)
+/// - Apple: no segmentation offloading available, use multiple buffers
+#[cfg(not(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "visionos"
+)))]
+const NUM_BUFS: usize = 1;
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "visionos"
+))]
+// Value approximated based on neqo-bin "Download" benchmark only.
+const NUM_BUFS: usize = 16;
+
+/// A UDP receive buffer.
+pub struct RecvBuf(Vec<Vec<u8>>);
+
+impl RecvBuf {
+    #[must_use]
+    pub fn new() -> Self {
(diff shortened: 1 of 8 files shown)


neqo-udp/src/lib.rs
```diff
@@ -20,7 +20,29 @@ use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
...
+    target_os = "visionos"
+)))]
+const BATCH_SIZE: usize = 1;
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "visionos"
+))]
+const BATCH_SIZE: usize = 32;
```
martinthomson, line 37: Needs some commentary about why there is a split and what the benefits of the additional allocation are.  2MB is fine, but knowing why is important.

mxinden: Added rational for the platform specific values. Unfortunately we don't have good data, thus the concrete values are based on local testing only. Please take another look.

neqo-udp/src/lib.rs
```diff
@@ -7,20 +7,66 @@
...
+const RECV_BUF_SIZE: usize = u16::MAX as usize;
+
+/// The number of buffers to pass to the OS on [`Socket::recv`].
+///
+/// Platforms without segmentation offloading, i.e. platforms not able to read
+/// multiple datagrams into a single buffer, can benefit from using multiple
+/// buffers instead.
+///
+/// Platforms with segmentation offloading don't benefit from additionally using
+/// multiple buffers.
```
larseggert, line 34: Is this actually true? At least for G**S**O, [Cloudflare seems to say there are additional benefits](https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/).

neqo-udp/src/lib.rs
```diff
@@ -7,20 +7,66 @@
...
+///
+/// - Linux/Android: use segmentation offloading via GRO
+/// - Windows: use segmentation offloading via URO (caveat see <https://github.com/quinn-rs/quinn/issues/2041>)
+/// - Apple: no segmentation offloading available, use multiple buffers
+#[cfg(not(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "visionos"
+)))]
```
larseggert, line 44: Wonder if we should start using `cfg_aliases`, similar to what `quinn` does: https://github.com/quinn-rs/quinn/blob/main/quinn-udp/build.rs

larseggert (approved):

neqo-udp/src/lib.rs
mxinden: I should have written that the marginal benefits have so far not been measurable. I can update the comment.

My mental model:

- Firefox uses a single socket per connection. Thus a syscall reads all queued UDP datagrams for at most one QUIC connection.
- With an MTU of 1_500 our current GRO buffer size allows for reading up to > 40 UDP datagrams in one syscall.
- Thus the marginal benefit of reading more than 40 UDP datagrams in one syscall is small. In other words when handling 40 datagrams at once, the syscall overhead becomes small.
- Based on conversations with CDNs at the last IETF, folks seem to use packet trains of ~10. In other words, say that reading on a connection is delayed, we could still read up to 4 packet trains in one syscall with our GRO-only implementation today.

@larseggert does the above make sense?

neqo-udp/src/lib.rs
mxinden: Good idea. I will look into it.

neqo-udp/src/lib.rs
```diff
@@ -7,20 +7,66 @@
...
+    pub fn new() -> Self {
+        Self(vec![vec![0; RECV_BUF_SIZE]; NUM_BUFS])
+    }
+}
+
+impl Default for RecvBuf {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
larseggert, line 69: This seems unused based on the coverage report. Use it or delete it?

mxinden: We enable the [`new_without_default`](https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default) clippy lint, which would complain as such if I remove the `Default` `impl`.

```
warning: you should consider adding a `Default` implementation for `RecvBuf`
  --> neqo-udp/src/lib.rs:60:5
   |
60 | /     pub fn new() -> Self {
61 | |         Self(vec![vec![0; RECV_BUF_SIZE]; NUM_BUFS])
62 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default
   = note: `#[warn(clippy::new_without_default)]` on by default
help: try adding this
   |
58 + impl Default for RecvBuf {
59 +     fn default() -> Self {
60 +         Self::new()
61 +     }
62 + }
```

Issue #2306, opened 2025-01-02
Potential Infinite loop in evict_to_internal function


KershawChang:
See [bug 1939690 comment 4](https://bugzilla.mozilla.org/show_bug.cgi?id=1939690#c4).

I've asked the reporter to provide STR. It's still not clear how easy this can be triggered.

mxinden:
The reporter on Bugzilla is hinting at this code:

https://github.com/mozilla/neqo/blob/922d266bba926532982ee3fa91574ae131d55248/neqo-qpack/src/table.rs#L263-L285



Off the top of my head I see the following scenario:

- `only_check` is `true`
- `self.dyamic.is_empty()` is false
- `e.can_reduce()` is true

This would lead to an infinite loop.

The following should fix the issue:

``` diff
diff --git a/neqo-qpack/src/table.rs b/neqo-qpack/src/table.rs
index 8c8b8d5e..abea8be2 100644
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -260,6 +260,7 @@ impl HeaderTable {
         self.evict_to_internal(reduce, true)
     }
 
     pub fn evict_to_internal(&mut self, reduce: u64, only_check: bool) -> bool {
         qtrace!(
             [self],
@@ -278,6 +279,8 @@ impl HeaderTable {
                 if !only_check {
                     self.used -= u64::try_from(e.size()).unwrap();
                     self.dynamic.pop_back();
+                } else {
+                    return true;
                 }
             }
         }
```

I will give this more thought and try to reproduce it in a unit test.

PR #2307, opened 2025-01-02
fix(qpack): make `can_evict_to` iter all evictable entries 

mxinden:
Previously we had `evict_to` and `can_evict_to`, where the former would actually evict values, whereas the latter would only check how much could be evicted. Both would delegate to `evict_to_internal`.

In the case of `evict_to`, `evict_to_internal` would look at the last entry, and evict it if possible, then repeat.

In the case of `can_evict_to`, `evict_to_internal` would look at the last value as well, and then repeat, but not actualy evict, thus continuously looking at the last entry, instead of all evictable entries.

This commit fixes `can_evict_to` by making it look at all evictable entries.

---

Partially fix https://github.com/mozilla/neqo/issues/2306.

### DIFF
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -253,21 +253,11 @@ impl HeaderTable {
     }
 
     fn evict_to(&mut self, reduce: u64) -> bool {
-        self.evict_to_internal(reduce, false)
-    }
-
-    pub fn can_evict_to(&mut self, reduce: u64) -> bool {
-        self.evict_to_internal(reduce, true)
-    }
-
-    // TODO: Needs to be pub?
-    pub fn evict_to_internal(&mut self, reduce: u64, only_check: bool) -> bool {
         qtrace!(
             [self],
-            "reduce table to {}, currently used:{} only_check:{}",
+            "reduce table to {}, currently used:{}",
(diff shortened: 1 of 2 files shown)


KershawChang (approved):
This change looks reasonable to me, but I think @larseggert or @martinthomson should also review this change.

neqo-qpack/src/table.rs
```diff
@@ -112,9 +112,9 @@ impl HeaderTable {
     ///
     /// # Errors
     ///
-    /// `ChangeCapacity` if table capacity cannot be reduced.
-    /// The table cannot be reduce if there are entries that are referred at the moment or their
-    /// inserts are unacked.
+    /// [`Error::ChangeCapacity`] if table capacity cannot be reduced.
+    /// The table cannot be reduced if there are entries that are referred at
+    /// the moment or their inserts are unacked.
```
larseggert, line 117:
```suggestion
    /// The table cannot be reduced if there are entries that are referred to at
    /// the moment, or whose inserts are unacked.
```

larseggert (approved):

mxinden:
Merging here to bring patch into Firefox. @martinthomson I am happy to do any follow-ups once you are back.

PR #2313, opened 2025-01-07
chore: Enable clippy `cfg_not_test`

larseggert:
https://rust-lang.github.io/rust-clippy/master/index.html#cfg_not_test


neqo-transport/src/connection/mod.rs
```diff
@@ -499,11 +496,13 @@ impl Connection {
...
-        {
-            assert!(!tparams::INTERNAL_TRANSPORT_PARAMETERS.contains(&tp));
-        }
+    #[cfg(test)]
+    pub fn set_local_tparam(
+        &self,
+        tp: tparams::TransportParameterId,
+        value: tparams::TransportParameter,
+    ) -> Res<()> {
+        assert!(cfg!(test) || tparams::INTERNAL_TRANSPORT_PARAMETERS.contains(&tp));
```
mxinden, line 505: I am confused. Given that the function has the `#[cfg(test)]` attribute, why is the `cfg!(test)` needed? Asked differently, isn't thus `tparams::INTERNAL_TRANSPORT_PARAMETERS.contains(&tp)` just dead code?

larseggert: I don't fully get this either. With just `assert!(tparams::INTERNAL_TRANSPORT_PARAMETERS.contains(&tp));`, these tests fail:
```
     Summary [   1.732s] 1174 tests run: 1171 passed, 3 failed, 2 skipped
        FAIL [   0.036s] neqo-transport connection::tests::handshake::anti_amplification
        FAIL [   0.030s] neqo-transport connection::tests::recovery::handshake_ack_pto
        FAIL [   0.030s] neqo-transport connection::tests::vn::compatible_upgrade_large_initial
```
With just `assert!(!tparams::INTERNAL_TRANSPORT_PARAMETERS.contains(&tp));`, (note the added negation) these tests fail:
```
     Summary [   1.827s] 1174 tests run: 1165 passed, 9 failed, 2 skipped
        FAIL [   0.030s] neqo-transport connection::tests::close::stateless_reset_client
        FAIL [   0.070s] neqo-transport connection::tests::handshake::bad_min_ack_delay
        FAIL [   0.050s] neqo-transport connection::tests::idle::tiny_idle_timeout
        FAIL [   0.034s] neqo-transport connection::tests::migration::preferred_address_client
        FAIL [   0.030s] neqo-transport connection::tests::migration::preferred_address_server_empty_cid
        FAIL [   0.034s] neqo-transport connection::tests::stream::max_data
        FAIL [   0.046s] neqo-transport connection::tests::vn::invalid_current_version_client
        FAIL [   0.046s] neqo-transport connection::tests::vn::invalid_current_version_server
        FAIL [   0.046s] neqo-transport connection::tests::vn::no_compatible_version
 ```

mxinden: Previously we have not been checking `INTERNAL_TRANSPORT_PARAMETERS` in tests at all. Thus I am not surprised by the above test failures both with and without `INTERNAL_TRANSPORT_PARAMETERS`.

Do we call `set_local_tparam` in any non-test code? If not, we might as well remove the check to `INTERNAL_TRANSPORT_PARAMETERS`. That would not be a change in behavior.

larseggert: I don't think we call `set_local_tparam` in non-test code.

mxinden:
Seems like we have been calling `set_local_param` from non-test code before:

https://github.com/mozilla/neqo/pull/1320/files#diff-c12ab6f4e19997beb7332cd18b610377c597bd04a6c75631653680f4e00b8299L1168-L1175

But that was removed in the same pull request (https://github.com/mozilla/neqo/pull/1320) that `INTERNAL_TRANSPORT_PARAMETERS` was introduced.

mxinden (approved):

larseggert:
Thanks for checking. So, a leftover.

PR #2316, opened 2025-01-07
chore: Enable clippy `decimal_literal_representation` lint

larseggert:
Not sure if we want this.

https://rust-lang.github.io/rust-clippy/master/index.html#decimal_literal_representation


larseggert:
Let's not do this.

mxinden:
> Especially for big powers of 2, a hexadecimal representation is usually more readable than a decimal representation.

:open_mouth: for me it is the opposite.

PR #2319, opened 2025-01-08
chore: Enable clippy `if_then_some_else_none` lint

larseggert:
And apply suggestions.


Cargo.toml
```diff
@@ -42,7 +42,7 @@ url = { version = "2.5.3", default-features = false, features = ["std"] }
 cargo = { level = "warn", priority = -1 }
 nursery = { level = "warn", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
-multiple_crate_versions = "allow"
```
mxinden, line 45: Why is this removal needed?

mxinden (approved):

Cargo.toml
larseggert: It's not needed (for this PR), I just stuck it in because we no longer need this `allow`.

PR #2320, opened 2025-01-08
chore: Make the TLS epoch an `enum` type

larseggert:
And replace `CryptoSpace` with it.


neqo-crypto/src/constants.rs
```diff
@@ -6,21 +6,60 @@
...
+    fn from(e: u16) -> Self {
+        match e {
+            0 => Self::Initial,
+            1 => Self::ZeroRtt,
+            2 => Self::Handshake,
+            3 => Self::ApplicationData,
+            _ => unreachable!(),
+        }
+    }
+}
```
mxinden, line 40: Instead of implementing `From<u16>` one could as well implement `TryFrom<u16>`. That would remove the `unreachable!` here, and would instead force the caller to handle the error case.

I would assume the caller knows better whether the input is sanitized, i.e. whether panicing is safe.

Thoughts?

neqo-crypto/src/secrets.rs
```diff
@@ -43,23 +44,16 @@ impl From<SSLSecretDirection::Type> for SecretDirection {
...
 
 impl DirectionalSecrets {
     fn put(&mut self, epoch: Epoch, key: SymKey) {
-        assert!(epoch > 0);
-        let i = (epoch - 1) as usize;
-        assert!(i < self.secrets.len());
-        // assert!(self.secrets[i].is_none());
-        self.secrets[i] = Some(key);
+        self.secrets[epoch] = Some(key);
     }
```
mxinden, line 53: Previously this function would panic on `epoch` `Initial`. Is it safe to change that behavior?

neqo-crypto/src/constants.rs
```diff
@@ -6,21 +6,60 @@
...
+impl std::fmt::Display for Epoch {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Self::Initial => write!(f, "Initial"),
+            Self::ZeroRtt => write!(f, "ZeroRtt"),
+            Self::Handshake => write!(f, "Handshake"),
+            Self::ApplicationData => write!(f, "ApplicationData"),
+        }
+    }
+}
```
mxinden, line 62: Why do we need a `Display` implementation?

mxinden (commented):
:+1: the more we can enforce constraints through the type system the better.

neqo-crypto/src/constants.rs
larseggert: Made the change. Also improved the other error handling in those functions a bit.

neqo-crypto/src/constants.rs
larseggert: Good catch! We used `{}` in a `qdebug!`. Changed that to `{:?}`.

mxinden (approved):
Looks good to me.

Let's give Martin a chance to comment on https://github.com/mozilla/neqo/pull/2320#discussion_r1910146468

neqo-crypto/src/secrets.rs
larseggert: Ping @martinthomson

neqo-crypto/src/constants.rs
```diff
@@ -6,21 +6,51 @@
...
+impl TryFrom<u16> for Epoch {
+    type Error = ();
+
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Self::Initial),
+            1 => Ok(Self::ZeroRtt),
+            2 => Ok(Self::Handshake),
+            3 => Ok(Self::ApplicationData),
+            _ => Err(()),
```
martinthomson, line 39: This could use a new Error type, rather than ().  Your call.

neqo-crypto/src/secrets.rs
```diff
@@ -44,23 +45,16 @@ impl From<SSLSecretDirection::Type> for SecretDirection {
 #[derive(Debug, Default)]
 pub struct DirectionalSecrets {
     // We only need to maintain 3 secrets for the epochs used during the handshake.
-    secrets: [Option<SymKey>; 3],
+    secrets: EnumMap<Epoch, Option<SymKey>>,
```
martinthomson, line 48: Does this need to be `Option<SymKey>` or just `EnumMap<Epoch, SymKey>` ?  Because we previously needed each to be optionally present, but the map already supports that.

neqo-crypto/src/secrets.rs
martinthomson: We can't get secrets for the Initial phase.  While we initialize the packet encryption with something, that is done at connection initialization time and we don't use the secrets store for that.

As for whether it makes sense to try to avoid the extra one value in the map, I don't think so.  The `SymKey` this holds is a simple pointer.

neqo-crypto/src/secrets.rs
```diff
@@ -78,7 +72,13 @@ impl Secrets {
         secret: *mut PK11SymKey,
         arg: *mut c_void,
     ) {
-        let secrets = arg.cast::<Self>().as_mut().unwrap();
+        let Ok(epoch) = Epoch::try_from(epoch) else {
+            // Don't touch secrets.
```
martinthomson, line 76: In both of these cases, we should complain loudly in a debug build.

neqo-transport/src/tracking.rs
```diff
@@ -48,8 +48,8 @@ impl PacketNumberSpace {
 impl From<Epoch> for PacketNumberSpace {
     fn from(epoch: Epoch) -> Self {
         match epoch {
-            TLS_EPOCH_INITIAL => Self::Initial,
-            TLS_EPOCH_HANDSHAKE => Self::Handshake,
+            Epoch::Initial => Self::Initial,
+            Epoch::Handshake => Self::Handshake,
             _ => Self::ApplicationData,
```
martinthomson, line 53:
```suggestion
            Epoch::ApplicationData | Epoch::ZeroRtt => Self::ApplicationData,
```

The underscore doesn't make sense any more.  Better to spell it out.

It might be worth documenting that we don't have a 0-RTT packet number space in the enum.  The basic story is that we use the application data space for that, using the appropriate keys based on availability.  (`CryptoSpace` holds the analogous values to `Epoch`.  Which leads me to the next comment...

martinthomson (commented):
Looks basically good.

One major request.  Would you mind deleting `CryptoSpace` and replacing it with the new `Epoch`?

PR #2322, opened 2025-01-08
chore: Enable the clippy `integer_division` lint

larseggert:
And fix all the warnings that it generates.


neqo-common/src/codec.rs
```diff
@@ -282,7 +282,7 @@ impl Encoder {
         let s = s.as_ref();
         assert_eq!(s.len() % 2, 0, "Needs to be even length");
 
-        let cap = s.len() / 2;
+        let cap = s.len() >> 1;
```
mxinden, line 285: To me `/ 2` is more intuitive than `>> 1`. In case this is about performance, I would be surprised if the compiler wouldn't do this potential optimization by itself.

All that said, I don't think this is the intention of this lint.

> When outside of some very specific algorithms, integer division is very often a mistake because it discards the remainder.

https://rust-lang.github.io/rust-clippy/master/index.html#integer_division

Using `>> 1` instead of `/ 2` still discards the remainder.

larseggert: To me, it;s the opposite, i.e., the shift makes it more clear to me that we're dropping the remainder. But YMMV.

Looking at this again, it's probably not worth doing.

larseggert:
Let's not do this.

PR #2328, opened 2025-01-09
chore: Enable the `semicolon_inside_block` lint

larseggert:
This is what we mostly use in the code.


mxinden (approved):
No opinion. The onliners are slightly more succinct.

PR #2329, opened 2025-01-09
chore: Enable the `shadow_same` lint

larseggert:
And fix the lint errors in the codebase.


mxinden (commented):
I consider the following a helpful pattern:

``` rust
let mut x = todo!();
// mutate x

// x won't be mutated from now on. Make it immutable.
let x = x;
```

Why explicitly disallow this?

larseggert:
I have not used that pattern. If you find it useful, let's not merge this.

mxinden:
We have been using it e.g. here https://github.com/mozilla/mtu/pull/4/files

PR #2330, opened 2025-01-09
chore: Enable the `str_to_string` lint

larseggert:
And fix all the warnings that it generates.


mxinden (approved):
Fine by me. Don't see a lot of value in it.

larseggert:
I'm not married to it. (I basically try each lint to see what it generates, and whatever seems mildly useful I make a PR for.)

PR #2332, opened 2025-01-09
chore: Enable the `unneeded_field_pattern` lint

larseggert:
And fix the lint errors in the codebase.


mxinden (approved):
I can see a case where one wants to explicitly name all fields, in order to get a compiler error when new fields are added.

That said, we can always locally disable the lint.

PR #2333, opened 2025-01-09
chore: Enable the `unused_result_ok` lint

larseggert:
Ignore errors explicitly.


neqo-transport/src/send_stream.rs
```diff
@@ -1591,10 +1589,11 @@ impl SendStreams {
...
 
-    #[allow(clippy::missing_panics_doc)]
-    #[allow(clippy::missing_errors_doc)]
-    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) -> Res<()> {
-        let stream: &mut SendStream = self.map.get_mut(&stream_id).ok_or(Error::InvalidStreamId)?;
+    pub fn set_fairness(&mut self, stream_id: StreamId, make_fair: bool) {
+        let Some(stream) = self.map.get_mut(&stream_id) else {
+            // We can get called with an invalid stream ID, and that is OK.
+            return;
+        };
```
mxinden, line 1596: Is silently `return`ing here the best option? I preferred the previous behavior where the caller decided whether calling with an invalid `StreamId` is OK.

Previously `Http3Connection::webtransport_create_stream_local` asserted (i.e. used `unwrap()`), in other words decided that it is not OK to call with invalid `StreamId`.

Silently `return`ing here might make it very hard to debug in the future.

mxinden (commented):
Other than the silent `return`, this looks good to me.

neqo-transport/src/send_stream.rs
larseggert: We had code that specifically ignored any passed-up errors in
https://github.com/mozilla/neqo/pull/2333/files#diff-55bcd011f453914d3922602a7c590756c17120748c1336dcb52e732e7818b21eL134

Should I just make that more explicit, and restore the old error handling?

mxinden: I prefer the explicit ignoring of the error on the caller side. In other words, moving the error-handling-by-ignoring-it as far downstream as possible.

larseggert: Ok, I will make the change.

neqo-http3/src/connection.rs
```diff
@@ -1284,7 +1288,7 @@ impl Http3Connection {
             .map_err(|e| Error::map_stream_create_errors(&e))?;
         // Set outgoing WebTransport streams to be fair (share bandwidth)
         // This really can't fail, panics if it does
-        conn.stream_fairness(stream_id, true).unwrap();
+        conn.stream_fairness(stream_id, true)?;
```
mxinden, line 1291: Comment is outdated now. Also I would be fine with a `stream_fairness(stream_id, true).expect("~this really can't fail ...");`

mxinden (approved):
Other than the outdated comment, this looks good to me.

PR #2336, opened 2025-01-09
ci: Remove some things from `env`

larseggert:
1. `CARGO_TERM_COLOR: always` is always set by `dtolnay/rust-toolchain`
2. `RUST_BACKTRACE` is better set where needed, rather than everywhere.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,9 +14,6 @@ on:
         type: boolean
         required: false
         default: false
-env:
-  CARGO_TERM_COLOR: always
-  RUST_BACKTRACE: 1
 
 concurrency:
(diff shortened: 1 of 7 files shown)


.github/workflows/check.yml
```diff
@@ -14,9 +14,6 @@ on:
         type: boolean
         required: false
         default: false
-env:
-  CARGO_TERM_COLOR: always
-  RUST_BACKTRACE: 1
```
mxinden, line 19: What is the drawback of setting `RUST_BACKTRACE`? In case we have a hard-to-reproduce unit test failure on CI, wouldn't it be good to have a stack trace?

larseggert: Ah yes, I meant to add that to the check step.

PR #2340, opened 2025-01-10
ci: Run `cargo-deny` during ci

larseggert:
Oh wow, an action that is actually already allowlisted :-)

### DIFF
--- a/.deny.toml
+++ b/.deny.toml
@@ -1,0 +1,31 @@
+[bans]
+multiple-versions = "deny"
+wildcards = "allow" # TODO: Is there a better way to allowlist local deps?
+allow-wildcard-paths = true
+
+[bans.workspace-dependencies]
+duplicates = "allow"
+include-path-dependencies = true
+unused = "allow"
+
+[licenses]
+unused-allowed-license = "allow"
+allow = [
+    "Apache-2.0",
+    "Apache-2.0 WITH LLVM-exception",
+    "BSD-2-Clause",
+    "BSD-3-Clause",
+    "CC0-1.0",
+    "ISC",
+    "MPL-2.0",
+    "MIT",
+    "MIT-0",
+    "Unicode-DFS-2016",
+    "Unicode-3.0",
+    "Zlib",
+]
+
+[sources]
+unknown-registry = "deny"
+unknown-git = "deny"
+required-git-spec = "rev"
--- a/.deny.toml
+++ b/.deny.toml
@@ -1,5 +1,5 @@
 [bans]
-multiple-versions = "deny"
+multiple-versions = "allow"
 
 [bans.workspace-dependencies]
 duplicates = "allow"
--- a/.deny.toml
+++ b/.deny.toml
@@ -1,11 +1,8 @@
 [bans]
 multiple-versions = "deny"
-wildcards = "allow" # TODO: Is there a better way to allowlist local deps?
-allow-wildcard-paths = true
 
 [bans.workspace-dependencies]
 duplicates = "allow"
-include-path-dependencies = true
 unused = "allow"
 
 [licenses]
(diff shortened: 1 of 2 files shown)


.github/workflows/deny.yml
```diff
@@ -0,0 +1,32 @@
...
+  contents: read
+
+jobs:
+  deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
```
mxinden, line 23:
```suggestion
          - bans
          - licenses
          - sources
```

Aren't these individual checks?

.github/workflows/deny.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
mxinden, line 26: Would we catch it, if it continues on error? I am fine failing CI and requiring a whitelisting pull request.

.deny.toml
```diff
@@ -0,0 +1,31 @@
+[bans]
+multiple-versions = "deny"
+wildcards = "allow" # TODO: Is there a better way to allowlist local deps?
```
mxinden, line 3: What is the failure without this?

mxinden (commented):
Seems like mozilla-central is using it too, but I can't find the "manifest". Did you?

https://searchfox.org/mozilla-central/rev/9f9fa5d2cb72b12ac7e168b7f6ee9820f63291e9/build/cargo/cargo-deny.yaml#7

.github/workflows/deny.yml
larseggert: Literally copied from https://github.com/EmbarkStudios/cargo-deny-action?tab=readme-ov-file#recommended-pipeline-if-using-advisories-to-avoid-sudden-breakages

.deny.toml
larseggert: Actually, none! :-)

(The moz config file I used as a template had an explicit `deny` there, which caused
```
error[wildcard]: found 4 wildcard dependencies for crate 'test-fixture'. allow-wildcard-paths is enabled, but does not apply to public crates as crates.io disallows path dependencies.
   ┌─ /Users/lars/Documents/Code/neqo/test-fixture/Cargo.toml:21:15
   │
21 │ neqo-common = { path = "../neqo-common", features = ["test-fixture"] }
   │               ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ wildcard dependency
22 │ neqo-crypto = { path = "../neqo-crypto" }
   │               ━━━━━━━━━━━━━━━━━━━━━━━━━━━ wildcard dependency
23 │ neqo-http3 = { path = "../neqo-http3", features = ["draft-29"] }
   │              ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ wildcard dependency
24 │ neqo-transport = { path = "../neqo-transport", features = ["draft-29"] }
   │                  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ wildcard dependency
```
but we can just remove that line entirely.)

.github/workflows/deny.yml
larseggert: I can make the change. This is taken from the recipe on the URL above.

.github/workflows/deny.yml
larseggert: I think what is there is correct, i.e., it is **one** run that checks **three** things.

.github/workflows/deny.yml
mxinden: No strong opinion. I assume it is easier to loosen this in the future, than manually checking whether the CI step succeeded.

mxinden (approved):

PR #2341, opened 2025-01-10
chore: We no longer need the `tpids!` macro

larseggert:


mxinden (approved):
I am fine either way.

PR #2342, opened 2025-01-10
feat: Try FxHasher to see if it makes a difference

larseggert:
WIP

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
 quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log", "fast-apple-datapath"] }
 regex = { version = "1.9", default-features = false, features = ["unicode-perl"] }
+rustc-hash = { version = "2.1", default-features = false, features = [ "std" ]}
 static_assertions = { version = "1.1", default-features = false }
 url = { version = "2.5.3", default-features = false, features = ["std"] }
 
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -6,13 +6,7 @@
 
 #![allow(clippy::module_name_repetitions)]
 
-use std::{
-    cell::RefCell,
-    collections::{BTreeSet, HashMap},
-    fmt::Debug,
-    mem,
-    rc::Rc,
-};
+use std::{cell::RefCell, collections::BTreeSet, fmt::Debug, mem, rc::Rc};
 
 use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn, Decoder, Header, MessageType, Role};
 use neqo_qpack::{decoder::QPackDecoder, encoder::QPackEncoder};
@@ -20,6 +14,7 @@ use neqo_transport::{
     streams::SendOrder, AppError, CloseReason, Connection, DatagramTracking, State, StreamId,
     StreamType, ZeroRttState,
 };
+use rustc_hash::FxHashMap;
 
 use crate::{
     client_events::Http3ClientEvents,
@@ -305,8 +300,8 @@ pub struct Http3Connection {
     pub qpack_decoder: Rc<RefCell<QPackDecoder>>,
     settings_state: Http3RemoteSettingsState,
     streams_with_pending_data: BTreeSet<StreamId>,
-    pub send_streams: HashMap<StreamId, Box<dyn SendStream>>,
-    pub recv_streams: HashMap<StreamId, Box<dyn RecvStream>>,
+    pub send_streams: FxHashMap<StreamId, Box<dyn SendStream>>,
+    pub recv_streams: FxHashMap<StreamId, Box<dyn RecvStream>>,
     webtransport: ExtendedConnectFeature,
 }
 
@@ -336,8 +331,8 @@ impl Http3Connection {
             local_params: conn_params,
             settings_state: Http3RemoteSettingsState::NotReceived,
             streams_with_pending_data: BTreeSet::new(),
-            send_streams: HashMap::new(),
-            recv_streams: HashMap::new(),
+            send_streams: FxHashMap::default(),
+            recv_streams: FxHashMap::default(),
             role,
         }
     }
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -9,7 +9,6 @@
 use std::{
     cell::RefCell,
     cmp::{max, min},
-    collections::HashMap,
     mem,
     ops::{Index, IndexMut, Range},
     rc::Rc,
@@ -25,6 +24,7 @@ use neqo_crypto::{
     TLS_GRP_EC_SECP256R1, TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1, TLS_GRP_EC_X25519,
     TLS_GRP_KEM_MLKEM768X25519, TLS_VERSION_1_3,
 };
+use rustc_hash::FxHashMap;
 
 use crate::{
     cid::ConnectionIdRef,
@@ -811,7 +811,7 @@ pub enum CryptoSpace {
 /// get other keys, so those have fixed versions.
 #[derive(Debug, Default)]
 pub struct CryptoStates {
-    initials: HashMap<Version, CryptoState>,
+    initials: FxHashMap<Version, CryptoState>,
     handshake: Option<CryptoState>,
     zero_rtt: Option<CryptoDxState>, // One direction only!
     cipher: Cipher,
@@ -1265,7 +1265,7 @@ impl CryptoStates {
             cipher: TLS_AES_128_GCM_SHA256,
             next_secret: hkdf::import_key(TLS_VERSION_1_3, &[0xaa; 32]).unwrap(),
         };
-        let mut initials = HashMap::new();
+        let mut initials = FxHashMap::default();
         initials.insert(
             Version::Version1,
             CryptoState {
@@ -1322,7 +1322,7 @@ impl CryptoStates {
             next_secret: secret.clone(),
         };
         Self {
-            initials: HashMap::new(),
+            initials: FxHashMap::default(),
             handshake: None,
             zero_rtt: None,
             cipher: TLS_CHACHA20_POLY1305_SHA256,
--- a/neqo-transport/src/tparams.rs
+++ b/neqo-transport/src/tparams.rs
@@ -8,7 +8,6 @@
 
 use std::{
     cell::RefCell,
-    collections::HashMap,
     net::{Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6},
     rc::Rc,
 };
@@ -19,6 +18,7 @@ use neqo_crypto::{
     ext::{ExtensionHandler, ExtensionHandlerResult, ExtensionWriterResult},
     random, HandshakeMessage, ZeroRttCheckResult, ZeroRttChecker,
 };
+use rustc_hash::FxHashMap;
 
 use crate::{
     cid::{ConnectionId, ConnectionIdEntry, CONNECTION_ID_SEQNO_PREFERRED, MAX_CONNECTION_ID_LEN},
@@ -312,7 +312,7 @@ impl TransportParameter {
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct TransportParameters {
-    params: HashMap<TransportParameterId, TransportParameter>,
+    params: FxHashMap<TransportParameterId, TransportParameter>,
 }
 
 impl TransportParameters {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,7 +86,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "shlex",
  "syn",
 ]
@@ -916,6 +916,7 @@ dependencies = [
  "neqo-qpack",
  "neqo-transport",
  "qlog",
+ "rustc-hash 2.1.0",
  "sfv",
  "test-fixture",
  "url",
@@ -929,6 +930,7 @@ dependencies = [
  "neqo-common",
  "neqo-transport",
  "qlog",
+ "rustc-hash 2.1.0",
  "static_assertions",
  "test-fixture",
 ]
@@ -946,6 +948,7 @@ dependencies = [
  "neqo-crypto",
  "neqo-transport",
  "qlog",
+ "rustc-hash 2.1.0",
  "smallvec",
  "static_assertions",
  "test-fixture",
@@ -1134,6 +1137,12 @@ name = "rustc-hash"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
+
+[[package]]
+name = "rustc-hash"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7fb8039b3032c191086b10f11f319a6e99e1e82889c5cc6046f515c9db1d497"
 
 [[package]]
 name = "ryu"
--- a/neqo-http3/Cargo.toml
+++ b/neqo-http3/Cargo.toml
@@ -24,6 +24,7 @@ neqo-crypto = { path = "./../neqo-crypto" }
 neqo-qpack = { path = "./../neqo-qpack" }
 neqo-transport = { path = "./../neqo-transport" }
 qlog = { workspace = true }
+rustc-hash = { workspace = true }
 sfv = { version = "0.9", default-features = false }
 url = { workspace = true }
 
--- a/neqo-http3/src/control_stream_local.rs
+++ b/neqo-http3/src/control_stream_local.rs
@@ -4,10 +4,11 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::VecDeque;
 
 use neqo_common::{qtrace, Encoder};
 use neqo_transport::{Connection, StreamId, StreamType};
+use rustc_hash::FxHashMap;
 
 use crate::{frames::HFrame, BufferedStream, Http3StreamType, RecvStream, Res};
 
@@ -50,7 +51,7 @@ impl ControlStreamLocal {
     pub fn send(
         &mut self,
         conn: &mut Connection,
-        recv_conn: &mut HashMap<StreamId, Box<dyn RecvStream>>,
+        recv_conn: &mut FxHashMap<StreamId, Box<dyn RecvStream>>,
     ) -> Res<()> {
         self.stream.send_buffer(conn)?;
         self.send_priority_update(conn, recv_conn)
@@ -59,7 +60,7 @@ impl ControlStreamLocal {
     fn send_priority_update(
         &mut self,
         conn: &mut Connection,
-        recv_conn: &mut HashMap<StreamId, Box<dyn RecvStream>>,
+        recv_conn: &mut FxHashMap<StreamId, Box<dyn RecvStream>>,
     ) -> Res<()> {
         // send all necessary priority updates
         while let Some(update_id) = self.outstanding_priority_update.pop_front() {
--- a/neqo-qpack/Cargo.toml
+++ b/neqo-qpack/Cargo.toml
@@ -22,6 +22,7 @@ neqo-common = { path = "./../neqo-common" }
 neqo-transport = { path = "./../neqo-transport" }
 qlog = { workspace = true }
 static_assertions = { workspace = true }
+rustc-hash = { workspace = true }
 
 [dev-dependencies]
 test-fixture = { path = "../test-fixture" }
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -4,10 +4,11 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::collections::{HashMap, HashSet, VecDeque};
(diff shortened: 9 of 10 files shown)


neqo-transport/src/tracking.rs
```diff
@@ -674,9 +674,8 @@ impl Default for AckTracker {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashSet;
-
     use neqo_common::Encoder;
+    use rustc_hash::FxHashSet;
```
martinthomson, line 678: A comment general to this changeset:

```suggestion
    use rustc_hash::FxHashSet as HashSet;
```

...would make this a smaller diff.

larseggert: Oh neat!

Cargo.toml
```diff
@@ -35,6 +35,7 @@ log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
 quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log", "fast-apple-datapath"] }
 regex = { version = "1.9", default-features = false, features = ["unicode-perl"] }
+rustc-hash = { version = "2.1", default-features = false, features = [ "std" ]}
```
mxinden, line 38: Note that mozilla-central uses `1.1.0`.

https://searchfox.org/mozilla-central/rev/e85232b4b28ecc970240d39203e417d1c320623c/Cargo.lock#5620-5624

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -4,9 +4,10 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::{cmp::max, collections::HashMap};
+use std::cmp::max;
 
 use neqo_common::{event::Provider as _, qdebug};
+use rustc_hash::FxHashMap as HashMap;
```
mxinden, line 10: What is the benefit of using `FxHashMap` in unit tests?

neqo-transport/src/crypto.rs
```diff
@@ -25,6 +24,7 @@ use neqo_crypto::{
     TLS_GRP_EC_SECP256R1, TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1, TLS_GRP_EC_X25519,
     TLS_GRP_KEM_MLKEM768X25519, TLS_VERSION_1_3,
 };
+use rustc_hash::FxHashMap as HashMap;
```
mxinden, line 27: Do we really see a performance gain in unit test runtime? Is it worth it?

mxinden (commented):
:+1: in general.

That said, I would prefer only replacing `std::collectionsHash*` where ever it proofs beneficial, e.g. not in unit tests.

neqo-transport/src/connection/tests/stream.rs
larseggert: None. I did a global replace :-) Will roll back for tests.

neqo-transport/src/crypto.rs
larseggert: I think we do. I'm currently mucking with the bencher setup so we get deltas-against-main for the performance table at the bottom of the benchmarks,

martinthomson: So the changes in THIS file relate to CryptoStates::initials, which might be better served by using EnumMap.

neqo-transport/src/tparams.rs
```diff
@@ -19,6 +18,7 @@ use neqo_crypto::{
     ext::{ExtensionHandler, ExtensionHandlerResult, ExtensionWriterResult},
     random, HandshakeMessage, ZeroRttCheckResult, ZeroRttChecker,
 };
+use rustc_hash::FxHashMap as HashMap;
```
martinthomson, line 21: This is used for our transport parameters, which is another place where an `EnumMap` could save us time.  We currently key on a `u64`, but we don't allow inserts unless we understand the transport parameter, so it would be possible to switch.

martinthomson (commented):
I've spotted two places where `EnumMap` could give us bigger wins.

I think that your performance gains largely derive from the changes to the client and server code.  There, the security risk is limited (we're not using this server in real deployments).  

Still, you should review the changes for security risk.  This hasher could expose us to DoS if the hashed values are controlled by an adversary.  I've checked the usage in our server code, which is fine because attackers don't get to control memory allocations (we use pointer values for the hash).  Still, that makes me wonder whether we should be using `Pin`.

larseggert:
@martinthomson thanks for the analysis. My plan is to add some benches first in another PR. I'll add some for those instances where you suggest to look into `EnumMap` as well.

Even if some of the macro benefits come from speeding up the demo client and server code, it's IMO still worthwhile doing, since eliminating those overheads makes it easier to spot other bottlenecks.

About security, I didn't do much of an analysis, but I think the main use of this insecure hasher would be when looking up items (streams, unacked chunks) that while under the control of an attacker are also quite limited in what valid values are that wouldn't immediately cause a connection clause.

neqo-http3/src/connection.rs
```diff
@@ -335,8 +332,8 @@ impl Http3Connection {
             local_params: conn_params,
             settings_state: Http3RemoteSettingsState::NotReceived,
             streams_with_pending_data: BTreeSet::new(),
-            send_streams: HashMap::new(),
-            recv_streams: HashMap::new(),
+            send_streams: HashMap::default(),
+            recv_streams: HashMap::default(),
```
larseggert, line 336: @martinthomson do you remember why we are using `HashMap` here, but `IndexMap` in `Connection`?

martinthomson: This is probably down to us paying more attention to the core transport than the HTTP/3 code.  I'd say that if we found `IndexMap` to be better at the lower layer, then it is probably every bit as good here.  The pattern of usage is nearly identical.

martinthomson:
I definitely agree with the point about removing the overheads from our toy code as much as possible.  This seems like a pretty substantial win there, so it's worth doing.  I doubt that my `EnumMap` suggestions will have a major impact, but the change did highlight the possibility (and it's not that much typing to switch over).

larseggert:
I think the `EnumMap` work should be factored out another PR, it will cause a bunch of changes throughout.

neqo-http3/src/connection.rs
```diff
@@ -303,9 +298,9 @@ pub struct Http3Connection {
     pub qpack_encoder: Rc<RefCell<QPackEncoder>>,
     pub qpack_decoder: Rc<RefCell<QPackDecoder>>,
     settings_state: Http3RemoteSettingsState,
-    streams_with_pending_data: BTreeSet<StreamId>,
-    pub send_streams: HashMap<StreamId, Box<dyn SendStream>>,
-    pub recv_streams: HashMap<StreamId, Box<dyn RecvStream>>,
+    streams_with_pending_data: HashSet<StreamId>,
+    pub send_streams: IndexMap<StreamId, Box<dyn SendStream>>,
```
mxinden, line 302: Conceptually, why is an `IndexMap` faster than a `HashMap` here? Both end up hashing the key, right?

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -43,8 +44,8 @@ pub struct WebTransportSession {
     state: SessionState,
     frame_reader: FrameReader,
     events: Box<dyn ExtendedConnectEvents>,
-    send_streams: BTreeSet<StreamId>,
-    recv_streams: BTreeSet<StreamId>,
+    send_streams: HashSet<StreamId>,
```
mxinden, line 47: What is the benefit of a `HashSet` over a BTreeSet` here?

neqo-transport/src/lib.rs
```diff
@@ -75,6 +78,8 @@ pub use self::{
     version::Version,
 };
 
+pub type IndexMap<K, V> = indexmap::IndexMap<K, V, BuildHasherDefault<FxHasher>>;
```
mxinden, line 81: Why is this more performant? Can you add a comment?

neqo-http3/src/connection.rs
```diff
@@ -1182,10 +1177,10 @@ impl Http3Connection {
                             events,
                             self.role,
                             self.recv_streams
-                                .remove(&stream_id)
+                                .shift_remove(&stream_id)
```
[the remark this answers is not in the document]
larseggert: Discussed in https://github.com/mozilla/neqo/pull/2342#discussion_r1948917641

neqo-http3/src/connection.rs
larseggert: It's not faster, it's possibly even slower. I made the change based on this comment: https://github.com/mozilla/neqo/pull/2342/files/bd061693b40e91b846c4d4cd1bc0ecfcd27c4e45#r1945509653

neqo-http3/src/features/extended_connect/webtransport_session.rs
larseggert: I was assuming `HashSet` to be faster?

neqo-transport/src/lib.rs
larseggert: It's probably nor, see above as to why.

neqo-http3/src/connection.rs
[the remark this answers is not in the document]
mxinden: Got it. Thanks.

neqo-http3/src/features/extended_connect/webtransport_session.rs
mxinden: I think we should only make these changes if we have proof that they are more performant.

Connecting this to the discussion above, might ordering (i.e. thus `BTreeSet`) be relevant here as well?

neqo-http3/src/connection.rs
martinthomson: So I'm not completely sure that we need consistently ordered iteration here.  We rely on that in the transport crate because we want frame sending to follow a predictable pattern (clear the old stuff out first).  However, that's not something we rely on at this layer of the stack, so whatever is fastest might be the right answer.

Maybe the first question we have is whether we need (insertion-)ordered iteration over this collection?

martinthomson:
> Though before we introduce the complexity of Pin, we might find a simple way around hashing the pointer values in the first place.

Definitely the right question to be asking.  I *think* that it might be possible to use the first connection ID as a key for this sort of thing, but we don't tend to keep that around today, once we stop using it.  Everything else -- as far as I know -- is ephemeral and therefore not suitable.

neqo-http3/src/features/extended_connect/webtransport_session.rs
larseggert: Fair point about only changing this after benchmarking.

neqo-http3/src/connection.rs
larseggert: First, I think we should differentiate between RX and TX.

On RX (both here and in transport), I don't see a point in using anything that maintains order or has semantics that go beyond a simple map or set. I think all we do here is lookups based on received data.

On TX, (both here and in transport), do we **really** care that iteration is consistently ordered? All these data structures _should_ mostly have a stable order while there are no insertions or removals. When there are, do we care that we'll have a different order afterwards? I can't convince myself that we do.

In other words, I think we should use whatever is fastest.

(Also, unrelated, I find the duplication/overlap between the transport and http3 crates really odd.)

larseggert:
I'm doing a benchmark to quantify the benefits first. (It's not going well, a **lot** of variation run-to-run for some reason.)

mxinden:
> a **lot** of variation run-to-run for some reason

That is counterintuitive for me, given that it uses `test-fixtures` and thus does no IO via the OS. Let me know if you want me to look into it.

larseggert:
I wonder if it's the CPU scheduler and frequency control on my Mac.

mxinden:
For what it is worth, here is #2444 on my machine:

```
➜  neqo-http3 git:(test-streams-bench) ✗ cargo bench --features bench

Benchmarking 1 streams of 1 bytes/multistream: Collecting 100 samples in estimated 5.1966 s (2400 
1 streams of 1 bytes/multistream
                        time:   [31.399 µs 31.555 µs 31.731 µs]
                        change: [-12.172% -10.263% -8.3468%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

Benchmarking 1000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 6.2030 s (40
1000 streams of 1 bytes/multistream
                        time:   [13.088 ms 13.117 ms 13.151 ms]
  2 (2.00%) high mild
  3 (3.00%) high severe

Benchmarking 10000 streams of 1 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 87.1s, or reduce sample count to 10.
Benchmarking 10000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 87.111 s (1
10000 streams of 1 bytes/multistream
                        time:   [876.43 ms 882.16 ms 888.29 ms]
  2 (2.00%) high mild

Benchmarking 1 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.0982 s (22
1 streams of 1000 bytes/multistream
                        time:   [33.435 µs 33.884 µs 34.409 µs]
  5 (5.00%) high mild
  2 (2.00%) high severe

Benchmarking 100 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.1397 s (
100 streams of 1000 bytes/multistream
                        time:   [1.5683 ms 1.5823 ms 1.5968 ms]

Benchmarking 1000 streams of 1000 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.2s, or reduce sample count to 60.
Benchmarking 1000 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 7.2433 s 
1000 streams of 1000 bytes/multistream
                        time:   [66.837 ms 67.100 ms 67.391 ms]
Found 20 outliers among 100 measurements (20.00%)
  5 (5.00%) high mild
  15 (15.00%) high severe

➜  neqo-http3 git:(test-streams-bench) ✗ cat /proc/cpuinfo     
                               
model name      : AMD Ryzen 7 7840U w/ Radeon  780M Graphics
```

I don't see much deviation. Am I running the wrong version @larseggert?

larseggert:
Can you run it again and see if there are changes run to run? That is where I see random improvements or regressions.

mxinden:
Here are two more runs with vanilla https://github.com/mozilla/neqo/pull/2444. No significant deviations. Note that I am not running your optimizations in this pull request.

```
➜  neqo-http3 git:(test-streams-bench) ✗ cargo bench --features bench
Benchmarking 1 streams of 1 bytes/multistream: Collecting 100 samples in estimated 5.0811 s 
1 streams of 1 bytes/multistream
                        time:   [31.514 µs 31.727 µs 32.013 µs]
                        change: [-0.3795% +0.5461% +1.6585%] (p = 0.28 > 0.05)
  3 (3.00%) high mild
  4 (4.00%) high severe

Benchmarking 1000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 6.2106
1000 streams of 1 bytes/multistream
                        time:   [13.032 ms 13.066 ms 13.104 ms]
                        change: [-0.7614% -0.3884% -0.0397%] (p = 0.04 < 0.05)
                        Change within noise threshold.
  1 (1.00%) low mild
  1 (1.00%) high mild
  9 (9.00%) high severe

Benchmarking 10000 streams of 1 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 87.3s, or reduce sample count to 10.
Benchmarking 10000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 87.26
10000 streams of 1 bytes/multistream
                        time:   [850.20 ms 852.13 ms 853.94 ms]
                        change: [-4.1112% -3.4050% -2.7470%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) low severe
  5 (5.00%) low mild
  3 (3.00%) high mild

Benchmarking 1 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.1258
1 streams of 1000 bytes/multistream
                        time:   [32.380 µs 32.615 µs 32.914 µs]
                        change: [-5.3650% -3.7472% -2.1786%] (p = 0.00 < 0.05)
                        Performance has improved.
  9 (9.00%) high mild
  2 (2.00%) high severe

Benchmarking 100 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.16
100 streams of 1000 bytes/multistream
                        time:   [1.4970 ms 1.5041 ms 1.5121 ms]
                        change: [-5.9242% -4.9438% -3.9764%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  4 (4.00%) high severe

Benchmarking 1000 streams of 1000 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.9s, or reduce sample count to 70.
Benchmarking 1000 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 6.9
1000 streams of 1000 bytes/multistream
                        time:   [66.039 ms 66.255 ms 66.489 ms]
                        change: [-1.7872% -1.2586% -0.7446%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 21 outliers among 100 measurements (21.00%)
  18 (18.00%) high mild
  3 (3.00%) high severe
```

```
➜  neqo-http3 git:(test-streams-bench) ✗ cargo bench --features bench
Benchmarking 1 streams of 1 bytes/multistream: Collecting 100 samples in estimated 5.0222 s 
1 streams of 1 bytes/multistream
                        time:   [31.196 µs 31.566 µs 32.008 µs]
                        change: [-1.9923% -0.5099% +1.0965%] (p = 0.52 > 0.05)
  2 (2.00%) high mild
  3 (3.00%) high severe

Benchmarking 1000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 6.2479
1000 streams of 1 bytes/multistream
                        time:   [12.863 ms 12.919 ms 12.980 ms]
                        change: [-1.6309% -1.1270% -0.5695%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 19 outliers among 100 measurements (19.00%)
  10 (10.00%) high mild
  9 (9.00%) high severe

Benchmarking 10000 streams of 1 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 87.5s, or reduce sample count to 10.
Benchmarking 10000 streams of 1 bytes/multistream: Collecting 100 samples in estimated 87.46
10000 streams of 1 bytes/multistream
                        time:   [862.91 ms 864.67 ms 866.53 ms]
                        change: [+1.1571% +1.4717% +1.7784%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe

Benchmarking 1 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.1478
1 streams of 1000 bytes/multistream
                        time:   [32.551 µs 32.892 µs 33.283 µs]
                        change: [-0.5530% +0.8511% +2.2636%] (p = 0.24 > 0.05)
  5 (5.00%) high mild
  2 (2.00%) high severe

Benchmarking 100 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 5.22
100 streams of 1000 bytes/multistream
                        time:   [1.5114 ms 1.5174 ms 1.5245 ms]
                        change: [+0.2271% +0.8880% +1.5818%] (p = 0.01 < 0.05)
                        Change within noise threshold.
  2 (2.00%) high severe

Benchmarking 1000 streams of 1000 bytes/multistream: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.0s, or reduce sample count to 70.
Benchmarking 1000 streams of 1000 bytes/multistream: Collecting 100 samples in estimated 7.0
1000 streams of 1000 bytes/multistream
                        time:   [66.765 ms 66.997 ms 67.247 ms]
                        change: [+0.6277% +1.1200% +1.6225%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe
```

larseggert:
Oh good. I think it is ore pinning being awkward on macOS then.

BTW, I came across https://manuel.bernhardt.io/posts/2023-11-16-core-pinning/ today, and we should change the bencher accordingly.

larseggert:
I'm redoing this PR in stages, to check if the new bench actually shows any improvements. The first push changes only the existing (non-test) uses of `HashMap` and `HashSet` to `FxHasher`.

larseggert:
Hm. The benches all show small improvements, while the client/server tests all show small regressions...

larseggert:
This shows enough benefit for me. @mxinden?

Cargo.toml
```diff
@@ -36,6 +36,7 @@ log = { version = "0.4", default-features = false }
 qlog = { version = "0.15.1", default-features = false }
 quinn-udp = { version = "0.5.11", default-features = false, features = ["direct-log"] }
 regex = { version = "1.9", default-features = false }
+rustc-hash = { version = "1.1", default-features = false, features = [ "std" ]}
```
mxinden, line 39: Firefox seems to use `rustc-hash` `2.1.1`:

https://searchfox.org/mozilla-central/rev/6a72a6d20eeb1b20b93862a79166938d6ce794a0/Cargo.lock#5846-5850

mxinden (commented):
I am fine merging here. Though I suggest we fix https://github.com/mozilla/neqo/issues/2657 first.

Also, how do we know whether each individual change has a positive impact? Maybe some are negative while the overall sum is positive.

larseggert:
I think FxHasher is strictly faster than the built-in one.

mxinden:
I don't feel strongly about it, i.e. I am fine merging here. Just missing the version change suggested above.

mxinden (approved):

PR #2352, opened 2025-01-13
Revert "perf(udp): multi-message receive on apple (#2302)"

mxinden:
This reverts commit 3c48567e164678acbeeea66391a378080bcee6c3.

`RPS` benchmark sometimes fails with `TransportError(PeerError(5)`. Git bisect shows 3c48567e164678acbeeea66391a378080bcee6c3 introduced the regression.

Until I figure out the reason, let's revert.


larseggert:
I wonder if I messed something up on the quinn-udp side. Off-by-one error on TX or RX when dealing with the buffer vector? Hm.

mxinden:
Debugging this further, the `TransportError(PeerError(5))` is originating here:

https://github.com/mozilla/neqo/blob/8bc926afc7aea9b9d2bda6b168ed38d94e229f31/neqo-transport/src/streams.rs#L420-L422

mxinden:
Ah, I believe `existed` is wrong:

https://github.com/mozilla/neqo/blob/8bc926afc7aea9b9d2bda6b168ed38d94e229f31/neqo-transport/src/streams.rs#L416-L422

Say that the stream is remote initiated, then `existed` is `false` and then `ss.is_none() && rs.is_none() && !existed` is `true`.

I will create a pull request.

mxinden:
Closing here in favor of https://github.com/mozilla/neqo/pull/2358.

PR #2357, opened 2025-01-14
ci: Use nightly wireshark for qns

larseggert:
Because some QNS tests require a newer version of Wireshark.


larseggert:
The nightly PPA appears abandoned :-(

PR #2358, opened 2025-01-14
fix(transport): handle out-of-order frame on remote init stream

mxinden:
https://github.com/mozilla/neqo/pull/2269 made a "stream control or data frame for a stream that is ours to initiate but we haven't yet" illegal.

https://github.com/mozilla/neqo/blob/7bbf900874eb02171dd893810d6895e14ed516e3/neqo-transport/src/streams.rs#L409-L424

Given a remote initiated stream that has been closed and forgotten since. Say that the local node receives an out-of-order frame on said forgotten stream. In such scenario:

- `stream_id.is_remote_initiated` is `true`
- thus `existed` is `false`
- `ss.is_none()` and `rs.is_none()` is `true`
- `ss.is_none() && rs.is_none() && !existed` is `true`
- thus `obtain_stream` falsely returns `Err(Error::StreamStateError);`

This commit fixes the above issue, only returning `Err(Error::StreamStateError);` on locally initiated streams.

### DIFF
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -539,23 +539,23 @@ fn do_not_accept_data_after_stop_sending() {
     );
 }
 
+struct Writer(Vec<u64>);
+
+impl crate::connection::test_internal::FrameWriter for Writer {
+    fn write_frames(&mut self, builder: &mut PacketBuilder) {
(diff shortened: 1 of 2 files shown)


larseggert (approved):

mxinden:
Updated PR description and added a unit test. @larseggert can you take another look?

larseggert (approved):

PR #2359, opened 2025-01-15
ci: Fix the `add_event_with_instant` test

larseggert:
It fails if the runtime exceeds 10 seconds, which happens sometimes in CI, e.g., https://github.com/mozilla/neqo/actions/runs/12772087190/job/35600809738?pr=2295#step:7:231


mxinden (approved):
Good catch. Thanks for debugging.

PR #2360, opened 2025-01-15
feat: Sock puppets for CHs

larseggert:
This is a proof of concept for an idea that was discussed with @martinthomson a while ago. That idea is to prepend the actual CH with an invalid "sock puppet" CH that is invalid but contains an innocuous SNI.

Update: I don't think we can make this work @martinthomson. When the server accepts the first CH enough so that it starts parsing the crypto data, all errors it encounters become connection errors and it will therefore not look further into the packet at the second coalesced CH. If we invalidate the first CH at the QUIC layer, the server will discard the entire packets, also without looking further into it. (We may need a TLS extension that says "ignore this CH and keep going"?)

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -278,6 +278,10 @@ pub struct Connection {
     release_resumption_token_timer: Option<Instant>,
     conn_params: ConnectionParameters,
     hrtime: hrtime::Handle,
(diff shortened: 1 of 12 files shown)


larseggert:
Is there maybe a way to do sock puppets via the ECH inner and outer CHs? CC @martinthomson and @dennisjackson, since they know that spec better than me.

larseggert:
I don't see a way forward for this w/o QUIC changes.

PR #2361, opened 2025-01-15
ci: Revamp the QNS CI action timeout

larseggert:
It seems that the `TIMEOUT` env variable applies to all steps?


mxinden (approved):
Oh, that is a surprise to me.

PR #2363, opened 2025-01-16
test(transport/vn): allow dropped_rx of 1

mxinden:
`compatible_upgrade_large_initial` tests whether Neqo can handle an ACK using QUIC version Y on a connection that ends up using QUIC version X.

https://github.com/mozilla/neqo/pull/2228 changed the test, asserting that the server will drop 2 instead of 1 packet. This spuriosly fails on CI:

https://github.com/mozilla/neqo/actions/runs/12796212878/job/35675377711?pr=2295

This commit allows both 1 and 2 packets dropped by the server.

---

@larseggert based on your changes in https://github.com/mozilla/neqo/pull/2228/, is this a valid assumption to make?

### DIFF
--- a/neqo-transport/src/connection/tests/vn.rs
+++ b/neqo-transport/src/connection/tests/vn.rs
@@ -251,7 +251,7 @@ fn compatible_upgrade_large_initial() {
     assert_eq!(server.version(), Version::Version2);
     // Only handshake padding is "dropped".
     assert_eq!(client.stats().dropped_rx, 1);
-    assert_eq!(server.stats().dropped_rx, 2);
+    assert!(matches!(server.stats().dropped_rx, 1 | 2));
 }
 
 /// A server that supports versions 1 and 2 might prefer version 1 and that's OK.


larseggert:
I can't recall why I made this change. So with #2228, is it that sometimes we have one and sometimes we have two packets we drop?

(As part of #2360, there will be a connection parameter to turn off SNI slicing. We can enable that here and revert the condition to the one from before #2228.)

mxinden:
> is it that sometimes we have one and sometimes we have two packets we drop?

Correct.

> (As part of #2360, there will be a connection parameter to turn off SNI slicing. We can enable that here and revert the condition to the one from before #2228.)

Sounds like a plan. That said, I would like merge https://github.com/mozilla/neqo/pull/2295 rather earlier than later, i.e. not block it on another pull request. We currently run Firefox with Cubic but our unit tests with New Reno. I fear we might introduce a regression.

@larseggert objections to the following plan?

1. Merge this pull request.
2. Merge https://github.com/mozilla/neqo/pull/2295.
3. Once https://github.com/mozilla/neqo/pull/2360 is merged, adjust the unit test touched here (i.e. `compatible_upgrade_large_initial`).

larseggert (approved):

Issue #2365, opened 2025-01-16
ci: nextest v0.9.88 requires rustc v1.81


mxinden:
Our rustc v1.76 CI is failing with: 

> cannot install package `cargo-nextest 0.9.88`, it requires rustc 1.81 or newer, while the currently active rustc version is 1.76.0

https://github.com/mozilla/neqo/actions/runs/12806683599/job/35705675701?pr=2295

nextest v0.9.88 was released 12h ago:

https://github.com/nextest-rs/nextest/releases/tag/cargo-nextest-0.9.88

I will try to pin `nextest`.

sunshowers:
Hi there -- FYI 0.9.86 bumped the MSRV: https://crates.io/crates/cargo-nextest/versions

Nextest's MSRV policy is "stable minus 2", as documented here: https://nexte.st/docs/stability/#minimum-supported-rust-version-msrv. It looks like a number of projects were relying on the MSRV having been steady across a few versions. In the future, I plan to bump nextest's MSRV more aggressively to ensure that projects aren't relying on a policy which doesn't exist.

(You're always welcome to use the pre-built binaries, of course!)

sunshowers:
Ah so the real issue here is that there is likely a Windows race which is preventing quickinstall from installing cargo binstall: https://github.com/mozilla/neqo/actions/runs/12806683599/job/35705675701?pr=2295#step:3:485

You could try using https://nexte.st/docs/installation/pre-built-binaries/#using-nextest-in-github-actions -- this will skip binstall and will fetch the binaries directly from GitHub Releases (what binstall is supposed to do, and successfully did on macOS and Linux).

larseggert:
Thanks for the analysis! We'd need to get `taiki-e/install-action` allowlisted first.

sunshowers:
Gotcha! You can also curl the binaries yourself via the URLs at the top of the page, if that makes it easier.

larseggert:
@mxinden do you still plan on landing one of the two PRs? (It seems the spurious fail hasn't happened again.)

mxinden:
I will close for now. Let's revisit in case we see failures in the future.

@sunshowers very much appreciate you reaching out directly! Thanks for the help.

PR #2337, opened 2025-01-09
ci: Enable `codecov/test-results-action`

larseggert:
https://github.com/codecov/test-results-action

Blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1940781

PR #2366, opened 2025-01-16
ci: remove nextest

mxinden:
Fixes https://github.com/mozilla/neqo/issues/2365.

Alternative to pinning nextest (https://github.com/mozilla/neqo/pull/2367).


larseggert:
WFM, too.

I'm kinda used to looking at nextest output, but I can see the benefits of not having another dependency.

larseggert:
If we're going this route, note that #2337 will add a `nextest` config that we'll need to port to `cargo test`.

PR #2372, opened 2025-01-17
ci(clippy): include benchmarks

mxinden:


mxinden:
Note that we can do the change below once async closures are stable:

``` diff
diff --git a/neqo-bin/benches/main.rs b/neqo-bin/benches/main.rs
index b2ef0987..43220b61 100644
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -57,7 +57,7 @@ fn transfer(c: &mut Criterion) {
         group.bench_function("client", |b| {
             b.to_async(Runtime::new().unwrap()).iter_batched(
                 || client::client(client::Args::new(&requests, upload)),
-                |client| async move {
+                async |client| {
                     client.await.unwrap();
                 },
                 BatchSize::PerIteration,
```

larseggert (approved):

PR #2385, opened 2025-01-23
feat: In-place crypto

### LINKED ISSUE #2246: perf: consider in-place en- and decryption
We currently re-allocate when en- and decrypting:

https://github.com/mozilla/neqo/blob/f3d01912b9a2d5962fffae1a192a3f62e738f745/neqo-transport/src/crypto.rs#L664-L665

https://github.com/mozilla/neqo/blob/f3d01912b9a2d5962fffae1a192a3f62e738f745/neqo-transport/src/crypto.rs#L689

Now that we tackled various low-hanging fruits around memory allocation (e.g. https://github.com/mozilla/neqo/pull/2184), memory allocations when en-/decrypting become more prominent. See e.g. the following recent Firefox profile doing a large file transfer:

https://profiler.firefox.com/public/y1ck0xgt3102jyqjjp93ggda99dt8bz2yg5c7k8/flame-graph/?globalTrackOrder=01&thread=1&transforms=ff-72~ff-82~ff-83&v=10

larseggert:
Only in-place encryption so far, and only for the main data path.

Fixes #2246 (eventually)

### DIFF
--- a/neqo-crypto/src/aead.rs
+++ b/neqo-crypto/src/aead.rs
@@ -6,7 +6,7 @@
 
 use std::{
     fmt,
-    ops::{Deref, DerefMut},
+    ops::{Deref, DerefMut, Range},
     os::raw::{c_char, c_uint},
     ptr::null_mut,
 };
@@ -124,6 +124,39 @@ impl RealAead {
             )
         }?;
         Ok(&output[0..(l.try_into()?)])
+    }
+
+    /// Encrypt a plaintext in place.
+    ///
+    /// The space provided in `data` needs to allow `Aead::expansion` more bytes to be appended.
+    ///
+    /// # Errors
+    ///
+    /// If the input can't be protected or any input is too large for NSS.
+    pub fn encrypt_in_place(
+        &self,
+        count: u64,
+        aad: Range<usize>,
+        input: Range<usize>,
+        data: &mut [u8],
+    ) -> Res<usize> {
+        let aad = &data[aad];
+        let input = &data[input];
+        let mut l: c_uint = 0;
+        unsafe {
+            SSL_AeadEncrypt(
+                *self.ctx,
+                count,
+                aad.as_ptr(),
+                c_uint::try_from(aad.len())?,
+                input.as_ptr(),
+                c_uint::try_from(input.len())?,
+                input.as_ptr(),
+                &mut l,
+                c_uint::try_from(input.len() + self.expansion())?,
+            )
+        }?;
+        Ok(l.try_into()?)
     }
 
     /// Decrypt a ciphertext.
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -418,9 +418,22 @@ impl PacketBuilder {
             hex(hdr),
             hex(body)
         );
-        let ciphertext = crypto.encrypt(self.pn, hdr, body)?;
+
+        // Add space for crypto expansion.
+        let data_end = self.encoder.len();
+        for _i in 0..crypto.expansion() {
+            self.encode_byte(123);
+        }
+
+        let ciphertext_len = crypto.encrypt(
+            self.pn,
+            self.header.clone(),
+            self.header.end..data_end,
+            self.encoder.as_mut(),
+        )?;
 
         // Calculate the mask.
+        let ciphertext = &self.encoder.as_mut()[self.header.end..self.header.end + ciphertext_len];
         let offset = SAMPLE_OFFSET - self.offsets.pn.len();
         if offset + SAMPLE_SIZE > ciphertext.len() {
             return Err(Error::InternalError);
@@ -434,9 +447,6 @@ impl PacketBuilder {
             self.encoder.as_mut()[j] ^= mask[i];
         }
 
-        // Finally, cut off the plaintext and add back the ciphertext.
-        self.encoder.truncate(self.header.end);
-        self.encoder.encode(&ciphertext);
         qtrace!("Packet built {}", hex(&self.encoder));
         Ok(self.encoder)
     }
@@ -998,7 +1008,8 @@ mod tests {
 
         // The spec uses PN=1, but our crypto refuses to skip packet numbers.
         // So burn an encryption:
-        let burn = prot.encrypt(0, &[], &[]).expect("burn OK");
+        let mut burn = [0; 16];
+        prot.encrypt(0, 0..0, 0..0, &mut burn).expect("burn OK");
         assert_eq!(burn.len(), prot.expansion());
 
         let mut builder = PacketBuilder::long(
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -40,7 +40,6 @@ use crate::{
     Error, Res,
 };
 
-const MAX_AUTH_TAG: usize = 32;
 /// The number of invocations remaining on a write cipher before we try
 /// to update keys.  This has to be much smaller than the number returned
 /// by `CryptoDxState::limit` or updates will happen too often.  As we don't
@@ -634,12 +633,18 @@ impl CryptoDxState {
         self.used_pn.end
     }
 
-    pub fn encrypt(&mut self, pn: PacketNumber, hdr: &[u8], body: &[u8]) -> Res<Vec<u8>> {
+    pub fn encrypt(
+        &mut self,
+        pn: PacketNumber,
+        hdr: Range<usize>,
+        body: Range<usize>,
+        data: &mut [u8],
+    ) -> Res<usize> {
         debug_assert_eq!(self.direction, CryptoDxDirection::Write);
         qtrace!(
-            "[{self}] encrypt pn={pn} hdr={} body={}",
-            hex(hdr),
-            hex(body)
+            "[{self}] encrypt_in_place pn={pn} hdr={} body={}",
+            hex(data[hdr.clone()].as_ref()),
+            hex(data[body.clone()].as_ref())
         );
 
         // The numbers in `Self::limit` assume a maximum packet size of `LIMIT`.
@@ -653,14 +658,12 @@ impl CryptoDxState {
         }
         self.invoked()?;
 
-        let size = body.len() + MAX_AUTH_TAG;
-        let mut out = vec![0; size];
-        let res = self.aead.encrypt(pn, hdr, body, &mut out)?;
+        let len = self.aead.encrypt_in_place(pn, hdr, body, data)?;
 
-        qtrace!("[{self}] encrypt ct={}", hex(res));
+        qtrace!("[{self}] encrypt ct={}", hex(data));
         debug_assert_eq!(pn, self.next_pn());
         self.used(pn)?;
-        Ok(res.to_vec())
+        Ok(len)
     }
 
     #[must_use]


larseggert:
@mxinden when you have a moment, would you take a look at the borrow-checker issue in `PublicPacket::decode`? It's the last one I couldn't figure out how to address.

mxinden:
Took a quick look.

``` rust
            let dcid = Self::opt(dcid_decoder.decode_cid(&mut decoder))?;
            if decoder.remaining() < SAMPLE_OFFSET + SAMPLE_SIZE {
                return Err(Error::InvalidPacket);
            }
            let header_len = decoder.offset();
            return Ok((
                Self {
                    packet_type: PacketType::Short,
                    dcid,
                    scid: None,
                    token: &[],
                    header_len,
                    version: None,
                    data,
                },
                &[],
                &mut [],
            ));
```

- `decoder` has an immutable reference to `data`.
- Through `decoder` `dcid` has an immutable reference to `data`.
- The function returns both `dcid` (immutable reference to `data` AND `data`. In other words, it returns both an immutable and a mutable reference to `data`, which is disallowed.

I can take a deeper look and try to fix it.

larseggert:
Thanks for the analysis! Wonder if we can make `dcid` a `Range` into `data`...

mxinden:
The above described issue, namely that of `dcid` and `data` being a conflicting (im-) mutable borrow, can be fixed by allocating `dcid` on the stack via an owned `ConnectionId` backed by a `SmallVec`:

``` diff
diff --git a/neqo-transport/src/packet/mod.rs b/neqo-transport/src/packet/mod.rs
index 73b47bcc..779ca72b 100644
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -563,7 +563,7 @@ pub struct PublicPacket<'a> {
     /// The packet type.
     packet_type: PacketType,
     /// The recovered destination connection ID.
-    dcid: ConnectionIdRef<'a>,
+    dcid: ConnectionId,
     /// The source connection ID, if this is a long header packet.
     scid: Option<ConnectionIdRef<'a>>,
     /// Any token that is included in the packet (Retry always has a token; Initial sometimes
```

That leaves us with another issue, namely rustc not being able to infer that early returns of `data` don't interfere with the final return of `remainder`. I have an idea which I will explore tomorrow.

mxinden:
Okay, I got it.

Let's take a look at `PacketBuilder` on `main`:

``` rust
/// `PublicPacket` holds information from packets that is public only.  This allows for
/// processing of packets prior to decryption.
pub struct PublicPacket<'a> {
    /// The packet type.
    packet_type: PacketType,
    /// The recovered destination connection ID.
    dcid: ConnectionIdRef<'a>,
    /// The source connection ID, if this is a long header packet.
    scid: Option<ConnectionIdRef<'a>>,
    /// Any token that is included in the packet (Retry always has a token; Initial sometimes
    /// does). This is empty when there is no token.
    token: &'a [u8],
    /// The size of the header, not including the packet number.
    header_len: usize,
    /// Protocol version, if present in header.
    version: Option<WireVersion>,
    /// A reference to the entire packet, including the header.
    data: &'a [u8],
}
```

`dcid`, `scid`, `token` and `data` are all **immutable** references into the same underlying memory allocation, here our long lived receive buffer.

This pull request introduces the following change:

``` diff
@@ -564,7 +574,7 @@ pub struct PublicPacket<'a> {
     /// Protocol version, if present in header.
     version: Option<WireVersion>,
     /// A reference to the entire packet, including the header.
-    data: &'a [u8],
+    data: &'a mut [u8],
 }
```

While `dcid`, `scid` and `token` are untouched, `data` is now a mutable reference. Having both immutable and mutable references to the same memory allocation is illegal, thus the compiler error.

An easy fix would be to make `dcid`, `scid` and `token` owned types. Given their small footprint, this is likely fine. There might be some additional optimizations, but I doubt they are worth it.

``` diff
diff --git a/neqo-transport/src/packet/mod.rs b/neqo-transport/src/packet/mod.rs
index 73b47bcc..dc85bbd0 100644
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -563,12 +563,12 @@ pub struct PublicPacket<'a> {
     /// The packet type.
     packet_type: PacketType,
     /// The recovered destination connection ID.
-    dcid: ConnectionIdRef<'a>,
+    dcid: ConnectionId,
     /// The source connection ID, if this is a long header packet.
-    scid: Option<ConnectionIdRef<'a>>,
+    scid: Option<ConnectionId>,
     /// Any token that is included in the packet (Retry always has a token; Initial sometimes
     /// does). This is empty when there is no token.
-    token: &'a [u8],
+    token: Vec<u8>,
     /// The size of the header, not including the packet number.
     header_len: usize,
     /// Protocol version, if present in header.
```

The above, plus a couple of smaller lifetime changes resolve the borrow checker issues.

I will propose a commit with my local changes.

mxinden:
@larseggert let me know what you think of https://github.com/larseggert/neqo/pull/34.

Note that it only addresses the borrow-checker issues in `neqo-transport/src/packet/mod.rs`.

Happy to look at the `neqo-http3` failures as well.

neqo-crypto/src/aead.rs
```diff
@@ -126,6 +126,39 @@ impl RealAead {
...
+    ///
+    /// The space provided in `data` needs to allow `Aead::expansion` more bytes to be appended.
+    ///
+    /// # Errors
+    ///
+    /// If the input can't be protected or any input is too large for NSS.
+    pub fn encrypt_in_place(
+        &self,
+        count: u64,
+        aad: Range<usize>,
```
martinthomson, line 139: I think that it would be better to take a slice for the AAD.  Then it could be kept elsewhere, if that makes more sense.

neqo-crypto/src/aead.rs
```diff
@@ -126,6 +126,39 @@ impl RealAead {
         Ok(&output[0..(l.try_into()?)])
     }
 
+    /// Encrypt `data` consisting of `aad` and plaintext `input` in place.
+    ///
+    /// The space provided in `data` needs to allow `Aead::expansion` more bytes to be appended.
+    ///
+    /// # Errors
```
martinthomson, line 133: This needs panics docs.  You are taking a range argument and indexing a slice argument.  That's going to break.

neqo-crypto/src/aead.rs
```diff
@@ -126,6 +126,39 @@ impl RealAead {
...
+    /// If the input can't be protected or any input is too large for NSS.
+    pub fn encrypt_in_place(
+        &self,
+        count: u64,
+        aad: Range<usize>,
+        input: Range<usize>,
+        data: &mut [u8],
+    ) -> Res<usize> {
+        let aad = &data[aad];
+        let input = &data[input];
```
martinthomson, line 144: I would be happier if `input` is required to be the plaintext plus `self.expansion()` in length, then you would construct the cipher input as `&mut input[..input.len() - self.expansion()]`.  If you do that, you can drop the `usize` value from the return (success means that you filled the buffer).  The cost of that is that the caller needs to understand how much padding to include, which is pretty-much already the case anyway.

neqo-crypto/src/aead.rs
```diff
@@ -158,6 +191,41 @@ impl RealAead {
         }?;
         Ok(&output[0..(l.try_into()?)])
     }
+
+    /// Decrypt a ciphertext in place.
+    ///
+    /// Note that NSS insists upon having extra space available for decryption, so
+    /// the buffer for `output` should be the same length as `input`, even though
+    /// the final result will be shorter.
```
martinthomson, line 199: I would move this note to a comment on the last argument to the function call below.  This API doesn't need to bother API users with this particular detail.

neqo-crypto/src/aead.rs
```diff
@@ -158,6 +191,41 @@ impl RealAead {
...
+    ///
+    /// # Errors
+    ///
+    /// If the input isn't authenticated or any input is too large for NSS.
+    pub fn decrypt_in_place(
+        &self,
+        count: u64,
+        aad: Range<usize>,
+        input: Range<usize>,
+        data: &mut [u8],
```
martinthomson, line 209: As before, I would prefer a slice for the AAD.  However, in this case, you can insulate the caller from knowing about `self.expansion()`, so returning the `usize` is maybe OK (or you could return a slice).

neqo-transport/src/packet/mod.rs
```diff
@@ -418,9 +418,22 @@ impl PacketBuilder {
             hex(hdr),
             hex(body)
         );
-        let ciphertext = crypto.encrypt(self.pn, hdr, body)?;
+
+        // Add space for crypto expansion.
+        let data_end = self.encoder.len();
+        for _i in 0..crypto.expansion() {
+            self.encode_byte(123);
+        }
```
martinthomson, line 426: We can do better than this.  There is a `pad_to()` function that can be used.  Or you can define `pad_by()`.

neqo-transport/src/packet/mod.rs
```diff
@@ -839,14 +860,18 @@ impl<'a> PublicPacket<'a> {
...
-            &self.data[self.header_len + pn_len..],
+            self.header_len + pn_len..self.data.len(),
         ))
     }
 
     /// # Errors
     ///
     /// This will return an error if the packet cannot be decrypted.
-    pub fn decrypt(&self, crypto: &mut CryptoStates, release_at: Instant) -> Res<DecryptedPacket> {
+    pub fn decrypt(
```
martinthomson, line 870: I'm looking at the calling code and I *think* that we can change this to take `self` instead of `&mut self`.  There is some use of the public packet after decryption, but it seems like the necessary pieces can be moved to the decrypted packet.

martinthomson:
FWIW, I looked at the `aead` crate interface [here](https://docs.rs/aead/0.5.2/aead/trait.AeadMutInPlace.html).  I don't think that is going to fit our needs very well.

neqo-crypto/src/aead.rs
larseggert: I'm not quite sure how to do this without copying `aad`, since there'd then be two references into `data`?

martinthomson: You will have a `&mut [u8]` that can be split (in this case temporarily) into an AAD and plaintext or ciphertext part. See https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_at_mut

neqo-transport/src/packet/mod.rs
larseggert: I'd suggest a separate PR for this.

larseggert:
> FWIW, I looked at the `aead` crate interface [here](https://docs.rs/aead/0.5.2/aead/trait.AeadMutInPlace.html). I don't think that is going to fit our needs very well.

I actually think they might? Our API after the last round of changes is very similar. (Or I'm missing something.)

neqo-crypto/src/aead.rs
```diff
@@ -156,7 +180,38 @@ impl RealAead {
                 c_uint::try_from(output.len())?,
             )
         }?;
-        Ok(&output[0..(l.try_into()?)])
+        Ok(&output[..l.try_into()?])
+    }
+
+    /// Decrypt a ciphertext in place.
+    ///
```
martinthomson, line 187:
```suggestion
    /// Returns a subslice of `data` (without the last `<self as Aead>::expansion()` bytes),
    /// that has been decrypted in place.
    ///
```

neqo-crypto/src/aead.rs
```diff
@@ -156,7 +180,38 @@ impl RealAead {
...
+                aad.as_ptr(),
+                c_uint::try_from(aad.len())?,
+                data.as_ptr(),
+                c_uint::try_from(data.len())?,
+                data.as_ptr(),
+                &mut l,
+                c_uint::try_from(data.len())?,
+            )
+        }?;
+        Ok(&mut data[..l.try_into()?])
```
martinthomson, line 214:
```suggestion
        debug_assert_eq!(usize::try_from(l), data.len() - self.expansion());
        Ok(&mut data[..l.try_into()?])
```

neqo-crypto/src/aead.rs
```diff
@@ -123,14 +123,35 @@ impl RealAead {
...
     ///
-    /// Note that NSS insists upon having extra space available for decryption, so
-    /// the buffer for `output` should be the same length as `input`, even though
-    /// the final result will be shorter.
+    /// The space provided in `data` needs to allow `Aead::expansion` more bytes to be appended.
+    ///
+    /// # Errors
+    ///
+    /// If the input can't be protected or any input is too large for NSS.
+    pub fn encrypt_in_place(&self, count: u64, aad: &[u8], data: &[u8]) -> Res<usize> {
```
martinthomson, line 136: Given that `decrypt_in_place` returns `&'a mut [u8]`, should this do the same?

IMPORTANT: You need to do this, at a minimum.

```suggestion
    pub fn encrypt_in_place(&self, count: u64, aad: &[u8], data: &mut [u8]) -> Res<usize> {
```

Though with the above...


```suggestion
    pub fn encrypt_in_place<'a>(&self, count: u64, aad: &[u8], data: &'a mut [u8]) -> Res<&'a mut [u8]> {
```

neqo-crypto/src/aead_null.rs
```diff
@@ -42,36 +42,60 @@ impl AeadNull {
...
+        Ok(data.len())
+    }
+
+    fn decrypt_check(&self, _count: u64, _aad: &[u8], input: &[u8]) -> Res<usize> {
+        if input.len() < self.expansion() {
             return Err(Error::from(SEC_ERROR_BAD_DATA));
         }
 
-        let len_encrypted = input.len() - AEAD_NULL_TAG.len();
+        let len_encrypted = input.len() - self.expansion();
```
martinthomson, line 61: If you change this to a checked subtraction, you can use 

```suggestion
        let len_encrypted = input.len().checked_sub(self.expansion())
          .ok_or(Error::from(SEC_ERROR_BAD_DATA))?;
```

neqo-transport/src/packet/mod.rs
```diff
@@ -611,69 +614,73 @@ impl<'a> PublicPacket<'a> {
...
     /// This will return an error if the packet could not be decoded.
     #[allow(clippy::similar_names)]
-    pub fn decode(data: &'a [u8], dcid_decoder: &dyn ConnectionIdDecoder) -> Res<(Self, &'a [u8])> {
+    pub fn decode(
+        data: &'a mut [u8],
+        dcid_decoder: &dyn ConnectionIdDecoder,
+    ) -> Res<(Self, &'a mut [u8])> {
         let mut decoder = Decoder::new(data);
-        let first = Self::opt(decoder.decode_uint::<u8>())?;
+        let first = PublicPacket::opt(decoder.decode_uint::<u8>())?;
```
martinthomson, line 622: Am I to guess that this needs to be `PublicPacket` and not `Self` because the lifetimes are different?  Subtle shit like that is where rust loses points in my book.

martinthomson (approved):
I have a few gotchas here, but I like how this is shaping up.

The real question I have is: does this really make it go faster?  The benchmarks show some improvements.  Are those consistent enough for you to be happy?  I see that the improvements aren't 100% consistent.

neqo-transport/src/packet/mod.rs
larseggert: I was able to revert all but one of those changes. Reverting the one that is left makes the borrow checker freak out.

larseggert:
> I have a few gotchas here, but I like how this is shaping up.
> 
> The real question I have is: does this really make it go faster? The benchmarks show some improvements. Are those consistent enough for you to be happy? I see that the improvements aren't 100% consistent.

I see a few percentage points (3-5%) locally. It's not a lot; I guess those extra heap allocations aren't that costly.
```
                        time:   [293.84 ms 294.75 ms 295.65 ms]
                        thrpt:  [338.24 MiB/s 339.27 MiB/s 340.32 MiB/s]
                 change:
                        time:   [-4.0668% -3.3831% -2.8379%] (p = 0.00 < 0.05)
                        thrpt:  [+2.9208% +3.5015% +4.2392%]
                        Performance has improved.
  1 (1.00%) high mild
```

larseggert:
I'm kinda surprised the `Upload` test isn't seeing any improvement though.

martinthomson:
It's quite possible that the work we do to generate a packet is still dominated by other factors.  There is a non-significant improvement according to the runs, maybe try running it 10x more to get the noise down some more.

PR #2387, opened 2025-01-23
feat: Option to turn off SNI slicing

larseggert:
Wireshark can't reassemble sliced CRYPTO frames, which causes QNS tests to fail bcause it then can't parse all packets.

This PR adds an option to disable SNI slicing, and we do so by default when running in QNS.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -241,10 +241,11 @@ impl Args {
         // This is the default for all tests except http3.
         self.shared.alpn = String::from("hq-interop");
         // Wireshark can't reassemble sliced CRYPTO frames, which causes tests to fail.
-        // So let's turn that off.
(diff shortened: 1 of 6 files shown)


mxinden:
That is a bummer. Especially since this gave us a signal whether all server implementations supported the sni-slicing feature. Do you see this being an issue? Is this temporary?

larseggert:
I'd be very surprised if major servers wouldn't support it.

We can file an issue with wireshark, but it's also kind of a feature when our connections can't be parsed :-)

mxinden:
> but it's also kind of a feature when our connections can't be parsed :-)

Haha, true.

mxinden (approved):

larseggert:
@mxinden turns out we can leave slicing on for some of the tests, so we will get alerts if some servers don't deal with it well.

mxinden:
Good news. Thanks.

PR #2388, opened 2025-01-24
chore: Invert SNI slicing option to `--no-sni-slicing`

larseggert:
To be consistent with other options, and so we can actually turn it off.

Also add it to `test.sh`, so that the `tshark` parsing works.


mxinden (commented):
:+1: on the consistency.

Regarding turning the previous flag off, didn't `--sni-slicing=false` work?

mxinden (approved):

larseggert:
`error: unexpected value 'false' for '--sni-slicing' found; no more were expected`

PR #2396, opened 2025-01-29
chore: Factor out packet logging

larseggert:
Because `output_path` and `input_path` are getting long, and there is a lot of redundancy between the calls to `dump` and `qlog`.


neqo-transport/src/qlog.rs
```diff
@@ -296,23 +296,18 @@ pub fn packets_lost(qlog: &NeqoQlog, pkts: &[SentPacket], now: Instant) {
 
 pub fn packet_received(
     qlog: &NeqoQlog,
-    public_packet: &PublicPacket,
-    payload: &DecryptedPacket,
+    pt: PacketType,
+    pn: PacketNumber,
+    plen: usize,
+    body: &[u8],
```
martinthomson, line 302: Would it make sense to build a struct for this chunk of values?  That might help with your clippy warning (which I think should be fixed).

neqo-transport/src/connection/dump.rs
```diff
@@ -50,3 +55,26 @@ pub fn dump_packet(
...
+
+#[allow(clippy::too_many_arguments)]
+pub fn log_packet(
+    conn: &Connection,
+    path: &PathRef,
+    tos: IpTos,
+    dir: &Direction,
+    pt: PacketType,
+    pn: PacketNumber,
+    plen: usize,
```
martinthomson, line 67: IMPORTANT: This is the length of the datagram, not the packet.

martinthomson (commented):
Just a few nits.  I think that we were mixing up some lengths, so maybe look into that more.

neqo-transport/src/connection/dump.rs
```diff
@@ -50,3 +55,26 @@ pub fn dump_packet(
         path.borrow()
     );
 }
+
+#[allow(clippy::too_many_arguments)]
+pub fn log_packet(
+    conn: &Connection,
+    path: &PathRef,
+    tos: IpTos,
+    dir: &Direction,
```
martinthomson, line 64: An idea occurred as I was hitting "submit"...

It might be possible to move this direction to a marker trait.  Imagine if `QlogDirection` were a trait that was implemented by `Qlog{Send|Receive}Direction` with `fn qlog(...args) { qlog::packet_sent(...args) }` or `fn qlog(...args) { qlog::packet_received(...args) }` as necessary.  Then you could `use QLogSendDirection as Tx` and call this function as `log_packet::<Tx>(...args)`.  Just a thought.

neqo-transport/src/connection/mod.rs
```diff
@@ -3545,6 +3528,27 @@ impl Connection {
     pub fn plpmtu(&self) -> usize {
         self.paths.primary().unwrap().borrow().plpmtu()
     }
+
+    fn log_packet(&self, meta: packet::MetaData, now: Instant) {
+        if log::STATIC_MAX_LEVEL == log::LevelFilter::Off || !log::log_enabled!(log::Level::Debug) {
```
martinthomson, line 3533: I think that the first check here is redundant:

https://docs.rs/log/0.4.25/src/log/macros.rs.html#218

```suggestion
        if !log::log_enabled!(log::Level::Debug) {
```

neqo-transport/src/qlog.rs
```diff
@@ -234,18 +226,31 @@ pub fn packet_sent(
...
+                    is_coalesced: None,
+                    retry_token: None,
+                    stateless_reset_token: None,
+                    supported_versions: None,
+                    raw: Some(raw),
+                    datagram_id: None,
+                    send_at_time: None,
+                    trigger: None,
+                })),
+                Direction::Rx => Some(EventData::PacketReceived(PacketReceived {
```
martinthomson, line 242: Does `PacketReceived` implement `Default`?  Because it seems like you could replace a bunch of this code with `..PacketReceived::default()`.

martinthomson (approved):
I like this.  A little more work, sure, but it is a lot nicer to look at.

neqo-transport/src/qlog.rs
larseggert: I had the same thought, but it doesn't. Maybe I'll do a PR against qlog...

larseggert: https://github.com/cloudflare/quiche/pull/1931

PR #2404, opened 2025-02-02
ci: Add `ubuntu-24.04-arm` to the `check` matrix

### LINKED ISSUE #2401: Add `ubuntu-arm` runner
Which should be faster, they are only in preview and so availability isn't quite there yet (jobs get queued for quite a while).

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

larseggert:
Fixes #2401


larseggert:
Nothing is ever easy :-)

PR #2413, opened 2025-02-03
ci: Check CI workflows with `zizmor`

larseggert:
And fix the security issues it found.


larseggert:
Need to commit this to `main` to verify that the actions are correctly running. PR is otherwise running actions from `main` without the needed changes :-(

PR #2423, opened 2025-02-05
fix: Reduce `bench` transfer sizes by about 1/3

larseggert:
Because they take many minutes to run on the bencher.

### DIFF
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -29,13 +29,13 @@ fn transfer(c: &mut Criterion) {
         upload,
     } in [
         Benchmark {
-            name: format!("1-conn/1-100mb-resp{mtu} (aka. Download)"),
-            requests: vec![100 * 1024 * 1024],
+            name: format!("1-conn/1-64mb-resp{mtu} (aka. Download)"),
+            requests: vec![64 * 1024 * 1024],
             upload: false,
         },
         Benchmark {
-            name: format!("1-conn/10_000-parallel-1b-resp{mtu} (aka. RPS)"),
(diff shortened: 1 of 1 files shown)


mxinden:
As long as criterion's `p` value doesn't change noticeably, I am fine with this change.

neqo-bin/benches/main.rs
```diff
@@ -29,13 +29,13 @@ fn transfer(c: &mut Criterion) {
...
-            name: format!("1-conn/1-100mb-resp{mtu} (aka. Download)"),
-            requests: vec![100 * 1024 * 1024],
+            name: format!("1-conn/1-64mb-resp{mtu} (aka. Download)"),
+            requests: vec![64 * 1024 * 1024],
             upload: false,
         },
         Benchmark {
-            name: format!("1-conn/10_000-parallel-1b-resp{mtu} (aka. RPS)"),
-            requests: vec![1; 10_000],
+            name: format!("1-conn/8192-parallel-1b-resp{mtu} (aka. RPS)"),
```
martinthomson, line 37: You said 1/3, but this is closer to 1/6.  Would 6144 do?

martinthomson (approved):

neqo-bin/benches/main.rs
larseggert: Yeah, I changed this one as an afterthought.

larseggert:
I'm also talking to the SRE team about more bench runners. That way, it doesn't matter so much how long each run takes, if it doesn't block others.

larseggert:
Let's leave things as-is. I'm working with the SRE team to get more bench runners.

PR #2425, opened 2025-02-06
ci: Fix QNS log failure

larseggert:
OK, fixed. I managed to accidentally remove a backslash, which caused everything to explode. Put it back.

Also did a little bonus maintenance while I was here, by adding `tquic` and `openssl` to the interop matrix, and by using the correct highlighting for lines normally hidden under "All results".

I still hate GitHub CI.

### DIFF
--- a/.github/actions/quic-interop-runner/action.yml
+++ b/.github/actions/quic-interop-runner/action.yml
@@ -112,11 +112,8 @@ runs:
       env:
         ARTIFACT_URL: ${{ steps.upload-logs.outputs.artifact-url }}
       run: |
-        cat result.json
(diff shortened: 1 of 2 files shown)


.github/actions/quic-interop-runner/action.yml
```diff
@@ -112,9 +112,8 @@ runs:
       env:
         ARTIFACT_URL: ${{ steps.upload-logs.outputs.artifact-url }}
       run: |
-        jq ". + {log_url: \"$ARTIFACT_URL\"}"
```
larseggert, line 115: The whole issue is due to the missing backslash in this line 🤦

PR #2427, opened 2025-02-07
fix: only enable quinn-udp fast-apple-datapath on arm

mxinden:
Only enable `fast-apple-datapath` on aarch64.

Old MacOS versions (e.g. 10.15) don't fully support `sendmsg_x` and `recvmsg_x`, both enabled by `fast-apple-datapath`. Instead of feature gating by OS version, simply do so by CPU arch, leveraging the fact that newer MacOS devices use ARM.

---

Draft for now. [Debugging mozilla-central failure on MacOS 10.15.](https://bugzilla.mozilla.org/show_bug.cgi?id=1942325#c4)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ rust-version = "1.76.0"
 enum-map = { version = "2.7", default-features = false }
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.13", default-features = false }
-quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log", "fast-apple-datapath"] }
 regex = { version = "1.9", default-features = false, features = ["unicode-perl"] }
 static_assertions = { version = "1.1", default-features = false }
 url = { version = "2.5.3", default-features = false, features = ["std"] }
--- a/neqo-bin/Cargo.toml
+++ b/neqo-bin/Cargo.toml
@@ -38,7 +38,7 @@ neqo-http3 = { path = "./../neqo-http3" }
 neqo-transport = { path = "./../neqo-transport" }
 neqo-udp = { path = "./../neqo-udp" }
 qlog = { workspace = true }
-quinn-udp = { workspace = true }
+quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log"] }
 regex = { workspace = true }
 tokio = { version = "1", default-features = false, features = ["net", "time", "macros", "rt", "rt-multi-thread"] }
 url = { workspace = true }
--- a/neqo-udp/Cargo.toml
+++ b/neqo-udp/Cargo.toml
@@ -19,7 +19,16 @@ workspace = true
 # Checked against https://searchfox.org/mozilla-central/source/Cargo.lock 2024-11-11
 log = { workspace = true }
 neqo-common = { path = "./../neqo-common" }
-quinn-udp = { workspace = true }
+quinn-udp = { version = "0.5.6", default-features = false, features = ["direct-log"] }
+
+# Only enable `fast-apple-datapath` on aarch64.
+#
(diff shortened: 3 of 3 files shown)


larseggert:
This is odd, because IIRC these syscalls predate Catalina by years.

mxinden:
Thanks @larseggert for the info.

All suspicions for now. I just know it is related to Neqo `v0.12` and MacOS 10.15.

Might as well be a syscall from the `mtu` crate which we are introducing through Neqo `v0.12` as well.

(Thus far I have no other way to reproduce than Mozilla's CI, a single run take ~6h, and logs have not been helpful.)

I will continue to debug this.

mxinden:
For the record, Mozilla CI:

- failes with Neqo `v0.12.2`: https://treeherder.mozilla.org/jobs?repo=try&revision=0b1ba0bbf8036fe982fb727e4a6205aeb7ba513f&selectedTaskRun=PaCBY_IPRny5j2SyiWU4CQ.0
- succeeds with Neqo `v0.12.2` no `mtu` crate and no `fast-apple-datapath` feature: https://treeherder.mozilla.org/jobs?repo=try&revision=f62750f5262f198079717fa5a0aa1471c68bdcd3&selectedTaskRun=ecJ2FQs0Rky5FsAkTcSM4Q.0
- fails with Neqo `v0.12.2` no `mtu` crate: https://treeherder.mozilla.org/jobs?repo=try&revision=e1d384ee16dceed38e23dc6d0ac02f92705ea78b&selectedTaskRun=KGM7n4kaRZiHU6OT8i8ziA.0
- succeeds with Neqo `v0.12.2` and no `fast-apple-datapath` feature: https://treeherder.mozilla.org/jobs?repo=try&revision=c94781bd7362c8d47f6065af6a7426f040d6784c&selectedTaskRun=br1Tj1VkSwCklxBs1LYN2w.0
- succeeds with this pull request: https://treeherder.mozilla.org/jobs?repo=try&revision=c9b3083ebbc1c6c7470ca5c0c1b53e926073c1f3&selectedTaskRun=WVg1w7h2RDOyjq-U3xsiMA.0

larseggert:
@mxinden in case it helps, I have an old Intel MacBook laying around somewhere.

mxinden:
A datapoint for `sendmsg_x` failure on MacOS <= 13:

https://github.com/oven-sh/bun/commit/26526cba3828bdd82731ed30152625fe2b12f7f5#diff-8060260f9888a3a03c4d236f22769baacc4c9aec45abf7f49b5b2bbab7e39579R307

mxinden:
> @mxinden in case it helps, I have an old Intel MacBook laying around somewhere.

@larseggert that would be very helpful! Could you checkout the `quinn` project and run its unit tests with the `fast-apple-datapath` feature enabled:

```
cargo test -p quinn-udp --features fast-apple-datapath
```

Which MacOS version is it running?

mxinden:
That is good news. Thanks for reproducing. This matches what we are seeing on Mozilla's CI and above linked commit on oven-sh/bun.

mxinden:
Another reference to the issue: https://bun.sh/blog/bun-v1.1.22#udp-socket-hangs-on-macos-13

larseggert:
So what's happening is that in the receive path, the `cmsg_iter` loop in `decode_recv` never terminates, because the iterator returns a `cmsghdr` where all fields are zero, which hits the `_` case in the loop.

I wonder if `libc::CMSG_NXTHDR` is not correct for older macOS versions... Anyway, we can work around this in `quinn-udp`.

mxinden:
With https://github.com/quinn-rs/quinn/pull/2154 merged, released and landed in mozilla-central, this issue is resolved.

PR #2434, opened 2025-02-10
feat: Replace `HashMap` with `EnumMap`

larseggert:
In several places.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -55,7 +55,15 @@ use crate::{
     stats::{Stats, StatsCell},
     stream_id::StreamType,
     streams::{SendOrder, Streams},
-    tparams::{self, TransportParameterId, TransportParameters, TransportParametersHandler},
+    tparams::{
+        self,
+        TransportParameterId::{
+            self, AckDelayExponent, ActiveConnectionIdLimit, DisableMigration, GreaseQuicBit,
+            InitialSourceConnectionId, MaxAckDelay, MaxDatagramFrameSize, MinAckDelay,
+            OriginalDestinationConnectionId, RetrySourceConnectionId, StatelessResetToken,
+        },
+        TransportParameters, TransportParametersHandler,
+    },
     tracking::{AckTracker, PacketNumberSpace, RecvdPackets},
     version::{Version, WireVersion},
     AppError, CloseReason, Error, Res, StreamId,
@@ -373,10 +381,8 @@ impl Connection {
         let mut cid_manager =
             ConnectionIdManager::new(cid_generator, local_initial_source_cid.clone());
         let mut tps = conn_params.create_transport_parameter(role, &mut cid_manager)?;
-        tps.local.set_bytes(
-            TransportParameterId::InitialSourceConnectionId,
-            local_initial_source_cid.to_vec(),
-        );
+        tps.local
+            .set_bytes(InitialSourceConnectionId, local_initial_source_cid.to_vec());
 
         let tphandler = Rc::new(RefCell::new(tps));
         let crypto = Crypto::new(
@@ -498,7 +504,7 @@ impl Connection {
     #[cfg(test)]
     pub fn set_local_tparam(
         &self,
-        tp: tparams::TransportParameterId,
+        tp: TransportParameterId,
         value: tparams::TransportParameter,
     ) -> Res<()> {
         if *self.state() == State::Init {
@@ -525,14 +531,8 @@ impl Connection {
         qtrace!("[{self}] Retry CIDs: odcid={odcid} remote={remote_cid} retry={retry_cid}");
         // We advertise "our" choices in transport parameters.
         let local_tps = &mut self.tps.borrow_mut().local;
-        local_tps.set_bytes(
-            TransportParameterId::OriginalDestinationConnectionId,
-            odcid.to_vec(),
-        );
-        local_tps.set_bytes(
-            TransportParameterId::RetrySourceConnectionId,
-            retry_cid.to_vec(),
-        );
+        local_tps.set_bytes(OriginalDestinationConnectionId, odcid.to_vec());
+        local_tps.set_bytes(RetrySourceConnectionId, retry_cid.to_vec());
 
         // ...and save their choices for later validation.
         self.remote_initial_source_cid = Some(remote_cid);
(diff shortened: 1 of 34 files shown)


neqo-transport/src/tparams.rs
```diff
@@ -27,28 +27,72 @@ use crate::{
...
-pub const ACTIVE_CONNECTION_ID_LIMIT: TransportParameterId = 0x0e;
-pub const INITIAL_SOURCE_CONNECTION_ID: TransportParameterId = 0x0f;
-pub const RETRY_SOURCE_CONNECTION_ID: TransportParameterId = 0x10;
-pub const VERSION_INFORMATION: TransportParameterId = 0x11;
-pub const GREASE_QUIC_BIT: TransportParameterId = 0x2ab2;
-pub const MIN_ACK_DELAY: TransportParameterId = 0xff02_de1a;
-pub const MAX_DATAGRAM_FRAME_SIZE: TransportParameterId = 0x0020;
+#[allow(clippy::enum_clike_unportable_variant)]
+#[derive(Debug, Clone, Enum, PartialEq, Eq)]
+pub enum TransportParameterId {
```
mxinden, line 32: Very much in favor of this `enum`!

neqo-transport/src/tparams.rs
```diff
@@ -128,7 +172,10 @@ pub enum TransportParameter {
 
 impl TransportParameter {
     fn encode(&self, enc: &mut Encoder, tp: TransportParameterId) {
-        qtrace!("TP encoded; type 0x{tp:02x} val {self:?}");
+        qtrace!(
+            "TP encoded; type {tp:?}(0x{:02x}) val {self:?}",
+            u64::from(tp.clone())
+        );
```
mxinden, line 178: Instead of `u64::from(tp.clone())`, how about implementing `Display` for `TransportParameterId`.

Also, given that `TransportParameterId` is a [field-less `enum`](https://doc.rust-lang.org/reference/items/enumerations.html#r-items.enum.fieldless) it can `derive` `Copy`, thus no need for the above call to `.clone()`.

mxinden (approved):
Thanks for the cleanup!

neqo-transport/src/connection/mod.rs
```diff
@@ -525,8 +525,14 @@ impl Connection {
         qtrace!("[{self}] Retry CIDs: odcid={odcid} remote={remote_cid} retry={retry_cid}");
         // We advertise "our" choices in transport parameters.
         let local_tps = &mut self.tps.borrow_mut().local;
-        local_tps.set_bytes(tparams::ORIGINAL_DESTINATION_CONNECTION_ID, odcid.to_vec());
-        local_tps.set_bytes(tparams::RETRY_SOURCE_CONNECTION_ID, retry_cid.to_vec());
+        local_tps.set_bytes(
+            TransportParameterId::OriginalDestinationConnectionId,
```
martinthomson, line 529: I wonder whether we should change the use statement at the top so that we can write: `OriginalDestinationConnectionId` rather than `TransportParameterId::OriginalDestinationConnectionId` (which takes a whole line on its own).

neqo-transport/src/connection/mod.rs
```diff
@@ -2621,7 +2627,7 @@ impl Connection {
             .tps
             .borrow()
             .remote()
-            .get_integer(tparams::IDLE_TIMEOUT);
+            .get_integer(TransportParameterId::IdleTimeout);
```
martinthomson, line 2630: This might be the one transport parameter that requires the prefix.

neqo-transport/src/crypto.rs
```diff
@@ -821,6 +824,10 @@ pub struct CryptoStates {
 }
 
 impl CryptoStates {
+    fn initials_is_empty(&self) -> bool {
+        self.initials.iter().all(|(_, v)| v.is_none())
```
martinthomson, line 828: You can use `Iterator::flatten` and `Iterator::count` with `EnumMap::values` for this.  I doubt that the code it generates would be more efficient, but it might be cheaper.

neqo-transport/src/qlog.rs
```diff
@@ -31,53 +31,80 @@ use crate::{
...
+            #[allow(clippy::cast_possible_truncation)] // Nope.
+            let ev_data =
+                EventData::TransportParametersSet(qlog::events::quic::TransportParametersSet {
+                    owner: None,
+                    resumption_allowed: None,
+                    early_data_enabled: None,
+                    tls_cipher: None,
+                    aead_tag_length: None,
+                    original_destination_connection_id: remote
+                        .get_bytes(TransportParameterId::OriginalDestinationConnectionId)
```
martinthomson, line 51: Another case where the extra long name might not be so useful.

neqo-transport/src/tparams.rs
```diff
@@ -27,28 +27,72 @@ use crate::{
...
-pub const DISABLE_MIGRATION: TransportParameterId = 0x0c;
-pub const PREFERRED_ADDRESS: TransportParameterId = 0x0d;
-pub const ACTIVE_CONNECTION_ID_LIMIT: TransportParameterId = 0x0e;
-pub const INITIAL_SOURCE_CONNECTION_ID: TransportParameterId = 0x0f;
-pub const RETRY_SOURCE_CONNECTION_ID: TransportParameterId = 0x10;
-pub const VERSION_INFORMATION: TransportParameterId = 0x11;
-pub const GREASE_QUIC_BIT: TransportParameterId = 0x2ab2;
-pub const MIN_ACK_DELAY: TransportParameterId = 0xff02_de1a;
-pub const MAX_DATAGRAM_FRAME_SIZE: TransportParameterId = 0x0020;
+#[allow(clippy::enum_clike_unportable_variant)]
```
martinthomson, line 30:
```suggestion
#[repr(u64)]
```

Suppressing the warning seems fine, but you could instead acknowledge the true size of this type.

neqo-transport/src/version.rs
```diff
@@ -6,13 +6,14 @@
 
 #![allow(clippy::module_name_repetitions)]
 
+use enum_map::Enum;
```
martinthomson, line 9: Unrelated?

martinthomson (approved):

neqo-transport/src/version.rs
larseggert: No, I need it for the `derive` below.

PR #2436, opened 2025-02-11
fix: Only try SNI slicing at offset 0

larseggert:
And also check that the SNI length makes sense. And rename some files for clarity.

This should fix the spurios failures for `compatible_upgrade_large_initial`.


neqo-transport/src/shuffle.rs
```diff
@@ -121,6 +124,19 @@ mod tests {
...
     }
 
+    #[test]
+    fn find_sni_invalid_sni_length() {
+        // ClientHello with an SNI extension with an invalid length
+        let mut buf = Vec::from(BUF_WITH_SNI);
+        let len = buf.len();
+
+        assert!(buf[len - 23] == 0x00 && buf[len - 22] == 0x0c); // Check Server Name List length
+                                                                 // Set Server Name List length to 0
```
mxinden, line 134: Bad formatting?

neqo-transport/src/shuffle.rs
```diff
@@ -48,6 +48,9 @@ pub fn find_sni(buf: &[u8]) -> Option<Range<usize>> {
         if ext_type == 0 {
             // SNI!
             let sni_len: u16 = dec.decode_uint()?;
```
martinthomson, line 50: Driveby:

```suggestion
            let sni_len = dec.decode_uint::<u16>()?;
```

I find the type annotation to be action at a distance, which makes the turbofish the more explicit option.  (I'm not myself free from this particular blemish; this is more characters, so I might have exploited type inference in the past, but no time like the present to start rectifying it :)

neqo-transport/src/shuffle.rs
martinthomson: I'm more concerned about the constants you have littered around.  22 and 23 here, 15 and 39 above.  Still, I can't see how that would be easy to fix.

neqo-transport/src/shuffle.rs
```diff
@@ -121,6 +124,19 @@ mod tests {
...
+    #[test]
+    fn find_sni_invalid_sni_length() {
+        // ClientHello with an SNI extension with an invalid length
+        let mut buf = Vec::from(BUF_WITH_SNI);
+        let len = buf.len();
+
+        assert!(buf[len - 23] == 0x00 && buf[len - 22] == 0x0c); // Check Server Name List length
+                                                                 // Set Server Name List length to 0
+        buf[len - 23] = 0x00;
+        buf[len - 22] = 0x00;
```
martinthomson, line 136: Zero length seems a bit much.  Why not make it 2?

neqo-transport/src/shuffle.rs
```diff
@@ -121,6 +124,19 @@ mod tests {
         assert!(super::find_sni(truncated).is_none());
     }
 
+    #[test]
+    fn find_sni_invalid_sni_length() {
+        // ClientHello with an SNI extension with an invalid length
+        let mut buf = Vec::from(BUF_WITH_SNI);
+        let len = buf.len();
+
+        assert!(buf[len - 23] == 0x00 && buf[len - 22] == 0x0c); // Check Server Name List length
```
martinthomson, line 133:
```suggestion
        assert_eq!(&buf[len-23..len-22], &[0x00, 0x0c], "check SNI length");
```

...will fix your odd comment shape at the same time

neqo-transport/src/shuffle.rs
martinthomson, line 136:
```suggestion
        buf[len - 22] = 0x02; // leave buf[len-23] unchanged at 0x00
```

neqo-transport/src/crypto.rs
```diff
@@ -1575,7 +1575,7 @@ impl CryptoStreams {
         let Some((offset, data)) = cs.tx.next_bytes() else {
             return;
         };
-        let written = if sni_slicing {
+        let written = if sni_slicing && offset == 0 {
```
martinthomson, line 1578: Good!  Test?

larseggert: I'm struggling a bit how to structure that test. Got any suggestions?

martinthomson: So in our ordinary setup with splitting, we'll send out two packets.  If we lose the right one of those (I think we reorder, so we'd want to lose the first), we'll have delivered the first part of the handshake data.  Then, we'll be forced to retransmit, which means that the offset will be non-zero when we hit this code again.

To verify that we've not split, we can verify that we get just one frame, not two.  Does that work?

larseggert: It almost works, except for the verification bit, because that RTX'ed chunk of crypto frame will not look like a valid CH and so the slicing logic will always exit and leave one slice, even if it was called. We'd need to feed in something at a non-zero offset that is a valid-looking CH. Hm...

martinthomson: Right, so the test would not have failed prior to you adding this check.

Except that it would have, just not reliably.  And I think that's OK.  If the test won't fail without this check, I'm sure that cargo mutants will complain and we can dismiss that, but I'd rather have the test and not have it reliably able to catch the problem than not have anything.

Thinking more on this, the odds of hitting this issue is probably higher than you would think.  We look for a more or less fixed ClientHello shape.  There's a few things where we skip vectors, but there we only need to ensure that we don't overrun the buffer, so there will be pure noise that will pass those tests.  The real check is where we hit the SNI extension parsing.  If we have something that is [0, 0, x, y] where x*256+y is between 3 and the remaining length of the packet, we will treat that as SNI.  (We do not check the SNI type or inner length, so this is looser than you might think.)  We also get a few swipes at that, jumping forward every time we see [?, ?, x, y].

So while there's a good chance that we'll skip past the end of the buffer and believe there is no SNI, the odds of us hitting this sequence is not as low as I had thought.  It's not as high as I'd like, but with a padding extension in play still, it's going to be non-trivial.

neqo-transport/src/crypto.rs
```diff
@@ -1562,51 +1562,53 @@ impl CryptoStreams {
...
-                let limit = data.len() / packets_needed;
-                let ((left_offset, left), (right_offset, right)) =
-                    limit_chunks((offset, left), (offset + mid as u64, right), limit);
-                (
-                    write_chunk(right_offset, right, builder),
-                    write_chunk(left_offset, left, builder),
-                )
+        while let Some((offset, data)) = cs.tx.next_bytes() {
+            let written = if sni_slicing && offset == 0 {
+                qdebug!("XXX SNI slicing enabled");
```
martinthomson, line 1576: Oops

neqo-transport/src/shuffle.rs
```diff
@@ -145,4 +147,19 @@ mod tests {
         let buf = [1; 1];
         assert!(super::find_sni(&buf).is_none());
     }
+
+    #[test]
```
martinthomson, line 151: I'd stick some commentary in here to cover this better.

martinthomson (approved):

larseggert:
Moving this to draft until I can add another test.

PR #2438, opened 2025-02-11
chore: Convert `FrameType` to an enum

larseggert:


neqo-transport/src/ackrate.rs
```diff
@@ -43,7 +43,7 @@ impl AckRate {
 
     pub fn write_frame(&self, builder: &mut PacketBuilder, seqno: u64) -> bool {
         builder.write_varint_frame(&[
-            FRAME_TYPE_ACK_FREQUENCY,
+            FrameType::AckFrequency.into(),
```
martinthomson, line 46:
```suggestion
            u64::from(FrameType::AckFrequency),
```

Just a preference for not hiding types.

neqo-transport/src/connection/tests/resumption.rs
```diff
@@ -129,7 +129,7 @@ fn ticket_rtt(rtt: Duration) -> Duration {
     dec.skip_vvec(); // Skip over the payload.
 
     // Replace the ACK frame with PADDING.
-    plaintext[..ACK_FRAME_1.len()].fill(FRAME_TYPE_PADDING.try_into().unwrap());
+    plaintext[..ACK_FRAME_1.len()].fill(FrameType::Padding.into());
```
martinthomson, line 132:
```suggestion
    plaintext[..ACK_FRAME_1.len()].fill(u64::from(FrameType::Padding));
```

Of course, I think that this is now really `u8::try_from()` instead, because this is `&mut [u8]`.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -610,7 +607,12 @@ fn legal_out_of_order_frame_on_remote_initiated_closed_stream() {
     // Deliver an out-of-order `FRAME_TYPE_MAX_STREAM_DATA` on forgotten stream.
     let dgram = send_with_extra(
         &mut client,
-        Writer(vec![FRAME_TYPE_MAX_STREAM_DATA, stream_id.as_u64(), 0, 0]),
+        Writer(vec![
+            FrameType::MaxStreamData.into(),
```
martinthomson, line 611:
```suggestion
            u64::from(FrameType::MaxStreamData),
```

neqo-transport/src/frame.rs
```diff
@@ -18,68 +18,152 @@ use crate::{
...
-pub const FRAME_TYPE_ACK_FREQUENCY: FrameType = 0xaf;
-// draft-ietf-quic-datagram
-pub const FRAME_TYPE_DATAGRAM: FrameType = 0x30;
-pub const FRAME_TYPE_DATAGRAM_WITH_LEN: FrameType = 0x31;
-const DATAGRAM_FRAME_BIT_LEN: u64 = 0x01;
-
-const STREAM_FRAME_BIT_FIN: u64 = 0x01;
-const STREAM_FRAME_BIT_LEN: u64 = 0x02;
-const STREAM_FRAME_BIT_OFF: u64 = 0x04;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
martinthomson, line 21:
```suggestion
#[repr(u64)]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```

neqo-transport/src/frame.rs
```diff
@@ -18,68 +18,152 @@ use crate::{
...
-        match self {
-            Self::Transport(_) => 0,
-            Self::Application(_) => 1,
+impl From<FrameType> for u8 {
+    fn from(val: FrameType) -> Self {
+        val as Self
+    }
+}
+
+impl TryFrom<u64> for FrameType {
```
martinthomson, line 73: It seems to me like this is ripe for a derive-like macro.  I wonder if there are any of those floating around?

neqo-transport/src/frame.rs
```diff
@@ -18,68 +18,152 @@ use crate::{
...
+}
+
+impl TryFrom<FrameType> for StreamType {
+    type Error = Error;
+
+    fn try_from(value: FrameType) -> Result<Self, Self::Error> {
+        match value {
+            FrameType::MaxStreamsBiDi | FrameType::StreamsBlockedBiDi => Ok(Self::BiDi),
+            FrameType::MaxStreamsUniDi | FrameType::StreamsBlockedUniDi => Ok(Self::UniDi),
+            _ => Err(Error::FrameEncodingError),
```
martinthomson, line 155: Is this the right error?  I'm not sure about usage though: what cases lead to this error?

neqo-transport/src/frame.rs
```diff
@@ -18,68 +18,152 @@ use crate::{
...
+        )
+    }
+    const fn is_stream_with_fin(self) -> bool {
+        matches!(
+            self,
+            Self::StreamWithFin
+                | Self::StreamWithLenFin
+                | Self::StreamWithOffFin
+                | Self::StreamWithOffLenFin
+        )
```
martinthomson, line 144: I wonder whether this produces the same code.  (i.e., does the compiler know how to apply `self as u64 & mask == pattern`)

It would be a shame if the cleaner code here results in worse performance.  Do you think that it is worth checking?

neqo-transport/src/qlog.rs
```diff
@@ -543,7 +543,7 @@ impl From<Frame<'_>> for QuicFrame {
             Frame::HandshakeDone => Self::HandshakeDone,
             Frame::AckFrequency { .. } => Self::Unknown {
                 frame_type_value: None,
-                raw_frame_type: frame.get_type(),
+                raw_frame_type: frame.get_type().into(),
                 raw: None,
             },
```
martinthomson, line 548: Interesting.  I guess we can look at whether qlog supports the frame when we get around to updating that spec.

martinthomson (commented):
Seems pretty good to me.

neqo-transport/src/frame.rs
larseggert: Oh yeah, there are a ton of useful things in `strum`.

neqo-transport/src/frame.rs
larseggert: We could also use `Error::DecodingFrame`? Not quite sure about the difference in meaning. 

We shouldn't hit this error unless there is a pilot error, i.e., calling trying to convert a `FrameType` into a `StreamType` for which that doesn't make sense (everything other than `MaxStreams` and `StreamsBlocked`).

larseggert: `Error::DecodingFrame` is only used in one place. I would suggest just changing that over to `Error::FrameEncodingError`.

neqo-transport/src/frame.rs
larseggert: It doesn't seem to be much worse at least: https://rust.godbolt.org/z/z3o1aMcnv

martinthomson: Yep, I tried a few variations and they all seem good.  But the code is neater this way.  Wins all around.

neqo-transport/src/frame.rs
martinthomson: Yeah, roll them into one.

mxinden (approved):

PR #2443, opened 2025-02-13
chore: Use `EnumSet` instead of a Boolean `EnumMap`

larseggert:
For `PacketNumberSpaceSet`. May be faster?

### DIFF
--- a/neqo-transport/src/tracking.rs
+++ b/neqo-transport/src/tracking.rs
@@ -16,7 +16,7 @@ use enum_map::{enum_map, Enum, EnumMap};
 use enumset::{EnumSet, EnumSetType};
 use neqo_common::{qdebug, qinfo, qtrace, qwarn, IpTosEcn};
 use neqo_crypto::Epoch;
-use strum::{EnumIter, IntoEnumIterator as _};
+use strum::EnumIter;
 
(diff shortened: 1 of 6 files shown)


neqo-transport/src/tracking.rs
```diff
@@ -76,63 +76,7 @@ impl From<PacketType> for PacketNumberSpace {
...
-                    f.write_str("+")?;
-                    first = false;
-                }
-                std::fmt::Display::fmt(sp, f)?;
-            }
-        }
-        f.write_str(")")
-    }
-}
+pub type PacketNumberSpaceSet = EnumSet<PacketNumberSpace>;
```
mxinden, line 79: Even if there is no performance win, this seems like a nice clean-up.

neqo-transport/src/tracking.rs
```diff
@@ -1127,43 +1072,37 @@ mod tests {
     #[test]
     fn pnspaceset_default() {
```
martinthomson, line 1073: We can probably just drop these tests.  We're testing other people's code.

martinthomson (approved):

mxinden (approved):

PR #2453, opened 2025-02-24
chore: Fix clippy warnings

larseggert:
Ans suppress some false positives, see https://github.com/rust-lang/rust-clippy/issues/14091

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -125,13 +125,13 @@ The API consists of:
   - `webtransport_session_accept` -  only used by the server-side implementation
   - `webtransport_close_session`
   - `webtransport_create_stream_local` -  this function is called when an application wants to open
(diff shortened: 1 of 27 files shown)


neqo-http3/src/connection.rs
```diff
@@ -171,15 +171,15 @@ There are the following types of streams:
 - `Decoder`: there is only a receiver stream of this type and the handler is `DecoderRecvStream`.
 - `Encoder`: there is only a receiver stream of this type and the handler is `EncoderRecvStream`.
 - `NewStream`: there is only a receiver stream of this type and the handler is
-               `NewStreamHeadReader`.
+  `NewStreamHeadReader`.
```
martinthomson, line 174: Wow, clippy is getting much more opinionated about stuff...

martinthomson (approved):
All seem good.  I like the removal of `#[must_use]` on trait implementations.  That's a good change.

PR #2454, opened 2025-02-25
ci: Define components for codecov

larseggert:
Might make navigating the coverage info easier? Let's see.


martinthomson (commented):
Seems reasonable.

mxinden (approved):

PR #2455, opened 2025-02-25
ci: Use CPUs 2 and 3 for benchmarking

larseggert:
They have been isolated per https://manuel.bernhardt.io/posts/2023-11-16-core-pinning/.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -111,10 +111,10 @@ jobs:
 
       - name: Run cargo bench
         run: |
+          sudo ip link set dev lo mtu "$MTU"
           taskset -c 2 nice -n -20 setarch --addr-no-randomize \
             cargo "+$TOOLCHAIN" bench --workspace --exclude neqo-bin --features bench -- --noplot | tee results.txt
-          sudo ip link set dev lo mtu "$MTU"
-          nice -n -20 setarch --addr-no-randomize \
+          taskset -c 2 nice -n -20 setarch --addr-no-randomize \
             cargo "+$TOOLCHAIN" bench --package neqo-bin --features bench -- --noplot | tee -a results.txt
 
       # Compare various configurations of neqo against msquic and google/quiche, and gather perf data
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -104,16 +104,14 @@ jobs:
           mkdir -p hyperfine
 
       # Disable turboboost, hyperthreading and use performance governor.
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
How many cores do we have?

mxinden (approved):

mxinden:
> 
>        time:   [2.1738 s 2.1812 s 2.1888 s]
>        thrpt:  [45.688 MiB/s 45.846 MiB/s 46.003 MiB/s]
> change:
>        time:   [+110.86% +113.68% +116.54%] (p = 0.00 < 0.05)
>        thrpt:  [-53.819% -53.201% -52.576%]

Since this pull request, our Download benchmark runs ~ 2x slower. See for example https://github.com/mozilla/neqo/pull/2481 and https://github.com/mozilla/neqo/pull/2488 both now below 50 MiB/s.

@larseggert is that deliberate? E.g. do we trade max throughput in favor of consistency?

For comparison, here is a run on my local machine, with PMTUD disabled:

```
➜  neqo git:(optimistic-send) ✗ cargo bench --features bench --bench main -- "Download"

                        time:   [312.42 ms 340.84 ms 368.79 ms]
                        thrpt:  [271.16 MiB/s 293.39 MiB/s 320.08 MiB/s]
                 change:
                        time:   [-14.955% -4.0479% +8.5757%] (p = 0.48 > 0.05)
                        thrpt:  [-7.8984% +4.2186% +17.585%]

larseggert:
I think the issue is that because the CPUs are isolated now, they aren't being controlled by the OS frequency governor and hence run at minimum frequency. I need to figure out how to run them at something higher.

larseggert:
@mxinden also, we're using the `tokio` multi-threaded runtime but force everything onto a single (isolated) core. I don't know how much inefficiencies that entails?

mxinden:
> @mxinden also, we're using the `tokio` multi-threaded runtime but force everything onto a single (isolated) core. I don't know how much inefficiencies that entails?

As in, now that the CPUs are pinned, we should use a single threaded runtime and thus get back to our initial throughput (i.e. 2x)?

larseggert:
> As in, now that the CPUs are pinned, we should use a single threaded runtime and thus get back to our initial throughput (i.e. 2x)?

Well, we pinned the process earlier as well (to a not-completely-isolated CPU) so there _should_ not be a difference, but who knows.

mxinden:
@larseggert how should we proceed here? I suggest reverting this pull request and the related changes on the benchmark runner, such that we get back to our previous ~100 Mbit/s. Thoughts?

PR #2456, opened 2025-02-26
chore: Bump MSRV to 1.82

larseggert:
And address some TODOs that were enabled by the new MSRV.


martinthomson:
What is going on with those benchmarks?  I'm seeing huge swings.  Have you landed that CPU pinning change?  I would have thought that we would be less exposed to noise with that.

martinthomson:
The failure here looks legit.  We're doing a sign erasure cast.  The problem I see is that we already suppress the lint, so I don't know what is going on there.

larseggert:
> What is going on with those benchmarks? I'm seeing huge swings. Have you landed that CPU pinning change? I would have thought that we would be less exposed to noise with that.

So did I. I need to do some more debugging to understand and fix why it is (still) doing that. One weird aspect is that the isolated cores report a clock speed of 800MHz (vs. 3Ghz). That might be due to the way the isolation was done (via `cmdline`); if so there is a newer `cpuset`-based approach that I can try. Or the report is just wrong, since the benchmark speed difference is not _that_ large...

larseggert:
> The failure here looks legit. We're doing a sign erasure cast. The problem I see is that we already suppress the lint, so I don't know what is going on there.

We didn't suppress `clippy::cast_sign_loss`, which only fires on Windows. Should be fixed now.

neqo-transport/src/crypto.rs
```diff
@@ -1253,14 +1250,14 @@ impl CryptoStates {
             cipher: TLS_AES_128_GCM_SHA256,
             next_secret: hkdf::import_key(TLS_VERSION_1_3, &[0xaa; 32]).unwrap(),
         };
-        let initials = enum_map! {
-            Version::Version1 => Some(CryptoState {
-                    tx: CryptoDxState::test_default(),
-                    rx: read(0),
+        let initials = EnumMap::from_array([
```
martinthomson, line 1253: Doesn't this make the code more fragile?  If we change the order in the enum (we won't, but still) then this code will still compile, but it will be wrong.

larseggert: The old code made `clippy::mem_forget` fire, and that seemed like a memory leak possibility. Maybe there is another way to avoid that?

martinthomson: I went to the lengths of making the change to the enum_map code, but then I tried to test it and I don't see the error.  Isn't clippy supposed to avoid warnings like this?  I'm going to suggest it to them anyway, but when I can't reproduce it, it seems almost pointless.

larseggert: You need to turn on this warning in Cargo.toml.

martinthomson: Of course 🤦.  The maintainer was receptive, so maybe we'll be able to get a revision that works for us.

larseggert: `enum-map` seems pretty dead a as a project, unfortunately: https://codeberg.org/xfix/enum-map/issues/147 I doubt we'll see another release...

PR #2457, opened 2025-02-26
ci: Use all cores to build on bencher

larseggert:
Also the isolated ones.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -72,22 +72,22 @@ jobs:
 
       - name: Build neqo
         run: |
-          cargo "+$TOOLCHAIN" bench --workspace --features bench --no-run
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
What are you overriding here?  The default for any of these tasks is running on all cores.  So what changed?  Is this because the machine is configured specially?  Why does that configuration not limit itself to services and not the workflow execution (which should be free to use all cores by default).

larseggert:
The bencher is booting with `nohz_full=2-3,6-7 isolcpus=domain,managed_irq,2-3,6-7 irqaffinity=0-1,4-5` and so won't schedule anything on CPUs 2 and 3 (and their hyperthreading cores 6 and 7). You need to use `taskset` to get anything scheduled on the isolated CPUs.

But I checked and this doesn't work. It seems like a range that includes isolated cores is still not scheduling anything on them.

PR #2460, opened 2025-02-28
fix: Convert large pass-by-value args to references

larseggert:
Let's see if this matters.


larseggert:
This seems to not help.

PR #2470, opened 2025-03-03
fix: Use 1MB socket RX buffers

### LINKED ISSUE #1962: Consider increasing OS socket buffer (`SO_SNDBUF`, `SO_RCVBUF`)
For increased performance, we might want to increase the OS UDP socket send and receive buffer.

See also:

- https://man7.org/linux/man-pages/man7/socket.7.html
    - `SO_SNDBUF`
    - `SO_RCVBUF`
- https://github.com/google/quiche/blob/95857ee25d30c1dc09f1c36cbe7a29b969a11f6d/quiche/quic/core/quic_udp_socket_posix.inc#L195
- https://github.com/quic-go/quic-go/blob/8f1110450bad0b41f8dd03ea8cf24c299a9c9a20/sys_conn_helper_linux.go#L35
- https://github.com/quinn-rs/quinn/blob/693c9b7cfbf89c541ba99523237594499984ffed/README.md#usage-notes

larseggert:
We need to do the same for `neqo-glue`.

Fixes #1962

### DIFF
--- a/neqo-bin/src/udp.rs
+++ b/neqo-bin/src/udp.rs
@@ -30,26 +30,28 @@ impl Socket {
         let socket = std::net::UdpSocket::bind(addr)?;
         let state = quinn_udp::UdpSocketState::new((&socket).into())?;
 
-        let send_buf = state.send_buffer_size((&socket).into())?;
+        let send_buf_before = state.send_buffer_size((&socket).into())?;
         // FIXME: We need to experiment if increasing this actually improves performance.
         // Also, on BSD and Apple targets, this seems to increase the `net.inet.udp.maxdgram`
         // sysctl, which is not the same as the socket buffer.
-        // if send_buf < ONE_MB {
-        //     qdebug!("Increasing socket send buffer size from {send_buf} to {ONE_MB}");
+        // if send_buf_before < ONE_MB {
         //     state.set_send_buffer_size((&socket).into(), ONE_MB)?;
+        //     let send_buf_after = state.send_buffer_size((&socket).into())?;
+        //     qdebug!("Increasing socket send buffer size from {send_buf_before} to {ONE_MB}, now: {send_buf_after}");
         // } else {
-        //     qdebug!("Default socket send buffer size is {send_buf}, not changing");
+        //     qdebug!("Default socket send buffer size is {send_buf_before}, not changing");
         // }
-        qdebug!("Default socket send buffer size is {send_buf}");
+        qdebug!("Default socket send buffer size is {send_buf_before}");
 
-        let recv_buf = state.recv_buffer_size((&socket).into())?;
-        if recv_buf < ONE_MB {
-            qdebug!("Increasing socket receive buffer size from {recv_buf} to {ONE_MB}");
+        let recv_buf_before = state.recv_buffer_size((&socket).into())?;
+        if recv_buf_before < ONE_MB {
             // Same as Firefox.
             // <https://searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#13474-13478>
             state.set_recv_buffer_size((&socket).into(), ONE_MB)?;
+            let recv_buf_after = state.recv_buffer_size((&socket).into())?;
+            qdebug!("Increasing socket recv buffer size from {recv_buf_before} to {ONE_MB}, now: {recv_buf_after}");
         } else {
-            qdebug!("Default socket receive buffer size is {recv_buf}, not changing");
+            qdebug!("Default socket receive buffer size is {recv_buf_before}, not changing");
         }
 
         Ok(Self {
--- a/neqo-bin/src/udp.rs
+++ b/neqo-bin/src/udp.rs
@@ -31,19 +31,23 @@ impl Socket {
         let state = quinn_udp::UdpSocketState::new((&socket).into())?;
 
         let send_buf = state.send_buffer_size((&socket).into())?;
-        if send_buf < ONE_MB {
-            qdebug!("Increasing send buffer size from {send_buf} to {ONE_MB}");
-            state.set_send_buffer_size((&socket).into(), ONE_MB)?;
-        } else {
-            qdebug!("Default send buffer size is {send_buf}, not changing");
-        }
+        // FIXME: We need to experiment if increasing this actually improves performance.
+        // Also, on BSD and Apple targets, this seems to increase the `net.inet.udp.maxdgram`
+        // sysctl, which is not the same as the socket buffer.
+        // if send_buf < ONE_MB {
+        //     qdebug!("Increasing socket send buffer size from {send_buf} to {ONE_MB}");
+        //     state.set_send_buffer_size((&socket).into(), ONE_MB)?;
+        // } else {
+        //     qdebug!("Default socket send buffer size is {send_buf}, not changing");
+        // }
+        qdebug!("Default socket send buffer size is {send_buf}");
 
         let recv_buf = state.recv_buffer_size((&socket).into())?;
         if recv_buf < ONE_MB {
-            qdebug!("Increasing receive buffer size from {recv_buf} to {ONE_MB}");
+            qdebug!("Increasing socket receive buffer size from {recv_buf} to {ONE_MB}");
             state.set_recv_buffer_size((&socket).into(), ONE_MB)?;
         } else {
-            qdebug!("Default receive buffer size is {recv_buf}, not changing");
+            qdebug!("Default socket receive buffer size is {recv_buf}, not changing");
         }
 
         Ok(Self {
--- a/neqo-bin/src/udp.rs
+++ b/neqo-bin/src/udp.rs
@@ -8,7 +8,7 @@
 
 use std::{io, net::SocketAddr};
 
-use neqo_common::{qdebug, Datagram};
+use neqo_common::{qwarn, Datagram};
 use neqo_udp::{DatagramIter, RecvBuf};
 
 /// Ideally this would live in [`neqo-udp`]. [`neqo-udp`] is used in Firefox.
@@ -32,18 +32,18 @@ impl Socket {
 
         let send_buf = state.send_buffer_size((&socket).into())?;
         if send_buf < ONE_MB {
-            qdebug!("Increasing send buffer size from {send_buf} to {ONE_MB}");
+            qwarn!("Increasing send buffer size from {send_buf} to {ONE_MB}");
             state.set_send_buffer_size((&socket).into(), ONE_MB)?;
         } else {
-            qdebug!("Default send buffer size is {send_buf}, not changing");
+            qwarn!("Default send buffer size is {send_buf}, not changing");
         }
 
         let recv_buf = state.recv_buffer_size((&socket).into())?;
         if recv_buf < ONE_MB {
-            qdebug!("Increasing receive buffer size from {recv_buf} to {ONE_MB}");
+            qwarn!("Increasing receive buffer size from {recv_buf} to {ONE_MB}");
             state.set_recv_buffer_size((&socket).into(), ONE_MB)?;
         } else {
-            qdebug!("Default receive buffer size is {recv_buf}, not changing");
+            qwarn!("Default receive buffer size is {recv_buf}, not changing");
         }
 
         Ok(Self {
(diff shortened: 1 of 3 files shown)


neqo-bin/src/udp.rs
```diff
@@ -28,6 +28,13 @@ impl Socket {
     pub fn bind<A: std::net::ToSocketAddrs>(addr: A) -> Result<Self, io::Error> {
         let socket = std::net::UdpSocket::bind(addr)?;
 
+        // Try to incresase the send and receive buffer sizes.
+        let socket = socket2::Socket::from(socket);
+        let one_mb: usize = 1 << 20;
+        // socket.set_send_buffer_size(one_mb)?;
+        socket.set_recv_buffer_size(one_mb)?;
```
mxinden, line 35: For the record, we use a 1MiB buffer in Firefox as well:

https://searchfox.org/mozilla-central/rev/9c395b6371eaea0d15f9c8a37889022be350cf0b/modules/libpref/init/StaticPrefList.yaml#14560-14564

https://searchfox.org/mozilla-central/rev/01aaa47e62a2015e7641f26ab0bc2bb00ab579b8/netwerk/protocol/http/HttpConnectionUDP.cpp#107-115

neqo-bin/src/udp.rs
```diff
@@ -28,6 +28,13 @@ impl Socket {
     pub fn bind<A: std::net::ToSocketAddrs>(addr: A) -> Result<Self, io::Error> {
         let socket = std::net::UdpSocket::bind(addr)?;
 
+        // Try to incresase the send and receive buffer sizes.
+        let socket = socket2::Socket::from(socket);
+        let one_mb: usize = 1 << 20;
+        // socket.set_send_buffer_size(one_mb)?;
```
mxinden, line 34: Not yet ready to merge.

mxinden (commented):
Instead of implementing this in `neqo-bin`, how about proposing it upstream `quinn-udp`:

https://github.com/quinn-rs/quinn/blob/8d6e48c20b71f7e915c13c33b66e2a09b6d59888/quinn-udp/src/unix.rs#L85-L86

I assume close to all QUIC implementations will want a large send and receive buffer.

larseggert:
> Instead of implementing this in `neqo-bin`, how about proposing it upstream `quinn-udp`:

Sounds good. Or maybe `quinn-udp` can offer functions to set those values.

I don't understand why the benches show a performance improvement, but the runs over loopback show a regression.

larseggert:
@mxinden where would you suggest to add this in `quinn-udp`? As part of `UdpSocketState` or somewhere else?

larseggert:
@mxinden alternatively, I'd suggest merging this to neqo now and switching over to any `quinn-udp` mechanism in a future PR.

mxinden:
@larseggert how about something along the lines of:

``` diff
diff --git a/quinn-udp/src/unix.rs b/quinn-udp/src/unix.rs
index c39941d5..0b5bbd21 100644
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -257,6 +257,14 @@ impl UdpSocketState {
         self.may_fragment
     }
 
+    pub fn set_send_buffer_size(&self, bytes: usize)-> io::Result<()>  {
+        todo!();
+    }
+
+    pub fn set_rec_buffer_size(&self, bytes: usize) -> io::Result<()> {
+        todo!();
+    }
+
     /// Returns true if we previously got an EINVAL error from `sendmsg` syscall.
     fn sendmsg_einval(&self) -> bool {
         self.sendmsg_einval.load(Ordering::Relaxed)
@@ -543,7 +551,7 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
     Ok(1)
 }
```

larseggert:
> @larseggert how about something along the lines of

Do we also need the equivalent for Windows?

mxinden:
Yes. I would say the more platforms the better.

larseggert:
@mxinden let me know if https://github.com/quinn-rs/quinn/pull/2179 is what you had in mind.

larseggert:
On my Mac, I now see:
```
0.003 DEBUG Increasing send buffer size from 9216 to 1048576
0.003 DEBUG Increasing receive buffer size from 786896 to 1048576
```
Note that the send buffer size is apparently based on the `net.inet.udp.maxdgram` `sysctl`? Weird. Need to check this on other platforms.

larseggert:
Linux (from a QNS run):
```
0.002 DEBUG Increasing send buffer size from 212992 to 1048576
0.002 DEBUG Default receive buffer size is 1048576, not changing
```

larseggert:
I also don't understand why loopback benchmark performance is going _down_.

mxinden:
I assume more buffer space is not always better. This reminds me of the _Source Buffer Management_ draft.

https://stuartcheshire.github.io/draft-cheshire-sbm/draft-cheshire-sbm.html

mxinden:
> @mxinden do we see `ENOBUFS` and are we handling that similar to `EAGAIN`?

We currently don't. Something along the following lines should work. Want to try running a benchmark on your local Mac to see whether either of them triggers?

``` diff
diff --git a/Cargo.lock b/Cargo.lock
index 64bfcc1f..d6b8b6a7 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1028,6 +1028,7 @@ name = "neqo-udp"
 version = "0.12.2"
 dependencies = [
  "cfg_aliases",
+ "libc",
  "log",
  "neqo-common",
  "quinn-udp",
diff --git a/neqo-udp/Cargo.toml b/neqo-udp/Cargo.toml
index 4abd402a..8ebc2e83 100644
--- a/neqo-udp/Cargo.toml
+++ b/neqo-udp/Cargo.toml
@@ -19,6 +19,7 @@ workspace = true
 log = { workspace = true }
 neqo-common = { path = "./../neqo-common" }
 quinn-udp = { workspace = true }
+libc = "0.2"
 
 [build-dependencies]
 cfg_aliases = "0.2"
diff --git a/neqo-udp/src/lib.rs b/neqo-udp/src/lib.rs
index d498f5aa..0cfe0e74 100644
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -74,7 +74,19 @@ pub fn send_inner(
         src_ip: None,
     };
 
-    state.try_send(socket, &transmit)?;
+    if let Err(e) = state.try_send(socket, &transmit) {
+        match e.raw_os_error() {
+            Some(libc::ENOBUFS) => {
+                todo!();
+            }
+            Some(libc::EAGAIN) => {
+                todo!();
+            }
+            Some(_) | None => {}
+        }
+
+        return Err(e);
+    }
 
     qtrace!(
         "sent {} bytes from {} to {}",
```

Related: Neither of them are currently exposed through `std::io::ErrorKind`. See https://github.com/rust-lang/rust/pull/79965 for some discussion on adding these to `std`.

larseggert:
> I assume more buffer space is not always better. This reminds me of the _Source Buffer Management_ draft.
> 
> [stuartcheshire.github.io/draft-cheshire-sbm/draft-cheshire-sbm.html](https://stuartcheshire.github.io/draft-cheshire-sbm/draft-cheshire-sbm.html)

It's different, in that I doubt that we actually fill that larger buffer, and they do. (If we filled it, we should IMO also see a throughput increase.)

larseggert:
OK, with the fixed bencher the erratic performance seems to be fixed 🎉 

I guess we'd want this one merged to see the full impact of  #1868?

neqo-bin/src/udp.rs
```diff
@@ -26,10 +26,28 @@ pub struct Socket {
...
     /// Create a new [`Socket`] bound to the provided address, not managed externally.
     pub fn bind<A: std::net::ToSocketAddrs>(addr: A) -> Result<Self, io::Error> {
+        const ONE_MB: usize = 1 << 20;
         let socket = std::net::UdpSocket::bind(addr)?;
+        let state = quinn_udp::UdpSocketState::new((&socket).into())?;
+
+        let send_buf = state.send_buffer_size((&socket).into())?;
+        if send_buf < ONE_MB {
+            qdebug!("Increasing send buffer size from {send_buf} to {ONE_MB}");
+            state.set_send_buffer_size((&socket).into(), ONE_MB)?;
```
larseggert, line 36: If this fails, should we retry with a smaller value? (Also for `recv_buf`.)

neqo-bin/src/udp.rs
```diff
@@ -26,10 +26,28 @@ pub struct Socket {
...
+            qdebug!("Default send buffer size is {send_buf}, not changing");
+        }
+
+        let recv_buf = state.recv_buffer_size((&socket).into())?;
+        if recv_buf < ONE_MB {
+            qdebug!("Increasing receive buffer size from {recv_buf} to {ONE_MB}");
+            state.set_recv_buffer_size((&socket).into(), ONE_MB)?;
+        } else {
+            qdebug!("Default receive buffer size is {recv_buf}, not changing");
+        }
```
larseggert, line 48: Should we add telemetry/stats about which sizes were actually in use for a connection?

mxinden: :+1: we can do that in `neqo_glue`.

larseggert: I'd like to keep the glue as minimal as possible. So adding this to `Stats` seems preferable.

mxinden:
**On send buffer increase:**

> > I assume more buffer space is not always better. This reminds me of the _Source Buffer Management_ draft.
> > [stuartcheshire.github.io/draft-cheshire-sbm/draft-cheshire-sbm.html](https://stuartcheshire.github.io/draft-cheshire-sbm/draft-cheshire-sbm.html)
> 
> It's different, in that I doubt that we actually fill that larger buffer, and they do. (If we filled it, we should IMO also see a throughput increase.)

Can you expand on why the above issue of source buffer bloat is not an issue here? In addition, if we never fill the larger buffer, why increase it in the first place?

Given that Neqo paces the sending of packets, and given that this pacing is on the order of milliseconds, assuming that each pacing batch fits into the send buffer, shouldn't that resolve the need for a large OS UDP send buffer size, as each pacing gab gives the OS enough time to empty the buffer? 

Also on the matter of pacing, wouldn't filling a large send buffer undo our user space pacing effort?

Now that we do [`PR_POLL_WRITE` in Firefox](https://phabricator.services.mozilla.com/D239162), I am less worried about a small send buffer.

**On receive buffer increase:**

Increasing the receive buffer size in `neqo-client` and `neqo-server` sounds good to me.

Note that we already increase the receive buffer size in Firefox. Thus there are no changes required in `neqo_glue`. We could however now do the increase through `quinn-udp` instead of NSPR.

neqo-bin/src/udp.rs
mxinden: > I'd like to keep the glue as minimal as possible.

Agreed.

Note however that this would be the first time we set something in `neqo_transport::Stats` from outside of `neqo_transport`.

larseggert:
If we want to experiment more, should we drop the increase to the send buffer for now?

(Also, I think `SO_SNDBUF` on BSD/Apple is not the same as on Linux, because on these platforms it actually seems to control `net.inet.udp.maxdgram`.)

mxinden:
> If we want to experiment more, should we drop the increase to the send buffer for now?

Sounds good to me.

neqo-bin/src/udp.rs
```diff
@@ -26,10 +26,36 @@ pub struct Socket {
...
+        //     qdebug!("Increasing socket send buffer size from {send_buf_before} to {ONE_MB}, now:
+        // {send_buf_after}"); } else {
+        //     qdebug!("Default socket send buffer size is {send_buf_before}, not changing");
+        // }
+        qdebug!("Default socket send buffer size is {send_buf_before}");
+
+        let recv_buf_before = state.recv_buffer_size((&socket).into())?;
+        if recv_buf_before < ONE_MB {
+            // Same as Firefox.
+            // <https://searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#13474-13478>
```
martinthomson, line 49: When citing searchfox, don't use the "latest" links, use the versioned links.

https://searchfox.org/mozilla-central/rev/fa5b44a4ea5c98b6a15f39638ea4cd04dc271f3d/modules/libpref/init/StaticPrefList.yaml#13474-13477 is what you are looking for here.

PR #2471, opened 2025-03-03
chore: `cargo clippy --fix`

larseggert:


mxinden (approved):
Oh `Error::other` is neat!

PR #2474, opened 2025-03-05
ci: Don't use `mold` when building for Android

larseggert:


larseggert:
> What is the rational?

The Android CI cross-compilation is otherwise failing on Linux.

PR #2477, opened 2025-03-06
fix: Add some crate features for performance

larseggert:
Let's see if they do.

Also, @mxinden, I was wondering why we went with a multi-threaded `tokio` client and server. I'm wondering if the thread-management overheads are worth it compared to using just the `rt` scheduler?


mxinden:
> Also, @mxinden, I was wondering why we went with a multi-threaded `tokio` client and server. 

I chose multi-threaded as it is the de-facto default. No other reason.

> I'm wondering if the thread-management overheads are worth it compared to using just the `rt` scheduler?

:+1: worth experimenting. Intuitively, given that it is a single future only, there is no cross-thread communication and thus no significant overhead.

mxinden:
I am fine merging here. That said, I would prefer individual pull requests per feature, to ensure each change, and not just all changes as a whole, have a positive performance impact. In addition, I don't think we should merge here before we have reliable benchmarks, i.e. not merge here before https://github.com/mozilla/neqo/issues/2657 is fixed.

larseggert:
I think there is some weak signal that there is some performance benefit here.

Cargo.toml
```diff
@@ -37,7 +37,7 @@ hex = { version = "0.4", default-features = false }
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.15.1", default-features = false }
 quinn-udp = { version = "0.5.13", default-features = false, features = ["direct-log"] }
-regex = { version = "1.9", default-features = false }
+regex = { version = "1.9", default-features = false, features = ["std", "perf", "unicode-perl"] }
```
martinthomson: What do we use regex for anyway?  I see it in qlog tests and http09.  Seems like we can take the extra dependencies purely on the promise that these are faster.

Not sure about "unicode-perl".  That seems to only increase binary size if we aren't using those patterns.  I see that http09 has two patterns that mean we might need "unicode-bool" instead.

```suggestion
regex = { version = "1.9", default-features = false, features = ["std", "perf", "unicode-bool"] }
```

larseggert:
I don't see enough performance benefits here.

PR #2478, opened 2025-03-06
ci: Switch from `perf` to `samply` for profiling

larseggert:


larseggert:
`samply` doesn't capture the kernel level.

Issue #2480, opened 2025-03-09
ci: pin OS versions


mxinden:
We mostly use `<SOME_OS>-latest` on CI, for example:

https://github.com/mozilla/neqo/blob/318b611512cb40235fc251a4870b43ecfd4a71fa/.github/workflows/check.yml#L51 

I suggest we pin the OS version, e.g. changing the above to:

``` diff
-         os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
+         os: [ubuntu-24.04, ubuntu-24.04-arm, macos-14, windows-2025]
```

Benefit:
- **When cutting patch releases of old `neqo-*` releases, CI is more likely to still work, given it using the same OS version as previous (patch-) releases of same release family.**
- No CI breakage due to automatic OS version updates (e.g. `macos-14` -> `macos-15`) across e.g. `main` CI runs.
- Supply-chain security

Objections? If not, I will prepare a pull request.

PR #2481, opened 2025-03-09
ci: pin GitHub-hosted runners OS versions

mxinden:
Fixes https://github.com/mozilla/neqo/issues/2480.

See [GitHub docs for latest runner versions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for--private-repositories).

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   actionlint:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


larseggert:
Do we need the same for `mtu`?

mxinden:
Yes, good point. I created https://github.com/mozilla/mtu/pull/105.

larseggert (approved):

PR #2488, opened 2025-03-10
perf(bin): optimistically send datagram, only await on failure

mxinden:
Previously `neqo-bin` would poll a UDP socket to be writable before attempting to send a datagram. This results in 2 syscalls per send.

Instead, optimistically attempt to send the datagram first, on failure `WouldBlock`, await and then try again. Under the assumption that the majority of sends work on the first try, this reduces the number of syscalls.

Follow-up to https://phabricator.services.mozilla.com/D239162#8315719.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -464,10 +464,19 @@ impl<'a, H: Handler> Runner<'a, H> {
     async fn process_output(&mut self) -> Result<(), io::Error> {
         loop {
             match self.client.process_output(Instant::now()) {
-                Output::Datagram(dgram) => {
-                    self.socket.writable().await?;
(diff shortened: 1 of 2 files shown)


mxinden:
Blocking here until https://github.com/mozilla/neqo/pull/2455#issuecomment-2711443972 is fixed. @larseggert let me know if I can help in any way.

larseggert:
@mxinden I did make a change a few days ago, let's see if the benchmark numbers change.

mxinden:
Thanks. Merging `main` tro trigger another benchmark run.

larseggert:
No change on the bencher it seems. Let's merge this now, since the underlying issue is not related to this PR.

larseggert (approved):

PR #2491, opened 2025-03-11
docs(README): mention `disable_when_third_party_roots_found` pref

mxinden:
In order to connect to a neqo-server with a self-signed certificate, one needs to set `network.http.http3.disable_when_third_party_roots_found` to `false`.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1925014 for details.

//CC @justus237

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,9 @@ something has changed.
 ### Connect with Firefox to local neqo-server
 
 1. Run `neqo-server` via `cargo run --bin neqo-server -- 'localhost:12345' --db ./test-fixture/db`.
-2. On Firefox, set `about:config` preference `network.http.http3.alt-svc-mapping-for-testing` to `localhost;h3=":12345"`.
+2. On Firefox, set `about:config` preferences:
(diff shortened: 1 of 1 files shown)


larseggert (approved):

mxinden:
This is a `.md` only change. We don't run CI on `.md` only changes:

https://github.com/mozilla/neqo/blob/99144585db35fb7dd7e60a78019df5d184ecad51/.github/workflows/check.yml#L6-L8

Yet we have various `check.yml` jobs marked as required. Thus I can not merge here.

@larseggert mind forcing a merge?

larseggert:
Should we run CI on all changes?

mxinden:
Given that we rarely have docs-only pull requests, I don't think the `paths-ignore` optimization saves us much CI-runner-time. Thus, yes, I am in favor of running CI on all changes.

martinthomson:
Given that this is fairly rare, I'd prefer to avoid the waste and just override the merge rules.

Issue #2476, opened 2025-03-05
Lost Packet Observed After 0RTT Connection Establishment


KershawChang:
I've been investigating why the upload goodput of Firefox sometimes drops and found that this drop usually happens on 0RTT connections. The behavior can be demonstrated by the test below.
```
#[test]
fn zero_rtt_negotiate() {
    // Note that the two servers in this test will get different anti-replay filters.
    // That's OK because we aren't testing anti-replay.
    let mut client = default_client();
    let mut server = default_server();
    connect(&mut client, &mut server);

    let token = exchange_ticket(&mut client, &mut server, now());
    let mut client = default_client();
    client
        .enable_resumption(now(), token)
        .expect("should set token");
    let mut server = resumed_server(&client);

    connect_with_rtt(&mut client, &mut server, now(), Duration::from_millis(100));
    assert!(client.tls_info().unwrap().early_data_accepted());
    assert!(server.tls_info().unwrap().early_data_accepted());

    // Why do we have a lost packet?
    eprintln!("lost:{:?}", client.stats().lost);
}
```

It appears that the `ZeroRtt` packet is always marked as lost. For example, the log shows:
```
0.016 DEBUG [Client fed0439569b46ed1] pn=0 type=ZeroRtt pri-path:08000000008ededd [fe80::1]:443->[fe80::1]:443 IpTos(Cs0, Ect0) len 1098
  TX -> NewConnectionId { sequence_number: 1, retire_prior: 0, connection_id: [8, 0, 0, 0, 1, 250, 20, 56], stateless_reset_token: [9, 35, 2, 93, 245, 164, 247, 61, 90, 42, 1, 169, 24, 48, 123, 186] }
0.016 TRACE [LossRecovery] packet ap-0 sent
0.016 TRACE [Pacer 1232/1232] spend 70 over 12320, 100ms
0.016 DEBUG packet_sent this=0x158809900, pn=0, ps=70

...
0.020 DEBUG packet_lost this=0x158809900, pn=0, ps=70
0.020 DEBUG [Cubic [last_max_cwnd: 12320, k: 0, w_max: 0, ca_epoch_start: None] CongCtrl 0/8624 ssthresh 8624] Cong event -> recovery; cwnd 8624, ssthresh 8624
0.020 DEBUG [Cubic [last_max_cwnd: 12320, k: 0, w_max: 0, ca_epoch_start: None] CongCtrl 0/8624 ssthresh 8624] state -> RecoveryStart
[...]

Issue #2490, opened 2025-03-11
Consider reducing MAX_PATH_PROBES with ECN


mxinden:
Currently we probe a path up to 6 times, i.e. 3 times with ECN, 3 times without ECN.

https://github.com/mozilla/neqo/blob/99144585db35fb7dd7e60a78019df5d184ecad51/neqo-transport/src/path.rs#L35-L39

On Firefox Nightly we see around ~2% of connection attempts seeing an ECN black-hole. In other words for 2% of connection attempts the first 3 path probes with ECN fail and one of the consecutive path probes without ECN succeed.

![Image](https://github.com/user-attachments/assets/09ad69c6-e53b-4b7a-b5fa-c8819958dc36)

https://yardstick.mozilla.org/d/aeak3dvriig3kd/http3?orgId=1&from=now-7d&to=now&timezone=browser&viewPanel=panel-3

If I understand correctly our initial PTO should be ~100ms. Thus ~2% of HTTP3 connections get delayed by 300ms. I would assume most of these ~2% thus loose the race to a concurrent HTTP2 connection attempt

Given the insights from Firefox Nightly, should we reduce the first `MAX_PATH_PROBES` with ECN from 3 to e.g. 1? Or do we consider ~2% not significant enough?

larseggert:
Should we do an experiment?

martinthomson:
What metric would you seek to optimize?  HTTP/3 usage?  ECN usage?  Connection establishment time?

mxinden:
> Should we do an experiment?

We can. That said, having done one myself now, I am not sure it is worth the bureaucratic overhead.

Given that Firefox Nightly is a small population only, and given the overhead of an experiment, how about we enable ECN on Firefox _Early Beta_ first. Rolling out to more devices gives us higher confidence in the percentage of ECN black holes seen by Firefox users. In case the percentage of black holes is still relatively high, we either (a) do an experiment or (b) make a change like the one suggested above.

> What metric would you seek to optimize? HTTP/3 usage? ECN usage? Connection establishment time?

Metrics worth monitoring:
- Number of HTTP responses by version via https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/http_response_version
- Number of blackholed connection attempts via https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/networking_http_3_ecn_path_capability
[...]

PR #2492, opened 2025-03-11
 fix(transport/recovery): don't use loss_delay before first ACK

mxinden:
RFC 9002 6.1.2 says:

> Once a later packet within the same packet number space has been acknowledged, an endpoint SHOULD declare an earlier packet lost if it was sent a threshold amount of time in the past.

https://www.rfc-editor.org/rfc/rfc9002.html#section-6.1.2

In other words, before the first ACK, one can not use the time threshold (i.e. here `loss_delay`) for loss detection, but instead has to depend on the Probe Timeout (i.e. here `cleanup_delay`).

This is e.g. relevant in the following scenario:

0. Assume a resumed 0-RTT connection with a small RTT (e.g. 4ms).
1. Say the client connects to the server and sends 0-RTT stream data along with the first flight.
2. The server will acknowledge Initial and Handshake packets, but won't acknowledge the 0-RTT packet, as it (a) doesn't yet have 1-RTT keys and (b) will delay ACKs by the default ACK delay (i.e. 25ms).
3. The client should not declare the 0-RTT stream data as lost via the time threshold, i.e. within ~1RTT, but instead wait for the PTO to fire, or an ACK to arrive. If the client prematurely declares the 0-RTT packet as lost, it will unnecessarily enter congestion recovery phase.

Fixes https://github.com/mozilla/neqo/issues/2476.

---

Is my above read of RFC 9002 correct?

Does this align with your suggestion in https://github.com/mozilla/neqo/issues/2476#issuecomment-2712145921 @martinthomson?

@KershawChang has validated that this would address the bug he is seeing in https://github.com/mozilla/neqo/issues/2476.

I will address the failures in other tests once we know this is a valid approach in the first place.

### DIFF
--- a/neqo-transport/src/connection/tests/zerortt.rs
+++ b/neqo-transport/src/connection/tests/zerortt.rs
@@ -12,7 +12,7 @@ use test_fixture::{assertions, now};
 
 use super::{
     super::Connection, connect, default_client, default_server, exchange_ticket, new_server,
-    resumed_server, CountingConnectionIdGenerator,
+    resumed_server, CountingConnectionIdGenerator, Output,
 };
 use crate::{
     events::ConnectionEvent, ConnectionParameters, Error, StreamType, Version,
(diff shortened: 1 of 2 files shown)


mxinden:
Looking at some old pull requests, this change has actually been proposed before by @larseggert as part of https://github.com/mozilla/neqo/pull/1998. That pull request was never merged as a whole, only subsets of it.

@larseggert any reason why the core of https://github.com/mozilla/neqo/pull/1998, i.e. _No time-based loss detection before `largest_acked`_, has never merged? In other words, do you see any reason why we shouldn't proceed with this pull request?

larseggert:
We should totally go ahead with this PR. I guess I never completed refactoring #1998 into its many components! Check if there is anything else that I forgot?

mxinden:
Status update: this pull request depends on https://github.com/mozilla/neqo/pull/2129/ which depends on https://github.com/mozilla/neqo/pull/2509.

larseggert:
> Status update: this pull request depends on #2129 which depends on #2509.

Those two are merged now, which should unblock this one.

mxinden:
I will need to finish https://github.com/mozilla/neqo/pull/2516 before we can proceed here.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -88,10 +88,13 @@ fn handshake_delay_with_ecn_blackhole() {
...
 
+    assert!(client.state().connected());
+    assert!(server.state().connected());
+
     assert_eq!(
         (finish - start).as_millis() / DEFAULT_RTT.as_millis(),
-        15,
-        "expected 6 RTT for client to detect blackhole, 6 RTT for server to detect blackhole and 3 RTT for handshake to be confirmed",
+        45,
+        "expected 3 RTT for first client PTO + 6 RTT for second PTO + 12 RTT for third PTO + another 21 RTT for the same on the server side + 3 RTT for handshake to be confirmed",
```
mxinden, line 97: See also discussion in https://github.com/mozilla/neqo/issues/2490#issuecomment-2759386437.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -255,43 +256,51 @@ fn pto_handshake_complete() {
...
+    let pto = pto * 2;
     let cb = client.process_output(now).callback();
-    assert_eq!(cb, HALF_RTT * 12);
+    assert_eq!(cb, pto);
 
-    // We still have only a single PTO
+    // Two PTOs in a row.
+    pto_counts[0] = 0;
+    pto_counts[1] = 1;
     assert_eq!(client.stats.borrow().pto_counts, pto_counts);
```
mxinden: See documentation for `Stats::pto_counts`:

https://github.com/mozilla/neqo/blob/3b8fac0f33b2a36337ed36fa6cc8d7e7b01ca42b/neqo-transport/src/stats.rs#L184-L186

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -203,7 +203,7 @@ fn pto_initial() {
     assert_eq!(delay, INITIAL_PTO * 3);
 }
 
-/// A complete handshake that involves a PTO in the Handshake space.
+/// A complete handshake that involves two PTOs in the Handshake space.
 #[test]
 fn pto_handshake_complete() {
```
mxinden, line 208: If I understand correctly, this unit test has been testing time threshold retransmits and not PTOs before. Thus with this pull request, the adjustments below are needed.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -255,38 +256,47 @@ fn pto_handshake_complete() {
...
+    let pto = pto * 2;
     let cb = client.process_output(now).callback();
-    assert_eq!(cb, HALF_RTT * 12);
+    assert_eq!(cb, pto);
 
-    // We still have only a single PTO
+    // Not one but two PTOs in a row, thus resetting first bit and increasing
+    // second bit.
+    pto_counts[0] = 0;
+    pto_counts[1] = 1;
```
mxinden, line 299: See documentation on `Stats::pto_counts` for data layout description:

https://github.com/mozilla/neqo/blob/3b8fac0f33b2a36337ed36fa6cc8d7e7b01ca42b/neqo-transport/src/stats.rs#L184-L186

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -255,38 +256,47 @@ fn pto_handshake_complete() {
...
     let pkt2 = client.process_output(now).dgram();
     assert_handshake(pkt2.as_ref().unwrap());
 
     pto_counts[0] = 1;
     assert_eq!(client.stats.borrow().pto_counts, pto_counts);
 
+    // PTO has been doubled.
+    let pto = 2 * pto;
+    let cb = client.process_output(now).callback();
+    assert_eq!(cb, pto);
```
mxinden, line 277: After a PTO expiry, the PTO doubles. Previously the test did not account for that, and thus trigger a time threshold based loss detection, but not a second PTO.

With the fix of this pull request, and the above accounting for PTO doubling, the test now actually sees a second PTO as expected.

mxinden:
@larseggert do you have some time to give this pull request a review?

neqo-transport/src/connection/tests/zerortt.rs
```diff
@@ -341,3 +341,89 @@ fn zero_rtt_loss_accepted() {
...
+        Output::Callback(_)
+    ));
+
+    now += RTT / 2;
+
+    // Despite the missing ACK from the server, expect client to neither declare
+    // 0-RTT stream data lost, nor attempt to retransmit it. Given that thus far
+    // no Application space ACK has been received, time threshold based loss
+    // detection does not yet apply.
+    assert!(client.process_output(dbg!(now)).dgram().is_none());
```
martinthomson, line 427:
```suggestion
    assert!(client.process_output(now).dgram().is_none());
```

Are you sure that the amount of elapsed time is more than the time loss recovery threshold?  I think that you are at a bit more than 1.5RTT at this point (plus the pacing delay).  That's more than 9/8 RTT, so I guess that's OK.  If my logic is correct, maybe just say that.

martinthomson (approved):
This LGTM.  The change is simple enough, even if the test changes are not.

neqo-transport/src/connection/tests/zerortt.rs
mxinden: > Are you sure that the amount of elapsed time is more than the time loss recovery threshold?

Yes, the elapsed time is more than the time loss recovery threshold. Executing this test without the fix in this pull request, the `client` does falsely declare the 0-RTT application data as lost in `client.process_output(dbg!(now)).dgram().is_none()` above.

More specifically it prints the following log lines:

```
0.023 TRACE detect lost ap: now=Instant { tv_sec: 20664, tv_nsec: 609522841 } delay=10.125ms
0.023 TRACE lost=0, time sent Instant { tv_sec: 20664, tv_nsec: 598522841 } is before lost_delay 10.125ms
```

Executing the following `if` arm:

https://github.com/mozilla/neqo/blob/ddb88ac993330379335d921eb68dbd6211fe2840/neqo-transport/src/recovery/mod.rs#L336-L340

>  That's more than 9/8 RTT, so I guess that's OK. If my logic is correct, maybe just say that.

Good idea. Will add a comment above. Thanks for the thorough review @martinthomson.

mxinden: For the record, done in https://github.com/mozilla/neqo/pull/2492/commits/30cdfaefd1ba3ceb7b81e8bfabdce005f2108054.

mxinden:
Thanks Lars, Kershaw and Martin for discovering the bug, the previous work through https://github.com/mozilla/neqo/pull/1998 and the reviews!

PR #2495, opened 2025-03-13
refactor(transport): consolidate max ACK delay constants

mxinden:
- Introduce `DEFAULT_REMOTE_ACK_DELAY` constant.
- Rename `DEFAULT_ACK_DELAY` to `DEFAULT_LOCAL_ACK_DELAY`.


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -613,7 +613,6 @@ mod tests {
     };
 
     const PTO: Duration = RTT;
-    const RTT_ESTIMATE: RttEstimate = RttEstimate::from_duration(RTT);
```
mxinden, line 616: `RttEstimate::from_duration` is no longer `const` as it uses `Default::default` of `PerrAckDelay`.

This affects test code only, thus simply replacing this constant with `RttEstimate::default`.

larseggert: Can we make it `#[cfg(test)]` in that case?

neqo-transport/src/tparams.rs
```diff
@@ -398,7 +399,10 @@ impl TransportParameters {
             | MaxDatagramFrameSize => 0,
             MaxUdpPayloadSize => 65527,
             AckDelayExponent => 3,
-            MaxAckDelay => 25,
+            MaxAckDelay => DEFAULT_REMOTE_ACK_DELAY
+                .as_millis()
+                .try_into()
+                .expect("delay fit u64"),
```
martinthomson, line 405:
```suggestion
                .expect("default remote ack delay in ms can't overflow u64"),
```

PR #2496, opened 2025-03-14
fix: Skip interface MTU query on OpenBSD

larseggert:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1952304

Since we'd now need to pass multiple `bool`s from the `ConnectionParameters` around, just pass the entire `struct`.

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -125,6 +125,7 @@ impl Default for ConnectionParameters {
             pmtud: false,
             // Firefox on OpenBSD is sandboxed and cannot access the routing socket that the mtu
             // crate uses to obtain the local interface MTU. (Trying to do so crashes the process.)
+            // See https://bugzilla.mozilla.org/show_bug.cgi?id=1952304
             pmtud_iface_mtu: cfg!(not(target_os = "openbsd")),
             sni_slicing: true,
             mlkem: true,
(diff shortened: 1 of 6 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -121,6 +123,9 @@ impl Default for ConnectionParameters {
             disable_migration: false,
             pacing: true,
             pmtud: false,
+            // Firefox on OpenBSD is sandboxed and cannot access the routing socket that the mtu
+            // crate uses to obtain the local interface MTU. (Trying to do so crashes the process.)
```
emilio, line 127: Maybe link to [bug 1952304](https://bugzilla.mozilla.org/show_bug.cgi?id=1952304) for more context?

mxinden: Given that the OpenBSD issue is Firefox specific, I prefer moving the `cfg!(not(target_os = "openbsd"))` to `neqo_glue`, and instead, set the default to always `true` here in `neqo-transport`.

We can then link to the Bugzilla Bug in `neqo_glue`.

mxinden (approved):
Thanks!

I set myself a reminder to set the parameter to `false` on OpenBSD in `neqo_glue` once we downstream `neqo-*` `v0.13.0`.

PR #2499, opened 2025-03-17
feat: Randomize CI packet number

### LINKED ISSUE #2462: Initial packet number need not be zero
We should investigate using a random non-zero small integer as the initial packet number for a connection, to increase entropy a bit.

larseggert:
This randomizes the starting packet number the client uses for the Initial packet number space.

We don't randomize this on the server, since otherwise we'd need even more changes to the tests to account for that.

Fixes #2462

CC @omansfeld


neqo-transport/src/crypto.rs
```diff
@@ -569,17 +585,17 @@ impl CryptoDxState {
...
+                self.min_pn,
                 prev.used_pn.end,
             );
             Err(Error::PacketNumberOverlap)
         } else {
-            self.used_pn.start = next;
+            self.used_pn = next..max(next, self.used_pn.end);
             Ok(())
         }
     }
```
omansfeld, line 601: I don't understand why this function is being changed and how that relates to the initial packet number being non-zero. Maybe it has something to do with https://github.com/mozilla/neqo/issues/2462#issuecomment-2708205236?

Per line 585f `next == prev.used_pn.end == self.min_pn` so aren't we effectively evaluating
```rust
} else if prev.used_pn.end > prev.used_pn.end {
...
}
```
in line 590? Which can never be true and is also logically very different from what it was before (comparing to `self.used_pn.start`).

I also don't quite understand the debug message (line 592), because aren't we checking for an already used packet number in `self` here? To me "too new packet number" would mean it's too high but the expression checks for it being too low.

larseggert: This PR was a quick hack. Take it more as a pointer in the direction that you might want to take.

omansfeld: Ah, misunderstood, thanks! I'll re-add it to my to-do's.

larseggert: Thanks for your feedback, that was useful and let to me properly understanding the code. I think the current version is correct now.

neqo-transport/src/crypto.rs
```diff
@@ -979,15 +988,21 @@ impl CryptoStates {
             Role::Server => (SERVER_INITIAL_LABEL, CLIENT_INITIAL_LABEL),
         };
 
+        let min_pn = if randomize_ci_pn {
+            packet::Number::from((random::<1>()[0] >> 4) + 1)
```
martinthomson: Why skip 0?  Isn't it a good idea to include that in the range?  If you want lower odds: (r >> 4) + (r  & 0xf) returns a value of 0 pretty rarely.  Though I tend to like (r >> 3) + (r & 0x7) for keeping it in one byte.

Also, what logic are you using to decide that 16 is the largest value?  Obviously, if we bump this up to 64, we'll have to contend with the whole thing where the packet number needs to be longer.  But that seems harmless.  We use largest_acknowledged, so it's probably fine to use any starting number.  And we can afford to spend a byte on this.

Given that we are safe in including any value here, could we instead do `let r = random::<2>(); r[0] & r[1]` ?  That has a 7/16 chance of slipping into the next byte, but no more than that.  It also pushes more toward having smaller values overall.

neqo-transport/src/stats.rs
```diff
@@ -274,6 +274,9 @@ pub struct Stats {
     /// Whether the connection was resumed successfully.
     pub resumed: bool,
 
+    /// The packet number of the first Initial packet sent.
+    pub first_initial_pn: Option<packet::Number>,
```
martinthomson: We shouldn't need to record this.  Maybe if that is your method for testing this, but I'd prefer that the packet be decrypted instead.

neqo-transport/src/crypto.rs
```diff
@@ -471,17 +471,18 @@ impl CryptoDxState {
         epoch: Epoch,
         secret: &SymKey,
         cipher: Cipher,
+        min_pn: packet::Number,
```
martinthomson, line 474: The alternative here is to have a `skip_pn()` method that you invoke for client initials specially.  This would be factored out of the `continuation()` method.

martinthomson (changes requested):
You don't test that the client Initial has a non-zero initial packet number.

neqo-transport/src/crypto.rs
larseggert: I skipped zero, so that it would be easy to do a test for it (which I then failed to do.)

The logic behind 16 was simply to pick a number that gave us some small randomization without eating up too much of the 0-255 range. Am very open to doing something different.

neqo-transport/src/stats.rs
larseggert: Didn't want this for tests, mostly wanted it for telemetry in case something goes wrong in Gecko. We don't have an obvious other way to determine if the initial pn wasn't zero?

omansfeld (approved):
Looks good to me with the added test and pending the discussion about randomization.

neqo-transport/src/crypto.rs
larseggert: There is now a test.

neqo-transport/src/stats.rs
martinthomson: If you are concerned about this causing problems (others have been doing this for ages, so I very much doubt that) then shouldn't you run an experiment with this on/off rather than just measuring it?

In general, I don't think we should be adding metrics for these tiny features.  The value of this tweak is pretty marginal and we have to carry 9 bytes of state and all the extra code just to populate it.

neqo-transport/src/crypto.rs
martinthomson: You should make it more random, I think.  It doesn't hurt if this uses more of the range.  Even beyond 63.  Though I'd aim for something that did less than 63 most times, either way.

martinthomson (approved):

neqo-transport/src/crypto.rs
larseggert: I tried your last suggestion, but we then need to fix the `reorder_handshake` test, which currently fails when we get two-byte packet numbers. Going with 
```rust
packet::Number::from((r >> 3) + (r & 0x7)) + 1
```

martinthomson: I can't easily see how `reorder_handshake` fails.  If the failure happens after two-bytes, just disable the initial packet number tweak for that test.

larseggert:
```
    thread 'connection::tests::handshake::reorder_handshake' (12105956) panicked at neqo-transport/src/connection/tests/handshake.rs:566:5:
    assertion `left == right` failed
      left: 1
     right: 3
```
I.e., it's only receiving one packet by then.

I can make the change, would then do 
```rust
let r = random::<2>();
(r[0] & r[1]) + 1
```

larseggert: @martinthomson so, about that failing test, I wonder if we actually have a bug somewhere. What seems to happen with two-byte packet numbers is that Handshake keys are not available after the 
```rust
client.process_input(s_initial_2, now);
```
line (`neqo-transport/src/connection/tests/handshake.rs:564`), whereas they are available (and saved packets are hence processed) with single-byte packet numbers. Odd...

martinthomson:
Nice way to find a boundary condition, @larseggert.

Our tests operate with a 1232-byte MTU.  In this particular test, the packet was being filled completely and precisely.  The ServerHello is 1178 bytes.  In a CRYPTO frame, that's 1182 bytes.  Add an ACK frame with a one-byte largest acknowledged and a gap for the dropped packet (7 bytes), the header (27) and an auth tag (16) and you get EXACTLY 1232 bytes.

In this case, we're adding one byte to the size of the ACK frame because the largest acknowledged needs two bytes to encode.  So the CRYPTO frame needs to have its last byte cut off to fit.

~~The reason we don't bother with another Initial packet is unclear to me.~~  (Edit: worked it out...)  If we added the remainder to the second packet, then that would have been processed and the keys would be available.  We have sent an Initial before and that one had a CRYPTO frame that fits the packet (because that didn't have a gap in the ACK frame, so it just fit) and that was marked as needing retransmission so that it could be retransmitted.  So all of the bytes were marked as needing retransmission.

The challenge here is that we have lost ONE packet, but we need to transmit TWO packets to repair the loss properly.  But our PTO code only sends two packets for Application packets, whereas it only sends ONE for Initial and Handshake packets.  That is because we want to avoid pointless PING-only packets when repairing the handshake.  So we limited handshake PTOs to a single packet.  That means that the PTO code moves on to sending a Handshake packet, rather than sending the last byte of that CRYPTO frame.

If we had sent TWO packets, we could maybe have made the PTO code pay attention to the number of packets that were marked as needing retransmission, but in this very narrow case, we need to repair the loss of one packet with two.  The alternative is to add some smarts to check for outstanding CRYPTO data when sending a second PTO, rather than capping the number of PTOs to 1.

A hacky workaround here might be to increase the MTU.  Though that just defers the problem for later.

larseggert:
Closing in favor of #2885.

PR #2508, opened 2025-03-18
ci: Don't fail QNS CI if a test times out

larseggert:
This gets rid of the :x: on all commits to `main` just because one or two QNS runs timed out.


mxinden:
How do we catch a bug early that results in QNS timeouts?

larseggert:
You'd need to check the reported errors.

AFAICT the current timeouts are bugs in the other implementations. (But maybe someone else should take a look.)

mxinden (approved):

PR #2509, opened 2025-03-19
fix(transport): make idle timeout PING piggy back on PTO PING 

mxinden:
Given a connection with both PTO and idle timeout firing, do not send two PINGs. Instead, have the idle timeout piggy back on the PTO PING.

---

Broken out of https://github.com/mozilla/neqo/pull/2129 and added unit test.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2222,24 +2222,26 @@ impl Connection {
             return true;
         }
 
-        let probe = if untracked && builder.packet_empty() || force_probe {
+        let pto = path.borrow().rtt().pto(self.confirmed());
+        let mut probe = if untracked && builder.packet_empty() || force_probe {
             // If we received an untracked packet and we aren't probing already
             // or the PTO timer fired: probe.
             true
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2222,24 +2222,26 @@ impl Connection {
...
-            }
+            false
         };
+
+        if self.streams.need_keep_alive() {
+            // We need to keep the connection alive, including sending a PING
+            // again. If a PING is already scheduled (i.e. `probe` is `true`)
+            // piggy back on it. If not, schedule one.
+            probe |= self.idle_timeout.send_keep_alive(now, pto, tokens);
+        }
```
mxinden, line 2243: Slightly different implementation compared to https://github.com/mozilla/neqo/pull/2129 for the following reasons:

- Single call to `self.idle_timeotu.send_keep_alive` only.
- No call to `self.idle_timeotu.send_keep_alive` on `self.streams.need_keep_alive()` `false`.

larseggert (approved):

PR #2033, opened 2024-08-02
chore: Remove some `unwrap`s where it is straightforward

larseggert:
This will hopefully include fixes for this issue:
```
core::result::unwrap_failed | neqo_transport::crypto::CryptoDxState::new_initial
```

https://crash-stats.mozilla.org/signature/?product=Firefox&signature=core%3A%3Aresult%3A%3Aunwrap_failed%20%7C%20neqo_transport%3A%3Acrypto%3A%3ACryptoDxState%3A%3Anew_initial&date=%3E%3D2024-01-01T12%3A31%3A00.000Z&date=%3C2024-08-02T12%3A31%3A00.000Z&_columns=date&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=reason&_columns=address&_columns=install_time&_columns=startup_crash&_sort=-date&page=1#reports

PR #2511, opened 2025-03-20
fix: Bubble up errors from `Crypto::confirm_version`

larseggert:
I don't think they should have been suppressed.


mxinden (approved):

martinthomson:
Do you have any analysis to support this change?  The suppression was obviously deliberate.

larseggert:
I don't think the suppression was deliberate. I added it in #2033. Before that PR, `confirm_version` would panic under the error conditions that are now passed up (and then IMO erroneously) ignored.

martinthomson:
OK, I still think that a test is a good idea.  If it is erroneous, that error should be reachable.

larseggert:
Well, I didn't manage to write a test for _that_ condition yet, but this one causes an endless loop :-)
```rust
fn compatible_xxx() {
    let mut client = new_client(
        ConnectionParameters::default().versions(Version::Version1, vec![Version::Version1]),
    );
    let mut server = new_server(
        ConnectionParameters::default().versions(Version::Draft29, vec![Version::Draft29]),
    );

    connect(&mut client, &mut server);
    assert_eq!(client.version(), Version::Version1);
    assert_eq!(server.version(), Version::Version1);
}
```

martinthomson:
Of course it does.  Our client never gives up and you don't have a version in common.

larseggert:
Ah, you're right of course that our testing `connect` logic doesn't handle that.

larseggert:
I can't come up with a test :shrug:

martinthomson:
Oh well, low impact.  Ship the fix anyway.  It's not like it is a terrible idea to surface this particular error.

PR #2513, opened 2025-03-21
fix: Don't count saved packets as received twice

larseggert:
We count packets as received when they enter `input_path`. Then we may save them, since we can't decrypt them yet. When we re-process them, they again enter `input_path` and are counted as received. Fix that double-counting by only counting them when they are actually processed.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1319,6 +1319,9 @@ impl Connection {
         );
         self.saved_datagrams.save(epoch, d, now);
         self.stats.borrow_mut().saved_datagrams += 1;
+        // We already counted the datagram as received in `[input_path]`. We
+        // will do so again when we (re-)process it, so reduce the count now.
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1319,6 +1319,9 @@ impl Connection {
         );
         self.saved_datagrams.save(epoch, d, now);
         self.stats.borrow_mut().saved_datagrams += 1;
+        // We already counted the datagram as received in `[input_path]`. We
```
mxinden, line 1322:
```suggestion
        // We already counted the datagram as received in [`input_path`]. We
```

Convention is to put the ticks inside.

https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

Note that this will not be picked up by `cargo docs`, given that it is a `//` and not a `///`.

mxinden (approved):

PR #2516, opened 2025-03-23
fix(transport/ecn): handle ECN loss on PTO

mxinden:
Previously we would only call `neqo_transport::ecn::Info::on_packets_lost` when an ECN ECT0 marked packet was considered lost due to the [time-threshold]. It would *not* be called when a packet was considered lost due to a PTO.

With https://github.com/mozilla/neqo/pull/2492 one no longer considers a packet lost based on the time-threshold before the first ACK, entirely depending on the PTO until then.

Thus with https://github.com/mozilla/neqo/pull/2492 the ECN path validation mechanism would no longer detect an ECN blackhole, dropping all initial packets.

With this commit the ECN path validation state machine (i.e. `neqo_transport::ecn::Info`) is notified on PTO and can thus react to an ECN blackhole, disabling ECN markings after `TEST_COUNT_INITIAL_PHASE`.

`neqo_transport` already provides the `RecoveryToken` mechanism to track losses. This commit introduces `RecoveryToken::EcnEct0` to track losses of ECN ECT0 marked packets *both* based on [time-threshold] and [PTO].

---

Given that we currently do time threshold based loss detection before the first ACK (bug to be fixed by https://github.com/mozilla/neqo/pull/2492), and given that the time threshold is smaller than the PTO, the fix in this pull request will only be exercised once https://github.com/mozilla/neqo/pull/2492 is merged.

Once https://github.com/mozilla/neqo/pull/2492/ is merged, the bug fix of this pull request will then also be covered by unit tests.

Blocks https://github.com/mozilla/neqo/pull/2492/.

[time-threshold]: https://www.rfc-editor.org/rfc/rfc9002.html#name-time-threshold
[PTO]: https://www.rfc-editor.org/rfc/rfc9002.html#name-probe-timeout

### DIFF
--- a/neqo-transport/benches/sent_packets.rs
+++ b/neqo-transport/benches/sent_packets.rs
@@ -7,7 +7,6 @@
 use std::time::Instant;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use neqo_common::IpTosEcn;
 use neqo_transport::{
     packet::{PacketNumber, PacketType},
     recovery::sent::{SentPacket, SentPackets},
@@ -21,7 +20,6 @@ fn sent_packets() -> SentPackets {
         pkts.track(SentPacket::new(
             PacketType::Short,
             PacketNumber::from(i),
-            IpTosEcn::default(),
             now,
             true,
             Vec::new(),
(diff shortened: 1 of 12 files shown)


larseggert (approved):
LGTM

PR #2524, opened 2025-03-24
fix: Convert large pass-by-value args to references

larseggert:
Let's see if this matters.


martinthomson:
Are those gains real?  I no longer know whether the benchmarks can be trusted.

PR #2525, opened 2025-03-25
fix: Don't try `write_maintenance_frames` at all priorities

larseggert:
The old code would try and call `write_maintenance_frames` at priorities other than `TransmissionPriority::Important`, which was useless. Just call it directly once.


martinthomson (approved):
This is exactly identical, so it's fine.

PR #2528, opened 2025-03-25
fix: Switch from `tokio` `multi_thread` to `current_thread`

larseggert:
This seems to speed up things quite a bit.

Also fix the percentage change calculation in the CI table. I don't know what I was thinking when I did the original formula. Math is hard I guess.


martinthomson (approved):
This seems to favour large request/response sizes.  I'm OK with that.

larseggert:
I'd like @mxinden to take a look, since the original code was his.

mxinden (approved):
Oh, my intuition was very wrong. Thank you for testing this out!

(Next to the performance improvements, this also gets us closer to how Firefox uses Neqo.)

PR #2530, opened 2025-03-26
ci: `bench` improvements

larseggert:
- Use `cset` to isolate benchmarks
- Stop `google-quiche`'s `bazel` server
- Show throughputs in results table


larseggert:
I'm landing this now, so we can get updated bench baselines. Can tweak more after.

PR #2535, opened 2025-03-28
CUBIC RFC 9438

omansfeld:
This initial draft PR just scopes the update from our implementation to RFC 9438 in code comments, updates RFC references and doesn't change any code yet.

I mostly did this for myself to scope the update and check what changed, but thought I might upload it as a draft already so people can give their input if I might've missed something or need some additional context somewhere.

Notes about what changed are prefixed with the keyword `UPDATE` and open questions that I still need to look into or gather feedback on with the keyword `QUESTION`.

I'll have to focus on some other things the next weeks, but will chip away on this on the side.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -552,6 +552,10 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
             self.acked_bytes,
             self.max_datagram_size(),
         );
+        // UPDATE: Add condition for `self.cc_algorithm == cubic` here to set `ssthresh` and
+        // `congestion_window` according to RFC 9438.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#figure-5>
         self.congestion_window = max(cwnd, self.cwnd_min());
         self.acked_bytes = acked_bytes;
         self.ssthresh = self.congestion_window;
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -15,21 +15,35 @@ use neqo_common::qtrace;
 
 use crate::cc::classic_cc::WindowAdjustment;
 
-/// > C is a constant fixed to determine the aggressiveness of window
-/// > increase  in high BDP networks.
+/// > Constant that determines the aggressiveness of CUBIC in
+/// > competing with other congestion control algorithms in high-BDP networks.
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-constants-of-interest>
 ///
-/// See discussion for rational for concrete value.
+/// See section 5.1 for discussion on how to set the concrete value.
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-5.1>
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-fairness-to-reno>
+///
+/// UPDATE: no change to C.
 pub const CUBIC_C: f64 = 0.4;
-/// TCP-friendly region additive factor
-///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7);
 
-/// `CUBIC_BETA` = 0.7;
+/// > CUBIC additive increase factor used in the Reno-friendly region
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-constants-of-interest>
+///
+/// > The model used to calculate CUBIC_ALPHA is not absolutely precise,
+/// > but analysis and simulation \[...\], as well as over a decade of experience with
+/// > CUBIC in the public Internet, show that this approach produces acceptable
+/// > levels of rate fairness between CUBIC and Reno flows.
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+///
+/// UPDATE: no change to alpha.
+pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7); // with CUBIC_BETA = 0.7
+
+/// > CUBIC multiplicative decrease factor
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-constants-of-interest>
 ///
 /// > Principle 4: To balance between the scalability and convergence speed,
 /// > CUBIC sets the multiplicative window decrease factor to 0.7 while Standard
@@ -37,16 +51,22 @@ pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7);
 /// > of this decision is slower convergence, especially under low statistical
 /// > multiplexing environments.
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-3>
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-principle-4-for-the-cubic-d>
+///
+/// `CUBIC_BETA` = 0.7;
+///
+/// UPDATE: no change to beta.
 pub const CUBIC_BETA_USIZE_DIVIDEND: usize = 7;
 pub const CUBIC_BETA_USIZE_DIVISOR: usize = 10;
 
-/// The fast convergence ratio further reduces the congestion window when a
-/// congestion event occurs before reaching the previous `W_max`.
+/// The fast convergence ratio further reduces the next `W_max` when a
+/// congestion event occurs while `cwnd < W_max`.
 ///
 /// See formula defined below.
 ///
-/// <https://www.rfc-editor.org/rfc/rfc8312#section-4.6>
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence>
+///
+/// UPDATE: no change to fast convergence ratio.
 pub const CUBIC_FAST_CONVERGENCE: f64 = 0.85; // (1.0 + CUBIC_BETA) / 2.0;
 
 /// The minimum number of multiples of the datagram size that need
@@ -54,11 +74,15 @@ pub const CUBIC_FAST_CONVERGENCE: f64 = 0.85; // (1.0 + CUBIC_BETA) / 2.0;
 /// When there is no loss, Cubic can return to exponential increase, but
 /// this value reduces the magnitude of the resulting growth by a constant factor.
 /// A value of 1.0 would mean a return to the rate used in slow start.
+///
+/// UPDATE: Not found in RFC. What is the reason for this and should we reconsider?
 const EXPONENTIAL_GROWTH_REDUCTION: f64 = 2.0;
 
 /// Convert an integer congestion window value into a floating point value.
 /// This has the effect of reducing larger values to `1<<53`.
 /// If you have a congestion window that large, something is probably wrong.
+///
+/// UPDATE: not found in RFC.
 pub fn convert_to_f64(v: usize) -> f64 {
     let mut f_64 = f64::from(u32::try_from(v >> 21).unwrap_or(u32::MAX));
     f_64 *= 2_097_152.0; // f_64 <<= 21
@@ -78,33 +102,67 @@ pub struct Cubic {
     /// > event.
     ///
     /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.6>
+    ///
+    /// UPDATE: The algorithm for fast convergence was improved in RFC 9438
+    /// to not need the extra variable anymore.
+    ///
+    /// ACTION: Delete variable when algorithm is adapted.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence>
     last_max_cwnd: f64,
-    /// Estimate of Standard TCP congestion window for Cubic's TCP-friendly
-    /// Region.
+    /// > An estimate for the congestion window in segments in the Reno-friendly region -- that
+    /// > is, an estimate for the congestion window of Reno.
     ///
-    /// > Standard TCP performs well in certain types of networks, for example,
-    /// > under short RTT and small bandwidth (or small BDP) networks.  In
-    /// > these networks, we use the TCP-friendly region to ensure that CUBIC
-    /// > achieves at least the same throughput as Standard TCP.
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-variables-of-interest>
     ///
-    /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
+    /// > Reno performs well in certain types of networks -- for example, under short RTTs and small
+    /// > bandwidths (or small BDPs). In these networks, CUBIC remains in the Reno-friendly region to
+    /// > achieve at least the same throughput as Reno.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+    ///
+    /// ACTION: Rename to `w_est` to conform with spec.
     estimated_tcp_cwnd: f64,
-    /// > K is the time period that the above function takes to increase the
-    /// > current window size to W_max if there are no further congestion events
+    /// > The time period in seconds it takes to increase the congestion window size
+    /// > at the beginning of the current congestion avoidance stage to `w_max`.
     ///
-    /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-variables-of-interest>
+    ///
+    /// For formula definition see:
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-window-increase-function>
     k: f64,
-    /// > W_max is the window size just before the window is reduced in the last
-    /// > congestion event.
+    /// > Size of `cwnd` in segments just before `cwnd` was reduced in the last congestion
+    /// > event when fast convergence is disabled (same as `cwnd_prior` on a congestion event).
+    /// > However, if fast convergence is enabled, `w_max` may be further reduced based on
+    /// > the current saturation point.
     ///
-    /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-variables-of-interest>
+    ///
+    /// This acts as the plateau for the cubic function where it switches from the concave to the
+    /// convex region.
+    ///
+    /// For formula definition see:
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence>
     w_max: f64,
-    /// > the elapsed time from the beginning of the current congestion
-    /// > avoidance
+    /// > The time in seconds at which the current congestion avoidance stage started.
     ///
-    /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-variables-of-interest>
+    ///
+    /// UPDATE/ACTION: Rename to `t_epoch` to adhere to spec.
     ca_epoch_start: Option<Instant>,
     /// Number of bytes acked since the last Standard TCP congestion window increase.
+    ///
+    /// UPDATE: We are using bytes but the spec recommends using segments.
+    ///
+    /// > Implementations can use bytes to express window sizes, which would require
+    /// > factoring in the SMSS wherever necessary and replacing segments_acked (Figure 4)
+    /// > with the number of acknowledged bytes.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-definitions>
+    ///
+    /// ACTIONS: Discuss if we want to switch to segments, confirm we are doing all conversions.
     tcp_acked_bytes: f64,
 }
 
@@ -146,18 +204,49 @@ impl Cubic {
     /// K = cubic_root((W_max - W_cubic) / C / MSS);
     ///
     /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
+    ///
+    /// UPDATE: Formula for K changed.
+    ///
+    /// Original equation is:
+    /// `k = cubic_root((w_max - cwnd_epoch)/C)` with `cwnd_epoch` being the congestion
+    /// window at the start of the current congestion avoidance stage (so at time t_epoch).
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#figure-2>
+    ///
+    /// Taking into account that we're using bytes not MSS units, the formula becomes:
+    ///
+    /// `k = cubic_root((w_max - cwnd_epoch)/C/MSS)`
+    ///
+    /// ACTION: change `calc_k`, add `cwnd_epoch`
     fn calc_k(&self, curr_cwnd: f64, max_datagram_size: usize) -> f64 {
         ((self.w_max - curr_cwnd) / CUBIC_C / convert_to_f64(max_datagram_size)).cbrt()
     }
 
-    /// W_cubic(t) = C*(t-K)^3 + W_max (Eq. 1)
-    /// t is relative to the start of the congestion avoidance phase and it is in seconds.
+    /// `w_cubic(t) = C*(t-K)^3 + w_max`
     ///
-    /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
+    /// with `t = t_current - t_epoch`.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#figure-1>
+    ///
+    /// We're using bytes not MSS units so we need to convert.
+    ///
+    /// UPDATE: Nothing changed.
     fn w_cubic(&self, t: f64, max_datagram_size: usize) -> f64 {
         (CUBIC_C * (t - self.k).powi(3)).mul_add(convert_to_f64(max_datagram_size), self.w_max)
     }
 
+    /// > w_est is set equal to cwnd_epoch at the start of the congestion avoidance stage.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+    ///
+    /// UPDATE: With the change to fast convergence, we can remove the `last_max_cwnd` logic here.
+    /// We also must set `w_max` and `k`. I think `k` should be here but `w_max` should happen in
+    /// `reduce_cwnd()` where we currently do fast convergence. Also need to set `w_est` here,
+    /// see above. I don't think we need a `cwnd_epoch` variable, as it's only used here and equals
+    /// `curr_cwnd_f64`. Might rename the parameter.
+    ///
+    /// Not sure about the `last_max_cwnd <= curr_cwnd_f64` logic and how that relates to the new
+    /// fast convergence. Need to look into that.
     fn start_epoch(
         &mut self,
         curr_cwnd_f64: f64,
@@ -203,10 +292,27 @@ impl WindowAdjustment for Cubic {
             self.tcp_acked_bytes += new_acked_f64;
         }
 
-        // Cubic concave or convex region
+        // Calculate `target` for the concave or convex region
         //
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.3>
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.4>
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-concave-region>
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-convex-region>
+        //
+        // UPDATE: New logic for how `target` is calculated.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1>
+        //
+        // And the `cwnd` increase of `target - cwnd / cwnd` only applies here,
+        // not in the Reno-friendly region. So that needs to be adjusted. See wording:
+        //
+        // > cwnd SHOULD be set to w_est
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
+        //
+        // vs.
+        //
+        // > cwnd MUST be incremented by `target - cwnd / cwnd`
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-convex-region>
         let time_ca = self
             .ca_epoch_start
             .map_or(min_rtt, |t| {
@@ -221,9 +327,13 @@ impl WindowAdjustment for Cubic {
             .as_secs_f64();
         let target_cubic = self.w_cubic(time_ca, max_datagram_size);
 
-        // Cubic TCP-friendly region
+        // Reno-friendly region
         //
-        //  <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+        //
+        // UPDATE: This is handled differently in the new RFC, but it's also
+        // different from the original RFC. Still need to understand what exactly
+        // is going on here.
         let max_datagram_size = convert_to_f64(max_datagram_size);
         let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
         let incr = (self.tcp_acked_bytes / tcp_cnt).floor();
@@ -232,15 +342,22 @@ impl WindowAdjustment for Cubic {
             self.estimated_tcp_cwnd += incr * max_datagram_size;
         }
 
-        // Take the larger cwnd of Cubic concave or convex and Cubic
-        // TCP-friendly region.
+        // Take the larger cwnd of Cubic concave or convex and Cubic Reno-friendly region.
         //
-        // > When receiving an ACK in congestion avoidance (cwnd could be
-        // > greater than or less than W_max), CUBIC checks whether W_cubic(t) is
-        // > less than W_est(t).  If so, CUBIC is in the TCP-friendly region and
-        // > cwnd SHOULD be set to W_est(t) at each reception of an ACK.
+        // > When receiving a new ACK in congestion avoidance (where cwnd could be greater than
+        // > or less than Wmax), CUBIC checks whether Wcubic(t) is less than w_est.  If so, CUBIC
+        // > is in the Reno-friendly region and cwnd SHOULD be set to w_est at each reception of a new ACK.
         //
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
+        //
+        // UPDATE: We should compare `w_cubic` (not `target`) and `w_est` here.
+        // Maybe it might make sense to change the order of things a bit:
+        // 1. Calculate `w_cubic` and `w_est`
+        // 2. Compare `w_cubic` and `w_est`
+        // 2.b. Maybe get `target` from `w_cubic`
+        // 3.a. Either get `acked_to_increase` from `target` (with `target - cwnd / cwnd`) OR
+        // 3.b. Get `acked_to_increase` from `w_est` by calculating the difference between
+        // `curr_cwnd` and `w_est`.
         let target_cwnd = target_cubic.max(self.estimated_tcp_cwnd);
 
         // Calculate the number of bytes that would need to be acknowledged for an increase
@@ -249,6 +366,8 @@ impl WindowAdjustment for Cubic {
         // The amount of data required therefore reduces asymptotically as the target increases.
         // If the target is not significantly higher than the congestion window, require a very
         // large amount of acknowledged data (effectively block increases).
+        //
+        // UPDATE: See above.
         let mut acked_to_increase =
             max_datagram_size * curr_cwnd_f64 / (target_cwnd - curr_cwnd_f64).max(1.0);
 
@@ -302,3 +421,7 @@ impl WindowAdjustment for Cubic {
         self.last_max_cwnd = last_max_cwnd;
     }
 }
+
+// UPDATE: Things to remember that didn't make it anywhere else:
+//
+// - what about `ssthresh`
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -75,14 +75,14 @@ pub const CUBIC_FAST_CONVERGENCE: f64 = 0.85; // (1.0 + CUBIC_BETA) / 2.0;
 /// this value reduces the magnitude of the resulting growth by a constant factor.
 /// A value of 1.0 would mean a return to the rate used in slow start.
 ///
-/// UPDATE: Not found in RFC. What is the reason for this and should we reconsider?
+/// UPDATE: Not found in RFC. I don't exactly understand why we're doing this?
 const EXPONENTIAL_GROWTH_REDUCTION: f64 = 2.0;
 
 /// Convert an integer congestion window value into a floating point value.
 /// This has the effect of reducing larger values to `1<<53`.
 /// If you have a congestion window that large, something is probably wrong.
 ///
-/// UPDATE: not found in RFC.
+/// UPDATE: not part of RFC.
 pub fn convert_to_f64(v: usize) -> f64 {
     let mut f_64 = f64::from(u32::try_from(v >> 21).unwrap_or(u32::MAX));
     f_64 *= 2_097_152.0; // f_64 <<= 21
@@ -302,15 +302,17 @@ impl WindowAdjustment for Cubic {
         // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1>
         //
         // And the `cwnd` increase of `target - cwnd / cwnd` only applies here,
-        // not in the Reno-friendly region. So that needs to be adjusted. See wording:
+        // not in the Reno-friendly region. So that needs to be adjusted. Right now
+        // we are doing the `target - cwnd / cwnd` part for all regions I think.
+        // See wording:
         //
-        // > cwnd SHOULD be set to w_est
+        // > cwnd SHOULD be set to w_est ("set to")
         //
         // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
         //
         // vs.
         //
-        // > cwnd MUST be incremented by `target - cwnd / cwnd`
+        // > cwnd MUST be incremented by `target - cwnd / cwnd` ("incremented by")
         //
         // <https://datatracker.ietf.org/doc/html/rfc9438#name-convex-region>
         let time_ca = self
@@ -331,9 +333,17 @@ impl WindowAdjustment for Cubic {
         //
         // <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
         //
-        // UPDATE: This is handled differently in the new RFC, but it's also
-        // different from the original RFC. Still need to understand what exactly
-        // is going on here.
+        // UPDATE/QUESTION: This is handled differently in the new RFC, but our implementation is also
+        // different from the original RFC. Still need to understand what exactly is going on here.
+        //
+        // Also note:
+        //
+        // > Once w_est has grown to reach the cwnd at the time of most recently setting
+        // > ssthresh -- that is, w_est >= cwnd_prior -- the sender SHOULD set CUBIC_ALPHA to
+        // > 1 to ensure that it can achieve the same congestion window increment rate
+        // > as Reno, which uses AIMD(1, 0.5).
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-11>
         let max_datagram_size = convert_to_f64(max_datagram_size);
         let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
         let incr = (self.tcp_acked_bytes / tcp_cnt).floor();
@@ -345,7 +355,7 @@ impl WindowAdjustment for Cubic {
         // Take the larger cwnd of Cubic concave or convex and Cubic Reno-friendly region.
         //
         // > When receiving a new ACK in congestion avoidance (where cwnd could be greater than
-        // > or less than Wmax), CUBIC checks whether Wcubic(t) is less than w_est.  If so, CUBIC
+        // > or less than w_max), CUBIC checks whether W_cubic(t) is less than w_est.  If so, CUBIC
         // > is in the Reno-friendly region and cwnd SHOULD be set to w_est at each reception of a new ACK.
         //
         // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
@@ -377,6 +387,41 @@ impl WindowAdjustment for Cubic {
         acked_to_increase as usize
     }
 
+    // UPDATE: CUBIC RFC 9438 changes the logic for multiplicative decrease and uses
+    // `bytes_in_flight` instead of `cwnd` for it's calculation:
+    //
+    // > ssthresh = bytes_in_flight * CUBIC_BETA
+    // > cwnd_prior = cwnd
+    // > if (reduction_on_loss) {
+    // >     cwnd = max(ssthresh, 2*MSS)
+    // > } else if (reduction_on_ece) {
+    // >     cwnd = max(ssthresh, 1*MSS)
+    // > }
+    // > ssthresh = max(ssthresh, 2*MSS)
+    //
+    // <https://datatracker.ietf.org/doc/html/rfc9438#figure-5>
+    //
+    // With the note:
+    //
+    // > A QUIC sender that uses a cwnd value to calculate new values for cwnd and ssthresh after
+    // > detecting a congestion event is REQUIRED to apply similar mechanisms [RFC9002].
+    //
+    // "similar mechanisms" refering to taking "measures to prevent cwnd from growing when the
+    // volume of bytes in flight is smaller than cwnd".
+    //
+    // QUESTION: Do we already do this somewhere?
+    //
+    // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.6>
+    //
+    // We are setting ssthresh in `on_congestion_event()` in `classic_cc.rs` after having returned `cwnd`
+    // from `reduce_cwnd()` below. That is code that isn't CUBIC specific though, so we can add a condition
+    // there to check which cc algorithm we're using and do the `ssthresh|cwnd = max(...)` outlined above
+    // for CUBIC only.
+    //
+    // (or maybe don't do it at all because QUIC has a minimum congestion window of 2*MSS, as mentioned
+    // further below)
+    // QUESTION: Which takes precedence? The minimum congestion window of 2*MSS for QUIC or CUBIC wanting to set
+    // cwnd to 1*MSS on reduction by ECE?
     fn reduce_cwnd(
         &mut self,
         curr_cwnd: usize,
@@ -386,12 +431,26 @@ impl WindowAdjustment for Cubic {
         let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
         // Fast Convergence
         //
-        // If congestion event occurs before the maximum congestion window before the last
-        // congestion event, we reduce the the maximum congestion window and thereby W_max.
-        // check cwnd + MAX_DATAGRAM_SIZE instead of cwnd because with cwnd in bytes, cwnd may be
+        // > During a congestion event, if the current cwnd is less than w_max, this indicates
+        // > that the saturation point experienced by this flow is getting reduced because of
+        // > a change in available bandwidth. This flow can then release more bandwidth by
+        // > reducing w_max further. This action effectively lengthens the time for this flow
+        // > to increase its congestion window, because the reduced w_max forces the flow to
+        // > plateau earlier. This allows more time for the new flow to catch up to its
+        // > congestion window size.
+        //
+        // Check cwnd + MAX_DATAGRAM_SIZE instead of cwnd because with cwnd in bytes, cwnd may be
         // slightly off.
         //
-        // <https://www.rfc-editor.org/rfc/rfc8312#section-4.6>
+        // UPDATE: New logic for fast convergence.
+        //
+        // if (cwnd < w_max AND fast_convergence_enabled) {
+        //     w_max = cwnd * CUBIC_FAST_CONVERGENCE;
+        // } else {
+        //     w_max = cwnd;
+        // }
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence>
         self.last_max_cwnd =
             if curr_cwnd_f64 + convert_to_f64(max_datagram_size) < self.last_max_cwnd {
                 curr_cwnd_f64 * CUBIC_FAST_CONVERGENCE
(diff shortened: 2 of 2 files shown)


neqo-transport/src/cc/cubic.rs
```diff
@@ -302,3 +480,35 @@ impl WindowAdjustment for Cubic {
...
+// (as is currently implemented)?
+//
+// 2. Timeout: <https://datatracker.ietf.org/doc/html/rfc9438#section-4.8>
+//
+// > QUIC does not collapse the congestion window when the PTO expires, since a single
+// > packet loss at the tail does not indicate persistent congestion.
+//
+// <https://datatracker.ietf.org/doc/html/rfc9002#section-4.7>
+//
+// QUESTION: Timeout (RTO/PTO) does not apply to us then?
```
mxinden, line 510: That is my understanding. See also https://github.com/mozilla/neqo/issues/2476 where we currently falsely assume a congestion event on PTO.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -552,6 +552,10 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
             self.acked_bytes,
             self.max_datagram_size(),
         );
+        // UPDATE: Add condition for `self.cc_algorithm == cubic` here to set `ssthresh` and
+        // `congestion_window` according to RFC 9438.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#figure-5>
```
mxinden, line 558: Not sure I am following. `cwnd` here is already based on the Cubic `reduce_cwnd` function, no?

neqo-transport/src/cc/cubic.rs
```diff
@@ -78,33 +102,67 @@ pub struct Cubic {
...
+    ///
+    /// UPDATE: We are using bytes but the spec recommends using segments.
+    ///
+    /// > Implementations can use bytes to express window sizes, which would require
+    /// > factoring in the SMSS wherever necessary and replacing segments_acked (Figure 4)
+    /// > with the number of acknowledged bytes.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-definitions>
+    ///
+    /// ACTIONS: Discuss if we want to switch to segments, confirm we are doing all conversions.
```
mxinden, line 165: For the protocol, limited previous discussion: https://github.com/mozilla/neqo/pull/1074#discussion_r546719993

mxinden (commented):
Thanks! Will need more time to work through the comments.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Yes it is. In RFC 9438 the logic for reducing `cwnd` changes, see pseudocode here:
 
https://github.com/mozilla/neqo/blob/c8499e8cb40e78df93696f22acf1a60554d87960/neqo-transport/src/cc/cubic.rs#L393-L400

As you can see it want's to change `cwnd` to something else than `ssthresh` under certain conditions, but we only have access to both here, not in `reduce_cwnd(...)` so it would need to be handed down.

Since QUIC recommends a minimum congestion window of `2*MSS` though I'm not sure this applies (See comments in `cubic.rs L500-501`) in which case we wouldn't need to change anything here because both `cwnd` and `ssthresh` would be `max(ssthresh, 2*MSS)` in any case.

neqo-transport/src/cc/cubic.rs
```diff
@@ -15,50 +15,74 @@ use neqo_common::qtrace;
...
 pub const CUBIC_FAST_CONVERGENCE: f64 = 0.85; // (1.0 + CUBIC_BETA) / 2.0;
 
 /// The minimum number of multiples of the datagram size that need
 /// to be received to cause an increase in the congestion window.
 /// When there is no loss, Cubic can return to exponential increase, but
 /// this value reduces the magnitude of the resulting growth by a constant factor.
 /// A value of 1.0 would mean a return to the rate used in slow start.
+///
+/// UPDATE: Not found in RFC. I don't exactly understand why we're doing this?
 const EXPONENTIAL_GROWTH_REDUCTION: f64 = 2.0;
```
mxinden: For the record, this was introduced here:

https://github.com/mozilla/neqo/pull/1074#discussion_r559255537

neqo-transport/src/cc/cubic.rs
```diff
@@ -146,18 +204,49 @@ impl Cubic {
...
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#figure-2>
+    ///
+    /// Taking into account that we're using bytes not MSS units, the formula becomes:
+    ///
+    /// `k = cubic_root((w_max - cwnd_epoch)/C/MSS)`
+    ///
+    /// ACTION: change `calc_k`, add `cwnd_epoch`
     fn calc_k(&self, curr_cwnd: f64, max_datagram_size: usize) -> f64 {
         ((self.w_max - curr_cwnd) / CUBIC_C / convert_to_f64(max_datagram_size)).cbrt()
```
mxinden: Do I understand correctly that you are suggesting to change `curr_cwnd` to `cwnd_epoch`, in other words to use the congestion window at the start of the epoch instead of the current congestion window?

From a first read of RFC 9438, that checks out. But I need to better understand why we did what we do today.

mxinden: Note that we are only calling `calc_k` in `start_epoch`. Thus, I assume, `curr_cwnd` is always equal to `cwnd_epoch`.

neqo-transport/src/cc/cubic.rs
```diff
@@ -232,15 +352,22 @@ impl WindowAdjustment for Cubic {
...
+        // > is in the Reno-friendly region and cwnd SHOULD be set to w_est at each reception of a new ACK.
         //
-        // > When receiving an ACK in congestion avoidance (cwnd could be
-        // > greater than or less than W_max), CUBIC checks whether W_cubic(t) is
-        // > less than W_est(t).  If so, CUBIC is in the TCP-friendly region and
-        // > cwnd SHOULD be set to W_est(t) at each reception of an ACK.
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
         //
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
+        // UPDATE: We should compare `w_cubic` (not `target`) and `w_est` here.
```
mxinden: > `        let target_cubic = self.w_cubic(time_ca, max_datagram_size);`

`target_cubic` should be `w_cubic`. Am I missing something?

neqo-transport/src/cc/cubic.rs
```diff
@@ -302,3 +480,35 @@ impl WindowAdjustment for Cubic {
...
+// > with exponential backoff, as described in [RFC3168].
+//
+// <https://datatracker.ietf.org/doc/html/rfc9438#section-4.6-7> VERSUS
+//
+// > QUIC therefore recommends that the minimum congestion window be two packets.
+//
+// <https://datatracker.ietf.org/doc/html/rfc9002#section-4.8>
+//
+// QUESTION: So this probably does not apply to CUBIC on QUIC and we never go below 2*MSS
+// (as is currently implemented)?
```
mxinden: > QUESTION: So this probably does not apply to CUBIC on QUIC and we never go below 2*MSS (as is currently implemented)?

In favor, i.e. keep the current minimum of 2*MSS. That said, let's keep the comment here a bit longer to give others a chance to voice their opinion as well.

neqo-transport/src/cc/cubic.rs
```diff
@@ -302,3 +480,35 @@ impl WindowAdjustment for Cubic {
...
+// > QUIC does not collapse the congestion window when the PTO expires, since a single
+// > packet loss at the tail does not indicate persistent congestion.
+//
+// <https://datatracker.ietf.org/doc/html/rfc9002#section-4.7>
+//
+// QUESTION: Timeout (RTO/PTO) does not apply to us then?
+//
+// 3. Spurious fast retransmits: <https://datatracker.ietf.org/doc/html/rfc9438#section-4.9.2>
+//
+// QUESTION: Do we want to implement this? If yes then we may be able to do it in a follow up PR.
```
mxinden: I am in favor of doing this in a follow-up pull request. I suggest tracking it in a GitHub issue.

neqo-transport/src/cc/cubic.rs
```diff
@@ -203,10 +292,29 @@ impl WindowAdjustment for Cubic {
...
 
-        // Cubic concave or convex region
+        // Calculate `target` for the concave or convex region
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-concave-region>
+        // <https://datatracker.ietf.org/doc/html/rfc9438#name-convex-region>
+        //
+        // UPDATE: New logic for how `target` is calculated.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1>
```
mxinden: Maybe link to Linux's 1.5 slow start limit?

https://github.com/torvalds/linux/blob/546b1c9e93c2bb8cf5ed24e0be1c86bb089b3253/net/ipv4/tcp_cubic.c#L318-L321

larseggert:
Also see https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/congestion/cubic.rs

I kinda wonder if it would make sense to have a standalone Rust crate implementing these CCAs for sharing between QUIC implementations...

neqo-transport/src/cc/cubic.rs
omansfeld: Yes, sorry the comment is poorly worded making it confusing.

What I am suggesting is exactly that, change `curr_cwnd` to `cwnd_epoch`. It's also correct that in our implementation they are always equal, I noted that somewhere further down in the `start_epoch()` function. I might rename the parameter so it better reflects the RFC.

We did what we do today because Dragana apparently already changed the original formula from RFC 8312 to look like the newer one (they are functionally identical). See original comment from Dragana further up.

Overall the change to K seems purely cosmetic to make the formula more clear without having to rely on things like the "in our implementation `curr_cwnd` and `cwnd_epoch` are always equal".

neqo-transport/src/cc/cubic.rs
omansfeld: The new RFC changes how `target` is calculated with upper and lower bounds, so it does not always equal `w_cubic` anymore, the comparison for the reno-friendliness still compares with `w_cubic` though. Thus the need for a change here.

neqo-transport/src/cc/cubic.rs
omansfeld: See #2694

neqo-transport/src/cc/cubic.rs
omansfeld: https://ntap.github.io/rfc8312bis/issues.html

also see reasoning about adding the bounds in the draft discussions (cannot properly link, search that page for "window increase")

omansfeld:
> Also see [cloudflare/quiche@`master`/quiche/src/recovery/congestion/cubic.rs](https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/congestion/cubic.rs?rgh-link-date=2025-06-03T10%3A36%3A41Z)
> 
> I kinda wonder if it would make sense to have a standalone Rust crate implementing these CCAs for sharing between QUIC implementations...

Thanks for the link! 

That'd be nice, though at this point I'd wonder if anybody would actually end up using it instead of iterating on theirs. I guess making a standalone crate fit into the existing implementations would also be a lot of work.

And more people implementing the CCAs means more people able to give feedback to the RFCs/drafts.

neqo-transport/src/cc/cubic.rs
omansfeld: If I read it correctly this is implementing the [`target = 1.5 * cwnd`](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1) maximum from the RFC.

https://github.com/mozilla/neqo/blob/1a42c5429731ece410b1836ed1bd47eac6394ad0/neqo-transport/src/cc/cubic.rs#L242-L246

Our code is following the [linux implementation](https://github.com/torvalds/linux/blob/546b1c9e93c2bb8cf5ed24e0be1c86bb089b3253/net/ipv4/tcp_cubic.c#L321) here (which I'm guessing is also where the new RFC got it's `1.5 * cwnd` maximum in the first place):

```c
	/* The maximum rate of cwnd increase CUBIC allows is 1 packet per
	 * 2 packets ACKed, meaning cwnd grows at 1.5x per RTT.
	 */
	ca->cnt = max(ca->cnt, 2U);
```

I'd suggest more closely following the RFC and putting that limit on `target` instead of indirectly limiting it through `acked_to_increase`. At least to me that's a lot easier to read and understand and should be functionally identical. It's also what e.g. [quiche](https://github.com/cloudflare/quiche/blob/a5311660df1b990a845757a5720d8506950fa30f/quiche/src/recovery/congestion/cubic.rs#L293) is doing:

```rust
        // target = w_cubic(t + rtt)
        let target = r
            .cubic_state
            .w_cubic(t + *rtt_stats.min_rtt, r.max_datagram_size);

        // Clipping target to [cwnd, 1.5 x cwnd]
        let target = f64::max(target, r.congestion_window as f64);
        let target = f64::min(target, r.congestion_window as f64 * 1.5);
```

mxinden: > If I read it correctly this is implementing the [`target = 1.5 * cwnd`](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1) maximum from the RFC.

Great catch!

> I'd suggest more closely following the RFC and putting that limit on `target` instead of indirectly limiting it through `acked_to_increase`.

In favor. Sounds good. I suggest adding a reference to both the RFC and the Linux implementation.

Instead of using `f64:max` and `f64::min`, how about using `f64::clamp`?

https://doc.rust-lang.org/std/primitive.f64.html#method.clamp

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -100,10 +100,15 @@ pub trait WindowAdjustment: Display + Debug {
     ) -> (usize, usize);
     /// Cubic needs this signal to reset its epoch.
     fn on_app_limited(&mut self);
+
+    // Those are only needed as helpers for CUBIC tests and are properly implemented in `cubic.rs`
```
```suggestion
    /// Those are only needed as helpers for CUBIC tests and are overridden in [`cubic::Cubic`].
```

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -342,16 +349,18 @@ fn congestion_event_congestion_avoidance_no_overflow() {
...
     ack_packet(&mut cubic, 1, now());
 
-    assert_within(cubic.last_max_cwnd(), cwnd_initial_f64 * 10.0, f64::EPSILON);
+    assert_within(cubic.w_max(), cwnd_initial_f64 * 10.0, f64::EPSILON);
     assert_eq!(cubic.cwnd(), cubic.cwnd_initial());
 
     // Now ack packet that was send earlier.
     ack_packet(&mut cubic, 0, now().checked_sub(PTO).unwrap());
+
+    // TODO: This test never asserts anything after the initial setup??
```
mxinden: Based on the test name (`congestion_event_congestion_avoidance_no_overflow`) I assume this was added as part of a patch fixing an overflow panic in `ack_packet`. Can you find something in the git history?

omansfeld: Good idea, I will take a look.

omansfeld: That's the case, it was committed together with the following code:

https://github.com/mozilla/neqo/blob/0bd3933b2ccfed140b903cc76e7a7f8c1daa3c56/neqo-transport/src/cc/cubic.rs#L199-L207

which coincidentally also answers why we were doing that 👍

neqo-transport/src/cc/cubic.rs
omansfeld: Just for understanding, that issue looks to me like we didn't falsely assume a congestion event on PTO but falsely assumed a lost packet leading to a congestion event, right? 

And PTO shouldn't trigger a congestion event, only persistent congestion does (but resets back to minimum window size, not a normal decrease). Am I reading that correctly?

If so then that should all be correctly implemented and doesn't need further action here.

neqo-transport/src/cc/cubic.rs
omansfeld: For the record e.g. quiche is doing the same thing (keeping the min cwnd of 2*MSS).

https://github.com/cloudflare/quiche/blob/4b720fcd783d5f820ef02ccf2e1ec91598ee1102/quiche/src/recovery/congestion/cubic.rs#L350-L354

```rust
        let ssthresh = (r.congestion_window as f64 * BETA_CUBIC) as usize;
        let ssthresh = 
            // MINIMUM_WINDOW_PACKETS = 2
            cmp::max(ssthresh, r.max_datagram_size * MINIMUM_WINDOW_PACKETS);
        r.ssthresh.update(ssthresh, r.hystart.in_css());
        r.congestion_window = ssthresh;
```

Are you okay with removing the comment now?

neqo-transport/src/cc/cubic.rs
omansfeld: The RFC is fine with using bytes anyways:

> Implementations can use bytes to express window sizes, which would require factoring in the SMSS wherever necessary and replacing segments_acked (Figure 4) with the number of acknowledged bytes.

https://datatracker.ietf.org/doc/html/rfc9438#section-4.1-1

Marking as resolved, I don't think there is any reason for us to switch this.

neqo-transport/src/cc/cubic.rs
mxinden: > but falsely assumed a lost packet leading to a congestion event, right?

That is correct.

> And PTO shouldn't trigger a congestion event, only persistent congestion does

Yes

> Similar to the RACK-TLP loss detection algorithm for TCP [[RFC8985](https://www.rfc-editor.org/rfc/rfc9002.html#RFC8985)], QUIC does not collapse the congestion window when the PTO expires, since a single packet loss at the tail does not indicate persistent congestion. Instead, QUIC collapses the congestion window when persistent congestion is declared; see [Section 7.6](https://www.rfc-editor.org/rfc/rfc9002.html#persistent-congestion).

https://www.rfc-editor.org/rfc/rfc9002.html#section-4.7

> (but resets back to minimum window size, not a normal decrease)

Yes

> When persistent congestion is declared, the sender's congestion window MUST be reduced to the minimum congestion window (kMinimumWindow), similar to a TCP sender's response on an RTO [[RFC5681](https://www.rfc-editor.org/rfc/rfc9002.html#RFC5681)].

https://www.rfc-editor.org/rfc/rfc9002.html#section-7.6.2

> If so then that should all be correctly implemented and doesn't need further action here.

My understanding as well. Thanks for diving into it.

neqo-transport/src/cc/cubic.rs
mxinden: :+1: fine removing the comment. Though I do think it is worth noting it somewhere in `cubic.rs` that we deviate from the Cubic RFC (1MSS) in favor of the QUIC RFC (2MSS).

neqo-transport/src/cc/cubic.rs
omansfeld: didn't end up linking to linux's 1.5 limit because this is now part of the upper/lower bound for `target` as per RFC, so I linked to that at the appropriate place.

neqo-transport/src/cc/cubic.rs
```diff
@@ -252,12 +262,16 @@ impl WindowAdjustment for Cubic {
         now: Instant,
     ) -> usize {
         let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
-        let new_acked_f64 = convert_to_f64(new_acked_bytes);
-        if self.t_epoch.is_none() {
-            // This is a start of a new congestion avoidance phase.
-            self.start_epoch(curr_cwnd_f64, new_acked_f64, max_datagram_size, now);
+        self.bytes_acked = convert_to_f64(new_acked_bytes);
```
omansfeld: We used to set `bytes_acked` in multiple places.

1. setting it `= new_acked_f64` in `start_epoch()`
2. setting it `+= new_acked_f64` here when not entering `start_epoch()`
3. setting it `-= incr * tcp_cnt` after having calculated the reno friendly increase, which if you do the math equals `-= bytes_acked`, so effectively setting it to zero.

https://github.com/mozilla/neqo/blob/0f3262121c1c390fd70cb24d97c63660ada90fb8/neqo-transport/src/cc/cubic.rs#L214-L220

Thus it was always `0` when entering `bytes_for_cwnd_increase()`, and we then set it to `new_acked_f64` (either by adding it to `0` or assigning it).

This streamlines that logic by just setting it in one place.

neqo-transport/src/cc/cubic.rs
```diff
@@ -252,12 +262,16 @@ impl WindowAdjustment for Cubic {
...
+        let t_epoch;
+        if let Some(t) = self.t_epoch {
```
omansfeld: Safely unwrapping `self.t_epoch` here, because we need it as an `Instant` type when calculating the time difference between `t_epoch` and `t_current` for the cubic function with `now.saturating_duration_since(t_epoch)`.

neqo-transport/src/cc/cubic.rs
```diff
@@ -262,6 +253,7 @@ impl WindowAdjustment for Cubic {
         now: Instant,
     ) -> usize {
         let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
+        let max_datagram_size_f64 = convert_to_f64(max_datagram_size);
```
omansfeld, line 282: we only use `max_datagram_size` as a `f64` value in this function, so now it converts once and then uses that value instead of converting at each use.

neqo-transport/src/cc/cubic.rs
```diff
@@ -346,70 +319,81 @@ impl WindowAdjustment for Cubic {
...
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.4-1>
+        //
+        // In neqo target_cwnd is in bytes.
+        let target_cwnd = if self.w_cubic(t.as_secs_f64(), max_datagram_size_f64) < self.w_est {
+            // Reno-friendly region sets cwnd = w_est
+            self.w_est
+        } else {
+            // Concave or convex region increments cwnd by ((target - cwnd) / cwnd) segments
+            ((target - curr_cwnd_f64) / curr_cwnd_f64).mul_add(max_datagram_size_f64, curr_cwnd_f64)
```
omansfeld: We have to multiply with `max_datagram_size_f64` again here, because the ratio we build with `((target - cwnd) / cwnd)` returns an increase in segments, but we want bytes.

neqo-transport/src/cc/cubic.rs
```diff
@@ -346,70 +319,81 @@ impl WindowAdjustment for Cubic {
...
-        // Limit increase to max 1 MSS per EXPONENTIAL_GROWTH_REDUCTION ack packets.
-        // This effectively limits target_cwnd to (1 + 1 / EXPONENTIAL_GROWTH_REDUCTION) cwnd.
-        acked_to_increase = acked_to_increase.max(EXPONENTIAL_GROWTH_REDUCTION * max_datagram_size);
-        acked_to_increase as usize
+        // other options:
+        // 1. get rid of the acked_to_increase logic and just do something akin to
+        //    `self.congestion_window = target_cwnd` OR
+        // 2. if we are actually bound to only increase in SMSS blocks then do `if cwnd_increase >=
+        //    max_datagram_size` and only increase if it is. track cwnd_increase across calls of
+        //    this function.
```
omansfeld: To me the way we set the updated congestion window seems a bit cumbersome by looking at how many bytes we need to ack for the increase of 1*SMSS to match the calculated increase of `cwnd_increase`. See code comment above.

Maybe it'd make sense to change that. For me the main question would be if we are actually bound to only increase in SMSS blocks for some technical reason, or if that's purely to mirror the RFC working with segments?

Anyways I'm aware that that change would probably not make any practical difference, but it would make the code much easier to understand imo, and thus also less prone to mistakes.

neqo-transport/src/cc/cubic.rs
martinthomson: The way to do this is instead:

```rust
let t_epoch = self.t_epoch.unwrap_or_else(|| {
    // Put your comment here.
    self.start_epoch(...);
    now
});
```

neqo-transport/src/cc/cubic.rs
mxinden: > if we are actually bound to only increase in SMSS blocks for some technical reason, or if that's purely to mirror the RFC working with segments?

I am not aware of a reason, other than to mirror the RFC.

> Anyways I'm aware that that change would probably not make any practical difference, but it would make the code much easier to understand imo, and thus also less prone to mistakes.

I believe it would improve readability of this code significantly, in other words, I am in favor. That said, to reduce complexity of this pull request, I suggest doing this either before or after in a separate pull request.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,206 @@ impl WindowAdjustment for Cubic {
...
-        let incr = (self.tcp_acked_bytes / tcp_cnt).floor();
-        if incr > 0.0 {
-            self.tcp_acked_bytes -= incr * tcp_cnt;
-            self.estimated_tcp_cwnd += incr * max_datagram_size;
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-11>
+        //
+        // TODO: Maybe add a test for this correctly working (and resetting back to default after
+        // congestion event)
+        if self.w_est >= self.cwnd_prior {
+            self.alpha = 1.0;
```
omansfeld, line 339: for reference, also see RFC discussion here: https://ntap.github.io/rfc8312bis/issues.html#n(2)

neqo-transport/src/cc/cubic.rs
omansfeld: Thanks for the suggestion, that is much more readable than what I had with `if let Some(t)...`

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
+        // > 1 to ensure that it can achieve the same congestion window increment rate
+        // > as Reno, which uses AIMD(1, 0.5).
         //
-        //  <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-        let max_datagram_size = convert_to_f64(max_datagram_size);
-        let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
-        let incr = (self.tcp_acked_bytes / tcp_cnt).floor();
-        if incr > 0.0 {
-            self.tcp_acked_bytes -= incr * tcp_cnt;
-            self.estimated_tcp_cwnd += incr * max_datagram_size;
```
omansfeld, line 219: This old w_est calculation is functionally identical to the new formula:

`w_est += incr * mss = mss* acked_bytes / tcp_cnt = mss * acked_bytes / (w_est / alpha) = mss * acked_bytes * alpha / w_est`

which is the same as the new increase:

`self.w_est += self.alpha * (self.new_bytes_acked / curr_cwnd_f64) * max_datagram_size_f64;`

The only difference is that the new RFC uses `cwnd` as a divisor instead of `w_est`. While in the Reno region this should be basically the same thing, as `cwnd` is equal to `w_est` at the last increase. It also makes sense that we'd want to compare to the actual current `cwnd`, instead of the old `w_est` value, which maybe was never applied.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
-        // > cwnd SHOULD be set to W_est(t) at each reception of an ACK.
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
         //
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-        let target_cwnd = target_cubic.max(self.estimated_tcp_cwnd);
+        // While the RFC specifies that we should compare W_cubic(t) with w_est we are rather
+        // comparing the previously calculated target here, since that is the value that would
+        // actually be used when deciding it's the cubic region.
+        //
+        // That is in line with what e.g. the Linux Kernel CUBIC implementation is doing.
```
omansfeld, line 355: For reference: https://github.com/torvalds/linux/blob/d7ee5bdce7892643409dea7266c34977e651b479/net/ipv4/tcp_cubic.c#L310-L315

```c
if (ca->tcp_cwnd > cwnd) {	/* if bic is slower than tcp */
	delta = ca->tcp_cwnd - cwnd;
	max_cnt = cwnd / delta;
	if (ca->cnt > max_cnt) {
		ca->cnt = max_cnt;
	}
```

The `if (ca->cnt > max_cnt)` line is of interest. `ca->cnt` is the cubic value, `max_cnt` is the reno value.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
+        // (https://datatracker.ietf.org/doc/html/rfc9438#section-4.4-2.1)
+        //
+        // Since we want to know how much we need to ack to increase by 1 segment we need the
+        // inverse of that, which would be `cwnd/cwnd_increase`.
+        // (E.g. if we'd increase by `1/4 * mss` per cwnd acked then we need to ack `4 * cwnd` to
+        // increase by `1 * mss`)
+        //
+        // The RFC only applies this increase per acked cwnd to the Cubic (concave/convex) region.
+        // We also apply it to the Reno region, as that is what the Linux Kernel CUBIC
+        // implementation does, too.
```
omansfeld, line 380: For reference, Linux implementation: 

Reno region: https://github.com/torvalds/linux/blob/d7ee5bdce7892643409dea7266c34977e651b479/net/ipv4/tcp_cubic.c#L311-L312
```c
delta = ca->tcp_cwnd - cwnd;
max_cnt = cwnd / delta;
```

Cubic region: https://github.com/torvalds/linux/blob/d7ee5bdce7892643409dea7266c34977e651b479/net/ipv4/tcp_cubic.c#L287
```c
ca->cnt = cwnd / (bic_target - cwnd);
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
+        }
 
-        // Calculate the number of bytes that would need to be acknowledged for an increase
-        // of `max_datagram_size` to match the increase of `target - cwnd / cwnd` as defined
-        // in the specification (Sections 4.4 and 4.5).
-        // The amount of data required therefore reduces asymptotically as the target increases.
-        // If the target is not significantly higher than the congestion window, require a very
-        // large amount of acknowledged data (effectively block increases).
-        let mut acked_to_increase =
-            max_datagram_size * curr_cwnd_f64 / (target_cwnd - curr_cwnd_f64).max(1.0);
```
omansfeld, line 240: This made sure that if we had an increase of less than `1.0` we'd return `mss * curr_cwnd` ("large amount of data").

I don't think it's necessary, as if the increase is smaller than `1.0` we'd return a value even bigger than `mss * curr_cwnd`, so I didn't implement that check in the new version.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
+        let cwnd_increase = target - curr_cwnd_f64;
 
-        // Limit increase to max 1 MSS per EXPONENTIAL_GROWTH_REDUCTION ack packets.
-        // This effectively limits target_cwnd to (1 + 1 / EXPONENTIAL_GROWTH_REDUCTION) cwnd.
-        acked_to_increase = acked_to_increase.max(EXPONENTIAL_GROWTH_REDUCTION * max_datagram_size);
```
omansfeld, line 244: This was functional identical to the `target < cwnd * 1.5` check that we now have when calculating the cubic value.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
-        acked_to_increase as usize
```
omansfeld, line 245: This was functional identical to the `target < cwnd * 1.5` check that we now have when calculating the cubic value.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -150,59 +149,22 @@ fn tcp_phase() {
...
-    // number of ack needed is t_to_increase / time_increase.
-    let expected_ack_cubic_increase =
-        (((CUBIC_C.mul_add((elapsed_time).as_secs_f64().powi(3), 1.0) / CUBIC_C).cbrt()
-            - elapsed_time.as_secs_f64())
-            / time_increase.as_secs_f64())
-        .ceil() as u64;
-    // num_acks is very close to the calculated value. The exact value is hard to calculate
-    // because the proportional increase (i.e. curr_cwnd_f64 / (target - curr_cwnd_f64) *
-    // MAX_DATAGRAM_SIZE_F64) and the byte counting.
-    assert_eq!(num_acks2, expected_ack_cubic_increase + 2);
```
omansfeld, line 196: This complicated check wouldn't pass with the new code.

As we added the logic where `alpha` is sometimes `1.0` the cwnd was now already a bit higher here than it was in the original code, so we'd also need more acks for the next increase. This formula would not increase the acks needed in a similar fashion.

From the code-comment above I'll guess that the formula wasn't really foolproof before either.

So I suggest removing this second round of cubic function increases, especially since the test is supposed to test the tcp phase anyways, which it then still does (including testing that the tcp phase is exited).

neqo-transport/src/connection/tests/cc.rs
```diff
@@ -244,7 +244,7 @@ fn cc_cong_avoidance_recovery_period_to_cong_avoidance(cc_algorithm: CongestionC
     assert!(cwnd_before_loss > cwnd_after_loss);
     qinfo!("moving to congestion avoidance {}", cwnd(&client));
 
-    for i in 0..6 {
+    for i in 0..7 {
```
omansfeld, line 249: Had to increase by 1 iteration to make the test pass. At first this seems like we grow cwnd slower with the new patch, but it's actually another reason:

In `on_packets_acked(...)` we sometimes increase cwnd 2x per ack.

https://github.com/mozilla/neqo/blob/ea3f7a521f9971cb43ceb099d81b802d96ae939b/neqo-transport/src/cc/classic_cc.rs#L256-L264

This is contrary to what RFC 9002 allows:
> A sender in congestion avoidance uses an Additive Increase Multiplicative Decrease (AIMD) approach that MUST limit the increase to the congestion window to at most one maximum datagram size for each congestion window that is acknowledged.

(https://datatracker.ietf.org/doc/html/rfc9002#section-7.3.3-2)

That double increase leads to the below test output for both main and cubic_9438. You'll see that in iteration 6 on main the cwnd gets increased twice from `11088` to `13552` while on cubic_9438 it is increased once in iteration 5 going from `11088` to `12320`, which leads to it not increasing in iteration 6 (the current cwnd is part of the calculations for cubic). In iteration 7 its unchanged on main while double increasing to `14784` on cubic_9438, so the patch is actually faster, just lacks behind on iteration 6 due to the double increase.

Main:
```
0.032 INFO client sending 8624 bytes into cwnd of 8624
0.032 INFO iteration 1
0.033 INFO client sending 8624 bytes into cwnd of 8624
0.033 INFO iteration 2
0.034 INFO client sending 11088 bytes into cwnd of 11088
0.034 INFO iteration 3
0.035 INFO client sending 11088 bytes into cwnd of 11088
0.035 INFO iteration 4
0.036 INFO client sending 11088 bytes into cwnd of 11088
0.036 INFO iteration 5
0.037 INFO client sending 11088 bytes into cwnd of 11088
0.037 INFO iteration 6
0.038 INFO client sending 13552 bytes into cwnd of 13552
0.038 INFO cwnd_before_loss = 12320 cwnd(&client) = 13552
```

CUBIC_9438:
```
0.033 INFO iteration 0
0.033 INFO client sending 8624 bytes into cwnd of 8624
0.034 INFO iteration 1
0.034 INFO client sending 8624 bytes into cwnd of 8624
0.035 INFO iteration 2
0.035 INFO client sending 11088 bytes into cwnd of 11088
0.036 INFO iteration 3
0.036 INFO client sending 11088 bytes into cwnd of 11088
0.037 INFO iteration 4
0.037 INFO client sending 11088 bytes into cwnd of 11088
0.038 INFO iteration 5
0.038 INFO client sending 12320 bytes into cwnd of 12320
0.039 INFO iteration 6
0.039 INFO client sending 12320 bytes into cwnd of 12320
0.040 INFO cwnd_before_loss = 12320 cwnd(&client) = 14784
```

I will open an issue to look into the double increase (and happy to hear opinions if it is actually a problem that we go against RFC 9002 there).

larseggert: Not so much an issue as unexpected - this smells like a bug. Great work tracking it down!

omansfeld: For reference: https://github.com/mozilla/neqo/issues/2875

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -100,10 +100,22 @@ pub trait WindowAdjustment: Display + Debug {
     ) -> (usize, usize);
     /// Cubic needs this signal to reset its epoch.
     fn on_app_limited(&mut self);
+
+    /// Helper for CUBIC tests and overridden in [`cubic::Cubic`]
     #[cfg(test)]
-    fn last_max_cwnd(&self) -> f64;
+    fn w_max(&self) -> f64 {
+        0.0
+    }
```
```suggestion
    fn w_max(&self) -> f64 {
        unreachable!()
    }
```

`0.0` is a dummy value and not a valid value, right? I.e. there is no scenario where one would want to use the trait's default implementation, correct? If so, I suggest panicking instead.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -100,10 +100,22 @@ pub trait WindowAdjustment: Display + Debug {
...
+
+    /// Helper for CUBIC tests and overridden in [`cubic::Cubic`]
+    #[cfg(test)]
+    fn set_w_max(&mut self, _w_max: f64) {}
```
mxinden: Same here. Panicking will ensure wrong usage won't go unnoticed.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -100,10 +100,22 @@ pub trait WindowAdjustment: Display + Debug {
...
+
+    /// Helper for CUBIC tests and overridden in [`cubic::Cubic`]
     #[cfg(test)]
-    fn set_last_max_cwnd(&mut self, last_max_cwnd: f64);
+    fn alpha(&self) -> f64 {
+        0.0
+    }
```
mxinden: Same here.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -552,14 +577,14 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
+        let (ssthresh, acked_bytes) = self.cc_algorithm.reduce_cwnd(
             self.congestion_window,
             self.acked_bytes,
             self.max_datagram_size(),
         );
-        self.congestion_window = max(cwnd, self.cwnd_min());
+        self.ssthresh = max(ssthresh, self.cwnd_min());
         self.acked_bytes = acked_bytes;
-        self.ssthresh = self.congestion_window;
+        self.congestion_window = self.ssthresh;
```
mxinden: I know the two implementations are effectively the same. Why this change?

Doc comments on `reduce_cwnd`:

```
    /// This function is called when a congestion event has been detected and it
    /// returns new (decreased) values of `curr_cwnd` and `acked_bytes`.
    /// This value can be very small; the calling code is responsible for ensuring that the
    /// congestion window doesn't drop below the minimum of `CWND_MIN`.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -15,46 +15,72 @@ use neqo_common::qtrace;
...
+///
+/// Formula:
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7);
+/// `CUBIC_ALPHA = 3.0 * (1.0 - CUBIC_BETA) / (1.0 + CUBIC_BETA)`
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+///
+/// This constant is used to set the default value of `neqo_transport::cc::cubic::Cubic::alpha`.
```
```suggestion
/// This constant is used to set the default value of [`neqo_transport::cc::cubic::Cubic::alpha`].
```

Do Rust docs to `struct` fields work? Would make sure this doc comment stays up-to-date.

neqo-transport/src/cc/cubic.rs
```diff
@@ -15,46 +15,72 @@ use neqo_common::qtrace;
...
+/// > To balance between the scalability and convergence speed, CUBIC sets the multiplicative window
+/// > decrease factor to 0.7 while Standard TCP uses 0.5. While this improves the scalability of
+/// > CUBIC, a side effect of this decision is slower convergence, especially under low statistical
 /// > multiplexing environments.
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-3>
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-principle-4-for-the-cubic-d>
+///
+/// For implementation reasons neqo uses a dividend and divisor approach with `usize` typing to
+/// construct `CUBIC_BETA = 0.7`
```
```suggestion
/// construct `CUBIC_BETA = 0.7`.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -15,46 +15,72 @@ use neqo_common::qtrace;
...
+/// Formula:
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7);
+/// `CUBIC_ALPHA = 3.0 * (1.0 - CUBIC_BETA) / (1.0 + CUBIC_BETA)`
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9438#name-reno-friendly-region>
+///
+/// This constant is used to set the default value of `neqo_transport::cc::cubic::Cubic::alpha`.
+pub const CUBIC_ALPHA: f64 = 3.0 * (1.0 - 0.7) / (1.0 + 0.7); // with CUBIC_BETA = 0.7
```
mxinden, line 45: I interpret the `Missed mutant` warnings below as us basically having no test coverage of `CUBIC_ALPHA`.

Can you think of an easy way to increase that coverage? Ideally in a separate pull request that we can merge first, thus ensure that the new implementation matches the old.

neqo-transport/src/cc/cubic.rs
```diff
@@ -70,99 +96,144 @@ pub fn convert_to_f64(v: usize) -> f64 {
...
+        self.alpha = CUBIC_ALPHA;
+        self.t_epoch = Some(now);
+        self.w_est = cwnd_epoch;
+        // If `w_max < cwnd_epoch` we take the cubic root from a negative value in `calc_k()`. That
+        // could only happen if somehow `cwnd` get's increased between calling `reduce_cwnd()` and
+        // `start_epoch()`, which is only possible after persistent congestion. It could also happen
+        // if we never had a congestion event, so never called `reduce_cwnd()` thus `w_max` was
+        // never set (so is still it's default `0.0` value). In any case we reset/initialize
+        // `w_max`, `cwnd_prior` and `k` here. We also set `alpha` to `1.0`, since
+        // `w_est >= cwnd_prior` is true here.
```
mxinden: Why do we know that `w_est >= cwnd_prior`?

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
         let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
-        let new_acked_f64 = convert_to_f64(new_acked_bytes);
-        if self.ca_epoch_start.is_none() {
-            // This is a start of a new congestion avoidance phase.
-            self.start_epoch(curr_cwnd_f64, new_acked_f64, max_datagram_size, now);
-        } else {
-            self.tcp_acked_bytes += new_acked_f64;
-        }
+        let max_datagram_size_f64 = convert_to_f64(max_datagram_size);
+        self.new_bytes_acked = convert_to_f64(new_acked_bytes);
```
```suggestion
        self.new_bytes_acked += convert_to_f64(new_acked_bytes);
```

The previous implementation did `self.tcp_acked_bytes += new_acked_f64;`. Wouldn't the above be necessary to maintain the status quo?

`new_bytes_acked` docs:

``` rust
    /// Number of bytes acked since the last congestion window increase.
```

I guess this is not quite right, as a call to `bytes_for_cwnd_increase` does not always lead to a congestion window increase.

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
-        // > greater than or less than W_max), CUBIC checks whether W_cubic(t) is
-        // > less than W_est(t).  If so, CUBIC is in the TCP-friendly region and
-        // > cwnd SHOULD be set to W_est(t) at each reception of an ACK.
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-8>
         //
-        // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.2>
-        let target_cwnd = target_cubic.max(self.estimated_tcp_cwnd);
+        // While the RFC specifies that we should compare W_cubic(t) with w_est we are rather
+        // comparing the previously calculated target here, since that is the value that would
+        // actually be used when deciding it's the cubic region.
```
```suggestion
        // actually be used when deciding whether it's the cubic region.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
+    // "similar mechanisms" refering to taking "measures to prevent cwnd from growing when the
+    // volume of bytes in flight is smaller than cwnd".
+    //
+    // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.6-4>
+    //
+    // TODO: Do we already have those mechanisms somewhere?
+    //
+    // This function only returns the value for `cwnd * CUBIC_BETA` and sets some variables for the
+    // start of a new congestion avoidance phase. Actually setting the congestion window happens in
+    // `ClassicCongestionControl::on_congestion_event` where this function is called.
```
```suggestion
    // [`super::ClassicCongestionControl::on_congestion_event`] where this function is called.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -182,110 +253,195 @@ impl WindowAdjustment for Cubic {
...
-        } else {
-            self.tcp_acked_bytes += new_acked_f64;
-        }
+        let max_datagram_size_f64 = convert_to_f64(max_datagram_size);
+        self.new_bytes_acked = convert_to_f64(new_acked_bytes);
+
+        let t_epoch = self.t_epoch.unwrap_or_else(|| {
+            // If we get here with `self.t_epoch == None` this is a new congestion evoidance state.
+            // It's been either set to `None` in `cubic::reduce_cwnd()` or needs to be
+            // initialized after slow start.
```
mxinden: Could as well be due to being application limited via `app_limited` below?

neqo-transport/src/cc/classic_cc.rs
omansfeld: I changed this to match the logic in the RFC, but yes, it's effectively the same.

https://datatracker.ietf.org/doc/html/rfc9438#section-4.6-5.1.1

Also see those prior discussions for some context:

https://github.com/mozilla/neqo/pull/2535#discussion_r2024536512
https://github.com/mozilla/neqo/pull/2535#discussion_r2120797395

I don't have strong opinions on this change, but I figure since it's fundamentally the same it doesn't hurt to follow RFC wording. If we keep it I'll change the doc comments.

neqo-transport/src/cc/cubic.rs
omansfeld: I wanted to use rust docs but it didn't properly work for me. It always opens a link to docs.rs where it then says "The requested crate does not exist".

neqo-transport/src/cc/cubic.rs
omansfeld: In line 221 we reset `w_est = cwnd_epoch` in `start_epoch()` as [per the RFC](https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-9).

In line 230 we also initialize `cwnd_prior = cwnd_epoch` for the case where `w_max < cwnd_epoch`.

That means we always have `w_est == cwnd_prior` here. So we set `alpha = 1.0` as the `>=` condition is fulfilled.

I will add a cross-reference to the rfc section that discusses increasing `alpha` to the comment so it's easier to understand in the future.

neqo-transport/src/cc/cubic.rs
omansfeld: True, I have actually missed that so far. I'll need to take a look into what exactly that means and why we are doing that after being app limited.

neqo-transport/src/cc/cubic.rs
omansfeld: See https://github.com/mozilla/neqo/pull/2535#discussion_r2172011406 that explains why I changed it to `=` instead of `+=` (TLDR: It is always `0` when we enter this codepath, so it doesn't make a difference).

~~Good catch on the docs, I'll rephrase it.~~

Actually this made me question if the status quo is correct (or maybe if I made a mistake in my reasoning about the status quo above). We use `new_bytes_acked` in the calculation for the reno friendly region. If we reset it between calls to `bytes_for_cwnd_increase` that means the calculation is off in the case where we didn't increase the congestion window (Thus not using all the bytes acked since the last increase in our calculation). 

I think we should only reset `new_bytes_acked` if `bytes_for_cwnd_increase` actually resulted in an increase of the congestion window. Otherwise they should be summed up. Then the code would match the doc comment, too. But I'm pretty sure it didn't match it before either.

I need to think about it some more, but do let me know your thoughts.

neqo-transport/src/cc/cubic.rs
omansfeld: I'm not sure what warnings you mean. All I see is the failing `codecov` check, but I don't see any `Missed mutant` warnings in there.

What should we be testing about `CUBIC_ALPHA`? The only thing that comes to mind is the change from `CUBIC_ALPHA` to `1.0` (and back) as per [the RFC](https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-11).

`alpha` is the actual value that is used in calculations. `CUBIC_ALPHA` is just the predefined constant that we use to reset `alpha` after the `1.0` increase. Previously we used the constant directly in calculations, but that changed because the RFC section linked above is new, so `alpha` is not constant anymore.

Thus the new implementation does NOT match the old in that regard, as that RFC section is new.

neqo-transport/src/cc/cubic.rs
mxinden: Yes, we don't publish to crates.io and thereby Neqo docs are not published to docs.rs.

It still ensures that these references are up-to-date:

https://github.com/mozilla/neqo/blob/bf80acf58b71f4e39c424c42367469958ef49fb8/.github/workflows/clippy.yml#L54

neqo-transport/src/cc/cubic.rs
mxinden: Do you see any of these?

image

mxinden: > What should we be testing about CUBIC_ALPHA? The only thing that comes to mind is the change from CUBIC_ALPHA to 1.0 (and back) as per [the RFC](https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-11).

Yes, that is a good idea.

> Thus the new implementation does NOT match the old in that regard, as that RFC section is new.

Fine if the test does not succeed on the old implementation.

I guess this is a more general problem. The fact that `cargo mutant` can change a central component, e.g. here `CUBIC_ALPHA` in such drastic ways, without a test failure, hints at low test coverage. I assume basic tests already get us a long way.

neqo-transport/src/cc/cubic.rs
mxinden: Can you add `debug_assert!`s for the assumptions that you are describing in this comment? Will ensure we uphold them in the future.

neqo-transport/src/cc/cubic.rs
mxinden: Why do we no longer set `new_bytes_acked` in `start_epoch` like we did before?

https://github.com/mozilla/neqo/blob/9bc3472e2f672a7ab38440356bbc34d9047fac27/neqo-transport/src/cc/cubic.rs#L148-L158

omansfeld: I took another look at the way we tracked `self.tcp_acked_bytes` vs how we should be doing it now and I hope the following comment clears stuff up. Thanks for challenging me about it, because it did lead to some deeper understanding about the whole setup (even though what I had before seems to have been correct).


- **Baseline:**
  - we need to make sure that all acked bytes are accounted for in `w_est`/`estimated_tcp_cwnd` either by tracking unused increases or by tracking acked bytes that weren't used or by both methods

- **our status quo for each call of `bytes_for_cwnd_increase`:**

https://github.com/mozilla/neqo/blob/820b658ff3bc77d5608b8acba80a125386ab5450/neqo-transport/src/cc/cubic.rs#L157

  - we set `self.tcp_acked_bytes = new_acked_f64` if we are calling `cubic::start_epoch`

https://github.com/mozilla/neqo/blob/820b658ff3bc77d5608b8acba80a125386ab5450/neqo-transport/src/cc/cubic.rs#L185-L191

  - we increment `self.tcp_acked_bytes += new_acked_f64` if we are not calling `cubic::start_epoch` (line 190)

https://github.com/mozilla/neqo/blob/820b658ff3bc77d5608b8acba80a125386ab5450/neqo-transport/src/cc/cubic.rs#L211-L220

  - we calculate the increment for `self.estimated_tcp_cwnd` in `incr` (line 216)
  - if the increment is atleast `1*mss` in size (that's the `if incr > 0.0` line combined with `.floor`'ing the `incr` value) we add it to `self.estimated_tcp_cwnd` and reduce `self.tcp_acked_bytes` by the used bytes (line 217-220)
  - we compare reno and cubic and if we have accumulated enough acks we increase by `1*mss` in `classic_cc::on_packets_acked` (where `bytes_for_cwnd_increase` was called)
  - the leftover bytes are stored in `self.tcp_acked_bytes` while the increase is stored in `self.estimated_tcp_cwnd`

---

- **new suggestion for each call of `bytes_for_cwnd_increase`:**
  - we calculate the increase to `self.w_est` with the passed in `new_acked_bytes` value and get rid of the struct field (equal to resetting the struct field with each call, thus functionally equal to what the code does in the current commit)
  - we compare reno and cubic and if we have accumulated enough acks we increase by `1*mss` in `classic_cc::on_packets_acked` (where `bytes_for_cwnd_increase` was called)
  - if we don't increase cwnd the acked bytes from the last function call are "stored" inside the increased `self.w_est`, thus we don't need to store them in `self.new_bytes_acked`

--- 

This would ensure all acked bytes are applied to `w_est` while getting rid of a whole lot of complexity and making the code a lot easier to read imo.

neqo-transport/src/cc/cubic.rs
omansfeld: This reasoning wasn't completely correct, see https://github.com/mozilla/neqo/pull/2535#discussion_r2315913966

neqo-transport/src/cc/cubic.rs
omansfeld: Leaving the current logic for now to reduce scope. See https://github.com/mozilla/neqo/issues/2875#issuecomment-3245127162 for future tracking.

neqo-transport/src/cc/cubic.rs
omansfeld: I found the warnings now, was looking in the wrong place. I agree that it's a problem and I will try to increase test coverage. As I've said offline, I think currently we never use the `self.alpha == CUBIC_ALPHA` in our tests, just either `self.alpha == 1.0` or the result of the cubic function wins (alpha is only used in the reno region).

neqo-transport/src/cc/cubic.rs
mxinden: Thanks. Above sounds reasonable to me. Can you push a commit with the change to see the changes required?

omansfeld: I commited a change that removes the `self.new_bytes_acked` struct field and uses the passed down `new_acked_bytes` value directly. On top of the previous changes already commited the current PR should now be in the state outlined in "new suggestion ..." above.

neqo-transport/src/cc/classic_cc.rs
omansfeld: On second thought a function `reduce_ssthresh` doesn't make much sense being called on a congestion event. And a function `reduce_cwnd` returning into a `ssthresh` value is weird, too.

So I reverted the change to match the function name and docs.

neqo-transport/src/cc/cubic.rs
omansfeld: As per [RFC section 5.8](https://datatracker.ietf.org/doc/html/rfc9438#app-limited) we shouldn't be counting the time we spend being app limited towards the `t` value in the cubic function. We use `self.t_epoch = None` to reset the timing, which seems correct. I added some references.

omansfeld:
I've added a test for the dynamic changing of the`alpha` value.

There are some CI failures currently, most of them due to other causes I think (Sanitize is addressed in #2954, the netbsd failures seem to be due to their cdn being down).

For me the transfer bench failure (https://github.com/mozilla/neqo/pull/2535#issuecomment-3265896667) seems to be most concerning. We see a regression only on `pacing-false` + `same-seed`. Could it be that the regression is just due to seed-luck? `varying-seeds` passes for the same benches. I need some input here though, as I don't know the exact intricacies of the simulator regarding seeding.

@mxinden could you take another look? I'm hoping we can push CUBIC along this week. Let me know if I can help in any way with making review easier or can do anything to further de-risk. We can also go over changes in person/a call this week if it helps.

larseggert:
I think this PR is becoming unwieldy - it touches too many things. Suggest to split out smaller pieces that can be merged now, and maybe using a tracking issue as an overview?

omansfeld:
> I think this PR is becoming unwieldy - it touches too many things. Suggest to split out smaller pieces that can be merged now, and maybe using a tracking issue as an overview?

@larseggert I agree this is becoming unwieldy, splitting is an excellent idea, thanks. I'll go do that now.

PR #2536, opened 2025-03-28
ci: Schedule benches on CPUs 2 and 3

larseggert:
Because the `transfer` bench creates two threads, and forcing both onto a single (isolated) CPU is hurting performance. As one might have expected.


mxinden:
> thrpt:  [139.11 MiB/s 139.98 MiB/s 140.86 MiB/s]

:tada:  better than before https://github.com/mozilla/neqo/pull/2455.

PR #2539, opened 2025-03-31
chore: Turn on some `smallvec` optimizations

larseggert:


neqo-transport/Cargo.toml
```diff
@@ -25,7 +25,7 @@ neqo-common = { path = "../neqo-common" }
 neqo-crypto = { path = "../neqo-crypto" }
 mtu = { version = "0.2", default-features = false } # neqo is only user currently, can bump freely
 qlog = { workspace = true }
-smallvec = { version = "1.13", default-features = false }
+smallvec = { version = "1.13", default-features = false, features = ["union", "const_generics", "const_new"] }
```
mxinden, line 28: Do we use `const_new` anywhere?

> This feature exposes the functions [SmallVec::new_const](https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html#method.new_const), [SmallVec::from_const](https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html#method.from_const), and [smallvec_inline](https://docs.rs/smallvec/latest/smallvec/macro.smallvec_inline.html) which enables the SmallVec to be initialized from a const context. For details, see the [Rust Reference](https://doc.rust-lang.org/reference/const_eval.html#const-functions).

https://docs.rs/smallvec/latest/smallvec/#const_new

mxinden (approved):
Other than `const_new` looks good to me.

For what it is worth, it will be enabled automatically anyways:

https://searchfox.org/mozilla-central/source/build/workspace-hack/Cargo.toml#64

neqo-transport/Cargo.toml
larseggert: Nope. But I thought there is no harm in including it. Can remove if you prefer.

mxinden: Preference for only enabling features that we make use of. If you don't feel strongly about it, please remove.

mxinden (approved):

Issue #2538, opened 2025-03-28
Why is there a large throughtput difference between upload and download tests?


larseggert:
For example, in https://github.com/mozilla/neqo/actions/runs/14130968402/attempts/1#summary-39591639225:
```Benchmarking 1-conn/1-100mb-resp/mtu-1504 (aka. Download)/client: Analyzing
1-conn/1-100mb-resp/mtu-1504 (aka. Download)/client
                        time:   [445.07 ms 449.70 ms 454.30 ms]
                        thrpt:  [220.12 MiB/s 222.37 MiB/s 224.69 MiB/s]
```
vs.
```
1-conn/1-100mb-resp/mtu-1504 (aka. Upload)/client
                        time:   [1.6145 s 1.6388 s 1.6636 s]
                        thrpt:  [60.110 MiB/s 61.022 MiB/s 61.938 MiB/s]
```

martinthomson:
Why is the latter called "100mb-resp" and not "100mb-req" ?

larseggert:
My guess is copy&paste error - @mxinden?

mxinden:
For what it is worth, here is a Flamegraph of a local `Upload` benchmark run. I have not had any luck identifying any bottleneck thus far.

![Image](https://github.com/user-attachments/assets/7a5f61a6-c558-479f-a2e2-8dc62372f680)

larseggert:
How does that flamegraph compare to a download one?

Also, have you looked at the flamegraphs generated on the bencher?

larseggert:
I think this got fixed.

mxinden:
Yes, https://github.com/mozilla/neqo/pull/2734 was the missing piece.

PR #2546, opened 2025-03-31
refactor(bin/bench): be specific about up- and download size

mxinden:
`neqo-bin` has a benchmark for download throughput, requests per second, handshakes per second and upload throughput. Previously this was specified as:

 ``` rust
struct Benchmark {
    name: String,
    requests: Vec<usize>,
    upload: bool,
}
```

This pull request makes the up- and download size explicit:

```
struct Benchmark {
    num_requests: usize,
    upload_size: usize,
    download_size: usize,
}
```

In addition, while previously one could either do an up- or a download benchmark, one could not do a combination of the two. This is due to the following line. Note that it ignores the URL response size parameter.

https://github.com/mozilla/neqo/blob/e72ae0ca8b3b413f854b136957c5d2bc83d69968/neqo-bin/src/server/http3.rs#L99-L102

This pull request fixes the above. When `neqo-server` receives a POST request, it waits for the client to close its write side, then responds with the amount of bytes specified in the URL path.


larseggert:
Oh, good catch.

larseggert:
Can we keep a bidirectional transfer as part of the benches?

mxinden:
> Can we keep a bidirectional transfer as part of the benches?

For sure. I will do a follow-up pull request. Sounds good?

mxinden:
I was wrong. Updated the PR description. Still worth merging.
 
> While a bug, this does not fix #2538, given that the HTTP3 server would simply ignore the `/104857600` URL path on receiving a `POST`.
> 
> https://github.com/mozilla/neqo/blob/e72ae0ca8b3b413f854b136957c5d2bc83d69968/neqo-bin/src/server/http3.rs#L99-L102

martinthomson:
Can we add separate options for upload and download size, so that the request URL contains the download size and the client just uploads/POSTs the upload size?

larseggert:
@mxinden is this OBE with #2599 merged?

mxinden:
Closing here in favor of https://github.com/mozilla/neqo/issues/2792.

PR #2551, opened 2025-04-02
Test that ECH and 0-RTT work

martinthomson:
Adds a comment to enable_resumption to explain why that needs to be called *after* properly configuring the client.


larseggert (approved):

neqo-crypto/min_version.txt
```diff
@@ -1 +1 @@
-3.105
+3.110
```
larseggert, line 1: If this required for the test, or do you just want to keep us current?

martinthomson: NSS crashes without the fix to https://bugzilla.mozilla.org/show_bug.cgi?id=1949677 which is in NSS 3.110.

PR #2552, opened 2025-04-02
fix: Unbreak the fuzzer more

larseggert:
This changes the way we link NSS to a fully static link when fuzzing or doing debug builds.

It also allows setting the `NSS_DB_PATH` env variable to `$ARGV0`, which indicates that the test fixture should look for the NSS database in the same directory as the executing binary.

Thanks, `oss-fuzz` for making us need all of this.

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -178,6 +178,9 @@ fn static_link() {
         "softokn_static",
         "ssl",
     ];
+    if env::consts::OS != "macos" {
+        static_libs.push("sqlite3");
+    }
     // Hardware specific libs.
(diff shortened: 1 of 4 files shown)


neqo-crypto/build.rs
```diff
@@ -177,10 +175,46 @@ fn static_link() {
         } else {
             "plds4"
         },
-        "smime",
         "softokn_static",
         "ssl",
     ];
+    // macOS always dynamically links against the system sqlite library.
+    // See https://github.com/nss-dev/nss/blob/a8c22d8fc0458db3e261acc5e19b436ab573a961/coreconf/Darwin.mk#L130-L135
+    if env::consts::OS == "macos" {
```
mxinden, line 183: When cross-compiling, `env::consts::OS` might not be the target OS.

I guess in this particular case it doesn't matter, but more generally, I assume using `env::var("CARGO_CFG_TARGET_OS")` would be more accurate?

PR #2555, opened 2025-04-03
fix: Simplify `Paths::path_response()`

### LINKED ISSUE #2309: Restructure `Paths::path_response`
I think that you can restructure this little piece of code to avoid the extra break.

```rust
let Some(primary) = self.migration_target.as_mut() else { break; }
if Rc::ptr_eq(target, primary) {
  let p = primary.take();
  mem::drop(self.select_primary(&p, now));
  return true;
}
break;
```

Once we're on 1.80 MSRV, we can use `Option::take_if` and simplify further.

```rust
if let Some(primary) = self.migration_target.take_if(|p| Rc::ptr_eq(target, p)) 
  mem::drop(self.select_primary(&primary, now));
  return true;
}
break;
```

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2294#discussion_r1895171720_

larseggert:
Fixes #2309

### DIFF
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -309,14 +309,10 @@ impl Paths {
             if p.borrow_mut().path_response(response, now, stats) {
                 // The response was accepted.  If this path is one we intend
                 // to migrate to, then migrate.
-                if self
+                if let Some(primary) = self
                     .migration_target
-                    .as_ref()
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Wasn't aware of `take_if`. Neat.

martinthomson (approved):
take_if is new-ish, so it wasn't an option originally.

PR #2556, opened 2025-04-03
chore: Get rid of some code by using `derive` more

larseggert:


neqo-crypto/src/lib.rs
```diff
@@ -103,7 +103,9 @@ fn version_check() -> Res<()> {
 /// This allows us to use SSLTRACE in all of our unit tests and programs.
 #[cfg(debug_assertions)]
 fn enable_ssl_trace() -> Res<()> {
-    let opt = Opt::Locking.as_int();
+    use ssl::PRInt32;
+
+    let opt = Opt::Locking as PRInt32;
```
mxinden, line 108: Not a strong opinion. It was nice that the potentially unsafe cast (i.e. `as`) was previously closer to the implementation (i.e. `Opt`).

neqo-transport/src/packet/mod.rs
```diff
@@ -48,41 +49,37 @@ pub use metadata::MetaData;
...
+        debug_assert!(
+            matches!(
+                self,
+                Self::Initial | Self::ZeroRtt | Self::Handshake | Self::Retry
+            ),
+            "is a long header packet type"
+        );
         // Version2 adds one to the type, modulo 4
-        (t + u8::from(v == Version::Version2)) & 3
+        (self as u8 + u8::from(v == Version::Version2)) & 3
```
mxinden, line 82: Previously on a short header type, this function would panic before this line in release mode. Now it only panics in debug mode, but still executes it in release mode. Is that intended?

mxinden (approved):

PR #2560, opened 2025-04-04
fix(transport/ecn): start after handshake and path validation

mxinden:
Fixes https://github.com/mozilla/neqo/issues/2490.

Only start ECN marking and thus ECN-Path validation when:

- Connection handshake is _confirmed_.
- AND, for any secondary path, path is validated (e.g. via `PATH_RESPONSE`).

----

**Draft for now. Opening up early to prevent duplicate work.**

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2509,7 +2509,7 @@ impl Connection {
 
             // If we don't have a TOS for this UDP datagram yet (i.e. `tos` is `None`), get it,
             // adding a `RecoveryToken::EcnEct0` to `tokens` in case of loss.
-            let tos = packet_tos.get_or_insert_with(|| path.borrow().tos(&mut tokens));
+            let tos = packet_tos.get_or_insert_with(|| path.borrow().tos(pt, &mut tokens));
             self.log_packet(
                 packet::MetaData::new_out(
                     path,
@@ -2948,6 +2948,7 @@ impl Connection {
                 .pmtud_mut()
                 .start(now, &mut self.stats.borrow_mut());
         }
+        self.paths.start_ecn();
         Ok(())
     }
 
--- a/neqo-transport/src/ecn.rs
+++ b/neqo-transport/src/ecn.rs
@@ -26,8 +26,12 @@ const TEST_COUNT_INITIAL_PHASE: usize = 3;
 
 /// The state information related to testing a path for ECN capability.
 /// See RFC9000, Appendix A.4.
-#[derive(Debug, PartialEq, Clone, Copy)]
+#[derive(Debug, PartialEq, Clone, Copy, Default)]
 enum ValidationState {
+    /// ECN validation not started yet. Reason might e.g. be still handshaking
+    /// or not being the primary path.
+    #[default]
+    NotStarted,
     /// The path is currently being tested for ECN capability, with the number of probes sent so
     /// far on the path during the ECN validation.
     Testing {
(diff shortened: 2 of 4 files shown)


neqo-transport/src/tracking.rs
```diff
@@ -453,11 +453,8 @@ impl RecvdPackets {
             return;
         }
 
-        builder.encode_varint(if self.ecn_count.is_some() {
-            FrameType::AckEcn
-        } else {
-            FrameType::Ack
-        });
+        // TODO: Send Ack and not AckEcn when no ecn counts available.
+        builder.encode_varint(FrameType::AckEcn);
```
mxinden, line 457: This change is a hack.

With this pull request we no longer need to send `FrameType::AckEcn` during the connection handshake, as the other side doesn't mark their datagrams during the handshake.

A QUIC AckECN is larger than a QUIC Ack, as it carries the various ECN counts.

The above size difference causes the server initial to fit into a single flight, thus breaking a lot of [`handshake.rs`](https://github.com/mozilla/neqo/blob/main/neqo-transport/src/connection/tests/handshake.rs) tests, specifically those that assume that the server initial doesn't fit in one flight and is thus later on coalesced with the handshake space.

Before I attempt fixing these, which is quite involved, double checking that this is the way we want to go.

@larseggert objections?

larseggert: I think this is the way to go, also given #2741.

larseggert:
@mxinden I tried to "help" and I think I messed up the rebase...

mxinden:
No worries. Rebased on to latest `main` now.

mxinden:
~~Only 10 more tests to go:~~

```
failures:
    connection::tests::keys::discarded_initial_keys
    connection::tests::migration::migrate_same_fail
    connection::tests::priority::critical
    connection::tests::priority::high_normal
    connection::tests::priority::important
    connection::tests::recovery::pto_handshake_complete
    connection::tests::resumption::ticket_rtt_larger_than_default
    connection::tests::resumption::ticket_rtt_less_than_default
    connection::tests::vn::compatible_upgrade
    connection::tests::vn::compatible_upgrade_large_initial
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1737,6 +1737,7 @@ impl Connection {
         // Handle each packet in the datagram.
         while !slc.is_empty() {
             self.stats.borrow_mut().packets_rx += 1;
+            self.stats.borrow_mut().dscp_rx[tos.into()] += 1;
```
mxinden, line 1740: Moved from below. Tests rightfully assert `packets_rx == dscp_rx[Cs0]`.

Previously, if any of the packets would be padding, i.e. garbage below, the two stats would diverge, as the former is counted above, the latter is counted below.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -208,7 +208,7 @@ fn dup_server_flight1() {
     assert!(out.as_dgram_ref().is_some());
     qdebug!("Output={:0x?}", out.as_dgram_ref());
 
-    assert_eq!(4, client.stats().packets_rx);
+    assert_eq!(3, client.stats().packets_rx);
```
mxinden, line 211: Most of the changes below simply revert changes made in https://github.com/mozilla/neqo/pull/2389/.

https://github.com/mozilla/neqo/pull/2389/ adjusted the tests of larger flight sizes due to MLKEM.

No longer doing ECN during the handshake replaces all ACK-ECN with ACK, thus oftentimes reducing the flight sizes again.

neqo-transport/src/tracking.rs
```diff
@@ -462,6 +462,7 @@ impl RecvdPackets {
         } else {
             FrameType::Ack
         });
+
```
mxinden:
```suggestion
```

neqo-transport/src/ecn.rs
```diff
@@ -313,11 +318,6 @@ impl Info {
             .count()
             .try_into()
             .expect("usize fits into u64");
-        if newly_acked_sent_with_ect0 == 0 {
-            qwarn!("ECN validation failed, no ECT(0) packets were newly acked");
-            self.disable_ecn(stats, ValidationError::Bleaching);
-            return;
-        }
```
mxinden, line 320: Say the packets acked here were sent during `ValidationState::NotStarted` or `ValidationState::Unknown`, they would thus not be marked, as expected. This is not a reason to disable ECN. 

This would fix https://github.com/mozilla/neqo/issues/2741.

@larseggert please double check the above reasoning.

larseggert: The logic is "if we sent any of the newly acked packets with ECN marks, but the count didn't increase in the ACK, disable".

mxinden: > if we sent any of the newly acked packets with ECN marks, but the count didn't increase in the ACK, disable

@larseggert that check is happening further below. I am not intending to delete it.

https://github.com/mozilla/neqo/blob/531cb0b6c21d27ff1ccec5502ecc346a6a922d71/neqo-transport/src/ecn.rs#L323-L327

mxinden: With the thumbps-up emoji, I am considering this resolved.

larseggert (approved):

PR #2564, opened 2025-04-07
chore: update tokio to v1.42.0

mxinden:
Previous version has unsoundness.

https://rustsec.org/advisories/RUSTSEC-2025-0023

Note, tokio is a dev-dependency only. (`dependency` in `neqo-bin` which itself is not used in Firefox.)


mxinden:
Closing in favor of https://github.com/mozilla/neqo/pull/2565.

Issue #2422, opened 2025-02-05
`hrtime::test::update_multi` fails occasionally in CI on Ubuntu


larseggert:
Debug and fix or disable.

larseggert:
I thought I had fixed that in #2636 but I guess not.

mxinden:
https://github.com/mozilla/neqo/pull/2636 disables the test on ARM.

https://github.com/mozilla/neqo/actions/runs/15073341987/job/42374916088?pr=2639 is a recent failure on x86-64.

martinthomson:
https://github.com/mozilla/neqo/actions/runs/15073341987/job/42374916088?pr=2639#step:7:291

larseggert:
Probably fixed by #2641

mxinden:
Reopening again, given recent failure:

https://github.com/mozilla/neqo/actions/runs/15418677161/job/43387638309?pr=2691

larseggert:
Another: https://github.com/mozilla/neqo/actions/runs/15485123773/job/43598054549

larseggert:
Two more:
* https://github.com/mozilla/neqo/actions/runs/15528951618/job/43713676263?pr=2717
* https://github.com/mozilla/neqo/actions/runs/15528951618/job/43713676248?pr=2717

@martinthomson is there anything we can tweak, or should we disable this test on `ubuntu-24.04-arm`?

martinthomson:
The slack we allow for timers is 1.5ms.  Here, I'm seeing up to twice that amount of slip for ARM.  We could have different constants for the different platforms, to allow for increased slippage on ARM.

PR #2565, opened 2025-04-07
chore(audit): ignore RUSTSEC-2025-0023

mxinden:
RUSTSEC-2025-0023 discloses a vulnerability in Tokio: "Broadcast channel calls clone in parallel, but does not require `Sync`".

https://rustsec.org/advisories/RUSTSEC-2025-0023

Tokio is only used in neqo-bin. neqo-bin does not make use of Tokio's broadcast channels. neqo-bin is not used by Firefox.

All Tokio versions with a fix for RUSTSEC-2025-0023 require a libc version bump, i.e. require us to use a libc version other than what mozilla-central uses.

Thus, instead of updating Tokio, this commit simply ignores the CVE.


mxinden:
CI failure due to https://github.com/mozilla/neqo/issues/2422.

.deny.toml
```diff
@@ -26,3 +26,6 @@ allow = [
 unknown-registry = "deny"
 unknown-git = "deny"
 required-git-spec = "rev"
+
+[advisories]
+ignore = ["RUSTSEC-2025-0023"]
```
larseggert, line 31:
```suggestion
ignore = ["RUSTSEC-2025-0023"] // FIXME: Remove this once Firefox moves to libc >= XX.YY
```
Where `XX.YY` is whatever is correct.

larseggert (approved):

PR #2568, opened 2025-04-07
fix: Fix QNS test errors with `nginx`

### LINKED ISSUE #2566: `nginx` and `haproxy` QNS tests broken
`git bisect` says it was 41036fc34fc8be1f881aad770996b7c0144815ba. Need to figure out why & fix it.

larseggert:
And hopefully `haproxy`.

The old code caused ACK-ECN frames to be sent that had all-zero counts, which I don't think is a violation of the RFC, but we still shouldn't do.

Fixes #2566.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,7 +511,7 @@ dependencies = [
 
 [[package]]
 name = "fuzz"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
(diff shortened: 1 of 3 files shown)


mxinden (approved):
Good catch!

Can you bump `version` in the root `Cargo.toml` to `0.13.1` in this pull request?

PR #2569, opened 2025-04-07
fix: Fix the `compatible_upgrade_large_initial` test

larseggert:
Equivalent of #2363 for `assert_dhcp`.


larseggert:
Merging this now to unbreak the test.

PR #2512, opened 2025-03-20
chore: Update `qlog` to 0.15.1

larseggert:
And use the newly derived `default`s to shorten some initializations in our code.

Upstreaming this will take some `cargo vet` work in Firefox.

PR #2571, opened 2025-04-08
chore(Cargo.toml): require qlog v0.15.1

mxinden:
https://github.com/mozilla/neqo/pull/2512/ bumped `qlog` to `v0.15.0`, but made use of features only available in `qlog` `v0.15.1`. Thus while using `qlog` `v0.15.0` would be allowed according to Neqo's root `Cargo.toml`, it would not compile.

This commit increases the `qlog` requirement to `v0.15.1`.


larseggert (approved):
Sorry for breaking things

mxinden:
No worries. I didn't catch this in my review either.

PR #2576, opened 2025-04-09
Introduce Certificate Compression Zlib Encoding/Decoding 

Frosne:
This patch: 

1) Extends bindings to support certificate compression set functionality and some functions to support it
2) Adds to flate2 to provide encoders/decoders (already present in https://searchfox.org/mozilla-central/source/third_party/rust/flate2, thus this specific implementation)
3) Provide a function **set_zlib_certificate_compression** that will enable advertising certificate compression in clientHello and decoding using zlib if we received a zlib encoded certificate
4) Introduce a test to check if connection is successful when both parties use zlib (important: server is encoding the certificate in this case) and only client supports zlib


neqo-crypto/Cargo.toml
```diff
@@ -17,6 +17,7 @@ workspace = true
 
 [dependencies]
 enum-map = { workspace = true }
+flate2 = "1.1.1"
```
larseggert, line 20: Could we add `default-features = false` and only enable those features actually needed?

neqo-crypto/src/agent.rs
```diff
@@ -571,6 +583,116 @@ impl SecretAgent {
         })
     }
 
+    /// `set_zlib_certificate_compression` enables using zlib encoding/decoding algorithms.
+    /// This asserts if no items are provided, or if any individual item is longer than
+    /// 255 octets in length.
```
larseggert, line 588: Why does it assert rather then return an error?

neqo-crypto/src/agent.rs
```diff
@@ -571,6 +583,116 @@ impl SecretAgent {
...
+                );
+                return rv;
+            });
+            match rv {
+                None => ssl::SECFailure,
+                Some(rv) => rv,
+            }
+        }
+
+        let encoding_zlib_name = CString::new("zlib").unwrap();
```
larseggert, line 678:
```suggestion
        let encoding_zlib_name = CString::new("zlib")?;
```

neqo-crypto/Cargo.toml
larseggert, line 20: Firefox is at a different version. Also pin in `Cargo.lock`.

```suggestion
flate2 = "1.0.30"
```

neqo-transport/src/crypto.rs
```diff
@@ -103,6 +103,7 @@ impl Crypto {
             // more configuration passed to server_enable_0rtt.
             c.enable_0rtt()?;
         }
+        agent.set_zlib_certificate_compression(false)?;
```
larseggert, line 106: Is there a reason why we don't turn this on by default?

martinthomson (changes requested):
I'm not sure that this is the right way to build this.  In NSS, we don't include the compressor.  That's so that the integration is left to the using code, especially Gecko, so that Gecko can reuse the (de)compressor that it has.  Having neqo bring its own (de)compressor denies the application that flexibility.

I would prefer to see a plugin-based approach, similar to the one used by NSS.

martinthomson:
Right.  I imagine you would define a trait that defines the compression interface.  Then the code would hold a `Box<dyn TheTrait>`.

Frosne:
> Right. I imagine you would define a trait that defines the compression interface. Then the code would hold a `Box<dyn TheTrait>`.

Is this what you meant? 
https://github.com/mozilla/neqo/pull/2592

I also wrote a prototype of how it should work in neqo_glue where we implement a trait: https://phabricator.services.mozilla.com/D245906 :) The phab patch will absolutely change, the intent was to check if we're on the same page about how it all should work (obviously, third_parts/rust will be gone)

larseggert:
Are we waiting on anything for this PR? Code changes by @Frosne or an updated review by @martinthomson?

martinthomson:
I don't think that this is the one we want.  #2592 seems better, but I'm still waiting on an update there.

larseggert:
OK. @Frosne, if this PR is OBE, please close?

PR #2579, opened 2025-04-10
ci: Use `samply` to convert `perf` data for the Firefox profiler

larseggert:
This should allow finer-grained stack- and flamegraphs that show inlined functions. Also might lead to smaller artifact sizes.


larseggert:
This is waiting for @mstange to add a `--presymbolicate` flag as discussed on Slack.

larseggert:
I'd like to merge this now. I added a TODO to switch back to installing binaries once there is a new samply release. Could I get a review?

PR #2580, opened 2025-04-13
ci(bench): fail on regression

mxinden:


mxinden:
The last couple of runs have shown no false positives. I suggest merging here.

In case we see e.g. the [`decoder.rs`](https://github.com/mozilla/neqo/blob/main/neqo-common/benches/decoder.rs) benchmark show more false positives, we can either revert here, or apply similar fixes like https://github.com/mozilla/neqo/pull/2587.

larseggert (approved):

PR #2587, opened 2025-04-14
bench(h3/streams): remove single stream variants

mxinden:
The single stream variants have a high variance.

With the perspective that they provide more noise than signal, remove them.


mxinden:
> 1000 streams of 1 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [26.054 ms 26.089 ms 26.123 ms]
>        change: [+5.3074% +5.5178% +5.7234%] (p = 0.00 < 0.05)
> 
> 10000 streams of 1 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [1.7037 s 1.7054 s 1.7070 s]
>        change: [+1.1424% +1.3007% +1.4576%] (p = 0.00 < 0.05)
> 
> Found 15 outliers among 100 measurements (15.00%)
> 11 (11.00%) low mild
> 3 (3.00%) high mild
> 1 (1.00%) high severe
> 
> 100 streams of 1000 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [3.5204 ms 3.5266 ms 3.5333 ms]
>        change: [+3.4699% +3.7635% +4.0449%] (p = 0.00 < 0.05)
> 
> Found 19 outliers among 100 measurements (19.00%)
> 19 (19.00%) high severe
> 
> 1000 streams of 1000 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [148.73 ms 148.82 ms 148.91 ms]
>        change: [+4.0261% +4.1115% +4.1981%] (p = 0.00 < 0.05)
> 
> Found 4 outliers among 100 measurements (4.00%)
> 4 (4.00%) high mild

Apparently the single-stream variants are not the only ones with high variance.

I have tried various tricks to make these more stable, without success thus far. For what it is worth, the majority of CPU time according to flamegraphs is spent in `neqo-crypto`.

neqo-http3/benches/streams.rs
```diff
@@ -76,21 +76,18 @@ fn connect() -> (Http3Client, Http3Server) {
...
 
-    for (streams, data_size) in [
-        (1, 1),
-        (1000, 1),
-        (10000, 1),
-        (1, 1000),
-        (100, 1000),
-        (1000, 1000),
-    ] {
+    for (streams, data_size) in [(1_000, 1), (1_000, 1_000)] {
```
mxinden, line 79: For the record, this benchmark was introduced in https://github.com/mozilla/neqo/pull/2444.

I don't have a strong opinion on the various variants. That said, for the sake of simplicity and for the sake of benchmark runtime, I would like to keep the number low.

If I understand https://github.com/mozilla/neqo/pull/2444 correctly, the initial intention of this benchmark was to measure the overhead of `neqo-http3` streams logic. I would argue that the two variants left (i.e. `[(1_000, 1), (1_000, 1_000)]` ) fulfill that need. The first will focus on the stream management overhead itself, given the tiny request payload of `1` byte only, the second on the overhead of large requests. Given that we want to measure stream overhead and not connection overhead, I don't see why we need runs with low number of stream, e.g. `(1, 1)`.

Am I missing an angle that is not covered with these two?

neqo-http3/benches/streams.rs
```diff
@@ -76,21 +76,18 @@ fn connect() -> (Http3Client, Http3Server) {
...
         let mut group = c.benchmark_group(format!("{streams} streams of {data_size} bytes"));
+
+        // High variance benchmark. Increase default sample size (100).
+        group.sample_size(500);
```
mxinden, line 83: The core goal of this pull request, reducing the number of false positive performance regressions.

neqo-http3/benches/streams.rs
```diff
@@ -76,21 +76,18 @@ fn connect() -> (Http3Client, Http3Server) {
...
+
         group.bench_function("multistream", |b| {
             let data = vec![0; data_size];
             b.iter_batched_ref(
                 connect,
                 |(client, server)| use_streams(client, server, streams, &data),
-                BatchSize::PerIteration,
+                BatchSize::SmallInput,
```
mxinden, line 90: I read the documentation of `BatchSize` as such, that one should only deviate from the default (i.e. `SmallInput`) on memory pressure issues. While `client` and `server` are likely large, I doubt they put the machine under memory pressure. Thus I suggest sticking with the default.

I don't feel strongly about it. Happy to revert.

https://docs.rs/criterion/latest/criterion/enum.BatchSize.html#variant.SmallInput

mxinden (commented):
The latest benchmark run still shows a regression. That said, the latest `main` only uses a _sample size_ of 100 and might be a false positive performance **improvement**.

With that in mind, I suggest we merge here. In case this doesn't achieve its goal, I can revert in the next couple of days.

@larseggert since I changed the PR significantly compared to its initial state, mind taking another look?

larseggert:
Also the neqo-common codec tests are sometimes flaky, so not sure if it's just this one.

larseggert (approved):

mxinden:
> Also the neqo-common codec tests are sometimes flaky, so not sure if it's just this one.

Thanks. I will look into it.

PR #2589, opened 2025-04-16
fix(transport/ecn): track ACK of marked packets

mxinden:
Alternative to https://github.com/mozilla/neqo/pull/2588.

**Problem**

(Same as https://github.com/mozilla/neqo/pull/2588.)

During ECN path testing, in case `TEST_COUNT_INITIAL_PHASE` ECN probes get lost, ECN is disabled.

https://github.com/mozilla/neqo/blob/a62c14068ad603f878f13a354f9ce0eddf97231b/neqo-transport/src/ecn.rs#L21-L25

Though only if `*probes_sent == *probes_lost`.

https://github.com/mozilla/neqo/blob/a62c14068ad603f878f13a354f9ce0eddf97231b/neqo-transport/src/ecn.rs#L241-L246

Say that a Neqo client has sent 2 ECN marked packets, i.e. probes, each of them eventually declared lost. Next the client sends two more packets, each again marked.

When the first of the two is declared lost, `probes_sent` is `4` and `probes_lost` is `3`, thus ECN is not disabled.

When the second of the two is declared lost, `probes_sent` is `4` and `probes_lost` is `4`. Given that `probes_lost` is not `== TEST_COUNT_INITIAL_PHASE`, ECN is once more not disabled.

**Solution**

(Alternative to https://github.com/mozilla/neqo/pull/2588.)

With this patch `neqo_transport::ecn` keeps track of ACKed marked packets. Instead of requiring `*probes_sent == *probes_lost`, only disable ECN if no marked packet has been acked already.

Note that as of now, this patch isn't relevant yet, as path probing is done at connection establishment time and thus `MAX_PATH_PROBES` already disables ECN marking beforehand.

https://github.com/mozilla/neqo/blob/a62c14068ad603f878f13a354f9ce0eddf97231b/neqo-transport/src/path.rs#L32-L36

That said, this will become relevant once we start ECN validation only after the handshake (https://github.com/mozilla/neqo/pull/2560).

---

@larseggert thank you for the suggestion in https://github.com/mozilla/neqo/pull/2588#issuecomment-2809048830. Is this what you had in mind?

Closes https://github.com/mozilla/neqo/pull/2588.

### DIFF
--- a/neqo-transport/src/ecn.rs
+++ b/neqo-transport/src/ecn.rs
@@ -32,6 +32,7 @@ enum ValidationState {
     /// far on the path during the ECN validation.
     Testing {
         probes_sent: usize,
+        initial_probes_acked: usize,
         initial_probes_lost: usize,
     },
     /// The validation test has concluded but the path's ECN capability is not yet known.
@@ -46,6 +47,7 @@ impl Default for ValidationState {
     fn default() -> Self {
         Self::Testing {
             probes_sent: 0,
+            initial_probes_acked: 0,
             initial_probes_lost: 0,
         }
     }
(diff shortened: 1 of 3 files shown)


neqo-transport/src/ecn.rs
```diff
@@ -225,20 +227,33 @@ impl Info {
...
     }
 
+    /// An [`IpTosEcn::Ect0`] marked packet has been acked.
+    pub(crate) fn acked_ecn(&mut self) {
+        if let ValidationState::Testing {
+            initial_probes_acked: probes_acked,
+            ..
+        } = &mut self.state
+        {
+            *probes_acked += 1;
```
larseggert, line 237: Should we immediately declare success when this happens?

mxinden: We certainly could.

I read the recommendation in RFC 9000 in a way that suggests to wait for more than one signal before declaring a path as _capable_:

> The goal is not to limit the duration of the testing period but to ensure that enough marked packets are sent for received ECN counts to provide a clear indication of how the path treats marked packets. [Section 13.4.2](https://www.rfc-editor.org/rfc/rfc9000.html#ecn-validation) suggests limiting this to ten packets or three times the PTO.

https://www.rfc-editor.org/rfc/rfc9000.html#ecn-alg

Do you expect paths to bleach intermittently? If so, I believe a testing period of 10 packets, will give us more accurate statistics.

Thoughts @larseggert?

larseggert: Sure. But I really wonder if there is intermittent bleaching anywhere.

larseggert (approved):

mxinden:
Thanks Lars!

PR #2592, opened 2025-04-17
Introduce SetCertificateCompression function

Frosne:
This patch introduces a function responsible for enabling using certificate compression during connection. 
This function takes as an input  a CertificateCompressor trait, that presents a certificate compression object. The trait provides 2 functions - encoder and decoder, name and a certificate compression identifier.


neqo-crypto/src/agent.rs
```diff
@@ -40,16 +43,67 @@ use crate::{
...
+));
+
+/// The trait is used to represent a certificate compression data structure
+/// Used in order to enable Certificate Compression extension during TLS connection
+pub trait CertificateCompressor {
+    /// Certificate Compression Algorithm identifier
+    /// zlib(1), brotli(2), zstd(3),
+    fn id(&self) -> u16;
+    /// Name of the certificate compression algorithm
+    fn name(&self) -> &'static str;
```
martinthomson, line 68:
```suggestion
    fn name(&self) -> &str;
```

I realize that the natural thing to do is have this be static, but we can have a narrower interface and allow compressors to be configured with strings that they own if they prefer.  That might make testing easier, for example.

neqo-crypto/src/agent.rs
```diff
@@ -40,16 +43,67 @@ use crate::{
...
+    /// The function that's used for encoding the certificate
+    /// Arguments:
+    ///
+    /// * `input`: Pointer to SECItem containing the bytes to encode
+    /// * `output`: Pointer to SECItem to store the result
+    ///
+    /// Note:  Certificate Compression encoding function is responsible
+    /// for allocating the output buffer itself.
+
+    fn encode(
```
martinthomson, line 79: This is not an encode function, but a getter for an encoder:

```suggestion
    fn encoder(
```

martinthomson (changes requested):
I realize that this is going to take a bunch more work to land, including in NSS.

neqo-crypto/src/agent.rs
```diff
@@ -40,16 +43,67 @@ use crate::{
...
+        &self,
+    ) -> ::std::option::Option<
+        unsafe extern "C" fn(input: *const SECItem, output: *mut SECItem) -> SECStatus,
```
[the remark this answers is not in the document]
Frosne: Yea, I agree. And I wanted to do it, but it did not work (maybe because my approach was not the correct one). 

So, this patch is supposed to work with this code in the neqo_glue: https://phabricator.services.mozilla.com/D245906. There is a function "decode_nss_secitem_with<F>" that takes a decoder as an input (as an F). The decoder is exactly what you expect 

```rust
 F: Fn(&[u8]) -> io::Result<Vec<u8>>
```

The function does some SecItem magic and applies the decoder. The function is then used with zlib, etc. 

And that's what I wanted to do in the trait, but rust requires me to either use a closure (and we still did not have enough of environment to catch I believe :)) or to change an approach

I.e. something like 
```rust
pub trait CertificateCompressor {
 fn decoder : Box<dyn (&[u8]) -> io::Result<Vec<u8>>>

....

decode_nss_secitem_with(encoder: Box<dyn CertificateCompressor>) {
    /// Read SecItem
   encoder.decode(...)
   /// Write SecItem
}
```
did not work.

neqo-crypto/src/agent.rs
```diff
@@ -571,6 +625,32 @@ impl SecretAgent {
...
+    /// This returns an error if the certificate compression could not be established
+    ///
+    /// [RFC8879]: https://datatracker.ietf.org/doc/rfc8879/
+
+    pub fn set_certificate_compression(
+        &mut self,
+        encoder: Box<dyn CertificateCompressor>,
+    ) -> Res<()> {
+        unsafe {
+            let compressor: SSLCertificateCompressionAlgorithm =
```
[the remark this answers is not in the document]
Frosne: Which approach seems to be the best for you? I can try to modify the NSS code (that would also cause some changes in the FF code, as it already uses the certificate compression) or I can leave the train with unsafe function pointers and use them in neqo_glue as in https://phabricator.services.mozilla.com/D245906

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: Dennis worked something out that is pretty nifty.  I iterated on it.  WDYT?  

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ea1c11b544d2f14cb7f1812c8de9c887

The only thing I'd do differently is avoid the allocation that comes from returning `Vec<u8>`, instead passing a mutable slice for the output and returning a subslice of that.  Something like:

```rust
trait Blah {
  const ID: u16;
  const NAME: &'static str;
  fn compress<'in, 'out>(input: &'in [u8], output: &'out mut [u8]) -> &'out [u8];
  fn decompress<'in, 'out>(input: &'in [u8], output: &'out mut [u8]) -> &'out [u8];
}
```

There is probably some trait magic we can employ to allow the `compress` function be optional, probably by splitting things between two traits, but maybe look at this for starters.

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: See [above](https://github.com/mozilla/neqo/pull/2592#discussion_r2055110334), where I think we can get away with no NSS changes.

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
Frosne: Sure, I will do it

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+
+                    let bytes_to_encode = null_safe_slice(input.as_ref().data, input.as_ref().len);
+                    let encoded_bytes = T::encode(bytes_to_encode);
+
+                    p11::SECITEM_MakeItem(
+                        null_mut(),
+                        // p11::SECItem is the same as ssl::SECItem
+                        output.cast::<p11::SECItemStr>(),
+                        encoded_bytes.as_ptr(),
+                        encoded_bytes.len().try_into().unwrap(),
```
martinthomson, line 149: I think that the general preference in this crate is to avoid `unwrap()`; you can instead do:

```rust
let Some(encoded_len) = c_uint::try_from(encoded_bytes.len()) else { return ssl::SECFailure; };
```

neqo-crypto/tests/agent.rs
```diff
@@ -535,3 +536,100 @@ fn ech_retry() {
...
+
+    assert!(client.state().is_connected());
+    assert!(server.state().is_connected());
+}
+
+struct CopyCompressionNoEncoder {}
+
+impl SafeCertificateCompression for CopyCompressionNoEncoder {
+    const ID: u16 = 0x4;
+    const NAME: &str = concat!("copy", "\0");
```
martinthomson, line 579: The trailing null here should not be necessary!  See above comments about CString usage.  It involves allocation, but that's fine in this case (and the compiler is likely able to optimize that out for this).

neqo-crypto/src/agent.rs
```diff
@@ -571,6 +675,30 @@ impl SecretAgent {
...
+        if T::ID == 0 {
+            return Err(Error::InvalidCertificateCompressionID);
+        }
+        let compressor: ssl::SSLCertificateCompressionAlgorithm =
+            ssl::SSLCertificateCompressionAlgorithm {
+                id: T::ID,
+                #[expect(clippy::as_ptr_cast_mut, reason = "requires to be const char")]
+                name: T::NAME.as_ptr() as *mut ::std::os::raw::c_char,
+                encode: T::ENABLE_ENCODING.then_some(<T as UnsafeCertCompression>::encode_callback),
+                decode: Some(<T as UnsafeCertCompression>::decode_callback),
```
martinthomson, line 697: Same here, though the raw pointer works.  (Again, untested.)

```suggestion
                decode: <T as UnsafeCertCompression>::decode_callback,
```

neqo-crypto/tests/agent.rs
```diff
@@ -535,3 +536,100 @@ fn ech_retry() {
...
+        .set_certificate_compression::<CopyCompression>()
+        .unwrap();
+    server
+        .set_certificate_compression::<CopyCompression>()
+        .unwrap();
+
+    connect(&mut client, &mut server);
+
+    assert!(client.state().is_connected());
+    assert!(server.state().is_connected());
```
martinthomson, line 572: These functions need to test that compression is active as well.  That means checking on whatever information about compression NSS has.

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+
+    unsafe extern "C" fn encode_callback(
+        input: *const ssl::SECItem,
+        output: *mut ssl::SECItem,
+    ) -> ssl::SECStatus;
+}
+
+/// The trait is used to represent a certificate compression data structure
+/// Used in order to enable Certificate Compression extension during TLS connection
+pub trait SafeCertificateCompression {
```
martinthomson, line 73: The public name for this doesn't need "safe" in it.

```suggestion
pub trait CertificateCompression {
```

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+    /// Certificate Compression identifier as in RFC8879
+    const ID: u16;
+    /// Certification Compression name (used only for logging/debugging)
```
martinthomson, line 76: This isn't used for reporting?  Seems like that could be improved in NSS.

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+    /// Certificate Compression could be used to encode and decode a certificate
+    /// though the encoding is not frequently used
+    /// Enable decoding field is used to signal to the implementation
+    /// to use the encoding as well
+    const ENABLE_ENCODING: bool = false;
+
+    /// Certificate Compression encoding function
+    /// If the implementation is not provided, we only copy the data
+    /// NB: If `ENABLE_ENCODING` is not set, the function pointer provided to NSS will be null
+    #[expect(clippy::must_use_candidate, reason = "Encoders are optional")]
```
martinthomson, line 87:
```suggestion
    #[must_use]
```

Anyone calling this function should use the returned value.  That's what `must_use` is for.  This interface allows it to be left `unimplemented!()` if `ENABLE_ENCODING` is false, that's all.

martinthomson (changes requested):
Here's an idea for this that might make it a little nicer for decompression-only users:

```rust
pub trait CertificateDecompression {
  const ID: u16;
  const NAME: &str;
  fn decode(data: &[u8]) -> Vec<u8>;
}

impl<T: CertificateDecompression> CertificateCompression for T {
  const ID = T::ID;
  const NAME = T::NAME;
  const ENABLE_ENCODING = false;
  fn encode(data: &[u8]) -> Vec<u8> { unimplemented!(); }
  fn decode(data: &[u8]) -> Vec<u8> { T::decode(data) }
}
```

WDYT?

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+                    }
+
+                    let bytes_to_decode_ptr =
+                        null_safe_slice(input.as_ref().data, input.as_ref().len);
+                    let decoded_bytes = T::decode(bytes_to_decode_ptr);
+                    if decoded_bytes.len() > output_len {
+                        return ssl::SECFailure;
+                    }
+
+                    std::ptr::copy_nonoverlapping(decoded_bytes.as_ptr(), output, output_len);
```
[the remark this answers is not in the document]
martinthomson: Oh, I forgot that the size should be known, so you can drop the return value entirely and simplify further.

neqo-crypto/src/agent.rs
Frosne: It's used for tracing, like we will log the message like "Server uses Zlib (for example) for decoding the received certificate"

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+            }
+        }
+    }
+
+    unsafe extern "C" fn encode_callback(
+        input: *const ssl::SECItem,
+        output: *mut ssl::SECItem,
+    ) -> ssl::SECStatus {
+        unsafe {
+            match std::ptr::NonNull::new(input.cast_mut()) {
```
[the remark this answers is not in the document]
Frosne: Extremely sorry, but what exactly do you mean? I am using ` std::ptr::NonNull::new(...) ` already.

neqo-crypto/src/agent.rs
Frosne: Agree, changed.

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+                None => ssl::SECFailure,
+                Some(input) => {
+                    if input.as_ref().data.is_null() || input.as_ref().len == 0 {
+                        return ssl::SECFailure;
+                    }
+
+                    let bytes_to_decode_ptr =
```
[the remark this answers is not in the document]
Frosne: You're absolutely correct! Thanks

neqo-crypto/src/agent.rs
Frosne: Done!

neqo-crypto/src/agent.rs
```diff
@@ -571,6 +675,30 @@ impl SecretAgent {
...
+    /// [RFC8879]: https://datatracker.ietf.org/doc/rfc8879/
+    pub fn set_certificate_compression<T: SafeCertificateCompression>(&mut self) -> Res<()> {
+        if T::ID == 0 {
+            return Err(Error::InvalidCertificateCompressionID);
+        }
+        let compressor: ssl::SSLCertificateCompressionAlgorithm =
+            ssl::SSLCertificateCompressionAlgorithm {
+                id: T::ID,
+                #[expect(clippy::as_ptr_cast_mut, reason = "requires to be const char")]
+                name: T::NAME.as_ptr() as *mut ::std::os::raw::c_char,
```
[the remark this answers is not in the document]
Frosne: That's what worries me here, please correct me if I'm missing anything:

CString will be allocated at the moment we create a compressor and then immediately deallocated? Do we get a freed pointer pointing to the name of the encoder/decoder?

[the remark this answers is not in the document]
Frosne: That's a pointer again: https://searchfox.org/mozilla-central/source/security/nss/lib/ssl/sslt.h#601

Should we rather allocate the name buffer in nss?

Frosne:
> @Frosne please run `cargo fmt` before pushing to fix formatting nits.

Will do!

neqo-crypto/src/agent.rs
Frosne: I'm not 100% sure? 
```
pub decode: ::std::option::Option<
        unsafe extern "C" fn(... ) -> SECStatus,
    >,
```

The compressor requires both encoder and decoder to be Option. Or it's not what you meant?

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,110 @@ use crate::{
...
+                None => ssl::SECFailure,
+                Some(input) => {
+                    if input.as_ref().data.is_null() || input.as_ref().len == 0 {
+                        return ssl::SECFailure;
+                    }
+
+                    let bytes_to_encode = null_safe_slice(input.as_ref().data, input.as_ref().len);
+                    let encoded_bytes = T::encode(bytes_to_encode);
+
+                    p11::SECITEM_MakeItem(
```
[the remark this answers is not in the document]
Frosne: I think the fact that we never used encoders encouraged us not to allocate anything on the NSS side...

neqo-crypto/tests/agent.rs
Frosne: I believe that NSS does not know anything about the compression we use. 

What we can do is to extend [SSLChannelInfoStr ](https://searchfox.org/mozilla-central/source/security/nss/lib/ssl/sslt.h#305) with the information of whether/which encoder is used. What do you think? P.s. `SSLCompressionMethod compressionMethod;` not to be mistaken for rfc8879 certificate compression

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
Frosne: What I was worried about is a malicious decoder:

We allocate some memory in NSS to store the decoded certificate. Then we provide this buffer to the decoder, right? (It's possible that I'm missing something). Then the decoder operates this memory to decode.

I'm worried that the decoder might not respect the allocated boundaries. So my plan was that the decoder allocates it's own memory, does its stuff and then we copy exactly the amount of bytes we need

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
Frosne: Also, a small remark, we can have something like let c_string = CString::new and then provide c_string to the compressor, but I'm not sure that it would really solve the problem

Frosne:
> Here's an idea for this that might make it a little nicer for decompression-only users:
> 
> ```rust
> pub trait CertificateDecompression {
>   const ID: u16;
>   const NAME: &str;
>   fn decode(data: &[u8]) -> Vec<u8>;
> }
> 
> impl<T: CertificateDecompression> CertificateCompression for T {
>   const ID = T::ID;
>   const NAME = T::NAME;
>   const ENABLE_ENCODING = false;
>   fn encode(data: &[u8]) -> Vec<u8> { unimplemented!(); }
>   fn decode(data: &[u8]) -> Vec<u8> { T::decode(data) }
> }
> ```
> 
> WDYT?

I'm trying to find a difference with this: 
```

pub trait CertificateCompression {
    /// Certificate Compression identifier as in RFC8879
    const ID: u16;
    /// Certification Compression name (used only for logging/debugging)
    const NAME: &str;
    /// Certificate Compression could be used to encode and decode a certificate
    /// though the encoding is not frequently used
    /// Enable decoding field is used to signal to the implementation
    /// to use the encoding as well
    const ENABLE_ENCODING: bool = false;

    /// Certificate Compression encoding function
    /// If the implementation is not provided, we only copy the data
    /// NB: If `ENABLE_ENCODING` is not set, the function pointer provided to NSS will be null
    #[must_use]
    fn encode(data: &[u8]) -> Vec<u8> {
         unimplemented!();
    }

    /// Certificate Compression decoding function
    #[must_use]
    fn decode(_data: &[u8]) -> Vec<u8>;
}
```

here we still have the benefit of `unimplemented()`, so the code is quite similar?

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: I'm saying that rather than spell out `std::ptr::NonNull` here, you can add a use statement to the top of the file:

```rust
use std::ptr::{NonNull, null, /* etc... */};
```

And then just have this as

```suggestion
            match NonNull::new(input.cast_mut()) {
```

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: How about a different approach then.  You call `SECITEM_MakeItem` to allocate something that is the same size as the input.  (Maybe with some fixed extra, if you think that's necessary to make testing easier.)  Then you make a slice from that and pass that slice to the caller.  Then we don't have two allocations (one in the trait, one for the new `SECItem`).

Right now, this is pretty inefficient.  Allocating more than we'd need isn't perfect, but it is better than performing two allocations.

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: Ah, I see the problem here.  This is a case where NSS depends on having a statically allocated string for naming.

Here's an alternative: change the interface to:

```rust
const NAME: &std::ffi::CStr = c"encoder-name";
```

Using a c-string here leaks that this is an FFI interface, but it avoids having to worry about lifetimes.

neqo-crypto/src/agent.rs
martinthomson: Oh, my bad.  I had no idea that this was indeed an `Option` in the bindings.  Forget this.  (I assumed that the pointer could just be null and that would be OK, but here we have an option.  The wonders never cease.)

neqo-crypto/tests/agent.rs
martinthomson: Yeah, I'm happy to defer that to a second change here, but I really do think that we need to be able to report a) whether compression was used, and b) what the compression saved us.  That probably means extending SSLChannelInfo with new fields.

neqo-crypto/src/agent.rs
martinthomson: See above.  We can do better (not for this change necessarily).

martinthomson:
The reason I suggested the different spelling is that it allows using code to implement something without having to type `unimplemented!()`.  It creates a guarantee that the code they write won't need special tests for the stub function.

larseggert:
Are we waiting on anything for this PR? Code changes by @Frosne or an updated review by @martinthomson?

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: This is not a situation you need to worry about.  That code is running rust, which can just crash itself or corrupt memory if it wants (we can't stop them from using `unsafe`).  More importantly, if we give them a slice and they stick to safe rust, that shouldn't happen.

neqo-crypto/src/agent.rs
```diff
@@ -50,6 +50,113 @@ use crate::{
...
+    ) -> ssl::SECStatus;
+}
+
+/// The trait is used to represent a certificate compression data structure
+/// Used in order to enable Certificate Compression extension during TLS connection
+pub trait CertificateCompression {
+    /// Certificate Compression identifier as in RFC8879
+    const ID: u16;
+    /// Certification Compression name (used only for logging/debugging)
+    const NAME: &str;
```
martinthomson, line 77: To make some of the below more concrete:

```suggestion
    const NAME: &CStr;
```

Frosne:
> Are we waiting on anything for this PR? Code changes by @Frosne or an updated review by @martinthomson?

Me! I will get back to the patch as soon as I can!

neqo-crypto/src/agent.rs
Frosne: Done

neqo-crypto/tests/agent.rs
Frosne: Do you mind if we land this patch first and then extend the SSLChannelInfo in the following patch?

Frosne:
> The reason I suggested the different spelling is that it allows using code to implement something without having to type `unimplemented!()`. It creates a guarantee that the code they write won't need special tests for the stub function.

Do you mean when we implement an encoder/decoder? But I believe it's to be already the case?

```
        struct CopyCompression {}
       
        impl CertificateCompression for CopyCompression {
        const ID: u16 = 0x4;
        const NAME: &CStr = c"copy";
        const ENABLE_ENCODING: bool = true;

        fn encode(data: &[u8]) -> Vec<u8> {
            data.to_vec()
        }

        fn decode(input: &[u8], output: &mut [u8]) -> usize {
            let len = std::cmp::min(input.len(), output.len());
            output[..len].copy_from_slice(&input[..len]);
            len
        }
    }
```

Implements both, whereas

``` struct CopyCompressionNoEncoder {}

impl CertificateCompression for CopyCompressionNoEncoder {
    const ID: u16 = 0x4;
    const NAME: &CStr = c"copy";

    fn decode(input: &[u8], output: &mut [u8]) -> usize {
        let len = std::cmp::min(input.len(), output.len());
        output[..len].copy_from_slice(&input[..len]);
        len
    }
}
```

is just a decoder

neqo-crypto/src/agent.rs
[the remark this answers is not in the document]
martinthomson: Don't worry about using an arena allocator here.  I just want to ensure that we aren't allocating unnecessarily.  We can justify some extra work here to make the interface clean -- this isn't that performance-critical -- so we don't need to make this totally perfect.

neqo-crypto/tests/agent.rs
martinthomson: Absolutely not.  I'd prefer it if we did it that way.  You'll need time to work out the NSS patch and I don't want this patch sitting here rotting in the meantime.

martinthomson:
I think that the ~ideal interface is:

```rust
pub trait CertificateDecoder {
  const ID: u16;
  const NAME: &CStr;
  fn decode(input: &[u8], output: &mut [u8]); 
}
```

Note that `decode()` has no need for a return value.  Decoding has to work and the output length has to match the length of `output`, because that is the way the spec works: we get the length.

And then:
```rust
pub trait CertificateEncoder: CertificateDecoder {
  fn encode(input: &[u8], output: &mut [u8]) -> usize; 
}
```

`encode()` needs a length returned, and we'd want to pass `input.len()` or `input.len() + 1` of pre-allocated memory for this; there are drawbacks to overallocation, but certificates are small enough that we don't need anything more complicated.  The other option is:

```rust
pub trait CertificateEncoder: CertificateDecoder {
  fn encode<'in, 'out>(input: &'in [u8], output: &'out mut [u8]) -> &'out [u8]; 
}
```

That is, the return value is a slice of the output.  That's much less good because it might not be a slice that starts at the beginning of the allocation and I don't want to deal with passing that off to NSS using a `SECItem`.

Frosne:
> I think that the ~ideal interface is:
> 
> ```rust
> pub trait CertificateDecoder {
>   const ID: u16;
>   const NAME: &CStr;
>   fn decode(input: &[u8], output: &mut [u8]); 
> }
> ```
> 
> Note that `decode()` has no need for a return value. Decoding has to work and the output length has to match the length of `output`, because that is the way the spec works: we get the length.
> 
> And then:
> 
> ```rust
> pub trait CertificateEncoder: CertificateDecoder {
>   fn encode(input: &[u8], output: &mut [u8]) -> usize; 
> }
> ```
> 
> `encode()` needs a length returned, and we'd want to pass `input.len()` or `input.len() + 1` of pre-allocated memory for this; there are drawbacks to overallocation, but certificates are small enough that we don't need anything more complicated. The other option is:
> 
> ```rust
> pub trait CertificateEncoder: CertificateDecoder {
>   fn encode<'in, 'out>(input: &'in [u8], output: &'out mut [u8]) -> &'out [u8]; 
> }
> ```
> 
> That is, the return value is a slice of the output. That's much less good because it might not be a slice that starts at the beginning of the allocation and I don't want to deal with passing that off to NSS using a `SECItem`.

I still believe that we need to return a result after the decoding operation in order to transfer the information whether the decoding was successful or not. Same for encoding. 

I think I understand your logic about having 2 public traits, for encoding and decoder, but when I was trying to implement that I realised that it causes code duplication (for example, if you look at the set_certificate_compression function)

martinthomson:
> I still believe that we need to return a result after the decoding operation in order to transfer the information whether the decoding was successful or not. Same for encoding.

I thought we'd established that both encode and decode are infallible.  Certainly that is the way that the NSS signatures operate.

> I think I understand your logic about having 2 public traits, for encoding and decoder, but when I was trying to implement that I realised that it causes code duplication (for example, if you look at the set_certificate_compression function)

A small amount of code duplication on our end should be manageable.  It's not a lot of code.

Frosne:
> > I still believe that we need to return a result after the decoding operation in order to transfer the information whether the decoding was successful or not. Same for encoding.
> 
> I thought we'd established that both encode and decode are infallible. Certainly that is the way that the NSS signatures operate.
> 
If we take zlib, https://docs.rs/flate2/latest/flate2/read/struct.ZlibDecoder.html, [read_to_end](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.read_to_string) function (the one to get the bytes)  could return an error. 

> > I think I understand your logic about having 2 public traits, for encoding and decoder, but when I was trying to implement that I realised that it causes code duplication (for example, if you look at the set_certificate_compression function)
> 
> A small amount of code duplication on our end should be manageable. It's not a lot of code.

But in the very least it would mean two `set_certificate_compression` functions, one for CompressionDecoder and one for CompressionEncoder. Do we really want this? I am afraid that it makes the outside code less clear..

martinthomson:
Yeah, my mistake on the error. We should make the function return a Result.

As for the other problem, there is a trick. I think. Give me a chance to check that out and we will see. In the meantime, a single trait is perfectly fine.

Frosne:
> Yeah, my mistake on the error. We should make the function return a Result.
> 
> As for the other problem, there is a trick. I think. Give me a chance to check that out and we will see. In the meantime, a single trait is perfectly fine.

There is a notion of downcasting in rust IIRC, maybe that's what you meant?: 

```
fn set_compression_alg(compressor: &dyn CertificateDecoder)
{
   if let Some(compressor) = compressor.as_any().downcast_ref::<CertificateEncoder>() 
   {
     // input is encoder 
   }
}
```

martinthomson:
I was hoping to avoid the need for a `dyn` reference, but it might work in this case.

The alternative I had in mind was to have something like this:

```rust
pub trait Decoder {
  /* ID and name properties omitted */
  fn decode(/* params */) -> Result<usize, ()>;
}
pub trait Encoder : Decoder {
  const HAS_ENCODER: bool = true;
  fn encode(/* params */) -> Result<usize, ()>;
}
pub struct DecodeOnly<T>;
impl<T: Decoder> Encoder for DecodeOnly<T> {
  const HAS_ENCODER: bool = false;
  fn encode(/* params */) -> Result<usize, ()> { unimplemented!(); }
}
```
Which allows call sites the option of calling this with:

```rust
fn set_compression<T: Encoder>(&mut self) { /* impl */ }
c.set_compression::<DecodeOnly<MyDecoder>>();
```

That avoids the downcast issue of needing an object reference to have it work, where we are using types without instances of anything.

Frosne:
@mxinden do you happen to know why I fail the CI? https://github.com/mozilla/neqo/actions/runs/15711244723/job/44269626849 for example.. [set_alpn](https://searchfox.org/mozilla-central/source/third_party/rust/neqo-transport/src/connection/mod.rs#558) seems absolutely ok

larseggert:
@Frosne The CI error message tells you what to fix? You can also run `cargo clippy` locally, or `cargo hack clippy --all-targets --benches --feature-powerset --exclude-features gecko -- -D warnings` if you want to run exactly what CI runs.

Frosne:
Ok, I seem to be left with only one red. 

@larseggert do you know how to improve the codecov?

larseggert:
If you look at https://app.codecov.io/gh/mozilla/neqo/commit/b33d3240ab68200a1355c92edbe744efdfbc6ef8, the `agent.rs` accordion shows that the `encode` function has no test coverage. Everything else looks OK.

Frosne:
> If you look at https://app.codecov.io/gh/mozilla/neqo/commit/b33d3240ab68200a1355c92edbe744efdfbc6ef8, the `agent.rs` accordion shows that the `encode` function has no test coverage. Everything else looks OK.

I'm curious why, because I do have a test using the encode function

larseggert:
Are you sure it's testing _this_ encode function? Maybe add a +1 somewhere and see if the test still passes.

Frosne:
> Are you sure it's testing _this_ encode function? Maybe add a +1 somewhere and see if the test still passes.

I understood the problem. I've introduced a default encoding, but have not tested specifically the default version

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+    const NAME: &CStr;
+    /// Certificate Compression could be used to encode and decode a certificate
+    /// though the encoding is not frequently used
+    /// Enable decoding field is used to signal to the implementation
+    /// to use the encoding as well
+    const ENABLE_ENCODING: bool = false;
+
+    /// Certificate Compression encoding function
+    /// If the implementation is not provided, we only copy the data
+    /// NB: If `ENABLE_ENCODING` is not set, the function pointer provided to NSS will be null
```
```suggestion
    ///
    /// This default implementation effectively does nothing.
    /// However, this is only run if `ENABLE_ENCODING` is `true`.
    /// Implementations that set `ENABLE_ENCODING` to `true` need to implement this function.
    ///
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+    /// # Errors
+    /// Encoding was unsuccessful, for example, not enough memory
+    fn encode(input: &[u8], output: &mut [u8]) -> Res<usize> {
+        let len = std::cmp::min(input.len(), output.len());
+        output[..len].copy_from_slice(&input[..len]);
+        Ok(len)
+    }
+
+    /// Certificate Compression decoding function. Returns the length of the decoded buffer,
+    /// or 0 if an error has occured.
```
```suggestion
    /// Certificate Compression decoding function.
    /// Returns the length of the decoded buffer.
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+/// The trait is responsible for calling `CertificateCompression` encoding and decoding
+/// functions using the NSS types
+unsafe impl<T: CertificateCompressor> UnsafeCertCompression for T {
+    unsafe extern "C" fn decode_callback(
+        input: *const ssl::SECItem,
+        output: *mut ::std::os::raw::c_uchar,
+        output_len: usize,
+        used_len: *mut usize,
+    ) -> ssl::SECStatus {
+        unsafe {
```
martinthomson, line 126: I prefer to focus the unsafe blocks on the code that is genuinely unsafe.  In this case, that's a much smaller scope than this (lines 116, 120, and 121 only in my current diff view).  Can you move the block to that smaller scope?

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                }
+            }
+        }
+    }
+
+    unsafe extern "C" fn encode_callback(
+        input: *const ssl::SECItem,
+        output: *mut ssl::SECItem,
+    ) -> ssl::SECStatus {
+        unsafe {
```
martinthomson, line 150: This has two bits of unsafe code, but it's still smaller-scoped than the whole function.

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                    let input_slice = null_safe_slice(input.as_ref().data, input_len);
+                    let output_slice: &mut [u8] = core::slice::from_raw_parts_mut(
+                        (*output).data,
+                        (*output).len.try_into().unwrap(),
+                    );
+
+                    let encode_result = T::encode(input_slice, output_slice);
+                    match encode_result {
+                        Err(_) => ssl::SECFailure,
+                        Ok(encoded_len) => {
```
martinthomson: This might be good for a:

```rust
let Ok(encoded_len) = T::encode(input_slice, output_slice) else {
    return ssl::SECFailure;
};
```
Just to keep the indenting reasonable.

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+    unsafe extern "C" fn decode_callback(
+        input: *const ssl::SECItem,
+        output: *mut ::std::os::raw::c_uchar,
+        output_len: usize,
+        used_len: *mut usize,
+    ) -> ssl::SECStatus {
+        unsafe {
+            match NonNull::new(input.cast_mut()) {
+                None => ssl::SECFailure,
+                Some(input) => {
```
martinthomson: I would suggest using:

```rust
let Some(input) = NonNull::new(input.cast_mut()) else {
    return ssl::SECFailure;
};
```

To keep indent levels sane.

neqo-crypto/src/agent.rs
```diff
@@ -572,6 +712,30 @@ impl SecretAgent {
...
+    /// [RFC8879]: https://datatracker.ietf.org/doc/rfc8879/
+    pub fn set_certificate_compression<T: CertificateCompressor>(&mut self) -> Res<()> {
+        if T::ID == 0 {
+            return Err(Error::InvalidCertificateCompressionID);
+        }
+
+        let compressor: ssl::SSLCertificateCompressionAlgorithm =
+            ssl::SSLCertificateCompressionAlgorithm {
+                id: T::ID,
+                name: T::NAME.as_ptr().cast_mut(),
```
martinthomson: The C type is `const char*`; does this really need the const cast?

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                        return ssl::SECFailure;
+                    }
+
+                    let input_len = input.as_ref().len;
+
+                    p11::SECITEM_AllocItem(
+                        null_mut(),
+                        // p11::SECItem is the same as ssl::SECItem
+                        output.cast::<p11::SECItemStr>(),
+                        input_len + 1,
```
```suggestion
                        // Compression shouldn't make the thing *longer*,
                        // but allocate one extra byte anyway to enable simple testing modes.
                        input_len + 1,
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                        Ok(encoded_len) => {
+                            if encoded_len == 0 {
+                                return ssl::SECFailure;
+                            }
+
+                            let rv = p11::SECITEM_ReallocItemV2(
+                                null_mut(),
+                                output.cast::<p11::SECItemStr>(),
+                                encoded_len.try_into().unwrap(),
+                            );
```
martinthomson: It's possible that you can avoid the reallocation.  Modifying the length of the `SECItem` should suffice.

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                    let output_slice: &mut [u8] = core::slice::from_raw_parts_mut(
+                        (*output).data,
+                        (*output).len.try_into().unwrap(),
+                    );
+
+                    let encode_result = T::encode(input_slice, output_slice);
+                    match encode_result {
+                        Err(_) => ssl::SECFailure,
+                        Ok(encoded_len) => {
+                            if encoded_len == 0 {
```
martinthomson: IMPORTANT: you need to check that `encoded_len <= output.len()`.  I would add a `debug_assert!()` for that in addition to a runtime check that returns SECFailure.

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+            match NonNull::new(input.cast_mut()) {
+                None => ssl::SECFailure,
+                Some(input) => {
+                    if input.as_ref().data.is_null() || input.as_ref().len == 0 {
+                        return ssl::SECFailure;
+                    }
+
+                    let input_slice = null_safe_slice(input.as_ref().data, input.as_ref().len);
+                    let output_slice: &mut [u8] =
+                        core::slice::from_raw_parts_mut(output, output_len);
```
martinthomson: Please use `use std::slice` instead.  And do you really need the type annotation for this?

If this is to avoid the errors that come from using those functions, it's OK to suppress the warnings.

(It's safe to assume that `output != NULL` and use this function directly, but we generally use functions from std.)

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +51,146 @@ use crate::{
...
+                        null_mut(),
+                        // p11::SECItem is the same as ssl::SECItem
+                        output.cast::<p11::SECItemStr>(),
+                        input_len + 1,
+                    );
+
+                    let input_slice = null_safe_slice(input.as_ref().data, input_len);
+                    let output_slice: &mut [u8] = core::slice::from_raw_parts_mut(
+                        (*output).data,
+                        (*output).len.try_into().unwrap(),
```
martinthomson: I generally prefer `x::try_from(y)`, but this is probably OK.  As above, don't use `core::slice` functions.

neqo-crypto/tests/agent.rs
```diff
@@ -530,3 +533,183 @@ fn ech_retry() {
...
+    const ID: u16 = 0x4;
+    const NAME: &CStr = c"copy";
+
+    const ENABLE_ENCODING: bool = true;
+
+    fn encode(input: &[u8], output: &mut [u8]) -> Res<usize> {
+        let len = std::cmp::min(input.len(), output.len());
+        output[..len].copy_from_slice(&input[..len]);
+        Ok(len)
+    }
```
martinthomson: This copies the default implementation

```suggestion
```

neqo-crypto/tests/agent.rs
```diff
@@ -530,3 +533,183 @@ fn ech_retry() {
...
+        .set_certificate_compression::<CopyCompressionNoEncoder>()
+        .unwrap();
+
+    connect(&mut client, &mut server);
+
+    assert!(client.state().is_connected());
+    assert!(server.state().is_connected());
+}
+
+struct CopyCompressionNoEncoderReturnsErr {}
```
martinthomson: This type is one that the DECODER returns an error.

You should add a test for when the ENCODER fails.

martinthomson (changes requested):
Getting close, I think.

neqo-crypto/src/agent.rs
Frosne: I think there are 4 pieces requiring unsafe, so I left it as it is 
(These are: as_ref, Alloc_Item, slice, Realloc)

neqo-crypto/src/agent.rs
Frosne: Done!

neqo-crypto/src/agent.rs
Frosne: I believe that there is no function like this. In theory, Realloc should check for the size before reallocating and make a reallocation only if the size is longer

neqo-crypto/src/agent.rs
Frosne: You're absolutely correct!

neqo-crypto/tests/agent.rs
Frosne: Done

neqo-crypto/tests/agent.rs
Frosne: Done

neqo-crypto/src/agent.rs
Frosne: I'm curious if it can happen because we already provide the output_slice of the size input_len + 1, but I added the check

neqo-crypto/src/agent.rs
martinthomson: If you do as I suggest regarding the `let Some(foo) = bar else { return; };` form, then the unsafe call can be scoped more tightly.

neqo-crypto/src/agent.rs
martinthomson: I'm suggesting that you set `output.len = encoded_len`, not use a function.

Frosne: Done!

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,143 @@ use crate::{
...
+        let output_slice = unsafe { slice::from_raw_parts_mut(output, output_len) };
+
+        let Ok(decoded_len) = T::decode(input_slice, output_slice) else {
+            return ssl::SECFailure;
+        };
+
+        if decoded_len != output_len {
+            return ssl::SECFailure;
+        }
+
```
martinthomson: Why offer this opportunity to fail?  If decode returns `Res<()>`, then you don't need this check.

Put differently, if there is no information carried in the return value of the function, there is no point in having the return value.

```suggestion
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,143 @@ use crate::{
...
+    /// # Errors
+    /// Encoding was unsuccessful, for example, not enough memory
+    fn encode(input: &[u8], output: &mut [u8]) -> Res<usize> {
+        let len = std::cmp::min(input.len(), output.len());
+        output[..len].copy_from_slice(&input[..len]);
+        Ok(len)
+    }
+
+    /// Certificate Compression decoding function.
+    /// Returns the length of the decoded buffer.
```
```suggestion
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,143 @@ use crate::{
...
+    /// # Errors
+    /// Decoding was unsuccessful, for example, not enough memory
+    fn decode(input: &[u8], output: &mut [u8]) -> Res<usize>;
```
```suggestion
    fn decode(input: &[u8], output: &mut [u8]) -> Res<()>;
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,143 @@ use crate::{
...
+            return ssl::SECFailure;
+        };
+        if unsafe { input.as_ref().data.is_null() || input.as_ref().len == 0 } {
+            return ssl::SECFailure;
+        }
+
+        let input_slice = unsafe { null_safe_slice(input.as_ref().data, input.as_ref().len) };
+        let output_slice = unsafe { slice::from_raw_parts_mut(output, output_len) };
+
+        let Ok(decoded_len) = T::decode(input_slice, output_slice) else {
```
```suggestion
        if T::decode(input_slice, output_slice).is_err() {
```

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,143 @@ use crate::{
...
+            return ssl::SECFailure;
+        };
```
```suggestion
        }
```

neqo-crypto/tests/agent.rs
```diff
@@ -530,3 +533,288 @@ fn ech_retry() {
...
+
+    client
+        .set_certificate_compression::<CompressionEncoderReturnsTooLong>()
+        .unwrap();
+
+    connect_fail(&mut client, &mut server);
+}
+
+#[test]
+fn connection_fails_decoder_returned_too_long() {
```
martinthomson: This test wouldn't be valid with the changes I recommend.

neqo-crypto/src/agent.rs
Frosne: RFC says that 
_"The length of the Certificate message once it is uncompressed. If, after decompression, the specified length does not match the actual length, the party receiving the invalid message MUST abort the connection with the "bad_certificate" alert. "_

If we don't return the decoded_len, we can't make this check, right? Or am I missing something?

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,141 @@ use crate::{
...
+            (input_ref.data, input_ref.len)
+        };
+
+        if input_data.is_null() || input_len == 0 {
+            return ssl::SECFailure;
+        }
+        let input_slice = unsafe { null_safe_slice(input_data, input_len) };
+
+        unsafe {
+            p11::SECITEM_AllocItem(
```
martinthomson, line 151: This needs a NULL check in case it fails.  Sorry I missed that earlier.

neqo-crypto/src/agent.rs
```diff
@@ -51,6 +52,141 @@ use crate::{
...
+            return ssl::SECFailure;
+        };
+
+        let output_slice = unsafe { slice::from_raw_parts_mut((*output).data, output_len) };
+
+        let Ok(encoded_len) = T::encode(input_slice, output_slice) else {
+            return ssl::SECFailure;
+        };
+
+        if encoded_len == 0 || encoded_len > output_len || encoded_len > u32::MAX as usize {
```
```suggestion
        if encoded_len == 0 || encoded_len > output_len {
```
The try_into call below will catch the out of bounds problem.

neqo-crypto/src/agent.rs
```diff
@@ -572,6 +708,30 @@ impl SecretAgent {
...
+    /// If the compression mechanism with the same id is already registered
+    /// If too many compression mechanisms are already registered
+    ///
+    /// This returns an error if the certificate compression could not be established
+    ///
+    /// [RFC8879]: https://datatracker.ietf.org/doc/rfc8879/
+    pub fn set_certificate_compression<T: CertificateCompressor>(&mut self) -> Res<()> {
+        if T::ID == 0 {
+            return Err(Error::InvalidCertificateCompressionID);
+        }
```
martinthomson, line 724: I like this, but you could make this structural if you change the type of ID to `NonZeroU16`.

It's a bit more annoying to use:

```rust
   const ID: NonZeroU16 = NonZeroU16::new(12).unwrap();
```

But you save on the need for a new error enum member.

martinthomson (approved):
Down to some discretionary stuff, aside from the NULL check.

larseggert:
The benchmark failure is due to me taking new runners into service and running out of time for bug fixing today. See #2760.

neqo-crypto/src/agent.rs
larseggert: @Frosne Wanted to check if you are planning to change this, or whether I should merge as-is?

neqo-crypto/src/agent.rs
larseggert: @Frosne Anything left to be done here? Are you waiting to hear from @martinthomson or is this resolved?

larseggert:
@Frosne There are a few open comments from @martinthomson; are you still planning any changes related to them? Otherwise, if you consider them resolved/addressed, please mark them as such.

neqo-crypto/src/agent.rs
Frosne: I would leave it as its now

Frosne:
> @Frosne There are a few open comments from @martinthomson; are you still planning any changes related to them? Otherwise, if you consider them resolved/addressed, please mark them as such.

I think now everything is resolved (if I have not miss anything)

Frosne:
Thanks!

PR #2605, opened 2025-04-29
chore: Fix new nightly `clippy` warnings

larseggert:


neqo-transport/src/rtt.rs
```diff
@@ -129,11 +129,7 @@ impl RttEstimate {
             self.first_sample_time = Some(now);
         } else {
             // Calculate EWMA RTT (based on {{?RFC6298}}).
-            let rttvar_sample = if self.smoothed_rtt > rtt_sample {
-                self.smoothed_rtt - rtt_sample
-            } else {
-                rtt_sample - self.smoothed_rtt
-            };
+            let rttvar_sample = self.smoothed_rtt.abs_diff(rtt_sample);
```
mxinden, line 132: clippy is quite good :bulb:

mxinden (approved):

martinthomson (approved):

neqo-transport/src/rtt.rs
martinthomson: The real implementation is the same as what we had, so I'm not that impressed.  What is impressive is the depth of the library of rules that clippy runs.

PR #2613, opened 2025-05-06
Adding robustness checking to the QUIC handshake

alizohaib:
This change allows clients to transmit a variable number of pre-initial packets (default 1).

### DIFF
--- a/neqo-bin/src/lib.rs
+++ b/neqo-bin/src/lib.rs
@@ -125,6 +125,10 @@ pub struct QuicParameters {
     /// Whether to slice the SNI.
     pub no_sni_slicing: bool,
 
+    #[arg(long)]
+    /// Whether to add dummy packets before client Initial packets.
+    pub no_pre_init_noise: bool,
+
     #[arg(name = "preferred-address-v4", long)]
     /// An IPv4 address for the server preferred address.
     pub preferred_address_v4: Option<String>,
@@ -148,6 +152,7 @@ impl Default for QuicParameters {
             preferred_address_v4: None,
             preferred_address_v6: None,
             no_sni_slicing: false,
+            no_pre_init_noise: false,
         }
     }
 }
@@ -221,7 +226,8 @@ impl QuicParameters {
             .cc_algorithm(self.congestion_control)
             .pacing(!self.no_pacing)
             .pmtud(!self.no_pmtud)
-            .sni_slicing(!self.no_sni_slicing);
+            .sni_slicing(!self.no_sni_slicing)
+            .pre_init_noise(!self.no_pre_init_noise);
         params = if let Some(pa) = self.preferred_address() {
             params.preferred_address(pa)
         } else {
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -25,9 +25,9 @@ use neqo_common::{
 use neqo_crypto::{
     agent::CertificateInfo, Agent, AntiReplay, AuthenticationStatus, Cipher, Client, Group,
     HandshakeState, PrivateKey, PublicKey, ResumptionToken, SecretAgentInfo, SecretAgentPreInfo,
-    Server, ZeroRttChecker,
+    Server, ZeroRttChecker, random, randomize,
 };
-use smallvec::SmallVec;
+use smallvec::{smallvec, SmallVec};
 use strum::IntoEnumIterator as _;
 
 use crate::{
@@ -87,6 +87,11 @@ pub use crate::send_stream::{RetransmissionPriority, SendStreamStats, Transmissi
 /// to receiving an undecryptable packet during the early part of the
 /// handshake.  This is a hack, but a useful one.
 const EXTRA_INITIALS: usize = 4;
+
+/// Maximum number of dummy datagrams to send before Client Initial packets.
+const MAX_PRE_INIT_PKTS: usize = 1;
+/// Payload size range (in bytes) for pre-Initial packets.
+const PRE_INIT_PAYLOAD_LEN_RANGE: (usize, usize) = (1200, 1400);
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum ZeroRttState {
@@ -291,6 +296,7 @@ pub struct Connection {
     release_resumption_token_timer: Option<Instant>,
     conn_params: ConnectionParameters,
     hrtime: hrtime::Handle,
+    pre_initial_pkts: usize,
 
     /// For testing purposes it is sometimes necessary to inject frames that wouldn't
     /// otherwise be sent, just to see how a connection handles them.  Inserting them
@@ -351,6 +357,10 @@ impl Connection {
             &mut c.stats.borrow_mut(),
         );
         c.setup_handshake_path(&Rc::new(RefCell::new(path)), now);
+
+        // Generate a random number between 1 and the maximum number of pre-initial packets
+        c.pre_initial_pkts = (random::<1>()[0] as usize % MAX_PRE_INIT_PKTS) + 1;
+        
         Ok(c)
     }
 
@@ -437,6 +447,7 @@ impl Connection {
             conn_params,
             hrtime: hrtime::Time::get(Self::LOOSE_TIMER_RESOLUTION),
             quic_datagrams,
+            pre_initial_pkts: 0,
             #[cfg(test)]
             test_frame_writer: None,
         };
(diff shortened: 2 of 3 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1153,25 @@ impl Connection {
 
         match (&self.state, self.role) {
             (State::Init, Role::Client) => {
+                // Before starting the client, send some noise
+                if self.conn_params.pre_init_noise_enabled() && self.pre_initial_pkts > 0 {
```
martinthomson, line 1157: It should be enough to just check the counter.  The counter should be zero otherwise.

martinthomson (commented):
@larseggert, would you mind taking a look at this as well?

I am not sure about this one: it's sending a number of *datagrams*, not packets.  That probably isn't right.  And I need to understand what the goals are for this feature.

neqo-bin/src/lib.rs
```diff
@@ -125,6 +125,10 @@ pub struct QuicParameters {
     /// Whether to slice the SNI.
     pub no_sni_slicing: bool,
 
+    #[arg(long)]
+    /// Whether to add dummy packets before client Initial packets.
```
larseggert, line 129:
```suggestion
    /// Whether to send random noise packets before client Initial packets.
```

neqo-transport/src/connection/mod.rs
```diff
@@ -88,6 +88,11 @@ pub use crate::send_stream::{RetransmissionPriority, SendStreamStats, Transmissi
 /// handshake.  This is a hack, but a useful one.
 const EXTRA_INITIALS: usize = 4;
 
+/// Maximum number of dummy datagrams to send before Client Initial packets.
+const MAX_PRE_INIT_PKTS: usize = 1;
+/// Payload size range (in bytes) for pre-Initial packets.
+const PRE_INIT_PAYLOAD_LEN_RANGE: (usize, usize) = (1200, 1400);
```
larseggert, line 94: I like the `noise` terminology you used; suggest to apply it more consistently, hence the rename here. Check if similar renamings make sense elsewhere?

```suggestion
const PRE_INIT_NOISE_LEN_RANGE: Range<usize> = (1200, 1400);
```
Is the 1200-1400 range chosen for any particular reason? If so, might want to add a comment.

neqo-transport/src/connection/mod.rs
```diff
@@ -291,6 +296,7 @@ pub struct Connection {
     release_resumption_token_timer: Option<Instant>,
     conn_params: ConnectionParameters,
     hrtime: hrtime::Handle,
+    pre_initial_pkts: usize,
```
larseggert, line 299:
```suggestion
    pre_initial_noise_pkts: usize,
```

neqo-transport/src/connection/mod.rs
```diff
@@ -351,6 +357,10 @@ impl Connection {
             &mut c.stats.borrow_mut(),
         );
         c.setup_handshake_path(&Rc::new(RefCell::new(path)), now);
+
+        // Generate a random number between 1 and the maximum number of pre-initial packets
+        c.pre_initial_pkts = (random::<1>()[0] as usize % MAX_PRE_INIT_PKTS) + 1;
```
larseggert, line 362:
```suggestion
        c.pre_initial_pkts = (usize::from_ne_bytes(random::<1>()[0]) % MAX_PRE_INIT_PKTS) + 1;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1153,25 @@ impl Connection {
 
         match (&self.state, self.role) {
             (State::Init, Role::Client) => {
+                // Before starting the client, send some noise
+                if self.conn_params.pre_init_noise_enabled() && self.pre_initial_pkts > 0 {
+                    self.pre_initial_pkts -= 1;
```
larseggert, line 1158: Decrement after the packet was generated? Otherwise you decrement but may not send.

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1153,25 @@ impl Connection {
...
+                    if let Some(path) = self.paths.primary() {
+                        let v = random::<1>()[0] as usize;
```
larseggert, line 1160:
```suggestion
                        let v = usize::from_ne_bytes(random::<1>()[0]);
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1153,25 @@ impl Connection {
...
+                        let len = PRE_INIT_PAYLOAD_LEN_RANGE.0 + (v % (PRE_INIT_PAYLOAD_LEN_RANGE.1 - PRE_INIT_PAYLOAD_LEN_RANGE.0 + 1));
```
larseggert, line 1161: Can this be simplified or at least explained better? `Range` may have some helper functions you can use.

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1153,25 @@ impl Connection {
...
+                        let mut dummy_buf: SmallVec<[u8; PRE_INIT_PAYLOAD_LEN_RANGE.1]> = smallvec![0u8; len];
```
larseggert, line 1162: Is a `SmallVec` really gaining us anything here?

larseggert (changes requested):

larseggert:
> And I need to understand what the goals are for this feature.

The first comment describing the commit could use some more content, agreed. (I also invited you to a related Signal chat just now.)

> Added: Also, I'm not seeing tests for this, which is probably why I am unable to make sense of it.

+1 on needing tests.

neqo-transport/src/connection/mod.rs
mxinden: Would a smaller payload, e.g. 10 bytes, do as well?

neqo-transport/src/connection/mod.rs
mxinden: Prefer a `Vec` unless a benchmark proves that a stack allocation is faster.

neqo-transport/src/connection/mod.rs
alizohaib: We chose that range to mimic actual Initial packet sizes. A smaller payload could work, but if it is less than the 1200-byte minimum defined in the RFC, it would be trivial to determine whether the first packet is a Client Initial.

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1155,28 @@ impl Connection {
 
         match (&self.state, self.role) {
             (State::Init, Role::Client) => {
+                // Before starting the client, send some noise
+                if self.pre_initial_noise_pkts > 0 {
+                    if let Some(path) = self.paths.primary() {
+                        // Choose a random payload size between the noise payload length range.
+                        let span = PRE_INIT_NOISE_LEN_RANGE.end() - PRE_INIT_NOISE_LEN_RANGE.start() + 1;
```
larseggert, line 1162:
```suggestion
                        let span = PRE_INIT_NOISE_LEN_RANGE.len() + 1;
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1155,28 @@ impl Connection {
...
+                        let v = usize::from(random::<1>()[0]);
+                        let len  = PRE_INIT_NOISE_LEN_RANGE.start() + (v % span);
+
+                        let mut dummy_buf= vec![0u8; len];
```
larseggert, line 1166:
```suggestion
                        let mut noise = vec![0u8; len];
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1142,6 +1155,28 @@ impl Connection {
 
         match (&self.state, self.role) {
             (State::Init, Role::Client) => {
+                // Before starting the client, send some noise
+                if self.pre_initial_noise_pkts > 0 {
+                    if let Some(path) = self.paths.primary() {
+                        // Choose a random payload size between the noise payload length range.
```
larseggert, line 1161:
```suggestion
                        // Choose a random payload size in the noise payload length range.
```

larseggert (changes requested):
Could I also ask that you run `cargo clippy`, `cargo fmt` and `cargo test` locally before you push, and fix any issues?

alizohaib:
> Could I also ask that you run `cargo clippy`, `cargo fmt` and `cargo test` locally before you push, and fix any issues?

Yes. Sorry about not doing that earlier.

larseggert:
@alizohaib do you intend to pursue this PR?

larseggert:
Making this a draft PR until it becomes reviewable again.

larseggert:
@alizohaib this is starting to bit-rot. Do you intend to see this merged?

larseggert:
Closing this for lack of activity. Please feel free to reopen.

PR #2617, opened 2025-05-08
fix: Eliminate expensive log call

larseggert:
That called `format!` even when the result was never logged.

CC @jrmuizel @mstange


martinthomson (approved):
I think that the original theory was that the compiler would know to remove this check when it wasn't necessary, but I guess it didn't manage to work that out.  The log message is pretty useless anyway.

PR #2618, opened 2025-05-08
fix: Avoid repeated lookups of NSS experimental API functions

larseggert:
CC @jrmuizel @mstange

### DIFF
--- a/neqo-crypto/src/exp.rs
+++ b/neqo-crypto/src/exp.rs
@@ -12,13 +12,19 @@ macro_rules! experimental_api {
         #[allow(clippy::allow_attributes, clippy::missing_safety_doc, reason = "Inherent in macro use.")]
         #[allow(clippy::allow_attributes, clippy::missing_errors_doc, reason = "Inherent in macro use.")]
         pub unsafe fn $n ( $( $a : $t ),* ) -> Result<(), $crate::err::Error> {
-            const EXP_FUNCTION: &str = stringify!($n);
-            let n = ::std::ffi::CString::new(EXP_FUNCTION)?;
-            let f = $crate::ssl::SSL_GetExperimentalAPI(n.as_ptr());
-            if f.is_null() {
+            struct ExperimentalAPI(*mut std::ffi::c_void);
+            unsafe impl Send for ExperimentalAPI {}
+            unsafe impl Sync for ExperimentalAPI {}
+            static EXP_API: ::std::sync::OnceLock<ExperimentalAPI> = ::std::sync::OnceLock::new();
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Something to consider is asking NSS to make the experimental APIs we use permanent fixtures, so that we don't need this at all.  A lot of the functions we originally had as experimental were barely experimental 5 years ago, which means that they probably aren't any more.

There's a bit of trickiness to the migration for those for the rust interface, but as long as `SSL_GetExperimentalAPI` retains knowledge of the string (mapping it to the public function rather than a private one), the migration should be seamless:
1. NSS switches to a public function, retaining the string in `SSL_GetExperimentalAPI` and removing the macro from `sslexp.h`.
2. neqo is updated to include the function in bindings.  neqo updates the minimum NSS version it depends on.
3. At some later date, the string is removed from the list that `SSL_GetExperimentalAPI` uses.

Do you think you could compile a list of functions for which this makes sense to do?

larseggert:
Here are the experimental functions called during the test suite, which should be all:
```
SSL_AeadDecrypt
SSL_AeadEncrypt
SSL_CreateAntiReplayContext
SSL_DestroyAead
SSL_DestroyResumptionTokenInfo
SSL_EncodeEchConfigId
SSL_GetEchRetryConfigs
SSL_GetResumptionTokenInfo
SSL_HelloRetryRequestCallback
SSL_HkdfExpandLabel
SSL_HkdfExpandLabelWithMech
SSL_HkdfExtract
SSL_InstallExtensionHooks
SSL_MakeAead
SSL_PeerCertificateChainDER
SSL_RecordLayerData
SSL_RecordLayerWriteCallback
SSL_ReleaseAntiReplayContext
SSL_SecretCallback
SSL_SendSessionTicket
SSL_SetAntiReplayContext
SSL_SetClientEchConfigs
SSL_SetMaxEarlyDataSize
SSL_SetResumptionToken
SSL_SetResumptionTokenCallback
SSL_SetServerEchConfigs
SSL_SetTimeFunc
```
@dennisjackson @jschanck would you be able to add those to the normal NSS API in the way @martinthomson suggested?

mxinden:
> Do you think you could compile a list of functions for which this makes sense to do?

At the SecPrivNecko work week in Toronto last week we (NSS and Necko team) discussed unifying the various NSS Rust wrappers into one.

Under the assumption that that unification is happening, does it make sense to invest into our current wrapper in the meantime?

martinthomson:
That change would be an asset no matter how we proceed with wrapping NSS.  It allows for direct binding to functions, rather than via the indirection.

PR #2593, opened 2025-04-18
perf: use GSO

mxinden:
Use generic send offloading (GSO) on Linux and UDP segment offloading (USO) on Windows.

GSO and USO allow us to batch multiple datagrams into one large payload (up to 64 KB) and pass it in a single system call to the kernel. The kernel either itself segments it, or has the NIC segment it, before sending it out on the network.

Early measurements show an up to 2x throughput improvement on artificial CPU bound localhost transfer benchmark.

---

Attempt 1: https://github.com/mozilla/neqo/commit/f25b0b77ff579b56a4ea882a3ca70404b3c38b03
Attempt 2: https://github.com/mozilla/neqo/pull/2532/

Compared to attempt 2:
- implements the datagram batching in `neqo-transport` instead of `neqo-bin`
- does not copy each datagram in the larger GSO buffer, but instead writes each into the GSO buffer right away.

Once this is merged, we can switch to a long-lived send buffer (see discussed in https://github.com/mozilla/neqo/issues/2670). https://github.com/mozilla/neqo/pull/2677 and this pull request lay the  groundwork for it.

PR #2620, opened 2025-05-09
fix: Move allocations outside of loops

larseggert:
Let's see if this moves the needle on anything...

### DIFF
--- a/neqo-http3/src/frames/reader.rs
+++ b/neqo-http3/src/frames/reader.rs
@@ -100,7 +100,7 @@ pub struct FrameReader {
     state: FrameReaderState,
     frame_type: HFrameType,
     frame_len: u64,
-    buffer: [u8; 2048],
+    buffer: [u8; MAX_READ_SIZE],
 }
 
 impl Default for FrameReader {
--- a/neqo-http3/src/frames/reader.rs
+++ b/neqo-http3/src/frames/reader.rs
@@ -8,7 +8,7 @@
     clippy::module_name_repetitions,
     reason = "<https://github.com/mozilla/neqo/issues/2284#issuecomment-2782711813>"
 )]
-use std::fmt::Debug;
+use std::{cmp::min, fmt::Debug};
 
 use neqo_common::{
     hex_with_len, qtrace, Decoder, IncrementalDecoderBuffer, IncrementalDecoderIgnore,
@@ -19,7 +19,7 @@ use neqo_transport::{Connection, StreamId};
 use super::hframe::HFrameType;
 use crate::{Error, RecvStream, Res};
 
-const MAX_READ_SIZE: usize = 4096;
+const MAX_READ_SIZE: usize = 2048; // Given a practical MTU of 1500 bytes, this seems reasonable.
 
 pub trait FrameDecoder<T> {
     fn is_known_type(frame_type: HFrameType) -> bool;
@@ -100,6 +100,7 @@ pub struct FrameReader {
     state: FrameReaderState,
     frame_type: HFrameType,
     frame_len: u64,
+    buffer: [u8; 2048],
 }
 
 impl Default for FrameReader {
@@ -117,6 +118,7 @@ impl FrameReader {
             },
             frame_type: HFrameType(u64::MAX),
             frame_len: 0,
+            buffer: [0; MAX_READ_SIZE],
         }
     }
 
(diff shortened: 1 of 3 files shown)


neqo-http3/src/frames/reader.rs
```diff
@@ -165,11 +165,12 @@ impl FrameReader {
         &mut self,
         stream_reader: &mut dyn StreamReader,
     ) -> Res<(Option<T>, bool)> {
+        let mut buf = [0; MAX_READ_SIZE];
```
mxinden, line 168: Given that `MAX_READ_SIZE` is `4096`, is a stack vs heap allocation the right way to go?

mxinden (commented):
Provided we see a benchmark improvement, I think this change is worth doing.

Long-term, I wonder whether we should redesign our stream receive buffer management both in `neqo-transport` and `neqo-http3`. Instead of each layer passing a buffer down to be copied into, I wonder whether we can pass existing buffers from the bottom (`neqo-transport`) to the top (`neqo-http3`) without copying.

After all, `neqo-transport` owns each stream chunk:

https://github.com/mozilla/neqo/blob/8add3f86cd5522ff2ccc51f32de6f515a76b14de/neqo-transport/src/recv_stream.rs#L141

Long-term plan in my mind for the receive path:

1. Long lived UDP receive buffer passed to the OS.
2. In-place decryption  (https://github.com/mozilla/neqo/pull/2385/).
3. Re-allocation of each stream chunk in `neqo-transport` (managed through above linked `data_ranges`).
4. `neqo-http3` immutably reading, or taking ownership of stream chunks, thus no allocation nor copying into long lived buffers.
5. (Potentially) re-allocation in `neqo-glue` to cross FFI boundary.

I will give this more thought. Also taking a look how other implementations solve this. Input welcome.

larseggert:
Yes, we should avoid allocations more, the recent flamegraphs show us spending noticeable time there.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -108,9 +110,10 @@ impl NewStreamHeadReader {
             reader, stream_id, ..
         } = self
         {
+            let mut buf = [0; MAX_READ_SIZE];
```
martinthomson, line 113: Does this buffer need to be this hard if it is just reading off the stream head?

larseggert: Not sure I understand what you mean by "this hard"?

martinthomson: I mistyped.  I should have said large, but my fingers have a mind of their own sometimes.

The reason I asked is that this type only exists to read at most 16 octets from the stream.  4k is quite a bit more than that.  It's entirely possible that reading more is unproductive.

larseggert: Is there any way to derive this upper limit from the type? (Otherwise I'll make it fixed and explain why in a comment.)

neqo-http3/src/frames/reader.rs
mxinden: Maybe. I don't have an intuition on that. In case it is not a clear win, I would prefer doing the above large refactoring instead.

neqo-http3/src/stream_type_reader.rs
martinthomson: A fixed value is fine.  The type doesn't inherently know how much it will need to read, at least not in a way that is easily derived.  I just know that from understanding what it does (which is to read two varints).

neqo-http3/src/stream_type_reader.rs
```diff
@@ -108,9 +110,11 @@ impl NewStreamHeadReader {
             reader, stream_id, ..
         } = self
         {
+            // This type only exists to read at most two varints (= 16 bytes) from the stream.
+            let mut buf = [0; 2 * size_of::<u64>()];
```
martinthomson, line 114:
```suggestion
            let mut buf = [0; 16];
```

varints aren't really `u64`, so why not just write 16 and be done.

martinthomson (approved):
How do I interpret the benchmark changes here?  Some improvements, some regressions.  This is HTTP-specific, so are the improvements implicated there?

larseggert:
> How do I interpret the benchmark changes here? Some improvements, some regressions. This is HTTP-specific, so are the improvements implicated there?

The `multistream` benchmark is a bit flaky (= not stable), so I would ignore those. There are some gains in the uplosd and download microbenchmarks, which I'd take as a signal in the right direction. They don't quite materialize in the demo client and server runs (the benchmark table). My theory is that we really need #2593 to get these to the point where they dependably are CPU-bound, at which point we should see improvements there as well.

PR #2623, opened 2025-05-12
ci: Avoid large pass-by-value instances

larseggert:
This is v2 of this PR, focusing on where it makes really sense.


neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -56,13 +56,13 @@ impl Stream for WebTransportRecvStream {
 
 impl RecvStream for WebTransportRecvStream {
     fn receive(&mut self, _conn: &mut Connection) -> Res<(ReceiveOutput, bool)> {
-        self.events.data_readable(self.get_info());
+        self.events.data_readable(&self.get_info());
```
martinthomson, line 59: This `get_info()` method is creating a new instance of `Http3StreamInfo`.  That seems wasteful.  Any reason why that value can't be owned?  ... so that `get_info()` can return a reference to it.

neqo-http3/src/recv_message.rs
```diff
@@ -211,7 +211,7 @@ impl RecvMessage {
             RecvMessageState::WaitingForData { .. }
             | RecvMessageState::WaitingForFinAfterTrailers { .. } => {
                 if post_readable_event {
-                    self.conn_events.data_readable(self.get_stream_info());
+                    self.conn_events.data_readable(&self.get_stream_info());
```
martinthomson, line 214: Same as above.

neqo-http3/src/server_connection_events.rs
```diff
@@ -68,30 +68,40 @@ pub struct Http3ServerConnEvents {
...
+            stream_info: *stream_info,
+        });
     }
 }
 
 impl RecvStreamEvents for Http3ServerConnEvents {
     /// Add a new `DataReadable` event
-    fn data_readable(&self, stream_info: Http3StreamInfo) {
-        self.insert(Http3ServerConnEvent::DataReadable { stream_info });
+    fn data_readable(&self, stream_info: &Http3StreamInfo) {
```
martinthomson, line 91: This requires a value, so it shouldn't take a reference.

martinthomson (changes requested):
Was this conversion automated?  Because there are a few patterns that I'm seeing that are concerning.  I stopped half way, but hopefully the patterns are clear.

There's some real improvement possible here, but those problems are enough that this isn't a good idea as-is.

.clippy.toml
```diff
@@ -7,3 +7,5 @@ disallowed-macros = [
 allow-mixed-uninlined-format-args = false
 allow-unwrap-in-tests = true
 allow-dbg-in-tests = true
+pass-by-value-size-limit = 16
+# avoid-breaking-exported-api = false # TODO: Enable this in a follow-on PR.
```
larseggert, line 11: See https://github.com/larseggert/neqo/pull/36

larseggert:
@martinthomson maybe I'm misunderstanding something about Rust and this lint. I thought avoiding pass-by-value inherently meant to pass a reference, even if the called function needs a value. If the called function needs a value, it dereferences, which does not involve a stack allocation (but still involved a memory copy).

neqo-http3/src/features/extended_connect/webtransport_streams.rs
larseggert: You mean adding a `stream_info` field, initializing that in `new`, and then returning a ref?

mxinden:
Rust has the convention that the function signature should express what the function implementation needs. Example:

- The function needs an allocated `Vec`, request an owned `Vec`, not a `&Vec` to be cloned later.
- The function needs a `u8`, request an owned `u8`, not a `&u8` that is then dereferenced.

In the context of this pull request, I suggest only changing to a reference if the function only needs a reference.

larseggert:
Ok, I will give this a do-over tomorrow.

martinthomson:
Max got it.  It is my understanding that this moves things as much as that is possible:

```rust
fn do_something(huge: ItsEnormous) -> ItsGargantuan {
  ItsGargantuan { huge, ...Default::default() }
}
```

That's not a firm guarantee, but the compiler can be somewhat smart about these things by using moves for these sorts of things.

larseggert:
I don't think there is a performance angle here, but it does seem like an OK cleanup to me.

martinthomson (approved):

PR #2628, opened 2025-05-14
feat: `clippy::field_scoped_visibility_modifiers`

larseggert:
Do we want this?


martinthomson (approved):
This seems like good hygiene.

PR #2629, opened 2025-05-14
feat: `clippy::impl_trait_in_params`

larseggert:
Do we want this?


martinthomson (commented):
I'm take-it-or-leave-it on these.  It's nice to be consistent and this gets most of the way there, so it's probably worth it.

mxinden (approved):
Not a strong opinion. Fine to merge here.

PR #2631, opened 2025-05-14
fix: `clippy::return_and_then`

larseggert:
Do we want this?

(Also sets `precedence_bits` and `redundant_test_prefix`, but those don't fire. @martinthomson should like the latter...)


martinthomson (approved):
I like ALL of these.  Thanks Lars.

Issue #49, opened 2019-06-13
Current Qpack is a very simple one


ddragana:
the qpack algorithm needs optimization.

martinthomson:
I wouldn't put this at the top of the list for a while.  Simple will suffice until we have more of the stack in place.  I'm sure that when it becomes a performance issue we'll get around to tweaking it.

ddragana:
Yes, this is for later, way later. The issue is just to know that it exists.

larseggert:
@martinthomson is this OBE?

martinthomson:
We could maybe do better.  Basically, Dragana and I decided that a simple strategy would suffice in the short term, but we knew (from talking to others) that a little bit of work could probably improve performance a fair bit.

There's a lot of performance on the table here, particularly in those first few round trips where the congestion window is small.  If we can pack one or two extra requests in, that could be a big deal for page load metrics.  I'd say keep this (or something with a less casual name) on the books.

larseggert:
Related discussion in https://github.com/mozilla/neqo/pull/2632#discussion_r2089914571

PR #2632, opened 2025-05-14
fix: `clippy::iter_over_hash_type`

larseggert:
Do we want this? Also see questions below.


neqo-qpack/src/encoder.rs
```diff
@@ -146,6 +146,10 @@ impl QPackEncoder {
     fn recalculate_blocked_streams(&mut self) {
         let acked_inserts_cnt = self.table.get_acked_inserts_cnt();
         self.blocked_stream_cnt = 0;
+        #[expect(
+            clippy::iter_over_hash_type,
+            reason = "OK to loop over unACKed blocks in an undefined order."
```
larseggert, line 151: Is it though? Would an ordered iteration have advantages? (Also applies to the next two instances of this in this file.)

neqo-crypto/build.rs
```diff
@@ -4,7 +4,11 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![expect(clippy::unwrap_used, reason = " OK in a build script.")]
+#![expect(
+    clippy::unwrap_used,
+    clippy::iter_over_hash_type,
+    reason = " OK in a build script."
```
martinthomson, line 10:
```suggestion
    reason = "OK in a build script."
```

neqo-qpack/src/encoder.rs
martinthomson: Are you suggesting that maybe we could get more efficiency from these if we knew what was most active?  I suspect that headers would be more likely to be changing for more recent streams, but I don't know how to avoid iterating the whole collection for these functions.  We'd need a completely different structure, I guess.  One that only stored blocked streams, in this case.  That might be faster.

What's the data structure?  `BinaryHeap`?

martinthomson (approved):
This seems good.  Though it's always uncomfortable when enabling the new lint only results in suppressions, not changes.

neqo-qpack/src/encoder.rs
larseggert: You thought about this more than I did - it just felt to me that this was the one suppression I added where maybe something else might be done. We have a long-standing bug (#49) on optimizing QPACK, any changes can be done as part of that work if we ever get to it.

PR #2633, opened 2025-05-15
chore: `use std::fmt` more

larseggert:
Instead of qualifying in place.

### DIFF
--- a/test-fixture/src/lib.rs
+++ b/test-fixture/src/lib.rs
@@ -9,7 +9,7 @@
 use std::{
     cell::{OnceCell, RefCell},
     cmp::max,
-    fmt::Display,
+    fmt::{self, Display, Formatter},
     io::{self, Cursor, Result, Write},
     mem,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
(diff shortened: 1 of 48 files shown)


test-fixture/src/lib.rs
```diff
@@ -432,16 +432,16 @@ impl Write for SharedVec {
...
 
 impl Display for SharedVec {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         f.write_str(
             &String::from_utf8(
                 self.buf
                     .lock()
-                    .map_err(|_| std::fmt::Error)?
+                    .map_err(|_| fmt::Error)?
```
martinthomson, line 440: For these, you could instead:

```rust
use std::fmt::Error as FmtError;
```

Not sure where I stand on that one though.

martinthomson (approved):

test-fixture/src/lib.rs
larseggert: Yeah, I thought about that. Not sure either which is better.

PR #2634, opened 2025-05-15
fix: `RecvdPackets` doesn't need a `VecDeque`

larseggert:
A simple `Vec` seems sufficient. Let's see if this has a performance impact.


martinthomson (commented):
The reason this used `VecDeque` was that we expect to be adding at the end and removing from the start continuously.  Removal from the start of a `Vec` means shuffling the whole collection in place, which is O(n).  Removal from a `VecDeque` is O(1).

Let's see what the benchmarks say.

martinthomson:
Those benchmarks are too unstable to make sense of the performance impact.  I find it hard to believe that this would result in a 20% improvement, as one suggests.  Others say 3-4% degradation, which is plausible, but inconclusive.  The 3% improvement to the frame reordering is bogus.

The amount of noise in these numbers makes them completely untrustworthy.  Here's a potentially crazy idea: we run benchmarks for a lot longer on our baseline.  That's daily, not every change.  Then, when a result indicates divergence from that baseline, we can ask to run just that benchmark again to see if that is significant.

larseggert:
I agree that the benchmarks are pretty useless. I spent quite a bit of time over the last few days trying out various things to make them better. I'm beginning to wonder if that machine is getting flaky or if other crazy things are happening, such as insufficient airflow making us hit thermal throttling.

We are getting additional runners (I CC'ed you on RELOPS-1310), maybe they will be more stable. We can then also up the runtime on the various benches; we're pretty blocked for CI time at the moment.

larseggert:
(Oh, but the current bench results are from last week, i.e., before my changes. The latest run failed. Let me re-run.)

larseggert:
This seems to make no difference.

PR #2635, opened 2025-05-15
fix: Collect ranges into a `SmallVec`

larseggert:
Let's see if this reduces memory churn/improves performance in this function.

### DIFF
--- a/neqo-transport/src/tracking.rs
+++ b/neqo-transport/src/tracking.rs
@@ -16,6 +16,7 @@ use enum_map::{Enum, EnumMap};
 use enumset::{EnumSet, EnumSetType};
 use neqo_common::{qdebug, qinfo, qtrace, qwarn, IpTosEcn, MAX_VARINT};
 use neqo_crypto::Epoch;
+use smallvec::SmallVec;
 use strum::{Display, EnumIter};
 
 use crate::{
@@ -448,7 +449,7 @@ impl RecvdPackets {
             .filter(|r| r.ack_needed())
             .take(max_ranges)
             .cloned()
-            .collect::<Vec<_>>();
+            .collect::<SmallVec<[_; MAX_TRACKED_RANGES]>>();
         if ranges.is_empty() {
             return;
         }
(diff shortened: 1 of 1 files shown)


neqo-transport/src/tracking.rs
```diff
@@ -500,7 +501,7 @@ impl RecvdPackets {
 
         tokens.push(RecoveryToken::Ack(AckToken {
             space: self.space,
-            ranges,
+            ranges: ranges.into_vec(),
```
martinthomson, line 504: This could also be a SmallVec, couldn't it?

martinthomson (approved):
Seems like a good plan to me.

neqo-transport/src/tracking.rs
martinthomson: Hmm, it seems like that is a good reason not to use `SmallVec` there.  Either that or do the boxing it suggests.  The net effect would be the same either way. `Vec` is heap allocation, `Box<SmallVec>` is a more complicated way of saying the same thing.  Maybe comment to that effect somewhere.

martinthomson: The boxed slice seems like a good compromise.

neqo-transport/src/stats.rs
```diff
@@ -259,6 +259,9 @@ pub struct Stats {
     /// Acknowledgments for packets that contained data that was marked
     /// for retransmission when the PTO timer popped.
     pub pto_ack: usize,
+    /// Number of times we had to drop an unacknowledged ACK range.
+    // TODO: Export this via glean.
```
mxinden: I don't think the `TODO` is needed, given that you created https://bugzilla.mozilla.org/show_bug.cgi?id=1966894, especially since https://bugzilla.mozilla.org/show_bug.cgi?id=1966894 will be fixed in mozilla-central and not in mozilla/neqo.

mxinden (commented):
Fine by me. Though do we see a clear signal that the `SmallVec` is an improvement?

PR #2637, opened 2025-05-15
ci: Add `cloudflare/quiche` and `aws/s2n-quic` to bench

larseggert:
@mxinden said that @jrmuizel suggested this, so we can compare against other Rust QUIC stacks.

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -120,7 +120,7 @@ runs:
       uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: dist
-        key: nss-${{ inputs.target && inputs.target || runner.os }}-${{ runner.arch }}-${{ inputs.type }}-${{ steps.heads.outputs.nss_head }}-${{ steps.heads.outputs.NSPR_HEAD }}
+        key: nss-${{ inputs.target && inputs.target || runner.os }}-${{ runner.arch }}-${{ inputs.type }}-${{ steps.heads.outputs.nss_head }}-${{ steps.heads.outputs.nspr_head }}
 
     - name: Check if build is needed
       id: check_build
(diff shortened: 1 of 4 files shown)


.github/actions/rust/action.yml
```diff
@@ -22,7 +22,7 @@ runs:
   using: composite
   steps:
     - name: Install Rust
-      uses: dtolnay/rust-toolchain@21dc36fb71dd22e3317045c0c31a3f4249868b17 # master
+      uses: dtolnay/rust-toolchain@master # zizmor: ignore[unpinned-uses]
```
larseggert, line 25: Since I had a lot of time to wait for the CI runs during debugging, I noticed some unrelated issues and fixed them. This is one of them.

larseggert: Dependabot doesn't update SHAs that aren't associated with tags, and `dtolnay/rust-toolchain` doesn't tag any releases, so we were stuck on an old version.

.github/actions/nss/action.yml
```diff
@@ -120,7 +120,7 @@ runs:
       uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: dist
-        key: nss-${{ inputs.target && inputs.target || runner.os }}-${{ runner.arch }}-${{ inputs.type }}-${{ steps.heads.outputs.nss_head }}-${{ steps.heads.outputs.NSPR_HEAD }}
+        key: nss-${{ inputs.target && inputs.target || runner.os }}-${{ runner.arch }}-${{ inputs.type }}-${{ steps.heads.outputs.nss_head }}-${{ steps.heads.outputs.nspr_head }}
```
larseggert, line 123: This is another unrelated change. Not sure if this even matters (are `outputs` case sensitive?) but no harm.

.github/actions/rust/action.yml
mxinden: Preference for a hash, even if it idoesn't auto-update. In my eyes, the more deterministic our CI is the better.

mxinden (approved):
:rocket: thanks!

PR #2638, opened 2025-05-16
feat: make fast-apple-datapath optional, off by default

mxinden:
The `quinn-udp` `fast-apple-datapath` Cargo feature allows Neqo to use the private MacOS multi-message IO functions `sendmsg_x` and `recvmsg_x`.

We have seen issues with `fast-apple-datapath`:

- https://github.com/quinn-rs/quinn/issues/2214
- https://github.com/quinn-rs/quinn/pull/2154

To reduce the risk of the upcoming larger roll out of the [Fast UDP for Firefox project](https://bugzilla.mozilla.org/show_bug.cgi?id=1901292) to Firefox Release, disable `fast-apple-datapath` for now.  Enabling it can then be a separate project afterwards.

---

Objections? I know it is a step backwards from a performance perspective. That said, I would like to prioritize getting the larger _Fast UDP for Firefox_ project onto Firefox Release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ enumset = { version = "1.1", default-features = false }
 hex = { version = "0.4", default-features = false }
 log = { version = "0.4", default-features = false }
 qlog = { version = "0.15.1", default-features = false }
-quinn-udp = { version = "0.5.11", default-features = false, features = ["direct-log", "fast-apple-datapath"] }
+quinn-udp = { version = "0.5.11", default-features = false, features = ["direct-log"] }
 regex = { version = "1.9", default-features = false }
 static_assertions = { version = "1.1", default-features = false }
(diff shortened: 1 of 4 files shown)


larseggert:
I'm OK with enabling this in a controlled way via an experiment. (We will likely need to do something similar for Windows GSO once Microsoft fixes their issue.)

larseggert:
Is it possible to enable/disable this via a pref, rather than via a Rust feature?

mxinden:
As it is implemented in quinn-udp today, it is not possible, i.e. can only be enabled or disabled at compile time. That said, we can make the required changes to quinn-udp in order to make it optional at runtime instead of compiletime.

larseggert (approved):

larseggert:
(I cancelled the bench run, since I am mucking around on the bencher machine.)

PR #2641, opened 2025-05-19
Make hrtime tests more robust

martinthomson:
As the timers get tighter, give them more chances, but make the test pass two times for every failure.

Plus better pretty printing.

For #2422, but not claiming to fix it.

### DIFF
--- a/neqo-common/src/hrtime.rs
+++ b/neqo-common/src/hrtime.rs
@@ -392,8 +392,9 @@ mod test {
 
     use super::Time;
 
-    const ONE: Duration = Duration::from_millis(1);
-    const ONE_AND_A_BIT: Duration = Duration::from_micros(1500);
+    const ONE_MS: Duration = Duration::from_millis(1);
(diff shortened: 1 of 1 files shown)


martinthomson:
The two CI failures here are both bogus.  This is test code only, so coverage going down has no effect, especially when that code exists to deal with instability in runners, like this.  Similarly, this cannot affect benchmarks.  The one regression is pretty significant, but there is no way this code could have produced that much of a change, as it is not compiled in for that benchmark.

larseggert (commented):
Can we maybe do something similar for other tests, so we can take out the `#[cfg(all(target_os = "linux", not(neqo_sanitize)))]` on line 385? Always has been bugging me that this is necessary...

martinthomson:
My experience here is that this is down to a lack of availability of good timer resolution in CI runners, more than anything else.

larseggert (approved):

larseggert:
There are still failures, e.g., https://github.com/mozilla/neqo/actions/runs/15418677161/job/43387638309?pr=2691

PR #2649, opened 2025-05-21
chore: Update `criterion` to 0.6.0

larseggert:


martinthomson (approved):

martinthomson:
Oh, one thing you will need to change before landing is to move to `std::hint::black_box` rather than the alias in criterion.

PR #2650, opened 2025-05-21
chore: New nightly clippy lints & remove unneeded suppressions

larseggert:
Remove a bunch of no-longer-needed clippy nightly bug suppressions, and fix some others.

### DIFF
--- a/neqo-crypto/src/p11.rs
+++ b/neqo-crypto/src/p11.rs
@@ -47,7 +47,11 @@ macro_rules! scoped_ptr {
             /// # Errors
             ///
             /// When passed a null pointer generates an error.
-            #[allow(clippy::allow_attributes, dead_code, reason = "False positive.")]
+            #[allow(
+                clippy::allow_attributes,
+                dead_code,
+                reason = "False positive; is used in code calling the macro."
+            )]
             pub fn from_ptr(ptr: *mut $target) -> Result<Self, $crate::err::Error> {
                 if ptr.is_null() {
                     Err($crate::err::Error::last_nss_error())
--- a/neqo-http3/src/frames/hframe.rs
+++ b/neqo-http3/src/frames/hframe.rs
@@ -150,7 +150,7 @@ impl HFrame {
                 update_frame.encode_varint(*element_id);
 
                 let mut priority_enc: Vec<u8> = Vec::new();
-                write!(priority_enc, "{priority}").unwrap();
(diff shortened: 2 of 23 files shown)


neqo-crypto/src/p11.rs
```diff
@@ -47,7 +47,11 @@ macro_rules! scoped_ptr {
             /// # Errors
             ///
             /// When passed a null pointer generates an error.
-            #[allow(clippy::allow_attributes, dead_code, reason = "False positive.")]
+            #[allow(
```
martinthomson, line 50:
```suggestion
            #[expect(
```

neqo-http3/src/frames/hframe.rs
```diff
@@ -150,7 +150,7 @@ impl HFrame {
                 update_frame.encode_varint(*element_id);
 
                 let mut priority_enc: Vec<u8> = Vec::new();
-                write!(priority_enc, "{priority}").unwrap();
+                write!(priority_enc, "{priority}").expect("write OK");
```
martinthomson, line 153: This might be a good case for [`encode_vvec_with`](https://github.com/mozilla/neqo/blob/bcb89cddc68474f73947d65a823cb0458e0fc682/neqo-common/src/codec.rs#L387)

Making a temporary buffer is pretty heavyweight.  Implementing `Write` for `Encoder` is more work, but it's neater and pretty easy.

neqo-crypto/src/p11.rs
larseggert: If I change it to `expect`, I get `warning: this lint expectation is unfulfilled` inside the macro :-(

martinthomson: image

There was another one, but that might be the same.

neqo-http3/src/frames/hframe.rs
larseggert: I opened https://github.com/mozilla/neqo/issues/2652. Suggest to merge this one as-is, then change it there.

(We do seem to do a lot of `Vec::new` in the http3 code...)

neqo-crypto/src/p11.rs
larseggert: Yeah, there are a bunch. Basically, everywhere I used `clippy::allow_attributes` it's because we can't simply do `expect` for some reason or other.

martinthomson (approved):

PR #2651, opened 2025-05-21
ci: Don't run baseline bench if cache exists

larseggert:
Because as long as a cache entry exists, GitHub doesn't allow it to be overwritten, so why run at all in that case.

Also check more frequently if we need to re-create the baseline bench cache.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -3,8 +3,8 @@ on:
   workflow_call:
   workflow_dispatch:
   schedule:
-    # Run at minute 0 past every 8th hour, so there is a `main`-branch baseline in the cache.
-    - cron: '0 */8 * * *'
+    # Run at minute 0 past every 4th hour, so there is a `main`-branch baseline in the cache.
+    - cron: '0 */4 * * *'
 env:
   CARGO_PROFILE_BENCH_BUILD_OVERRIDE_DEBUG: true
   CARGO_PROFILE_RELEASE_DEBUG: true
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -38,7 +38,18 @@ jobs:
...
+        uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: |
+            neqo/target/criterion
+            hyperfine
+          key: bench-results-${{ runner.name }}-${{ github.sha }}
+          restore-keys: bench-results-${{ runner.name }}-
+
       - name: Checkout microsoft/msquic
+        if: ${{ github.ref != 'refs/heads/main' || steps.cache.outputs.cache-hit == 'false' }}
```
martinthomson, line 52: Is there a way to pack this check into a variable, rather than repeat it?

martinthomson (approved):

.github/workflows/bench.yml
larseggert: I don't think so. There also isn't a way to make a workflow exit early w/success.

martinthomson: A variable might be easier.  I was thinking:

```yaml
  - step:
    id: has
    run: |
      echo "cache=${{ github.ref != 'refs/heads/main' || steps.cache.outputs.cache-hit == 'false' }}" >>$GITHUB_OUTPUT

# ...
    if: ${{ steps.has.cache }}
```

PR #2653, opened 2025-05-21
feat: Implement `Write` for `Encoder`

### LINKED ISSUE #2652: Implementing `Write` for `Encoder`
> This might be a good case for [`encode_vvec_with`](https://github.com/mozilla/neqo/blob/bcb89cddc68474f73947d65a823cb0458e0fc682/neqo-common/src/codec.rs#L387)
>
> Making a temporary buffer is pretty heavyweight.  Implementing `Write` for `Encoder` is more work, but it's neater and pretty easy.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2650#discussion_r2099334691_

larseggert:
And use it.

Fixes #2652

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -4,7 +4,10 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::{fmt::{self, Debug, Formatter}, io::Write};
+use std::{
+    fmt::{self, Debug, Formatter},
+    io::{self, Write},
+};
 
 use crate::hex_with_len;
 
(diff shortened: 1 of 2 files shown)


neqo-http3/src/frames/hframe.rs
```diff
@@ -146,15 +146,12 @@ impl HFrame {
...
-                write!(priority_enc, "{priority}").unwrap();
-
-                update_frame.encode(&priority_enc);
-                enc.encode_varint(update_frame.len() as u64);
-                enc.encode(update_frame.as_ref());
+                enc.encode_vvec_with(|enc_inner| {
+                    enc_inner.encode_varint(*element_id);
+                    enc_inner
+                        .write_all(format!("{priority}").as_bytes())
+                        .expect("write OK");
```
martinthomson, line 153: I mean std::fmt::Write.  Then you could do this:

```suggestion
                    format!{enc_inner, "{priority}").expect("write OK");
```

neqo-common/src/codec.rs
```diff
@@ -472,6 +475,17 @@ impl From<Encoder> for Vec<u8> {
     }
 }
 
+impl Write for Encoder {
```
martinthomson, line 478: `std::fmt::Write` has just one method to implement: `fn write_str(&mut self, s: &str) -> Result<(), Error>`

martinthomson (changes requested):

martinthomson (approved):
Wow, that's less code overall, when you get right down to it.  Nice.

PR #2654, opened 2025-05-21
fix: Eliminate some variables that are only used for logging

larseggert:
And some functions that did nothing but log.

### DIFF
--- a/test-fixture/src/sim/mod.rs
+++ b/test-fixture/src/sim/mod.rs
@@ -260,13 +260,12 @@ impl Simulator {
             n.init(Rc::clone(&self.rng), start);
         }
 
-        let setup_start = Instant::now();
         let now = self.process_loop(start, start);
-        let setup_time = now - start;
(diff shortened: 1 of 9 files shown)


test-fixture/src/sim/mod.rs
```diff
@@ -260,13 +260,12 @@ impl Simulator {
...
         }
 
-        let setup_start = Instant::now();
         let now = self.process_loop(start, start);
-        let setup_time = now - start;
         qinfo!(
             "{t}: Setup took {wall:?} (wall) {setup_time:?} (simulated)",
             t = self.name,
-            wall = setup_start.elapsed(),
+            wall = Instant::now().elapsed(),
```
martinthomson, line 267: This one doesn't work.  Same for the next `Instant::now().elapsed()` call.

martinthomson (approved):
Mostly good, but check your use of `Instant::now().elapsed()`

PR #2656, opened 2025-05-21
ci: Bump `hyperfine` runs to 100

larseggert:


martinthomson:
Wow, that's a lot of regressions on the QNS side.  I'm going to suggest that the multistream case is always going to be a tough one.  That this is now more stable is interesting.

larseggert:
The QNS mess is orthogonal; https://interop.seemann.io has also been very flaky.

I wonder if the regressions with upping the iterations points to issues with having neqo-server run for longer periods. Are we maybe running into memory pressure, or are we not clearing out closed connections from various data structures?

larseggert:
I changed the PR to also start a fresh server for each `hyperfine` iteration, rather than leaving the same one running for all iterations. Let's see.

.github/workflows/bench.yml
```diff
@@ -316,27 +316,24 @@ jobs:
                   transmogrify "${server_cmd[$server]}" "$cc" "$pacing" "${neqo_flags[$client]}"
                   FILENAME="$client-$server$EXT"
                   SERVER_CMD="$CMD"
-                  # shellcheck disable=SC2086
-                  nice -n -20 setarch --addr-no-randomize cset proc --set=cpu2 --exec ${CMD/ / -- } &
-                  transmogrify "${client_cmd[$client]}" "$cc" "$pacing" "${neqo_flags[$server]}"
-                  # shellcheck disable=SC2086
+                  # pkill only matches on the first 15 characters of the command?!
+                  SERVER_TAG="$(basename "$(echo "${server_cmd[$server]}" | cut -f1 -d' ')" | cut -c1-15)"
```
martinthomson, line 320: You could always use `$&` instead.  It's possible that the long command line is causing this to hit some limit.

martinthomson: Oh, that doesn't work.  The pkill invocation is not made by the shell, but by hyperfine itself.  For that to work, you'd need a wrapper that writes the PID to a file.

mxinden (approved):

larseggert:
@mxinden maybe `neqo-bin/benches/main.rs` should also start a fresh server for each benchmark iteration? (I tried to do this, but my async-Rust-fu isn't good enough.)

larseggert:
No, it doesn't explain the noise on other benches. I am really not understanding why we have such a spread.

larseggert:
> We might just have to change the server port on each iteration, as a previous iteration might not yet have fully cleaned up their socket bind.

Are we not doing SOREUSEPORT or whatever it's called?

PR #2658, opened 2025-05-23
chore: Update `codecov/test-results-action` to 1.1.1

larseggert:
See https://github.com/codecov/test-results-action/issues/109#issuecomment-2903180210


martinthomson:
Here's a thing: maybe we can avoid running benchmarks unless the code changes (including the benchmark workflow file, which might suggest splitting things up more).

PR #2663, opened 2025-05-26
ci: Fix bench skip logic

larseggert:
Support for Boolean logic in `env` statements sucks.


mxinden:
Should be resolved now: https://www.githubstatus.com/

PR #2665, opened 2025-05-26
bench: revert custom benc run length on stable benchmarks

mxinden:
https://github.com/mozilla/neqo/pull/2655 increased the benchmark runtime across all benchmarks. This made benchmarks more stable.

Looking at the results of a recent run, this does not seem to be necessary for smaller benchmarks. Thus this commit reverts the custom setting for some in order to save CI runtime.

https://github.com/mozilla/neqo/actions/runs/15251928213/job/42891415533

---

Minor optimization. Don't feel strongly about it.

### DIFF
--- a/neqo-common/benches/decoder.rs
+++ b/neqo-common/benches/decoder.rs
@@ -6,7 +6,7 @@
 
 #![expect(clippy::unwrap_used, reason = "OK in a bench.")]
 
-use std::{hint::black_box, time::Duration};
+use std::hint::black_box;
 
 use criterion::{criterion_group, criterion_main, Criterion};
 use neqo_common::Decoder;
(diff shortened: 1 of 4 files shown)


mxinden:
For the record, currently our `cargo bench` step takes > 1h:

image

https://github.com/mozilla/neqo/actions/runs/15251928213/job/42891415533

larseggert (commented):
So with this change, some of the benches appear to have become more flaky again. I'm hoping we'll have additional runners soon, which means the long runtime is less of a blocker for other PRs in flight.

larseggert:
> I wonder whether we should run the benchmarks for both the current pull request and `main` on each Action execution. Thoughts @larseggert?

The issue is that GiHub won't ever update a cache entry, it will only create a new one when none is present. We would need to figure out how to use the web API to delete a cache entry. Or we would need to switch to storing bench results as artifacts somehow.

I also have https://bugzilla.mozilla.org/show_bug.cgi?id=1966839 open to investigate using https://bencher.dev/, which would replace much of our custom scripting around this.

mxinden:
Closing here since https://github.com/mozilla/neqo/pull/2682 merged first.

PR #2666, opened 2025-05-27
Split transport parameters for ECH

martinthomson:
This ensures that our outer ClientHello contains only the minimum set of transport parameters that are necessary for the fallback handshake. That includes the version negotiation and connection ID parameters.

Obviously, transport parameters that relate to the use of streams and datagrams are unnecessary here.  We won't be using any of those.

I've excluded a couple of transport parameters that might seem like they are necessary.  Here's the analysis:

* `OriginalDestinationConnectionId` = necessary for Retry
* `IdleTimeout` = unnecessary because we are not going to time out
* `StatelessResetToken` = unnecessary because we aren't going to be around long enough for this to matter
* `MaxUdpPayloadSize` = arguably necessary if we set a lower limit than the default, but we presently cannot
* `InitialMaxData` = only relevant if we receive data, which we won't
* `InitialMaxStreamDataBidiLocal` = only relevant if we receive data, which we won't
* `InitialMaxStreamDataBidiRemote` = only relevant if we receive data, which we won't
* `InitialMaxStreamDataUni` = only relevant if we receive data, which we won't
* `InitialMaxStreamsBidi` = only relevant if we receive data, which we won't
* `InitialMaxStreamsUni` = only relevant if we receive data, which we won't
* `AckDelayExponent` = relevant and included
* `MaxAckDelay` = relevant and included
* `DisableMigration` = irrelevant for a client
* `PreferredAddress` = irrelevant for a client
* `ActiveConnectionIdLimit` = the default is fine
* `InitialSourceConnectionId` = necessary for Retry
* `RetrySourceConnectionId` = necessary for Retry
* `VersionInformation` = necessary for version negotiation
* `GreaseQuicBit` = arguably needed, but we can save space and still use the feature
* `MinAckDelay` = not included, which will disable the frames that thiss enables, which is totally fine
* `MaxDatagramFrameSize` = only relevant if we receive data, which we won't


neqo-crypto/tests/ext.rs
```diff
@@ -6,16 +6,20 @@
...
+    generate_ech_keys, AuthenticationStatus, Client, Error, HandshakeState, Server,
 };
-use test_fixture::fixture_init;
+use test_fixture::{damage_ech_config, fixture_init, now};
 
 mod handshake;
 use crate::handshake::connect;
 
+const ECH_CONFIG_ID: u8 = 7;
+const ECH_PUBLIC_NAME: &str = "public.example";
```
larseggert, line 21: Do we want to make this configurable?

neqo-transport/src/tparams.rs
```diff
@@ -722,10 +733,29 @@ impl TransportParametersHandler {
     pub const fn remote_handshake(&self) -> Option<&TransportParameters> {
         self.remote_handshake.as_ref()
     }
+
+    /// Filter to retain only those transport parameters that are necessary for an outer `ClientHello`.
+    const fn filter_ch_outer(tp: TransportParameterId, _v: Option<&TransportParameter>) -> bool {
```
larseggert, line 738: It might be a good idea to add the analysis from the first comment in the PR here?

larseggert (approved):

neqo-crypto/tests/ext.rs
martinthomson: It's just a test.  Isn't a fixed value enough?

neqo-transport/src/tparams.rs
martinthomson: I was in the process of doing that as you asked :)  kazuho suggested that we also share with the QUIC mailing list, which I'll do as well.

martinthomson:
@larseggert codecov is reporting a huge drop here, but the report itself shows only drops for stuff that this patch doesn't touch.  Is that something we need to look into?  I was about to merge, but it seems sensible to double-check.

larseggert:
CodeCov has been weird lately. The only direct changes I see where something isn't covered is `set_alpn` in `connection/mod.rs` and `enable_ech` in `connection_client.rs`. (Look for the only files on https://app.codecov.io/gh/mozilla/neqo/commit/0ce75ec1fb9d2299c2598e7f458111f0f1509a16 where coverage isn't 100% and then check for the red highlights.)

larseggert:
> I don't know how to interpret the interop test matrix here. Is this something I should be concerned about?

No. The entire QNS tests have been flaky for weeks, also on interop.seeman.io. I have been trying to debug, but it only seems to happen on amd64 platforms, not arm ones.

mxinden (approved):
:+1: For what my review is worth.

Thanks for the write-up on the mailing list.

PR #2669, opened 2025-05-27
ci(.github/rust): replace `quickinstall` with `cargo install` with `--locked`

mxinden:
Previously we would use https://github.com/cargo-bins/cargo-quickinstall to install various Rust based command line tools, e.g. `nextest`.

`quickinstall` does not respect our `Cargo.lock`, nor does it have the option to enable it, e.g. via `--locked`.

The problem with not respecting our `Cargo.lock`:

- The `quickinstall` `nexttest` version uses `backtrace@0.3.75`.
- `backtrace@0.3.75` MSRV is `1.82.0` (https://crates.io/crates/backtrace/0.3.75)
- Our `Cargo.lock` specifies `backtrace@0.3.74`, which `quickinstall` ignores.
- https://github.com/mozilla/neqo/pull/2661/ tries to lower our MSRV to 1.81.0.
- The `check (ubuntu-24.04, 1.81.0, debug)` fails as it is using rustc `1.81.0`, but `backtrace@0.3.75` requires rustc `1.82.0`. https://github.com/mozilla/neqo/actions/runs/15275053671/job/42963962708

This commit removes `quickinstall` and replaces it with a simple `cargo install`. In addition it passes `--locked` to `cargo install`, thus respecting our `Cargo.lock`.

From a quick look at past builds, this does not add significant CI runtime.

### DIFF
--- a/.github/actions/rust/action.yml
+++ b/.github/actions/rust/action.yml
@@ -103,13 +103,6 @@ runs:
       if: ${{ runner.os == 'Windows' }}
       run: rm /usr/bin/link.exe || true
 
-    - name: Install cargo-quickinstall
-      shell: bash
-      if: ${{ inputs.tools != '' }}
-      env:
-        GITHUB_TOKEN: ${{ inputs.token }}
-      run: cargo install --locked cargo-quickinstall
-
     - name: Install Rust tools
       shell: bash
       if: ${{ inputs.tools != '' }}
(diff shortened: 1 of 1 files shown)


larseggert:
Can we use `cargo-binstall` instead of compiling from source?

mxinden:
I don't see how `cargo binstall` would infer the `Cargo.lock` file, i.e. the concrete dependency versions, that were used to build a binary that it fetches.

It does have a `--locked` option, but as far as I can tell, that only respects our `Cargo.lock` on fall-back, i.e. when it can not find a pre-build binary.

https://github.com/cargo-bins/cargo-binstall/blob/268643a6b5ea099f5718ee5cd3ff7dc89a5eb49b/crates/bin/src/args.rs#L329-L333

larseggert (approved):

mxinden:
@larseggert can you point me to the failure? Given [that latest `main` is all green](https://github.com/mozilla/neqo/actions/runs/15308870005/job/43068364023), I assume this is intermittent?

Related: I don't yet understand the benefit of `nextest` over `cargo test`. Are we using it for the output only? Does it provide a significant performance benefit? Is it worth the added complexity? Maybe resurrecting https://github.com/mozilla/neqo/pull/2366/ is an option?

larseggert:
Failure here, for example: https://github.com/mozilla/neqo/actions/runs/15341691060/job/43168984253

 I think they bumped the version after the latest merge on `main`. 

Yes, maybe `nextest` in CI is something we should stop doing, esp. if we need to compile it from source. I'll do a PR.

PR #2673, opened 2025-05-28
ci: Don't cache `main` bench results

### LINKED ISSUE #2672: Don't cache benchmark results
> I am suggesting to not cache benchmark results at all, but do something along the lines of:
>
> 1. `git checkout $CURRENT_PR`
> 2. `cargo bench -- --save-baseline current`
> 3. `git checkout `main`
> 4. `cargo bench -- --save-baseline main`
> 5. compare `current` and `main`

_Originally posted by @mxinden in https://github.com/mozilla/neqo/issues/2665#issuecomment-2912832447_

larseggert:
Regenerate them every time.

Fixes #2672

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -400,9 +400,11 @@ jobs:
                       FORMAT=""
                     fi
                     printf "| %s %s%.1f%s | %s%.1f%%%s |\n" "$SYMBOL" "$FORMAT" "$DELTA" "$FORMAT" "$FORMAT" "$PERCENT" "$FORMAT" >> steps.md
-                  else
+                  elif [ "$client" == "neqo" ] || [ "$server" == "neqo" ]; then
                     echo No cached baseline from main found.
                     echo '| :question: | :question: |' >> steps.md
+                  else
+                    echo '| | |' >> steps.md
                   fi
                 done
               done


larseggert:
Regarding one vs. two checkouts, I was wondering if two separate target directories would avoid recompilation.

.github/workflows/bench.yml
```diff
@@ -103,104 +82,105 @@ jobs:
...
       - name: Run cargo bench
-        if: steps.run.outputs.needed == 'true'
         env:
           NSS_DB_PATH: ${{ github.workspace }}/neqo/test-fixture/db
         run: |
           sudo ip link set dev lo mtu "$MTU"
           mkdir -p binaries
-          cd neqo
+          cd neqo-main
+          rm -rf target/criterion
```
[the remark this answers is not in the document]
larseggert: To make sure there is no old criterion date. We're now checking out with 'clean: false' to avoid recompilations when possible.

mxinden:
> 1-conn/1-100mb-resp/mtu-1504 (aka. Download)/client: 💚 Performance has improved.
> 
>        time:   [625.39 ms 625.96 ms 626.57 ms]
>        thrpt:  [159.60 MiB/s 159.76 MiB/s 159.90 MiB/s]
> change:
>        time:   [−5.9466% −5.5630% −5.1966%] (p = 0.00 < 0.05)
>        thrpt:  [+5.4814% +5.8907% +6.3225%]
> 
> 1 (1.00%) high mild
> 1 (1.00%) high severe

The variance across benchmark iterations is now very low:

>        thrpt:  [159.60 MiB/s 159.76 MiB/s 159.90 MiB/s]

I assume this is due to https://github.com/mozilla/neqo/pull/2681.

This pull request does not change any Rust code. Thus the variance between this pull request and `main` should be close to zero and not +5%.

Thus we still have some unknown noise in here, right?

larseggert:
> Thus we still have some unknown noise in here, right?

Probably. Maybe compare the criterion reports for the two runs of this bench; the detailed performance data might show something?

larseggert:
So there are distinct differences between those two runs:
image

.github/workflows/bench.yml
```diff
@@ -38,32 +34,28 @@ jobs:
...
+
+      - name: Checkout mozilla/neqo@main
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: mozilla/neqo
+          ref: main
+          path: neqo-main
+          submodules: 'recursive'
+          persist-credentials: false
+          clean: false
```
mxinden, line 56: Is this third checkout necessary @larseggert?

Above we checkout `neqo/` and `neqo-main/`. Why checkout `neqo-main/` a second time?

larseggert: Probably a botched rebase, will fix

.github/workflows/bench.yml
```diff
@@ -38,32 +34,28 @@ jobs:
...
-            neqo/target/criterion
-            hyperfine
-          key: bench-results-${{ runner.name }}-${{ github.sha }}
-          restore-keys: bench-results-${{ runner.name }}-
+          repository: mozilla/neqo
+          ref: main
+          path: neqo-main
+          submodules: 'recursive'
+          persist-credentials: false
+          clean: false
```
mxinden, line 45: Why deviate from the default here? In other words, why `clean: false`?

.github/workflows/bench.yml
```diff
@@ -104,104 +93,105 @@ jobs:
...
-        if: steps.run.outputs.needed == 'true'
         env:
           NSS_DB_PATH: ${{ github.workspace }}/neqo/test-fixture/db
         run: |
           sudo ip link set dev lo mtu "$MTU"
           mkdir -p binaries
-          cd neqo
+          cd neqo-main
+          rm -rf target/criterion
+          for BENCH in $BENCHES; do
```
mxinden, line 179: `$BENCHES` contains the benchmark binaries of the PR builds. Sample:

```
➜  neqo git:(main) ✗ cargo bench --profile=dev --locked --workspace --features bench --no-run 2>&1  | tee benches.txt
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
  Executable benches/main.rs (target/debug/deps/main-d713735815617f13)
  Executable benches/decoder.rs (target/debug/deps/decoder-15595b2a50ff4039)
  Executable benches/streams.rs (target/debug/deps/streams-696b71404547e60c)
  Executable benches/min_bandwidth.rs (target/debug/deps/min_bandwidth-fd86000de3dceeb9)
  Executable benches/range_tracker.rs (target/debug/deps/range_tracker-a78947e7c536dd73)
  Executable benches/rx_stream_orderer.rs (target/debug/deps/rx_stream_orderer-06f95e3d033c1cdd)
  Executable benches/sent_packets.rs (target/debug/deps/sent_packets-698f9499d852f3bc)
  Executable benches/transfer.rs (target/debug/deps/transfer-169bd66b648e1c78)
```

Note the hashes at the end.

The binaries of the `main` branch do not necessarily have the same hash.

Was this deliberate @larseggert? If not, I can create a pull request.

larseggert: Ah, true, we want both. PR appreciated!

mxinden: :+1: https://github.com/mozilla/neqo/pull/2691

PR #2674, opened 2025-05-28
ci: Pin `google-quiche`

larseggert:
You'd think a company the size of Google could invest in some CI that would prevent them from constantly breaking their `main` branch...


larseggert:
Not sure where the NSPR/Windows failure is coming from, but not this change.

PR #2675, opened 2025-05-28
chore: Get NSPR via release tarball

larseggert:
I'm giving up on waiting for the sync for https://github.com/nss-dev/nspr to get fixed.


larseggert:
> Wonder if we should do this for NSS, too.

Decided against it. We want the ability to have a neqo PR depend on an unreleased NSS version.

PR #2676, opened 2025-05-28
ci(check): use Rust 1.82 instead of 1.81 on Windows

mxinden:
NSS build fails on Windows with Rust 1.81.0. Firefox does not use Rust 1.81.0 (except for testing, see
https://bugzilla.mozilla.org/show_bug.cgi?id=1968057#c1). Firefox uses Rust 1.82.0. Thus this is not worth fixing. Let's explicitly use Rust 1.82.0 here for now.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -59,6 +59,25 @@ jobs:
           - os: windows-2025
             rust-toolchain: stable
             type: release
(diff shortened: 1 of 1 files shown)


mxinden:
For the record, failure on Windows-2025 with Rust 1.81.0:

https://github.com/mozilla/neqo/actions/runs/15295143638/job/43022375602

Issue #2054, opened 2024-08-17
Make an encoder that does not require ownership of the underlying buffer


dlrobertson:
See [neqo-transport/src/tparams.rs:690](https://github.com/mozilla/neqo/blob/165fff6c11f1c863ec0b2ae978dd572f65155383/neqo-transport/src/tparams.rs#L690):

```rust
    // TODO: Modify to avoid a copy.
    let mut enc = Encoder::default();
    self.local.encode(&mut enc);
    assert!(enc.len() <= d.len());
    d[..enc.len()].copy_from_slice(enc.as_ref());
```

It should be possible to create a codec that doesn't require ownership of the underlying buffer. My underlying long-term goal would be something akin to what we had for [smoltcp](https://github.com/smoltcp-rs/smoltcp), where the packet structure was able to abstract over mutability (e.g. [the ipv6 Packet structure](https://github.com/smoltcp-rs/smoltcp/blob/8e3ea5c7f09a76f0a4988fda20cadc74eacdc0d8/src/wire/ipv6.rs#L537)).

dlrobertson:
This TODO has bugged me for a while and I have a long flight with a good bit of time in the airport today 😆 so I'll make a first pass at doing this... Comments and critiques on this are very much appreciated.

mxinden:
:+1: we will need an `Encoder` encoding into an externally managed buffer (send buffer) as part of https://github.com/mozilla/neqo/issues/1693 as well.

dlrobertson:
### TL;DR

I let my enthusiasm get the better of me 😆 The current usage of the `Encoder` structure doesn't really play well with a backing buffer that is not growable.

### Problem

There are some cases where the `Encoder`'s `capacity` is treated as a limit (e.g. the [neqo-thransport PacketBuilder](https://github.com/mozilla/neqo/blob/main/neqo-transport/src/packet/mod.rs#L121)). I think cases like this could play nice with a backing buffer of a fixed size with a managed offset. There are however many locations where this doesn't work so well, so I think it would be a significant lift to get a strict "[smoltcp]-like" packet structure.

### Possible solution

[...]

Issue #2670, opened 2025-05-27
Consider not pre-allocating each UDP datagram in `output_path`


mxinden:
When generating a new UDP datagram, we currently pre-allocate the maximum datagram size upfront:

https://github.com/mozilla/neqo/blob/5e6e7baa1d6ea169c67ca5a1fa69d7d392b07146/neqo-transport/src/connection/mod.rs#L2449-L2451

For small datagrams, this is wasteful.

How wasteful? Not too much according to our requests-per-second (i.e. RPS) benchmark and my local machine.

``` diff
diff --git a/neqo-transport/src/connection/mod.rs b/neqo-transport/src/connection/mod.rs
index 929982c1..d8fb5403 100644
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2448,7 +2448,7 @@ impl Connection {
 
         // Frames for different epochs must go in different packets, but then these
         // packets can go in a single datagram
-        let mut encoder = Encoder::with_capacity(profile.limit());
+        let mut encoder = Encoder::new();
```

```
➜  neqo-bin git:(main) ✗ critcmp main no-pre-alloc
group                                               main                                   no-pre-alloc
-----                                               ----                                   ------------
1-conn/10_000-parallel-1b-resp (aka. RPS)/client    1.01    189.4±4.91ms 51.6 KElem/sec    1.00    188.0±4.12ms 51.9 KElem/sec
```

That said, this might not represent reality well.

Also note that it is not as easy as simply applying the above patch, given that the `Encoder` capacity is implicitly used, e.g. in `PacketBuilder::infer_capacity`:

https://github.com/mozilla/neqo/blob/5e6e7baa1d6ea169c67ca5a1fa69d7d392b07146/neqo-transport/src/packet/mod.rs#L154-L160

martinthomson:
Isn't there a risk that reallocation will cost more than overallocation?  Those numbers are very much inconclusive.

larseggert:
Are we really over-allocating here? Isn't `profile.limit()` limited by the MTU?

larseggert:
I think I'd prefer using once-allocated "static" buffers here, and changing Encoder to encode into those rather than owning the memory.

martinthomson:
[...]

PR #2677, opened 2025-05-29
perf(common): make Encoder generic over borrowed or owned buffer

mxinden:
This allows users of `Encoder` to encode into existing buffers, thus reducing memory allocations and copying.

Example

Previously:

``` rust
let some_existing_buffer = vec![];
// ...
let mut encoder = Encoder::new(); // allocating a new temporary Vec
encoder.encode_varint(42); // writing into the temporary Vec
some_existing_buffer.extend_from_slice(&encoder); // copying latter into former.
```

Now:

``` rust
let some_existing_buffer = vec![];
// ...
let mut encoder = Encoder::new_borrowed_vec(&mut some_existing_buffer);
encoder.encode_varint(42); // writing directly into some_existing_buffer
```

---

Fixes https://github.com/mozilla/neqo/issues/2054.
Discussed in https://github.com/mozilla/neqo/issues/2670.
Simplifies https://github.com/mozilla/neqo/pull/2593.

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -469,15 +469,15 @@ impl Debug for Encoder {
     }
 }
 
-impl AsRef<[u8]> for Encoder {
+impl<B: Buffer> AsRef<[u8]> for Encoder<B> {
     fn as_ref(&self) -> &[u8] {
-        self.buf.as_ref()
+        &self.buf.as_slice()[self.start..]
     }
 }
 
-impl AsMut<[u8]> for Encoder {
+impl<B: Buffer> AsMut<[u8]> for Encoder<B> {
     fn as_mut(&mut self) -> &mut [u8] {
-        self.buf.as_mut()
+        &mut self.buf.as_mut()[self.start..]
     }
 }
 
@@ -550,6 +550,8 @@ pub trait Buffer: io::Write {
 
     fn as_slice(&self) -> &[u8];
 
+    fn as_mut(&mut self) -> &mut [u8];
+
     // Functions needed for `Encoder::encode_vvec_with` and `Encoder::encode_vec_with`.
 
     fn write_zeroes(&mut self, n: usize);
@@ -566,6 +568,10 @@ impl Buffer for Vec<u8> {
 
     fn as_slice(&self) -> &[u8] {
         self.as_ref()
+    }
+
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.as_mut_slice()
     }
 
     fn write_zeroes(&mut self, n: usize) {
@@ -590,6 +596,10 @@ impl Buffer for &mut Vec<u8> {
         self.as_ref()
     }
 
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.as_mut_slice()
+    }
+
     fn write_zeroes(&mut self, n: usize) {
         self.resize(self.len() + n, 0);
     }
@@ -610,6 +620,10 @@ impl Buffer for Cursor<&mut [u8]> {
 
     fn as_slice(&self) -> &[u8] {
         self.get_ref()
+    }
+
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.get_mut()
     }
 
     fn write_zeroes(&mut self, n: usize) {
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -203,27 +203,32 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
 #[derive(Clone, PartialEq, Eq)]
 pub struct Encoder<B = Vec<u8>> {
     buf: B,
+    /// Tracks the starting position of the buffer when the [`Encoder`] is created.
+    /// This allows distinguishing between bytes that existed in the buffer before
+    /// encoding began and those written by the [`Encoder`] itself.
+    start: usize,
 }
 
 impl<B: Buffer> Encoder<B> {
-    /// Get the length of the underlying buffer: the number of bytes that have
-    /// been written to the buffer.
+    /// Get the length of the [`Encoder`].
+    ///
+    /// Note that the length of the underlying buffer might be larger.
     #[must_use]
     pub fn len(&self) -> usize {
-        self.buf.len()
+        self.buf.len() - self.start
     }
 
     /// Returns true if the encoder buffer contains no elements.
     #[must_use]
     pub fn is_empty(&self) -> bool {
-        self.buf.is_empty()
+        self.len() == 0
     }
 
     /// Create a view of the current contents of the buffer.
     /// Note: for a view of a slice, use `Decoder::new(&enc[s..e])`
     #[must_use]
     pub fn as_decoder(&self) -> Decoder<'_> {
-        Decoder::new(self.buf.as_slice())
+        Decoder::new(&self.buf.as_slice())
     }
 
     /// Generic encode routine for arbitrary data.
@@ -409,6 +414,7 @@ impl Encoder<Vec<u8>> {
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
             buf: Vec::with_capacity(capacity),
+            start: 0,
         }
     }
 
@@ -439,17 +445,21 @@ impl Encoder<Vec<u8>> {
         self.buf.truncate(len);
     }
 
-    /// Pad the buffer to `len` with bytes set to `v`.
+    /// Pad the [`Encoder`] to `len` with bytes set to `v`.
     pub fn pad_to(&mut self, len: usize, v: u8) {
-        if len > self.buf.len() {
-            self.buf.resize(len, v);
+        let buffer_len = len + self.start;
+        if buffer_len > self.buf.len() {
+            self.buf.resize(buffer_len, v);
         }
     }
 }
 
 impl Default for Encoder {
     fn default() -> Self {
-        Self { buf: Vec::new() }
+        Self {
+            buf: Vec::new(),
+            start: 0,
+        }
     }
 }
 
@@ -481,6 +491,7 @@ impl From<&[u8]> for Encoder {
     fn from(buf: &[u8]) -> Self {
         Self {
             buf: Vec::from(buf),
+            start: 0,
         }
     }
 }
@@ -515,6 +526,7 @@ impl<'a> Encoder<Cursor<&'a mut [u8]>> {
     pub fn new_borrowed_slice(buf: &'a mut [u8]) -> Self {
         Encoder {
             buf: Cursor::new(buf),
+            start: 0,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


neqo-common/src/codec.rs
```diff
@@ -197,12 +202,120 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
...
+    }
+
+    /// Encode a vector with a varint length using a closure.
+    ///
+    /// # Panics
+    ///
+    /// When `f()` writes more than 2^62 bytes.
+    ///
+    /// TODO: Worse performance than previous implementation. Worth optimizing?
+    pub fn encode_vvec_with<F: FnOnce(&mut Encoder<&mut SmallVec<[u8; 128]>>)>(
```
martinthomson, line 298: I'd say.  Is the problem that you need to set a type for the argument to `f`?

neqo-common/src/codec.rs
```diff
@@ -197,12 +202,120 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
...
+    buf: B,
+}
+
+impl Default for Encoder {
+    fn default() -> Self {
+        Self { buf: Vec::new() }
+    }
+}
+
+impl<B: std::io::Write> Encoder<B> {
```
martinthomson, line 216: Rather than rely on `io::Write`, can you instead use `AsMut<[u8]>` as the target type of `B`?  That would remove all of the concerns about writing and restore all the previous concerns about indexing.  But it would make some of these functions easier to copy over.  You'd need to track the write position, so it's a smidge more memory for the `Vec` case, but I think that's acceptable on balance.

neqo-common/src/codec.rs
```diff
@@ -197,12 +202,120 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
...
+            () => panic!("Varint value too large"),
+        };
+        self
+    }
+
+    /// Encode a vector in TLS style.
+    ///
+    /// # Panics
+    ///
+    /// When `v` is longer than 2^64.
```
martinthomson, line 272:
```suggestion
    /// When `v` is longer than 2^`n`.
```

neqo-common/src/codec.rs
```diff
@@ -197,12 +202,120 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
...
+            u64::try_from(v.as_ref().len()).expect("v is longer than 2^64"),
+        )
+        .encode(v)
+    }
+
+    /// Encode a vector with a varint length.
+    ///
+    /// # Panics
+    ///
+    /// When `v` is longer than 2^64.
```
martinthomson, line 285:
```suggestion
    /// When `v` is longer than 2^62.
```

neqo-common/src/codec.rs
mxinden: I chose `io::Write` instead of `AsMut<[u8]>` for the `Encoder<Vec<u8>>` case, where the `Vec<u8>` is empty at first, e.g. due to the caller not knowing how much memory is needed. Note that this is the most common case across all of Neqo today.

**[`Vec<u8>`'s `io::Write`](https://doc.rust-lang.org/src/std/io/impls.rs.html#503-506) implementation will grow the `Vec`. `AsMut<[u8]>` can not.**

`encode_vvec_with` implementation is the only function that really changes, i.e. is affected by the above. I still need to do more benchmarks. Intuitively it shouldn't come with a performance hit. E.g. none of our unit tests write more than 128 bytes, i.e. the `SmallVec` never spills. Where ever this does have a performance impact, we can simply replace `encode_vvec_with` with the corresponding manual steps.

Alternatively I can keep the old `encode_vvec_with` implementation on `Encoder<Vec<u8>>` and replace the usages of `encode_vvec_with` on all non-`Encoder<Vec<u8>>` `Encoder`s, which aren't many.

We could as well introduce our own custom trait combining `Write` and `AsMut`. Thus far I didn't think it is worth the complexity.

With all the above in mind, can you think of a better way forward @martinthomson?

mxinden: Friendly ping @martinthomson.

neqo-common/src/codec.rs
mxinden: Problem described in detail in https://github.com/mozilla/neqo/pull/2677/files#r2115271882.

neqo-common/src/codec.rs
mxinden: Simply copied from previous implementation. Will fix in a separate pull request.

neqo-common/src/codec.rs
mxinden: Simply copied from previous implementation. Will fix in a separate pull request.

neqo-common/src/codec.rs
martinthomson: Sorry about that.  My view is that we could use a trait and multiple implementations if you care about enabling both growing in two different ways.

neqo-common/src/codec.rs
mxinden: Given this simply being copied from `main`, I propose we fix this in a separate pull request.

https://github.com/mozilla/neqo/pull/2735

neqo-common/src/codec.rs
mxinden: Given this simply being copied from `main`, I propose we fix this in a separate pull request.

https://github.com/mozilla/neqo/pull/2735

neqo-common/src/codec.rs
mxinden: Thanks for the input. I updated the pull request to use a trait and multiple implementations.

neqo-common/src/codec.rs
mxinden: Resolved given the latest updates. See https://github.com/mozilla/neqo/pull/2677#discussion_r2145100660.

neqo-common/src/codec.rs
```diff
@@ -197,59 +200,12 @@ impl<'b> PartialEq<Decoder<'b>> for Decoder<'_> {
...
-            buf: Vec::with_capacity(capacity),
-        }
-    }
-
-    /// Get the capacity of the underlying buffer: the number of bytes that can be
-    /// written without causing an allocation to occur.
-    #[must_use]
-    pub fn capacity(&self) -> usize {
-        self.buf.capacity()
-    }
```
mxinden: These just moved down, as they need to be implemented on `Encode<Vec<u8>>` and not any `Encode<B>`.

neqo-common/src/codec.rs
```diff
@@ -267,40 +223,30 @@ impl Encoder {
...
 
-    /// Don't use this except in testing.
+    /// Generic encode routine for arbitrary data.
     ///
     /// # Panics
     ///
-    /// When `s` contains non-hex values or an odd number of values.
-    #[cfg(any(test, feature = "test-fixture"))]
-    #[must_use]
-    pub fn from_hex<A: AsRef<str>>(s: A) -> Self {
```
mxinden, line 280: Same here, moved down.

neqo-http3/src/buffered_send_stream.rs
```diff
@@ -46,14 +47,22 @@ impl BufferedStream {
         };
     }
 
+    pub fn encode_with<F: FnOnce(&mut Encoder<&mut Vec<u8>>)>(&mut self, f: F) {
+        if let Self::Initialized { buf, .. } = self {
+            f(&mut Encoder::new_borrowed_vec(buf));
+        } else {
+            debug_assert!(false, "Do not encode data before the stream is initialized");
```
mxinden, line 54: Same as in `fn buffer` below.

mxinden:
@martinthomson can you take another look?

neqo-common/src/codec.rs
```diff
@@ -472,13 +491,129 @@ impl From<Encoder> for Vec<u8> {
...
+    fn len(&self) -> usize {
+        usize::try_from(self.position()).expect("memory allocation not to exceed usize")
+    }
+
+    fn write_zeroes(&mut self, n: usize) {
+        let start = usize::try_from(self.position()).expect("Buffer length does not exceed usize");
+        let end = start + n;
+
+        self.get_mut()[start..end].fill(0);
+        self.set_position(end as u64);
```
martinthomson: Does this need:
```suggestion
        self.set_position(u64::try_from(end).expect("Position cannot exceed u64"));
```

neqo-common/src/codec.rs
```diff
@@ -472,13 +491,129 @@ impl From<Encoder> for Vec<u8> {
...
+        self[pos] = data;
+    }
+
+    fn rotate_right(&mut self, start: usize, count: usize) {
+        self[start..].rotate_right(count);
+    }
+
+    fn as_slice(&self) -> &[u8] {
+        self.as_ref()
+    }
```
martinthomson, line 603: Should the order match the trait shape?

neqo-common/src/codec.rs
```diff
@@ -472,13 +491,129 @@ impl From<Encoder> for Vec<u8> {
...
+
+    fn as_slice(&self) -> &[u8];
+
+    // Functions needed for `Encoder::encode_vvec_with` and `Encoder::encode_vec_with`.
+
+    fn write_zeroes(&mut self, n: usize);
+
+    fn write_at(&mut self, pos: usize, data: u8);
+
+    fn rotate_right(&mut self, start: usize, count: usize);
```
martinthomson, line 567: You could require `impl AsMut<[u8]>` and then provide default implementations of these.

neqo-common/src/codec.rs
```diff
@@ -472,13 +491,129 @@ impl From<Encoder> for Vec<u8> {
...
+}
+
+pub trait Buffer: io::Write {
+    fn len(&self) -> usize;
+
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    fn as_slice(&self) -> &[u8];
```
martinthomson, line 557: Maybe require `impl AsRef<[u8]>`?

martinthomson (approved):

neqo-common/src/codec.rs
mxinden: Unfortunately `Cursor<[T]>` does not implement `AsRef<[T]>` and `AsMut<[T]>`. In addition, given that we only have a `Buffer` implementation for `Vec<u8>`, `&mut Vec<u8>` and `&mut [u8]`, the first two and the last requiring different `rotate_right` implementations, the default implementation would not be used.

neqo-common/src/codec.rs
mxinden: Thanks.



https://github.com/mozilla/neqo/pull/2677/commits/8b5e8c299c0bf2b6b9d6fd3ca3d488f778760a84

neqo-common/src/codec.rs
mxinden: Ah neat, rustanalyzer can do this.

https://github.com/mozilla/neqo/pull/2677/commits/49088fcda4d5b3326988bfae7e5cafa5f4315681

neqo-common/src/codec.rs
mxinden: See https://github.com/mozilla/neqo/pull/2677#discussion_r2152459508. Unfortunately `Cursor<[T]>` does not implement `AsRef<[T]>`.

PR #2678, opened 2025-05-30
ci: Use bencher.dev

larseggert:
Feed our benchmark data into bencher.dev, to be shown over time at https://bencher.dev/console/projects/neqo.

### DIFF
--- a/.github/workflows/bencher-archive.yml
+++ b/.github/workflows/bencher-archive.yml
@@ -10,7 +10,7 @@ jobs:
     name: Archive closed fork PR branch with Bencher
     runs-on: ubuntu-latest
     env:
-      BENCHER_PROJECT: ${{ github.repository}}
+      BENCHER_PROJECT: neqo
       BENCHER_API_TOKEN: ${{ secrets.BENCHER_API_TOKEN }}
       BENCHER_BRANCH: ${{ github.ref_name}}
     steps:
--- a/.github/workflows/bencher-archive.yml
+++ b/.github/workflows/bencher-archive.yml
@@ -1,5 +1,6 @@
 on:
   pull_request_target:
+    # SAFETY: We are not running any code from the triggering PR, so this is safe.
     types: [closed] # zizmor: ignore[dangerous-triggers]
 
 permissions:
--- a/.github/workflows/bencher-archive.yml
+++ b/.github/workflows/bencher-archive.yml
@@ -6,8 +6,8 @@ permissions:
   contents: read
 
 jobs:
-  archive_fork_pr_branch:
-    name: Archive closed fork PR branch with Bencher
+  bencher_archive:
+    name: Archive bencher.dev PR branch
     runs-on: ubuntu-latest
     env:
       BENCHER_PROJECT: neqo
@@ -17,5 +17,5 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           persist-credentials: false
-      - uses: bencherdev/bencher@v0.5.1 # zizmor: ignore[unpinned-uses] bencherdev/bencher@54d78b95feda17a1b2761d3332c488d37261ca4e #v0.5.1
+      - uses: bencherdev/bencher@v0.5.1 # zizmor: ignore[unpinned-uses]
       - run: bencher archive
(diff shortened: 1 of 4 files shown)


larseggert:
@mxinden would you mind giving this a review? It needs to be merged to `main` to make the two new workflows trigger. I expect there will be a follow-on PR to fix any nits once they start running.

.github/workflows/bencher-archive.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  pull_request_target:
+    types: [closed] # zizmor: ignore[dangerous-triggers]
```
mxinden: Using `pull_request_target` comes with security risks:

> For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

Can you add a comment here why this is safe?

.github/workflows/bencher-archive.yml
```diff
@@ -0,0 +1,21 @@
...
+    env:
+      BENCHER_PROJECT: neqo
+      BENCHER_API_TOKEN: ${{ secrets.BENCHER_API_TOKEN }}
+      BENCHER_BRANCH: ${{ github.ref_name}}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - uses: bencherdev/bencher@v0.5.1 # zizmor: ignore[unpinned-uses]
+      - run: bencher archive
```
mxinden, line 22: Do I understand correctly that this step simply marks the bencher.dev reports for this pull request as archived? Nothing more?

.github/workflows/bencher-upload.yml
```diff
@@ -0,0 +1,105 @@
+name: bencher.dev Upload
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed # zizmor: ignore[dangerous-triggers]
```
mxinden: Can you add a comment why this is safe? Something along the lines of: "we don't execute code produced by the triggering workflow"?

.github/workflows/bencher-archive.yml
larseggert: That's what I think it does. I adapted these from https://bencher.dev/docs/how-to/github-actions/.

thomaseizinger:
> @thomaseizinger I know you are a fan of bencher.dev and added it to [firezone/firezone#5915](https://github.com/firezone/firezone/pull/5915). If you have some time, would you mind taking a peek here?

LGTM, I am not familiar with neqo's benchmark infrastructure so I can't comment on this much. We've not had an issues with our integration.

The only thing I noticed is that you are specifying the threshold config as part of the upload? We have those configured in the backend and are just making reports to bencher. It may be though that the recommended way of doing this has changed recently. I haven't spent much time on our benchmarking infra lately.

larseggert:
> The only thing I noticed is that you are specifying the threshold config as part of the upload?

I took this from https://bencher.dev/docs/how-to/github-actions/. What would be a better way to do this?

thomaseizinger:
> > The only thing I noticed is that you are specifying the threshold config as part of the upload?
> 
> I took this from [bencher.dev/docs/how-to/github-actions](https://bencher.dev/docs/how-to/github-actions/). What would be a better way to do this?

We are just uploading the report: https://github.com/firezone/firezone/blob/48bbb0f9b691f3133a25766d43a94a77681f352c/.github/workflows/ci.yml#L227-L259

The threshold is then just configured in the backend via the UI. I am not sure if this is _better_ though. I don't recall having an option to set the threshold as part of the upload.

thomaseizinger:
> > The only thing I noticed is that you are specifying the threshold config as part of the upload?
> 
> I took this from [bencher.dev/docs/how-to/github-actions](https://bencher.dev/docs/how-to/github-actions/). What would be a better way to do this?

I see that you are using `bencher run`. We run our perf tests separately because we use iperf3 and that isn't supported by bencher. So that might be the reason why your setup is different.

PR #2679, opened 2025-05-30
ci: Drop `nextest`

larseggert:
Use `cargo test`.


larseggert:
@mxinden I'd like to revert this. The `junit-test` replacement that I found compiles and executes the entire test suite again, which really makes CI times longer. Objections?

PR #2681, opened 2025-05-30
bench(bin): spawn new server for each iteration

mxinden:
Spawn a new `neqo-server` for each benchmark iteration.

---

We see `neqo-bin` benchmark performance degrading over time from criterion iteration to iteration. This is especially obvious now that we have https://github.com/mozilla/neqo/pull/2655.

Without this patch I can reproduce the degradation locally, looking at the timing of each iteration (see upwards trend below).

Example:

image

With this patch, timing across iterations is stable, with no noticeable increase from first to last.

Still needs cleanups, but proofs the point.

Shout-out to @larseggert who discovered the degradation and actually proposed this fix in the past.

### DIFF
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -6,7 +6,9 @@
 
 #![expect(clippy::unwrap_used, reason = "OK in a bench.")]
 
-use std::{env, hint::black_box, path::PathBuf, str::FromStr as _, time::Duration};
+use std::{
+    env, hint::black_box, net::SocketAddr, path::PathBuf, str::FromStr as _, time::Duration,
+};
 
 use criterion::{criterion_group, criterion_main, BatchSize, Criterion, Throughput};
 use neqo_bin::{client, server};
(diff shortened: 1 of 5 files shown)


mxinden:
> 1-conn/1-100mb-resp/mtu-1504 (aka. Download)/client: 💚 Performance has improved.
> 
>        time:   [635.37 ms 636.08 ms 636.81 ms]
>        thrpt:  [157.03 MiB/s 157.21 MiB/s 157.39 MiB/s]



> 1-conn/1-100mb-req/mtu-1504 (aka. Upload)/client: 💚 Performance has improved.
> 
>        time:   [645.98 ms 647.23 ms 648.47 ms]
>        thrpt:  [154.21 MiB/s 154.50 MiB/s 154.80 MiB/s]
> change:
>        time:   [−91.158% −91.133% −91.108%] (p = 0.00 < 0.05)
>        thrpt:  [+1024.6% +1027.8% +1030.9%]

Oh, I only now notice that this pull request, while reducing the variance, also brought our Upload benchmark speed up to par with our Download benchmark speed.

//CC @larseggert since we discussed the discrepancy before.

PR #2683, opened 2025-05-30
ci: Inexact bench cache hits are OK

larseggert:
Probably redundant with #2673.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -51,7 +51,8 @@ jobs:
       - id: run
         env:
           IS_MAIN: ${{ github.ref == 'refs/heads/main' }}
-          CACHE_HIT: ${{ steps.cache.outputs.cache-hit == 'true' }}
+          # See https://github.com/actions/cache#outputs; we are OK with inexact hits.
+          CACHE_HIT: ${{ steps.cache.outputs.cache-hit != '' }}
         run: |
           # Always run benchmarks on PRs or when there isn't a cache hit.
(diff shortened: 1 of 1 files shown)


mxinden:
Interesting. Curious what the benchmark runner reports.

Do I understand correctly that shorter runtime is the only benefit over https://github.com/mozilla/neqo/pull/2673?

larseggert:
Yeah this mostly reduces the runtime of the scheduled run, which we completely eliminate with the other PR.

mxinden:
I was referring to https://github.com/mozilla/neqo/pull/2673 running both `main` and PR benchmarks, i.e. thus having a longer runtime.

But of course, you are right as well that we no longer need the scheduled run.

mxinden:
I am a bit confused by this change.

If I understand correctly, our goal is to compare each pull request to the most recent `main`. If we compare a pull request to an old main, the diff will be more than the pull request itself, i.e. the diff will contain other pull requests.

>           # See https://github.com/actions/cache#outputs; we are OK with inexact hits.

Do I understand this change correctly, that we are now skipping `main` benchmark runs if we have an inexact cache hit? That would make us skip some `main` commits, right? Thus pull requests would not be compared against latest `main`. Thus comparisons would contain more than the pull request itself.

@larseggert am I missing something?

larseggert:
I think this is OBE; I just queued #2673 for merge.

mxinden:
Sounds good. Thanks.

PR #2690, opened 2025-06-02
ci: Don't check out twice

larseggert:
Sorry for botching the rebase earlier.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -44,17 +44,6 @@ jobs:
           persist-credentials: false
           clean: false
 
-
-      - name: Checkout mozilla/neqo@main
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -34,17 +34,6 @@ jobs:
           persist-credentials: false
           clean: false
```
mxinden, line 35: Related, also raised in the other PR, why do we need `clean: false`?

mxinden (approved):
Thanks for the follow-up!

.github/workflows/bench.yml
larseggert: So the `target` directories or other compile state isn't blown away. This should decrease build times.

PR #2688, opened 2025-06-02
ci(bench): use single Neqo Git checkout

mxinden:
Our criterion benchmarks report a difference between `main` and non-Rust PRs. That is unexpected. On each run we checkout Neqo twice, once for `main`, once for the corresponding PR. I assume the deviation is due to Criterion not being able to compare across checkouts.

Instead of using two Neqo checkouts, this pull request uses a single checkout, switching between `main` and PR branch.

Fixes https://github.com/mozilla/neqo/issues/2657.

See https://github.com/mozilla/neqo/pull/2673#issuecomment-2930071659.

PR #2693, opened 2025-06-03
ci: Don't use `nice` or `setarch` for bench run

larseggert:
I don't think either of these should be the reason for the performance differences, but 🤷‍♂️

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -167,8 +167,7 @@ jobs:
           rm -rf target/criterion
           for BENCH in $BENCHES; do
             # shellcheck disable=SC2086
-            nice -n -20 setarch --addr-no-randomize cset proc --set=cpu23 --exec \
-              $BENCH -- --bench --save-baseline main
+            cset proc --set=cpu23 --exec $BENCH -- --bench --save-baseline main
           done
           cd ../neqo
           rm -rf target/criterion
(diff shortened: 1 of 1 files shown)


mxinden (approved):
The simpler our setup becomes, the better.

larseggert:
As expected, this doesn't help :-(

mxinden:
> 
>        time:   [689.16 ms 690.30 ms 691.60 ms]
>        thrpt:  [144.59 MiB/s 144.86 MiB/s 145.10 MiB/s]
> change:
>        time:   [+7.5429% +7.8381% +8.1295%] (p = 0.00 < 0.05)
>        thrpt:  [−7.5183% −7.2684% −7.0138%]
> 
> Found 3 outliers among 100 measurements (3.00%)
> 1 (1.00%) high mild
> 2 (2.00%) high severe

First time I see a regression instead of an improvement. Ignoring the fact that there should be no difference at all, I have always seen a ~4% improvement, see e.g. recent benchmark results on https://github.com/mozilla/neqo/pull/2688.

mxinden:
https://github.com/mozilla/neqo/pull/2688 seems to fix the benchmark noise we see on unrelated changes.

I am in favor of proceeding here, but I suggest to wait till https://github.com/mozilla/neqo/pull/2688 lands to make sure this doesn't actually regress anything.

larseggert:
Prefer to leave it in, since we know have more stable runs.

PR #2696, opened 2025-06-04
ci: Fix `bencher.dev` upload

larseggert:
WIP

### DIFF
--- a/.github/workflows/bencher-upload.yml
+++ b/.github/workflows/bencher-upload.yml
@@ -34,6 +34,10 @@ jobs:
           name: ${{ env.PR_EVENT }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - run: |
+          ls -lt
+          cat event.json
+
       - name: Export PR event data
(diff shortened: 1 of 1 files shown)


larseggert:
This needs to be merged to `main` before we'll know whether it works.

.github/workflows/bencher-upload.yml
```diff
@@ -34,49 +34,27 @@ jobs:
           name: ${{ env.PR_EVENT }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - run: |
+          ls -lt
+          cat event.json
+
```
mxinden, line 40: Debugging leftover?

mxinden (approved):

.github/workflows/bencher-upload.yml
larseggert: Yeah. I'd like to leave this in for now to be able to debug in case there are still failures. (Since this won't execute until after merge to `main`.)

PR #2697, opened 2025-06-04
ci: Unpin `google/quiche` during bench

larseggert:
Since @martinduke asked whether it compiles again.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -57,7 +57,6 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           repository: google/quiche
-          ref: 23785904d54861f341ffef0ce5513b6c0a5beebc # TODO: Unpin when `main` compiles again.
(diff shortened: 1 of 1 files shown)


mxinden:
Preference for keeping it pinned, but updating it regularly, in order for our CI to become more reproducible.

larseggert:
Would be nice if Dependabot could do this...

mxinden:
> Would be nice if Dependabot could do this...

If we had google/quiche as a Git submodule, it seems like dependabot could keep it up-to-date. Search "git submodule" below.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

Also some users:

https://grep.app/search?q=package-ecosystem%3A+%22gitsubmodule

larseggert:
Let's not go the submodule route for something we only use for testing.

mxinden:
:+1: Not a strong opinion. I am fine either way.

larseggert:
I would suggest to unpin again. If we want to pin, we should pin all the comparison implementations.

PR #2699, opened 2025-06-05
ci: More bencher.dev fixes

larseggert:
- Also try an upload on workflow failure (because we fail the bench workflow on a performance regression but still want to track performance)
- Create fewer artifacts


larseggert:
As before, we'll only know if this works after it's merged to `main`.

mxinden (approved):

PR #2704, opened 2025-06-05
refactor(transport): don't implicitly infer PacketBuilder limit

mxinden:
Previously when creating a new `PacketBuilder`, it would first infer its limit implicitly based on the capacity of the provided `Encoder`:

```
fn infer_limit(encoder: &Encoder) -> usize {
    if encoder.capacity() > 64 {
        encoder.capacity()
    } else {
        2048
    }
}
```

Once created, it would then later on depend on the user to call `PacketBuilder::set_initial_limit` to set the exact explicit limit:

```
/// Set the initial limit for the packet, based on the profile and the PMTUD state.
/// Returns true if the packet needs padding.
pub fn set_initial_limit(
    &mut self,
    profile: &SendProfile,
    aead_expansion: usize,
    pmtud: &Pmtud,
) -> bool {
    if pmtud.needs_probe() {
        debug_assert!(pmtud.probe_size() >= profile.limit());
        self.limit = pmtud.probe_size() - aead_expansion;
        true
    } else {
        self.limit = profile.limit() - aead_expansion;
        false
    }
}
```

This commit simplifies the process, explicitly passing the exact limit from the start.

---

Draft for now. Unit tests need to be updated.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -27,10 +27,9 @@ use crate::{
     cid::{ConnectionId, ConnectionIdDecoder, ConnectionIdRef, MAX_CONNECTION_ID_LEN},
     crypto::{CryptoDxState, CryptoStates, Epoch},
     frame::FrameType,
-    recovery::SendProfile,
     tracking::PacketNumberSpace,
     version::{Version, WireVersion},
-    Error, Pmtud, Res,
+    Error, Res,
 };
 
 /// `MIN_INITIAL_PACKET_SIZE` is the smallest packet that can be used to establish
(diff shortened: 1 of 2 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -151,14 +150,6 @@ impl PacketBuilder {
     /// The minimum useful frame size.  If space is less than this, we will claim to be full.
     pub const MINIMUM_FRAME_SIZE: usize = 2;
 
-    fn infer_limit(encoder: &Encoder) -> usize {
-        if encoder.capacity() > 64 {
-            encoder.capacity()
-        } else {
-            2048
-        }
-    }
```
mxinden, line 160: The constants `2048` and `64` where introduced in https://github.com/mozilla/neqo/pull/998/files#diff-8ddbcc4657a65b23d868c4d46ac64145618c83d1266fd3eea9b3ad4eb77d662fR153-R159.

@martinthomson do you recall why you used these values in particular? Asked differently, what is the rational for the concrete values?

martinthomson: The 64 is a limit for an ACK-only packet.  It's based on a rough approximation of our maximum ACK size (which is in theory a lot bigger, but in practice 64 is pretty unlikely, even with high fragmentation).  2048 was picked as a fallback for when we didn't have an MTU.  Neither are that special.

mxinden:
@larseggert do you have some time to review this pull request?

larseggert (approved):

PR #2708, opened 2025-06-06
chore: Flip `avoid-breaking-exported-api` to `false`

larseggert:
And deal with the fallout.

We have one consumer and can afford to break the API, and this unlocks more linting.

### DIFF
--- a/neqo-crypto/src/aead.rs
+++ b/neqo-crypto/src/aead.rs
@@ -69,6 +69,7 @@ impl RealAead {
     }
 
     #[must_use]
+    #[expect(clippy::unused_self, reason = "We may need this in the future.")]
     pub const fn expansion(&self) -> usize {
         16
     }
--- a/neqo-crypto/src/aead.rs
+++ b/neqo-crypto/src/aead.rs
@@ -69,8 +69,7 @@ impl RealAead {
     }
 
     #[must_use]
-    #[expect(clippy::unused_self, reason = "We may need this in the future.")]
-    pub const fn expansion(&self) -> usize {
+    pub const fn expansion() -> usize {
         16
     }
 
@@ -153,7 +152,7 @@ impl RealAead {
                 aad.as_ptr(),
                 c_uint::try_from(aad.len())?,
                 data.as_ptr(),
-                c_uint::try_from(data.len() - self.expansion())?,
+                c_uint::try_from(data.len() - Self::expansion())?,
                 data.as_ptr(),
                 &mut l,
                 c_uint::try_from(data.len())?,
@@ -224,7 +223,7 @@ impl RealAead {
                 c_uint::try_from(data.len())?,
             )
         }?;
-        debug_assert_eq!(usize::try_from(l)?, data.len() - self.expansion());
+        debug_assert_eq!(usize::try_from(l)?, data.len() - Self::expansion());
         Ok(&mut data[..l.try_into()?])
     }
 }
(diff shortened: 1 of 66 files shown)


neqo-crypto/src/aead.rs
```diff
@@ -69,6 +69,7 @@ impl RealAead {
     }
 
     #[must_use]
+    #[expect(clippy::unused_self, reason = "We may need this in the future.")]
     pub const fn expansion(&self) -> usize {
```
mxinden: We can remove it now and re-introduce it in the future?

I am fine either way.

neqo-crypto/src/aead_null.rs
```diff
@@ -29,10 +33,15 @@ impl AeadNull {
...
+    #[expect(clippy::unused_self, reason = "We may need this in the future.")]
     pub const fn expansion(&self) -> usize {
         AEAD_NULL_TAG.len()
     }
 
+    #[expect(
+        clippy::unnecessary_wraps,
+        reason = "Need to replicate the API of aead::RealAead."
+    )]
     pub fn encrypt<'a>(
```
mxinden: Would it be possible for both to implement a trait? Would the lint then go away?

Either way, fine to ignore.

neqo-http3/src/conn_params.rs
```diff
@@ -30,9 +30,9 @@ impl Default for Http3Parameters {
         Self {
             conn_params: ConnectionParameters::default(),
             qpack_settings: QpackSettings {
-                max_table_size_encoder: QPACK_MAX_TABLE_SIZE_DEFAULT,
-                max_table_size_decoder: QPACK_MAX_TABLE_SIZE_DEFAULT,
-                max_blocked_streams: QPACK_MAX_BLOCKED_STREAMS_DEFAULT,
+                table_size_encoder: QPACK_MAX_TABLE_SIZE_DEFAULT,
+                table_size_decoder: QPACK_MAX_TABLE_SIZE_DEFAULT,
+                blocked_streams: QPACK_MAX_BLOCKED_STREAMS_DEFAULT,
```
mxinden: Why remove the `max_` here?

mxinden (approved):
Great. I wondered in the past why clippy ignores some types.

neqo-crypto/src/aead.rs
larseggert: Question for @martinthomson I think.

neqo-http3/src/conn_params.rs
larseggert: Clippy lint complaining about all fields having the same prefix. Can also ignore the lint?

larseggert: I opened #2724 for this.

neqo-crypto/src/aead_null.rs
larseggert: I opened #2724 for this.

neqo-crypto/src/aead.rs
larseggert: I made this change in 8331c75dc52c22c38eb2ae4634c5fdd7e7998966. LMKWYT.

neqo-http3/src/conn_params.rs
mxinden: I think the _max_ needs to be somewhere. After all it is the _maximum_ number of blocked streams, not the number of blocked streams. The `max_` prefix seems easiest to me. Alternatively we can include _max_ in the struct name.

PR #2709, opened 2025-06-06
chore: Bump macOS runners to `macos-15`

larseggert:
And also convert a few remaining `ubuntu-latest` instances to `ubuntu-24.04`.


larseggert:
After this is merged, need to update the `main` branch protection rules to require the new `macos-15` checks.

mxinden (approved):

PR #2720, opened 2025-06-09
fix: Disable some `hrtime` tests on `aarch64`

### LINKED ISSUE #2422: `hrtime::test::update_multi` fails occasionally in CI on Ubuntu

larseggert:
Resolves #2422

### DIFF
--- a/neqo-common/src/hrtime.rs
+++ b/neqo-common/src/hrtime.rs
@@ -495,7 +495,7 @@ mod test {
         check_delays(ONE_MS_AND_A_BIT);
     }
 
-    #[cfg(not(target_arch = "arm"))] // This test is flaky on linux/arm.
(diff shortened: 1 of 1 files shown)


neqo-common/src/hrtime.rs
```diff
@@ -487,6 +492,7 @@ mod test {
         thr.join().unwrap();
     }
 
+    #[cfg(not(target_arch = "aarch64"))] // This test is flaky on linux/arm.
```
mxinden, line 495: I only recall `update_multi` to be flaky. If that is really the case, I would prefer only disabling `update_multi` on `aarch64`.

mxinden (approved):

neqo-common/src/hrtime.rs
larseggert: It's other ones, too. Check the various links to failures at #2422.

PR #2722, opened 2025-06-10
fix: Strip Markdown formatting from bencher uploads

larseggert:


.github/workflows/bencher-upload.yml
```diff
@@ -58,6 +57,7 @@ jobs:
             --github-actions '${{ secrets.GITHUB_TOKEN }}' \
             --file results-main.txt
           for file in hyperfine-main/*.json; do
+            jq 'walk(if type == "object" and has("command") and .command[0:2] == "**" then.command |= .[2:-2] else . end)' < "$file" > stripped.json
```
mxinden: For my understanding, this would ideally live in `bencher`'s `shell_hyperfine` adapter?

mxinden (approved):

.github/workflows/bencher-upload.yml
larseggert: The `**` get added in `bench.yml` for some benchmarks we want to highlight. This strips it back out. Maybe it would be cleaner to strip it out over there; let me try that.

PR #2726, opened 2025-06-12
fix: Cache `RxStreamOrderer::bytes_ready`

larseggert:
Avoid iterating over the ranges when we don't need to. Let's see if this moves the needle on the benches.

TODO: If it does, we should consider maintaining a `bytes_ready` count as part of managing the ranges, and eliminate the cache.


larseggert:
Doesn't seem to matter.

PR #2727, opened 2025-06-12
ci: Check if we can benchmark and run `perf` at the same time

larseggert:
This would save 1/3 of the `cargo bench` run time.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -174,20 +174,13 @@ jobs:
           done
           # Copy the main branch results to a separate directory for bencher.dev.
           cp -r target/criterion ../criterion-main
+          # Run the builds from the PR branch and compare to baseline.
           git checkout -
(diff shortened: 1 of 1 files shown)


larseggert:
Based on this, it's clear that we can't.

mxinden:
> 
>        time:   [679.21 ms 680.57 ms 682.04 ms]
>        thrpt:  [146.62 MiB/s 146.94 MiB/s 147.23 MiB/s]
> change:
>        time:   [+3.0907% +3.3568% +3.6376%] (p = 0.00 < 0.05)
>        thrpt:  [−3.5100% −3.2478% −2.9981%]
> 
> Found 1 outliers among 100 measurements (1.00%)
> 1 (1.00%) high severe

Expected this to have a larger impact. That said, agreed, I am in favor of running `perf` separately as we do today.

Issue #2664, opened 2025-05-26
Consider optimizing `RecvStream::read` memory management


mxinden:
When reading off a stream trough `RecvStream::read`, unsurprisingly a lot of time is spent in (1) `copy_from_slice` and (2) de-allocating the now-no-longer-needed memory.

https://github.com/mozilla/neqo/blob/33ec4462d21b681df15e1647b79be05d426d8df5/neqo-transport/src/recv_stream.rs#L334-L374

Instead of the caller providing a `buf` to be copied into, maybe there is a way for `RecvStream` to return a chunk of the memory already allocated.

![Image](https://github.com/user-attachments/assets/eacb6a3e-2400-46e8-a335-57134771d7c2)

Also discussed in https://github.com/mozilla/neqo/pull/2620#pullrequestreview-2831178488.

PR #2729, opened 2025-06-13
fix: Use 256K for ranges

larseggert:
4K may lead to larger B-trees with higher overheads.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -157,6 +157,8 @@ impl RxStreamOrderer {
     /// Only when `u64` values cannot be converted to `usize`, which only
     /// happens on 32-bit machines that hold far too much data at the same time.
     pub fn inbound_frame(&mut self, mut new_start: u64, mut new_data: &[u8]) {
+        const MAX_RANGE: usize = 64 * 1024; // 64 KiB is the maximum range we allow in a single frame.
+
         qtrace!("Inbound data offset={new_start} len={}", new_data.len());
 
         // Get entry before where new entry would go, so we can see if we already
@@ -197,7 +199,7 @@ impl RxStreamOrderer {
                 // If it is small enough, extend the previous buffer.
                 // This can't always extend, because otherwise the buffer could end up
                 // growing indefinitely without being released.
-                prev_vec.len() < 4096 && prev_end == new_start
+                prev_vec.len() < MAX_RANGE && prev_end == new_start
             } else {
                 // PPPPPP    ->  PPPPPP
                 //   NNNN
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -157,7 +157,7 @@ impl RxStreamOrderer {
     /// Only when `u64` values cannot be converted to `usize`, which only
     /// happens on 32-bit machines that hold far too much data at the same time.
     pub fn inbound_frame(&mut self, mut new_start: u64, mut new_data: &[u8]) {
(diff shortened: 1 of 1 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -157,6 +157,8 @@ impl RxStreamOrderer {
     /// Only when `u64` values cannot be converted to `usize`, which only
     /// happens on 32-bit machines that hold far too much data at the same time.
     pub fn inbound_frame(&mut self, mut new_start: u64, mut new_data: &[u8]) {
+        const MAX_RANGE: usize = 64 * 1024; // 64 KiB is the maximum range we allow in a single frame.
```
larseggert:
```suggestion
        const MAX_RANGE: usize = 64 * 1024; // 64 KB is the maximum length of a range.
```

neqo-transport/src/recv_stream.rs
```diff
@@ -157,6 +157,8 @@ impl RxStreamOrderer {
     /// Only when `u64` values cannot be converted to `usize`, which only
     /// happens on 32-bit machines that hold far too much data at the same time.
     pub fn inbound_frame(&mut self, mut new_start: u64, mut new_data: &[u8]) {
+        const MAX_RANGE: usize = 64 * 1024; // 64 KB is the maximum length of a range.
```
larseggert:
```suggestion
        const MAX_RANGE: usize = 256 * 1024; // 256 KB is the maximum length of a range.
```

martinthomson:
What led to the choice of 256K?  Is the [RxStreamOrderer::inbound_frame()](https://bencher.dev/perf/neqo/benchmarks/rxstreamorderer-inbound-frame) test what you used to tune this?

larseggert:
> What led to the choice of 256K? Is the [RxStreamOrderer::inbound_frame()](https://bencher.dev/perf/neqo/benchmarks/rxstreamorderer-inbound-frame) test what you used to tune this?

No, I looked at the flamegraphs in the profiler. I first tried 64KB, then 256KB to see if that made more of an improvement. It's a bit inconsistent. I hope that once #2593 lands and we pump more bytes through the code, things will be clearer.

larseggert:
I think we may save more if we could change this
```rust
self.data_ranges.insert(new_start, to_add.to_vec());
```
to something that inserted a 256KB `Vec` with `to_add` at the front. There isn't something like `to_vec_with_capacity` is there?

mxinden:
Long term, I would like to explore https://github.com/mozilla/neqo/issues/2664, i.e. instead of copying into a user's provided buffer, simply return the existing, no-longer-needed allocation from `RxStreamOrdered`.

---

Short term - thinking out loud:

I assume we extend existing ranges for the sake of de-fragmentation and thus keeping `RxStreamOrdered` `BTreeMap` small.

I would assume  on the `Downlaod` and `Upload` benchmarks, given that they use a single stream only, and given that they fill each UDP Datagram to the max, that most ranges are large.

Under the above assumption, if we don't extend existing ranges at all in `RxStreamOrdered::inbound_frame`, we would save on a lot of memory copies. Assuming memory copies are more expensive than the additional fragmentation and thus larger `BTreeMap`, we would thus speed those two benchmarks up.

@larseggert is this worth trying? (Mostly challenging my intuition here.)

larseggert:
Worth a try! Don't have good intuition on whether the memory copy or data structure management overheads are larger.

larseggert:
Agree #2664 is the long term goal.

larseggert:
This seems to have worse performance. Dropping it.

PR #2730, opened 2025-06-13
fix: Don't construct `Http3StreamInfo` on every call to `get_info`

larseggert:
Suggested by @martinthomson in https://github.com/mozilla/neqo/pull/2623#discussion_r2084522167

### DIFF
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -544,14 +544,14 @@ impl Http3StreamInfo {
 }
 
 trait RecvStreamEvents: Debug {
-    fn data_readable(&self, _stream_info: Http3StreamInfo) {}
-    fn recv_closed(&self, _stream_info: Http3StreamInfo, _close_type: CloseType) {}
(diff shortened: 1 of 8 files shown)


neqo-http3/src/lib.rs
```diff
@@ -544,14 +544,14 @@ impl Http3StreamInfo {
...
-    fn recv_closed(&self, _stream_info: Http3StreamInfo, _close_type: CloseType) {}
+    fn data_readable(&self, _stream_info: &Http3StreamInfo) {}
+    fn recv_closed(&self, _stream_info: &Http3StreamInfo, _close_type: CloseType) {}
 }
 
 trait HttpRecvStreamEvents: RecvStreamEvents {
     fn header_ready(
         &self,
-        stream_info: Http3StreamInfo,
+        stream_info: &Http3StreamInfo,
```
martinthomson, line 554: I took a brief look and it seems like half of these functions need a value, but others don't.  That makes this a bit awkward, but I think you have the right balance.

martinthomson (approved):

Issue #2728, opened 2025-06-13
Add http3 support to `test-fixtures` `Simulator`


mxinden:
The `test-fixtures` `Simulator` currently supports `neqo-transport` `Connection`s only. In other words, it can only simulate QUIC connections.

https://github.com/mozilla/neqo/blob/37c3aeebb79aef9f9649c54c3bbfae84fee523b3/test-fixture/src/sim/connection.rs#L55-L56

It would be helpful to be able to simulate HTTP3 connections. E.g. in the `neqo-http3` `streams.rs` benchmark we currently re-implement the package-exchange logic, though less sophisticated, e.g. without timers, without a check for success, ...

https://github.com/mozilla/neqo/blob/37c3aeebb79aef9f9649c54c3bbfae84fee523b3/neqo-http3/benches/streams.rs#L22-L37 

(The `streams.rs` benchmark sees some false positives. Moving to a sophisticated `Simulator` like we have in `neqo-transport` might resolve the issue.)

PR #2734, opened 2025-06-13
bench(bin/server): process input datagram in batches

mxinden:
Have `neqo-server` batch process input datagrams. Note that `neqo-client` already does this, i.e. `neqo-server` is simply catching up.


mxinden:
Looking at benchmark execution logs, the server never reads more than one datagram from the socket. I assume this is due to missing GSO support (https://github.com/mozilla/neqo/pull/2593) on the client. Once https://github.com/mozilla/neqo/pull/2593 merged, we should re-evaluate here. Keeping as draft for now.

mxinden:
> 1-conn/1-100mb-req/mtu-1504 (aka. Upload)/client: 💚 Performance has improved.
> 
>        time:   [226.65 ms 226.94 ms 227.31 ms]
>        thrpt:  [439.92 MiB/s 440.64 MiB/s 441.21 MiB/s]
> change:
>        time:   [−65.945% −65.654% −65.374%] (p = 0.00 < 0.05)
>        thrpt:  [+188.80% +191.15% +193.64%]
> 

Confirming the suspicion from https://github.com/mozilla/neqo/pull/2734#issuecomment-2972411999 and brings Upload benchmark up to par with Download benchmark. Still need to look into the regressions.

neqo-bin/src/server/mod.rs
```diff
@@ -204,29 +209,34 @@ fn qns_read_response(filename: &str) -> Result<Vec<u8>, io::Error> {
...
-    fn process_multiple(
+    fn new(
+        args: &Args,
+        anti_replay: AntiReplay,
+        cid_mgr: Rc<RefCell<dyn ConnectionIdGenerator>>,
+    ) -> Self;
+    fn process_multiple<'a>(
         &mut self,
-        dgram: Option<Datagram<&mut [u8]>>,
+        dgrams: impl IntoIterator<Item = Datagram<&'a mut [u8]>>,
```
mxinden, line 219: This change makes `HttpServer` no longer [_dyn compatible_](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility).

Instead of `Runner` owning a `server: Box<dyn HttpServer>`, it now owns a `server: S` where `S` is a new trait parameter.

In my eyes the additional complexity (trait parameter `S`) is worth it. Dynamic dispatching is difficult to work with in Rust, especially when using Rust's Futures. In addition, given that `neqo-bin` is at the outskirt of our code base, the new trait parameter `S` does not infect other places.

neqo-bin/src/server/mod.rs
```diff
@@ -321,16 +335,14 @@ impl Runner {
...
-                .await?;
-            }
+            Self::process_inner(
+                &mut self.server,
+                &mut self.timeout,
+                &mut self.sockets,
+                &self.now,
+                Some(input_dgrams),
+            )
+            .await?;
```
mxinden, line 345: This is the core of this pull request. Instead of looping over all input datagrams, `process`ing each individually, pass it down as a batch.

neqo-transport/src/server.rs
```diff
@@ -351,113 +360,134 @@ impl Server {
...
+                continue;
+            };
+
+            // Finding an existing connection. Should be the most common case.
+            if let Some(c) = self
+                .connections
+                .iter_mut()
+                .find(|c| c.borrow().is_valid_local_cid(packet.dcid()))
+            {
+                c.borrow_mut().process_input(dgram, now);
```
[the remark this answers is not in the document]
mxinden: Check again. `process_input` has no return value.

mxinden:
This pull request is ready for review.

---

This is the main change, a 200% improvement in throughput. Note that this simply our server implementation catching up to our client implementation. In other words, this will not be visible in Firefox.

> 1-conn/1-100mb-req/mtu-1504 (aka. Upload)/client: 💚 Performance has improved.
> 
>        time:   [202.71 ms 202.99 ms 203.32 ms]
>        thrpt:  [491.84 MiB/s 492.65 MiB/s 493.32 MiB/s]
> change:
>        time:   [−68.468% −68.168% −67.860%] (p = 0.00 < 0.05)
>        thrpt:  [+211.13% +214.15% +217.14%]

---

> 1000 streams of 1 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [42.424 ns 42.841 ns 43.257 ns]
>        change: [+50.786% +52.616% +54.599%] (p = 0.00 < 0.05)
> 
> 1000 streams of 1000 bytes/multistream: 💔 Performance has regressed.
> 
>        time:   [43.138 ns 43.575 ns 44.009 ns]
>        change: [+56.158% +58.103% +60.133%] (p = 0.00 < 0.05)

I have seen these benchmarks give a lot of false positive signals. I don't yet know why. Here, I suggest merging even with the signal. Long term I would like to do https://github.com/mozilla/neqo/issues/2728.

---

> [codecov/patch](https://github.com/mozilla/neqo/pull/2734/checks?check_run_id=45880755610)Failing after 1s — 94.11% of diff hit (target 94.92%)

The difference in code coverage is a combination of `unreachable`s and `assert`s that ensure unreachable states are unreachable. By definition, those won't be covered by tests. Other minor changes are simply moves, that weren't covered before.

neqo-transport/src/server.rs
martinthomson, line 389: How often is it that we get a batch of datagrams and they are from different connections?  That is, does it make sense to do a full scan for every datagram?  Or could we cache the result and test that connection for any subsequent items.

Consider:

```rust
if let Some(c) = last_c.take().into_iter().as_mut()
    .chain(self.connections.iter_mut())
    .find(|c| ...) {
    last_c = Some(Rc::clone(&c);
    c.borrow_mut().process_input(...);
}
```

neqo-transport/src/server.rs
```diff
@@ -121,6 +123,12 @@ pub struct Server {
     qlog_dir: Option<PathBuf>,
     /// Encrypted client hello (ECH) configuration.
     ech_config: Option<EchConfig>,
+    /// Remaining datagrams of a batch of datagrams provided via
+    /// [`Server::process_multiple`]. An earlier datagram in the batch required
+    /// an immediate return without further processing of the remaining
+    /// datagrams. To be processed on consecutive calls to
+    /// [`Server::process_multiple`].
+    saved_datagrams: VecDeque<SavedDatagram>,
```
martinthomson, line 131: Why `VecDeque`?  It seems like this typically is a one-off: you save datagrams only for VN and the first Initial, but then are most likely going to drain the whole thing next time.

martinthomson (approved):
This seems right.

The performance numbers for those tests that regress here seems bad though.  If this is just flakiness, what can we do to remove that flakiness?

neqo-transport/src/server.rs
mxinden: Yes, but there is an edge case, where datagrams are queued to `saved_datagrams` over multiple `process_multiple` calls, building a queue. I would like to processing them in the right order, i.e. not introduce complexity of out-of-order processing, thus using a `VecDeque`.

(Note this is implemented as a unit test in this pull request. See `neqo-transport/tests/server.rs`.)

Let me know if you are still concerned with using a `VecDeque` here.

neqo-transport/src/server.rs
mxinden: > How often is it that we get a batch of datagrams and they are from different connections?

On a connection over the internet, I would assume close to never. On a localhost connection with close-to zero latency, still rarely I suppose.

That said, given that `self.connections` is likely small, especially given this being a test server only, I doubt iterating the connections is expensive. 

Below a CPU flamegraph of a server during our _Upload_ benchmark. Iteration of `self.connections` does not show up.

image

I would only introduce the above suggested optimization if we have proof it has an impact. Thus far no proof, thereby I prefer not introducing it just yet.

Let me know if you object.

neqo-transport/src/server.rs
martinthomson: Ah, so you are concerned that there might be a batch of datagrams, of which you process one or a small number, repeated several times.  And you are concerned that the removal from the head will result in less efficiency.  Isn't that still a problem if you need to scale up?  When you need to resize up, you do a lot more work with a VecDeque.

neqo-transport/src/server.rs
martinthomson: Wow, I don't think we have much room to improve there if we are spending that large a portion of the time just decrypting.

Note that we are not testing a situation where there are multiple connections, so of course this doesn't show.  That said, this code is no good for that case anyway.

larseggert:
@mxinden anything you want to change here before this can land?

neqo-transport/src/server.rs
mxinden: > Ah, so you are concerned that there might be a batch of datagrams, of which you process one or a small number, repeated several times.

Yes, correct. We need to handle the case where we don't drain `saved_datagrams` each time.

> And you are concerned that the removal from the head will result in less efficiency.

I am not concerned about efficiency here. I want to ensure we handle incoming datagrams in order. I consider `VecDeque` the idiomatic `std::collection` when a queue (push at the end, pop at the front) is needed.

mxinden:
Just Martin's concern above, that this pull request regresses the `multistream` benchmarks a lot.

> The performance numbers for those tests that regress here seems bad though. If this is just flakiness, what can we do to remove that flakiness?

I haven't found these to be useful in the past. That said, before merging here, it makes sense to me to either remove them or do https://github.com/mozilla/neqo/issues/2728.

larseggert:
I added because I wanted to measure whether changing to `FxHasher` was giving a benefit. I am fine with removing them.

larseggert:
#2805 disables the `stream` benches. Doing #2728 would be nice. I'll merge this PR now.

mxinden:
:tada: we have symetric up- and download performance now. Thanks for merging.

PR #2739, opened 2025-06-16
fix: Flip `NeqoQlog` types around

larseggert:
From `Rc<RefCell<Option<NeqoQlogShared>>>` to `Option<Rc<RefCell<NeqoQlogShared>>>`.

We spend 1% of runtime in `borrow_mut()` even when qlogs are disabled (= the inner `Option` is `None`.)

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -27,7 +27,7 @@ use crate::Role;
 
 #[derive(Debug, Clone, Default)]
 pub struct NeqoQlog {
-    inner: Rc<RefCell<Option<NeqoQlogShared>>>,
+    inner: Option<Rc<RefCell<NeqoQlogShared>>>,
 }
 
 pub struct NeqoQlogShared {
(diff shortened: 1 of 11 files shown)


mxinden:
I am as well surprised that `RefCell::borrow_mut` is so expensive. I looked into it in the past, but couldn't spot why.

https://doc.rust-lang.org/src/core/cell.rs.html#1089

Note that we set `Option::None` at runtime. Moving the `Option` outside the `Rc` would prevent this change from propagating to all references.

https://github.com/mozilla/neqo/blob/8f02fd962c60a5792a6d27ca266205cd21253e53/neqo-common/src/qlog.rs#L161-L163

larseggert:
I think I have an issue open that we should put `qlog` behind a feature. We don't use it in Gecko AFAIK.

mxinden:
> We don't use it in Gecko AFAIK.

It can be enabled via a Firefox Pref.

https://firefox-source-docs.mozilla.org/networking/http/logging.html#how-to-enable-quic-logging

larseggert:
Have we ever used that?

mxinden:
Yes, I have used it to debug congestion control issues, e.g. when uploading to Google Drive.

I don't know how to get real-world qlogs (e.g. Google Drive upload) with `neqo-client` alone. Maybe I am missing something?

mxinden:
> We spend 1% of runtime in `borrow_mut()` even when qlogs are disabled (= the inner `Option` is `None`.)

Can you share this profile? I can only reproduce < 1%.

larseggert:
I don't seem to have the trace anymore. And given the issue you identified, I don't think we can ship this.

PR #2740, opened 2025-06-16
fix: Call `Instant::now` less frequently in client

larseggert:
It makes up 1% of CPU time.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -431,7 +431,8 @@ impl<'a, H: Handler> Runner<'a, H> {
     async fn run(mut self) -> Res<Option<ResumptionToken>> {
         loop {
             let handler_done = self.handler.handle(&mut self.client)?;
-            self.process_output().await?;
+            let now = Instant::now();
+            self.process_output(now).await?;
             if self.client.has_events() {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

neqo-bin/src/client/mod.rs
```diff
@@ -431,7 +431,8 @@ impl<'a, H: Handler> Runner<'a, H> {
     async fn run(mut self) -> Res<Option<ResumptionToken>> {
         loop {
             let handler_done = self.handler.handle(&mut self.client)?;
-            self.process_output().await?;
+            let now = Instant::now();
+            self.process_output(now).await?;
```
mxinden: The only other usage of `now` is in calling `self.client.close`. I assume the latter happens only once per connection. Thus I don't think the change has any performance benefit. Am I missing something?
```suggestion
            self.process_output(Instant::now()).await?;
```

mxinden (approved):

neqo-bin/src/client/mod.rs
larseggert: You're right. I had `let now = Instant::now();` outside the loop before, but then I realized `self.process_output(now).await?;` can block, which means we would need a fresh timestamp.

larseggert: (Or should it be outside the loop?)

mxinden: Outside the loop would mean that a Neqo client would have a single timestamp for the entire runtime.

mxinden (commented):
Code looks good to me.

Does this resolve the 1% CPU time of `Instant::now`? Unfortunately, this does not seem to improve benchmarks. Maybe the 1% CPU time was worth it, given the more accurate time resolution in e.g. the congestion controller?

larseggert:
I need to check the flamegraph once the latest bench run finishes.

larseggert:
It does reduce, but only by 0.1%, because there are several other places where we (and `tokio`) call `Instant::now`. So probably not useful.

PR #2742, opened 2025-06-17
fix: Keep tracking ECN marks in `ValidationState::Unknown`

### LINKED ISSUE #2741: ECN validation failed, no ECT(0) packets were newly acked
When I run `RUST_LOG=warn cargo bench --features bench --bench transfer` I see
```
Benchmarking transfer/pacing-false/varying-seeds: Warming up for 5.0000 s0.000 WARN [Server ...] Dropped received packet: Coalesced packet has different DCID; Total: 1
0.000 WARN [Client ...] Dropped received packet: Coalesced packet has different DCID; Total: 1
0.000 WARN [LossRecovery] ignoring in-1 from dropped space
0.027 WARN [Server ...] Dropped received packet: Coalesced packet has different DCID; Total: 1
0.028 WARN [Client ...] Dropped received packet: Coalesced packet has different DCID; Total: 1
0.028 WARN [LossRecovery] ignoring in-1 from dropped space
0.028 WARN ECN validation failed, no ECT(0) packets were newly acked
```
and many more instances of the last line. That is surprising. Why is ECN validation failing over loopback?

larseggert:
This doesn't quite work, because we never receive an ACK for the ECN probes in the Initial packet number space before dropping it, so the probe and ACK counts never align.

I think #2560 is the better approach here, i.e., do ECN once we're in established state.

Fixes #2741

### DIFF
--- a/neqo-transport/src/ecn.rs
+++ b/neqo-transport/src/ecn.rs
@@ -36,7 +36,11 @@ enum ValidationState {
         initial_probes_lost: usize,
     },
     /// The validation test has concluded but the path's ECN capability is not yet known.
-    Unknown,
+    Unknown {
+        probes_sent: usize,
+        initial_probes_acked: usize,
+        initial_probes_lost: usize,
+    },
     /// The path is known to **not** be ECN capable.
     Failed(ValidationError),
     /// The path is known to be ECN capable.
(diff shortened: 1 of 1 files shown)


larseggert:
Closing in favor of https://github.com/mozilla/neqo/pull/2560.

PR #2747, opened 2025-06-21
perf: don't allocate in UDP send path

### LINKED ISSUE #2670: Consider not pre-allocating each UDP datagram in `output_path`

mxinden:
Maintain a long-lived send buffer. When producing outbound UDP datagrams
write them into the buffer. Then pass the buffer to the OS to be sent
out on the network.

In other words, don't heap-allocate in the UDP send path.

---

Corresponding past patch for the receive path https://github.com/mozilla/neqo/pull/2184
Fixes https://github.com/mozilla/neqo/issues/2670

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -11,7 +11,7 @@ use std::{
     collections::VecDeque,
     fmt::Display,
     fs::File,
-    io::{BufWriter, Write as _},
+    io::{BufWriter, Cursor, Write as _},
     net::SocketAddr,
     num::NonZeroUsize,
     path::PathBuf,
@@ -190,12 +190,13 @@ impl TryFrom<&State> for CloseState {
 }
 
 impl super::Client for Connection {
-    fn process_multiple_output(
+    fn process_multiple_output<'a>(
         &mut self,
         now: Instant,
+        send_buffer: Cursor<&'a mut [u8]>,
         max_datagrams: NonZeroUsize,
-    ) -> OutputBatch {
-        self.process_multiple_output(now, max_datagrams)
+    ) -> OutputBatch<Cursor<&'a mut [u8]>> {
+        self.process_multiple_output(now, send_buffer, max_datagrams)
     }
 
     fn process_multiple_input<'a>(
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -11,7 +11,7 @@ use std::{
     collections::VecDeque,
     fmt::Display,
     fs::File,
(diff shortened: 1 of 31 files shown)


mxinden:
I don't think at this point the performance improvement (~4% see https://github.com/mozilla/neqo/pull/2861#issuecomment-3194223848)) is worth the added complexity. Complexity being:

- new trait parameter `B: Buffer` on most public `process*` functions
- a necessary refactor of `test_frame_writer`
  - as of today, it needs to be object safe, which is in conflict with the new trait parameter `B`
  - alternatively we could require `Cursor<&mut [u8]>` everywhere
  - or don't store `test_frame_writer` in `Connection, but instead pass it along the `process*` functions

I will close here for now. We can revisit this once we want to spend more time on optimizing IO performance.

larseggert:
@mxinden revived this to do another benchmark run, now that we addressed some performance issues elsewhere in the code. Thanks for the reminder about this one!

mxinden:
Good idea. Thanks.

PR #2753, opened 2025-06-23
fix: Remove `module_name_repetitions`

### LINKED ISSUE #2284: chore(transport): remove `module_name_repetitions`
We allow the clippy lint [`module_name_repetitions`](https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions):

https://github.com/mozilla/neqo/blob/a7581770ff138c16727e92ecb7b47d9ceb95425e/neqo-transport/src/lib.rs#L7

We e.g. prefix various types in the `ecn` module with `Ecn`:

https://github.com/mozilla/neqo/blob/3001a3a56f2274eaafaa956fb394f0817f526ae7/neqo-transport/src/ecn.rs#L18-L30

As discussed in https://github.com/mozilla/neqo/pull/2270#discussion_r1874515991, I suggest removing the allow and instead fix the module name repetition.

larseggert:
I'm not sure if we really want this, or all of it. It's quite noisy.

Fixes #2284

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use neqo_common::{qdebug, qerror, qinfo, qtrace, qwarn, Decoder, Header, MessageType, Role};
-use neqo_qpack::qpack;
+use neqo_qpack as qpack;
 use neqo_transport::{
     streams::SendOrder, AppError, CloseReason, Connection, DatagramTracking, State, StreamId,
     StreamType, ZeroRttState,
--- a/neqo-qpack/src/lib.rs
+++ b/neqo-qpack/src/lib.rs
@@ -24,9 +24,7 @@ use std::fmt::{self, Display, Formatter};
 
 pub use stats::Stats;
 
-pub mod qpack {
-    pub use crate::{decoder::Decoder, encoder::Encoder, Settings};
-}
+pub use crate::{decoder::Decoder, encoder::Encoder};
 
 type Res<T> = Result<T, Error>;
 
--- a/neqo-http3/src/conn_params.rs
+++ b/neqo-http3/src/conn_params.rs
@@ -6,7 +6,7 @@
 
 use std::cmp::min;
 
-use neqo_qpack::qpack;
+use neqo_qpack as qpack;
 use neqo_transport::ConnectionParameters;
 
 const QPACK_MAX_TABLE_SIZE_DEFAULT: u64 = 65536;
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -1294,7 +1294,7 @@ mod tests {
 
     use neqo_common::{event::Provider as _, qtrace, Datagram, Decoder, Encoder};
     use neqo_crypto::{AllowZeroRtt, AntiReplay, ResumptionToken};
-    use neqo_qpack::qpack;
+    use neqo_qpack as qpack;
     use neqo_transport::{
         CloseReason, ConnectionEvent, ConnectionParameters, Output, State, StreamId, StreamType,
         Version, INITIAL_RECV_WINDOW_SIZE, MIN_INITIAL_PACKET_SIZE,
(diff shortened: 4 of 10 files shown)


neqo-crypto/src/hp.rs
```diff
@@ -44,6 +39,7 @@ experimental_api!(SSL_HkdfExpandLabelWithMech(
 ));
 
 #[derive(Clone)]
+#[expect(clippy::module_name_repetitions, reason = "This is OK.")]
```
martinthomson: I wouldn't object to this one changing, though we'd need to alias it to HpKey almost everywhere it is used...

neqo-http3/src/connection.rs
```diff
@@ -1605,12 +1605,12 @@ impl Http3Connection {
     }
 
     #[must_use]
-    pub const fn qpack_encoder(&self) -> &Rc<RefCell<QPackEncoder>> {
+    pub const fn qpack_encoder(&self) -> &Rc<RefCell<encoder::QPack>> {
```
martinthomson: This seems like less of a win.  I wouldn't mind qpack::Encoder/qpack::Decoder being exported from a qpack module.

neqo-transport/src/connection/mod.rs
```diff
@@ -3234,24 +3240,26 @@ impl Connection {
...
-                    RecoveryToken::Stream(stream_token) => self.streams.acked(stream_token),
-                    RecoveryToken::Ack(at) => self.acks.acked(at),
-                    RecoveryToken::Crypto(ct) => self.crypto.acked(ct),
-                    RecoveryToken::NewToken(seqno) => self.new_token.acked(*seqno),
-                    RecoveryToken::NewConnectionId(entry) => self.cid_manager.acked(entry),
-                    RecoveryToken::RetireConnectionId(seqno) => self.paths.acked_retire_cid(*seqno),
-                    RecoveryToken::AckFrequency(rate) => self.paths.acked_ack_frequency(rate),
-                    RecoveryToken::KeepAlive => self.idle_timeout.ack_keep_alive(),
-                    RecoveryToken::Datagram(dgram_tracker) => self
+                    recovery::Token::Stream(stream_token) => self.streams.acked(stream_token),
```
martinthomson, line 3247: For this change, I would have aliased this:

```rust
use recovery::Token as RecoveryToken;
```

neqo-qpack/src/lib.rs
```diff
@@ -22,15 +22,17 @@ mod table;
 
 use std::fmt::{self, Display, Formatter};
 
-pub use decoder::QPackDecoder;
-pub use encoder::QPackEncoder;
 pub use stats::Stats;
 
+pub mod qpack {
```
martinthomson: Why the module?  Isn't the crate enough?

neqo-qpack/src/qpack_send_buf.rs
```diff
@@ -8,14 +8,14 @@ use std::ops::Deref;
 
 use neqo_common::Encoder;
 
-use crate::{huffman::encode_huffman, prefix::Prefix};
+use crate::{huffman, prefix::Prefix};
 
 #[derive(Default, Debug, PartialEq, Eq)]
-pub struct QpackData {
+pub struct Data {
```
martinthomson, line 14: This name is awful.  What is it?  From the implementation, it looks like a buffer, not a unit of data.

Not for this PR, but this looks like it could be replaced by a trait that defines the encode_prefixed_encoded_int (also a name that could be better) or maybe just the encode_literal methods and attaches them to `neqo_common::Encoder`.

martinthomson (approved):
Impossible to really get a handle on something at this scale.  But everything I've seen is generally an improvement.

neqo-qpack/src/qpack_send_buf.rs
larseggert: I opened https://github.com/mozilla/neqo/issues/2754

neqo-qpack/src/lib.rs
larseggert: Good idea, fixed in fd63931bceb0f47ee51f94724d40bfb791fae47d.

larseggert:
Forcing the merge. The bench workflow fails because a bench name changed and hence there is no baseline to compare to.

Issue #2752, opened 2025-06-23
`send_ignore_emsgsize` fails on BSDs


larseggert:
On FreeBSD:
```
---- tests::send_ignore_emsgsize stdout ----
  0.000 DEBUG Failed to send datagram of size 65536 from 127.0.0.1:51845 to 127.0.0.1:18401. PMTUD probe? Ignoring error: Message too long (os error 40)
  0.000 TRACE sent 12 bytes from 127.0.0.1:51845 to 127.0.0.1:18401
  Error: Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

larseggert:
Also fails on macOS:
```
---- tests::send_ignore_emsgsize stdout ----
0.000 DEBUG Logging initialized
0.000 DEBUG Failed to send datagram of size 65536 from 127.0.0.1:61867 to 127.0.0.1:60316. PMTUD probe? Ignoring error: Message too long (os error 40)
0.000 TRACE sent 12 bytes from 127.0.0.1:61867 to 127.0.0.1:60316
Error: Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

mxinden:
```
  0.000 DEBUG Failed to send datagram of size 65536 from 127.0.0.1:51845 to 127.0.0.1:18401. PMTUD probe? Ignoring error: Message too long (os error 40)
```

This is expected.

Before https://github.com/quinn-rs/quinn/pull/2199 `quinn-udp` would always ignore `EMSGSIZE`. With https://github.com/quinn-rs/quinn/pull/2199 `EMSGSIZE` is returned by `try_send`, for the user (i.e. `neqo-udp`) to handle.

```
  Error: Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

This is expected. That said, I am surprised it is printed. Not sure where it is printed. The test should only check for it.

https://github.com/mozilla/neqo/blob/0f3262121c1c390fd70cb24d97c63660ada90fb8/neqo-udp/src/lib.rs#L394

I will take a look.

Edit: I assume it is a timing issue, where the last send hasn't yet arrived at the socket. I will create a patch.

PR #2755, opened 2025-06-24
test(udp): set emsgsize test socket to blocking

mxinden:
The `send_ignore_emsgsize` unit test first sends an oversized datagram, expecting it to be ignored, then sends a normal datagram, expecting it to be received.

In the second check, to avoid race conditions, set the receiving socket to blocking before the call to `receiver.recv`.

Fixes https://github.com/mozilla/neqo/issues/2752


mxinden:
Fuzz failures will be fixed in https://github.com/mozilla/neqo/pull/2756.

larseggert (approved):

PR #2759, opened 2025-06-26
ci: Bump the cache keys

larseggert:
Invalidate them, to see if the recent weird CI failures are due to cache corruption.


larseggert:
Yeah, that didn't help. I have no clue what broke Linux and Windows Rust compilation; I don't think it's a change in neqo.

PR #2763, opened 2025-06-30
ci: Run `cargo bench` and perf compare in parallel

larseggert:
Now that we have multiple self-hosted runners, this should speed up things.


larseggert:
So is it better to block two runners for 30min or one runner for an hour?

mxinden:
The two tasks should be close to perfectly parallelizable, as in, splitting them does not add much duplication, right? Thus the former gives us a speed up of 30 minutes when CI is otherwise idle, while maintaining the status quo when CI is busy.

larseggert:
This needs to be on `main` before it can be debugged further.

PR #2764, opened 2025-06-30
feat: Use `HashMap` for `RecvStreams`

larseggert:
Does this matter in terms of performance?

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -10,7 +10,7 @@
 use std::{
     cell::RefCell,
     cmp::max,
-    collections::BTreeMap,
+    collections::{BTreeMap, HashMap},
     fmt::Debug,
     mem,
     rc::{Rc, Weak},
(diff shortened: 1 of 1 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -63,6 +63,10 @@ impl RecvStreams {
         now: Instant,
         rtt: Duration,
     ) {
+        #[expect(
+            clippy::iter_over_hash_type,
+            reason = "OK to loop over streams in an undefined order."
```
martinthomson, line 68: Our prioritization scheme depends on in-order progress in the order of creation.  It's not fatal, but we've always assumed that this is in-order.  Otherwise, you get bytes in an indeterminate order.  In particular, we don't tend to adjust priority of control streams in HTTP.

larseggert: OK, then we need to add a test for that.

larseggert: https://github.com/mozilla/neqo/issues/2765

PR #2769, opened 2025-07-01
ci: Run bencher in GCP

larseggert:
Or at least let's see what happens...


larseggert:
This needs to be on `main` for it to become debuggable. Merging now.

PR #2773, opened 2025-07-02
ci: Don't delay the start of CI benchmarks

larseggert:
Now that we have multiple runners, we don't need to be so careful to avoid wasting runtime there, and can instead trade for earlier completion.


larseggert:
This needs to be on `main` for the PR comment workflows to fire. Merging now.

PR #2777, opened 2025-07-03
Introduce a `recovery::Tokens` type

larseggert:
~~For some reason, I wasn't allowed to push a rebase straight to #1657. So this is a new PR. Wanted to check if all the perf work we did in the meantime changes things here, and to get perf traces.~~

This PR now just introduces the `recovery::Tokens` type, which is still based on a `Vec`.

### DIFF
--- a/neqo-transport/benches/sent_packets.rs
+++ b/neqo-transport/benches/sent_packets.rs
@@ -7,7 +7,10 @@
 use std::{hint::black_box, time::Instant};
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use neqo_transport::{self, packet, recovery::sent};
+use neqo_transport::{
+    self, packet,
+    recovery::{sent, RecoveryTokenVec},
+};
 
 fn sent_packets() -> sent::Packets {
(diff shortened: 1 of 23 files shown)


martinthomson:
This seems fairly substantially worse.  I guess the question to ask - if you wanted to keep on at this - is this because the size of the array is too small?  If we pick a too-small value, we might be reallocating more often for tokens.  Maybe we can test that by logging the size of the store at the end of packet construction.  Plotting the distribution might point to a cutoff that makes more sense.

larseggert:
Yeah, I'll do some digging. It _seems_ like there should be a win here.

martinthomson:
I agree.  It's just not with 8 items, it seems.

One thing you might consider is landing this change with `type RecoveryTokenVec = Vec<RecoveryToken>` as the definition, so that future changes are one-liners.

larseggert:
I was about to do
```rust
pub type Tokens = SmallVec<[Token; 8]>;
```
and then use it e.g. like
```rust
let mut tokens = recovery::Tokens::new();
```

PR #2779, opened 2025-07-04
feat: Use `SmallVec` for `recovery::Token`

larseggert:
With size 16, just to see if it matters. Alternative to #1657.


larseggert:
It's worse with 16. I think the issue may be that the size of the object is getting (much) bigger, and we do clone tokens in a few places (`detect_lost_packets`, `maybe_fire_pto`) which probably becomes more costly.

PR #2780, opened 2025-07-04
feat: Put tokens into `Rc`

larseggert:
The theory here is that by putting `Tokens` into an `Rc`, we can avoid copying them when a `Packet` gets cloned.


martinthomson:
Do you need reference counting, or would `Box` suffice?

larseggert:
My theory was that `Box::clone` would do a `memcpy` whereas `Rc::clone` would not. Am I wrong?

mxinden:
Seems correct, yes.

That said, what is the intention behind this pull request? The following confuses me. `Token` uses a `SmallVec` now, i.e. is stack allocated. That stack allocation is then wrapped by a pointer (`Rc`) thus heap allocated.

larseggert:
It's a bit of an experiment. In 1f6d802f365fce6fed4304ad689e75a03adf1d55, I put the `Vec` into an `Rc` to see if that made a difference. I didn't expect the change to `SmallVec` to matter, and it seems like it doesn't.

martinthomson (approved):
Shame about the loss of `const`, but this seems fine.

larseggert:
I thought we had enabled a clippy lint to flag that already, but let me check.

larseggert:
We clone `Packet`s here
https://github.com/mozilla/neqo/blob/420a9311a0a50d7f7ef515818a138ec6508e0e2c/neqo-transport/src/recovery/mod.rs#L352
and here
https://github.com/mozilla/neqo/blob/420a9311a0a50d7f7ef515818a138ec6508e0e2c/neqo-transport/src/recovery/mod.rs#L842-L856
with a derived `Clone`, so I don't see a way to make that more explicit.

PR #2782, opened 2025-07-05
perf(transport): reserve profile.limit() space in send_buffer

mxinden:


mxinden:
This change does not show a positive performance improvement, thus closing it.

Related to it is https://github.com/mozilla/neqo/pull/2747, which instead of trying to guess the right allocation size before building a packet, gets rid of the allocation all together, replacing it with a long-lived send buffer.

PR #2784, opened 2025-07-07
ci: Debug/fix `cpupower` setting on GCP runners

larseggert:


larseggert:
This needs to be on `main` before it can be tested/debugged.

Issue #2283, opened 2024-12-12
perf(http3): don't re-allocate entire datagram to prefix with varint


mxinden:
When sending an HTTP3 datagram, we prefix it with the session id. In order to do so, we write the session ID into a new `Encoder` and then copy the entire datagram over into that `Encoder`.

https://github.com/mozilla/neqo/blob/3001a3a56f2274eaafaa956fb394f0817f526ae7/neqo-http3/src/features/extended_connect/webtransport_session.rs#L415-L418

We might be able to do better, e.g. passing session ID and datagram down separately.

mansf-osk:
This is not easily possible I think. We need ownership of the data in `QuicDatagrams::add_datagram` when we push it into the queue as a new `QuicDatagram` so we need to reallocate at some point anyways, might as well do it here while we prefix the session ID.

https://github.com/mozilla/neqo/blob/3001a3a56f2274eaafaa956fb394f0817f526ae7/neqo-transport/src/quic_datagrams.rs#L170

Maybe we could do something with lifetimes there, but I'm not experienced enough with Rust to assess if that would be at all feasible (and going by Max's *"if you're using lifetimes you're probably doing something wrong"* advice, probably not).

Issue #1206, opened 2021-08-17
TTry to avoid copying QuicDatagrams data 


ddragana:
The copy that happens here is a little unfortunate.  It might be a good idea to mark this for future investigation.  I don't think that a copy is going to be so bad for us, but we might want to enable a more efficient handling process.  That's a project though, so this is fine.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1201#discussion_r689962436_

larseggert:
Code moved around; copy now happens in https://github.com/mozilla/neqo/blob/ad8f3909d9ae96219e56a7c31e6822d6e64a6dc9/neqo-transport/src/events.rs#L195-L200

mxinden:
See also https://github.com/mozilla/neqo/issues/2283 and https://github.com/mozilla/neqo/issues/2854 for datagram copying on the HTTP/3 layer.

PR #2796, opened 2025-07-17
feat: implement MASQUE connect-udp

mxinden:
Implements [RFC 9298](https://datatracker.ietf.org/doc/html/rfc9298).

---

Planned follow-ups:

- Re-introduce `neqo-bin` client and server implementation (see https://github.com/mozilla/neqo/pull/2796/commits/d583a02b89fa673cb3c9e4b5f8e6299af4cabdd7).
- https://github.com/mozilla/neqo/issues/2843
- https://github.com/mozilla/neqo/issues/2932
- Add a benchmark test now that we have https://github.com/mozilla/neqo/pull/2890.
- Measure impact of https://github.com/mozilla/neqo/issues/2854, https://github.com/mozilla/neqo/issues/1206, https://github.com/mozilla/neqo/issues/2283.


echo.py
```diff
@@ -0,0 +1,12 @@
+import socket
+
+HOST = '127.0.0.1'  # Listen on all interfaces
+PORT = 12345      # Arbitrary non-privileged port
+
+with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as server_socket:
+    server_socket.bind((HOST, PORT))
+    print(f"UDP Echo Server listening on {HOST}:{PORT}")
+    while True:
+        data, addr = server_socket.recvfrom(1024)
```
martinthomson: Is this big enough?

echo.py
```diff
@@ -0,0 +1,12 @@
...
+        print(f"Received from {addr}: {data.decode()}")
+        server_socket.sendto(data, addr)
```
martinthomson: Is this [black](https://pypi.org/project/black/)?

neqo-http3/src/features/extended_connect/connect_udp.rs
```diff
@@ -0,0 +1,422 @@
...
+}
+
+impl SessionState {
+    pub const fn closing_state(&self) -> bool {
+        matches!(self, Self::FinPending | Self::Done)
+    }
+}
+
+#[derive(Debug)]
+pub struct ConnectUdpSession {
```
martinthomson: Consider naming this:

```suggestion
pub struct Session {
```

neqo-http3/src/server_events.rs
```diff
@@ -243,12 +243,23 @@ pub struct WebTransportRequest {
...
+    stream_handler: StreamHandler,
+}
+
 impl Display for WebTransportRequest {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "WebTransport session {}", self.stream_handler)
     }
 }
 
+impl Display for ConnectUdpRequest {
```
martinthomson: Ordering is a bit odd here.  Either all types first, then all implementations; or, all definitions for the same type together.

neqo-http3/tests/connect_udp.rs
```diff
@@ -0,0 +1,136 @@
...
+use neqo_crypto::AuthenticationStatus;
+use neqo_http3::{
+    ConnectUdpServerEvent, ConnectUdpSessionAcceptAction, Http3Client, Http3ClientEvent,
+    Http3Parameters, Http3Server, Http3ServerEvent, Http3State,
+};
+use test_fixture::{
+    anti_replay, fixture_init, now, CountingConnectionIdGenerator, DEFAULT_ADDR, DEFAULT_ALPN_H3,
+    DEFAULT_KEYS, DEFAULT_SERVER_NAME,
+};
+#[test]
```
martinthomson:
```suggestion

#[test]
```

echo.py
```diff
@@ -0,0 +1,12 @@
+import socket
+
+HOST = '127.0.0.1'  # Listen on all interfaces
+PORT = 12345      # Arbitrary non-privileged port
```
martinthomson: Consider using `process.argv[]` for these.  I'd want tests that include IPv6 as well.

larseggert:
Should we have some MASQUE bench?

mxinden:
Yes. I have a work-in-progress MASQUE proxy server implementation here. Adding a benchmark similar to the Download benchmark will thus be doable.

echo.py
mxinden: Removed file. Checked in for early testing only.

echo.py
mxinden: Removed file. Checked in for early testing only.

neqo-http3/tests/httpconn.rs
```diff
@@ -91,31 +91,6 @@ fn process_client_events(conn: &mut Http3Client) {
...
-    assert_eq!(hconn_c.state(), Http3State::Connected);
-    let out = hconn_s.process(out.dgram(), now()); // Handshake
-    let out = hconn_c.process(out.dgram(), now());
-    let out = hconn_s.process(out.dgram(), now());
-    // assert!(hconn_s.settings_received);
-    let out = hconn_c.process(out.dgram(), now());
-    // assert!(hconn_c.settings_received);
-
-    out.dgram()
-}
```
mxinden: Moved to `test-fixtures` (see below) in order to be callable from `neqo-http3/tests/connect_udp.rs`.

neqo-http3/src/connection.rs
```diff
@@ -797,12 +818,24 @@ impl Http3Connection {
...
+                Header::new(":authority", target.authority()),
+            ]
+        } else {
+            vec![
+                Header::new(":method", request.method),
+                Header::new(":scheme", target.scheme()),
+                Header::new(":authority", target.authority()),
+                Header::new(":path", target.path()),
+            ]
+        };
```
mxinden: I will revert this change before merging. It will be merged as part of https://github.com/mozilla/neqo/pull/2886.

neqo-http3/src/connection_client.rs
```diff
@@ -2302,7 +2357,7 @@ mod tests {
 
     #[test]
     fn settings_frame_on_push_stream() {
-        test_wrong_frame_on_push_stream(&[0x4, 0x4, 0x6, 0x4, 0x8, 0x4]);
+        test_wrong_frame_on_push_stream(&[0x4, 0x4, 0x6, 0x4, 0x8, 0x1]);
```
mxinden, line 2417: Default SETTINGS frame now includes connect-udp.

neqo-http3/src/connection_client.rs
```diff
@@ -2390,7 +2445,7 @@ mod tests {
 
     #[test]
     fn settings_frame_on_request_stream() {
-        test_wrong_frame_on_request_stream(&[0x4, 0x4, 0x6, 0x4, 0x8, 0x4]);
+        test_wrong_frame_on_request_stream(&[0x4, 0x4, 0x6, 0x4, 0x8, 0x1]);
```
mxinden, line 2505: Default SETTINGS frame now includes connect-udp.

neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -4,42 +4,29 @@
...
-                Self::Error(e)
-            }
-            CloseType::Done => Self::Clean {
-                error: 0,
-                message: String::new(),
-            },
-        }
-    }
-}
-
```
mxinden, line 42: Moved into new module `neqo-http3/src/features/extended_connect/session.rs`.

neqo-http3/src/features/extended_connect/webtransport_session.rs
mxinden, line 1: A WebTransport session and a ConnectUDP session have a lot of logic in common.

Common logic removed here, and extracted in to `neqo-http3/src/features/extended_connect/session.rs` such that it can be reused by both WebTransport sessions and ConnectUDP sessions.

neqo-http3/src/features/extended_connect/session.rs
mxinden, line 1: The majority of the code here is extracted from `neqo-http3/src/features/extended_connect/webtransport_session.rs` in order to be used by both a WebTransport session and a connect-udp session.

neqo-http3/src/frames/tests/reader.rs
```diff
@@ -58,18 +58,22 @@ impl FrameReaderTest {
 fn frame_reading_with_stream_settings1() {
     let mut fr = FrameReaderTest::new();
 
-    // Send and read settings frame 040406040804
+    // Send and read settings frame 040406040801
```
mxinden, line 61: HTTP/3 settings now contain CONNECT parameter.

mxinden:
This  pull request is ready for review. I have a couple of minor clippy and test changes pending, but those don't need to block review. Please let me know if there is any way I can make reviewing this large patch easier.

neqo-bin/src/client/mod.rs
```diff
@@ -602,20 +608,43 @@ pub async fn client(mut args: Args) -> Res<()> {
...
+        }) else {
             qerror!("No compatible address found for: {host}");
             exit(1);
         };
-        let mut socket = crate::udp::Socket::bind(local_addr_for(&remote_addr, 0))?;
+
+        let proxy_addr = args.proxy.as_ref().map(|proxy_url| {
+            let Origin::Tuple(_scheme, proxy_host, proxy_port) = proxy_url.origin() else {
+                panic!();
+            };
```
larseggert: `.expect()`?

neqo-http3/src/server_events.rs
```diff
@@ -371,6 +371,122 @@ impl WebTransportRequest {
...
+    /// The value will change over time depending on the encoded size of the
+    /// packet number, ack frames, etc.
+    ///
+    /// # Errors
+    ///
+    /// The function returns `NotAvailable` if datagrams are not enabled.
+    ///
+    /// # Panics
+    ///
+    /// This cannot panic. The max varint length is 8.
```
larseggert:
```suggestion
```
I don't see code that would panic in this function? Delete the comment?

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![expect(clippy::unwrap_used, reason = "This is example code.")]
+
+//! An HTTP/3 MASQUE connect-udp proxy client.
```
larseggert: The `proxied_http3.rs` file name doesn't scream "MASQUE" to me. If this is MASQUE-specific, maybe rename?

Also for the "proxied" term used in the code.

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
...
+
+use super::{Client, CloseState, Res};
+
+pub struct Handler {}
+
+impl Handler {
+    pub(crate) const fn new() -> Self {
+        Self {}
+    }
+}
```
larseggert: Can we derive and use `default()` instead?

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
```diff
@@ -286,7 +285,7 @@ impl WtTest {
     pub fn check_session_closed_event_server(
         &self,
         wt_session: &WebTransportRequest,
-        expected_reeason: &SessionCloseReason,
+        expected_reeason: &CloseReason,
```
larseggert:
```suggestion
        expected_reason: &CloseReason,
```

neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -117,5 +123,36 @@ impl ExtendedConnectFeature {
         self.feature_negotiation.enabled()
     }
 }
-#[cfg(test)]
-mod tests;
+
+#[derive(Debug, Default)]
+struct Listener {
+    headers: Option<(Vec<Header>, bool, bool)>,
```
larseggert, line 113: Would a struct instead of a tuple be more readable?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -0,0 +1,560 @@
...
+            state: State::Negotiating,
+            events,
+            protocol,
+        }
+    }
+
+    /// # Panics
+    ///
+    /// This function is only called with `RecvStream` and `SendStream` that also implement
+    /// the http specific functions and `http_stream()` will never return `None`.
```
larseggert: _Can_ this still panic?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    /// # Errors
+    ///
+    /// The function can only fail if supplied headers are not valid http headers.
+    ///
+    /// # Panics
+    ///
+    /// `control_stream_send` implements the  http specific functions and `http_stream()`
+    /// will never return `None`.
```
larseggert: _Can_ this still panic?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -0,0 +1,560 @@
...
+                self.session_id,
+                CloseReason::from(close_type),
+                None,
+            );
+        }
+    }
+
+    /// # Panics
+    ///
+    /// This cannot panic because headers are checked before this function called.
```
larseggert: Remove the coment then?

neqo-http3/src/client_events.rs
```diff
@@ -358,10 +412,18 @@ impl Http3ClientEvents {
...
+                self.insert(Http3ClientEvent::WebTransport(
+                    WebTransportEvent::Negotiated(succeeded),
+                ));
+            }
+            HSettingType::EnableConnect => {
+                self.insert(Http3ClientEvent::ConnectUdp(ConnectUdpEvent::Negotiated(
+                    succeeded,
+                )));
+            }
+            _ => qtrace!("HSetting {:?} {succeeded} not handled", feature_type),
```
larseggert, line 435: Is this not more serious?

neqo-http3/src/connection.rs
```diff
@@ -787,6 +807,7 @@ impl Http3Connection {
         output
     }
 
+    // TODO: Rename?
```
larseggert: TODO for now or later?

neqo-bin/src/client/mod.rs
mxinden: This is matching on a `Origin`, not a `Result` or `Option`, thus no `.expect()` available.

I can't say that I love the pattern I am using here. That said, I am missing a better alternative. Ideas?

larseggert:
There is also some test failures with obvious fixes.

neqo-bin/src/client/mod.rs
```diff
@@ -632,17 +661,48 @@ pub async fn client(mut args: Args) -> Res<()> {
...
-
-                Runner::new(real_local, &mut socket, client, handler, &args)
+                let mut client_args = args.clone();
+                // Don't forward proxy authorization to origin.
+                client_args.headers = client_args
+                    .headers
+                    .into_iter()
+                    .filter(|h| {
+                        let name_lower = h.name().to_ascii_lowercase();
+                        name_lower != "authorization" && name_lower != "proxy-authorization"
```
martinthomson: I don't think that you should strip Authorization from these.

neqo-bin/src/client/proxied_http3.rs
martinthomson: "Proxied" is fine.  It's accurate and doesn't need to use a particular brand name.  It's a proxy, that uses CONNECT-UDP, which we sometimes label "MASQUE".

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![expect(clippy::unwrap_used, reason = "This is example code.")]
```
martinthomson: Isn't this suppression already at the binary level?  I thought that the rule passed down to modules.

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    fn take_token(&mut self) -> Option<ResumptionToken> {
+        None
+    }
+}
+
+pub struct ProxiedHttp3 {
+    proxied_conn: Http3Client,
+    handler: super::http3::Handler,
```
martinthomson: If this were generic on `H : Handler` instead, could this be used for the other client code as well?

neqo-http3/src/features/extended_connect/mod.rs
mxinden: Just a copy:

https://github.com/mozilla/neqo/blob/67ad82d7ae7a72e67034ab285854522918a8b0af/neqo-http3/src/features/extended_connect/webtransport_session.rs#L507-L510

I will add a TODO for myself to change it once this pull request is merged.

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
...
+            local_addr,
+            remote_addr,
+            Http3Parameters::default()
+                .connection_parameters(
+                    ConnectionParameters::default()
+                        .datagram_size(1500)
+                        .pmtud(true),
+                )
+                .connect(true)
+                .http3_datagram(true),
```
martinthomson: Should we make these part of the defaults?

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,245 @@
...
+                            DatagramTracking::None,
+                        )
+                        .unwrap();
+                }
+            }
+        };
+
+        // Second, service the proxy connection.
+        let maybe_proxy_conn_callback =
+            match self.proxy_conn.process_multiple_output(now, max_datagrams) {
```
martinthomson: The names `proxy_conn` and `proxied_conn` are too similar.  Consider `outer` and `inner`.  Or `tunnel` and `inner`.  Or something else that isn't basically the same name.

neqo-http3/src/client_events.rs
mxinden: Among the `HSettingsType`s we negotiate, only subset emit a client event, more specifically only WebTransport and connect-udp.

https://github.com/mozilla/neqo/blob/67ad82d7ae7a72e67034ab285854522918a8b0af/neqo-http3/src/settings.rs#L33-L40

In case we introduce new features in the future that do require a client event, I added the `qtrace` above.

All others should be fine to ignore.

neqo-bin/src/client/proxied_http3.rs
mxinden: I added multiple commits, documenting the usage of MASQUE connect-udp as the proxy protocol in use here.

Would you like me to rename to symbols themselves? To me MASQUE is an implementation detail here and I don't think `masque_connect_udp_http3.rs` is easy to read compared to `proxied_http3.rs`.

No strong opinion.

mxinden:
Thanks for the review @larseggert. Mind taking another look?

neqo-bin/src/client/proxied_http3.rs
larseggert: I don't feel strongly either.

neqo-bin/src/client/proxied_http3.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct ProxiedHttp3 {
+    /// HTTP/3 connection to the origin, proxied through
+    /// [`ProxiedHttp3::proxy_conn`].
+    proxied_conn: Http3Client,
+    handler: super::http3::Handler,
+    /// HTTP/3 connection to the proxy server, providing a MASQUE connect-udp
+    /// session.
+    proxy_conn: Http3Client,
+    url: Url,
+    /// The MASQUE connect-udp session ID, i.e. the HTTP EXTENDED CONNECT stream ID.
```
larseggert:
```suggestion
    /// The MASQUE connect-udp session ID, i.e., the HTTP EXTENDED CONNECT stream ID.
```

neqo-common/src/datagram.rs
```diff
@@ -240,9 +240,22 @@ impl DatagramBatch {
...
+        })
+    }
+
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = Datagram<&mut [u8]>> {
+        self.d.chunks_mut(self.datagram_size).map(|d| Datagram {
+            src: self.src,
+            dst: self.dst,
+            tos: self.tos,
+            d,
+        })
```
larseggert: I guess there isn't some helper crate that would let us derive these?

neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -52,42 +44,56 @@ pub(crate) trait ExtendedConnectEvents: Debug {
...
+    ConnectUdp,
+}
+
+impl fmt::Display for ExtendedConnectType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::WebTransport => write!(f, "webtransport"),
+            Self::ConnectUdp => write!(f, "connect-udp"),
+        }
+    }
```
larseggert: You might be able to use https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html

neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -52,42 +44,56 @@ pub(crate) trait ExtendedConnectEvents: Debug {
...
-        clippy::unused_self,
-        reason = "This will change when we have more features using ExtendedConnectType."
-    )]
     pub const fn string(self) -> &'static str {
-        "webtransport"
+        match self {
+            Self::WebTransport => "webtransport",
+            Self::ConnectUdp => "connect-udp",
+        }
     }
```
larseggert: You might be able to use https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -5,534 +5,152 @@
...
     recv_streams: HashSet<StreamId>,
     role: Role,
 }
 
 impl Display for Session {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        write!(f, "WebTransportSession session={}", self.id)
+        write!(f, "WebTransportSession")
     }
 }
```
larseggert, line 42: Maybe https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html again.

neqo-http3/src/client_events.rs
```diff
@@ -6,29 +6,32 @@
...
     settings::HSettingType,
     CloseType, Error, Http3StreamInfo, HttpRecvStreamEvents, PushId, RecvStreamEvents, Res,
     SendStreamEvents,
 };
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum WebTransportEvent {
-    Negotiated(bool),
+    Negotiated(
+        /// Whether the negotiation succeeded.
```
martinthomson:
```suggestion
        /// Whether WebTransport was negotiated.
```

neqo-http3/src/client_events.rs
```diff
@@ -41,6 +44,28 @@ pub enum WebTransportEvent {
     },
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum ConnectUdpEvent {
+    Negotiated(
+        /// Whether the negotiation succeeded.
```
martinthomson:
```suggestion
        /// Whether CONNECT-UDP was negotiated.
```

neqo-http3/src/client_events.rs
martinthomson: I'd lift this to `info`, perhaps.  Probably `debug`.  No more than that.  Unless you can't imagine landing on this function in any other way.  I think that this is just a setting that we didn't understand.

neqo-http3/src/conn_params.rs
```diff
@@ -14,6 +14,8 @@ const QPACK_TABLE_SIZE_LIMIT: u64 = (1 << 30) - 1;
 const QPACK_MAX_BLOCKED_STREAMS_DEFAULT: u16 = 20;
 const MAX_PUSH_STREAM_DEFAULT: u64 = 0;
 const WEBTRANSPORT_DEFAULT: bool = false;
+/// Do not support HTTP Extended CONNECT by default.
+const CONNECT_DEFAULT: bool = false;
 const HTTP3_DATAGRAM_DEFAULT: bool = false;
```
martinthomson: It seems like we could flip these defaults, as I noted above.

For a client, these all require that the client take special steps to initiate something, so there is no risk in them just happening spontaneously.

For a server, the server still receives a request that they have to explicitly accept before anything happens.  They would need to handle datagrams as well, but the default is that those get ignored, so I can't see why this would be a problem if it were enabled.

The reason I say that is that having a single connection profile is far easier to reason about for building applications.  Then, if the feature is needed, either the connections you have established already have the feature or they don't and you don't need to think about making new connections with the capability enabled.

This is more true for Firefox than it is this code, but the logic that applies to Firefox applies to any application that uses QUIC.

martinthomson (commented):
I just realized that I'm not going to be able to do more on this.  I'm strapped for time this week and unlikely to be able to recover.  I've added some comments, but these are rough and not comprehensive.

I did want to ask whether it was possible to split this up a little more.  It's very big and it seems like some of the client/server parts could be disentangled from the core, at the very least.

neqo-bin/src/client/proxied_http3.rs
mxinden: "other client code" as in the [HTTP09](https://github.com/mozilla/neqo/blob/main/neqo-bin/src/client/http09.rs) implementation? That would allow us to proxy an HTTP09 connection, whereas currently we can only proxy an HTTP3 connection.

I consider the HTTP09 implementation to only be useful in the context of the QUIC interop runner. Thus I consider adding support here not worth the complexity. Do you think differently?

neqo-http3/src/conn_params.rs
mxinden: Agreed with the above. I suggest continuing this discussion in https://github.com/mozilla/neqo/issues/2932.

neqo-bin/src/client/proxied_http3.rs
mxinden: In favor. Let's discuss in https://github.com/mozilla/neqo/issues/2932.

neqo-bin/src/client/proxied_http3.rs
mxinden: Moved to the crate level (i.e. `neqo-bin/src/lib.rs`).

https://github.com/mozilla/neqo/pull/2796/commits/662524f13b0b20b1493670283c5f67db98759476

neqo-http3/src/features/extended_connect/mod.rs
mxinden: Good idea. See https://github.com/mozilla/neqo/pull/2796/commits/5a1be706fac0b6e19ca0372245bf889878229bcb.

neqo-http3/src/features/extended_connect/mod.rs
mxinden: Good idea. See https://github.com/mozilla/neqo/pull/2796/commits/5a1be706fac0b6e19ca0372245bf889878229bcb.

neqo-http3/src/features/extended_connect/webtransport_session.rs
mxinden: Unfortunately `strum`'s `Display` derive macro only supports `enum`s.

neqo-common/src/datagram.rs
mxinden: Given that this is based on the type specific `datagram_size`, I can not think of a generic way to do this.

The `std` library `chunks` and `chunks_mut` functions are already a big help.

larseggert:
Could we add some tests here? Also check the coverage report for functions that appear unused.

neqo-bin/src/client/proxied_http3.rs
mxinden: Renamed in 8209c59154cea6e8d7fa49a5c59599661123009c.

mxinden:
> I did want to ask whether it was possible to split this up a little more. It's very big and it seems like some of the client/server parts could be disentangled from the core, at the very least.

I pushed https://github.com/mozilla/neqo/pull/2796/commits/d583a02b89fa673cb3c9e4b5f8e6299af4cabdd7, removing all `neqo-bin` client and server changes for now. I will propose them as a separate pull request once this one is merged.

larseggert:
Any chance we can get coverage up a bit?

mxinden:
> Any chance we can get coverage up a bit?

I added multiple additional commits. Most of the missing coverage is not new, but simply due to moving WebTransport logic.

Given Lars approval and Martin being short on time, could we move forward here, thus unblock the corresponding Firefox efforts @larseggert and @martinthomson? I would then follow up on any reviews post merge.

larseggert:
Let's get this in.

mxinden:
@larseggert thanks. Would you mind doing a manual merge?

neqo-http3/src/features/extended_connect/mod.rs
mxinden: For the record, happening in https://github.com/mozilla/neqo/pull/3002.

PR #2802, opened 2025-07-21
fix(http3): reduce visibility of internal functions

mxinden:
Mark functions not used in Firefox as pub(crate). This enables us to detect unused code and simplifies future refactorings.


larseggert:
Is there a way to test whether this (1) has caught _all_ the respective functions and (2) something we can do to correctly tag future ones?

Also, should we do something similar for types?

mxinden:
> whether this (1) has caught all the respective functions

I can not think of an easy way.

> and (2) something we can do to correctly tag future ones?

Same as above.

> Also, should we do something similar for types?

We can. But here too, I am not aware of an automated way.

For types that are produced in mozilla/neqo and consumed in mozilla-firefox/firefox, we should keep their constructor private, thus noticing when they are never constructed.

In the meantime, fine moving forward here @larseggert?

Issue #2804, opened 2025-07-22
Stream benches have high variability


larseggert:
The `stream` benches in `neqo-http3/benches` have high variability and are hence not very useful. They should be disabled until they can be improved.

PR #2805, opened 2025-07-22
ci: Disable `stream` benches

larseggert:
Until they can be improved

See #2804

### DIFF
--- a/neqo-http3/Cargo.toml
+++ b/neqo-http3/Cargo.toml
@@ -11,6 +11,9 @@ description.workspace = true
 keywords.workspace = true
 categories.workspace = true
 readme.workspace = true
+# FIXME: These benches have a high rate of variability and are hence not very useful
+# in their current form. See https://github.com/mozilla/neqo/issues/2804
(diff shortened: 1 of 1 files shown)


mxinden:
Instead of commenting out the section in `Cargo.toml`, you could as well comment out this line:

https://github.com/mozilla/neqo/blob/13b17e4ea6c0057ab47119a182fb2b3e2589f939/neqo-http3/benches/streams.rs#L100

Thus the code would still be statically checked by CI.

I am also fine removing it completely for now. Given that we have version control, we can always revert. Seems to be the convention in this repository. I am not aware of any other dead code.

larseggert:
> Instead of commenting out the section in `Cargo.toml`, you could as well comment out this line:
> 
> https://github.com/mozilla/neqo/blob/13b17e4ea6c0057ab47119a182fb2b3e2589f939/neqo-http3/benches/streams.rs#L100

That doesn't quite work, because then I'd need to add an empty `main` function to the bench.

Issue #1372, opened 2022-10-06
Buffer WebTransport streams and datagrams if they arrive before the response for the CONNECT request arrives


ddragana:

larseggert:
CC @jesup

mxinden:
Corresponding draft section:

> Clients can, however, send a SETTINGS frame, multiple WebTransport CONNECT requests, WebTransport data streams, and WebTransport datagrams all within a single flight. As those can arrive out of order, a WebTransport server could be put into a situation where it receives a stream or a datagram without a corresponding session. Similarly, a client may receive a server-initiated stream or a datagram before receiving the CONNECT response headers from the server.To handle this case, WebTransport endpoints SHOULD buffer streams and datagrams until they can be associated with an established session.
>
> To avoid resource exhaustion, endpoints MUST limit the number of buffered streams and datagrams. When the number of buffered streams is exceeded, a stream SHALL be closed by sending a RESET_STREAM and/or STOP_SENDING with the WT_BUFFERED_STREAM_REJECTED error code. When the number of buffered datagrams is exceeded, a datagram SHALL be dropped. It is up to an implementation to choose what stream or datagram to discard.

https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#section-4.5

mxinden:
Note that the issue title includes both streams and datagrams that might arrive early. #2806 only fixes the issue for streams.

I still think we should close here. Datagrams need to be treated as best-effort by the sender anyways. Unless we have a good use-case, I don't think doing the same fix for datagrams is worth the complexity.

Please comment if you think otherwise or have a use-case.

afrind:
If a QUIC DATAGRAM is in the same QUIC packet as the 200 Ok, will it be correctly delivered to the application?  It would be kind of nice if this case was handled.

mxinden:
@afrind yes, as long as the 200 OK is before the WebTransport datagram in the QUIC packet, the client will correctly handle both.

https://github.com/mozilla/neqo/pull/2979 will add a test for it.

PR #2806, opened 2025-07-23
test(http3):  WebTransport handles new stream before session accept

mxinden:
Tests race condition in https://github.com/mozilla/neqo/issues/1372.

### DIFF
--- a/neqo-http3/tests/webtransport.rs
+++ b/neqo-http3/tests/webtransport.rs
@@ -332,3 +332,81 @@ fn wt_server_stream_bidi() {
         wt_server_stream.stream_id()
     );
 }
+
+#[test]
+fn wt_race_condition_server_stream_before_confirmation() {
+    let now = now();
+    let (mut client, mut server) = connect();
+
+    // Client creates a WebTransport session.
+    client
+        .webtransport_create_session(now, &("https", "something.com", "/"), &[])
+        .unwrap();
+    exchange_packets(&mut client, &mut server);
+    assert_eq!(server.process_output(now).dgram(), None);
+    while client.next_event().is_some() {}
+
+    // Server accepts the session, but hold back the UDP datagram.
+    let wt_server_session = server
+        .events()
+        .find_map(|event| {
+            if let Http3ServerEvent::WebTransport(WebTransportServerEvent::NewSession {
+                session,
+                ..
+            }) = event
+            {
+                Some(session)
+            } else {
+                None
+            }
+        })
+        .expect("Should receive WebTransport session request");
+    wt_server_session
+        .response(&WebTransportSessionAcceptAction::Accept)
+        .unwrap();
+    let server_accept_dgram = server
+        .process_output(now)
+        .dgram()
+        .expect("Expected server to produce session acceptance datagram");
+    assert_eq!(server.process_output(now).dgram(), None);
+
+    // Server creates a stream, but hold back the UDP datagram.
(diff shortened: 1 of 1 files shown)


neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -28,7 +28,12 @@ use crate::{
 
 #[derive(Debug, PartialEq)]
 enum SessionState {
-    Negotiating,
+    Negotiating {
+        /// Remote initiated streams received before session confirmation.
+        ///
+        /// [`HashSet`] size limited by QUIC connection stream limit.
+        pending_streams: HashSet<StreamId>,
```
mxinden, line 35: Might need some more thought as an attacker might open and then close more than the QUIC connection stream limit.

martinthomson: Given that this is just stream IDs and we limit those at the connection level, I'm not concerned.  The session will time out and these will all go away.  That seems like it's OK.

To reiterate an idea I had, maybe we can fix this at the API level.  Rather than return a session once the server confirms, we might be able to return a usable object before then.  It might delay the sending of anything, but it could at least surface incoming datagrams and streams through the API, so that things are not lost.

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -279,6 +286,28 @@ impl WebTransportSession {
...
+                            self.events.extended_connect_new_stream(
+                                Http3StreamInfo::new(
+                                    stream_id,
+                                    ExtendedConnectType::WebTransport
+                                        .get_stream_type(self.session_id),
+                                ),
+                                // Explicitly emit a stream readable event. Such
+                                // event was previously suppressed as the
+                                // session was still negotiating.
+                                true,
```
martinthomson, line 308: Are you sure that the stream truly is readable?  It is possible to have a stream created without data being sent.  Out of order data or flow control streams cause the stream to be created, but not be readable.

mxinden: Good point. Did some debugging. Turns out, we always emit a `DataReadable` on a new stream, whether the stream is actually readable or not.

I suggest tackling this in a separate pull request. I don't think it is high priority.

Tracked in https://github.com/mozilla/neqo/issues/2916.

neqo-http3/src/features/extended_connect/webtransport_session.rs
mxinden: > Given that this is just stream IDs and we limit those at the connection level, I'm not concerned. The session will time out and these will all go away. That seems like it's OK.

:+1: Let's keep as is then.

> To reiterate an idea I had, maybe we can fix this at the API level. Rather than return a session once the server confirms, we might be able to return a usable object before then. It might delay the sending of anything, but it could at least surface incoming datagrams and streams through the API, so that things are not lost.

Intuitively, I would expect this to be useful for power users only. Unless we have someone asking for it, I would assume it is not worth the complexity it introduces.

Either way, I don't think we should introduce it in this bug fix.

mxinden:
@larseggert or @martinthomson could you give this another review? I would like to get this fix into the next release.

larseggert:
@mxinden approved but needs a rebase.

mxinden:
Merged `main`.

Most lines here are covered. All lines related to the patch are. The missing ones have been uncovered before. @larseggert would you mind merging manually once more?

Issue #2807, opened 2025-07-23
Connection error: HTTP/3 error: HttpFrame


larseggert:
Bisected to 13b17e4ea6c0057ab47119a182fb2b3e2589f939

Repeat with, in the neqo directory:
```
docker build --progress=plain -f qns/Dockerfile -t ghcr.io/mozilla/neqo-qns:latest .
```
and then in the interop-runner
```
./run.py --client quic-go --server neqo -t handshake -d
```
Error is
```
server  | 1.025 INFO [Http3 server connection] Connection error: HTTP/3 error: HttpFrame
```
because likely a buffer length is being calculated or passed incorrectly.

larseggert:
If we debug why this is, we should also add a unit test for it, because this only showed up when I started taking the interop runner back into CI rotation.

mxinden:
Fixed by #2808 .

PR #2808, opened 2025-07-25
fix(interop): update ALPN before starting server

mxinden:
The Neqo server changes its ALPN depending on the QUIC interop test:

https://github.com/mozilla/neqo/blob/9e52e922343609dba5171c0adb869cff7bd8d3a0/neqo-bin/src/server/mod.rs#L431-L437

Thereafter it decides whether to spawn an HTTP3 server or an HTTP09 server:

https://github.com/mozilla/neqo/blob/531cb0b6c21d27ff1ccec5502ecc346a6a922d71/neqo-bin/src/server/mod.rs#L468-L474

https://github.com/mozilla/neqo/pull/2734 moved the latter in front of the other. Thus QUIC interop related ALPN overrides where no longer accounted for. We would spawn an HTTP3 server for tests requiring an HTTP09 server. The HTTP3 server would then incorrectly try to parse HTTP09 messages as `HFrame`s.

This pull request fixes the order, first updating the ALPN depending on the QUIC interop test, then spawning the correct server.

Given that this is specific to the QUIC interop tests only, I don't think a dedicated test is needed.

----

In addition, this pull request updates our CI workflow.  quic-interop-runner now requires docker v28.1.

See https://github.com/quic-interop/quic-interop-runner/pull/433

### DIFF
--- a/.github/actions/rust/action.yml
+++ b/.github/actions/rust/action.yml
@@ -113,8 +113,8 @@ runs:
       run: |
         for tool in $(echo $TOOLS | tr -d ","); do
           if [ "$tool" == "samply" ]; then
-            # TODO: Install released version once `--unstable-presymbolicate` is released.
-            cargo install --git https://github.com/mstange/samply samply
+            # TODO: Install released version once `--presymbolicate` (https://github.com/mstange/samply/pull/634) is released.
+            cargo install --git https://github.com/mstange/samply --rev f6ff5dedc73ab84a8ef45231f41cd3e721fb5cd4 samply
           else
             # FIXME: See https://github.com/Swatinem/rust-cache/issues/204 for why `--force`.
             cargo install --locked --force "$tool"
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -48,13 +48,13 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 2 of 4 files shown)


mxinden:
Actual error: `interface_name requires Docker Engine v28.1 or later`

mxinden:
> `server  |   TX -> ConnectionClose { error_code: Application(262), frame_type: 0, reason_phrase: "HTTP/3 error: HttpFrame" }`

Great. Interop is running again. Now showing https://github.com/mozilla/neqo/issues/2807.

I will look into https://github.com/mozilla/neqo/issues/2807 next.

//CC @larseggert to make sure we are not duplicating work.

martinthomson:
The benchmarks look like they are victim to version drift:

```
error: unexpected argument '--unstable-presymbolicate' found
  tip: a similar argument exists: '--presymbolicate'
```

larseggert:
@mstange likely updated `samply`.

mstange:
Oh you weren't using a pinned revision? Heh

I implemented proper presymbolication. So you can now update the script to use --presymbolicate and remove the packaging of the system.json files which are no longer produced.

larseggert:
No pinned version. YOLO.

We should also install via `binstall` again, instead of compiling from source.

mstange:
I still haven't made a release though.

mxinden:
https://github.com/mozilla/neqo/pull/2811 fixed the `samply` flag and pins the `samply` version. Thank you for the help everyone!

PR #2811, opened 2025-07-28
ci: fix samply unexpected argument '--unstable-presymbolicate'

mxinden:
Follow-up to https://github.com/mozilla/neqo/pull/2808#issuecomment-3127665044.

QNS failures will be fixed in https://github.com/mozilla/neqo/pull/2808.


mxinden:
//CC @mstange

PR #2814, opened 2025-07-30
SCONE

martinthomson:
This turned out to be relatively easy to implement.

Does doesn't do anything with the SCONE indications that are received; as this is intended for use in a browser, there isn't an easy way to plumb those through to something that might use those indications.  For the moment, these are just logged.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2894,6 +2894,12 @@ impl Connection {
         debug_assert_eq!(self.role, Role::Client);
         if let Some(path) = self.paths.primary() {
             qlog::client_connection_started(&mut self.qlog, &path, now);
+            qlog::recovery_parameters_set(
+                &mut self.qlog,
+                path.borrow().plpmtu(),
+                self.conn_params.get_cc_algorithm(),
+                now,
+            );
         }
         qlog::client_version_information_initiated(
             &mut self.qlog,
@@ -3516,6 +3522,7 @@ impl Connection {
             now,
         );
         let largest_acknowledged = acked_packets.first().map(sent::Packet::pn);
+        qlog::packets_acked(&mut self.qlog, space, &acked_packets, now);
         for acked in acked_packets {
             for token in acked.tokens() {
                 match token {
@@ -3589,6 +3596,12 @@ impl Connection {
             path.borrow_mut().set_valid(now);
             // Generate a qlog event that the server connection started.
             qlog::server_connection_started(&mut self.qlog, &path, now);
+            qlog::recovery_parameters_set(
+                &mut self.qlog,
+                path.borrow().plpmtu(),
+                self.conn_params.get_cc_algorithm(),
+                now,
+            );
         } else {
             self.zero_rtt_state = if self
                 .crypto
@@ -3625,12 +3638,16 @@ impl Connection {
     fn set_state(&mut self, state: State, now: Instant) {
         if state > self.state {
             qdebug!("[{self}] State change from {:?} -> {state:?}", self.state);
+            let old_state = self.state.clone();
             self.state = state.clone();
             if self.state.closed() {
                 self.streams.clear_streams();
             }
             self.events.connection_state_change(state);
-            qlog::connection_state_updated(&mut self.qlog, &self.state, now);
+            qlog::connection_state_updated(&mut self.qlog, &old_state, &self.state, now);
+            if let State::Closed(reason) = &self.state {
+                qlog::connection_closed(&mut self.qlog, reason, now);
+            }
         } else if mem::discriminant(&state) != mem::discriminant(&self.state) {
             // Only tolerate a regression in state if the new state is closing
             // and the connection is already closed.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -985,9 +985,9 @@ dependencies = [
 
 [[package]]
 name = "qlog"
-version = "0.15.2"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f15b83c59e6b945f2261c95a1dd9faf239187f32ff0a96af1d1d28c4557f919"
+checksum = "16689c500fede96aa7d5e1f8924abe08fafb65146d7e192826bb5fb57a1e1ad0"
 dependencies = [
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ hex = { version = "0.4", default-features = false }
 http = { version = "0.2.9", default-features = false }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
-qlog = { version = "0.15.1", default-features = false }
+qlog = { version = "0.16.0", default-features = false }
 quinn-udp = { version = "0.6", default-features = false, features = ["log"] }
 rustc-hash = { version = "2.1", default-features = false, features = [ "std" ]}
 static_assertions = { version = "1.1", default-features = false }
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -65,7 +65,7 @@ impl Qlog {
             None,
             now,
             new_trace(role),
-            qlog::events::EventImportance::Base,
+            qlog::events::EventImportance::Extra,
             Box::new(BufWriter::new(file)),
         );
         Self::enabled(streamer, qlog_path)
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -22,7 +22,7 @@ use crate::{
 };
 
 pub const CWND_INITIAL_PKTS: usize = 10;
-const PERSISTENT_CONG_THRESH: u32 = 3;
+pub const PERSISTENT_CONG_THRESH: u32 = 3;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum Phase {
@@ -194,6 +194,7 @@ impl<T> ClassicCongestionControl<T> {
 
 impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
     fn set_qlog(&mut self, qlog: Qlog) {
+        self.pmtud.set_qlog(qlog.clone());
         self.qlog = qlog;
     }
 
@@ -273,7 +274,6 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
 
             if self.current.phase.in_recovery() {
                 self.set_phase(Phase::CongestionAvoidance, None, now);
-                qlog::metrics_updated(&mut self.qlog, &[qlog::Metric::InRecovery(false)], now);
             }
 
             new_acked += pkt.len();
@@ -803,7 +803,6 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
             &[
                 qlog::Metric::CongestionWindow(self.current.congestion_window),
                 qlog::Metric::SsThresh(self.current.ssthresh),
-                qlog::Metric::InRecovery(true),
             ],
             now,
         );
--- a/neqo-transport/src/cc/mod.rs
+++ b/neqo-transport/src/cc/mod.rs
@@ -20,9 +20,7 @@ mod classic_cc;
 mod cubic;
 mod new_reno;
 
-#[cfg(test)]
-pub use classic_cc::CWND_INITIAL_PKTS;
-pub use classic_cc::ClassicCongestionControl;
+pub use classic_cc::{CWND_INITIAL_PKTS, ClassicCongestionControl, PERSISTENT_CONG_THRESH};
 pub use cubic::Cubic;
 pub use new_reno::NewReno;
 
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -1109,7 +1109,8 @@ impl Path {
 
     /// Update the `QLog` instance.
     pub fn set_qlog(&mut self, qlog: Qlog) {
-        self.sender.set_qlog(qlog);
+        self.sender.set_qlog(qlog.clone());
+        self.qlog = qlog;
     }
 }
 
--- a/neqo-transport/src/pmtud.rs
+++ b/neqo-transport/src/pmtud.rs
@@ -9,13 +9,13 @@ use std::{
     time::{Duration, Instant},
 };
 
-use neqo_common::{Buffer, qdebug, qinfo};
+use neqo_common::{Buffer, qdebug, qinfo, qlog::Qlog};
 use static_assertions::const_assert;
 
 use crate::{
     Stats,
     frame::{FrameEncoder as _, FrameType},
-    packet,
+    packet, qlog,
     recovery::{self, sent},
 };
 
@@ -59,6 +59,7 @@ pub struct Pmtud {
     probe_count: usize,
     probe_state: Probe,
     raise_timer: Option<Instant>,
+    qlog: Qlog,
 }
 
 impl Pmtud {
@@ -93,6 +94,22 @@ impl Pmtud {
             probe_count: 0,
             probe_state: Probe::NotNeeded,
             raise_timer: None,
+            qlog: Qlog::disabled(),
+        }
+    }
+
+    pub fn set_qlog(&mut self, qlog: Qlog) {
+        self.qlog = qlog;
+    }
+
+    fn set_mtu(&mut self, idx: usize, stats: &mut Stats, now: Instant) {
+        let old_mtu = self.plpmtu();
+        self.mtu = self.search_table[idx];
+        stats.pmtud_pmtu = self.mtu;
+        let new_mtu = self.plpmtu();
+        if old_mtu != new_mtu {
+            let done = !self.needs_probe();
+            qlog::mtu_updated(&mut self.qlog, old_mtu, new_mtu, done, now);
         }
     }
 
@@ -185,18 +202,20 @@ impl Pmtud {
 
         // A probe was ACKed, confirm the new MTU and try to probe upwards further.
         stats.pmtud_ack += acked;
-        self.mtu = self.search_table[self.probe_index];
-        stats.pmtud_pmtu = self.mtu;
-        qdebug!("PMTUD probe of size {} succeeded", self.mtu);
+        let confirmed_idx = self.probe_index;
+        qdebug!(
+            "PMTUD probe of size {} succeeded",
+            self.search_table[confirmed_idx]
+        );
         self.next(now, stats);
+        self.set_mtu(confirmed_idx, stats, now);
     }
 
     /// Stops the PMTUD process, setting the MTU to the largest successful probe size.
     fn stop(&mut self, idx: usize, now: Instant, stats: &mut Stats) {
         self.probe_state = Probe::NotNeeded; // We don't need to send any more probes
         self.probe_index = idx; // Index of the last successful probe
-        self.mtu = self.search_table[idx]; // Leading to this MTU
-        stats.pmtud_pmtu = self.mtu;
+        self.set_mtu(idx, stats, now); // Leading to this MTU
         self.probe_count = 0; // Reset the count
         self.raise_timer = Some(now + PMTU_RAISE_TIMER);
         qinfo!(
@@ -238,11 +257,10 @@ impl Pmtud {
     /// Starts PMTUD from the minimum MTU, probing upward.
     pub fn start(&mut self, now: Instant, stats: &mut Stats) {
         self.probe_index = 0;
-        self.mtu = self.search_table[self.probe_index];
-        stats.pmtud_pmtu = self.mtu;
         self.raise_timer = None;
-        qdebug!("PMTUD started, PLPMTU is now {}", self.mtu);
         self.next(now, stats);
+        self.set_mtu(0, stats, now);
+        qdebug!("PMTUD started, PLPMTU is now {}", self.mtu);
     }
 
     /// Starts the next upward PMTUD probe.
--- a/neqo-transport/src/qlog.rs
+++ b/neqo-transport/src/qlog.rs
@@ -13,22 +13,29 @@ use std::{
 
 use neqo_common::{Decoder, Ecn, hex, qinfo, qlog::Qlog};
 use qlog::events::{
-    EventData, RawInfo,
-    connectivity::{ConnectionStarted, ConnectionState, ConnectionStateUpdated},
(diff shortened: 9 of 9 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -468,14 +469,14 @@ impl ConnectionParameters {
         tps.local_mut()
             .set_integer(InitialMaxStreamsUni, self.max_streams_uni);
         tps.local_mut().set_integer(
-            TransportParameterId::IdleTimeout,
+            IdleTimeout,
```
larseggert, line 537: Unrelated?

neqo-transport/src/connection/params.rs
```diff
@@ -468,14 +469,14 @@ impl ConnectionParameters {
...
             u64::try_from(self.idle_timeout.as_millis()).unwrap_or(0),
         );
         if let PreferredAddressConfig::Address(preferred) = &self.preferred_address {
             if role == Role::Server {
                 let (cid, srt) = cid_manager.preferred_address_cid()?;
                 tps.local_mut().set(
-                    TransportParameterId::PreferredAddress,
+                    PreferredAddressTp,
```
larseggert: Unrelated?

neqo-transport/src/packet/mod.rs
```diff
@@ -644,32 +645,45 @@ impl<'a> Public<'a> {
...
+                        version: None,
+                        data,
+                    },
+                    &mut [],
+                ))
+            }
+            Version::SCONE1 | Version::SCONE2 => {
+                // Note: this outright ignores SCONE packets.
+                // It does not even validate that the connection ID is correct.
+                info!(
```
larseggert:
```suggestion
                debug!(
```
Or even `trace`?

neqo-transport/src/connection/params.rs
martinthomson: While the patient was open.  I wanted to make the usage consistent.

neqo-transport/src/connection/mod.rs
```diff
@@ -2817,6 +2811,36 @@ impl Connection {
...
+            profile.limit()
+        );
+        let pad_amount = profile.limit() - encoder.len();
+        initial.track_padding(pad_amount);
+        // This ensures that the last bytes are a SCONE indication, if there is enough space.
+        // This is not tracked, other than for congestion control (above)
+        if pad_amount >= Self::SCONE_INDICATION.len() {
+            encoder.pad_to(
+                profile.limit() - Self::SCONE_INDICATION.len() + 1,
+                Self::SCONE_INDICATION[0],
```
mxinden, line 2864: I understand we want to change the last two padding bytes to `0xc8` and `0x13`. 

Why change all other padding bytes to `0xc8`? For performance?

neqo-transport/src/connection/mod.rs
```diff
@@ -2817,6 +2811,36 @@ impl Connection {
...
+            );
+            encoder.encode(&Self::SCONE_INDICATION[1..]);
```
mxinden, line 2866: Do you expect `SCONE_INDICATOR` to increase in length in the future?

neqo-transport/src/connection/mod.rs
```diff
@@ -373,6 +373,8 @@ impl Connection {
     /// A long default for timer resolution, so that we don't tax the
     /// system too hard when we don't need to.
     const LOOSE_TIMER_RESOLUTION: Duration = Duration::from_millis(50);
+    /// The SCONE indicator.
+    const SCONE_INDICATION: &[u8] = &[0xc8, 0x13];
```
mxinden, line 351: I can't find these constants, nor the client -> server indicator mechanism, in the SCONE editors copy, SCONE draft 2, nor the TRAIN draft. Can you point me in the right direction?

neqo-transport/src/connection/mod.rs
```diff
@@ -2664,7 +2666,7 @@ impl Connection {
             let limit = if path.borrow().pmtud().needs_probe() {
                 needs_padding = true;
                 debug_assert!(path.borrow().pmtud().probe_size() >= profile.limit());
-                path.borrow().pmtud().probe_size() - aead_expansion
+                path.borrow().pmtud().probe_size() - aead_expansion - Self::SCONE_INDICATION.len()
```
mxinden: Do I understand correctly, that we are effectively giving up 2 bytes of every outgoing MTU-sized datagram?

neqo-transport/src/connection/mod.rs
martinthomson: They are in pull requests for the moment: 

https://github.com/ietf-wg-scone/scone/pull/51 for version
https://github.com/ietf-wg-scone/scone/pull/49 for indication
https://github.com/ietf-wg-scone/scone/pull/50 for transport parameter (now merged)

neqo-transport/src/connection/mod.rs
martinthomson: Only for packets that we need to pad (Initial and any probing packet).  I figure that this is worth it.

neqo-transport/src/connection/mod.rs
martinthomson: I'm taking precautions for now.  I expect that this will be reduced to a single assignment by the compiler.

neqo-transport/src/connection/mod.rs
martinthomson: Not performance, but I figured that it would be worth a change.  It changes how we treat the padding bytes (0xc8 is a long header packet), which should help with robustness tests on other implementations.

neqo-transport/src/packet/mod.rs
```diff
@@ -644,32 +645,45 @@ impl<'a> Public<'a> {
 
         // Generic long header.
         let version = Self::opt(decoder.decode_uint())?;
-        let dcid = ConnectionIdRef::from(Self::opt(decoder.decode_vec(1))?).into();
-        let scid = ConnectionIdRef::from(Self::opt(decoder.decode_vec(1))?).into();
+        let dcid = ConnectionIdRef::from(Self::opt(decoder.decode_vec(1))?);
+        let scid = ConnectionIdRef::from(Self::opt(decoder.decode_vec(1))?);
```
martinthomson, line 688: This part should have a modest positive performance impact, because it avoids assignments on packets that end up being discarded.

I'm a little concerned that we're seeing a drop in performance in benchmarks though; there are no other changes that should alter performance.  Unless you count having three more bytes of transport parameters.

martinthomson:
I'm still stumped on the performance hit here.  Any ideas?

mxinden:
Nothing comes to mind. I will try to reproduce locally.

mxinden:
Hypothesis:

- We no longer pad with `0x00`s, but `0xc8` instead.
- `0x00` padding is filtered out early in our processing pipeline
    https://github.com/mozilla/neqo/blob/4ad2f9a3a98db0b164b6106d79959831799dd84b/neqo-transport/src/packet/mod.rs#L619-L628
- `0xc8` is interpreted as a long packet, then later on discarded.
- I assume the additional processing itself is expensive. In addition we now have a `qwarn!` triggering in our hot path.
  https://github.com/mozilla/neqo/blob/4ad2f9a3a98db0b164b6106d79959831799dd84b/neqo-transport/src/connection/mod.rs#L1498-L1500
- Executing the `transfer` benchmark locally, stdout is spammed with the `qwarn!` above.

Can we special case `0xc8` in `packet/mod.rs` `encode` to discard the SCONE padding early @martinthomson?

larseggert:
We should also add a preference for this (default on).

neqo-transport/src/connection/mod.rs
larseggert: `c8` has more zeros than ones, so that _should_ still helps with filtering, but we might want to check (or make the value configurable).

martinthomson:
I'm not sure about that hypothesis: we don't send these packets very often and an all-0xc8 packet usually fails decryption (because it's a long header packet with a connection ID longer than the remaining bytes).  That's FASTER path to rejecting the packet than attempting to decrypt a short header packet.

As for the `qwarn!()`, we should downgrade that to `qinfo!()` or even `qdebug!()`, because dropping packets is totally normal.  I'll try that out.

martinthomson:
Oh, I just realized what might be going on here:

https://github.com/mozilla/neqo/blob/4ad2f9a3a98db0b164b6106d79959831799dd84b/neqo-transport/src/packet/mod.rs#L667-L680

This might be more work than following the short header path.  Maybe this needs a flag where that option produces an error instead.  That code only makes sense in the very early phases of server setup.

martinthomson:
Testing on my machine shows less-than-ideal performance, hinting at a slowdown, but I can't spot the problem still:

```
$ target/release/deps/transfer-ea5f6c8313464190 --bench -b main
transfer/pacing-false/varying-seeds
                        time:   [20.598 ms 20.658 ms 20.723 ms]
                        change: [−0.1420% +0.3377% +0.8482%] (p = 0.18 > 0.05)
  2 (2.00%) high mild
  1 (1.00%) high severe

transfer/pacing-true/varying-seeds
                        time:   [20.808 ms 20.894 ms 20.988 ms]
                        change: [+1.8460% +2.2808% +2.7827%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  8 (8.00%) high mild
  3 (3.00%) high severe

transfer/pacing-false/same-seed
                        time:   [20.479 ms 20.539 ms 20.608 ms]
                        change: [+0.9765% +1.3664% +1.7723%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  1 (1.00%) high mild
  6 (6.00%) high severe

transfer/pacing-true/same-seed
                        time:   [20.728 ms 20.785 ms 20.847 ms]
                        change: [−0.0452% +0.3311% +0.6927%] (p = 0.08 > 0.05)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

mxinden:
For what it is worth, I too am unable to reproduce the regression locally with the latest patches. Here using the `Download` benchmark:

```
critcmp main scone    
group                                         main                                   scone
-----                                         ----                                   -----
1-conn/1-100mb-resp (aka. Download)/client    1.01    299.1±6.00ms   334.3 MB/sec    1.00    297.1±3.57ms   336.6 MB/sec
```

larseggert:
Do the flamegraphs look different?

martinthomson:
@mxinden, are you able to help out here?  I know that this isn't high priority, but I really can't see how this change would materially affect performance.  I'm unable to get flamegraphs from my machine easily, so your help would be appreciated.

larseggert:
> I'm unable to get flamegraphs from my machine easily, so your help would be appreciated.

You can "[Download data for profiler.firefox.com](https://github.com/mozilla/neqo/actions/runs/16724643916/artifacts/3682410006)" and then (with [`samply`](https://github.com/mstange/samply) installed do, for example,
```
samply load neqo-neqo-cubic.client.perf.samply.json.gz
```

neqo-transport/src/connection/mod.rs
mxinden: Reverting this line resolves the performance regression I am seeing locally on this branch.

```
$ critcmp main scone scone-without-2                                                       
group                                         main                                   scone                                  scone-without-2
-----                                         ----                                   -----                                  ---------------
1-conn/1-100mb-resp (aka. Download)/client    1.01     49.3±4.30ms  2026.3 MB/sec    2.78    135.8±8.62ms   736.4 MB/sec    1.00     48.8±2.63ms     2.0 GB/sec
```

We reduce the `limit` here, accounting for an additional 2 bytes for SCONE, but then below stay within the limit, i.e. don't use the 2 bytes:

``` rust
        let pad_amount = profile.limit() - encoder.len();
```

Testing locally: we never fully pack the datagrams (missing 2 bytes), thus we never reach the PMTUD probe size and thus we never increase the MTU. 

@martinthomson is the ` - Self::SCONE_INDICATION.len()` necessary?

martinthomson: Yikes, that tiny change?  That only reserves two bytes in Initial packets for the indication.  That's a huge change in your stats.  I think that there's a mistake here.  It should only cut this for Initial.  Let me try changing that.

larseggert: Well, if it disables PMTUD, the effect will be substantial.

mxinden: Answering the question why the change visible here is so substantial:

- I did not limit the interface MTU on my local machine, i.e. with a working PMTUD the packet MTU goes up to 64k on my machine.
- The benchmark machine has an interface MTU of 1_500 bytes, thus the performance difference reported by the benchmark machine is significantly smaller.

martinthomson: That makes sense.  Given the recent change, I suspect that we'll be in better shape.  It won't disable probing in the same way because it will only affect Initial packets (where we don't do PMTUD, right?).

martinthomson: The recent change shows no difference in benchmarks.  Do you want to poke it again?  It seems like you correctly identified the performance bug and the change I made addresses that.

mxinden: > it will only affect Initial packets (where we don't do PMTUD, right?)

Correct. PMTUD is only enabled once the connection is confirmed:

https://github.com/mozilla/neqo/blob/b01d9662cf6756880b2888dd95af663f372f2c10/neqo-transport/src/connection/mod.rs#L3147-L3156

mxinden: > The recent change shows no difference in benchmarks.

Just to avoid any confusion, "no difference" compared to `main`. In other words, the performance regression is fixed. The latest benchmark results show no regression.

neqo-transport/src/packet/mod.rs
mxinden: > This part should have a modest positive performance impact, because it avoids assignments on packets that end up being discarded.

Not calling `into()` right away prevents a stack allocation only, right?

https://github.com/mozilla/neqo/blob/b01d9662cf6756880b2888dd95af663f372f2c10/neqo-transport/src/cid.rs#L88-L92

I want to make sure I am not missing something. I am fine either way.

neqo-transport/src/connection/mod.rs
```diff
@@ -2664,10 +2666,17 @@ impl Connection {
             let limit = if path.borrow().pmtud().needs_probe() {
                 needs_padding = true;
                 debug_assert!(path.borrow().pmtud().probe_size() >= profile.limit());
-                path.borrow().pmtud().probe_size() - aead_expansion
+                path.borrow().pmtud().probe_size()
+                    - if space == PacketNumberSpace::Initial {
+                        // Note that this will not reserve space for the indication
+                        // if packets are coalesced (with Handshake or 0-RTT).
+                        Self::SCONE_INDICATION.len()
```
mxinden, line 2698: I don't yet understand the subtraction of `Self::SCONE_INDICATION.len()` here.

It reserves two extra bytes for scone, taking them away from `profile.limit()` which is calculated here.

But then below, we make sure we stay within the limit of `profile.limit()` when writing the SCONE bytes. 

``` rust
        let pad_amount = profile.limit() - encoder.len();
```

In other words, we never use the 2 reserved bytes. Am I missing something?

martinthomson: Oh, I see what's going on here.  I should not do the reservation on this arm anyway.  Let me correct that.  Thanks for checking.  My logic was that this was "address validation" probing, but it's PMTUD.

martinthomson:
So this is now up to date with the latest in the spec.  And all the performance problems are addressed (thanks Max for picking that out).  We only IGNORE scone signals, so we don't need to worry about https://github.com/ietf-wg-scone/scone/pull/54 and friends.  The only concern is whether we can ship this with the unallocated codepoints.

larseggert:
Should the WG as for early IANA allocation?

martinthomson:
That's the plan, though there is so much bickering over irrelevant details, that I'm reluctant to do that.  (Which I guess answers the question about landing this...)

larseggert:
Add a connection parameter so we can land and pref it off?

larseggert:
@mxinden is in the process of vendoring in a neqo release, so merge what you want to land soon.

martinthomson:
Let's skip that release then.  It might be better to skip a release while things settle out.

larseggert:
Blocked by lack of final IANA codepoints.

larseggert:
@martinthomson any news re: IANA codepoints?

larseggert:
@martinthomson is this now ready to review/land, or are we still waiting for IANA codepoints?

martinthomson:
So, I think we should land it, but I have not yet sent the email to IANA.  I'm drafting it right now.

neqo-transport/src/connection/tests/priority.rs
```diff
@@ -244,7 +244,7 @@ fn critical() {
     let stats_before = server.stats().frame_tx;
     let dgram = server.process_output(now()).dgram();
     let stats_after = server.stats().frame_tx;
-    assert_eq!(stats_after.crypto, stats_before.crypto + 1);
```
larseggert, line 247: Gratuitous change (also below)? Or is the number of crypto frames now less predicable?

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -229,11 +229,15 @@ fn pto_handshake_complete() {
 
     now += HALF_RTT;
     let pkt = server.process(pkt, now).dgram();
-    assert_handshake(pkt.as_ref().unwrap());
+    // This is probably a Handshake packet, but it might also contain
+    // extra Initial data at the start.
+    // assert_handshake(pkt.as_ref().unwrap());
```
larseggert, line 234: Did you mean to leave this commented-out? (Also below.)

neqo-transport/src/connection/params.rs
larseggert: Gratuitous change?

test-fixture/src/lib.rs
```diff
@@ -477,19 +477,27 @@ pub fn split_datagram(d: &Datagram) -> (Datagram, Option<Datagram>) {
...
-        dgram
+
+    let mut remainder = &d[..];
+    while let (_first, Some(second)) = split_packet(remainder) {
+        if is_padding(second) {
+            return Datagram::new(
+                d.source(),
+                d.destination(),
+                d.tos(),
+                d[..d.len() - second.len()].to_vec(),
```
larseggert, line 495: Ask Claude to fix these mutants :-)

neqo-transport/src/connection/tests/priority.rs
martinthomson: Yes, because we are right up against the limit for CRYPTO frames in a server Initial.  I was seeing this fluctuate quite a bit.

neqo-transport/src/connection/tests/recovery.rs
martinthomson: I did, though I think that I left the comment because I wanted to discuss it.  Given where things stand, it seems like it might pay to have a `assert_contains_handshake()` or similar, which would ignore a small Initial packet.  There's a lot of fragility in our current tests due to packet sizing.  Maybe we'll get past that quickly and server Initial packets will grow past the point where they fit in a single datagram, but that seems fragile.

neqo-transport/src/connection/mod.rs
martinthomson: This has been resolved now, I think.

neqo-transport/src/connection/mod.rs
martinthomson: We're not currently filtering that way.  So we are reduced to treating `c8` as the length of connection IDs (because `c8` is a long header byte).  That leads to overflow problems.  I doubt we'd ever get a "real" packet from decoding a bunch of fixed bytes that start with `0x80`, because of the way we encode lengths, but the AEAD is our ultimate protection against this sort of garbage.  Anyone not relying on the AEAD might get into trouble.

neqo-transport/src/connection/params.rs
martinthomson: Yes, but one that improves consistency.  Happy to revert if you think it necessary.

neqo-transport/src/packet/mod.rs
martinthomson: Yeah, that's just a (small) allocation.  Not a big deal, but since I was here.

test-fixture/src/lib.rs
martinthomson: I'm not really interested in fixing this in the way that Claude would suggest: it's test code.  We could write a test for it, but I'm not sure that it's worth the effort.

The problem is that this indicates that we have tests that appear to rely on `strip_padding`, but they aren't checking the `dropped_rx` stat properly.

I'll ask Claude to look at tests that include `strip_padding` and see whether it can find ways to tighten up the assertions around `dropped_rx`.

martinthomson: I asked Claude to look at the tests.  It found an error in a comment.  That's it.

It took several iterations to get it to do something approximating what I wanted.

The underlying problem here is that we have direct tests of this function in the code, but those tests all depend on the padding being present.  With this change, we are right up against datagram size limits.  To fix those mutants changes, we'd need to directly test this function; the existing tests are not going to catch this.  Maybe we should do that, but I don't think that we need to.  The size of packets will inevitably change and we'll not need to worry about these particular mutants.  (We'll probably have a bunch of bigger problems with `assert_handshake()` calls to handle though.)

neqo-transport/src/connection/mod.rs
larseggert: No, I mean `c8` is probably not worse than `00` when it comes to confusing firewalls that count bit values to determine whether a packet is encrypted,

neqo-transport/src/connection/tests/priority.rs
larseggert: Would disabling randomizing initial packet numbers make things predicatble?

neqo-transport/src/connection/mod.rs
martinthomson: Ah, yeah, that's going to be a bit hard to predict.  I guess it depends on how many `c8` values are there.

neqo-transport/src/connection/tests/recovery.rs
larseggert: I made #3447 for this.

larseggert:
> So, I think we should land it, but I have not yet sent the email to IANA. I'm drafting it right now.

So we should wait with landing this until we have codepoints? Or land it and then change them in a follow-on PR?

martinthomson:
> So we should wait with landing this until we have codepoints? Or land it and then change them in a follow-on PR?

Given how QUIC registries work (registrant suggests a codepoint), I think we're good to land.  The risk of a codepoint change is low.

PR #2815, opened 2025-07-30
refactor(http3): improve `FrameReader` `Debug` implementation

mxinden:
`FrameReader` owns a 2048 byte `buffer`

``` rust
pub struct FrameReader {
    state: FrameReaderState,
    frame_type: HFrameType,
    frame_len: u64,
    buffer: [u8; MAX_READ_SIZE],
}
```

This buffer is only used when reading from a stream, then consecutively decoding off the buffer. There is no need to print the entire buffer in its `Debug` implementation. See sample logging output below.

```
0.036 DEBUG [RecvMessage stream_id:0] state=WaitingForResponseHeaders { frame_reader: FrameReader { state: GetLength { decoder: IncrementalDecoderUint { v: 0, remaining: None } }, frame_type: HFrameType(1), frame_len: 0, buffer: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ... ] } }
```

### DIFF
--- a/neqo-http3/src/frames/reader.rs
+++ b/neqo-http3/src/frames/reader.rs
@@ -7,8 +7,8 @@
 use std::{cmp::min, fmt::Debug};
 
 use neqo_common::{
-    hex_with_len, qtrace, Decoder, IncrementalDecoderBuffer, IncrementalDecoderIgnore,
-    IncrementalDecoderUint,
+    hex_snip_middle, hex_with_len, qtrace, Decoder, IncrementalDecoderBuffer,
+    IncrementalDecoderIgnore, IncrementalDecoderUint,
 };
 use neqo_transport::{Connection, StreamId};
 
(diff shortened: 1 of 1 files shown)


neqo-http3/src/frames/reader.rs
```diff
@@ -101,6 +100,16 @@ pub struct FrameReader {
     buffer: [u8; MAX_READ_SIZE],
 }
 
+impl Debug for FrameReader {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FrameReader")
+            .field("state", &self.state)
+            .field("frame_type", &self.frame_type)
+            .field("frame_len", &self.frame_len)
```
martinthomson: The alternative is to use `neqo_common::hex_snip_middle()` as follows:

```suggestion
            .field("frame", hex_snip_middle(&self.buffer[..self.frame_len]))
```

mxinden: Good idea. Thank you!

larseggert (approved):

PR #2816, opened 2025-07-31
Flip equality for const_min/const_max

martinthomson:
The specification for max() and min() when the arguments are equal has min() take the first argument and max() take the second.  Our code does the opposite.  Flip them.

This has no effect, other than to avoid propagating a logical error.


mxinden (approved):

mxinden:
Android failure seems unrelated. I can take a look later today. Happy to proceed here in the meantime.

larseggert (approved):

PR #2819, opened 2025-08-04
fix: Define all DSCP values

### LINKED ISSUE #2818: Dscp::from() doesn't cover all values
https://github.com/mozilla/neqo/blob/8c6524015e07834a72d566804041544d746e93ef/neqo-common/src/tos.rs#L119 
...has an `expect()` call. This throws if passed the wrong value.  If we happen to get an IP packet with an as-yet-unregistered DSCP value, we'll crash, such as:

```rust
_ = Dscp::from(0b0000_1100);
```

This is because we only cover 23 of the 64 possible values in the enum.  (The full range of 255 values isn't covered because we mask off the ECN bits.)

That's probably not what we want.  It's possible that operating systems will shield us from that, or that there is no plans to register more codepoints, ever, but I can't find evidence of that.  We should just define the rest of the codepoints in the enum.  The backward compatibility strategy there is a bit questionable if they ever grow a real name, but we can probably work something out then.

larseggert:
And add tests for all `u8` values for DSCP and ECN conversion confirming no panics.

Fixes #2818

### DIFF
--- a/neqo-common/src/tos.rs
+++ b/neqo-common/src/tos.rs
@@ -287,7 +287,7 @@ mod tests {
 
     #[test]
     fn u8_into_dscp_all() {
-        for i in 0..u8::MAX {
+        for i in 0..=u8::MAX {
             _ = Dscp::from(i);
         }
     }


neqo-common/src/tos.rs
```diff
@@ -210,6 +251,13 @@ mod tests {
         assert_eq!(Ecn::from(0b11), Ecn::Ce);
     }
 
+    #[test]
+    fn u8_into_ecn_all() {
+        for i in 0..u8::MAX {
```
martinthomson: Oops, I missed something!

```suggestion
        for i in 0..=u8::MAX {
```

neqo-common/src/tos.rs
```diff
@@ -237,6 +285,13 @@ mod tests {
         assert_eq!(u8::from(Dscp::Le), 0b0000_0100);
     }
 
+    #[test]
+    fn u8_into_dscp_all() {
+        for i in 0..u8::MAX {
```
martinthomson:
```suggestion
        for i in 0..=u8::MAX {
```

PR #2820, opened 2025-08-04
feat: Compile less of NSS

larseggert:


martinthomson:
It looks like this hit the cache.  Can you point to where this improves performance?  (I think this is probably worthwhile, but if we're using cached NSS, it might be close to zero benefit for CI builds.  Do you know if it makes local builds quicker?)

larseggert:
It saves me about 3 seconds locally (17.51 -> 14.78).

It will stop hitting the cache whenever the next NSS patch lands in its `main`.

martinthomson:
Good enough for me.  You should ship this given that the tests pass.

larseggert:
Thanks, needs a review.

PR #2824, opened 2025-08-04
ci: Create cache on `merge_group` trigger

larseggert:
Rust caches are currently only being created for workflows that have an `on push` trigger, which some don't (e.g., `bench.yml` and `perfcompare.yml`. Make it so that Rust caches are also generated for these.

### DIFF
--- a/.github/actions/rust/action.yml
+++ b/.github/actions/rust/action.yml
@@ -90,7 +90,8 @@ runs:
         cache-all-crates: true
         key: ${{ inputs.targets }}-${{ env.RUNNER_OS }}
         workspaces: ${{ inputs.workspaces }}
-        save-if: ${{ github.ref == 'refs/heads/main' || github.event.merge_group.base_ref == 'refs/heads/main' }} # Only cache runs from `main`.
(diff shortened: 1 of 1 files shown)


larseggert:
This will only have effect when merged to `main`.

.github/actions/rust/action.yml
```diff
@@ -90,7 +90,7 @@ runs:
         cache-all-crates: true
         key: ${{ inputs.targets }}-${{ env.RUNNER_OS }}
         workspaces: ${{ inputs.workspaces }}
-        save-if: ${{ github.ref == 'refs/heads/main' }} # Only cache runs from `main`.
+        save-if: ${{ github.ref == 'refs/heads/main' || github.event.merge_group.base_ref == 'refs/heads/main' }} # Only cache runs from `main`.
```
mxinden: Comment would need updating.

larseggert:
The perfcompare workflow currently doesn't seem to create a rust cache ever. Trying to fix that.

mxinden (approved):

PR #2825, opened 2025-08-04
Make initial RTT configurable

aochagavia:
Inspired by #2041, this PR makes the inital RTT configurable. It includes a handshake test to ensure the PTO is calculated based on the custom initial RTT.


aochagavia:
FYI I ran `cargo test` locally and saw all tests pass

aochagavia:
Just pushed a change that should fix the `cargo clippy` CI job.

Re the bot complaining about test coverage, I think what we have now is reasonable (but I'm open to suggestions from a human 😉).

mxinden:
Thank you for the contribution. Would you mind sharing what your use-case is? Are you as well experimenting with the space-application scenario?

neqo-bin/src/lib.rs
```diff
@@ -116,6 +116,10 @@ pub struct QuicParameters {
     /// The idle timeout for connections, in seconds.
     pub idle_timeout: u64,
 
+    #[arg(long = "init_rtt")]
+    /// The initial round-trip time. Defaults to [`DEFAULT_INITIAL_RTT`] if not specified.
+    pub initial_rtt_ms: Option<u64>,
```
mxinden:
```suggestion
    #[arg(long = "init_rtt", default_value = DEFAULT_INITIAL_RTT)]
    /// The initial round-trip time. Defaults to [`DEFAULT_INITIAL_RTT`] if not specified.
    pub initial_rtt_ms: u64,
```

Would this work as well?

aochagavia:
Indeed! I'm collaborating with the [tiptop IETF working group](https://datatracker.ietf.org/doc/charter-ietf-tiptop/) to test QUIC in a deep-space setting. Here's an [analogous PR](https://github.com/cloudflare/quiche/pull/2053) to Cloudflare's quiche, for reference.

neqo-transport/src/lib.rs
```diff
@@ -250,3 +253,6 @@ impl From<CloseError> for CloseReason {
 }
 
 pub type Res<T> = Result<T, Error>;
+
+// Defined in -recovery 6.2 as 333ms but using lower value.
+pub const DEFAULT_INITIAL_RTT: Duration = Duration::from_millis(100);
```
mxinden: Why move this out of `rtt.rs`?

I think I have a slight preference for defining this in `rtt.rs` as before and exposing it here as `pub use rtt::DEFAULT_INITIAL_RTT` at the top of the file.

neqo-bin/src/lib.rs
aochagavia: Not really, because `DEFAULT_INITIAL_RTT` is a `Duration` and we need a `u64` here

neqo-transport/src/lib.rs
aochagavia: Good idea, fixed

neqo-bin/src/lib.rs
larseggert: `DEFAULT_INITIAL_RTT.as_millis()` doesn't work?

larseggert:
> Re the bot complaining about test coverage, I think what we have now is reasonable (but I'm open to suggestions from a human 😉).

Would it be complicated to add tests to improve the coverage?

aochagavia:
I'm just guessing, but since coverage actually _increased_ with the test I added, and the bot mentions only 2 lines from my change are missing coverage, it must be referring to the `ConnectionParameters::get_initial_rtt` function (which is indeed not exercised by any test). I personally consider it redundant to test a one-line getter function, but I'm more than happy to add a test for it if you or other maintainers think it makes sense for neqo. What do you think?

neqo-bin/src/lib.rs
aochagavia: I played around with the code and the following diff would work:

```diff
     /// The idle timeout for connections, in seconds.
     pub idle_timeout: u64,
 
-    #[arg(long = "init_rtt")]
-    /// The initial round-trip time. Defaults to [`DEFAULT_INITIAL_RTT`] if not specified.
-    pub initial_rtt_ms: Option<u64>,
+    #[arg(long = "init_rtt", default_value_t = DEFAULT_INITIAL_RTT.as_millis() as u64)]
+    /// The initial round-trip time.
+    pub initial_rtt_ms: u64,
 
     #[arg(long = "cc", default_value = "cubic")]
     /// The congestion controller to use.
@@ -153,7 +153,7 @@ impl Default for QuicParameters {
             max_streams_bidi: 16,
             max_streams_uni: 16,
             idle_timeout: 30,
-            initial_rtt_ms: None,
+            initial_rtt_ms: DEFAULT_INITIAL_RTT.as_millis() as u64,
             congestion_control: CongestionControlAlgorithm::Cubic,
             no_pacing: false,
             no_pmtud: false,
@@ -230,10 +230,7 @@ impl QuicParameters {
             .max_streams(StreamType::BiDi, self.max_streams_bidi)
             .max_streams(StreamType::UniDi, self.max_streams_uni)
             .idle_timeout(Duration::from_secs(self.idle_timeout))
-            .initial_rtt(
-                self.initial_rtt_ms
-                    .map_or(DEFAULT_INITIAL_RTT, Duration::from_millis),
-            )
+            .initial_rtt(Duration::from_millis(self.initial_rtt_ms))
             .cc_algorithm(self.congestion_control)
             .pacing(!self.no_pacing)
             .pmtud(!self.no_pmtud)
```

Would you prefer this version over the `Option<u64>` one? The `as u64` cast is a bit ugly, but now the CLI help message is slightly clearer (it uses clap's `[default: ...]` instead of relying on a comment that might go stale in the future). Below is the relevant fragment:

```
      --init_rtt <INITIAL_RTT_MS>
          The initial round-trip time
          
          [default: 100]
```

larseggert:
It's the lines in `path.rs` and `connection/mod.rs`. See https://app.codecov.io/gh/mozilla/neqo/pull/2825?dropdown=coverage&src=pr&el=h1#29b8ec23a6da1889e637eb4b10691eae-R440

(I haven't checked whether they are in tests or not.)

aochagavia:
Ah, thanks for pointing out the specific files. Is it correct that the original lines didn't have coverage, and that this lack of coverage has now surfaced because of my changes? Here are my thoughts about the specific files btw:

1. The missing coverage in `path.rs` happens inside `Paths::rtt`, which is a test function (it has a `#[cfg(test)]` at the function signature). I assume we can treat that portion of the missing coverage as a false positive, in which test code is mistakenly being considered application code.
2. The missing coverage in `connection/mod.rs` happens inside `Connection::write_appdata_frames`. It seems that the specific line of code missing coverage is currently unreachable: the `else` closure of the `map_or_else` runs when there is no primary path available, but as far as I can see `write_appdata_frames` is only called when there _is_ a primary path. IMO we should either allow this missing line too, or otherwise refactor `write_appdata_frames` to take the primary path as a function parameter so we can get rid of the unreachable code altogether (the primary path is available at the call site). Let me know if you prefer the refactor route, then I'll push an additional commit implementing my suggestion.

neqo-bin/src/lib.rs
mxinden: Yes, preference for your diff above. Thank you.

mxinden:
> IMO we should either allow this missing line too

That is fine by me.

> or otherwise refactor `write_appdata_frames` to take the primary path as a function parameter so we can get rid of the unreachable code altogether (the primary path is available at the call site).

Sounds promising. I suggest experimenting with this in a separate pull request.

neqo-bin/src/lib.rs
aochagavia: Sure, just applied the diff and force-pushed

aochagavia:
I think this should be ready to go! Let me know if there's anything else that needs to change.

mxinden (approved):
:pray: Thanks. Will merge here tomorrow unless anyone has any comments in the meantime.

aochagavia:
The CI failed because of a clippy pedantic lint when compiling with the `bench` feature on. Just addressed it by adding an `expect` as shown below (in `neqo-bin/src/lib.rs`):

```rust
#[cfg(any(test, feature = "bench"))]
impl Default for QuicParameters {
    fn default() -> Self {
        Self {
            quic_version: vec![],
            max_streams_bidi: 16,
            max_streams_uni: 16,
            idle_timeout: 30,
            #[expect(
                clippy::cast_possible_truncation,
                reason = "DEFAULT_INITIAL_RTT.as_millis() is statically known to be < u64::MAX"
            )]
            initial_rtt_ms: DEFAULT_INITIAL_RTT.as_millis() as u64,
            congestion_control: CongestionControlAlgorithm::Cubic,
            no_pacing: false,
            no_pmtud: false,
            preferred_address_v4: None,
            preferred_address_v6: None,
            no_sni_slicing: false,
        }
    }
}
```

martinthomson:
We don't use `as` here, if we can avoid it.  Try using `TryFrom` instead.

aochagavia:
> We don't use `as` here, if we can avoid it. Try using `TryFrom` instead.

Sure, just replaced it by `u64::try_from(DEFAULT_INITIAL_RTT.as_millis()).unwrap_or(u64::MAX),`

martinthomson:
That's an awfully large default.  It might be better to use `expect()` instead.  It's not like the static value will ever trip that.

aochagavia:
All right, just pushed a version that uses `expect` instead (also rebased on top of main)

mxinden:
All of Martin's comments seem to be addressed. Merging here.

aochagavia:
@mxinden @larseggert @martinthomson Many thanks for finding the time to review and merge this!

PR #2827, opened 2025-08-06
feat: Optimize neqo-common

larseggert:
I'm playing around with Claude and asked it
> Check the neqo_common crate for any avoidable memory copies, and eliminate them.

This is what it came up with. Let's see what the benches say...

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -98,15 +98,47 @@ impl<'a> Decoder<'a> {
         if self.remaining() < n {
             return None;
         }
-        Some(if n == 1 {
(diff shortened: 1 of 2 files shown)


larseggert:
Seems to have no effect.

martinthomson (commented):
This seems like a bit of an F for Claude.

For the one thing, the copies are being made anyway.  And copies of less than 16 bytes is not enough to be relevant for any meaningful performance, because that is often just registers.

larseggert:
> And copies of less than 16 bytes is not enough to be relevant for any meaningful performance, because that is often just registers.

I think @claude learned this pattern from parsing things in a tight loop, where unrolling like this might help.

PR #2828, opened 2025-08-06
feat: Optimize `neqo-transport`

larseggert:
Similar to #2827.


neqo-transport/src/recovery/mod.rs
```diff
@@ -630,7 +630,7 @@ impl Loss {
             return (Vec::new(), Vec::new());
         };
         let loss_delay = primary_path.borrow().rtt().loss_delay();
-        let mut lost = Vec::new();
+        let mut lost = Vec::with_capacity(8); // Typically few packets are lost at once
```
martinthomson, line 633: Did Claude come up with this number on its own?  Because I have questions.  It's a reasonable thing to do, but a number as small as 8 is probably low enough that the first allocation will be that large anyway.  And if there are NO lost packets, you made things slower.

neqo-transport/src/connection/mod.rs
```diff
@@ -3211,7 +3214,7 @@ impl Connection {
                     .streams_mut()
                     .inbound_frame(space, offset, data)?;
                 if self.crypto.streams().data_ready(space) {
-                    let mut buf = Vec::new();
+                    let mut buf = Vec::with_capacity(16384); // Typical handshake message size
```
martinthomson, line 3226: This number I know to be wrong.  I know that we have certificates (which as a client we never send...) that can be large-ish, but not that large.  In practice, the size we might pre-allocate here would be around 2k, not 16k.

neqo-transport/src/recovery/mod.rs
```diff
@@ -873,7 +873,7 @@ impl Loss {
         let loss_delay = primary_path.borrow().rtt().loss_delay();
         let confirmed = self.confirmed();
 
-        let mut lost_packets = Vec::new();
+        let mut lost_packets = Vec::with_capacity(16); // Pre-allocate for typical PTO scenarios
```
martinthomson, line 876: Again, no evidence for the number.

neqo-transport/src/connection/mod.rs
```diff
@@ -2569,7 +2569,10 @@ impl Connection {
         max_datagrams: NonZeroUsize,
     ) -> Res<SendOptionBatch> {
         let packet_tos = path.borrow().tos();
-        let mut send_buffer = Vec::new();
+        let mut send_buffer = Vec::with_capacity(min(
+            DatagramBatch::MAX,
+            path.borrow().plpmtu() * max_datagrams.get(),
+        ));
```
martinthomson, line 2584: This looks pretty clever, until you realize that this is just copied from a few lines down.  Also, nope.

neqo-transport/src/connection/mod.rs
larseggert, line 3226: This is obviously way to big, but I wonder if 1500 makes sense.

neqo-transport/src/recovery/mod.rs
larseggert, line 633: This might make sense, but I wish we had some data on loss events rather than a guess.

neqo-transport/src/recovery/mod.rs
larseggert, line 876: Ditto.

martinthomson (commented):
Oops, I forgot to hit Submit.  Are you suggesting that this is a performance win, so we should take it, even if we have no idea why?  Jus' Vibe'n?

neqo-transport/src/recovery/mod.rs
larseggert: Yeah, this is @claude's number.

larseggert:
> Are you suggesting that this is a performance win, so we should take it, even if we have no idea why? 

There seem to be some modest improvements here from the preallocations, but the preallocation amounts should be informed by some data.

> Jus' Vibe'n?

Kinda. I'm asking @claude to do various things to see what it can do, and the (few) ones where the results look like they might the beginning of something I do a draft PR for, mostly so that benches run. (I tried to ask @claude to run benches locally to verify improvements - and even make new benches for each specific change it proposes - but it kinda fails hard when confronted with that ask.)

neqo-transport/src/connection/mod.rs
mxinden: Previous attempt with no sign of impact: https://github.com/mozilla/neqo/pull/2782

larseggert:
There doesn't seem to be anything here.

PR #2830, opened 2025-08-07
 feat: Optimize `neqo-bin`

larseggert:
Similar to #2827


martinthomson (commented):
The preallocation here probably won't hurt, but nor is it really going to make any meaningful difference.  The refactoring seems a bit off-spec, but it's fine.

larseggert:
This is @mxinden's code, so he can decide if this makes sense for him.

mxinden (commented):
Intuitively, I don't think these pre-allocations have an impact. The positive signal from the benchmarks isn't significant. The one benchmark where I would expect an improvement (HPS) shows a regression. Thus I suggest closing here without merge.

PR #2831, opened 2025-08-07
 feat: Optimize `neqo-qpack`

larseggert:
Similar to #2827


neqo-qpack/src/decoder.rs
```diff
@@ -54,7 +54,7 @@ impl Decoder {
             local_stream_id: None,
             max_table_size: qpack_settings.max_table_size_decoder,
             max_blocked_streams: usize::from(qpack_settings.max_blocked_streams),
-            blocked_streams: Vec::new(),
+            blocked_streams: Vec::with_capacity(qpack_settings.max_blocked_streams.into()),
```
martinthomson: Running the conversion twice is extra work.  This is probably fine, but it isn't going to do much, if anything, for performance.

neqo-qpack/src/huffman.rs
```diff
@@ -77,7 +77,7 @@ impl<'a> BitReader<'a> {
 /// Never, but rust can't know that.
 pub fn decode(input: &[u8]) -> Res<Vec<u8>> {
     let mut reader = BitReader::new(input);
-    let mut output = Vec::new();
+    let mut output = Vec::with_capacity(input.len()); // Huffman typically expands, so start with input size
```
martinthomson: This is a reasonably sensible starting point.  It might be even better if we had an understanding of what the typical expansion is.  Even doubling might do wonders here, given what we know about Huffman coding.

neqo-qpack/src/huffman.rs
```diff
@@ -116,7 +116,7 @@ fn decode_character(reader: &mut BitReader) -> Res<Option<u16>> {
 /// Never, but rust doesn't know that.
 #[must_use]
 pub fn encode(input: &[u8]) -> Vec<u8> {
-    let mut output: Vec<u8> = Vec::new();
+    let mut output: Vec<u8> = Vec::with_capacity((input.len() * 5) / 8 + 1); // Huffman compresses ~37.5%, with buffer
```
martinthomson: ... and here we have an actual formula.  A non-crazy one.  I wonder where it comes from.

neqo-qpack/src/header_block.rs
```diff
@@ -215,7 +215,7 @@ impl<'a> HeaderDecoder<'a> {
             );
             return Ok(HeaderDecoderResult::Blocked(self.req_insert_cnt));
         }
-        let mut h: Vec<Header> = Vec::new();
+        let mut h: Vec<Header> = Vec::with_capacity(8); // Typical number of headers in HTTP request/response
```
martinthomson: This is nonsense and probably won't help at all.

martinthomson (changes requested):
This one seems like it could at least help a little.  Though the Huffman encode/decode probably needs to be made symmetrical.  That is, if the expansion estimate is correct.  It probably isn't because the shortest bit sequence for a character is 5 bits, so it would be extremely optimistic to expect a header to be 5/8 of the length of the uncompressed byte sequence.

neqo-qpack/src/huffman.rs
```diff
@@ -116,7 +116,7 @@ fn decode_character(reader: &mut BitReader) -> Res<Option<u16>> {
 /// Never, but rust doesn't know that.
 #[must_use]
 pub fn encode(input: &[u8]) -> Vec<u8> {
-    let mut output: Vec<u8> = Vec::new();
+    let mut output: Vec<u8> = Vec::with_capacity(input.len()); // Huffman compresses, so this should be big enough to not reallocate.
```
martinthomson, line 119: It can reallocate when given pathological data, but we don't do that.  So this is probably fine.

neqo-qpack/src/decoder.rs
```diff
@@ -53,8 +54,8 @@ impl Decoder {
             send_buf,
             local_stream_id: None,
             max_table_size: qpack_settings.max_table_size_decoder,
-            max_blocked_streams: usize::from(qpack_settings.max_blocked_streams),
-            blocked_streams: Vec::new(),
+            max_blocked_streams,
+            blocked_streams: Vec::with_capacity(max_blocked_streams),
```
martinthomson, line 58: Want to try SmallVec here?

martinthomson (approved):

neqo-qpack/src/decoder.rs
larseggert: With what constant size?

martinthomson: You know, zero seems about right, the more I think on it.  (Which means reverting the change.)  Blocking should be avoided in most cases, so starting with no allocation is best.  If this ever needs to be allocated, paying a tax then is reasonable.

martinthomson: Oh, with one potential tweak: We could allocate the maximum if we ever need to allocate.  I think that you can use `Vec::reserve()` for that.

larseggert: I don't think that's what `reserve` does: https://users.rust-lang.org/t/is-vec-with-capacity-like-vec-new-with-vec-reserve-or-vec-new-with-vec-reserve-exact/80282. In other words, we might as well use `with_capacity` then it seems.

PR #2832, opened 2025-08-07
feat: Optimize `neqo-crypto`

larseggert:
Similar to #2827


neqo-crypto/src/agent.rs
```diff
@@ -459,9 +459,9 @@ impl SecretAgent {
             alert: Box::pin(None),
             now: TimeHolder::default(),
 
-            extension_handlers: Vec::new(),
+            extension_handlers: Vec::with_capacity(4), // Typical number of TLS extensions
```
larseggert: Likely too small.

martinthomson: WAT.   We add ONE.

We could move to `SmallVec<[1; _]>`, I guess.

neqo-crypto/src/agent.rs
```diff
@@ -459,9 +459,9 @@ impl SecretAgent {
...
 
-            ech_config: Vec::new(),
+            ech_config: Vec::with_capacity(1), // Usually 0 or 1 ECH config
```
martinthomson: Same here.

neqo-crypto/src/agent.rs
```diff
@@ -1040,7 +1040,7 @@ impl Client {
         let mut client = Self {
             agent,
             server_name,
-            resumption: Box::pin(Vec::new()),
+            resumption: Box::pin(Vec::with_capacity(256)), // Typical TLS resumption token size
```
martinthomson: This number is very wrong.  And trivially measurable.  We should be able to get traces with a better number.  Any of our resumption tests, with logging on, will spit out a value that we can round up by a little bit.

neqo-crypto/src/agentio.rs
```diff
@@ -222,13 +222,13 @@ pub struct AgentIo {
...
 impl AgentIo {
-    pub const fn new() -> Self {
+    pub fn new() -> Self {
         Self {
             input: AgentIoInput {
                 input: null(),
                 available: 0,
             },
-            output: Vec::new(),
+            output: Vec::with_capacity(1500), // Pre-allocate for typical TLS record output
```
martinthomson: Well, this is also wrong.  It's the analogue of the pre-allocation we'd use for CRYPTO frames.  It is not record size.

neqo-crypto/src/cert.rs
```diff
@@ -44,11 +44,12 @@ fn stapled_ocsp_responses(fd: *mut PRFileDesc) -> Option<Vec<Vec<u8>>> {
     let ocsp_nss = unsafe { SSL_PeerStapledOCSPResponses(fd) };
     match NonNull::new(ocsp_nss as *mut SECItemArray) {
         Some(ocsp_ptr) => {
-            let mut ocsp_helper: Vec<Vec<u8>> = Vec::new();
             let Ok(len) = isize::try_from(unsafe { ocsp_ptr.as_ref().len }) else {
                 qerror!("[{fd:p}] Received illegal OSCP length");
                 return None;
             };
+            #[expect(clippy::cast_sign_loss, reason = "OK because <= 2^24")]
+            let mut ocsp_helper: Vec<Vec<u8>> = Vec::with_capacity(len as usize);
```
martinthomson, line 52: This is solid.  I don't like the cast much, but the preallocation is legit.  (I should say, it won't do squat for performance though, so the extra complexity doesn't really pay off at all.)

neqo-crypto/src/agentio.rs
larseggert: In the one test (`neqo-crypto::agent basic`) where this isn't zero, it's 238.

neqo-crypto/src/agent.rs
larseggert: This is actually some number of `u8`s.

neqo-crypto/src/agent.rs
larseggert: Tokens are 840 or 856 bytes in tests.

martinthomson: I'd hedge and say 900.

neqo-crypto/src/agent.rs
martinthomson: Oh, so it's a byte array?  I'm going to suggest 64 then.

neqo-crypto/src/agentio.rs
martinthomson: That seems low.  Don't we have a handshake message with an ML-KEM share in it?

larseggert:
There doesn't seem to be any win here.

PR #2833, opened 2025-08-07
fix(bin): replace buggy to_headers with <Header as FromStr>::from_str

mxinden:
`neqo-bin`'s `to_headers` would never return an error due to misuse of `Iterator::scan` implementation.

This commit fixes the bug. In addition, it moves the implementation into a `FromStr` implementation on `Header` directly. Given the `FromStr` implementation, we can use it in `neqo-bin` client `Args` `clap` directly, no need for a dedicated parse.


neqo-common/src/header.rs
```diff
@@ -72,3 +74,60 @@ where
...
     }
 }
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum HeaderFromStrError {
+    MissingName,
+    MissingValue,
+}
+
+impl std::fmt::Display for HeaderFromStrError {
```
martinthomson: Want to use `thiserror` crate for this instead?  Do we use that already?

neqo-common/src/header.rs
```diff
@@ -72,3 +74,60 @@ where
...
+        }
+    }
+}
+
+impl std::error::Error for HeaderFromStrError {}
+
+impl FromStr for Header {
+    type Err = HeaderFromStrError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
martinthomson, line 98: Consider checking [RFC 9113](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.1) for some extra checks you could put in place for these.

neqo-common/src/header.rs
```diff
@@ -72,3 +74,60 @@ where
...
+        let mut parts = s.splitn(2, ':');
+        let name = parts
+            .next()
+            .ok_or(HeaderFromStrError::MissingName)?
+            .trim()
+            .to_lowercase();
```
martinthomson: I think that you want `to_ascii_lowercase()`

neqo-common/src/header.rs
mxinden: Neqo does not use `thiserror` today. I am very much in favor of `thiserror`, especially given its backtrace capability.

If we decide to introduce `thiserror`, I suggest we introduce it in a separate pull request.

To drive consensus: @larseggert I assume you don't object to introducing `thiserror` in Neqo?

larseggert: No objection, gecko is using it already. Separate PR makes sense.

neqo-common/src/header.rs
mxinden: We already have:

https://github.com/mozilla/neqo/blob/afef86a2eda8f5306a2850e2d4df50257ed9e9e5/neqo-http3/src/headers_checks.rs#L88-L154

Which is checked in:

https://github.com/mozilla/neqo/blob/afef86a2eda8f5306a2850e2d4df50257ed9e9e5/neqo-http3/src/send_message.rs#L63-L65

and:
https://github.com/mozilla/neqo/blob/afef86a2eda8f5306a2850e2d4df50257ed9e9e5/neqo-http3/src/recv_message.rs#L167

Let me know if you would like me to extend `headers_valid`. I would do it in a separate pull request.

neqo-common/src/header.rs
mxinden: Good point. Thanks. Done.

neqo-common/src/header.rs
mxinden: Tracked in https://github.com/mozilla/neqo/issues/2834.

larseggert (approved):

neqo-common/src/header.rs
```diff
@@ -72,3 +74,80 @@ where
...
+}
+
+impl std::fmt::Display for FromStrError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::MissingColon => write!(f, "Header string missing colon"),
+            Self::MissingName => write!(f, "Header string missing name"),
+        }
+    }
+}
```
mxinden, line 91: Codecov is complaining about these lines not being covered. I don't think we need tests for the `Display` implementation. Long term we will remove them with https://github.com/mozilla/neqo/issues/2834.

@larseggert in case you agree, would you mind manually merging here?

PR #2839, opened 2025-08-08
fix(transport/frame): prevent potential overflow in padding decode

mxinden:
Previously `neqo-transport` would use a `u16` to count the number of padding frames, i.e. `0`s. If an attacker sends more than `u16::MAX` `0`s, the counter would overflow.

This commit returns a `Error::TooMuchData` instead.

Note that the maximum UDP size is `u16::MAX`, thus in the real world, an attacker can not send more than `u16::MAX` padding frames.

---

Fixes https://github.com/mozilla/neqo/issues/2838.

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -72,6 +72,18 @@ impl<'a> Decoder<'a> {
         self.skip_inner(len);
     }
 
+    /// Skip while the current byte is `predicate`. Returns the number of bytes
+    /// skipped.
+    pub fn skip_while(&mut self, predicate: u8) -> usize {
+        let until = self
+            .as_ref() // remaining bytes
+            .iter()
+            .position(|v| *v != predicate)
+            .unwrap_or_else(|| self.remaining());
+        self.skip(until);
+        until
+    }
+
     /// Provides the next byte without moving the read position.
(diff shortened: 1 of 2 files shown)


neqo-common/src/codec.rs
```diff
@@ -835,6 +847,39 @@ mod tests {
         let mut dec = enc.as_decoder();
         dec.skip_vvec();
     }
+    #[test]
```
martinthomson, line 851:
```suggestion

    #[test]
```

neqo-transport/src/frame.rs
```diff
@@ -514,15 +514,11 @@ impl<'a> Frame<'a> {
...
-                    }
-                    length += 1;
-                    dec.skip(1);
-                }
-                Ok(Self::Padding(length))
+                // t itself + any additional `Frame::Padding`
+                (1 + dec.skip_while(u8::from(FrameType::Padding)))
+                    .try_into()
+                    .map(Self::Padding)
+                    .map_err(|_| Error::TooMuchData)
```
mxinden, line 521: Aside from the "bug fix", there might be a small performance improvement here, where instead of `skip(1)`ing each `0`, `skip_while` would skip all at once.

PR #2856, opened 2025-08-15
chore: Bump `sfv` to 0.14

larseggert:
So Gecko can move there, too.

Compile fails locally for reasons I don't understand.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -647,9 +647,9 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "2.10.0"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe4cd85333e22411419a0bcae1297d25e58c9443848b11dc6a86fefe8c78a661"
+checksum = "3954d50fe15b02142bf25d3b8bdadb634ec3948f103d04ffe3031bc8fe9d7058"
 dependencies = [
  "equivalent",
  "hashbrown",


larseggert:
Compile fails in CI, too. @valenting I think you are familiar with `sfv`, any clues?

valenting:
I think it's an incomplete implementation in sfv with some of the non-default features. I hit this too about a month ago, but then I started working on other things. Note my WIP patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1983230 for the Gecko updates.

I'll have a look at fixing sfv

valenting:
From what I can tell from my local build, I can bump the version if I use my local checkout of sfv. So all we'd need is a new release of sfv with the latest changes https://github.com/undef1nd/sfv/pull/212

valenting:
You should be able to update to 0.14
Let me know if that works.

larseggert:
I see the same failures with 0.14.

I can build 0.13 and 0.14 when I check out `sfv` separately. But when I try to build as part of `neqo`, I get these errors.

mxinden:
> I can build 0.13 and 0.14 when I check out sfv separately.

Can you get the version of `indexmap` on this checkout

> But when I try to build as part of neqo, I get these errors.

Then try to reuse that same version in neqo (`cargo update -p indexmap@xxx`) @larseggert?

larseggert:
Thanks, that was helpful. It fails with `indexmap@2.2` and works with `indexmap@2.10`.

mxinden:
`sfv` should require at least `indexmap@2.10` in its `Cargo.toml` then. Can you create a pull request? Happy to do it for you if you want.

PR #2857, opened 2025-08-15
ci: No `mold`

larseggert:
It takes longer to install the mold package then performing the link.


mxinden:
Also worth removing here to stay consistent across our various test and benchmark efforts?

https://github.com/mozilla/neqo/blob/d2ed80271ae5e39c9b1886caa849b09201b4f88c/test/ansible/install.yml#L80

PR #2861, opened 2025-08-17
perf: don't allocate in client UDP send path

mxinden:
Have client maintain a long-lived send buffer. When producing outbound UDP datagrams write them into the buffer. Then pass the buffer to the OS to be sent out on the network.

In other words, don't heap-allocate in the client UDP send path.

---

Corresponding past patch for the receive path https://github.com/mozilla/neqo/pull/2184

Fixes https://github.com/mozilla/neqo/issues/2670

Subset of https://github.com/mozilla/neqo/pull/2747

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -13,7 +13,7 @@ use std::{
     collections::VecDeque,
     fmt::Display,
     fs::File,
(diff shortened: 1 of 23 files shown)


mxinden:
Closing here, see https://github.com/mozilla/neqo/pull/2747#issuecomment-3194253414.

PR #2870, opened 2025-08-18
Move test_frame_writer use

martinthomson:
... to ensure that it can be used for handshake packet types.

Takes @larseggert's suggestion to assert that the packet isn't full.


larseggert:
@martinthomson the assertion seems to hit in the tests. I guess we're uncovering more issues...

martinthomson:
Great.  That meant that I needed to factor the test out into 5 different tests.  Worth it.

PR #2872, opened 2025-08-18
Force coverage up by simplifying a test

martinthomson:
This improves coverage and makes the test much, much shorter.


martinthomson:
Sorry for the churn, I realized that I could address the TODO at the same time.

PR #2877, opened 2025-08-19
Fix compatible upgrade

martinthomson:
This changeset should have zero obvious effect right now, but this is preparatory for the change that will randomize the first packet number.

This makes the whole version negotiation thing far more robust.  It ensures that we correctly track the received packet numbers across multiple crypto states at both client and server.

It changes how we confirm a compatible version upgrade, using the availability of Handshake keys as the definitive trigger.  At the server, that means that the server has processed the ClientHello (which contains the client transport parameters and that is what the server needs.  At the client, that means that the client has received at least one packet with the chosen value, which is what the client needs (any packet that makes progress on the handshake is enough).

This fixes two bugs:

1. We previously committed a version when receiving any CRYPTO frame. If we received an Initial packet with a TLS HelloRetryRequest, that would have been wrong.  That can happen if the server hadn't processed the ClientHello, but we were previously assuming that the server would have done so.

2. There are cases where a peer can send a v2 packet with a truncated packet number.  They make the choice to truncate based on the largest acknowledged, which might be > 256 if they are doing first packet number randomization.  Because we track the largest received packet number in the cryptographic send/receive state AND there is separate state for that for each version that might be used, if that truncated packet number is the first v2 packet we see, we fail to decrypt.  That situation is not recoverable.

There are now tests for these bugs, which are a little convoluted to setup, but not completely implausible.

I've retained the case for compatible upgrade if the packet version in an Initial packet changes during the handshake.  This is consistent with what the spec requires and it can avoid the need to do the fiddly stuff that follows.

Fixing the packet number continuation thing was easy on the server: it is the one that decides to use a new version, so it can commit before it ever needs to decrypt anything.  On the client, we have to be a little more careful.  There's a tiny performance hit for doing the update, which entails looking at the crypto state for every version, but it should be modest overall.

There was one other test that tests an error condition that becomes State::Closed rather than State::Closing as a result of deferring the version confirmation.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -3059,10 +3059,6 @@ impl Connection {
     }
 
     /// Commit to a particular version.
-    ///
-    /// `packet_version` is set to `Some` when the client is calling,
-    /// because that is what determines the version.
-    /// For a server, the version is taken from the transport parameters.
     fn compatible_upgrade(&mut self, packet_version: Version) -> Res<()> {
         if !matches!(self.state, State::WaitInitial | State::WaitVersion) {
             return Ok(());


neqo-transport/src/crypto.rs
```diff
@@ -916,11 +921,60 @@ impl CryptoStates {
...
 
+    /// When decrypting Initial packets, there are potentially multiple active versions.
+    /// The `used_pn` range tracks what has been received on the version that was used.
+    /// But if the version changes, the version we select might have a value of 0,
+    /// rather than the actual value, which can cause packet number recovery to fail.
+    /// To avoid that, have the indicated `version` continue from the previous version.
+    /// This only needs to be run once, so run it when getting header protection.
+    fn maybe_continue_initial_rx(&mut self, version: Version) {
+        // Only do this if this version hasn't been used...
+        // This would be better with `is_none_or`, but that needs rust 1.82
```
larseggert:
```suggestion
        // This would be better with `is_none_or`, but that needs MSRV >= 1.82
```
Because I search for "MSRV" in the code when it changes, to find spots to modify.

neqo-transport/tests/connection.rs
```diff
@@ -299,7 +299,7 @@ fn overflow_crypto() {
             return;
         }
     }
-    panic!("Was not able to overflow the crypto buffer");
+    panic!("unable to overflow the crypto buffer: {:?}", client.state());
```
larseggert:
```suggestion
    panic!("Unable to overflow the crypto buffer: {:?}", client.state());
```

larseggert (approved):

neqo-transport/src/connection/mod.rs
```diff
@@ -3064,31 +3054,38 @@ impl Connection {
...
-        self.crypto.confirm_version(v)?;
-        self.version = v;
-        Ok(())
+    const fn has_version(&self) -> bool {
+        self.crypto.has_handshake_keys()
     }
 
+    /// Commit to a particular version.
+    ///
+    /// `packet_version` is set to `Some` when the client is calling,
```
mxinden: `packet_version` is neither an `Option`, nor is it set in the function below. Is this comment outdated?

neqo-transport/src/crypto.rs
```diff
@@ -916,11 +921,60 @@ impl CryptoStates {
...
+        // having the borrow checker complain.
+        // It *ignores* errors from the `continuation()`
+        // so that the restore step isn't skipped.
+        //
+        // This doesn't need to be full anti-replay.
+        // Each version has separate keys, so nonce reuse is OK.
+        // After this, we might reject packets if the peer
+        // does reuse nonces, but they aren't allowed to do that.
+        //
+        // Note: these `if let Some(...)` conditions are always true.
```
mxinden, line 961: If so, why not `expect()` instead? Would allow us to catch if they ever aren't true.

martinthomson:
I'll leave this with @mxinden for a bit to give him a chance to contemplate whether this is the right approach.

mxinden (approved):
Thanks for the elaborate description. For what my review is worth, this looks good to me.

neqo-transport/src/connection/mod.rs
```diff
@@ -3064,31 +3054,38 @@ impl Connection {
...
+    /// because that is what determines the version.
+    /// For a server, the version is taken from the transport parameters.
```
```suggestion
```

neqo-transport/src/crypto.rs
martinthomson: This is literally impossible: see above where each is checked with `is_some_and()`.  The reason I didn't use expect is that it is more code in this case.  `self.initials[whatever].as_ref().expect("whatever").rx` or `self.initials[whatever].take().expect("whatever").rx` is nearly as much code all told, with the added ugliness of expect.

martinthomson:
The mutants call here suggests that we need a HelloRetryRequest test.  I'll open an issue to track that; it's not trivial.

martinthomson:
I'm going to take this, because it looks like my SCONE patch is blocked on a (new?) version negotiation bug.  The `reorder_handshake` test is failing because the server chooses QUICv2, but the client installs QUICv1 keys for the Handshake space.  Rather than dig in on that, I'm going to see if these changes help there.

martinthomson:
Looking at the coverage report here (the reason this bounced), these are not conditions that we can trigger.  But we have a policy of not using `.expect()` for stuff like that.  What gives?

larseggert:
Did this bounce because of coverage? I see green checkmarks.

martinthomson:
Apparently that was why.  Trying again, then I'll override.

PR #2880, opened 2025-08-19
chore: Remove dead code

larseggert:
Based on coverage. We should check if there is some code that _should_ have been called...

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -74,6 +74,7 @@ const fn alpn_from_quic_version(version: Version) -> &'static str {
 ///   - [`Http3Client::set_qlog`]
 /// - retrieving information about a connection:
 /// - [`Http3Client::peer_certificate`]
+///   - [`Http3Client::qpack_encoder_stats`]
 ///   - [`Http3Client::transport_stats`]
 ///   - [`Http3Client::state`]
 ///   - [`Http3Client::tls_info`]
@@ -89,6 +90,8 @@ const fn alpn_from_quic_version(version: Version) -> &'static str {
 ///   - [`Http3Client::cancel_fetch`]
 ///   - [`Http3Client::stream_reset_send`]
 ///   - [`Http3Client::stream_stop_sending`]
+/// - priority feature:
+///   - [`Http3Client::priority_update`]
 /// - `WebTransport` feature:
 ///   - [`Http3Client::webtransport_create_session`]
 ///   - [`Http3Client::webtransport_close_session`]


neqo-crypto/src/agent.rs
```diff
@@ -414,10 +414,6 @@ impl SecretAgentInfo {
     pub const fn alpn(&self) -> Option<&String> {
         self.alpn.as_ref()
     }
-    #[must_use]
-    pub const fn signature_scheme(&self) -> SignatureScheme {
```
martinthomson, line 418: Huh, I'd have thought that we'd want to have metrics for that in Firefox.  Oh well.

neqo-crypto/src/auth.rs
```diff
@@ -42,10 +42,4 @@ impl From<AuthenticationStatus> for PRErrorCode {
...
 }
 
-// Note that this mapping should be removed after gecko eventually learns how to
-// map into the enumerated type.
-impl From<PRErrorCode> for AuthenticationStatus {
-    fn from(v: PRErrorCode) -> Self {
-        Self::from_repr(v).unwrap_or(Self::Unknown)
-    }
-}
+//
```
martinthomson:
```suggestion
```

neqo-crypto/src/cert.rs
```diff
@@ -38,40 +29,6 @@ fn peer_certificate_chain(fd: *mut PRFileDesc) -> Option<ItemArray> {
     }
 }
 
-// As explained in rfc6961, an OCSPResponseList can have at most
-// 2^24 items. Casting its length is therefore safe even on 32 bits targets.
-fn stapled_ocsp_responses(fd: *mut PRFileDesc) -> Option<Vec<Vec<u8>>> {
```
martinthomson: Not having these in Firefox is probably OK.

neqo-crypto/src/cert.rs
```diff
@@ -38,40 +29,6 @@ fn peer_certificate_chain(fd: *mut PRFileDesc) -> Option<ItemArray> {
...
-                let item = unsafe { null_safe_slice((*itemp).data, (*itemp).len) };
-                ocsp_helper.push(item.to_owned());
-            }
-            Some(ocsp_helper)
-        }
-        None => None,
-    }
-}
-
-fn signed_cert_timestamp(fd: *mut PRFileDesc) -> Option<Vec<u8>> {
```
martinthomson: Not having this in Firefox might not be OK.  We're rolling out CT and I thought we needed this for that.

neqo-http3/src/connection.rs
```diff
@@ -940,36 +939,6 @@ impl Http3Connection {
         Ok(())
     }
 
-    /// Set the stream `SendOrder`.
-    ///
-    /// # Errors
-    ///
-    /// Returns `InvalidStreamId` if the stream id doesn't exist
-    pub fn stream_set_sendorder(
```
martinthomson: Are you *sure* that Firefox doesn't use this?  Because WebTransport is supposed to use these functions.

neqo-http3/src/connection_client.rs
```diff
@@ -789,32 +749,6 @@ impl Http3Client {
...
-    /// Sets the `SendOrder` for a given stream
-    ///
-    /// # Errors
-    ///
-    /// It may return `InvalidStreamId` if a stream does not exist anymore.
-    ///
-    /// # Panics
-    ///
-    /// This cannot panic.
-    pub fn webtransport_set_sendorder(
```
martinthomson, line 801: As above.  I'm fairly sure we'll need this for WebTransport.

neqo-transport/src/connection/mod.rs
```diff
@@ -507,13 +485,6 @@ impl Connection {
             .server_enable_0rtt(Rc::clone(&self.tps), anti_replay, zero_rtt_checker)
     }
 
-    /// # Errors
-    /// When the operation fails.
-    pub fn set_certificate_compression<T: CertificateCompressor>(&mut self) -> Res<()> {
```
martinthomson, line 512: I'm pretty sure that Firefox *is* using this one.

martinthomson (changes requested):
Are you absolutely certain that Firefox isn't using some of these?  There are some actual features this would neuter.  Certificate compression, stream prioritization in WebTransport, and CT are just a few of the things that look to be harmed here.

larseggert:
Yeah, good point. I'll check.

But we should really have new tests for those as well...

larseggert:
OK, I restored all the code that Gecko needs to be there.

neqo-http3/src/connection.rs
```diff
@@ -954,22 +954,6 @@ impl Http3Connection {
...
     }
 
-    /// Set the stream Fairness.   Fair streams will share bandwidth with other
-    /// streams of the same sendOrder group (or the unordered group).  Unfair streams
-    /// will give bandwidth preferentially to the lowest streamId with data to send.
-    ///
-    /// # Errors
-    ///
-    /// Returns `InvalidStreamId` if the stream id doesn't exist
-    pub fn stream_set_fairness(
```
martinthomson: This is likely also a gecko requirement.

neqo-http3/src/connection_client.rs
```diff
@@ -806,15 +764,6 @@ impl Http3Client {
         Http3Connection::stream_set_sendorder(&mut self.conn, stream_id, sendorder)
     }
 
-    /// Sets the `Fairness` for a given stream
-    ///
-    /// # Errors
-    ///
-    /// It may return `InvalidStreamId` if a stream does not exist anymore.
-    pub fn webtransport_set_fairness(&mut self, stream_id: StreamId, fairness: bool) -> Res<()> {
```
martinthomson, line 814: Also for WT.

neqo-http3/src/server_events.rs
```diff
@@ -350,11 +344,6 @@ impl WebTransportRequest {
             )
     }
 
-    #[must_use]
-    pub fn remote_datagram_size(&self) -> u64 {
```
martinthomson, line 354: I'm not 100% on this, but this might also be a WT requirement.

neqo-transport/src/connection/mod.rs
```diff
@@ -3777,11 +3757,6 @@ impl Connection {
         self.streams.keep_alive(stream_id, keep)
     }
 
-    #[must_use]
-    pub const fn remote_datagram_size(&self) -> u64 {
```
martinthomson: WT

neqo-transport/src/send_stream.rs
```diff
@@ -722,20 +722,6 @@ impl SendStream {
         ss
     }
 
-    pub fn write_frames<B: Buffer>(
```
martinthomson, line 725: This seems very wrong.  This is the bit where we write STREAM frames.

I see the trace call in logs often, I'd be happy if you deleted that...

larseggert:
So Gecko builds fine with all of these removed...

neqo-transport/src/send_stream.rs
larseggert: Are you sure that message isn't coming from `neqo_transport::send_stream::SendStreams::write_frames`?

We are also writing these frames in `neqo_transport::send_stream::SendStream::write_frames_with_early_return`.

neqo-http3/src/connection.rs
larseggert: It's not called in Gecko.

neqo-http3/src/connection_client.rs
larseggert: It's not called in Gecko.

neqo-http3/src/server_events.rs
larseggert: It's not called in Gecko.

neqo-transport/src/connection/mod.rs
larseggert: Same as above, it's not called in Gecko.

neqo-http3/src/connection_client.rs
martinthomson: Yes, though it should be.  I think that the problem is that we haven't used this in our WT implementation yet.  Maybe we can call on @jesup to help us with these.  I forget the details, exactly.

neqo-transport/src/send_stream.rs
martinthomson: Ah, you are right.  In that case, it can go, but can you rename write_frames_with_early_return to write_frames as part of this?

larseggert:
@martinthomson I restored the bits you flagged as probably needed for WT and did the rename.

PR #2881, opened 2025-08-19
fix: Don't call `finish` in `std::hash::Hash` implementations

larseggert:
At least I understand from the docs that is not needed: https://doc.rust-lang.org/std/hash/trait.Hash.html#implementing-hash

### DIFF
--- a/neqo-http3/src/server_events.rs
+++ b/neqo-http3/src/server_events.rs
@@ -42,7 +42,6 @@ impl std::hash::Hash for StreamHandler {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.conn.hash(state);
         state.write_u64(self.stream_info.stream_id().as_u64());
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

martinthomson:
Ignoring the bencher problems.  @larseggert is that going to be something we have to get used to?

larseggert:
> Ignoring the bencher problems. @larseggert is that going to be something we have to get used to?

I think so, unfortunately. Remember that I learned that the machines that were given to us for benchmarking were some that were taken out of the Gecko benchmarking pool due to having been upgraded to a BIOS version that made the results unstable :-(

I am discussing getting newer & faster machines with the SRE team. We _can_ already also run things in GCP, but it's unlikely to be more stable. (If you want to try, you can do manual runs for the `bench` and `perfcompare` workflows and select a `gcp` destination from the popup.)

PR #2884, opened 2025-08-21
Test and fix HRR bug

martinthomson:
I'm seemingly on a roll here.  I was looking to improve coverage, but I found a bug instead.  We were not able to do a compatible upgrade if we got HelloRetryRequest, because we were processing the transport parameters twice.

Now, it's OK to process parameters twice, but not if you crash the second time, which is what happened.

(We could check that they are the same, because they are supposed to be, but that's not worth the extra complexity given the likelihood that we'll ever see HelloRetryRequest in practice.)

### DIFF
--- a/neqo-transport/src/tparams.rs
+++ b/neqo-transport/src/tparams.rs
@@ -621,6 +621,7 @@ impl TransportParameters {
 pub struct TransportParametersHandler {
     role: Role,
     versions: version::Config,
+    version_selected: bool,
     local: TransportParameters,
     remote_handshake: Option<TransportParameters>,
     remote_0rtt: Option<TransportParameters>,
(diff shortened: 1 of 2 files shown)


neqo-transport/src/tparams.rs
```diff
@@ -666,49 +668,57 @@ impl TransportParametersHandler {
...
+        if self.version_selected {
+            // A second call to this is only possible on the server,
+            // if we need to send a TLS HelloRetryRequest.
+            debug_assert_eq!(self.role, Role::Server);
+            return Ok(());
+        }
+
+        let Some((current, other)) = remote_tp.get_versions() else {
+            return Ok(());
+        };
```
martinthomson, line 680: Early return here makes the changes look more significant than they are.  The change is the block above and the two places where the `version_selected` flag is set to `true`.

larseggert (approved):

PR #2885, opened 2025-08-21
Randomize the first packet number

### LINKED ISSUE #2462: Initial packet number need not be zero

martinthomson:
This is mostly the work of @larseggert, I'm just repackaging it.  This will target his branch with any changes.

This randomizes the starting packet number the client uses for the Initial packet number space.

We don't randomize this on the server, since otherwise we'd need even more changes to the tests to account for that.

Fixes #2462.
Closes #2499.

### DIFF
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -18,11 +18,11 @@ use enum_map::EnumMap;
 use neqo_common::{hex, hex_snip_middle, qdebug, qinfo, qtrace, Buffer, Encoder, Role};
 pub use neqo_crypto::Epoch;
 use neqo_crypto::{
-    hkdf, hp, Aead, Agent, AntiReplay, Cipher, Error as CryptoError, HandshakeState, PrivateKey,
-    PublicKey, Record, RecordList, ResumptionToken, SymKey, ZeroRttChecker, TLS_AES_128_GCM_SHA256,
-    TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, TLS_CT_HANDSHAKE, TLS_GRP_EC_SECP256R1,
-    TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1, TLS_GRP_EC_X25519, TLS_GRP_KEM_MLKEM768X25519,
-    TLS_VERSION_1_3,
+    hkdf, hp, random, Aead, Agent, AntiReplay, Cipher, Error as CryptoError, HandshakeState,
+    PrivateKey, PublicKey, Record, RecordList, ResumptionToken, SymKey, ZeroRttChecker,
+    TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, TLS_CT_HANDSHAKE,
+    TLS_GRP_EC_SECP256R1, TLS_GRP_EC_SECP384R1, TLS_GRP_EC_SECP521R1, TLS_GRP_EC_X25519,
+    TLS_GRP_KEM_MLKEM768X25519, TLS_VERSION_1_3,
 };
 
 use crate::{
@@ -477,8 +477,9 @@ impl CryptoDxState {
         epoch: Epoch,
         secret: &SymKey,
         cipher: Cipher,
+        min_pn: packet::Number,
     ) -> Res<Self> {
-        qdebug!("Making {direction:?} {epoch:?} CryptoDxState, v={version:?} cipher={cipher}",);
+        qdebug!("Making {direction:?} {epoch:?} CryptoDxState, v={version:?} cipher={cipher} min_pn={min_pn}",);
         let hplabel = String::from(version.label_prefix()) + "hp";
         Ok(Self {
             version,
@@ -486,8 +487,8 @@ impl CryptoDxState {
             epoch: usize::from(epoch),
             aead: Aead::new(TLS_VERSION_1_3, cipher, secret, version.label_prefix())?,
             hpkey: hp::Key::extract(TLS_VERSION_1_3, cipher, secret, &hplabel)?,
-            used_pn: 0..0,
-            min_pn: 0,
+            used_pn: min_pn..min_pn,
+            min_pn,
             invocations: Self::limit(direction, cipher),
             largest_packet_len: INITIAL_LARGEST_PACKET_LEN,
         })
@@ -498,6 +499,7 @@ impl CryptoDxState {
         direction: CryptoDxDirection,
         label: &str,
         dcid: &[u8],
+        min_pn: packet::Number,
     ) -> Res<Self> {
         qtrace!("new_initial {version:?} {}", ConnectionIdRef::from(dcid));
         let salt = version.initial_salt();
@@ -511,7 +513,7 @@ impl CryptoDxState {
 
         let secret = hkdf::expand_label(TLS_VERSION_1_3, cipher, &initial_secret, &[], label)?;
 
-        Self::new(version, direction, Epoch::Initial, &secret, cipher)
+        Self::new(version, direction, Epoch::Initial, &secret, cipher, min_pn)
     }
 
     /// Determine the confidentiality and integrity limits for the cipher.
@@ -732,6 +734,7 @@ impl CryptoDxState {
             CryptoDxDirection::Write,
             "server in",
             CLIENT_CID,
+            0,
         )
         .unwrap()
     }
@@ -794,7 +797,7 @@ impl CryptoDxAppData {
         cipher: Cipher,
     ) -> Res<Self> {
         Ok(Self {
-            dx: CryptoDxState::new(version, dir, Epoch::ApplicationData, secret, cipher)?,
+            dx: CryptoDxState::new(version, dir, Epoch::ApplicationData, secret, cipher, 0)?,
             cipher,
             next_secret: Self::update_secret(cipher, secret)?,
         })
@@ -1022,7 +1025,13 @@ impl CryptoStates {
 
     /// Create the initial crypto state.
     /// Note that the version here can change and that's OK.
-    pub fn init<'v, V>(&mut self, versions: V, role: Role, dcid: &[u8]) -> Res<()>
+    pub fn init<'v, V>(
+        &mut self,
+        versions: V,
+        role: Role,
+        dcid: &[u8],
+        randomize_first_pn: bool,
+    ) -> Res<()>
     where
         V: IntoIterator<Item = &'v Version>,
     {
@@ -1034,6 +1043,18 @@ impl CryptoStates {
             Role::Server => (SERVER_INITIAL_LABEL, CLIENT_INITIAL_LABEL),
         };
 
+        let min_pn = if randomize_first_pn {
+            let r = random::<4>();
+            let pn = packet::Number::from;
+            // A random starting packet number. 5 uniformly random bits.
+            // Then maybe add up to 256 to occasionally nudge into a second byte for
+            // both varint and packet number encodings, heavily weighted to small values.
+            // And a small offset to ensure that the value is always non-zero.
+            (pn(r[0]) & 0x1f) + (pn(r[1]) + pn(r[2]) + pn(r[3])).saturating_sub(512) + 1
+        } else {
+            0
+        };
+
         for v in versions {
             qdebug!(
                 "[{self}] Creating initial cipher state v={v:?}, role={role:?} dcid={}",
@@ -1041,8 +1062,8 @@ impl CryptoStates {
             );
 
             let mut initial = CryptoState {
-                tx: CryptoDxState::new_initial(*v, CryptoDxDirection::Write, write, dcid)?,
-                rx: CryptoDxState::new_initial(*v, CryptoDxDirection::Read, read, dcid)?,
+                tx: CryptoDxState::new_initial(*v, CryptoDxDirection::Write, write, dcid, min_pn)?,
+                rx: CryptoDxState::new_initial(*v, CryptoDxDirection::Read, read, dcid, 0)?,
             };
             if let Some(prev) = &self.initials[*v] {
                 qinfo!(
@@ -1063,9 +1084,14 @@ impl CryptoStates {
     /// This is maybe slightly inefficient in the first case, because we might
     /// not need the send keys if the packet is subsequently discarded, but
     /// the overall effort is small enough to write off.
-    pub fn init_server(&mut self, version: Version, dcid: &[u8]) -> Res<()> {
+    pub fn init_server(
+        &mut self,
+        version: Version,
+        dcid: &[u8],
+        randomize_first_pn: bool,
+    ) -> Res<()> {
         if self.initials[version].is_none() {
-            self.init(&[version], Role::Server, dcid)?;
+            self.init(&[version], Role::Server, dcid, randomize_first_pn)?;
         }
         Ok(())
     }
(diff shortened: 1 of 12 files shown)


larseggert (approved):

larseggert:
`benches/min_bandwidth.rs` failed. Not sure if this is a fluke. @mxinden?

mxinden:
Never seen it fail before. I thought I was quite conservative with 85% bandwidth usage only.

https://github.com/mozilla/neqo/blob/c535de9c66b7f99a50e5da12888c324434358d23/neqo-transport/benches/min_bandwidth.rs#L38

I re-triggered the benchmark. In case it is intermittent only, and there is no suspicion that this pull request introduces a regression, please ignore the failure. I will take a look tomorrow.

neqo-transport/src/crypto.rs
```diff
@@ -1034,15 +1043,27 @@ impl CryptoStates {
...
         };
 
+        let min_pn = if randomize_first_pn {
+            let r = random::<4>();
+            let pn = packet::Number::from;
+            // A random starting packet number. 5 uniformly random bits.
+            // Then maybe add up to 256 to occasionally nudge into a second byte for
+            // both varint and packet number encodings, heavily weighted to small values.
+            // And a small offset to ensure that the value is always non-zero.
+            (pn(r[0]) & 0x1f) + (pn(r[1]) + pn(r[2]) + pn(r[3])).saturating_sub(512) + 1
```
mxinden: This is clever., but hard to reason about. Out of curiosity, does it follow some standard pattern, or invented just for this use case?

neqo-transport/src/crypto.rs
```diff
@@ -1034,15 +1043,27 @@ impl CryptoStates {
             Role::Server => (SERVER_INITIAL_LABEL, CLIENT_INITIAL_LABEL),
         };
 
+        let min_pn = if randomize_first_pn {
+            let r = random::<4>();
+            let pn = packet::Number::from;
+            // A random starting packet number. 5 uniformly random bits.
```
mxinden: Why 5? I understand that we want some low number of bits to keep the variable length integer short.

neqo-transport/src/crypto.rs
mxinden: Why not just the below?

```suggestion
            pn(r[0].min(r[1]).min(r[2]).min(r[3])) + 1
```

- Easy to reason about.
- Always above `0`.
- 8 bits long in 6.25% (0.5^4) of the case, i.e. 2 byte in variable length encoding.
- Add or remove `min`s to increase or decrease probability for 2 byte long variable length encoding.

larseggert:
Still failing:
```
     Running benches/min_bandwidth.rs (target/debug/deps/min_bandwidth-05de51a29f1fe196)

thread 'main' (6959) panicked at neqo-transport/benches/min_bandwidth.rs:73:5:
expected to reach 0.85 of maximum bandwidth (1000 Mbit/s) but got 811.3804914445187 Mbit/s
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: bench failed, to rerun pass `-p neqo-transport --bench min_bandwidth`
Error: Process completed with exit code 101.
```

mxinden:
Taking a look. Will try to reproduce locally.

neqo-transport/src/crypto.rs
mxinden: Distribution should look something like this:

image

(missing the `+ 1`)

mxinden:
I can reproduce this locally via:

```
while cargo bench --bench min_bandwidth --features bench;do :; done
```

Around every 3rd iteration fails.

I can not reproduce this on `main`.

This is surprising.

larseggert:
Is there anything in our CC code that assumes packet numbers start at zero?

Just glancing at stuff, `first_app_limited` is for example initialized to `0`.

martinthomson:
Good question about CC.  Worth checking.

@mxinden we need the end of the tail to be >256 with non-negligible probability because that is the point at which a lot of the bugs I recently found trigger (because that is what gets the packet number encoding to truncate.

What you have never does that.  What we want is high probability of being less than 64, so we are typically most efficient, but then something that tends to spread across the range to about 512.  How about this?

```rust
pn(r[0] & 0x1f) + (pn(r[1].saturating_sub(224)) << 5) + 1
```

That's simple: most of the time it's even odds between 1 and 32 (inclusive).  Then about 1/8 of the time (tune as you like) the value is evenly distributed from 1 to 1024.  At 1024, we're well into the second byte for packet numbers and still well within the 2 byte encoding for varints.

That way, we're at about 3/32 (10%) over 256 and 15/128 (a little higher) over 64.

martinthomson:
Looking at `first_app_limited`, we start out not in the app limited state, so this code should just move the value up as needed:

https://github.com/mozilla/neqo/blob/b1c4293dde66f63993b9c13a8129ca71880ea015/neqo-transport/src/cc/classic_cc.rs#L379-L385

As long as we avoid being app limited (which doesn't depend on packet numbers) we should be OK.

Edit: the handshake starts out app limited anyway, because we are often sending less than half the congestion window.  That's true before as much as it is now.  We should tolerate the random PN easily, because we just bump the value up when sending a packet.  I can't see that being the issue.

mxinden:
Thanks for the explainer @martinthomson. Feel free to ignore the below. I mostly want to challenge my own understanding.

> pn(r[0] & 0x1f) + (pn(r[1].saturating_sub(224)) << 5) + 1

If I understand the `saturating_sub(224)` correctly it gives us a range of `0..31` in the bottom 5 bits. With the `<< 5` and `r[0]r & 0x1f` we thus get the "high probability of being less than 64", while occasionally going up to 1024.

Intuitively I would have said `saturating_sub(224)` is the same as a `>> 3` followed by a `<< 5`, but that would give us a uniform distribution across the possible values, instead of a distribution heavily biased towards `0` (i.e. with `r[0] & 0x1f` heavily biased towards `< 64`).

Both solutions are fine with me, preference for the latter. Maybe you can add a sentence to explain `saturating_sub(224)` for future me.

martinthomson:
OK, I've fixed the fuzzing issue, but the bandwidth benchmark is still stuck.  What is most interesting is that disabling randomization doesn't change the outcome.  The problem is in other parts of the code.

The code on main doesn't get as high a congestion window in my test, but it hits the first loss much sooner.  The main code pushes more packets through and loses more of them as a result.

I logged the cwnd increases and there are 4k+ times that it is increased by a little on main.  On the branch, it increases less than 200 times.  So, while the starting point is higher on the branch, it seems like it isn't increasing the congestion window in congestion avoidance anywhere near as much on the branch.

larseggert:
CC @omansfeld, see @martinthomson's latest comment above - got any ideas?

martinthomson:
OK, I have it.  I don't understand it completely though.  It seems like our Cubic implementation is a bit unreliable at dialing in a bandwidth estimate.  Stability relies a LOT on the path characteristics being stable and our test here isn't that.

On main, we set a target ACK rate of 25.5 times per RTT.  That's an awful lot, but it means that we end up with a lot of high quality feedback.  This patch changed to 4 times per RTT.  That's a big improvement in terms of its ability to operate efficiently.  

I'm still collecting results, but I'll create a table here:

| ACKS per RTT | Bandwidth | SSthresh | Drops |
|----|----|----|----|
| 1 | 837.8016171598957 | n/a | 0 |
| 2 | 869.3036825291894 | 9710318, 6797222, 4768404 | 2199 |
| 3 | 833.6612203683653 | 8987533 | 105 |
| 4 | 812.1666041813006 | 9164203 | 312 |
| 6 | 859.5269729211594 | 9152156, 6406509, 4494042 | 1771 |
| 8 | 866.8427432985857 | 8483680 | 40 |
| 12 | 876.1943349293886 | 12511871, 8758309, 6142027 | 3617 |
| 16 | 874.6212767488254 | 12508118, 8755682, 6668839 | 3130 |
| 20 | 876.2535979874053 | 12518751, 8763125, 6145398 | 3874 |
| 24 | 870.5149282368627 | 12361204, 8652842, 6667568, 4681958 | 3813 |

I'm only logging the cwnd when we leave slow start, which sometimes doesn't happen at all, but generally happens first at close to the same value (~9M in this test).  Two interesting things:

1. We seem to sit around the same point once we hit 10-ish acks per RTT, whereas it is unstable before that point.
2. The state seems to exit slow start more often as the bandwidth increases in this test.  Why the congestion controller goes back to slow start is not obvious without more logging (I can't log very much here without the logs being very large).

This is a very simple simulation, so it's not really indicative of anything in particular, we have a fixed delay and a fixed-depth tail-drop queue that simulates a gigabit link.  There's no jitter, we aren't doing PMTUD, and there is no baseline drop rate other than what happens if we exceed the link capacity.  I used a fixed seed for the simulator, but that seed won't be used in this set; the variation we're getting is from things like the connection ID size, which can vary between runs.

martinthomson:
An observation here is that @mxinden's estimate of achievable throughput seems pretty aggressive.  Though the theoretical throughput of a 1Gb/s is high, our overheads are around 48 (IPv6) + 5 (short + 4 byte packet number) + 16 (tag) + 5 (stream frame) = 74, reducing that to around 940Mb/s.  But the test seems like it settles on 870&ndash;880Mb/s.

It seems like the taildrop queue here is a problem, allowing the bandwidth to spike up.  I'm going to take another look at the setup and see if it can't be made a little more sensible.

mxinden:
> This patch changed to 4 times per RTT.

Oh, how does this patch, i.e. the randomization of packet numbers, lead to a reduction in the ACK frequency?

martinthomson:
> Oh, how does this patch, i.e. the randomization of packet numbers, lead to a reduction in the ACK frequency?

It was an accident.  I changed the test setup to remove the `ack_rate(255)` in the setup.  Which was something that couldn't be overridden, so it seemed like a good idea to remove it.  But then I failed to notice that this was a pretty substantial change.

martinthomson:
OK, so I changed the buffer in the taildrop queue from 5Mb to 0.5Mb.  The results are very interesting.  Overall throughput tanks and we get a whole lot of new warnings.

The logs warn about dropping unacknowledged ACK ranges at all ACK rate settings, which suggests that there's something pretty unhealthy with this arrangement.  We obviously have a lot of packets in flight and we're not getting enough ACK frames through.  I wonder if we shouldn't be sending PING occasionally to help avoid this problem.

| ACKS per RTT | Bandwidth | SSthresh | Drops | unackdrop |
|----|----|----|----|---|
| 1 | 344.96659873445 | 2886452, 2025690, 1419707 | 2236 | 12 |
| 2 | 336.61208527596307 | 2656172, 1860182, 1306439 | 2540 | 0 |
| 3 | 347.59838007400384 | 2910687, 2038343, 1429427 | 2537 | 0 |
| 4 | 315.80611372231414 | 3039456, 2128481, 1489936, 1045542 | 2320 | 131 |
| 6 | 323.061221953434 | 3034733, 2125175, 1487622, 1043922 | 2325 | 132 |
| 8 | 339.9060953139216 | 1727903, 1214706 | 687 | 28 |
| 12 | 327.2888327880112 | 3039481, 2128499, 1489949, 1045551 | 2418 | 123 |
| 16 | 328.4334509604926 | 3038452, 2128641, 1490048, 1045620 | 2400 | 130 |
| 20 | 328.8943037679999 | 3037414, 2127914, 1489539, 1045264 | 2445 | 127 |
| 24 | 317.0190205281322 | 2456640, 1722235, 1206426, 849672 | 2622 | 0 |

mxinden:
> OK, so I changed the buffer in the taildrop queue from 5Mb to 0.5Mb. 

Given that we currently pace over half the RTT, wouldn't a loss-based congestion controller (e.g. our Cubic) require some proportional fraction of the bandwidth delay product as a buffer at the bottleneck queue (i.e. here the `TailDrop`)?

https://github.com/mozilla/neqo/blob/1b16f830ba25d127b38ff6a19ea1e2e28b3a92fd/neqo-transport/src/pace.rs#L17-L27

martinthomson:
The `Delay` node buffers for as long as it needs to.  The reason to have a tail-drop queue at the sending end of a link is to ensure that the rate of egress onto that link is consistent.  The size of the buffer that the queue maintains isn't really that important.  Or at least, it only determines how much of a spike the buffer can absorb over the line rate.

Setting the buffer to the size of the BDP as the current test does means that you effectively have 3x the BDP sitting on the link somewhere.  One each near each sender and half a BDP on each `Delay` node.

mxinden:
Why does the `TailDrop` queue not matter? It enforces a bandwidth `X`. Given our pacer configuration, we send at `2*X` in the first half of the RTT, then `0*X` in the second half. Thus, without a large queue at the `TailDrop`, we will always drop `X/2` of the rate in the first half of the RTT.

martinthomson:
That's a fair point, but why would the network play nice for us like that?  It doesn't have to buffer 1GiB if we want to send that fast.

My understanding of the way that this ends up working in practice is that once you hit steady state, you are ACK-clocked.  That is, each ACK that comes in releases some more bandwidth and you send another batch of datagrams based on the increase to the CWND.  You aren't relying on that pacer at that point as much as you are on the rate of incoming ACK frames.  That's why the ACKs per RTT measure tends to matter.

The pacer operating at 4x the estimated rate shouldn't make a huge difference if we're getting 8 ACKs per RTT.  Each ACK will release 1/8 of the estimated BDP to the sender, which is about 625KiB.  In this recent tuning, that's more than my 500KiB buffer, but not by a whole lot.  That means there's probably a little overspill on the queue for sure.

I wonder if we shouldn't be doing ECN marking in `TailDrop`...

martinthomson:
OK, I'm going to suggest that we take this problem off to another thread and see if we can't work out what to do about the test separately.  I've increased the ACK rate for the test to the maximum so that it passes.  There's probably something more to do to fix that when operating at these rates, but that's for future study, not this change.

mxinden:
> My understanding of the way that this ends up working in practice is that once you hit steady state, you are ACK-clocked. 

💡 never thought about ACKs forcing the sender to smooth over the whole RTT. Thanks. This is helpful.

> OK, I'm going to suggest that we take this problem off to another thread and see if we can't work out what to do about the test separately.

Agreed. Doesn't need to block this effort. Thank you for investigating!

>  I've increased the ACK rate for the test to the maximum so that it passes. There's probably something more to do to fix that when operating at these rates, but that's for future study, not this change.

May I suggest lowering the `MINIMUM_EXPECTED_UTILIZATION` instead? That is simpler to reason about and does not deviate from the default configuration we use in Firefox.

https://github.com/mozilla/neqo/blob/1b16f830ba25d127b38ff6a19ea1e2e28b3a92fd/neqo-transport/benches/min_bandwidth.rs#L38

martinthomson:
Let's look at tweaking the min_bandwidth setup in a separate effort.  This just restores it to where it was before this change.  Isolating it, in effect.  Then we can look at changing the test in light of what we've just learned here separately.

larseggert:
Where are we with this one? Should we merge this and investigate the CC issues separately?

martinthomson:
The CC issues are latent and need attention at some point, but not here.  This is good to go.

martinthomson:
https://github.com/mozilla/neqo/actions/runs/17235761436?pr=2885 looks pretty sketchy to me @mxinden it looks like the new transfer tests are getting mangled.  Can you take a look?

mxinden:
Thanks for the ping. I will take a look. In the meantime, I suggest we revert it.

https://github.com/mozilla/neqo/pull/2908

PR #2886, opened 2025-08-21
feat(http3): support classic HTTP CONNECT

mxinden:
For now, just a hack, to unblock @KershawChang on the Firefox side.

Will need more work. Probably, as @KershawChang suggested, a new function on `Http3Client`, as HTTP CONNECT strictly speaking isn't part of _fetch_.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -758,6 +758,7 @@ impl Http3Connection {
         output
     }
 
+    // TODO: Rename?
     fn create_fetch_headers<'b, 't, T>(request: &RequestDescription<'b, 't, T>) -> Res<Vec<Header>>
     where
         T: AsRequestTarget<'t> + ?Sized + Debug,
@@ -770,10 +771,21 @@ impl Http3Connection {
         // Transform pseudo-header fields
         let mut final_headers = vec![
             Header::new(":method", request.method),
-            Header::new(":scheme", target.scheme()),
             Header::new(":authority", target.authority()),
-            Header::new(":path", target.path()),
         ];
+        let is_classic_connect = request.connect_type.is_none()
+            && request
+                .headers
+                .iter()
+                .any(|h| h.name() == ":method" && h.value() == "CONNECT");
+        if is_classic_connect {
+            // > The :scheme and :path pseudo-header fields are omitted
+            //
+            // <https://datatracker.ietf.org/doc/html/rfc9114#section-4.4>
+        } else {
+            final_headers.push(Header::new(":scheme", target.scheme()));
+            final_headers.push(Header::new(":path", target.path()));
+        }
         if let Some(conn_type) = request.connect_type {
             final_headers.push(Header::new(":protocol", conn_type.string()));
         }
@@ -835,6 +847,7 @@ impl Http3Connection {
     {
         let final_headers = Self::create_fetch_headers(request)?;
 
+        // TODO: Why is this here? WebTransport goes a different route, no?
         let stream_type = if request.connect_type.is_some() {
             Http3StreamType::ExtendedConnect
         } else {


neqo-http3/src/connection.rs
```diff
@@ -770,10 +771,21 @@ impl Http3Connection {
...
             Header::new(":method", request.method),
-            Header::new(":scheme", target.scheme()),
             Header::new(":authority", target.authority()),
-            Header::new(":path", target.path()),
         ];
+        let is_classic_connect = request.connect_type.is_none()
+            && request
+                .headers
+                .iter()
+                .any(|h| h.name() == ":method" && h.value() == "CONNECT");
```
KershawChang: I think we should just check if `request.method == "CONNECT"`?

neqo-http3/src/connection.rs
```diff
@@ -796,22 +799,60 @@ impl Http3Connection {
...
+                qwarn!(
+                    "Method {} with CONNECT type {:?}",
+                    request.method,
+                    request.connect_type
+                );
+                debug_assert!(false);
+            }
+            None if request.method == "CONNECT" => {
+                qwarn!(
+                    "Method {} with CONNECT type {:?}",
```
martinthomson, line 811: Maybe fix the message here: you just tested that the method is CONNECT and that the connect type is `None`, so the message can be fixed.

neqo-http3/src/connection_client.rs
```diff
@@ -515,6 +516,46 @@ impl Http3Client {
...
+    /// The function establishes a classic HTTP CONNECT tunnel using `target`
+    /// and `headers`. Data can be send into the tunnel via
+    /// [`Http3Client::send_data`] and received from the tunnel via
+    /// [`Http3Client::read_data`]. The tunnel can be closed via
+    /// [`Http3Client::stream_close_send`].
+    ///
+    /// # Errors
+    ///
+    /// If a new stream cannot be created an error will be return.
+    pub fn classic_connect<'x, 't: 'x, T>(
```
martinthomson: Thinking about the name here...  I think that we want:

* `fn connect()`
* `fn connect_webtransport()`
* `fn connect_udp()`
* and maybe `connect_tcp()` (though I remain skeptical there)

I know that `connection.connect()` has some awkward implications for the connection itself, but at this level, I think we are OK to use the HTTP connect.

neqo-http3/tests/classic_connect.rs
```diff
@@ -0,0 +1,117 @@
...
+    };
+    assert_eq!(stream_id, stream.stream_id());
+    let mut data = vec![0; 4];
+    let (amount, _fin) = client.read_data(now(), stream_id, &mut data).unwrap();
+    assert_eq!(&data[..amount], b"pong");
+}
+
+#[test]
+#[should_panic(expected = "assertion failed: false")]
+fn classic_connect_via_fetch_panics_in_debug() {
```
martinthomson, line 100: Why `_in_debug()` only?  It seems like this can be an unconditional failure/panic.

neqo-http3/src/connection_client.rs
```diff
@@ -515,6 +516,46 @@ impl Http3Client {
...
+        &mut self,
+        now: Instant,
+        target: &'t T,
```
martinthomson: Is this a case where we want the target to be something else?  An HTTP connect uses a name and port, or an IP and port.  Could we use that here?  We could maybe have the type we choose implement `AsRequestTarget` so that we don't need to tweak `RequestDescription` at all, but a narrower type might avoid misuse of the API.

neqo-http3/src/connection.rs
```diff
@@ -796,22 +799,60 @@ impl Http3Connection {
...
-        if let Some(conn_type) = request.connect_type {
-            final_headers.push(Header::new(":protocol", conn_type.string()));
+        match request.connect_type {
+            Some(_) if request.method != "CONNECT" => {
+                qwarn!(
+                    "Method {} with CONNECT type {:?}",
+                    request.method,
+                    request.connect_type
+                );
+                debug_assert!(false);
```
martinthomson: Is this really the way we want to handle this?  I see the test (which will fail in a non-debug test run...) but it seems like the right thing to do is abort: the request we're talking about isn't something that makes sense.

It's understandable that you want to pass a request through at this point, but you aren't validating earlier in the stack either.

neqo-http3/src/connection.rs
mxinden: :+1: done in https://github.com/mozilla/neqo/pull/2886/commits/f2ac6e41421ac21e41b7eeea8ae610f8872fe6a1.

neqo-http3/src/connection_client.rs
mxinden: Done in https://github.com/mozilla/neqo/pull/2886/commits/694c2b28dfbfa1107124a90133686131df3ae46b.

neqo-http3/tests/classic_connect.rs
mxinden: Changed in https://github.com/mozilla/neqo/pull/2886/commits/577307073ff708e4db240c2609527a94660045a7. Let me know if this addresses your concerns above.

neqo-http3/src/connection_client.rs
mxinden: Good point. I added https://github.com/mozilla/neqo/pull/2886/commits/1b6a760e27566e455e92ffad3a223bff8694b9e9 here for now. As a follow-up I created https://github.com/mozilla/neqo/pull/2966.

neqo-http3/src/connection.rs
mxinden: Addressed in https://github.com/mozilla/neqo/pull/2886/commits/577307073ff708e4db240c2609527a94660045a7. Let me know if I misunderstood your suggestion.

mxinden:
@martinthomson  mentioned in https://github.com/mozilla/neqo/pull/2796#pullrequestreview-3194685313 that he is short on time. Maybe @larseggert can you help review?

larseggert:
Could we add a test to improve coverage around the error conditions arising in `connect`?

mxinden:
I am not sure how to trigger a connection level error from `fetch` or `connect`. E.g. providing a `CONNECT` through `fetch` is now caught early in the method. E.g. one can no longer provide a scheme or path to `connect`, as it is prevented at the type level.

Long term I suggest we split the main `neqo_http3::Error` up, such that each method only returns the subset it can actually return. Though I suggest doing so in a separate pull request.

In the meantime, @larseggert are you ok proceeding?

I did add another commit to cover the other previously uncovered lines in `create_request_headers`.

PR #2887, opened 2025-08-21
fix(transport/server): call process_input with all saved_datagrams 

mxinden:
Given a server with two saved datagrams, where the first does not trigger an immediate response, the server would execute multiple `process_input` calls, one per datagram, and thus run into the following `assert`:

https://github.com/mozilla/neqo/blob/4d17dea10f163189022ec1c11e82602da6a7e198/neqo-transport/src/server.rs#L476-L479

With this commit, all saved datagrams are passed to `process_input` at once, processing it through the function's `while` loop.

As a side effect, replaces `VecDeque` with `Vec` as suggested by Martin in https://github.com/mozilla/neqo/pull/2734#discussion_r2204644834.

Bug introduced in https://github.com/mozilla/neqo/pull/2734.

Caught in https://treeherder.mozilla.org/jobs?repo=try&revision=05ccc36f73b45f2cff11b3f2c21044af706b779c&selectedTaskRun=NBGOUuvrRUiejEbiJFK4EA.0


neqo-transport/src/server.rs
```diff
@@ -533,23 +532,27 @@ impl Server {
...
-                        d: d.to_owned(),
-                        t: now,
-                    }));
-                // Return immediately. Do any maintenance on next call.
-                return OutputBatch::DatagramBatch(b);
-            }
+        // Process input datagrams from previous call.
+        let saved_datagrams = std::mem::take(&mut self.saved_datagrams)
+            .into_iter()
+            .map(|SavedDatagram { d, .. }| d);
```
martinthomson: You can't drop the time at which the datagram was saved.  Otherwise, you foul up the RTT estimates.

neqo-transport/src/server.rs
```diff
@@ -533,23 +532,27 @@ impl Server {
...
+        if let OutputBatch::DatagramBatch(b) = self.process_input(saved_datagrams, now) {
```
martinthomson: This needs to be passed `SavedDatagram::t`, not `now`.

I know that you want to feed a batch in, but you can't erase the time from saved datagrams.  You could look for a run of datagrams with the same time and feed those in together, I guess.  But then you are probably back into the `VecDeque` domain with `take_until` or similar (or one of the `drain()` variants if those ever ship).

martinthomson:
```suggestion
        if let b @ OutputBatch::DatagramBatch(_) = self.process_input(saved_datagrams, FIXME) {
```
Then you can return `b` later.

neqo-transport/src/server.rs
mxinden: :facepalm: of course. Thanks for the catch.

neqo-transport/src/server.rs
mxinden: > This needs to be passed SavedDatagram::t, not now.

Fixed. Thanks for the catch!

> You could look for a run of datagrams with the same time and feed those in together, I guess

I believe the `saved_datagrams` mechanism serves an edge case only, thus not worth the optimization / complexity.

neqo-transport/tests/server.rs
```diff
@@ -900,40 +900,56 @@ fn has_active_connections() {
...
+        })
+        .collect();
+
+    // Server sends a version negotation immediately. Saves second and third
+    // input datagram for later.
     server
         .process_multiple(
-            vec![damaged_dgram.clone(), damaged_dgram.clone()],
+            vec![
+                invalid_dgrams.pop().unwrap(),
```
martinthomson: Rather than collect datagrams into a vec and then pop them, why not make an `invalid_dgram()` function, passing it the damage you want it to inflict.

You could even avoid making a new client, instead damaging `valid_dgram`, though you would have to clone it.

martinthomson (commented):
Just a suggestion for the test.  And, yes, we don't have to be super-performant when it comes to saved datagrams.  The extra complexity of batching is probably not worth it.

neqo-transport/tests/server.rs
mxinden: > Rather than collect datagrams into a vec and then pop them, why not make an invalid_dgram() function, passing it the damage you want it to inflict.

:+1: will give it a try. Thanks.

> You could even avoid making a new client, instead damaging valid_dgram, though you would have to clone it.

I believe I need different connection IDs to force the server to allocate new connections.

martinthomson (approved):

PR #2891, opened 2025-08-23
bench(transport/transfer): measure both wallclock and simulated time

mxinden:
The `transfer.rs` benchmark uses the `test-fixtures` `Simulator`. The wallclock time of a simulator run tells us the compute efficiency (e.g. cost of memory allocations) of our code, the simulated time tells us the network protocol efficiency (e.g. cost of a congestion control strategy).

("network protocol efficiency" in lack of a better description.)

This commit extends the benchmark to additionally measure simulated time.

---

I discovered `criterion`'s [`Bencher::iter_custom`](https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_custom) today. In combination with the `Simulator`, we can now benchmark simulated time across various seeds.

### DIFF
--- a/neqo-transport/benches/transfer.rs
+++ b/neqo-transport/benches/transfer.rs
@@ -82,6 +82,7 @@ fn benchmark_transfer(c: &mut Criterion, label: &str, seed: Option<impl AsRef<st
                 let mut d_sum = Duration::ZERO;
                 for _i in 0..iters {
                     // run() returns the simulated time, excluding setup time.
+                    // No need for black_box(), because this doesn't measure compute.
                     d_sum += setup().run();
                 }
                 d_sum


.github/workflows/bench.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             cp "$BENCH" ../binaries/neqo/
             # shellcheck disable=SC2086
             nice -n -20 setarch --addr-no-randomize cset proc --set=cpu23 --exec \
-              $BENCH -- --bench --baseline main | { grep -v '^cset' || test $? = 1; } | tee -a ../results.txt
+              $BENCH -- --bench --baseline-lenient main | { grep -v '^cset' || test $? = 1; } | tee -a ../results.txt
```
mxinden, line 137: This pull request introduces new benchmarks. They are not yet present on `main`.

Without `-lenient` `cargo bench` fails, as it can not find a `main` execution of those very benchmarks to compare against.

```
Baseline 'main' must exist before comparison is allowed; try --save-baseline main
```

```
  -b, --baseline <baseline>
          Compare to a named baseline. If any benchmarks do not have the specified baseline this command fails.
      --baseline-lenient <baseline-lenient>
          Compare to a named baseline. If any benchmarks do not have the specified baseline then just those benchmarks are not compared against the baseline while every other benchmark is compared against the baseline.
```

mxinden:
```
transfer/pacing-true/varying-seeds/simulated-time
       time:   [36.725 s 36.787 s 36.847 s]
       thrpt:  [111.16 KiB/s 111.34 KiB/s 111.53 KiB/s]
  1 (1.00%) low mild
```

Benchmark uses a simulated 1 Mbyte/s link, thus 100 Kbyte/s effective throughput is low.

https://github.com/mozilla/neqo/blob/1b16f830ba25d127b38ff6a19ea1e2e28b3a92fd/test-fixture/src/sim/taildrop.rs#L82-L87

neqo-transport/benches/transfer.rs
```diff
@@ -27,45 +27,69 @@ const TRANSFER_AMOUNT: usize = 1 << 22; // 4Mbyte
...
+        // efficiency.
+        //
+        // Note: Given that this is using simulated time, we can measure actual
+        // throughput.
+        group.throughput(criterion::Throughput::Bytes(TRANSFER_AMOUNT as u64));
+        group.bench_function("simulated-time", |b| {
+            b.iter_custom(|iters| {
+                let mut d_sum = Duration::ZERO;
+                for _i in 0..iters {
+                    d_sum += setup().run();
```
martinthomson, line 86: I guess that you don't need to worry here about the setup overhead and `blackbox()` because this only cares about the simulated time; it might be worth noting that.

.github/workflows/bench.yml
martinthomson: Want to open an issue to remove this once we're stable?  I can see why having the ability to add benchmarks is useful.

It might pay to have a comment here if you keep it: `# --baseline-lenient allows us to add new benchmarks in pull requests`

Or plan to remove it: `# TODO: remove --baseline-lenient once #2891 is on main`

neqo-transport/benches/transfer.rs
```diff
@@ -27,45 +27,69 @@ const TRANSFER_AMOUNT: usize = 1 << 22; // 4Mbyte
...
         });
+
+        // Benchmark with simulated time, i.e. measure the network protocol
+        // efficiency.
+        //
+        // Note: Given that this is using simulated time, we can measure actual
+        // throughput.
+        group.throughput(criterion::Throughput::Bytes(TRANSFER_AMOUNT as u64));
+        group.bench_function("simulated-time", |b| {
+            b.iter_custom(|iters| {
```
martinthomson, line 81: How does this work out in terms of performance?  Usually, the test infrastructure runs the function a few times to see how long it takes.  Does it use the wall clock for that process, or the time that this function reports?  The wall clock would be fine, but if it uses the time this function reports, that will give some very bad results.

larseggert (approved):

neqo-transport/benches/transfer.rs
mxinden: From a quick glance at the code, `iter_custom` tracks both the elapsed time (i.e. wallclock time) as well as the time returned here (i.e. simulated time). It will use the wallclock time to e.g. find a reasonable warm up duration and it will use the simulated time for the actual measurement.

https://github.com/bheisler/criterion.rs/blob/af5cc00ef1ad5e32b2d36a5be4d9cad8ed0c6ec9/src/bencher.rs#L44-L45

https://github.com/bheisler/criterion.rs/blob/af5cc00ef1ad5e32b2d36a5be4d9cad8ed0c6ec9/src/bencher.rs#L132-L140

If I understand your comment above correctly, that addresses your concerns.

.github/workflows/bench.yml
mxinden: :+1: would like to keep it. I expect to add more benchmarks in future pull requests. Added a comment in https://github.com/mozilla/neqo/pull/2891/commits/5c53f72ddbde8c2d5d81c7f3ab74b7927648ca47.

neqo-transport/benches/transfer.rs
mxinden: Added a comment for the setup overhead and moved `blackbox` to the wallclock test only https://github.com/mozilla/neqo/pull/2891/commits/c2b15526c9b79d6952f2bc0b91eb73fd2d915a7a. Hope this makes it easier to read.

neqo-transport/benches/transfer.rs
martinthomson: Yep, they did the sensible thing.  That's not always something you can rely on.

neqo-transport/benches/transfer.rs
```diff
@@ -27,45 +27,67 @@ const TRANSFER_AMOUNT: usize = 1 << 22; // 4Mbyte
...
+                let mut d_sum = Duration::ZERO;
+                for _i in 0..iters {
+                    // run() returns the simulated time, excluding setup time.
```
martinthomson, line 84:
```suggestion
                    // run() returns the simulated time, excluding setup time.
                    // No need for black_box(), because this doesn't measure compute.
```

martinthomson (approved):

PR #2893, opened 2025-08-25
Simplify the fixed delay node

martinthomson:
This should be faster as the structure is less complex, but we'll need to test against the changes that are coming to test wall-clock time for the simulation runs.

### DIFF
--- a/test-fixture/src/sim/delay.rs
+++ b/test-fixture/src/sim/delay.rs
@@ -103,7 +103,7 @@ impl Debug for RandomDelay {
 
 pub struct Delay {
     delay: Duration,
-    queue: BTreeMap<Instant, VecDeque<Datagram>>,
+    queue: VecDeque<(Instant, Datagram)>,
 }
 
 impl Delay {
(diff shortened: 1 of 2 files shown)


test-fixture/src/sim/delay.rs
```diff
@@ -128,19 +128,15 @@ impl Node for Delay {
...
-                if same_time_queue.is_empty() {
-                    self.queue.remove(&k);
-                }
-
-                return Output::Datagram(d);
+        if let Some((t, _)) = self.queue.front() {
+            if *t <= now {
+                let Some((_, d)) = self.queue.pop_front() else {
+                    unreachable!();
+                };
```
mxinden:
```suggestion
                let Some((_, d)) = self.queue.pop_front().expect("fron() check above");
```

A bit simpler?

mxinden (approved):
We already test wallclock time in `transfer.rs` today. https://github.com/mozilla/neqo/pull/2891 only adds simulated time. In other words, if you see a significant improvement in the benchmarks, that is likely caused by this change.

mxinden:
No impact visible via the `transfer.rs` benchmark. That said, I am in favor of merging here anyways.

PR #2895, opened 2025-08-25
chore: Update some `dev-dependencies`

larseggert:


larseggert:
Yeah, please update your license in a point release.
```error[rejected]: failed to satisfy license requirements
  ┌─ registry+https://github.com/rust-lang/crates.io-index#libfuzzer-sys@0.4.10:4:36
  │
4 │ license = "(MIT OR Apache-2.0) AND NCSA"
  │            ────────────────────────━━━━
  │            │                       │
  │            │                       rejected: license is not explicitly allowed
  │            license expression retrieved via Cargo.toml `license`
  │
  ├ NCSA - University of Illinois/NCSA Open Source License:
  ├   - OSI approved
  ├   - FSF Free/Libre
  ├ libfuzzer-sys v0.4.10
    └── fuzz v0.14.1
```

mxinden (approved):

larseggert:
Sorry, @mxinden, more changes.

Cargo.lock
```diff
@@ -1298,17 +1325,19 @@ dependencies = [
...
 [[package]]
 name = "tokio"
-version = "1.43.1"
+version = "1.47.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "492a604e2fd7f814268a378409e6c92b5525d747d10db9a229723f55a417958c"
+checksum = "89e49afdadebb872d3145a5638b59eb0691ea23e46ca484037cfab3b76b95038"
 dependencies = [
  "backtrace",
+ "io-uring",
```
mxinden: Oh, that is surprising. Where is this coming from?

Cargo.lock
```diff
@@ -1481,6 +1510,15 @@ dependencies = [
  "windows-targets",
 ]
 
+[[package]]
+name = "windows-sys"
+version = "0.59.0"
```
mxinden: Only a dev-dependency, still deviates from what Firefox is using. Worth it?

https://github.com/mozilla-firefox/firefox/blob/772c2d9c9704ef53aca712aece449d0a03e0be40/Cargo.lock#L7893-L7900

Cargo.lock
larseggert: `tokio`. I don't see a way to disable it, do you?

Cargo.lock
larseggert: Addressed by going down to `tokio@1.46.1`.

Cargo.lock
mxinden: Discussed in https://github.com/tokio-rs/tokio/issues/7546. Potential fix: https://github.com/tokio-rs/tokio/pull/7547.

Fine to proceed or wait until a fix is available. Your call.

larseggert:
Blocked until `tokio` defines a way to get rid of the `io-uring` dep.

mxinden:
For the record, https://github.com/tokio-rs/tokio/pull/7621 merged, which should unblock the `tokio` issue.

larseggert:
> For the record, [tokio-rs/tokio#7621](https://github.com/tokio-rs/tokio/pull/7621) merged, which should unblock the `tokio` issue.

ACK. I think we still need to wait until they cut a release.

Cargo.lock
larseggert: `tokio` has now been updated, but pulling it in duplicates some Windows crates. @mxinden do you consider this a problem?

Cargo.lock
```diff
@@ -1223,7 +1163,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ce305eb0b4296696835b71df73eb912e0f1ffd2556a501fcede6e0c50349191c"
 dependencies = [
  "libc",
- "windows-sys",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "socket2"
+version = "0.6.1"
```
mxinden, line 1152: Firefox uses `v0.5.7`. Is the deviation worth it? While it won't spill into Firefox itself, it will spill into mozilla-central, as the `http3server` uses it. We would either need to update mozilla-central to `socket2` `v0.5.7`, or deviate from the no-2-versions-of-a-crate rule.

https://github.com/mozilla-firefox/firefox/blob/04f343bf04ce512276a27788ef21f06e995b9496/Cargo.lock#L6473-L6481

Cargo.lock
```diff
@@ -1506,13 +1455,19 @@ dependencies = [
  "syn",
 ]
 
+[[package]]
+name = "windows-link"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f0805222e57f7521d6a62e36fa9163bc891acd422f971defe97d64e70d0a4fe5"
```
mxinden, line 1465: for the record, same as Firefox:

https://github.com/mozilla-firefox/firefox/blob/main/Cargo.lock#L8089

Cargo.lock
```diff
@@ -1531,7 +1486,25 @@ version = "0.52.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "282be5f36a8ce781fad8c8ae18fa3f9beff57ec1b52cb3de0789201425d9a33d"
 dependencies = [
- "windows-targets",
+ "windows-targets 0.52.6",
+]
+
+[[package]]
+name = "windows-sys"
+version = "0.60.2"
```
mxinden, line 1506: Firefox uses `0.52.0`. This is a `dependency`. Not a `dev-dependency. I doubt we should break the no-2-versions-of-a-crate rule here.

Cargo.lock
```diff
@@ -1531,7 +1486,25 @@ version = "0.52.0"
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2f500e4d28234f72040990ec9d39e3a6b950f9f22d3dba18416c35882612bcb"
+dependencies = [
+ "windows-targets 0.53.5",
+]
+
+[[package]]
+name = "windows-sys"
+version = "0.61.2"
```
mxinden, line 1515: Why is the third `windows-sys` dependency needed?

Cargo.lock
larseggert: `tokio` needs `socket2 = "^0.6.0"`. Updating `http3server` should be fine.

Cargo.lock
larseggert: Yeah, this blocks us.

Cargo.lock
larseggert: I think `tokio` and `socket2` have different requirements.

larseggert:
Let's wait until Fx deps have been uplifted.

PR #2890, opened 2025-08-23
feat(test-fixtures): add http3 support to Simulator

mxinden:
Enables benchmarking of MASQUE connect-udp implementation (https://github.com/mozilla/neqo/pull/2796).

Fixes https://github.com/mozilla/neqo/issues/2728.
Fixes https://github.com/mozilla/neqo/issues/2804.

PR #2897, opened 2025-08-25
feat: Import the `mtu` crate

larseggert:
In preparation of archiving the standalone repo. There's little point in having it be separate from the main consumer.


mxinden:
In favor of this move!

In case you want to maintain the mozilla/mtu git history, there are various tricks to combine two git repositories including their history, e.g. here mozilla/neqo and mozilla/mtu.

https://stackoverflow.com/questions/5995826/import-a-git-repository-with-all-its-history-into-an-existing-git-repository

I don't have an opinion. In other words, I am fine loosing the mozilla/mtu git history.

larseggert:
The failing semver check is because `mtu` is not in `main` and hence can't be checked.

mtu/.clippy.toml
mxinden, line 1: Does our root `.clippy.toml` not apply here?

https://github.com/mozilla/neqo/blob/main/.clippy.toml

.github/workflows/check-mtu.yml
mxinden, line 1: Do we need the additional workflow? Can this be combined with our existing `check.yml`. E.g. the Android steps seem to be mostly duplicated.

neqo-http3/Cargo.toml
```diff
@@ -32,7 +32,7 @@ strum = { workspace = true}
 url = { workspace = true }
 
 [dev-dependencies]
-criterion = { version = "0.6", default-features = false, features = ["cargo_bench_support"] }
+# criterion = { version = "0.6", default-features = false, features = ["cargo_bench_support"] }
```
mxinden, line 35: Remove all together? I can re-introduce it in https://github.com/mozilla/neqo/pull/2890.

.github/workflows/check-mtu.yml
larseggert: I was going to open up an issue on this. We should definitely figure out a way to remove the duplication. But maybe better in a separate PR.

neqo-http3/Cargo.toml
larseggert: There is also the other commented-out stuff at the end of the file; so I'd prefer to just leave this as a comment, too.

mxinden:
Thanks for consolidating the CI actions!

larseggert:
I think this is finally good to go.

mxinden (approved):

PR #2889, opened 2025-08-22
feat: start using thiserror

mxinden:
Fixes https://github.com/mozilla/neqo/issues/2834.

PR #2912, opened 2025-08-27
refactor: replace all manual impl std::error::Error with thiserror

mxinden:
Continuation of https://github.com/mozilla/neqo/pull/2889.


neqo-crypto/src/err.rs
```diff
@@ -54,33 +51,51 @@ pub mod mozpkix {
 
 pub type Res<T> = Result<T, Error>;
 
-#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq, Error)]
 pub enum Error {
-    Aead,
```
mxinden, line 59: I believe this is unused.

neqo-crypto/src/err.rs
```diff
@@ -54,33 +51,51 @@ pub mod mozpkix {
...
     MixedHandshakeMethod,
+    #[error("No data available")]
     NoDataAvailable,
+    #[error("NSS error: {name} ({code}): {desc}")]
     Nss {
         name: String,
         code: PRErrorCode,
         desc: String,
     },
-    Overrun,
```
mxinden, line 79: I believe this is unused.

neqo-bin/src/client/mod.rs
```diff
@@ -43,14 +44,21 @@ mod http3;
...
+    #[error("I/O error: {0}")]
     Io(io::Error),
+    #[error("qlog error: {0}")]
     Qlog(qlog::Error),
+    #[error("transport error: {0}")]
     Transport(neqo_transport::Error),
+    #[error("application error: {0}")]
     Application(AppError),
+    #[error("crypto error: {0}")]
     Crypto(neqo_crypto::Error),
```
larseggert: I guess there isn't a way to automatically derive error messages for the entire `enum` type following this pattern?

mxinden:
> Isn't the workspace-level `Cargo.toml` missing from this PR?

The `thiserror` workspace declaration?

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/Cargo.toml#L44

That was added in https://github.com/mozilla/neqo/pull/2889.

neqo-bin/src/client/mod.rs
mxinden: Unfortunately not (that I am aware of). I was able to replace a couple of these with `#[error(transparent)]` just now though. Let me know what you think.

neqo-crypto/src/err.rs
larseggert: Maybe remove it then. @martinthomson?

neqo-http3/src/lib.rs
```diff
@@ -183,54 +179,97 @@ pub use server_events::{
...
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Error)]
 pub enum Error {
+    #[error("HTTP no error")]
     HttpNone,
+    #[error("HTTP general protocol error")]
     HttpGeneralProtocol,
-    HttpGeneralProtocolStream, /* this is the same as the above but it should only close a
-                                * stream not a connection. */
+    // this is the same as the above but it should only close a stream not a
```
larseggert:
```suggestion
    // This is the same as the above but it should only close a stream not a
```

neqo-qpack/src/lib.rs
```diff
@@ -36,29 +35,49 @@ pub struct Settings {
...
+    #[error("Critical stream closed")]
     ClosedCriticalStream,
+    #[error("Internal error")]
     Internal,
 
     // These are internal errors, they will be transformed into one of the above.
-    NeedMoreData, /* Return when an input stream does not have more data that a decoder
-                   * needs.(It does not mean that a stream is closed.) */
+
+    // Return when an input stream does not have more data that a decoder needs.(It
```
larseggert:
```suggestion
    // Return when an input stream does not have more data that a decoder needs. (It
```

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     Internal,
+    #[error("connection refused")]
     ConnectionRefused,
+    #[error("flow control error")]
     FlowControl,
+    #[error("stream limit exceeded")]
     StreamLimit,
+    #[error("stream state error")]
     StreamState,
+    #[error("final size error")]
```
martinthomson:
```suggestion
    #[error("stream final size error")]
```

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     ConnectionIdLimitExceeded,
+    #[error("connection IDs exhausted")]
     ConnectionIdsExhausted,
+    #[error("invalid connection state")]
     ConnectionState,
+    #[error("decryption error")]
     Decrypt,
+    #[error("disabled version")]
     DisabledVersion,
+    #[error("idle timeout")]
```
martinthomson:
```suggestion
    #[error("no packets received for longer than the idle timeout")]
```

neqo-qpack/src/lib.rs
```diff
@@ -36,29 +35,49 @@ pub struct Settings {
...
+    // does not mean that a stream is closed.)
```
martinthomson:
```suggestion
    /// Return when an input stream does not have more data that a decoder needs.
    /// It does not mean that a stream is closed.
```
Doesn't this need /// ?

neqo-http3/src/lib.rs
```diff
@@ -183,54 +179,97 @@ pub use server_events::{
...
-    Transport(TransportError),
+    #[error("Transport error: {0}")]
+    Transport(
+        #[from]
+        #[source]
+        TransportError,
+    ),
+    #[error("Transport stream does not exist")]
     TransportStreamDoesNotExist,
+    #[error("Unavailable")]
```
martinthomson: It would be helpful to have something a little clearer for these.

neqo-crypto/src/err.rs
martinthomson: Yeah, we don't bother transforming the errors from NSS here.  That was the original plan, but it never happened.

neqo-crypto/src/err.rs
```diff
@@ -19,6 +15,7 @@ mod codes {
     include!(concat!(env!("OUT_DIR"), "/nss_sslerr.rs"));
 }
 pub use codes::{SECErrorCodes as sec, SSLErrorCodes as ssl};
+use thiserror::Error;
 #[expect(dead_code, reason = "Code is bindgen-generated.")]
```
martinthomson, line 20:
```suggestion

#[expect(dead_code, reason = "Code is bindgen-generated.")]
```

Some elbow room between "use" and "mod" stuff.

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     IdleTimeout,
+    #[error("integer overflow")]
     IntegerOverflow,
+    #[error("invalid input")]
     InvalidInput,
+    #[error("invalid migration")]
     InvalidMigration,
+    #[error("invalid packet")]
```
martinthomson:
```suggestion
    #[error("an invalid packet was dropped (internal use only)")]
```

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     InvalidPacket,
+    #[error("invalid resumption token")]
     InvalidResumptionToken,
+    #[error("invalid retry")]
```
martinthomson:
```suggestion
    #[error("invalid retry packet dropped (internal use only)")]
```

There's a bunch of these...

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     UnexpectedMessage,
+    #[error("unknown connection ID")]
     UnknownConnectionId,
+    #[error("unknown frame type")]
     UnknownFrameType,
+    #[error("version negotiation")]
     VersionNegotiation,
+    #[error("wrong role")]
     WrongRole,
+    #[error("unknown transport parameter")]
```
martinthomson: This is probably internal-only as well.

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
     NoAvailablePath,
+    #[error("no more data")]
     NoMoreData,
+    #[error("not available")]
     NotAvailable,
+    #[error("not connected")]
     NotConnected,
+    #[error("packet number overlap")]
     PacketNumberOverlap,
+    #[error("peer application error: {0}")]
```
martinthomson:
```suggestion
    #[error("peer application error: 0x{0:x}")]
```

Might pay to use hex for these.

neqo-transport/src/lib.rs
```diff
@@ -82,66 +81,114 @@ const ERROR_APPLICATION_CLOSE: TransportError = 12;
...
+    #[error("keys discarded for epoch {0:?}")]
     KeysDiscarded(crypto::Epoch),
-    /// Packet protection keys are exhausted.
-    /// Also used when too many key updates have happened.
+    /// Packet protection keys are exhausted.  Also used when too many key
+    /// updates have happened.
+    #[error("keys exhausted")]
     KeysExhausted,
     /// Packet protection keys aren't available yet for the identified space.
+    #[error("keys pending for epoch {0:?}")]
```
martinthomson: This is, I think, internal.

martinthomson (commented):
Lots of internal errors that should never see the light of day here.  That's fine, but it might be worth indicating which is which.

neqo-crypto/src/err.rs
larseggert: Nix it.

PR #2914, opened 2025-08-28
Add ECN to the simulator

martinthomson:
This is still pretty crude, but it seems to approximate something
sensible now.

There was a big bug in the simulator where the TailDrop node wasn't
asking to be serviced when removing things from its queue.  So it was
only removing things from the queue when packets left the link.  That
meant that the only sensible way to use it was to have zero delay (as it
was being used).

This fixes that and adds ECN marking at a chosen threshold.  My
observation is that the marking threshold has a small impact on
performance (the output below shows that moving from marking at 0.7 to
marking at 0.8 improves throughput a little.  The first two were run at
0.7. Obviously, this comes at the cost of extra latency.  I have no real
idea what buffers make sense, this is just a synthetic test that sets
out some guard rails for things like CC and flow control under very,
very clean conditions.

This is still not ideal.  The min_bandwidth test should probably get a
higher throughput than it does.  Occasionally, this gets as high as
870Mbps.  But mostly it is 760Mbps or thereabouts.  I have no idea what
makes those occasional runs so much faster.

```
# while RUST_LOG=info cargo test --release -p neqo-transport --bench min_bandwidth -F bench |& grep Achieved; do :; done
5.384 INFO Achieved 756.5999835520076 Mb/s bandwidth (link rate 1000)
5.411 INFO Achieved 759.7709198759945 Mb/s bandwidth (link rate 1000)
5.536 INFO Achieved 769.6899939983132 Mb/s bandwidth (link rate 1000)
5.358 INFO Achieved 771.442097560842 Mb/s bandwidth (link rate 1000)
5.423 INFO Achieved 768.9489036615845 Mb/s bandwidth (link rate 1000)
5.551 INFO Achieved 870.1786336328342 Mb/s bandwidth (link rate 1000)
5.504 INFO Achieved 768.9484584413776 Mb/s bandwidth (link rate 1000)
5.380 INFO Achieved 768.9018540063886 Mb/s bandwidth (link rate 1000)
5.412 INFO Achieved 768.9489036615845 Mb/s bandwidth (link rate 1000)
5.492 INFO Achieved 768.9481621100322 Mb/s bandwidth (link rate 1000)
5.418 INFO Achieved 872.2894963620984 Mb/s bandwidth (link rate 1000)
5.423 INFO Achieved 768.9489036615845 Mb/s bandwidth (link rate 1000)
5.377 INFO Achieved 872.2894966278358 Mb/s bandwidth (link rate 1000)
5.488 INFO Achieved 768.9024512780954 Mb/s bandwidth (link rate 1000)
5.439 INFO Achieved 767.510863851431 Mb/s bandwidth (link rate 1000)
5.487 INFO Achieved 768.9489038680874 Mb/s bandwidth (link rate 1000)
```

### DIFF
--- a/neqo-transport/benches/min_bandwidth.rs
+++ b/neqo-transport/benches/min_bandwidth.rs
@@ -14,12 +14,13 @@
 
 use std::time::Duration;
 
+use neqo_common::{log::init as log_init, qinfo};
 use neqo_transport::{ConnectionParameters, State};
 use test_fixture::{
     boxed,
     sim::{
         connection::{Node, ReachState, ReceiveData, SendData},
-        network::{Delay, Mtu, TailDrop},
+        network::{Mtu, TailDrop},
         Simulator,
     },
 };
@@ -35,14 +36,22 @@ pub fn main() {
     const TRANSFER_AMOUNT: usize = GIB;
     const LINK_BANDWIDTH: usize = GBIT;
     const LINK_RTT_MS: usize = 40;
-    const MINIMUM_EXPECTED_UTILIZATION: f64 = 0.85;
+    /// The amount of delay that the link buffer will add, when full.
+    const BUFFER_LATENCY_MS: usize = 4;
+    /// The proportion of the link buffer at which marking starts.
+    const ECN_THRESHOLD: f64 = 0.8;
+    /// How much of the theoretical bandwidth we will expect to deliver.
+    const MINIMUM_EXPECTED_UTILIZATION: f64 = 0.7;
 
     let gbit_link = || {
         let rate_byte = LINK_BANDWIDTH / 8;
-        // BDP here is 5MB = 1e9 * 0.04 / 8: buffer an entire BDP here.
-        let capacity_byte = LINK_BANDWIDTH * LINK_RTT_MS / 1000;
-        TailDrop::new(rate_byte, capacity_byte, capacity_byte, Duration::ZERO)
+        let capacity_byte = LINK_BANDWIDTH * BUFFER_LATENCY_MS / 1000;
+        let mark_capacity = ((capacity_byte as f64) * ECN_THRESHOLD) as usize;
+        let delay = Duration::from_millis(LINK_RTT_MS as u64) / 2;
+        TailDrop::new(rate_byte, capacity_byte, mark_capacity, delay)
     };
+
+    log_init(None);
 
     let simulated_time = Simulator::new(
         "gbit-bandwidth",
@@ -54,7 +63,6 @@ pub fn main() {
             ),
             Mtu::new(1500),
             gbit_link(),
-            Delay::new(Duration::from_millis(LINK_RTT_MS as u64 / 2)),
             Node::new_server(
                 ConnectionParameters::default(),
                 boxed![ReachState::new(State::Confirmed)],
@@ -62,13 +70,17 @@ pub fn main() {
             ),
             Mtu::new(1500),
             gbit_link(),
-            Delay::new(Duration::from_millis(LINK_RTT_MS as u64 / 2)),
         ],
     )
     .setup()
     .run();
 
     let achieved_bandwidth = TRANSFER_AMOUNT as f64 * 8.0 / simulated_time.as_secs_f64();
+    qinfo!(
+        "Achieved {} Mb/s bandwidth (link rate {})",
+        achieved_bandwidth / MBIT as f64,
+        LINK_BANDWIDTH / MBIT
+    );
 
     assert!(
         LINK_BANDWIDTH as f64 * MINIMUM_EXPECTED_UTILIZATION < achieved_bandwidth,
--- a/neqo-transport/benches/min_bandwidth.rs
+++ b/neqo-transport/benches/min_bandwidth.rs
@@ -39,9 +39,9 @@ pub fn main() {
 
     let gbit_link = || {
         let rate_byte = LINK_BANDWIDTH / 8;
-        // Router buffer set to bandwidth-delay product.
-        let capacity_byte = rate_byte * LINK_RTT_MS / 1_000;
-        TailDrop::new(rate_byte, capacity_byte, Duration::ZERO)
+        // BDP here is 5MB = 1e9 * 0.04 / 8: buffer an entire BDP here.
+        let capacity_byte = LINK_BANDWIDTH * LINK_RTT_MS / 1000;
+        TailDrop::new(rate_byte, capacity_byte, capacity_byte, Duration::ZERO)
     };
 
     let simulated_time = Simulator::new(
--- a/test-fixture/src/sim/taildrop.rs
+++ b/test-fixture/src/sim/taildrop.rs
@@ -7,7 +7,7 @@
 #![expect(clippy::unwrap_used, reason = "This is test code.")]
 
 use std::{
-    cmp::max,
+    cmp::{max, min},
     collections::VecDeque,
     fmt::{self, Debug, Display},
     time::{Duration, Instant},
@@ -82,10 +82,8 @@ pub struct TailDrop {
     /// A queue of unsent bytes.
     queue: VecDeque<Datagram>,
     /// The time that the next datagram can enter the link.
-    next_deque: Option<Instant>,
+    next_deque: Option<(Instant, u32)>,
 
-    /// Any sub-ns delay from the last enqueue.
-    sub_ns_delay: u32,
     /// The time it takes a byte to exit the other end of the link.
     delay: Duration,
     /// The packets that are on the link and when they can be delivered.
@@ -98,15 +96,16 @@ impl TailDrop {
     /// Make a new taildrop node with the given rate, queue capacity, and link delay.
     #[must_use]
     pub const fn new(rate: usize, capacity: usize, mark_capacity: usize, delay: Duration) -> Self {
+        assert!(rate != 0, "zero rate gets you nowhere");
+        assert!(rate <= 1_000_000_000, "rates over 1Gbps are not supported");
         Self {
-            overhead: 64,
+            overhead: 80,
             rate,
             capacity,
             mark_capacity,
             used: 0,
             queue: VecDeque::new(),
             next_deque: None,
-            sub_ns_delay: 0,
             delay,
             on_link: VecDeque::new(),
             stats: Stats::new(),
@@ -134,7 +133,8 @@ impl TailDrop {
     }
 
     /// Start sending a datagram.
-    fn send(&mut self, d: Datagram, now: Instant) {
+    /// Send at the given time, with the given sub-nanosecond extra delay.
+    fn send(&mut self, d: Datagram, now: Instant, sub_ns: u32) {
         // How many bytes are we "transmitting"?
         let sz = u128::try_from(self.size(&d)).unwrap();
 
@@ -143,13 +143,13 @@ impl TailDrop {
         // This ensures that high rates and small packets don't result in rounding
         // down times too badly.
         // Duration consists of a u64 and a u32, so we have 32 high bits to spare.
-        let t = sz * (ONE_SECOND_NS << 32) / u128::try_from(self.rate).unwrap()
-            + u128::from(self.sub_ns_delay);
+        let t =
+            sz * (ONE_SECOND_NS << 32) / u128::try_from(self.rate).unwrap() + u128::from(sub_ns);
         let send_ns = u64::try_from(t >> 32).unwrap();
         assert_ne!(send_ns, 0, "sending a packet takes <1ns");
-        self.sub_ns_delay = u32::try_from(t & u128::from(u32::MAX)).unwrap();
         let deque_time = now + Duration::from_nanos(send_ns);
-        self.next_deque = Some(deque_time);
+        let sub_ns = u32::try_from(t & u128::from(u32::MAX)).unwrap();
+        self.next_deque = Some((deque_time, sub_ns));
 
         // Now work out when the packet is fully received at the other end of
         // the link. Setup to deliver the packet then.
@@ -158,12 +158,13 @@ impl TailDrop {
     }
 
     /// Enqueue for sending.  Maybe.  If this overflows the queue, drop it instead.
+    /// If the queue is too deep, CE mark it before saving.
     fn maybe_enqueue(&mut self, d: Datagram, now: Instant) {
         self.stats.received += 1;
         if self.next_deque.is_none() {
             // Nothing in the queue and nothing still sending.
             debug_assert!(self.queue.is_empty());
-            self.send(d, now);
+            self.send(d, now, 0);
         } else if self.used + self.size(&d) <= self.capacity {
             self.used += self.size(&d);
             self.stats.update_maxq(self.used);
@@ -178,21 +179,28 @@ impl TailDrop {
     /// If the last packet that was sending has been sent, start sending
     /// the next one.
     fn maybe_send(&mut self, now: Instant) {
-        if self.next_deque.as_ref().is_some_and(|t| *t <= now) {
+        if let Some((t, subns)) = &self.next_deque {
+            if now < *t {
+                return;
+            }
+
             if let Some(d) = self.queue.pop_front() {
                 self.used -= self.size(&d);
-                self.send(d, now);
+                self.send(d, *t, *subns);
             } else {
                 self.next_deque = None;
-                self.sub_ns_delay = 0;
             }
         }
     }
 
+    /// Apply ECN-CE markings to packets if they are ECT(0) marked.
+    /// This is classic, simple ECN; we don't know about L4S yet.
     fn maybe_mark(&mut self, d: Datagram) -> Datagram {
-        if self.used <= self.mark_capacity {
+        if self.used <= self.mark_capacity || Ecn::from(d.tos()) != Ecn::Ect0 {
             return d;
         }
+
+        qtrace!("taildrop marking {} bytes", d.len());
         self.stats.marked += 1;
         let tos = Tos::from((Dscp::from(d.tos()), Ecn::Ce));
         Datagram::new(d.source(), d.destination(), tos, d.to_vec())
@@ -213,7 +221,11 @@ impl Node for TailDrop {
                 self.stats.delivered += 1;
                 Output::Datagram(d)
             } else {
-                Output::Callback(*t - now)
+                let next = self
+                    .next_deque
+                    .as_ref()
+                    .map_or(*t, |(next, _)| min(*next, *t));
+                Output::Callback(next - now)
             }
         } else {
             Output::None
--- a/test-fixture/src/sim/taildrop.rs
+++ b/test-fixture/src/sim/taildrop.rs
@@ -9,17 +9,59 @@
 use std::{
     cmp::max,
     collections::VecDeque,
-    fmt::{self, Debug},
+    fmt::{self, Debug, Display},
     time::{Duration, Instant},
 };
 
-use neqo_common::{qinfo, qtrace, Datagram};
+use neqo_common::{qinfo, qtrace, Datagram, Dscp, Ecn, Tos};
 use neqo_transport::Output;
 
 use super::Node;
 
 /// One second in nanoseconds.
 const ONE_SECOND_NS: u128 = 1_000_000_000;
+
+#[derive(Clone)]
+struct Stats {
+    /// The number of packets received.
+    received: usize,
+    /// The number of packets marked.
+    marked: usize,
+    /// The number of packets dropped.
+    dropped: usize,
+    /// The number of packets delivered.
+    delivered: usize,
+    /// The maximum amount of queue capacity ever used.
+    /// As packets leave the queue as soon as they start being used, this doesn't
+    /// count them.
+    maxq: usize,
+}
+
+impl Stats {
+    const fn new() -> Self {
+        Self {
+            received: 0,
+            marked: 0,
+            dropped: 0,
+            delivered: 0,
+            maxq: 0,
+        }
+    }
+
+    fn update_maxq(&mut self, used: usize) {
+        self.maxq = max(self.maxq, used);
+    }
+}
+
+impl Display for Stats {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "rx {} mark {} drop {} tx {} maxq {}",
+            self.received, self.marked, self.dropped, self.delivered, self.maxq
+        )
+    }
+}
 
 /// This models a link with a tail drop router at the front of it.
 #[derive(Clone)]
@@ -31,6 +73,9 @@ pub struct TailDrop {
     rate: usize,
     /// The depth of the queue, in bytes.
     capacity: usize,
+    /// The depth of the queue where arriving packets get marked.
+    /// Set this to `capacity` or higher so that it has no effect.
+    mark_capacity: usize,
 
     /// A counter for how many bytes are enqueued.
     used: usize,
@@ -46,36 +91,25 @@ pub struct TailDrop {
     /// The packets that are on the link and when they can be delivered.
     on_link: VecDeque<(Instant, Datagram)>,
 
-    /// The number of packets received.
-    received: usize,
-    /// The number of packets dropped.
-    dropped: usize,
-    /// The number of packets delivered.
-    delivered: usize,
-    /// The maximum amount of queue capacity ever used.
-    /// As packets leave the queue as soon as they start being used, this doesn't
-    /// count them.
-    maxq: usize,
+    stats: Stats,
 }
 
 impl TailDrop {
     /// Make a new taildrop node with the given rate, queue capacity, and link delay.
     #[must_use]
-    pub const fn new(rate: usize, capacity: usize, delay: Duration) -> Self {
+    pub const fn new(rate: usize, capacity: usize, mark_capacity: usize, delay: Duration) -> Self {
         Self {
             overhead: 64,
             rate,
             capacity,
+            mark_capacity,
             used: 0,
             queue: VecDeque::new(),
             next_deque: None,
             sub_ns_delay: 0,
             delay,
             on_link: VecDeque::new(),
-            received: 0,
-            dropped: 0,
-            delivered: 0,
-            maxq: 0,
+            stats: Stats::new(),
         }
     }
 
@@ -83,13 +117,13 @@ impl TailDrop {
     /// with a fat 32k buffer (about 30ms), and the default forward delay of 50ms.
     #[must_use]
     pub const fn dsl_downlink() -> Self {
-        Self::new(1_000_000, 32_768, Duration::from_millis(50))
+        Self::new(1_000_000, 32_768, 32_768, Duration::from_millis(50))
     }
 
     /// Cut uplink to one fifth of the downlink (2Mbps), and reduce the buffer to 1/4.
     #[must_use]
     pub const fn dsl_uplink() -> Self {
-        Self::new(200_000, 8_192, Duration::from_millis(50))
+        Self::new(200_000, 8_192, 8_192, Duration::from_millis(50))
     }
 
     /// How "big" is this datagram, accounting for overheads.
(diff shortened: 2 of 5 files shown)


larseggert:
When you say "marking at 0.8", do you mean you mark **all** packets when occupancy exceeds 80% of the capacity and none before?

We probably need to do RED or WRED. See for example https://www.juniper.net/documentation/us/en/software/junos/cos/topics/example/cos-schedulers-explicit-congestion-notification-configuring.html

martinthomson:
<details>
<summary>In case anyone wants to investigate the anomaly...</summary>

I can confirm that setting SIMULATION_SEED doesn't lead to this being reproducible.  And that logging at the debug level is unworkable amounts of output; this only logs info.  I'm not seeing anything here that would indicate that the run was "special" in any way.

```
0.000 INFO gbit-bandwidth: seed 38e87509a754345d2d523b7c56fa8aae697651dcf725370b93f31337c3dd1688
0.000 INFO [Server ...] Received valid Initial packet with scid CID [17]: 11000000006f14ad1d9fb688d9eac9a52a dcid CID [8]: 68877b03a1939bca
0.001 INFO Compatible upgrade Version1 ==> Version2
0.006 WARN [recovery::Loss] ignoring in-3 from dropped space
0.006 INFO [Client 68877b03a1939bca] Connection established
0.006 INFO [Server 68877b03a1939bca] Connection established
0.006 INFO gbit-bandwidth: Setup took 6.916621ms (wall) 80.050192ms (simulated)
0.007 INFO ECN validation succeeded, path is capable
0.007 INFO ECN validation succeeded, path is capable
5.498 INFO gbit-bandwidth: Simulation took 5.491255838s (wall) 9.935127471s (simulated)
5.498 INFO gbit-bandwidth: stats for Client ...
  rx: 908781 drop 1 dup 0 saved 0
  tx: 11968 lost 0 lateack 0 ptoack 0 unackdrop 0
  pmtud: 0 sent 0 acked 0 lost 0 change 1500 iface_mtu 0 pmtu
  resumed: false
  frames rx:
    crypto 3 done 1 token 0 close 0
    ack 1390 (max 11920) ping 0 padding 0
    stream 907386 reset 0 stop 0
    max: stream 0 data 0 stream_data 0
    blocked: stream 0 data 0 stream_data 1393
    datagram 0
    ncid 7 rcid 0 pchallenge 0 presponse 0
    ack_frequency 3
  frames tx:
    crypto 4 done 0 token 0 close 0
    ack 10574 (max 909929) ping 0 padding 0
    stream 0 reset 0 stop 0
    max: stream 1 data 0 stream_data 1408
    blocked: stream 0 data 0 stream_data 0
    datagram 0
    ncid 7 rcid 0 pchallenge 0 presponse 0
    ack_frequency 1
  ecn:
    tx:
      Initial Count({NotEct: 4, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 2, Ect1: 0, Ect0: 0, Ce: 0})
      Short Count({NotEct: 13, Ect1: 0, Ect0: 11949, Ce: 0})
    acked:
      Short Count({NotEct: 0, Ect1: 0, Ect0: 11908, Ce: 0})
    rx:
      Initial Count({NotEct: 3, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 1, Ect1: 0, Ect0: 1, Ce: 0})
      Short Count({NotEct: 2, Ect1: 0, Ect0: 907550, Ce: 1223})
    path validation outcomes: ValidationCount({Capable: 1, NotCapable(BlackHole): 0, NotCapable(Bleaching): 0, NotCapable(ReceivedUnsentECT1): 0})
    mark transitions:
      First NotEct to Ect0 (Handshake, 1)
      First Ect0 to NotEct (Short, 10) to Ce (Short, 8014)
      First Ce to Ect0 (Short, 9696)
  dscp: Cs0: 908781

5.498 INFO gbit-bandwidth: taildrop: rx 11966 mark 0 drop 0 tx 11950 maxq 1312
5.498 INFO gbit-bandwidth: stats for Server ...
  rx: 11955 drop 3 dup 0 saved 0
  tx: 910014 lost 1230 lateack 0 ptoack 0 unackdrop 0
  pmtud: 0 sent 0 acked 0 lost 0 change 1500 iface_mtu 0 pmtu
  resumed: false
  frames rx:
    crypto 4 done 0 token 0 close 0
    ack 10561 (max 907937) ping 0 padding 0
    stream 0 reset 0 stop 0
    max: stream 0 data 0 stream_data 1406
    blocked: stream 0 data 0 stream_data 0
    datagram 0
    ncid 7 rcid 0 pchallenge 0 presponse 0
    ack_frequency 1
  frames tx:
    crypto 3 done 1 token 0 close 0
    ack 1394 (max 11942) ping 0 padding 0
    stream 908616 reset 0 stop 0
    max: stream 0 data 0 stream_data 0
    blocked: stream 0 data 0 stream_data 1393
    datagram 0
    ncid 7 rcid 0 pchallenge 0 presponse 0
    ack_frequency 3
  ecn:
    tx:
      Initial Count({NotEct: 3, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 1, Ect1: 0, Ect0: 1, Ce: 0})
      Short Count({NotEct: 2, Ect1: 0, Ect0: 910007, Ce: 0})
    acked:
      Short Count({NotEct: 0, Ect1: 0, Ect0: 905483, Ce: 1223})
    rx:
      Initial Count({NotEct: 4, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 2, Ect1: 0, Ect0: 0, Ce: 0})
      Short Count({NotEct: 13, Ect1: 0, Ect0: 11933, Ce: 0})
    path validation outcomes: ValidationCount({Capable: 1, NotCapable(BlackHole): 0, NotCapable(Bleaching): 0, NotCapable(ReceivedUnsentECT1): 0})
    mark transitions:
      First NotEct to Ect0 (Short, 1)
      First Ect0 to NotEct (Short, 11)
  dscp: Cs0: 11955

5.498 INFO gbit-bandwidth: taildrop: rx 910012 mark 1223 drop 1230 tx 908778 maxq 3999700
5.544 INFO Achieved 864.6023533239477 Mb/s bandwidth (link rate 1000)
```

</details>

martinthomson:
Oh, yeah, I should try RED.  Are there recommended curves for that?  I'm looking for one with simple config.

larseggert:
That (= too many knobs and little guidance how to set them) was always the problem with RED.

IIRC Sally's original paper had a heuristic. Or maybe just do what the Juniper config does and see what you get...

martinthomson:
How about this then:  For a queue of size `size` and capacity `capacity`, the formula is:

```rust
let r = clamp(0, 2 * ((size / capacity) - 0.3), 0.95);
if rand(0, 1) < r.pow(3) {
  mark(d)
} else {
  d
}
```

That is, follow a cubic curve that goes from 0 at 30% capacity to 1 at 80%, but cap the mark rate to 95% just before the 80% threshold.

That's close to what I see in the recommendations.  I'll sketch out what that looks like in practice.

martinthomson:
Here's the results of the min_bandwidth test without ECN:

```
5.664 INFO gbit-bandwidth: taildrop: rx 12664 mark 0 drop 0 tx 12651 maxq 1312
5.664 INFO gbit-bandwidth: taildrop: rx 909314 mark 0 drop 6506 tx 902803 maxq 3999968
5.710 INFO Achieved 861.4218453708036 Mb/s bandwidth (link rate 1000)
5.843 INFO gbit-bandwidth: taildrop: rx 14214 mark 0 drop 0 tx 14201 maxq 1312
5.843 INFO gbit-bandwidth: taildrop: rx 908123 mark 0 drop 6700 tx 901421 maxq 3999886
5.889 INFO Achieved 856.8743925466188 Mb/s bandwidth (link rate 1000)
5.638 INFO gbit-bandwidth: taildrop: rx 14310 mark 0 drop 0 tx 14294 maxq 1312
5.638 INFO gbit-bandwidth: taildrop: rx 906308 mark 0 drop 8926 tx 897382 maxq 3999498
5.685 INFO Achieved 818.5208580279693 Mb/s bandwidth (link rate 1000)
5.601 INFO gbit-bandwidth: taildrop: rx 14310 mark 0 drop 0 tx 14294 maxq 1312
5.601 INFO gbit-bandwidth: taildrop: rx 906308 mark 0 drop 8926 tx 897382 maxq 3999498
5.648 INFO Achieved 818.5208581059649 Mb/s bandwidth (link rate 1000)
```

And with RED, following that pattern:

```
5.277 INFO gbit-bandwidth: taildrop: rx 16570 mark 0 drop 0 tx 16555 maxq 1312
5.277 INFO gbit-bandwidth: taildrop: rx 899086 mark 118 drop 0 tx 899084 maxq 2635373
5.324 INFO Achieved 534.9536758403262 Mb/s bandwidth (link rate 1000)
5.313 INFO gbit-bandwidth: taildrop: rx 16570 mark 0 drop 0 tx 16555 maxq 1312
5.313 INFO gbit-bandwidth: taildrop: rx 899086 mark 118 drop 0 tx 899084 maxq 2635373
5.361 INFO Achieved 534.9537810497656 Mb/s bandwidth (link rate 1000)
5.256 INFO gbit-bandwidth: taildrop: rx 16570 mark 0 drop 0 tx 16555 maxq 1312
5.256 INFO gbit-bandwidth: taildrop: rx 899086 mark 118 drop 0 tx 899084 maxq 2635373
5.305 INFO Achieved 534.9539912355617 Mb/s bandwidth (link rate 1000)
5.253 INFO gbit-bandwidth: taildrop: rx 16570 mark 0 drop 0 tx 16555 maxq 1312
5.253 INFO gbit-bandwidth: taildrop: rx 899086 mark 118 drop 0 tx 899084 maxq 2635373
5.300 INFO Achieved 534.9539912355617 Mb/s bandwidth (link rate 1000)
```

I need to validate the code, but that seems like a pretty big change.  ECN is definitely working in that it prevents drops, but it seems to be causing us to flip into congestion avoidance much earlier and we seem to detect persistent congestion and drop back into slow start.  With the seed I'm using, this is just one persistent congestion patch, but the slow start threshold is very low relative to what I've seen in other tests.

We first exit slow start at 3455589 and then 2425811 (approximately).  If I disable ECN, we exit more often, but at much higher windows: 11633690, 8143583, 5700508, and 3998979.  All higher than when ECN is enabled, which is fine, but we don't have enough time for the window to grow large enough to matter in this test with a simulated time of 10s.

larseggert:
I have previously seen us declare persistent congestion pretty aggressively. I wonder if there is a bug there.

larseggert:
(Aside: Back when I was a student, we used to joke that if you wanted to do congestion control stuff with ns2, you first needed to fix ns2. I guess that - still! - holds independent of a particular simulator...)

martinthomson:
Yeah, that is not surprising :)

Here we go then, copying my commit comment:

I've added some stats that show some very interesting results.

The most interesting result here is the buffer utilization metric I've added as part of this.

A no-ECN run with a 4ms target buffer latency got 0.42051 utilization.
A run with ECN enabled uses a buffer that is twice the size.  It uses
the cubic curve that runs from 0 to 95% mark rate between 30 and 80% of
the buffer utilization ends up with 0.13187 utilization.
Those equate to no-ECN averaging 1682024 bytes in the buffer
and ECN averaging 1054930 bytes in the buffer.

Because the ECN does have a larger buffer, the maximum utilization in
bytes is 7744121 against the non-ECN 3999886 which completely fills up
in operation.  **ECN loses no packets**, no-ECN loses 6700.

The adjustment that doubles the buffer for the ECN-enabled case allows
throughput to be fairly similar. Without that, performance tanks badly.
This makes sense in a way: ECN means that we can't really use most of
the buffer (see above) so we lose the throughput benefits of buffering.

I suspected that we're hitting persistent congestion, but the log
statement there never shows in logs.  So it's not that.

What I see in the logs is that we get more congestion events with ECN
*disabled*.  We get fewer congestion events with ECN,
but they are at a smaller congestion window.

There's a bunch of variability in the tests, even with a fixed seed for
the simulator.  It might pay to have a different configuration that
doesn't vary the length of connection IDs for some tests.  We can even
disable connection IDs.  That's the only thing I can think of that might
affect reproducibility of tests.

test-fixture/src/sim/taildrop.rs
```diff
@@ -7,20 +7,107 @@
...
+            marked: 0,
+            dropped: 0,
+            delivered: 0,
+            maxq: 0,
+            capacity: 0,
+            start_time: None,
+            last_update: None,
+            cumulative_usage: 0,
+        }
+    }
```
larseggert, line 66: Derive `default` and use it instead?

test-fixture/src/sim/taildrop.rs
```diff
@@ -31,65 +118,67 @@ pub struct TailDrop {
...
+    /// Whether to apply ECN markings.
+    ecn: bool,
 
     /// A counter for how many bytes are enqueued.
     used: usize,
     /// A queue of unsent bytes.
     queue: VecDeque<Datagram>,
     /// The time that the next datagram can enter the link.
-    next_deque: Option<Instant>,
+    next_deque: Option<(Instant, u32)>,
```
larseggert, line 130: Comment doesn't say what the new `u32` is for. Is it the removed `sub_ns_delay`?

test-fixture/src/sim/taildrop.rs
```diff
@@ -124,38 +213,92 @@ impl TailDrop {
...
+    fn maybe_mark(&mut self, dgram: Datagram) -> Datagram {
+        if self.ecn && Ecn::from(dgram.tos()) == Ecn::Ect0 && self.should_mark(self.used) {
+            qtrace!("taildrop marking {} bytes", dgram.len());
+            self.stats.marked += 1;
+            let tos = Tos::from((Dscp::from(dgram.tos()), Ecn::Ce));
+            Datagram::new(dgram.source(), dgram.destination(), tos, dgram.to_vec())
+        } else {
+            dgram
+        }
+    }
```
larseggert: Also apply CE for ECT(1), they should be identical. (Not that we use ECT(1) yet.)

test-fixture/src/sim/taildrop.rs
```diff
@@ -124,38 +213,92 @@ impl TailDrop {
...
+
+    fn should_mark(&self, used: usize) -> bool {
+        // Apply RED which starts at 0 mark chance at 30% of the capacity.
+        // From there, follow a cubic curve that reaches 1 at 80% capacity.
+        // Cap at around 95% mark probability.
+        //
+        // let p = (2 * ((used / capacity) - 0.3));
+        // if rand(0, 1) < p.pow(3).clamp(0, 0.95) { mark(d) } else { d }
+        //
+        // This code scales that up by a factor of 1024 so we can use integers.
```
larseggert, line 279: Do we care that we can use integers? This isn't kernel code.

test-fixture/src/sim/taildrop.rs
martinthomson: `Default` isn't `const`, which turned out to be a necessity for some reason.  (I think that it's because of how we set up the simulator tests with macros.)

test-fixture/src/sim/taildrop.rs
martinthomson: Yeah, it was cleaner that way.

test-fixture/src/sim/taildrop.rs
martinthomson: OK, good.  I was under the impression that we'd need a different algorithm.  Easy fix.

test-fixture/src/sim/taildrop.rs
martinthomson: The numbers we are using are big enough that they won't fit in the mantissa of `f64`.  That might not be a problem, because we don't care about the rounding error accumulation really.  It's more that the random number generator only produces an integer.  Teaching it to produce a float is harder than moving to ints here.  It's a little fiddly here, but not overly so.

test-fixture/src/sim/taildrop.rs
mxinden: I assume we would only start marking ECT(1) once we implement some L4S congestion controller.

L4S requires the bottleneck to implement a [dual-queue AQM](https://www.rfc-editor.org/rfc/rfc9332.html?utm_source=chatgpt.com), separating ECT(0) and ECT(1) flows. In addition, ECT(1) flows should have a lower marking threshold.

This is to say that the `TailDrop` implementation here is not yet ready for L4s. Under the assumption that we don't do ECT(1) without L4S, I suggest simply panicking here when we see ECT(1). That way, once we implement an L4S congestion controller, we know what to fix.

neqo-transport/benches/min_bandwidth.rs
```diff
@@ -20,13 +20,15 @@ use test_fixture::{
     boxed,
     sim::{
         connection::{Node, ReachState, ReceiveData, SendData},
-        network::{Delay, Mtu, TailDrop},
+        network::{Mtu, TailDrop},
         Simulator,
     },
 };
 
+/// Run a trasfer of a gigabyte over a gigabit link.
```
mxinden:
```suggestion
/// Run a transfer of a gigabyte over a gigabit link.
```

neqo-transport/benches/min_bandwidth.rs
```diff
@@ -35,52 +37,61 @@ pub fn main() {
...
-        achieved_bandwidth / MBIT as f64,
+        "{name} expected to reach {MINIMUM_EXPECTED_UTILIZATION} of maximum bandwidth ({t} Mbit/s) but got {a} Mbit/s",
+        t = LINK_BANDWIDTH / MBIT,
+        a = achieved_bandwidth / MBIT as f64,
     );
 }
+
+fn main() {
+    gbit_bandwidth(false);
+    gbit_bandwidth(true);
```
mxinden, line 99: :+1: In favor of testing both with and without ECN. For what it is worth, we see ~50% of all connections being ECN capable.

neqo-transport/benches/min_bandwidth.rs
```diff
@@ -35,52 +37,61 @@ pub fn main() {
...
+    /// How much of the theoretical bandwidth we will expect to deliver.
+    const MINIMUM_EXPECTED_UTILIZATION: f64 = 0.75;
 
     let gbit_link = || {
         let rate_byte = LINK_BANDWIDTH / 8;
-        // Router buffer set to bandwidth-delay product.
-        let capacity_byte = rate_byte * LINK_RTT_MS / 1_000;
-        TailDrop::new(rate_byte, capacity_byte, Duration::ZERO)
+        // Set capacity to double when ECN is enabled.
+        let capacity_byte = LINK_BANDWIDTH * BUFFER_LATENCY_MS * (1 + usize::from(ecn)) / 1000;
```
mxinden:
```suggestion
        let capacity_byte = LINK_BANDWIDTH * BUFFER_LATENCY_MS * (1 + usize::from(ecn)) / 1000;
```

`LINK_BANDWIDTH` is in bits, yet the result is `capacity_byte`. Aren't you missing a `/ 8` here?

mxinden:
```suggestion
        let capacity_byte = LINK_BANDWIDTH * BUFFER_LATENCY_MS / 1000 * (1 + usize::from(ecn));
```

Eases understanding that `/ 1000` converts MS to S.

martinthomson:
I don't think that we want a 32ms buffer on a 40ms RTT, so I've fixed the computation for buffer depth there.  A shallow buffer has a pretty significant impact on transfer throughput.

With a 4ms maximum latency on the buffer, here's where things stand:

```
5.379 INFO gbit-bandwidth-noecn: taildrop: rx 27494 mark 0 drop 0 tx 27449 maxq 1312 util 0.00000000007909192866063154
5.379 INFO gbit-bandwidth-noecn: taildrop: rx 897018 mark 0 drop 2686 tx 894324 maxq 499872 util 0.020684066259229614
5.426 INFO gbit-bandwidth-noecn achieved 327.3515869778989 Mb/s bandwidth (link rate 1000)
10.607 INFO gbit-bandwidth-ecn: taildrop: rx 23889 mark 0 drop 0 tx 23858 maxq 1312 util 0.000000000022708908805627867
10.607 INFO gbit-bandwidth-ecn: taildrop: rx 895726 mark 248 drop 193 tx 895524 maxq 999744 util 0.012177479163127468
10.652 INFO gbit-bandwidth-ecn achieved 346.1879087008973 Mb/s bandwidth (link rate 1000)
5.314 INFO gbit-bandwidth-noecn: taildrop: rx 27494 mark 0 drop 0 tx 27449 maxq 1312 util 0.00000000007909192866063154
5.314 INFO gbit-bandwidth-noecn: taildrop: rx 897018 mark 0 drop 2686 tx 894324 maxq 499872 util 0.020684066259229614
5.360 INFO gbit-bandwidth-noecn achieved 327.3515869778989 Mb/s bandwidth (link rate 1000)
10.569 INFO gbit-bandwidth-ecn: taildrop: rx 26290 mark 0 drop 0 tx 26259 maxq 1312 util 0.000000000018445580879511176
10.569 INFO gbit-bandwidth-ecn: taildrop: rx 896395 mark 242 drop 151 tx 896236 maxq 999744 util 0.011802812326471894
10.614 INFO gbit-bandwidth-ecn achieved 342.98631400841225 Mb/s bandwidth (link rate 1000)
```

That is, the deeper buffer for ECN allows it to slightly outperform the smaller drop-only buffer.  ECN seems to be marking relatively few packets, but the shallower buffer means it loses a few as well.  Both are a lot lower than the overall non-ECN drop rate.  Latency is roughly the same for both.

Note the absolute stability of the estimate for the no-ECN case, which has no variability.  Using the same seed for ECN makes that completely stable also.

It seems like in both cases, we're not done probing for the link bandwidth, because increasing the transfer size leads to a doubling of the bandwidth (at a much higher cost in time, of course).  This is a 5GB transfer:

```
26.894 INFO gbit-bandwidth-noecn: taildrop: rx 138000 mark 0 drop 0 tx 137948 maxq 1312 util 0.00000000003186549070825237
26.894 INFO gbit-bandwidth-noecn: taildrop: rx 4494054 mark 0 drop 2686 tx 4491360 maxq 499872 util 0.2528884600997324
27.124 INFO gbit-bandwidth-noecn achieved 659.4401650325442 Mb/s bandwidth (link rate 1000)
53.754 INFO gbit-bandwidth-ecn: taildrop: rx 63312 mark 0 drop 0 tx 63302 maxq 1312 util 0.000000000011777127945920991
53.754 INFO gbit-bandwidth-ecn: taildrop: rx 4481470 mark 238 drop 197 tx 4481271 maxq 999744 util 0.05634958417594766
53.978 INFO gbit-bandwidth-ecn achieved 609.3594807656492 Mb/s bandwidth (link rate 1000)
26.748 INFO gbit-bandwidth-noecn: taildrop: rx 138000 mark 0 drop 0 tx 137948 maxq 1312 util 0.00000000003186549070825237
26.748 INFO gbit-bandwidth-noecn: taildrop: rx 4494054 mark 0 drop 2686 tx 4491360 maxq 499872 util 0.2528884600997324
26.976 INFO gbit-bandwidth-noecn achieved 659.4401650325442 Mb/s bandwidth (link rate 1000)
53.374 INFO gbit-bandwidth-ecn: taildrop: rx 65449 mark 0 drop 0 tx 65432 maxq 1312 util 0.000000000005516942351745791
53.374 INFO gbit-bandwidth-ecn: taildrop: rx 4478564 mark 249 drop 179 tx 4478383 maxq 999744 util 0.06162337817487789
53.600 INFO gbit-bandwidth-ecn achieved 639.6429688410446 Mb/s bandwidth (link rate 1000)
```

The lower throughput of ECN is concerning here.  A 10GB transfer hints at the ECN marks causing constant reductions in the estimated window throughout.  This significantly affects throughput relative to the non-ECN variant, which has absolute stability in this (unrealistic) test:

```
51.638 INFO gbit-bandwidth-noecn: taildrop: rx 275967 mark 0 drop 0 tx 275907 maxq 1312 util 0.000000000018266656165795572
51.638 INFO gbit-bandwidth-noecn: taildrop: rx 8989796 mark 0 drop 2686 tx 8987104 maxq 499872 util 0.3238115755224498
52.092 INFO gbit-bandwidth-noecn achieved 756.0406716896754 Mb/s bandwidth (link rate 1000)
102.995 INFO gbit-bandwidth-ecn: taildrop: rx 118263 mark 0 drop 0 tx 118248 maxq 1312 util 0.00000000000951029712311687
102.995 INFO gbit-bandwidth-ecn: taildrop: rx 8961494 mark 263 drop 184 tx 8961309 maxq 999744 util 0.05213906059487346
103.440 INFO gbit-bandwidth-ecn achieved 616.0275957113346 Mb/s bandwidth (link rate 1000)
```

I also played tried shifting the guard rails, moving from 0.3&ndash;0.8 to 0.4&ndash;0.9.  That produced a lot more dropping and marking overall, by a factor of around 2, plus a slightly improved throughput:

```
10.059 INFO gbit-bandwidth-ecn: taildrop: rx 19701 mark 0 drop 0 tx 19677 maxq 1312 util 0.00000000001967531511159024
10.059 INFO gbit-bandwidth-ecn: taildrop: rx 892521 mark 723 drop 1571 tx 890949 maxq 999744 util 0.01608063787815736
10.107 INFO gbit-bandwidth-ecn achieved 369.84572191083333 Mb/s bandwidth (link rate 1000)
```

Finally, I've moved to a quadratic, rather than a cubic.  There isn't a lot of difference in the tests I've run, but the code is simpler.  It also seems to avoid some of the instability I saw in the larger transfer sizes (though that takes long enough that I'm not enthusiastic about doing a more thorough set of tests).  Here's a couple of larger transfers (8GB and 12GB) with the updated settings:

```
42.983 INFO gbit-bandwidth-noecn: taildrop: rx 220780 mark 0 drop 0 tx 220722 maxq 1312 util 0.000000000022026483517401776
42.983 INFO gbit-bandwidth-noecn: taildrop: rx 7191656 mark 0 drop 2686 tx 7188961 maxq 499872 util 0.3041900787040877
43.348 INFO gbit-bandwidth-noecn achieved 729.3254333409886 Mb/s bandwidth (link rate 1000)
86.055 INFO gbit-bandwidth-ecn: taildrop: rx 84629 mark 0 drop 0 tx 84614 maxq 1312 util 0.000000000005776930965333724
86.055 INFO gbit-bandwidth-ecn: taildrop: rx 7160651 mark 809 drop 1548 tx 7159102 maxq 999856 util 0.0832548796697331
86.415 INFO gbit-bandwidth-ecn achieved 712.4773860878167 Mb/s bandwidth (link rate 1000)
64.558 INFO gbit-bandwidth-noecn: taildrop: rx 331148 mark 0 drop 0 tx 331093 maxq 1312 util 0.000000000015604258766155856
64.558 INFO gbit-bandwidth-noecn: taildrop: rx 10788062 mark 0 drop 2686 tx 10785367 maxq 499872 util 0.3377375525003566
65.118 INFO gbit-bandwidth-noecn achieved 775.0154747588863 Mb/s bandwidth (link rate 1000)
129.444 INFO gbit-bandwidth-ecn: taildrop: rx 114602 mark 0 drop 0 tx 114586 maxq 1312 util 0.000000000004038731317681646
129.444 INFO gbit-bandwidth-ecn: taildrop: rx 10740556 mark 837 drop 1551 tx 10739004 maxq 999856 util 0.08988986454302042
129.971 INFO gbit-bandwidth-ecn achieved 747.1544004343607 Mb/s bandwidth (link rate 1000)
```

Note how the drop rate is almost the same as the basic test.  It looks like Cubic CC is good at dialing in a bandwidth estimate given absolutely stable conditions.  Having ECN disabled gives an edge here, which is likely due to that absolute stability relative to the random markings from ECN, but the difference is pretty small.  ECN has a pretty significant advantage on latency throughout, even though it could use a deeper buffer.  Latency is around half with ECN enabled.  Given that we only start marking once the buffer hits 80% of the non-ECN buffer capacity, that seems pretty good.

larseggert:
> Finally, I've moved to a quadratic, rather than a cubic.

Have you tried a simple linear relation? IIRC that is what most routers offer.

martinthomson:
> Have you tried a simple linear relation? IIRC that is what most routers offer.

It doesn't seem to change the outcome much.  There's a small change if the starting threshold changes, not the shape.  That is, if we start dropping at 30% of the buffer, then little more than 30% of the buffer is available and we drop out of slow start quicker.  That determines when we drop out of the exponential window growth phase, which seems to be still the biggest contributor to this test case.

The variability from having different drop patterns (that is, the different seeds to the RNG) seem most important in terms of the final outcome.  A linear ramp is maybe less variable overall, because marks are applied more consistently close to the low threshold.  I'm still seeing a 7-8% variability in overall bandwidth.

My guess is that we're not going to see a lot of extra value trying to get more from this one contrived test.  We'll need to get a few more to test.  Including having some simulation of variable queue occupancy from cross-traffic.

Issue #2917, opened 2025-08-31
Q: Why not send PMTUD probe in coalesced packet?


mxinden:
Currently we don't send a PMTUD probe in a coalesced packet:

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2456-L2463

Why is that?

Do we, for the sake of stability, not send probes in initial and handshake space, and thus not when coalescing spaces?

@larseggert do you recall?

If the above is true, I don't think we implement it correctly.

1. We set the `PacketBuilder` `limit` early on in `output_dgram_on_path`. The `limit` is based on whether we want to send a probe or not. Note that it does not depend on whether the packet is coalesced.

    https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2644-L2662

2. Later on, iff it is `!coalesced`, we add a ping frame and mark it as a probing frame:

    https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2455-L2464

3. As a final step, we add the actual padding based on the limit in (1), but not under the condition of `!coalesced`.

    https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2496

In other words we can have a PMTUD probe limit (1), no ping and not marked as a probe (2), but still get full probe padding (3).

Given that we didn't mark the probe (2), we might attempt to send another probe right after.

larseggert:
I don't recall why I did it this way; I think the intent might have been to wait with PMTUD until after the handshake, in order to not complicate that mechanism further.

Issue #2915, opened 2025-08-29
GSO with PMTUD produces inconsistent segment sizes


mxinden:
With GSO enabled, i.e. `max_datagrams` `> 1`, Neqo produces a `DatagramBatch`.

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L1193-L1197

Each `Datagram` within the `DatagramBatch` is of size `DatagramBatch::datagram_size`. The last and only the last `Datagram` in the `DatagramBatch` is allowed to be smaller than `DatagramBatch::datagram_size`

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-common/src/datagram.rs#L141-L147

The above invariant is enforced in `Connection::output_dgram_batch_on_path`:

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2558-L2568

What the above code does not account for is the fact that the `path.borrow().plpmtu()` size might change between `Datagram`s when producing a `DatagramBatch` in `output_dgram_batch_on_path` when PMTUD is enabled. Thus a `Datagram` can be larger than its predecessor within a `DatagramBatch`.

In such case, the OS will not correctly send out the `DatagramBatch`, and thus all `Datagram`s will be lost.

In addition, in debug mode, the above can lead to an integer underflow in `DatagramBatch::MAX - send_buffer.len()`. In a previous iteration `plpmtud()` returned a datagram size that would barely fit into `send_buffer`, but while producing the `Datagram`, `Pmtud` would decide to increase the mtu and thus the datagram size, thus overfilling `send_buffer`. Thereby `send_buffer` would exceed `DatagramBatch::MAX` in the next iteration and thus the subtraction would underflow.

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/mod.rs#L2565C36-L2565C74

Panic reported by @valenting when testing MASQUE connect-udp Neqo server 🙏 

Note that this is not an issue in Firefox today. Firefox does do GSO (on Android and Linux), but does not yet do PMTUD. See [Bugzilla Bug 1909910](https://bugzilla.mozilla.org/show_bug.cgi?id=1909910).

[...]

PR #2918, opened 2025-08-31
chore(transport): assert GSO requirements and break on PMTUD probe

mxinden:
Assert GSO requirements in `output_dgram_batch_on_path` in debug mode. Might catch some future bugs when combining PMTUD and GSO. In release mode, ignore the error, causing the batch to be dropped by the receiver only.

In addition, on PMTUD probe, close GSO batch to not violate GSO requirements.

---

Related: https://github.com/mozilla/neqo/issues/2915, https://github.com/mozilla/neqo/issues/2917

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,12 +105,13 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.0.89"
+version = "1.2.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2580,12 +2589,20 @@ impl Connection {
                 packet_tos,
             )? {
                 SendOption::Yes => {
+                    debug_assert_eq!(
+                        mtu,
+                        path.borrow().plpmtu(),
+                        "mtu does not change within batch"
```
larseggert:
```suggestion
                        "MTU does not change within batch"
```

larseggert (approved):

PR #2926, opened 2025-09-01
refactor(bin): derive Clone for Args

mxinden:
Previousy `Args` would not implement `Clone`, thus requiring lifetime parameters for `Args` references in various places in `neqo-bin`.

This commit derives `Clone` for `Args`, thus simplifying the code.

Needed in and pulled out of https://github.com/mozilla/neqo/pull/2796/.

### DIFF
--- a/neqo-bin/src/client/http3.rs
+++ b/neqo-bin/src/client/http3.rs
@@ -34,16 +34,17 @@ use url::Url;
 use super::{get_output_file, qlog_new, Args, CloseState, Res};
 use crate::{send_data::SendData, STREAM_IO_BUFFER_SIZE};
 
-pub struct Handler<'a> {
+pub struct Handler {
     #[expect(clippy::struct_field_names, reason = "This name is more descriptive.")]
-    url_handler: UrlHandler<'a>,
+    url_handler: UrlHandler,
     token: Option<ResumptionToken>,
     output_read_data: bool,
     read_buffer: Vec<u8>,
 }
 
-impl<'a> Handler<'a> {
-    pub(crate) fn new(url_queue: VecDeque<Url>, args: &'a Args) -> Self {
+impl Handler {
+    pub(crate) fn new(url_queue: VecDeque<Url>, args: Args) -> Self {
+        let output_read_data = args.output_read_data;
         let url_handler = UrlHandler {
             url_queue,
             handled_urls: Vec::new(),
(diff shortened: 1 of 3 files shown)


neqo-bin/src/client/http3.rs
```diff
@@ -55,10 +56,18 @@ impl<'a> Handler<'a> {
...
         }
     }
+
+    fn reinit(&mut self) {
+        for url in self.url_handler.handled_urls.drain(..) {
+            self.url_handler.url_queue.push_front(url);
+        }
+        self.url_handler.stream_handlers.clear();
+        self.url_handler.all_paths.clear();
+    }
```
mxinden, line 70: Just moved up from below, as it can now share the same `impl Handler {` block.

neqo-bin/src/client/http3.rs
```diff
@@ -55,10 +56,18 @@ impl<'a> Handler<'a> {
...
-            output_read_data: args.output_read_data,
+            output_read_data,
             read_buffer: vec![0; STREAM_IO_BUFFER_SIZE],
         }
     }
+
+    fn reinit(&mut self) {
+        for url in self.url_handler.handled_urls.drain(..) {
+            self.url_handler.url_queue.push_front(url);
+        }
```
martinthomson, line 67: I know that this is not changed, but does this really need to pull one by one?  This could use `VecDeque::append` if we didn't care about ordering.  We clearly don't because this reverses the order of the drained URLs.  So this might be:

```suggestion
        self.url_handler.url_queue.append(self.url_handler.handled_urls.drain(..));
```

Or, if you prefer to retain order:


```suggestion
        let old_len = self.url_handler.url_queue.len();
        self.url_handler.url_queue.append(self.url_handler.handled_urls.drain(..));
        self.url_handler.url_queue.rotate_left(old_len);
```

Though that seems unlikely to be necessary.

Finally, you should move this method to `UrlHandler`, because the `Handler` should not be playing with someone else's toys.

larseggert (approved):

mxinden:
> This doesn't seem THAT much better. I know that lifetime parameters are annoying, but it was not so bad. Still, cloning here is cheap.

I should have expanded on the rational. https://github.com/mozilla/neqo/pull/2796/ needs `Args` in both the connection to the proxy and the proxied connection to the server. This lays the groundwork for it.

neqo-bin/src/client/http3.rs
mxinden: :+1: Will do this in a follow-up pull request.

mxinden: For the record, refactor happening in https://github.com/mozilla/neqo/pull/2931.

PR #2931, opened 2025-09-02
refactor(bin): move reinit to UrlHandler and maintain URL order

mxinden:
Moves Handler::reinit to UrlHandler::reinit.

Ensures URLs are re-enqueued in the same order.

Follow-up to
https://github.com/mozilla/neqo/pull/2926#discussion_r2314615625.

### DIFF
--- a/neqo-bin/src/client/http3.rs
+++ b/neqo-bin/src/client/http3.rs
@@ -60,14 +60,6 @@ impl Handler {
             read_buffer: vec![0; STREAM_IO_BUFFER_SIZE],
         }
     }
-
-    fn reinit(&mut self) {
-        for url in self.url_handler.handled_urls.drain(..) {
-            self.url_handler.url_queue.push_front(url);
-        }
-        self.url_handler.stream_handlers.clear();
-        self.url_handler.all_paths.clear();
-    }
 }
 
 pub fn create_client(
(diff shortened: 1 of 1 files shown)


neqo-bin/src/client/http3.rs
```diff
@@ -447,4 +439,12 @@ impl UrlHandler {
         self.stream_handlers.remove(&stream_id);
         self.process_urls(client);
     }
+
+    fn reinit(&mut self) {
+        for url in self.handled_urls.drain(..).rev() {
```
mxinden, line 442: Copy from above, except the inserted `rev()` to maintain the URL queue order.

mxinden: Today I learned: `rev()` uses [`DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html), which actually makes this fast. (Not implying speed is relevant here.)

martinthomson: This only maintains *approximately* the same order.  Any URL that needed a retry will land in `handled_urls` out of order.  I guess "some effort" is better than "no effort", but I would have instead done:

```rust
self.url_queue.append(&mut self.handled_urls);
```

Though this would have required changing the type of the other.

martinthomson (approved):

larseggert:
This still has the `cargo bench` panic?

neqo-bin/src/client/http3.rs
mxinden: Leaving as is. Let me know in case you feel strongly about the above.

Issue #2933, opened 2025-09-03
ci: cargo bench failure


mxinden:
Our `cargo bench` CI step is currently failing with:

```
thread 'main' panicked at /home/bench/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/criterion-0.6.0/src/kde.rs:10:66:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I am looking into it.

https://github.com/mozilla/neqo/actions/runs/17407629280/job/49431345474?pr=2931
https://github.com/mozilla/neqo/actions/runs/17403800945/job/49402890622?pr=2930

larseggert:
Have you tried pulling the update to `criterion@0.7` out of #2895 to see if this fixes it?

mxinden:
Yes. Unfortunately without success.

Panic seems to first occur in https://github.com/mozilla/neqo/pull/2897. That said, I have no clue why.

I need some more time.

larseggert:
> Panic seems to first occur in [#2897](https://github.com/mozilla/neqo/pull/2897). 

Sorry that I missed that there! I always now assume a failing bench is because of a performance regression, not a crash. (I wouldn't mind taking out the failure-on-regression...)

I have no clue what might trigger this either.

larseggert:
This seems to still happen? See e.g. https://github.com/mozilla/neqo/actions/runs/17553800617/job/49852623527?pr=2952#step:10:779

mxinden:
With https://github.com/mozilla/neqo/pull/2963 merged, closing here.

PR #2934, opened 2025-09-03
 fix(transport/benches): use separate group name

mxinden:
Criterion needs each group name to be unique. Otherwise it will panic as it does in https://github.com/mozilla/neqo/issues/2933.

---

Introduced in https://github.com/mozilla/neqo/pull/2909. Though given that no `main` baseline was available, first only surfaced in https://github.com/mozilla/neqo/pull/2897.

Fixes https://github.com/mozilla/neqo/issues/2933.

Blocked on #2943.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -176,7 +176,7 @@ jobs:
             echo
             sed -E -e 's/^                 //gi' \
                   -e '/cset:.*last message/d' \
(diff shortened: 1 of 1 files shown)


mxinden:
Ready for review. @larseggert I can fix the empty sections in the benchmark comment in a follow-up. No need to block all of you.

PR #2937, opened 2025-09-04
ci: More bencher debugging

larseggert:


larseggert:
Needs to be on `main` to run. Merging.

PR #2939, opened 2025-09-05
ci: Remove last traces of `mold`

larseggert:
We killed `mold` in #2857.

Also make our cache key a bit prettier while I'm here.


martinthomson (approved):

martinthomson:
Getting issues with --no-rosegment options to the compiler here.  Did you cut something important?

mxinden:
> Getting issues with --no-rosegment options to the compiler here. Did you cut something important?

Checking here: https://github.com/mozilla/neqo/pull/2942

PR #2941, opened 2025-09-05
ci: Debug bencher upload

larseggert:


larseggert:
Needs to be on `main` to take effect, merging.

PR #2942, opened 2025-09-05
Revert "ci: Remove last traces of `mold`"

mxinden:
Reverts mozilla/neqo#2939

Testing whether https://github.com/mozilla/neqo/pull/2939 is the source of the new `cargo bench` failure:

```
 error: linking with `cc` failed: exit status: 1
  |
  = note:  "cc" "-m64" "/tmp/rustc85mDwy/symbols.o" "<2 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,libcfg_if-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/rustc85mDwy/raw-dylibs" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/opt/actions-runner/_work/neqo/neqo/neqo/target/release/build/bindgen-77d0643b92e491a7/build_script_build-77d0643b92e491a7" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-nodefaultlibs" "-Wl,--no-rosegment,"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: /usr/bin/ld: unrecognized option '--no-rosegment'
          /usr/bin/ld: use the --help option for usage information
          collect2: error: ld returned 1 exit status
```

PR #2943, opened 2025-09-05
fix(.github/rust): remove custom linker args

mxinden:
https://github.com/mozilla/neqo/pull/2939/ removed the custom linker args leftover from when we used `mold`.

I believe now the argument order is relevant. Thus moving both to the respective usage points.

---

Follow-up to https://github.com/mozilla/neqo/pull/2939/.

Alternative to https://github.com/mozilla/neqo/pull/2942.


mxinden:
Failure in `cargo bench` command execution itself will be fixed in https://github.com/mozilla/neqo/pull/2934.

larseggert:
Thank you!

larseggert (approved):

PR #2944, opened 2025-09-05
ci: More bencher fixes

larseggert:
Fingers crossed.


larseggert:
Again, needs to live on `main` before we know if this works. Merging.

PR #2945, opened 2025-09-05
ci: More bencher upload fixes

larseggert:


larseggert:
Needs to be on `main`, merging

PR #2951, opened 2025-09-08
chore: Fix new clippy warnings

larseggert:
`cargo 1.91.0-nightly (761c4658d 2025-09-04)`

### DIFF
--- a/neqo-transport/src/connection/tests/migration.rs
+++ b/neqo-transport/src/connection/tests/migration.rs
@@ -4,6 +4,12 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#![allow(
+    clippy::allow_attributes,
+    clippy::unwrap_in_result,
+    reason = "OK in tests."
+)]
+
 use std::{
     cell::RefCell,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
@@ -705,11 +711,11 @@ fn migration_client_empty_cid() {
 fn fast_handshake(client: &mut Connection, server: &mut Connection) -> Option<Datagram> {
     let dgram = client.process_output(now()).dgram();
     let dgram2 = client.process_output(now()).dgram();
-    server.process_input(dgram?, now());
+    server.process_input(dgram.unwrap(), now());
     let dgram = server.process(dgram2, now()).dgram();
     let dgram = client.process(dgram, now()).dgram();
     let dgram = server.process(dgram, now()).dgram();
-    client.process_input(dgram?, now());
+    client.process_input(dgram.unwrap(), now());
     assert!(maybe_authenticate(client));
     let dgram = client.process_output(now()).dgram();
     server.process(dgram, now()).dgram()
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -658,6 +658,12 @@ impl IndexMut<StreamType> for LocalStreamLimits {
 
 #[cfg(test)]
 mod test {
+    #![allow(
+        clippy::allow_attributes,
+        clippy::unwrap_in_result,
+        reason = "OK in tests."
+    )]
+
     use std::{
         cmp::min,
         collections::VecDeque,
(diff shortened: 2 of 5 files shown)


neqo-bin/src/bin/client.rs
```diff
@@ -7,6 +7,11 @@
 use clap::Parser as _;
 
 #[tokio::main(flavor = "current_thread")]
+#[allow(
+    clippy::allow_attributes,
+    clippy::unwrap_in_result,
+    reason = "FIXME: False positive?"
```
martinthomson, line 13: Did you find an issue regarding this?  It's possible that there is a change in the async function desugaring that is tripping these.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -705,11 +705,11 @@ fn migration_client_empty_cid() {
 fn fast_handshake(client: &mut Connection, server: &mut Connection) -> Option<Datagram> {
     let dgram = client.process_output(now()).dgram();
     let dgram2 = client.process_output(now()).dgram();
-    server.process_input(dgram.unwrap(), now());
+    server.process_input(dgram?, now());
```
martinthomson: This is probably wrong.  When this function exits, the assumption is that the handshake is complete.  The `None` return option exists because the server might have nothing to say, not if the process fails.  This should suppress the warning.

neqo-transport/src/qlog.rs
```diff
@@ -122,7 +122,11 @@ fn connection_started(qlog: &Qlog, path: &PathRef, now: Instant) {
     );
 }
 
-#[expect(clippy::similar_names, reason = "new and now are similar.")]
+#[allow(
+    clippy::allow_attributes,
+    clippy::similar_names,
+    reason = "FIXME: Hits on MSRV <1.91."
```
martinthomson: You lost the new/now reason with this change.

neqo-udp/src/lib.rs
```diff
@@ -364,17 +368,14 @@ mod tests {
...
-                .expect("receive to succeed")
-                .for_each(|d| {
-                    assert_eq!(
-                        SEGMENT_SIZE,
-                        d.len(),
-                        "Expect received datagrams to have same length as sent datagrams"
-                    );
-                    num_received += 1;
-                });
+            receiver.recv(receiver_addr, &mut recv_buf)?.for_each(|d| {
```
martinthomson: I don't get this change: you are testing, so an expectation is fine.

neqo-udp/src/lib.rs
```diff
@@ -346,7 +348,9 @@ mod tests {
 
         let sender = socket()?;
         let receiver = socket()?;
-        let receiver_addr: SocketAddr = "127.0.0.1:0".parse().unwrap();
+        let receiver_addr: SocketAddr = "127.0.0.1:0"
+            .parse()
+            .map_err(|_| io::ErrorKind::InvalidInput)?;
```
martinthomson: I don't get this change: you are testing, so an expectation is fine.  Especially for this case, where it should be infallible.

test-fixture/src/header_protection.rs
```diff
@@ -26,7 +26,7 @@ pub use crate::{default_client, now, CountingConnectionIdGenerator};
 #[expect(clippy::type_complexity, reason = "OK in test.")]
 pub fn decode_initial_header(dgram: &Datagram, role: Role) -> Option<(&[u8], &[u8], &[u8], &[u8])> {
     let mut dec = Decoder::new(&dgram[..]);
-    let type_and_ver = dec.decode(5).unwrap().to_vec();
+    let type_and_ver = dec.decode(5)?.to_vec();
```
martinthomson: This one seems fine.  `None` means failure for this.

neqo-bin/src/bin/client.rs
larseggert: No, but I only googled for a few minutes.

neqo-transport/src/connection/tests/migration.rs
larseggert: Good catch. I mechanically replaced these and relied on tests to flag issues.

neqo-transport/src/fc.rs
```diff
@@ -658,6 +658,8 @@ impl IndexMut<StreamType> for LocalStreamLimits {
 
 #[cfg(test)]
 mod test {
+    #![expect(clippy::unwrap_in_result, reason = "OK in tests.")]
```
martinthomson: It's a shame that we can't `cfg_if()` this.  But if we do that at the top level of the crate, we lose these warnings when tests are enabled, which is in the normal validation flow.

PR #2952, opened 2025-09-08
feat: Optimize `RxStreamOrderer`

larseggert:
Courtesy of @claude. Let's see if this actually does anything.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -138,7 +138,7 @@ impl RecvStreams {
 /// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
-    // Contiguous buffer for in-order data.
+    // Buffer for arriving in-order data.
     data: VecDeque<u8>,
     offset: u64, // Offset where `data` starts.
 
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -187,7 +187,7 @@ impl RxStreamOrderer {
         }
 
         // Only use fast path for truly in-order data when there are no gaps.
-        if new_start == self.next_expected() && self.ooo.is_empty() {
+        if new_start <= self.next_expected() && new_start + new_end > self.next_expected() && self.ooo.is_empty() {
             self.store_in_order_data(new_start, new_data);
             return;
         }
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -206,8 +206,9 @@ impl RxStreamOrderer {
             self.offset = new_start;
         }
 
-        self.data.extend(new_data);
-        self.received += u64::try_from(new_data.len()).expect("usize fits in u64");
+        let overlap = self.next_expected() - new_start;
+        self.data.extend(&new_data[overlap..]);
+        self.received += u64::try_from(new_data.len() - overlap).expect("usize fits in u64");
     }
 
     fn store_out_of_order_data(&mut self, mut new_start: u64, mut new_data: &[u8]) {
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -132,22 +132,32 @@ impl RecvStreams {
 
 /// Holds data not yet read by application. Orders and dedupes data ranges
 /// from incoming STREAM frames.
+///
+/// Optimized for mostly in-order data with a hybrid structure:
+/// - Primary buffer for contiguous data at the head
+/// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
+    // Primary contiguous buffer for sequential data
+    primary_buffer: Vec<u8>,
+    primary_start: u64, // Offset where primary_buffer starts
+
+    // Out-of-order chunks with gaps
     data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
-    retired: u64,                        // Number of bytes the application has read
-    received: u64,                       // The number of bytes stored in `data_ranges`
+
+    retired: u64,  // Number of bytes the application has read
+    received: u64, // The number of bytes stored across all buffers
+
+    // Optimization: track the expected next sequential offset
+    next_expected: u64,
 }
 
 impl RxStreamOrderer {
-    #[must_use]
-    pub fn new() -> Self {
-        Self::default()
-    }
-
     /// Process an incoming stream frame off the wire. This may result in data
     /// being available to upper layers if frame is not out of order (ooo) or
     /// if the frame fills a gap.
+    ///
+    /// Optimized for mostly in-order data with fast-path for sequential frames.
     /// # Panics
     /// Only when `u64` values cannot be converted to `usize`, which only
     /// happens on 32-bit machines that hold far too much data at the same time.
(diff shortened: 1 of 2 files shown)


larseggert:
There might be some benefit here, but the benches are again failing with
```
thread 'main' panicked at /home/bench/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/criterion-0.6.0/src/kde.rs:10:66:
index out of bounds: the len is 0 but the index is 0
```

neqo-transport/src/recv_stream.rs
```diff
@@ -132,22 +132,32 @@ impl RecvStreams {
...
+    // Out-of-order chunks with gaps
     data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
-    retired: u64,                        // Number of bytes the application has read
-    received: u64,                       // The number of bytes stored in `data_ranges`
+
+    retired: u64,  // Number of bytes the application has read
+    received: u64, // The number of bytes stored across all buffers
+
+    // Optimization: track the expected next sequential offset
+    next_expected: u64,
```
mxinden: Shouldn't this always be `primary_start + primary_buffer.len()`?

neqo-transport/src/recv_stream.rs
```diff
@@ -132,22 +132,32 @@ impl RecvStreams {
...
 /// Holds data not yet read by application. Orders and dedupes data ranges
 /// from incoming STREAM frames.
+///
+/// Optimized for mostly in-order data with a hybrid structure:
+/// - Primary buffer for contiguous data at the head
+/// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
+    // Primary contiguous buffer for sequential data
+    primary_buffer: Vec<u8>,
```
mxinden: Instead of a `Vec`, why not a ring buffer, i.e. `VecDeque`? Say we transfer 1 GB of data on a stream, but only ever have 10 MB in-flight. The former would allocate 1 GB over its lifetime, the latter only 10 MB.

mxinden (commented):
Idea makes sense to me. What datastructure do other implementations use?

mxinden:
A comparison between this pull request and https://github.com/mozilla/neqo/issues/2664 can be summarized as:

- This pull request does two memory copies, (1) from the UDP datagram into the long lived `primary_buffer` and (2) into the users buffer in `RxStreamOrdered::read`. If `primary_buffer` is a  `VecDequeue`, there is no additional allocation.
- https://github.com/mozilla/neqo/issues/2664 does one memory allocation and one memory copy, allocating and copying each stream frame into a new `Vec`. The user would be provided with that `Vec`. No second memory copy needed.

I don't have an intuition which strategy is faster. Posting this here to document it. I am in favor of this pull request, especially since it is nicely contained within the module and doesn't need API changes like https://github.com/mozilla/neqo/issues/2664 does.

neqo-transport/src/recv_stream.rs
larseggert: @mxinden the `VecDeque` is slower (or I am using it wrong).

larseggert:
> Idea makes sense to me. What datastructure do other implementations use?

quiche seems to use a `BTreeMap` similar to what we originally had: https://github.com/cloudflare/quiche/blob/d97fe02bf9e5972246bccb43e9682223fba640a7/quiche/src/stream/recv_buf.rs#L53

neqo-transport/src/recv_stream.rs
mxinden: I am assuming it is the call to `make_contiguous()` , which transforms a wrapping ring buffer into a contiguous ring buffer, thus loosing all the benefits of a ring buffer. Instead, can you try doing two copies, once with the first slice from `as_slices`, once potentially with the second slice returned by `as_slice`.

neqo-transport/src/recv_stream.rs
```diff
@@ -132,31 +132,47 @@ impl RecvStreams {
...
+///
+/// Optimized for mostly in-order data with a hybrid structure:
+/// - in-order buffer for contiguous data at the head
+/// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
-    data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
-    retired: u64,                        // Number of bytes the application has read
-    received: u64,                       // The number of bytes stored in `data_ranges`
+    // Contiguous buffer for in-order data.
```
martinthomson:
```suggestion
    // Buffer for arriving in-order data.
```

neqo-transport/src/recv_stream.rs
```diff
@@ -132,31 +132,47 @@ impl RecvStreams {
...
+    data: VecDeque<u8>,
+    offset: u64, // Offset where `data` starts.
+
+    // Out-of-order data with gaps
+    ooo: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
```
martinthomson: not a great name either

```suggestion
    out_of_order: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
```

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,87 @@ impl RxStreamOrderer {
             new_start = self.retired;
         }
 
-        if new_data.is_empty() {
-            // No data to insert
+        // Only use fast path for truly in-order data when there are no gaps.
+        if new_start == self.next_expected() && self.ooo.is_empty() {
```
martinthomson:
```suggestion
        if new_start <= self.next_expected() && new_start + new_end > self.next_expected() && self.ooo.is_empty() {
```

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,87 @@ impl RxStreamOrderer {
...
+            "Fast path: appending in-order data {new_start}-{}",
+            new_start + u64::try_from(new_data.len()).expect("usize fits in u64")
+        );
+
+        if self.data.is_empty() {
+            self.offset = new_start;
+        }
+
+        self.data.extend(new_data);
+        self.received += u64::try_from(new_data.len()).expect("usize fits in u64");
```
martinthomson: It might be worth a few extra cycles to support chunks that overlap the end.

```suggestion
        let overlap = self.next_expected() - new_start;
        self.data.extend(&new_data[overlap..]);
        self.received += u64::try_from(new_data.len() - overlap).expect("usize fits in u64");
```

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,87 @@ impl RxStreamOrderer {
             new_start = self.retired;
         }
 
-        if new_data.is_empty() {
-            // No data to insert
+        // Only use fast path for truly in-order data when there are no gaps.
+        if new_start == self.next_expected() && self.ooo.is_empty() {
+            self.store_in_order_data(new_start, new_data);
             return;
```
martinthomson: Maybe use `else` here.

neqo-transport/src/recv_stream.rs
```diff
@@ -320,24 +419,123 @@ impl RxStreamOrderer {
...
     fn read(&mut self, buf: &mut [u8]) -> usize {
         qtrace!("Reading {} bytes, {} available", buf.len(), self.buffered());
         let mut copied = 0;
 
-        for (&range_start, range_data) in &mut self.data_ranges {
+        // First, try to read from in-order buffer
+        if !self.data.is_empty() && self.offset <= self.retired {
+            let data_end = self.offset + u64::try_from(self.data.len()).expect("usize fits in u64");
+            if data_end > self.retired {
+                let copy_offset =
```
martinthomson: Something here needs to be factored out.

neqo-transport/src/recv_stream.rs
```diff
@@ -132,31 +132,47 @@ impl RecvStreams {
...
+/// Optimized for mostly in-order data with a hybrid structure:
+/// - in-order buffer for contiguous data at the head
+/// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
-    data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
-    retired: u64,                        // Number of bytes the application has read
-    received: u64,                       // The number of bytes stored in `data_ranges`
+    // Contiguous buffer for in-order data.
+    data: VecDeque<u8>,
```
martinthomson: not a great name for this

Maybe:

```suggestion
    in_order: VecDeque<u8>,
```

neqo-transport/src/recv_stream.rs
```diff
@@ -251,59 +340,69 @@ impl RxStreamOrderer {
...
-                } else {
-                    false
-                }
-            })
-            // Accumulate, but saturate at usize::MAX.
-            .fold(0, |acc: usize, (_, data_len)| {
-                acc.saturating_add(usize::try_from(data_len).unwrap_or(usize::MAX))
-            })
+
+        // First check in-order buffer
```
martinthomson, line 385: Should this instead turn the in-order buffer and the out-of-order buffers into a single iterator, applying the algorithm we used to have to the outcome?  You can make an iterator by computing:

```rust

   let (a, b) = self.in_order.as_slices();
   for (&start_offset, data) in (!a.is_empty().then_some(|a| (self.offset, a))
        .chain((!b.is_empty().then_some(|b| (self.offset + a.len(), b))
        .chain(self.ooo.iter()) {
        if ...
    }
```

That iterator function can be used to deliver bytes and to count what is available).  It's a tiny bit more work, because you could skip some checks on the in-order chunk that this wouldn't) but it might be far more readable.

larseggert:
Sorry for flipping off "draft", have something else to fix.

neqo-transport/src/recv_stream.rs
```diff
@@ -132,31 +132,47 @@ impl RecvStreams {
...
+/// - in-order buffer for contiguous data at the head
+/// - `BTreeMap` for out-of-order chunks with gaps
 #[derive(Debug, Default)]
 pub struct RxStreamOrderer {
-    data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
-    retired: u64,                        // Number of bytes the application has read
-    received: u64,                       // The number of bytes stored in `data_ranges`
+    // Buffer for arriving in-order data.
+    in_order: VecDeque<u8>,
+    offset: u64, // Offset where `data` starts.
```
martinthomson: Is there ever any case where `self.offset != self.retired` ?  I can't think of why you would hold data *before* `retired` (i.e., `offset < retired`) and a gap doesn't make sense (i.e., `offset > retired`).

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,90 @@ impl RxStreamOrderer {
...
-            // No data to insert
-            return;
+        if new_start <= self.next_expected()
+            && new_start + new_end > self.next_expected()
+            && self.out_of_order.is_empty()
+        {
+            // Only use fast path for truly in-order data when there are no gaps.
+            self.store_in_order_data(new_start, new_data);
+        } else {
+            // Handle out-of-order data.
```
martinthomson, line 202:
```suggestion
```

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,90 @@ impl RxStreamOrderer {
...
+    }
+
+    fn store_out_of_order_data(&mut self, mut new_start: u64, mut new_data: &[u8]) {
+        qtrace!(
+            "Handling out-of-order data {new_start}-{}",
+            new_start + u64::try_from(new_data.len()).expect("usize fits in u64")
+        );
+        let new_end = new_start + u64::try_from(new_data.len()).expect("usize fits in u64");
+
+        // First, check for overlap with in-order buffer.
```
martinthomson, line 226: I would change the checks outside this function so that you don't have to worry about this case.

Suggestion above...

neqo-transport/src/recv_stream.rs
```diff
@@ -170,13 +186,90 @@ impl RxStreamOrderer {
...
 
-        if new_data.is_empty() {
-            // No data to insert
-            return;
+        if new_start <= self.next_expected()
+            && new_start + new_end > self.next_expected()
+            && self.out_of_order.is_empty()
+        {
+            // Only use fast path for truly in-order data when there are no gaps.
+            self.store_in_order_data(new_start, new_data);
```
martinthomson:
```suggestion
        {
            if new_start + new_end > self.next_expected() {
                // Only use fast path for truly in-order data when there are no gaps.
                self.store_in_order_data(new_start, new_data);
            }
```

You could move the inner check to an early return, if you prefer.

Two things:

1. You don't need to check if `out_of_order` is empty.  If the new chunk is contiguous, then it's fine.  However, you should maybe test the two approaches:
    a. Check if `out_of_order` is empty, as you have.  Then there is never a risk that you overlap `in_order` with `out_of_order`.
    b. Trim the new data based on the first byte of `out_of_order`.  That should be as simple as `let len = self.out_of_order.first().map_or_else(|| new_data.len(), |(end_offset, _)| usize::try_from(end_offset - new_start).expect("things"))`.
2. You don't need to store anything if there is something to store.

I would not concern myself if `in_order` and `out_of_order` are contiguous because stuff was stored in `out_of_order` before the gap was closed.  That will sort itself out.

martinthomson (commented):
I didn't look especially closely at the reading and status functions.  My suggestion from before might make the status functions a little simpler.  The reading function won't be able to use that, but I think that the comment about `offset` and `retired` might deal with that.

Overall, this looks like a good direction to take though.  I will ask though: do you think that the benchmarks we have are up to the task of validating the change?  I don't see this move the needle on headline benchmarks, just the microbench.  And those are pretty noisy.

larseggert:
I've noodled on this a bit more, and I'm not sure there is a big win here. While the `inbound_frame` microbenchmark improves, that contains a pretty unrealistic scenario, because things are just added in-order and nothing is ever _read_ from the data structure.

In reality, we'd see a bunch of in-order arrivals of data that go into the `VecDeque`, then a gap (a loss), and everything afterwards goes into the `BTreeMap`. Since we apparently don't read very frequently from the data structure, we're essentially again bound by the `BTreeMap` performance but added a bunch of extra complexity.

If we wanted to avoid that `VecDeque`-then-only-`BTreeMap` pattern, we'd need to append from the `BTreeMap` into the `VecDeque` when gaps fill, but that's an extra memory copy. So no real gain there either.

Issue #2855, opened 2025-08-15
`disable-random` not fully working


larseggert:
The `disable-random` feature was added to allow repeatable fuzzing runs. It doesn't seem to fully take effect however. This becomes pretty easy to spot if one runs
```
cargo nextest run --features build-fuzzing-corpus --no-fail-fast
```
twice in a row. That command, under the `fuzzer/corpus`, builds fuzzing corpora for the various fuzzers, i.e, collects input data for the various fuzzers generated during the run of the test suite. `build-fuzzing-corpus` turns on `disable-random` and `disable-encryption`.

There will be *x* files in the `fuzzer/corpus` directory tree after the first run. After a second run, there will be *y* files, where *x < y*.

The reason for the increase is that the second test run generated data that wasn't generated during the first run, which can really only be due to randomization still being present.

We should fix this, since the fuzzing runs should be repeatable.

One hunch is that either the `disable-random` feature isn't fully transitively applied throughout the various crates on compilation. Another hunch is that there is some inherent remaining randomization in the NSS TLS stack.

PR #2958, opened 2025-09-09
fix: Turn off more TLS randomization with `disable-random`

larseggert:
Might address some of #2855.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -560,11 +560,7 @@ impl SecretAgent {
         self.set_option(ssl::Opt::OcspStapling, true)?;
         self.set_option(
             ssl::Opt::Grease,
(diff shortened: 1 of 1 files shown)


neqo-crypto/src/agent.rs
```diff
@@ -558,8 +558,18 @@ impl SecretAgent {
...
         self.set_option(ssl::Opt::OcspStapling, true)?;
-        self.set_option(ssl::Opt::Grease, grease)?;
-        self.set_option(ssl::Opt::EnableChExtensionPermutation, true)?;
+        self.set_option(
+            ssl::Opt::Grease,
+            if cfg!(not(feature = "disable-random")) {
+                grease
+            } else {
+                false
+            },
```
```suggestion
            cfg!(not(feature = "disable-random")) && grease,
```

martinthomson (approved):

larseggert:
CC @moz-mdauer

PR #2957, opened 2025-09-09
Debug BACKTRACE

mxinden:
Debugging https://github.com/mozilla/neqo/issues/2933.

PR #2965, opened 2025-09-10
feat(.github/bench): enable full Rust backtraces for benches

mxinden:
Follow-up to https://github.com/mozilla/neqo/pull/2957#discussion_r2332695847.


.github/actions/rust/action.yml
```diff
@@ -21,6 +21,9 @@ inputs:
     description: Newline-separated list of workspaces
     required: false
 
+env:
+  RUST_BACKTRACE: full
```
larseggert: I think – but am not sure – that you need to add that to `$GITHUB_ENV`.

mxinden: GitHub does seem to accept it. Is it not carried to the actual action runs this way @larseggert?

larseggert: I don't know, check the env afterwards?

Issue #2967, opened 2025-09-10
CUBIC update to RFC 9438 tracking issue


omansfeld:
As suggested by @larseggert the big PR #2535 is becoming unwieldy so I'm splitting it into multiple smaller ones and tracking them here.

I'm trying to keep the PRs as independent as possible so they can be landed one-by-one. That means that there will likely be some merge conflicts between them though, because they sometimes have to touch the same things -- I will address those as they emerge.

My plan for PRs is as follows:

- [x] #2968 
  - all others will depend on this one
- [x] #2970 
- [x] #2973 
  - #3043 has been split out to further de-risk and land later
- [x] #2983 
  - [x] #2994 
  - [x] #2997  
- #2985 will also be tackled later

I'll update this issue description as we go along.

Update 2025-10-15: Everything merged and leftovers are tracked in #3053. I will close this issue.

larseggert:
Before we start reviewing/merging PRs that change behavior, I'd like to discuss better (CI) testing for CC, and getting _that_ merged first.

This isn't (only) about making sure we have good LoC coverage, but making sure we have test cases that validate correct behavior in CC edge cases. 

We probably need quite a few of these tests, so implementing them all in straight up Rust is going to be very wordy. Wondering if we can extend the simulator so one can feed it a test case with a script (think something like packetdrill) and have it execute that.

omansfeld:
> Before we start reviewing/merging PRs that change behavior, I'd like to discuss better (CI) testing for CC, and getting _that_ merged first.
> 
> This isn't (only) about making sure we have good LoC coverage, but making sure we have test cases that validate correct behavior in CC edge cases.
> 
> We probably need quite a few of these tests, so implementing them all in straight up Rust is going to be very wordy. Wondering if we can extend the simulator so one can feed it a test case with a script (think something like packetdrill) and have it execute that.

I agree it would be good to have confidence in our CI around CC and methods to properly evaluate CC -- for the upcoming work on Slow Start (#2858) , too.

Extending the simulator with e.g. packetdrill does sound like a new project-size effort to me though (packetdrill doesn't support quic for example, so would also need to be extended), so I'm wondering what the best way forward is.

[...]

PR #2970, opened 2025-09-10
feat(transport/cubic): multiplicative decrease updates

omansfeld:
Changes the following:
- new fast convergence algorithm works without an extra variable for saving state (see [RFC](https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence))
  - thus we get rid of `last_max_cwnd`
- add `cwnd_prior` for logging, this will also be used in a later PR for dynamically changing the value of `CUBIC_ALPHA`
- adding comments
- renaming to match RFC 9438

Split off of #2535. 
Part of #2967.
Depends on #2968.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -42,7 +42,7 @@ jobs:
     name: zizmor 🌈
     runs-on: ubuntu-24.04
     permissions:
-      security-events: write
+      security-events: write # Needs to upload SARIF info.
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
--- a/.github/workflows/bencher-upload.yml
+++ b/.github/workflows/bencher-upload.yml
@@ -15,8 +15,8 @@ jobs:
     # The bench workflow fails when a regression is detected, but we want to upload results regardless.
     if: github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion == 'failure'
     permissions:
-      pull-requests: write
-      checks: write
+      pull-requests: write # Writes comments on PRs.
+      checks: write # Writes comments on commits.
     runs-on: ubuntu-24.04
     env:
       BENCHER_PROJECT: ${{ github.event.repository.name }}
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -19,7 +19,7 @@ jobs:
   comment:
     name: Comment on PR
     permissions:
-      pull-requests: write
+      pull-requests: write # Writes comments on PRs.
     runs-on: ubuntu-24.04
     if: |
       github.event.workflow_run.event == 'pull_request' &&
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -31,7 +31,7 @@ jobs:
     outputs:
       imageID: ${{ steps.docker_build_and_push.outputs.imageID }}
     permissions:
-      packages: write
+      packages: write # To push to GitHub Container Registry.
     steps:
       - uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
         with:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -446,7 +446,7 @@ dependencies = [
 
 [[package]]
 name = "fuzz"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "libfuzzer-sys",
  "neqo-common",
@@ -805,7 +805,7 @@ dependencies = [
 
 [[package]]
 name = "neqo-bin"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "clap",
  "clap-verbosity-flag",
@@ -830,7 +830,7 @@ dependencies = [
 
 [[package]]
 name = "neqo-common"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "criterion",
  "enum-map",
@@ -848,7 +848,7 @@ dependencies = [
 
 [[package]]
 name = "neqo-crypto"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "bindgen",
  "enum-map",
@@ -866,7 +866,7 @@ dependencies = [
 
 [[package]]
 name = "neqo-http3"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "criterion",
  "enumset",
@@ -887,7 +887,7 @@ dependencies = [
 
 [[package]]
 name = "neqo-qpack"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "log",
  "neqo-common",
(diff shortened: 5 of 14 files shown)


mxinden:
> Keeping as a draft until #2968 is merged.

With #2968 merged, is this ready for review?

omansfeld:
> > Keeping as a draft until #2968 is merged.
> 
> With #2968 merged, is this ready for review?

It is, pending Lars's comment about general CI coverage

>Before we start reviewing/merging PRs that change behavior, I'd like to discuss better (CI) testing for CC, and getting that merged first.

https://github.com/mozilla/neqo/issues/2967#issuecomment-3275675183

omansfeld:
@mxinden This is just a refactor and the behavior is already tested here:

https://github.com/mozilla/neqo/blob/89cb6ea254d874dcf69498e254e3f06815982cf5/neqo-transport/src/cc/tests/cubic.rs#L310-L335

So this PR should be ready for review/merge right now.

The refactor is mainly that fast convergence has a slightly different algorithm now that uses one less variable to keep state. Functionally it does the same thing.

For reference:
Old RFC: https://datatracker.ietf.org/doc/html/rfc8312#section-4.6
New RFC: https://datatracker.ietf.org/doc/html/rfc9438#name-fast-convergence

neqo-transport/src/cc/cubic.rs
```diff
@@ -94,16 +109,33 @@ pub struct Cubic {
...
     w_max: f64,
-    /// > the elapsed time from the beginning of the current congestion
-    /// > avoidance
+    /// > The time in seconds at which the current congestion avoidance stage started.
     ///
-    /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
-    ca_epoch_start: Option<Instant>,
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#name-variables-of-interest>
+    ///
+    /// This also is reset on being application limited
```
```suggestion
    /// This also is reset on being application limited.
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -145,35 +177,48 @@ impl Cubic {
         (CUBIC_C * (t - self.k).powi(3)).mul_add(convert_to_f64(max_datagram_size), self.w_max)
     }
 
+    /// Resets all relevant parameters at the start of a new epoch (new congestion
+    /// avoidance stage) according to RFC 9438. The `w_max` variable is set in `reduce_cwnd()`. Also
```
mxinden: I don't understand the intention behind this sentence:

> The `w_max` variable is set in `reduce_cwnd()`.

`w_max` is set in both `start_epoch` and `reduce_cwnd`. Why is that relevant here? Would the following as well be correct?

```suggestion
    /// avoidance stage) according to RFC 9438. The `w_max` variable has been set in `reduce_cwnd()` prior to this call. Also
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -145,35 +177,48 @@ impl Cubic {
...
         self.tcp_acked_bytes = new_acked_f64;
         self.estimated_tcp_cwnd = curr_cwnd_f64;
-        if self.last_max_cwnd <= curr_cwnd_f64 {
+        // If `w_max < cwnd_epoch` we take the cubic root from a negative value in `calc_k()`. That
+        // could only happen if somehow `cwnd` get's increased between calling `reduce_cwnd()` and
+        // `start_epoch()`, which is only possible after persistent congestion or an app limited
+        // period. It could also happen if we never had a congestion event, so never called
+        // `reduce_cwnd()` thus `w_max` was never set (so is still it's default `0.0`
+        // value). In any case we reset/initialize `w_max`, `cwnd_prior` and `k` here.
+        self.k = if self.w_max < curr_cwnd_f64 {
```
mxinden: This is a non-refactoring change from `self.cwnd_prior` to `self.w_max`, correct? And I assume this is part of RFC 9438? I think this is fine. Do I understand correctly that `cwnd_prior` is used for logging only now, and will only really be used with https://github.com/mozilla/neqo/pull/2985/?

neqo-transport/src/cc/cubic.rs
omansfeld: It's usually set in `reduce_cwnd` and only set in `start_epoch` in the special case that it is still uninitialized (we have never called `reduced_cwnd`).

If I remember correctly I added that sentence because when first looking at the code I was confused why some parameters are set in `start_epoch` and `w_max` is not, so it's hopefully easier for other maintainers to understand in the future.

I will change the wording of that paragraph to be a bit more explicit.

neqo-transport/src/cc/cubic.rs
omansfeld: I don't quite understand what you mean by "from `self.cwnd_prior` to `self.w_max`".

In this PR we get rid of `last_max_cwnd` and directly use `w_max` instead, as per the RFC section linked up top in the PR description.

Because of fast convergence it can happen that `w_max` is actually smaller than the congestion window before it was reduced. That's why we track the actual old congestion window in `cwnd_prior`. In cases where fast convergence didn't trigger `cwnd_prior` will be equal to `w_max`.

And correct, `cwnd_prior` will only really be used with #2985 and is for now only used in `Display for Cubic`. I'm fine removing it from this PR if you want.

It is usually set in `reduce_cwnd` and here we initialize it for cases where `reduce_cwnd` was never called (which will happen a lot upon exiting slow start once we implement hystart or search).

mxinden: > I don't quite understand what you mean by "from `self.cwnd_prior` to `self.w_max`".

Sorry. I should have been more explicit. The previous implementation used `cwnd_prior` (previously `last_max_cwnd`. The new implementation is using `w_max`. That is fine. Though is not a pure refactoring, but an actual logic change, thus this pull request should be e.g. `feat(transport/cubic): multiplicative decrease updates`. This might be nit-picky. Though in case this patch introduces a regression, we will on first investigation ignore all `refactor` pull requests, assuming they don't change functionality.

> And correct, `cwnd_prior` will only really be used with #2985 and is for now only used in `Display for Cubic`. I'm fine removing it from this PR if you want.

Yes. Let's remove for now.

I think each pull request should only introduce what it needs and, say for the unlikely case where we stop working on Neqo tomorrow, leave the code in a consistent state by itself, not depending on follow-up pull requests yet to be merged.

mxinden: > I think each pull request should only introduce what it needs and, say for the unlikely case where we stop working on Neqo tomorrow, leave the code in a consistent state by itself, not depending on follow-up pull requests yet to be merged.

Good read on this guideline:

https://matklad.github.io/2024/01/03/of-rats-and-ratchets.html

omansfeld: Removed `cwnd_prior` for now.

I see your point r.e. regression investigations ignoring `refactor` PRs. I guess this is a logic change, even if it doesn't change the result.

For the record, removing `last_max_cwnd` in favor of `w_max` is more like this
```rust
// old way
let a = 0;
let b = a + 1;
a = b;

// new way
let a = 0;
a += 1;
```
`last_max_cwnd` was basically just saving the result to the new `w_max`.

That said, I don't mind calling this PR `feature`. Are you fine going to merge with this one?

Issue #2932, opened 2025-09-02
Consider always sending `SETTINGS_H3_DATAGRAM` `true`


mxinden:
Currently by default we set `SETTINGS_H3_DATAGRAM` to `false`:

https://github.com/mozilla/neqo/blob/c638e9768420b3d59fa1be43fa739af72e5ed833/neqo-http3/src/conn_params.rs#L17

RFC 9297 recommends always sending this setting, under the premise that QUIC datagrams are supported I assume:

> Since transmitting HTTP Datagrams using QUIC DATAGRAM frames requires sending the HTTP/3 SETTINGS_H3_DATAGRAM setting, it "sticks out". In other words, probing clients can learn whether a server supports HTTP Datagrams over QUIC DATAGRAM frames. As some servers might wish to obfuscate the fact that they offer application services that use HTTP Datagrams, it's best for all implementations that support this feature to always send this setting; see [Section 2.1.1](https://datatracker.ietf.org/doc/html/rfc9297#setting).

https://datatracker.ietf.org/doc/html/rfc9297#security

This is mostly relevant for servers. That said, we might as well do the same on the client side, reducing our fingerprinting surface.

Thoughts?

martinthomson:
Yes, always is best.  Note that this also makes it easier to find a connection that is datagram-ready from a connection pool.  If we have a connection, then it will have datagram support if the server supports it.

mxinden:
Should Firefox try to negotiate `SETTINGS_H3_DATAGRAM` when QUIC datagrams are disabled?

- As far as I can tell it would be spec compliant.
- Yet I can not think of a scenario where this would be useful.
- It is confusing.

---

`neqo-transport` currently disables QUIC datagrams by default:

https://github.com/mozilla/neqo/blob/6ea322a849fb0e03465e6bbc75363ee035b65538/neqo-transport/src/connection/params.rs#L118

Yet, Firefox enables it by default:

https://github.com/mozilla-firefox/firefox/blob/4e8871487a9f97992e00d5c8105ef1d8cf3c88f3/modules/libpref/init/StaticPrefList.yaml#L15524-L15528

I suggest enabling it in `neqo-transport` as well. Objections?

PR #2972, opened 2025-09-10
feat(transport,http3): enable SETTINGS_H3_DATAGRAM and max_datagram_frame_size

mxinden:
This commit changes the following two defaults:

- Sets max_datagram_frame_size to 1200 by default.
- Sets SETTINGS_H3_DATAGRAM to true by default.

---

See https://github.com/mozilla/neqo/issues/2932 for discussion.

Fixes https://github.com/mozilla/neqo/issues/2932.

### DIFF
--- a/neqo-http3/src/conn_params.rs
+++ b/neqo-http3/src/conn_params.rs
@@ -135,3 +135,20 @@ impl Http3Parameters {
         self.http3_datagram
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


larseggert:
Coverage shows that there is an error condition we don't have a test for.

mxinden:
@larseggert I pushed a commit to improve test coverage. The only uncovered line is in a test itself, not executed, as the previous call panics, as asserted in the test.

Mind merging here manually?

PR #2973, opened 2025-09-10
cubic: reno region updates

omansfeld:
Keeping as a draft until https://github.com/mozilla/neqo/pull/2968 is merged.

Changes the following:

- renaming to match RFC 9438
- new formula for `w_est` (see [RFC](https://datatracker.ietf.org/doc/html/rfc9438#section-4.3-10.1.1))
- removing the tracking of `new_acked_bytes` across acks (see https://github.com/mozilla/neqo/pull/2535#discussion_r2315913966 for discussion)
- comments and docs
- making tests pass (see [`33cc7c8` (#2535)](https://github.com/mozilla/neqo/pull/2535/commits/33cc7c8a2da93113388a0d72298874d7c8d8e522#r2281629718) for discussion)

Split off of https://github.com/mozilla/neqo/pull/2535.
Part of https://github.com/mozilla/neqo/issues/2967.
Depends on https://github.com/mozilla/neqo/pull/2968.

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -165,6 +165,16 @@ impl Cubic {
         }
         qtrace!("[{self}] New epoch");
     }
+
+    #[cfg(test)]
+    pub const fn last_max_cwnd(&self) -> f64 {
+        self.last_max_cwnd
+    }
+
+    #[cfg(test)]
+    pub fn set_last_max_cwnd(&mut self, last_max_cwnd: f64) {
+        self.last_max_cwnd = last_max_cwnd;
+    }
 }
 
 impl WindowAdjustment for Cubic {
@@ -277,15 +287,5 @@ impl WindowAdjustment for Cubic {
         // Reset ca_epoch_start. Let it start again when the congestion controller
         // exits the app-limited period.
         self.ca_epoch_start = None;
-    }
-
-    #[cfg(test)]
-    fn last_max_cwnd(&self) -> f64 {
-        self.last_max_cwnd
-    }
-
-    #[cfg(test)]
-    fn set_last_max_cwnd(&mut self, last_max_cwnd: f64) {
-        self.last_max_cwnd = last_max_cwnd;
     }
 }
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -4,7 +4,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! CUBIC congestion control
+//! CUBIC congestion control (RFC 9438)
 
 use std::{
     fmt::{self, Display},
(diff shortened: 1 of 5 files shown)


omansfeld:
The change to how we track acks is changing behavior (as can be seen in the benchmarks above and in the need to change the assert values in existing tests). I already have a local patch that keeps the old ack tracking but uses the new formula and doesn't change behavior anymore (no change in tests needed). Just need to integrate it with the other changes in this PR (comments, renames, etc) and then this should be ready to be reviewed, too.

We can revisit if we actually want the ack tracking or not later when we have more tests. But then this can land already.

neqo-transport/src/cc/cubic.rs
```diff
@@ -260,14 +277,26 @@ impl WindowAdjustment for Cubic {
...
+
+        // We first calculate the increase in segments and floor it to only include whole segments.
+        let increase = (CUBIC_ALPHA * self.reno_acked_bytes / curr_cwnd_f64).floor();
+        // Only apply the increase if it is at least by one segment.
+        if increase > 0.0 {
+            self.w_est += increase * max_datagram_size_f64;
+            // Because we floored the increase to whole segments we cannot just zero
+            // `reno_acked_bytes` but have to calculate the actual bytes used.
+            let acked_bytes_used = increase * curr_cwnd_f64 / CUBIC_ALPHA;
+            self.reno_acked_bytes -= acked_bytes_used;
```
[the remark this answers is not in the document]
omansfeld: The congestion window can never be `0`.

neqo-transport/src/cc/cubic.rs
```diff
@@ -232,11 +244,11 @@ impl WindowAdjustment for Cubic {
         now: Instant,
     ) -> usize {
         let curr_cwnd_f64 = convert_to_f64(curr_cwnd);
-        let new_acked_f64 = convert_to_f64(new_acked_bytes);
+        let new_acked_bytes_f64 = convert_to_f64(new_acked_bytes);
         let max_datagram_size_f64 = convert_to_f64(max_datagram_size);
```
larseggert: For any of these with names ending in `_f64`, I'd suggest removing that suffix if the original variable is not otherwise used in the function, for clarity.

omansfeld:
> There are a bunch of `f64` values in this code (not necessarily new with this PR, also before) that are inherently integers. For example, everything measured in bytes, etc. Could we re-examine this code to make sure we use integer types wherever possible, rather than `f64`? (At least if we can do so without more conversions.)

I looked at this when first starting the work on cubic. While all the byte values (cwnd, mss, acks etc.) are inherently `usize` they have to be multiplicated/divided with `f64` values (e.g. all the constants that are not whole numbers).

I don't know if there is a smarter rust way to do this, but the only thing I can think of is using a divisor/divident approach instead, like in the `reduce_cwnd` function:

https://github.com/mozilla/neqo/blob/a75f3b58c889917fd28ec4a7d69ead8573eb8567/neqo-transport/src/cc/cubic.rs#L345-L351

I believe doing the conversions is more readable in comparison. That's why I initially abandoned the idea of doing something about it -- happy to hear of another way though if there is one.

Anyways, I will open another issue to track this, because I don't think it has to be part of this specific PR.

Edit for reference: #3022

neqo-transport/src/cc/cubic.rs
```diff
@@ -258,16 +270,28 @@ impl WindowAdjustment for Cubic {
...
-        let tcp_cnt = self.estimated_tcp_cwnd / CUBIC_ALPHA;
-        let incr = (self.tcp_acked_bytes / tcp_cnt).floor();
-        if incr > 0.0 {
-            self.tcp_acked_bytes -= incr * tcp_cnt;
-            self.estimated_tcp_cwnd += incr * max_datagram_size_f64;
+        // > Note that this equation uses segments_acked and cwnd is measured in segments. An
+        // > implementation that measures cwnd in bytes should adjust the equation accordingly
+        // > using the number of acknowledged bytes and the SMSS.
+        //
+        // Formula: w_est += (alpha * bytes_acked / cwnd) * SMSS
```
omansfeld, line 282: See comment on the older PR for an explanation as to why the old formula does the same thing as the new one: https://github.com/mozilla/neqo/pull/2535#discussion_r2279396215

omansfeld:
Thanks for the + @larseggert , I will wait to give @mxinden a chance to review before merging, too.

mxinden (approved):
Thanks for splitting this out! Change looks good to me. Well done!

Can you open a pull request with https://github.com/mozilla/neqo/pull/2973/commits/0364e4c3b1adba1526409def42fd69c299e9175a reverted? Even if you don't find time to work on it, I think it is worth documenting this next step.

omansfeld:
> Thanks for splitting this out! Change looks good to me. Well done!
> 
> Can you open a pull request with [0364e4c](https://github.com/mozilla/neqo/commit/0364e4c3b1adba1526409def42fd69c299e9175a) reverted? Even if you don't find time to work on it, I think it is worth documenting this next step.

Will do once this lands!

PR #2977, opened 2025-09-11
ci: Exclude tests from coverage reporting

### LINKED ISSUE #2976: ci: Configure Codecov to ignore coverage of tests themselves
@larseggert I pushed a commit to improve test coverage. The only uncovered line is in a test itself, not executed, as the previous call panics, as asserted in the test.

Mind merging here manually?

_Originally posted by @mxinden in https://github.com/mozilla/neqo/issues/2972#issuecomment-3279122269_

larseggert:
This might make the Codecov CI check fail less often.

Also fix some other minor Codecov things while I'm here.

Fixes #2976

### DIFF
--- a/neqo-http3/src/frames/connect_udp_frame.rs
+++ b/neqo-http3/src/frames/connect_udp_frame.rs
@@ -23,6 +23,7 @@ impl FrameDecoder<Self> for Frame {
 }
 
 #[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
     use super::*;
     use crate::frames::reader::FrameDecoder;
--- a/neqo-http3/src/settings.rs
+++ b/neqo-http3/src/settings.rs
@@ -329,6 +329,7 @@ impl ZeroRttChecker for HttpZeroRttChecker {
 }
 
 #[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
     use neqo_common::Encoder;
 
--- a/neqo-qpack/src/reader.rs
+++ b/neqo-qpack/src/reader.rs
@@ -335,6 +335,7 @@ pub fn parse_utf8(v: &[u8]) -> Res<&str> {
 }
 
 #[cfg(test)]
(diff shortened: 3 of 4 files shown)


test-fixture/src/sim/http3_connection.rs
```diff
@@ -433,6 +433,7 @@ impl Goal for Responses {
 }
 
 #[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
```
martinthomson, line 436: I see you did this all over.  Is there any chance we could add a new macro thingy for this?

See https://docs.rs/cfg_aliases/latest/cfg_aliases/  (the cost is a build.rs in all crates...)

I don't think that cfg_aliases is quite right, because you have a `cfg_attr` rule here; we might even need a new proc macro, so maybe I'm over-thinking things.  Still, repeated code is a bit annoying.

(On the plus side, if you have a proc macro, you can bundle a suppression for clippy::unwrap_used with it.)

Edit: Or, you could just disable coverage for this code unconditionally...

larseggert: If I understand https://github.com/taiki-e/cargo-llvm-cov#exclude-code-from-coverage correctly, "unconditionally" won't work due to nightly-only.

martinthomson: Yeah, that's unfortunate.  If this were a proc macro, you could manage all that from the macro.

larseggert: I'd really like to avoid the `build.rs` thing, as it makes the builds slower IIRC.

larseggert: @martinthomson do you _really_ think going proc-macro is worth it here?

martinthomson: Not really.  It's just that these test configuration lines are getting repetitive to the point of being a little annoying.

larseggert: Would you approve the PR in that case?

martinthomson (approved):

PR #2979, opened 2025-09-13
test(webtransport): handle session OK and WT dgram in same UDP dgram

mxinden:
Test that if a QUIC DATAGRAM is in the same QUIC packet as the WebTransport session 200 Ok, will it be correctly delivered to the application.

Follow-up to https://github.com/mozilla/neqo/issues/1372

### DIFF
--- a/neqo-http3/tests/webtransport.rs
+++ b/neqo-http3/tests/webtransport.rs
@@ -427,3 +427,68 @@ fn wt_race_condition_server_stream_before_confirmation() {
         assert_eq!(events.next(), None);
     }
 }
+
+#[test]
+fn wt_session_ok_and_wt_datagram_in_same_udp_datagram() {
+    fixture_init();
+    let now = now();
+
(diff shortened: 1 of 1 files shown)


neqo-http3/src/server.rs
```diff
@@ -138,6 +138,7 @@ impl Http3Server {
         max_datagrams: NonZeroUsize,
     ) -> OutputBatch {
         qtrace!("[{self}] Process");
+        self.process_http3(now);
```
mxinden: Force the **server** to flush down the WebTransport Session 200 OK to the QUIC layer. Without it the server will first send the WebTransport datagram and then the WebTransport Session OK.

Will find a fix before merging here.

The proof of the test is still valid, i.e. the client does handle a WebTransport datagram when it is packed along with the Webtransport session 200 OK in the same QUIC packet and UDP datagram.

PR #2983, opened 2025-09-15
feat(transport/cubic): cubic region RFC9438 updates

omansfeld:
- comments and docs
- ~~using max_datagram_size_f64 across functions~~
  - (this was already merged in #2994)
- ~~using `saturating_duration_since` to get time since epoch start~~
  - (this was merged in #2997)
- new `target` formula clamping to min and max value (cwnd<=target<=cwnd*1.5)
- removing `EXPONENTIAL_GROWTH_REDUCTION` since it's use is now covered by the max value mentioned above

Split off of #2535.
Part of #2967.

EDIT 2025-10-13:

I rebased after parts of this have already landed (see above checklist). This PR now contains the last updates to the cubic region and should be ready for review.

In the old code we'd also apply the clamping to `1.5 * curr_cwnd` to the reno region, since it was applied to the return value, not to the cubic region calculation result. Since this clamping is supposed to prevent exponential growth ("less than the increase rate of slow start" - [RFC9438](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10)) it doesn't need to be applied to the reno region, which is growing additively per design.

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -310,16 +310,6 @@ impl WindowAdjustment for Cubic {
             // `reno_acked_bytes` but have to calculate the actual bytes used.
             let acked_bytes_used = increase * curr_cwnd / CUBIC_ALPHA;
             self.reno_acked_bytes -= acked_bytes_used;
-
-            // Enforce assumption that Reno adheres to the bounds in the RFC:
-            // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10>
-            //
-            // On very small `curr_cwnd` values it can happen that an increase by just `1 *
-            // max_datagram_size` would violate the bounds. To avoid asserting in that case the
-            // additional `|| max_datagram_size > curr_cwnd * 0.5` predicate was added.
-            debug_assert!(self.w_est <= curr_cwnd * 1.5 || max_datagram_size > curr_cwnd * 0.5,
-               "expect w_est to increase slower than slow start, curr_cwnd: {curr_cwnd}, w_est: {}, acked_bytes_used: {acked_bytes_used}, max_datagram_size: {max_datagram_size}",
-               self.w_est);
         }
 
         // > When receiving a new ACK in congestion avoidance (where cwnd could be greater than


neqo-transport/src/cc/cubic.rs
```diff
@@ -157,28 +156,34 @@ impl Display for Cubic {
...
     ///
-    /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
-    fn calc_k(&self, curr_cwnd: f64, max_datagram_size: f64) -> f64 {
-        ((self.w_max - curr_cwnd) / CUBIC_C / max_datagram_size).cbrt()
+    /// <https://datatracker.ietf.org/doc/html/rfc9438#figure-2>
+    ///
+    /// Taking into account that neqo is using bytes but the formula assumes segments for both
+    /// `w_max` and `cwnd_epoch` it becomes:
+    ///
+    /// `k = cubic_root((w_max - cwnd_epoch)/SMSS/C)`
```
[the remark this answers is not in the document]
omansfeld: Copilot is hallucinating something here. The implementation in line 174 is exactly the same division order as the original comment.

`w_max - cwnd_epoch` returns a byte value but the formula for `k` needs segments, so we divide that by `max_datagram_size` before dividing by `CUBIC_C` as per the `k` formula from the RFC.

neqo-transport/src/cc/cubic.rs
```diff
@@ -265,12 +270,24 @@ impl WindowAdjustment for Cubic {
...
+        // > congestion window increase rate is non-decreasing and is less than the increase rate of
+        // > slow start.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10>
+        //
+        // In neqo target is in bytes.
         let t = now.saturating_duration_since(t_epoch);
-        let target_cubic = self.w_cubic((t + min_rtt).as_secs_f64(), max_datagram_size);
+        // cwnd <= target <= cwnd * 1.5
+        let mut target = f64::clamp(
```
```suggestion
        let mut target_cubic = f64::clamp(
```

Slight preference for keeping the name. Objections?

neqo-transport/src/cc/cubic.rs
```diff
@@ -294,30 +311,46 @@ impl WindowAdjustment for Cubic {
...
+        // (https://datatracker.ietf.org/doc/html/rfc9438#section-4.4-2.1)
+        //
+        // Since we want to know how much we need to ack to increase by 1 segment we need the
+        // inverse of that, which would be `cwnd/cwnd_increase`.
+        // (E.g. if we'd increase by `1/4 * mss` per cwnd acked then we need to ack `4 * cwnd` to
+        // increase by `1 * mss`)
+        //
+        // The RFC only applies this increase per acked cwnd to the Cubic (concave/convex) region.
+        // We also apply it to the Reno region, as that is what the Linux Kernel CUBIC
+        // implementation does, too.
```
mxinden, line 352: Can you add a reference to the Linux kernel GitHub repository code here?

mxinden:
Once the latest congestion control changes landed in Nightly, we should closely monitor packet loss:

https://glam.telemetry.mozilla.org/firefox/probe/http3_loss_ratio/explore?activeBuckets=%5B%220%22%2C%221%22%2C%222%22%2C%224%22%2C%223%22%2C%225%22%5D&process=parent&timeHorizon=ALL&visiblePercentiles=%5B99.9%2C99%2C95%2C75%2C50%5D

mxinden:
> In the old code we'd also apply the clamping to `1.5 * curr_cwnd` to the reno region, since it was applied to the return value, not to the cubic region calculation result. Since this clamping is supposed to prevent exponential growth ("less than the increase rate of slow start" - [RFC9438](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10)) it doesn't need to be applied to the reno region, which is growing additively per design.

Worth adding a `debug_assert` asserting that the Reno target never grows exponentially?

neqo-transport/src/cc/cubic.rs
omansfeld: I changed it because I was using the variable further below as the general target (no specifically cubic or reno region). No objection to changing it and just doing `let target = target_cubic.max(self.w_est)` further below though, just as we had before.

omansfeld:
> > In the old code we'd also apply the clamping to `1.5 * curr_cwnd` to the reno region, since it was applied to the return value, not to the cubic region calculation result. Since this clamping is supposed to prevent exponential growth ("less than the increase rate of slow start" - [RFC9438](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10)) it doesn't need to be applied to the reno region, which is growing additively per design.
> 
> Worth adding a `debug_assert` asserting that the Reno target never grows exponentially?

I don't think it's necessary, as it really mathematically cannot happen (unless with unrealistically small cwnd values). 

Just as an example, if we had a cwnd of `10 * mss` (which is already really small) we would need to ack `50 * mss` bytes to reach the `curr_cwnd * 1.5` threshold. Which cannot really happen, because the congestion window is only `10 * mss`. That is with the increased `CUBIC_ALPHA = 1.0` which we currently don't even implement. And the magnitude of acked bytes needed for this gets bigger the bigger the cwnd is. Should be something like `curr_cwnd^2 / 2` acked bytes needed to reach the threshold.

That said, I'm not opposed to putting in a `debug_assert` there. You said "worth it" implying there is some downside to having it?

mxinden:
> That said, I'm not opposed to putting in a `debug_assert` there. You said "worth it" implying there is some downside to having it?

Just one more line of code. No other downside.

neqo-transport/src/cc/cubic.rs
```diff
@@ -265,12 +270,24 @@ impl WindowAdjustment for Cubic {
                 .expect("unwrapping `None` value -- it should've been set by `start_epoch`")
         };
 
-        // Cubic concave or convex region
+        // Calculate `target` for the concave or convex region
```
```suggestion
        // Calculate `target_cubic` for the concave or convex region
```

neqo-transport/src/cc/cubic.rs
```diff
@@ -285,6 +302,11 @@ impl WindowAdjustment for Cubic {
 
         // We first calculate the increase in segments and floor it to only include whole segments.
         let increase = (CUBIC_ALPHA * self.reno_acked_bytes / curr_cwnd).floor();
+
+        debug_assert!(increase < curr_cwnd / 2.0,
```
mxinden: Is this correct? You are saying we don't need the above `f64::clamp` for Reno, because Reno would never grow as fast. Though the above `clamp` compares `curr_cwnd` to `target_cubic`, while this `debug_assert` compares `curr_cwnd` to increase, i.e. the delta instead of the target.

Say that Reno increases by `curr_cwnd / 3.0` on each ACK, i.e. by one third, that is a lot, especially with a large `curr_cwnd`.

omansfeld: `target < curr_cwnd * 1.5` means we cannot grow the congestion window by more than 50% of what it is currently. So that should be equivalent to `increase < curr_cwnd / 2.0`. I could also put the assert somewhere else and assert on a target value instead of on the increase and use the `< curr_cwnd * 1.5`. But it should do the same thing.

I agree that `curr_cwnd / 3.0` on each ACK would be a lot. But `curr_cwnd / 2` is what the RFC enforces for the cubic region, so I think it makes sense to assert on the same max growth for reno.

In practice with the math outlined in my comment above any reno growth even near `curr_cwnd / 3.0` is unrealistic. Especially with a large `curr_cwnd`.

> Just as an example, if we had a cwnd of 10 * mss (which is already really small) we would need to ack 50 * mss bytes to reach the curr_cwnd * 1.5 threshold. Which cannot really happen, because the congestion window is only 10 * mss. That is with the increased CUBIC_ALPHA = 1.0 which we currently don't even implement. And the magnitude of acked bytes needed for this gets bigger the bigger the cwnd is. Should be something like curr_cwnd^2 / 2 acked bytes needed to reach the threshold.

omansfeld: Did find something else wrong with the assert though: `increase` is in segments while `curr_cwnd` is in bytes.  While at it I moved it to check on `w_est` instead using the same `<= curr_cwnd * 1.5` logic as the cubic region clamp is using. That should make it easier to understand.

neqo-transport/src/cc/cubic.rs
```diff
@@ -294,30 +312,54 @@ impl WindowAdjustment for Cubic {
             self.reno_acked_bytes -= acked_bytes_used;
         }
 
-        // Take the larger cwnd of Cubic concave or convex and Cubic
-        // TCP-friendly region.
+        debug_assert!(self.w_est <= curr_cwnd * 1.5,
+           "detected exponential increase of w_est - curr_cwnd: {curr_cwnd}, w_est: {}, reno_acked_bytes: {}",
```
mxinden: I don't think the usage of `exponential` is mathematically sound here. E.g. why is `1.4` non exponential and `1.6` is? How about:

```suggestion
          // Enforce assumption that Reno adheres to <https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-10>.
           "expect w_est to increase slower than slow start, curr_cwnd: {curr_cwnd}, w_est: {}, reno_acked_bytes: {}",
```

omansfeld: Agreed, I like your suggestion.

It seems like we are actually hitting that debug assert with very low cwnd values in some bench/simulator tests. I will circle back tomorrow.

omansfeld: Some of the assert messages from failing tests ([link to this particular one](https://github.com/omansfeld/neqo/actions/runs/18491735945/job/52686808353)):
`expect w_est to increase slower than slow start, curr_cwnd: 8094, w_est: 16237, acked_bytes_used: 15288.666666666662, max_datagram_size: 8143`

`expect w_est to increase slower than slow start, curr_cwnd: 27280, w_est: 43615, acked_bytes_used: 51528.888888888876, max_datagram_size: 16335`

As can be seen this only fails when we have a `max_datagram_size` that is bigger than `curr_cwnd * 0.5`, i.e. we have a congestion window that is so small that even increasing it by just one segment makes it an increase that is `> curr_cwnd * 1.5`.

The reason this even shows up is our byte tracking: We don't increase `w_est` every ACK by a bit, but we gather bytes across multiple ACKs until we have enough to increase by `1*mss` (see the `acked_bytes_used` values above that are much bigger than the `curr_cwnd`). We have #3043 logged for this.

To move forward I suggest asserting on `self.w_est <= curr_cwnd * 1.5 || max_datagram_size > curr_cwnd * 0.5`. This makes sure we don't assert when we have a congestion window so small that a `1*mss` increase would violate `w_est <= curr_cwnd * 1.5`.

EDIT: corrected the new predicate to use `||` and `>`.

PR #2984, opened 2025-09-15
feat(transport/server): add process_multiple_input 

mxinden:
Given a connection between client and server, take the following scenario:

1. Client opens two streams to server.
2. Server closes the first and sends data on the second.

Previously this would result in the HTTP/3 server to send two UDP datagrams, one
with the close frame for the first stream, one with the data frame for the
second stream. A close is forwarded right away from the Neqo HTTP/3 layer to the
Neqo QUIC layer. Data is buffered at the HTTP/3 layer.

The HTTP/3 server first processes any outbound data on the QUIC layer, and only
then, if nothing is to be send, does it flush down any pending data from the
HTTP/3 layer to the QUIC layer.

https://github.com/mozilla/neqo/blob/c5d1fc5a4f4ba28b043c811eef50174ee26a63cd/neqo-http3/src/server.rs#L134-L153

This commit introduces a `process_multiple_input` function on the QUIC server
layer. The HTTP/3 server layer can now first pass any incoming datagrams down to
the QUIC layer, then flush any outbound data from the HTTP/3 layer to the QUIC
layer, then produce outbound UDP datagrams at the UDP layer. That way, UDP
datagrams are more efficiently packed. E.g. in the above scenario, both the
close frame and the data frame end up in the same UDP datagram.

Relevant for https://github.com/mozilla/neqo/pull/2979.

### DIFF
--- a/neqo-http3/src/server.rs
+++ b/neqo-http3/src/server.rs
@@ -138,7 +138,7 @@ impl Http3Server {
         max_datagrams: NonZeroUsize,
     ) -> OutputBatch {
         qtrace!("[{self}] Process");
-        let out = self.server.process_multiple(dgrams, now, max_datagrams);
+        let out = self.server.process_multiple_input(dgrams, now);
         self.process_http3(now);
         // If we do not that a dgram already try again after process_http3.
         match out {
@@ -352,7 +352,7 @@ mod tests {
     };
 
     use neqo_common::{event::Provider as _, Encoder};
-    use neqo_crypto::{AuthenticationStatus, ZeroRttCheckResult, ZeroRttChecker};
+    use neqo_crypto::{AuthenticationStatus, ResumptionToken, ZeroRttCheckResult, ZeroRttChecker};
     use neqo_qpack as qpack;
     use neqo_transport::{
         CloseReason, Connection, ConnectionEvent, State, StreamId, StreamType, ZeroRttState,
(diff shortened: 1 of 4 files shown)


neqo-http3/src/server.rs
```diff
@@ -138,7 +138,7 @@ impl Http3Server {
         max_datagrams: NonZeroUsize,
     ) -> OutputBatch {
         qtrace!("[{self}] Process");
-        let out = self.server.process_multiple(dgrams, now, max_datagrams);
+        let out = self.server.process_multiple_input(dgrams, now);
```
mxinden, line 141: This is the core of this pull request.

Don't generate an output datagram on the QUIC layer early on (i.e. `process_multiple`). Instead, first handle any input datagrams (`process_multiple_input`), then flush the HTTP/3 layer down to the QUIC layer (`self.process_http3`), then produce any output datagrams (`self.server.process_multiple`).

neqo-http3/tests/classic_connect.rs
```diff
@@ -16,8 +16,7 @@ fn classic_connect() {
     let mut client = default_http3_client();
     let mut server = default_http3_server();
     let out = test_fixture::connect_peers(&mut client, &mut server);
-    let out = server.process(out, now()).dgram().unwrap();
-    client.process_input(out, now());
+    assert_eq!(server.process(out, now()).dgram(), None);
```
mxinden, line 19: Server is now packing everything in the first UDP datagram.

PR #2985, opened 2025-09-15
cubic: dynamic changing of alpha value

omansfeld:
Keeping as draft until #2973 is merged.

- added `self.alpha` field
- added `self.cwnd_prior` field
- existing functions now use `self.alpha` instead of constant value
- `self.alpha` changes and resets as per RFC
- added test helpers
- added test for alpha
- made tests pass (see [this comment on original PR](https://github.com/mozilla/neqo/pull/2535/files#r2279439160) for context)

Could also be made independent of #2973 if needed, but would require some rewriting on the test.

Split off of #2535.
Part of #2967.
Depends on #2973.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,8 @@
 
 version: 2
 updates:
+  # We don't really use Dependabot for Rust dependencies, since we are tracking Gecko,
+  # but we do want to be notified of security vulnerabilities.
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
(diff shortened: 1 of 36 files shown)


larseggert:
#2973 is in, will this progress now?

omansfeld:
> #2973 is in, will this progress now?

In #2973 we split off #3043 which was the part the test here depends on.

Because we don't increase our `w_est` value every ACK but rather wait until we have accumulated enough bytes across multiple ACKs to increase by at least `1 * mss` it is very hard to assert correct growth of `w_est` in the test for this change here.

Apart from that I believe we also decided that this is one of the CUBIC parts that we will rather defer to later, similar to #3043. It would make sense to rebase all the changes in here so it is in a similar state as #3043, ready to be picked up again when time comes.

I will update the description and rebase once all CUBIC changes that can be merged now are in.

omansfeld:
> In https://github.com/mozilla/neqo/pull/2973 we split off https://github.com/mozilla/neqo/pull/3043 which was the part the test here depends on.

I added our byte-tracking logic to the test and thus made it not depend on #3043 anymore. This can easily be removed again if at some point they are both supposed to land.

The description has been updated and the PR is theoretically now in a review-ready state and ready to be picked up again later when we want to get back to the task of de-risking landing it.

larseggert:
@omansfeld do you intend to pursue this PR?

PR #2987, opened 2025-09-16
ci: Upload multiple bencher files in one go

larseggert:
Supported as of 0.5.5, see https://github.com/bencherdev/bencher/issues/618#issuecomment-3289733885


larseggert:
Needs to be on `main` to debug. Merging.

PR #2988, opened 2025-09-16
feat: Don't re-allocate inbound datagram to remove varint

### LINKED ISSUE #2854: perf(http3): don't re-allocate inbound datagram to remove varint #2283
When receiving a WebTransport datagram (and soon MASQUE connect-udp datagram https://github.com/mozilla/neqo/pull/2796), we decode its session ID from the front, then re-allocate the remaining datagram in a new `Vec` (see `to_vec`):

https://github.com/mozilla/neqo/blob/a4eb5e53f69fad22edf25cd0087e434c13d1e21e/neqo-http3/src/connection.rs#L649-L658

Instead we could:

- convert the `Vec` to a `VecDeque` ([O(1)](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#impl-From%3CVec%3CT,+A%3E%3E-for-VecDeque%3CT,+A%3E)), then remove the session ID from the front
- `Vec::rotate_left` then `Vec::truncate` (O(n))
- use [`Bytes`](https://docs.rs/bytes/)
- ...

---

See also https://github.com/mozilla/neqo/issues/2283 for similar allocation for outbound datagrams.

larseggert:
Fixes #2854

@mxinden, is this what you had in mind with option 1?

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -240,11 +240,11 @@ impl ExtendedConnectEvents for Http3ClientEvents {
         Ok(())
     }
 
-    fn new_datagram(&self, session_id: StreamId, datagram: Vec<u8>) {
+    fn new_datagram(&self, session_id: StreamId, datagram: VecDeque<u8>) {
         self.insert(Http3ClientEvent::WebTransport(
             WebTransportEvent::Datagram {
                 session_id,
-                datagram,
+                datagram: datagram.into(),
             },
         ));
     }
--- a/neqo-http3/src/features/extended_connect/webtransport_session.rs
+++ b/neqo-http3/src/features/extended_connect/webtransport_session.rs
@@ -6,6 +6,7 @@
 
 use std::{
     cell::RefCell,
+    collections::VecDeque,
     fmt::{self, Display, Formatter},
     mem,
     rc::Rc,
@@ -477,7 +478,7 @@ impl Session {
         Ok(())
     }
 
-    pub fn datagram(&self, datagram: Vec<u8>) {
+    pub fn datagram(&self, datagram: VecDeque<u8>) {
         if self.state == State::Active {
             self.events.new_datagram(self.id, datagram);
         }
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -6,6 +6,7 @@
 
 use std::{
     cell::RefCell,
+    collections::VecDeque,
     fmt::{self, Debug, Display, Formatter},
     mem,
     rc::Rc,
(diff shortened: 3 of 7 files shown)


neqo-http3/src/client_events.rs
```diff
@@ -240,11 +240,11 @@ impl ExtendedConnectEvents for Http3ClientEvents {
         Ok(())
     }
 
-    fn new_datagram(&self, session_id: StreamId, datagram: Vec<u8>) {
+    fn new_datagram(&self, session_id: StreamId, datagram: VecDeque<u8>) {
         self.insert(Http3ClientEvent::WebTransport(
             WebTransportEvent::Datagram {
                 session_id,
-                datagram,
+                datagram: datagram.into(),
```
mxinden, line 247: This would re-allocate the `VecDeque` as a `Vec`, correct? Thus the performance hit just moves into `client_events.rs`?

I guess we could emit a `VecDeque` in the event itself, i.e. `WebTransportEvent::Datagram`'s `datagram`.

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -477,7 +478,7 @@ impl Session {
         Ok(())
     }
 
-    pub fn datagram(&self, datagram: Vec<u8>) {
+    pub fn datagram(&self, datagram: VecDeque<u8>) {
```
mxinden, line 481: https://github.com/mozilla/neqo/pull/2796/ will improve some of this by passing a `datagram: &[u8]` instead of a `datagram: Vec<u8>`, thus not needing a `Vec` allocation on the caller side.

See `session.rs` here:

https://github.com/mozilla/neqo/pull/2796/files#diff-01aa8f1aba4108ba00b94f619d3ee9de48583de59a33f05e526e480b1e8acf59R388-R402

mxinden (commented):
Do folks have opinions on exposing `VecDeque` vs [`Bytes`](https://docs.rs/bytes/latest/bytes/) in our public interface?

`Bytes` is already in Firefox's `Cargo.lock`. Given that we here don't make use of `VecDeque`'s ring buffer, but simply need to remove bytes off of the head, `Bytes` seems better suited.

neqo-http3/src/client_events.rs
larseggert: Oh, does it? I thought it was a O(1) conversion. If it allocates, that kills the approach.

larseggert:
> Do folks have opinions on exposing `VecDeque` vs [`Bytes`](https://docs.rs/bytes/latest/bytes/) in our public interface?

Not really. If it's part of `tokio` and already in Gecko, I assume it's relatively common and hopefully stable?

larseggert:
Closing here. Someone should look into `Bytes`.

Issue #1894, opened 2024-05-08
Put `qlog` behind a feature


larseggert:
And disable it by default for `release` builds. As discussed on the call today.

PR #2989, opened 2025-09-16
fix: Make `PublicPacket::decrypt` take non-`mut` `self`

### LINKED ISSUE #2397: Make `PublicPacket::decrypt` take `&self`
> I'm looking at the calling code and I *think* that we can change this to take `self` instead of `&mut self`.  There is some use of the public packet after decryption, but it seems like the necessary pieces can be moved to the decrypted packet.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2385#discussion_r1932949860_

larseggert:
Not sure if the complexity if worth it?

Fixes #2397


martinthomson (commented):
It seems like the real change here is the decryption error type.  Which means that we do lots more work if decryption fails.  Why is that helpful?  I usually find that when crypto fails, you have no hope at all.

larseggert:
They main reasons the error case is complex is needing to check for SRT and to QLOG the thing.

neqo-transport/src/packet/mod.rs
```diff
@@ -917,14 +928,58 @@ impl fmt::Debug for Public<'_> {
...
+    pub packet_type: Type,
+}
+
+impl From<(&Public<'_>, Error)> for DecryptionError {
+    fn from((packet, error): (&Public, Error)) -> Self {
+        Self {
+            error,
+            data: packet.data.to_vec(),
+            dcid: packet.dcid.clone(),
+            packet_type: packet.packet_type,
```
martinthomson, line 985: It would be nice if this code did nothing if logging was not enabled.

larseggert: We can make that so once we have fixed #1894. Let me take a stab at that.

neqo-transport/src/packet/mod.rs
```diff
@@ -883,29 +881,42 @@ impl<'a> Public<'a> {
             // This is OK in this case because we the only reason this can
             // fail is if the cryptographic module is bad or the packet is
             // too small (which is public information).
-            let (key_phase, pn, header) = self.decrypt_header(rx)?;
+            let (key_phase, pn, header) = self
+                .decrypt_header(rx)
+                .map_err(|e| DecryptionError::from((&self, e)))?;
```
martinthomson: Couldn't decrypt_header() return `Result<_, DecryptionError>` as well?

neqo-transport/src/packet/mod.rs
```diff
@@ -883,29 +881,42 @@ impl<'a> Public<'a> {
...
+                return Err(DecryptionError::from((&self, Error::Decrypt)));
             };
             let version = rx.version(); // Version fixup; see above.
-            let d = rx.decrypt(pn, header, self.data)?;
+            let d = match rx.decrypt(pn, header, self.data) {
+                Ok(data) => data,
+                Err(e) => {
+                    return Err(DecryptionError::from((&self, e)));
+                }
+            };
```
martinthomson: Another case where returning `DecryptionError` might improve ergonomics.

neqo-transport/src/packet/mod.rs
```diff
@@ -883,29 +881,42 @@ impl<'a> Public<'a> {
...
             }
-            crypto.check_pn_overlap()?;
             Ok(Decrypted {
                 version,
                 pt: self.packet_type,
                 pn,
-                data: d,
+                dcid: self.dcid,
+                scid: self.scid,
+                data: d.to_vec(),
```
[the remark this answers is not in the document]
martinthomson: I like this suggestion.  A good version would have `rx.decrypt()` operate in-place, mutating the buffer that `self` holds.  Then you can truncate that buffer to the requisite length and return it without an allocation.

neqo-transport/src/packet/mod.rs
```diff
@@ -939,14 +950,58 @@ impl fmt::Debug for Public<'_> {
...
+        Self {
+            error,
+            data: packet.data.to_vec(),
+            dcid: packet.dcid.clone(),
+            packet_type: packet.packet_type,
+        }
+    }
+}
+
+impl DecryptionError {
```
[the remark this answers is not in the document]
martinthomson: I don't think that this suggestion is compatible with the way that the code is structured.  The function that returns `DecryptionError` takes ownership of the buffer through the `self` argument.

It's OK to pass ownership to the `DecryptionError`.

neqo-transport/src/packet/mod.rs
```diff
@@ -939,14 +950,58 @@ impl fmt::Debug for Public<'_> {
...
+    #[must_use]
+    #[allow(
+        clippy::allow_attributes,
+        clippy::len_without_is_empty,
+        reason = "OK here."
```
martinthomson: If Copilot is correct, the message here should be "length cannot be 0 by construction" or similar.

martinthomson (changes requested):
The superficial change to the decryption error thing might need to be plumbed a bit deeper.

neqo-transport/src/packet/mod.rs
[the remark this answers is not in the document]
larseggert: I looked into this. I think it's simpler, after merging #3139, to have `Decrypted` hold a `&[u8]` rather than a `Vec`.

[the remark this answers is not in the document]
larseggert: Changed in the latest version

neqo-transport/src/packet/mod.rs
larseggert: I just looked at this again, and I think I lost context. What was your concern here @martinthomson?

neqo-transport/src/packet/mod.rs
```diff
@@ -885,31 +859,42 @@ impl<'a> Public<'a> {
...
             // using `rx`, make sure to initiate a key update.
             if rx.needs_update() {
-                crypto.key_update_received(release_at)?;
+                if let Err(e) = crypto.key_update_received(release_at) {
+                    return Err(DecryptionError::from((&self, e)));
+                }
+            }
+            if let Err(e) = crypto.check_pn_overlap() {
+                return Err(DecryptionError::from((&self, e)));
             }
```
martinthomson: This would be easier to follow as:


```suggestion
            crypto.check_pn_overlap().map_err(|e| DecryptionError::from((&self, e)))?;
```

neqo-transport/src/packet/mod.rs
martinthomson: So, with #1894, we won't do anything with logs, but this error will be constructed when an error occurs.  That construction involves a bunch of work that will be thrown out.

It's only error cases, so we probably won't be wasting too much effort overall, so don't contort the code too badly if it doesn't work out, but it seems like you could add an Option somewhere.

neqo-transport/src/packet/mod.rs
```diff
@@ -885,31 +859,42 @@ impl<'a> Public<'a> {
...
+                    return Err(DecryptionError::from((&self, e)));
+                }
+            };
             // If this is the first packet ever successfully decrypted
             // using `rx`, make sure to initiate a key update.
             if rx.needs_update() {
-                crypto.key_update_received(release_at)?;
+                if let Err(e) = crypto.key_update_received(release_at) {
+                    return Err(DecryptionError::from((&self, e)));
+                }
```
martinthomson: Like below, `map_err()` might make this easier to follow.  The error handling is pretty noisy without using `?`

neqo-transport/src/packet/mod.rs
```diff
@@ -885,31 +859,42 @@ impl<'a> Public<'a> {
...
+                .map_err(|e| DecryptionError::from((&self, e)))?;
             let Some(rx) = crypto.rx(version, epoch, key_phase) else {
-                return Err(Error::Decrypt);
+                return Err(DecryptionError::from((&self, Error::Decrypt)));
             };
             let version = rx.version(); // Version fixup; see above.
-            let header_end = header.end;
-            let payload_len = rx.decrypt(pn, header, self.data)?;
-            let data = &self.data[header_end..header_end + payload_len];
+            let d = match rx.decrypt(pn, header, self.data) {
```
martinthomson: Same here.

neqo-transport/src/packet/mod.rs
```diff
@@ -885,31 +859,42 @@ impl<'a> Public<'a> {
             // This is OK in this case because we the only reason this can
             // fail is if the cryptographic module is bad or the packet is
             // too small (which is public information).
-            let (key_phase, pn, header) = self.decrypt_header(rx)?;
+            let (key_phase, pn, header) = self
+                .decrypt_header(rx)
+                .map_err(|e| DecryptionError::from((&self, e)))?;
             let Some(rx) = crypto.rx(version, epoch, key_phase) else {
```
martinthomson: This one is good for `ok_or()`.

neqo-transport/src/packet/mod.rs
larseggert: I think I found a way to not copy at least the data in the error case.

neqo-transport/src/packet/mod.rs
```diff
@@ -824,7 +825,7 @@ impl<'a> Public<'a> {
             .data
             .get(sample_offset..(sample_offset + SAMPLE_SIZE))
             .ok_or(Error::NoMoreData)?;
```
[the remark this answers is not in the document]
martinthomson: This might be a reasonable case for `.expect()` instead.

neqo-transport/src/packet/mod.rs
```diff
@@ -824,7 +825,7 @@ impl<'a> Public<'a> {
...
-        let sample: &[u8; SAMPLE_SIZE] = sample.try_into()?;
+        let sample: &[u8; SAMPLE_SIZE] = sample.try_into().map_err(|_| Error::NoMoreData)?;
```
martinthomson:
```suggestion
        let sample: &[u8; SAMPLE_SIZE] = sample.try_into().expect("fixed size");
```

larseggert: This gets us a clippy warning, but I can just do `?`.

PR #2990, opened 2025-09-16
perf(codec): optimize encode functions for modern CPUs

larseggert:
(This is me vibing with Claude again.)

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -110,39 +110,15 @@ impl<'a> Decoder<'a> {
         if self.remaining() < n {
             return None;
         }
-        let data = &self.buf[self.offset..self.offset + n];
-        self.offset += n;
-
-        Some(match n {
-            1 => u64::from(data[0]),
-            2 => u64::from(u16::from_be_bytes(
-                data.try_into().expect("slice is exactly 2 bytes"),
-            )),
-            3 => {
-                let mut buf = [0; 4];
-                buf[1..].copy_from_slice(data);
-                u64::from(u32::from_be_bytes(buf))
-            }
-            4 => u64::from(u32::from_be_bytes(
-                data.try_into().expect("slice is exactly 4 bytes"),
-            )),
-            5 => {
-                let mut buf = [0; 8];
-                buf[3..].copy_from_slice(data);
-                u64::from_be_bytes(buf)
-            }
-            6 => {
-                let mut buf = [0; 8];
-                buf[2..].copy_from_slice(data);
-                u64::from_be_bytes(buf)
-            }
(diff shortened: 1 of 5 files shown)


larseggert:
So Claude hallucinated checking the assembly with godbolt.org :-) Second attempt with more explicit instructions.

larseggert:
Is this worth it?

neqo-common/src/codec.rs
```diff
@@ -272,30 +272,33 @@ impl<B: Buffer> Encoder<B> {
     /// # Panics
     ///
     /// When `n` is outside the range `1..=8`.
+    #[inline]
```
mxinden: Unless we have solid  proof, my rule of thumb for `#[inline]` is, that the compiler knows better.

neqo-common/src/codec.rs
```diff
@@ -272,30 +272,33 @@ impl<B: Buffer> Encoder<B> {
...
         let v = v.into();
         assert!(n > 0 && n <= 8);
-        for i in 0..n {
-            self.encode_byte(((v >> (8 * (n - i - 1))) & 0xff) as u8);
-        }
-        self
+        // Using to_be_bytes() generates better assembly, especially for odd sizes
+        // where it uses rev + strh instead of multiple strb instructions
+        let bytes = v.to_be_bytes();
+        self.encode(&bytes[8 - n..])
```
mxinden, line 280: If not for the sake of performance, I think we should do this for the sake of simplicity.

larseggert:
> prepend a commit

Sorry, not sure what you mean here?

mxinden:
Sorry for the brevity. Let me expand.

- `cargo mutant` is able to alter the code in significant ways without test failures.
- Thus I assume this code isn't well tested.
- Given that this pull request does substantial changes to the logic, it might introduce a bug, e.g. in the edge cases.
- Thus I suggest first writing a commit increasing test coverage, making sure it succeeds on `main`.
- Then running the new tests on this patch to make sure it still succeeds.

larseggert:
So I think the `mutants` check here is meaningless, because for the code in question
```rust
            () if v < (1 << 6) => self.encode_uint(1, v),
            () if v < (1 << 14) => self.encode_uint(2, v | (1 << 14)),
            () if v < (1 << 30) => self.encode_uint(4, v | (2 << 30)),
            () if v < (1 << 62) => self.encode_uint(8, v | (3 << 62)),
```
or (`|`) and xor (`^`) produce the same result. We're taking a value smaller than some boundary, and or-ing or xor-ing it with a value larger than the boundary. Same result.

martinthomson (approved):

PR #2993, opened 2025-09-17
ci: Ignore `bencher run` exit code

larseggert:
Bencher exits with error if it detects an alert, ignore that so the workflow doesn't fail.


larseggert:
Needs to be on `main`.

PR #2994, opened 2025-09-17
refactor(transport/cubic): only convert `max_datagram_size_f64` once

omansfeld:
As discussed with @mxinden I'm splitting out some even smaller PRs from #2967 that are purely refactors.

This one calls `convert_to_f64(max_datagram_size)` once and then uses that value in all other function calls and throughout `bytes_for_cwnd_increase` instead of converting on each use.

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -133,23 +133,23 @@ impl Cubic {
     /// K = cubic_root((W_max - W_cubic) / C / MSS);
     ///
     /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
-    fn calc_k(&self, curr_cwnd: f64, max_datagram_size: usize) -> f64 {
-        ((self.w_max - curr_cwnd) / CUBIC_C / convert_to_f64(max_datagram_size)).cbrt()
+    fn calc_k(&self, curr_cwnd_f64: f64, max_datagram_size_f64: f64) -> f64 {
+        ((self.w_max - curr_cwnd_f64) / CUBIC_C / max_datagram_size_f64).cbrt()
     }
 
     /// W_cubic(t) = C*(t-K)^3 + W_max (Eq. 1)
     /// t is relative to the start of the congestion avoidance phase and it is in seconds.
     ///
     /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
-    fn w_cubic(&self, t: f64, max_datagram_size: usize) -> f64 {
-        (CUBIC_C * (t - self.k).powi(3)).mul_add(convert_to_f64(max_datagram_size), self.w_max)
+    fn w_cubic(&self, t: f64, max_datagram_size_f64: f64) -> f64 {
+        (CUBIC_C * (t - self.k).powi(3)).mul_add(max_datagram_size_f64, self.w_max)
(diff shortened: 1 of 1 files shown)


neqo-transport/src/cc/cubic.rs
```diff
@@ -133,23 +133,23 @@ impl Cubic {
     /// K = cubic_root((W_max - W_cubic) / C / MSS);
     ///
     /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
-    fn calc_k(&self, curr_cwnd: f64, max_datagram_size: usize) -> f64 {
-        ((self.w_max - curr_cwnd) / CUBIC_C / convert_to_f64(max_datagram_size)).cbrt()
+    fn calc_k(&self, curr_cwnd_f64: f64, max_datagram_size_f64: f64) -> f64 {
```
```suggestion
    fn calc_k(&self, curr_cwnd: f64, max_datagram_size: f64) -> f64 {
```

Given that there is no non-f64 in scope here, I don't think the suffix is necessary.

neqo-transport/src/cc/cubic.rs
```diff
@@ -133,23 +133,23 @@ impl Cubic {
...
+    fn w_cubic(&self, t: f64, max_datagram_size_f64: f64) -> f64 {
+        (CUBIC_C * (t - self.k).powi(3)).mul_add(max_datagram_size_f64, self.w_max)
     }
 
     fn start_epoch(
         &mut self,
         curr_cwnd_f64: f64,
         new_acked_f64: f64,
-        max_datagram_size: usize,
+        max_datagram_size_f64: f64,
```
mxinden: Same here

neqo-transport/src/cc/cubic.rs
```diff
@@ -133,23 +133,23 @@ impl Cubic {
...
+        ((self.w_max - curr_cwnd_f64) / CUBIC_C / max_datagram_size_f64).cbrt()
     }
 
     /// W_cubic(t) = C*(t-K)^3 + W_max (Eq. 1)
     /// t is relative to the start of the congestion avoidance phase and it is in seconds.
     ///
     /// <https://www.rfc-editor.org/rfc/rfc8312#section-4.1>
-    fn w_cubic(&self, t: f64, max_datagram_size: usize) -> f64 {
-        (CUBIC_C * (t - self.k).powi(3)).mul_add(convert_to_f64(max_datagram_size), self.w_max)
+    fn w_cubic(&self, t: f64, max_datagram_size_f64: f64) -> f64 {
```
mxinden: Same

omansfeld:
Quick general question about our github workflow: Should I add patches to the merge queue myself once they are approved and comments are addressed (like here now), or is that done by the reviewer?

(also just noting, there might be merge conflicts with #2970 so they probably shouldn't both be added to the queue at the same time)

larseggert:
> Should I add patches to the merge queue myself once they are approved and comments are addressed (like here now), or is that done by the reviewer?

We don't really have a fixed policy. Sometimes the reviewer will, but esp. when I think that some nits may still be fixed I will just approve and leave it for the author to enqueue.

larseggert:
> here might be merge conflicts with #2970 so they probably shouldn't both be added to the queue at the same time

Well, whichever is second will need a rebase :-)

PR #2997, opened 2025-09-17
refactor(transport/cubic): use `saturating_duration_since` for t

omansfeld:
- renames `ca_epoch_start` to `t_epoch`
- use `saturating_duration_since` instead of manually checking for older
  timings
- also refactors how we unwrap `self.t_epoch`

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -103,7 +103,7 @@ pub struct Cubic {
     /// > avoidance
     ///
     /// <https://datatracker.ietf.org/doc/html/rfc8312#section-4.1>
-    ca_epoch_start: Option<Instant>,
+    t_epoch: Option<Instant>,
     /// Number of bytes acked since the last Standard TCP congestion window increase.
     tcp_acked_bytes: f64,
 }
@@ -112,8 +112,8 @@ impl Display for Cubic {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "Cubic [last_max_cwnd: {}, k: {}, w_max: {}, ca_epoch_start: {:?}]",
-            self.last_max_cwnd, self.k, self.w_max, self.ca_epoch_start
+            "Cubic [last_max_cwnd: {}, k: {}, w_max: {}, t_epoch: {:?}]",
+            self.last_max_cwnd, self.k, self.w_max, self.t_epoch
         )?;
         Ok(())
     }
@@ -152,7 +152,7 @@ impl Cubic {
         max_datagram_size: f64,
         now: Instant,
     ) {
-        self.ca_epoch_start = Some(now);
+        self.t_epoch = Some(now);
         // reset tcp_acked_bytes and estimated_tcp_cwnd;
         self.tcp_acked_bytes = new_acked;
         self.estimated_tcp_cwnd = curr_cwnd;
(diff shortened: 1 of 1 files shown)


larseggert (approved):

neqo-transport/src/cc/cubic.rs
```diff
@@ -194,30 +194,30 @@ impl WindowAdjustment for Cubic {
...
+            // avoidance stage. It's been set to `None` by
+            // [`super::ClassicCongestionControl::reduce_cwnd`] or needs to be
+            // initialized after slow start. It could also have been reset by
+            // [`super::ClassicCongestionControl::on_app_limited`] in which case we also start a
+            // new congestion avoidance stage for the purpose of resetting
+            // timing as per RFC 9438 section 5.8.
+            //
+            // <https://datatracker.ietf.org/doc/html/rfc9438#app-limited>
+            self.start_epoch(curr_cwnd_f64, new_acked_f64, max_datagram_size_f64, now);
+            now
```
mxinden: It took me quite some time to understand that this `else` arm returns `now`, which then subtracts `now` once again, effectively returning `0` here.

Why not move the `now.saturating_duration_since` into the above `if` arm, returning `0` in this `else` arm?

neqo-transport/src/cc/cubic.rs
```diff
@@ -194,30 +194,30 @@ impl WindowAdjustment for Cubic {
...
         //
         // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.3>
         // <https://datatracker.ietf.org/doc/html/rfc8312#section-4.4>
-        let time_ca = self
-            .ca_epoch_start
-            .map_or(min_rtt, |t| {
-                if now + min_rtt < t {
-                    // This only happens when processing old packets
-                    // that were saved and replayed with old timestamps.
-                    min_rtt
```
mxinden, line 254: Do I understand correctly that this condition is now ignored? Why is that?

neqo-transport/src/cc/cubic.rs
```diff
@@ -194,30 +194,30 @@ impl WindowAdjustment for Cubic {
...
-                } else {
-                    now + min_rtt - t
-                }
-            })
-            .as_secs_f64();
-        let target_cubic = self.w_cubic(time_ca, max_datagram_size_f64);
+        let t = now.saturating_duration_since(t_epoch);
```
mxinden, line 260: Previously we would add `now + min_rtt`, and then subtract `t`. In cases where `now < t_epoch` but `now + min_rtt > t_epoch`, this is a functional change. Is that fine?

neqo-transport/src/cc/cubic.rs
omansfeld: Just so I understand you correctly: You're suggesting to move the calculation for `t` (that is then used in the cubic function) into the `if` arm, and setting `t = 0.0` in the else?

Something like

```rust
let t = if let Some(epoch) = self.t_epoch {
  self.tcp_acked_bytes += new_acked_f64;
  epoch.saturating_duration_since(now)
} else {
  self.start_epoch(...);
  0.0
}
```

That should work. My initial thought was to first unwrap `self.t_epoch` into the local `let t_epoch` and then later do the `let t = now - t_epoch` calculation. To me that is more readable and kinda follows the RFC wording

> where t is the elapsed time in seconds from the beginning of the current congestion avoidance stage -- that is, `t = t_current - t_epoch`and where t_epoch is the time at which the current congestion avoidance stage starts.

https://datatracker.ietf.org/doc/html/rfc9438#name-window-increase-function

I don't have strong opinions on this and am fine with either way. What do you prefer?

neqo-transport/src/cc/cubic.rs
omansfeld: With `saturating_suration_since` this is effectively now

```rust
let t = if now < t_epoch {
   0.0
} else {
   now - t_epoch
}
```

And then adding `min_rtt` in the `w_cubic(t + min_rtt)`function call.

This follows the RFC wording and doesn't hide adding `min_rtt` in the `t` calculation. ([1](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-6.1), [2](https://datatracker.ietf.org/doc/html/rfc9438#section-4.2-11.1))

You are correct (see your other comment) that this doesn't cover the edge case where `now < t_epoch` but `now + min_rtt > t_epoch`. As `now < t_epoch` is already an edge case that shouldn't really happen I think it is fair to just treat `t` as `0` in that case.

neqo-transport/src/cc/cubic.rs
omansfeld: answered [here](https://github.com/mozilla/neqo/pull/2997/files#r2358590439)

neqo-transport/src/cc/cubic.rs
mxinden: I see your point above. I find it confusing that we call `self.start_epoch` which sets `self.t_epoch` to `Some(now)`, but then don't make use of that value, but instead simply return `now` in the line after.

How about the following:

``` rust
if self.t_epoch.is_none() {
            // If we get here with `self.t_epoch == None` this is a new congestion
            // avoidance stage. It's been set to `None` by
            // [`super::ClassicCongestionControl::reduce_cwnd`] or needs to be
            // initialized after slow start. It could also have been reset by
            // [`super::ClassicCongestionControl::on_app_limited`] in which case we also start a
            // new congestion avoidance stage for the purpose of resetting
            // timing as per RFC 9438 section 5.8.
            //
            // <https://datatracker.ietf.org/doc/html/rfc9438#app-limited>
            self.start_epoch(curr_cwnd_f64, new_acked_f64, max_datagram_size_f64, now);
}

let t_epoch = self.t_epoch.expect("start_epoch sets t_epoch");

File: neqo-transport/src/cc/cubic.rs (hunk 2 above)
Comment (mxinden) [reply neqo-transport/src/cc/cubic.rs:254]: > As `now < t_epoch` is already an edge case that shouldn't really happen

Worth adding a `debug_assert!`?

**Review by mxinden — APPROVED** [review]
Thanks for the explanations! Two minor comments above. Otherwise good to merge from my side.

File: neqo-transport/src/cc/cubic.rs (hunk 2 above)
Comment (omansfeld) [reply neqo-transport/src/cc/cubic.rs:254]: > Worth adding a `debug_assert!`?

I don't think so. It is an edge case, but it can still happen, see old code comments:

https://github.com/mozilla/neqo/blob/b3d8f0d21db5657ebafb14f9b60c8892d2eb3aa9/neqo-transport/src/cc/cubic.rs#L211-L213

My argument for the `now < t_epoch` but `now + min_rtt > t_epoch` edge case not mattering is the following:

- `t = now - t_epoch`, so the time since the current epoch started
- the cubic function grows with bigger `t`
- if we replay old packets where `now < t_epoch`, so `t = 0.0` we will get a really small value (most likely smaller than the current cwnd), that likely won't be used anyways
- for `now + min_rtt > t_epoch` we'd have `0.0 <= t <= min_rtt` so the value returned by the cubic function would still be really small, and would also likely not be used

File: neqo-transport/src/cc/cubic.rs (hunk 1 above)
Comment (omansfeld) [reply neqo-transport/src/cc/cubic.rs]: We need the `t_epoch != None` case for incrementing `tcp_acked_bytes` so this wouldn't work. I like using `t_epoch.expect` though.

I've changed it so instead of `now` we use that to assign `t_epoch` in the `else`.

```rust
else {
            // If we get here with `self.t_epoch == None` this is a new congestion
            // avoidance stage. It's been set to `None` by
            // [`super::ClassicCongestionControl::reduce_cwnd`] or needs to be
            // initialized after slow start. It could also have been reset by
            // [`super::ClassicCongestionControl::on_app_limited`] in which case we also start a
            // new congestion avoidance stage for the purpose of resetting
            // timing as per RFC 9438 section 5.8.
            //
            // <https://datatracker.ietf.org/doc/html/rfc9438#app-limited>
            self.start_epoch(curr_cwnd_f64, new_acked_f64, max_datagram_size_f64, now);
            self.t_epoch
                .expect("unwrapping `none` value -- it should've been set by `start_epoch`")
        };
```

larseggert:
@omansfeld do you have permissions to merge this? Otherwise ping @mxinden or me when you are ready.

PR #3000, opened 2025-09-18
fix(transport): don't underflow on max MTU and GSO

mxinden:
A GSO datagram batch is not allowed to exceed the MTU of the corresponding IP address family. In the case of IPv4, this is `65535 - 20 - 8 = 65507`. In the case of IPv6 this is `65535 - 40 -8 = 65487`.

Previously the GSO batch generation code has been conservative, assuming the smaller IPv6 limit, i.e. `65487`. That said, on an IPv4 flow, PMTUD could increase the MTU all the way to the IPv4 limit.

This was problematic here:

https://github.com/mozilla/neqo/blob/ae02833019335ad7dd19edd9f3ef7dc4b692b7c2/neqo-transport/src/connection/mod.rs#L2571

`DatagramBatch::MAX` is equal to the smaller IPv6 limit of `65487`. `send_buffer.len` on an IPv4 flow with PMTUD can be up to the IPv4 limit, i.e. `65507`. The subtraction can therefore underflow.

This commit fixes the issue, consulting the `PMTUD` module for the accurate address family MTU instead.

---

Panic discovered by @valenting using the new MASQUE connect-udp implementation in combination with PMTUD :pray: 

Note that PMTUD is [not currently enabled in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1909910), thus this does not affect Firefox as of today.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ description = "Neqo, the Mozilla implementation of QUIC in Rust."
 keywords = ["quic", "http3", "neqo", "mozilla", "ietf", "firefox"]
 categories = ["network-programming", "web-programming"]
 readme = "README.md"
-version = "0.16.0"
+version = "0.16.1"
 # Keep in sync with `.rustfmt.toml` `edition`.
 edition = "2021"
 license = "MIT OR Apache-2.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -446,7 +446,7 @@ dependencies = [
 
 [[package]]
 name = "fuzz"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ description = "Neqo, the Mozilla implementation of QUIC in Rust."
 keywords = ["quic", "http3", "neqo", "mozilla", "ietf", "firefox"]
 categories = ["network-programming", "web-programming"]
 readme = "README.md"
-version = "0.16.0"
+version = "0.16.1"
```
mxinden, line 24: Note that I would like to cut a new patch release to get this bug fix into Firefox. Objections?

PR #3001, opened 2025-09-19
chore: Some more clippy lints

larseggert:


neqo-crypto/src/selfencrypt.rs
```diff
@@ -126,8 +126,13 @@ impl SelfEncrypt {
...
     /// Returns an error when the self-encrypted object is invalid;
     /// when the keys have been rotated; or when NSS fails.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `ciphertext` is too short to be valid.
     #[expect(clippy::similar_names, reason = "aad is similar to aead.")]
     pub fn open(&self, aad: &[u8], ciphertext: &[u8]) -> Res<Vec<u8>> {
+        assert!(ciphertext.len() > 1);
         if ciphertext[0] != Self::VERSION {
```
mxinden: Isn't this effectively the same? In other words, wouldn't Rust also always panic in the line below if `ciphertext.len()` is `0`?

mxinden (approved):
Don't understand the value of `missing_asserts_for_indexing`. Otherwise in favor.

larseggert:
`missing_asserts_for_indexing`:
> Checks for repeated slice indexing without asserting beforehand that the length
> is greater than the largest index used to index into the slice.
>
> Asserting that the length of the slice is at least as large as the largest value
> to index beforehand gives the compiler enough information to elide the bounds checks,
> effectively reducing the number of bounds checks from however many times
> the slice was indexed to just one (the assert).

Agree that it's of questionable benefit, but _might_ be helpful if such indexing ever gets added in a hot-path loop.

neqo-crypto/src/selfencrypt.rs
martinthomson: Should this return `Err(Something)` rather than asserting?

larseggert: It could. At the moment, we'd also just crash with an out-of-bounds error though.

martinthomson: That seems like a better fix.  We use this code for stuff we send to untrusted entities, so I think that I'd prefer that to letting them crash us out.  I suspect that we have a minimum length somewhere that guards this, but it's harmless to do: `ciphertext.get(0).ok_or(Error::SelfEncrypt)?`

PR #3002, opened 2025-09-21
refactor(http3): replace (Vec<Header>, bool, bool) with named struct

mxinden:
Minor follow-up to Lars' comment in https://github.com/mozilla/neqo/pull/2796#discussion_r2325101347.

### DIFF
--- a/neqo-http3/src/features/extended_connect/mod.rs
+++ b/neqo-http3/src/features/extended_connect/mod.rs
@@ -108,30 +108,44 @@ impl ExtendedConnectFeature {
     }
 }
 
+#[expect(
+    clippy::struct_field_names,
+    reason = "wrapper type, providing additional info"
+)]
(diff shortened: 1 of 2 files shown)


neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -108,17 +108,23 @@ impl ExtendedConnectFeature {
...
+struct HeaderInfo {
+    interim: bool,
+    fin: bool,
+}
+
 #[derive(Debug, Default)]
 struct Listener {
-    headers: Option<(Vec<Header>, bool, bool)>,
+    headers: Option<(Vec<Header>, HeaderInfo)>,
 }
```
larseggert, line 131: Why not stick the Vec in there as well and avoid the tuple?

PR #3003, opened 2025-09-22
feat: Reduce reallocations in RxStreamOrderer

larseggert:
Let's see if this helps performance.


larseggert:
Hm. Transfer test regression, but one bench shows an improvement. Time to look at flamegraphs...

larseggert:
Hm. Simplifying `inbound_frame` drastically doesn't seem to make things slower.

larseggert:
So what we save in `memcpy` we spend on `malloc` now :-)

larseggert:
Dropping in favor of #3690

PR #3005, opened 2025-09-22
fix: Put `qlog` behind a feature flag

### LINKED ISSUE #1894: Put `qlog` behind a feature

larseggert:
Also remove `qlog::add_event_data_now`, since we need to pass a timestamp around more places now anyhow. And rename `QlogMetric` to `qlog::Metric`.

~This is pretty intrusive. Anyone got an idea how to make this cleaner?~ Edit: It's a _bit_ better now.

Fixes #1894


mxinden:
Thanks for exploring the change.

This adds a lot of noise. Thus I suggest only proceeding here in case it has a significant performance win.

What will we do on the Firefox side? Currently one can dynamically enable qlog logging at runtime. Would we always enable the feature in Firefox? If so, how we run our Neqo tests would diverge to how we run Neqo in Firefox.

larseggert:
I would disable in release builds. So we ever ask for qlogs for bugzilla tickets?

neqo-http3/src/connection.rs
```diff
@@ -419,12 +429,24 @@ impl Http3Connection {
...
+            &mut self.recv_streams,
+            #[cfg(feature = "qlog")]
+            now,
+        )?;
 
-        self.send_non_control_streams(conn)?;
+        self.send_non_control_streams(
+            conn,
+            #[cfg(feature = "qlog")]
+            now,
```
martinthomson: I would not worry about making these conditional.  It's far too invasive to have that throughout.

martinthomson (dismissed):
So it seems like the most invasive action here is to make some `Instant` arguments conditional.  I would not worry about that: we can pass them through all the way.  There are relatively few places where we need to invoke `Instant::now()` to have that work and the argument lists don't bloat too much as a result.

Overall, the way I'd deal with this is to limit the conditional compilation to the qlog wrapper.  That can make the inner object conditional and stub out all the functions.

Because most of the work is deferred and uses closures, the compiler will be able to recognize that the closures aren't run, so it can erase that code.  I don't know how far up the call tree that will go in terms of killing code, but it should be far enough.

martinthomson:
> Do we ever ask for qlogs for bugzilla tickets?

Sometimes, yes.  Though maybe we can deliver a custom build in those cases.

larseggert:
> Because most of the work is deferred and uses closures, the compiler will be able to recognize that the closures aren't run, so it can erase that code. I don't know how far up the call tree that will go in terms of killing code, but it should be far enough.

That's what I was wondering, because if I only stub it out in neqo-common, only that crate will gain a `qlog` feature and hence all the other crates should be unaware of what that is set to and generate code for qlogging, even if that feature is off?

Let me see what it would look like if we also stubbed out qlog in the other crates.

neqo-bin/src/client/mod.rs
```diff
@@ -528,19 +532,29 @@ impl<'a, H: Handler> Runner<'a, H> {
     }
 }
 
+#[cfg_attr(
+    not(feature = "qlog"),
+    expect(unused_variables, reason = "Only used with qlog.")
+)]
 fn qlog_new(args: &Args, hostname: &str, cid: &ConnectionId) -> Res<Qlog> {
```
martinthomson: You could use

```rust
if cfg!(not(feature = "qlog")) {
  return Err(Error::Argument("qlog...");
}
```

I think that avoids all the unused variable warnings and keeps the code simple.

This crate, like most of the binary crates (and test-fixture), isn't directly exposed to the qlog crate, so it should be possible to go through the motions of setting up logging, which would do nothing if the neqo-common crate had the feature disabled.

The -transport and -http crates need to be exposed to qlog types, which means that it makes sense for disabling there to be more invasive.  However, there might be a different approach.

neqo-common/src/qlog.rs
```diff
@@ -36,6 +58,7 @@ impl Qlog {
     /// # Errors
     ///
     /// Will return `qlog::Error` if it cannot write to the new file.
+    #[cfg(feature = "qlog")]
```
martinthomson, line 61: Again, I would not feature guard the entire function, unless it is to define an alternative function implementation that returns a neutered value (without `inner`).

neqo-common/src/qlog.rs
```diff
@@ -71,10 +94,12 @@ impl Qlog {
     /// # Errors
     ///
     /// Will return `qlog::Error` if it cannot write to the new log.
+    #[cfg(feature = "qlog")]
     pub fn enabled(mut streamer: QlogStreamer, qlog_path: PathBuf) -> Result<Self, qlog::Error> {
         streamer.start_log()?;
 
         Ok(Self {
+            #[cfg(feature = "qlog")]
```
martinthomson: Why have this if the function is also conditional?

neqo-common/src/qlog.rs
```diff
@@ -89,10 +114,15 @@ impl Qlog {
     }
 
     /// If logging enabled, closure may generate an event to be logged.
+    #[cfg_attr(
+        not(feature = "qlog"),
+        expect(unused_variables, clippy::unused_self, reason = "Only used with qlog.")
+    )]
```
martinthomson, line 120: This function is good.  You can't use `if cfg!(` because that means taking in the qlog crate dependency.

neqo-common/src/qlog.rs
```diff
@@ -193,7 +215,7 @@ pub fn new_trace(role: Role) -> TraceSeq {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "qlog"))]
```
martinthomson, line 218: This is nicer than your BSD example in the other crate.

neqo-transport/src/qlog.rs
```diff
@@ -20,28 +22,40 @@ use qlog::events::{
...
+    not(feature = "qlog"),
+    expect(
+        unused_variables,
+        clippy::missing_const_for_fn,
+        reason = "Only used with qlog."
+    )
+)]
 pub fn connection_tparams_set(qlog: &Qlog, tph: &TransportParametersHandler, now: Instant) {
+    #[cfg(feature = "qlog")]
     qlog.add_event_data_with_instant(
```
martinthomson, line 59: If `add_event_data_with_instant` does nothing when qlog is disabled, do you really need all the guards around it?  The same applies for all these functions.

test-fixture/src/lib.rs
```diff
@@ -187,6 +193,7 @@ where
     )
     .expect("create a client");
 
+    #[cfg(feature = "qlog")]
```
martinthomson, line 197: I wonder if you can also avoid having to do this.  The extra work here is pretty trivial if calling `set_qlog` ultimately hits a noop.  It's not like this is production code.

test-fixture/src/sim/http3_connection.rs
```diff
@@ -303,15 +303,15 @@ impl Requests {
         loop {
             let remaining = self.remaining.get_mut(&stream_id).unwrap();
             let end = min(*remaining, DATA.len());
-            let sent = c.send_data(stream_id, &DATA[..end]).unwrap();
+            let sent = c.send_data(stream_id, &DATA[..end], now()).unwrap();
```
martinthomson: All these added now parameters is a bit of a distraction from the core reason for the change.  Could that be split out?

neqo-http3/src/qlog.rs
```diff
@@ -6,42 +6,69 @@
 
 // Functions that handle capturing QLOG traces.
```
martinthomson, line 7: I wonder if this (like -transport) could be a lot simpler:

```rust
#[cfg(feature = "qlog")]
mod qlog;
#[cfg(not(feature = "qlog"))]
mod qlog { include!("qlog-stub.rs"); }
```

No need then to have warnings and whatnot, you just have to keep the definitions synchronized.

The only reason you need this is that these functions are exposed to the types in the qlog crate and you don't want to compile the qlog crate when the feature is disabled.

neqo-bin/src/lib.rs
martinthomson, line 1: You should make the documentation on the `qlog_dir` command-line argument reflect whether qlog is enabled.  Either that or remove it entirely when the feature is disabled, but that makes the setup more complicated.

martinthomson (commented):
You've picked up some improvements that probably need to be taken up separately.

test-fixture/src/sim/http3_connection.rs
larseggert: I'll break those out into another PR.

neqo-http3/src/qlog.rs
larseggert: I somewhat dislike having multiple source files that are conditionally used, but I can make that change if you have a string preference.

martinthomson: Not entirely necessary, but your conditional compilation is pretty ugly as it is.  It's almost the case that you might be better off with duplicate functions:

```rust
/// Docs for `the_function`.
#[cfg(feature = "qlog")]
fn the_function(arg: Arg) -> Whatever {
  // actual code...
}
#[cfg(not(feature = "qlog"))]
fn the_function(_arg: Arg) -> Whatever {
  // Neutered code
}

File: neqo-bin/src/client/mod.rs (hunk 2 above)
Comment (larseggert) [reply neqo-bin/src/client/mod.rs]: That gets us an `unreachable_code` warning...

File: neqo-transport/src/qlog.rs (hunk 7 above)
Comment (larseggert) [reply neqo-transport/src/qlog.rs:59]: I think so, because otherwise things from the `quiche/qlog` crate are used, which is not a dep.

File: test-fixture/src/lib.rs (hunk 8 above)
Comment (larseggert) [reply test-fixture/src/lib.rs:197]: If I leave `qlog` as a dep anywhere in `neqo`, it will still be pulled into Gecko. (`new_neqo_qlog` depends on things from it.)

File: neqo-http3/src/features/extended_connect/session.rs [hunk 11]
```diff
@@ -304,9 +304,9 @@ impl Session {
         Ok(())
     }
 
-    pub(crate) fn add_stream(&mut self, stream_id: StreamId) -> Res<()> {
+    pub(crate) fn add_stream(&mut self, stream_id: StreamId, now: Instant) -> Res<()> {
```
mxinden, line 307: I am in favor of explicitly passing `now` everywhere. In other words, I am in favor of these changes to merge.

mxinden (commented):
I am hesitant to disable qlog in Firefox at compile time. We plan to invest more time into our congestion controller. qlog is currently our only way to debug congestion control issues beyond basic logging, e.g. by visualizing the congestion window over time. I think qlog will become a valuable tool in this work.

If we don't disable qlog in Firefox by default, I don't think we should do so in Neqo either, as Neqo benchmarks should match Firefox as closely as possible.

I acknowledge the problem. I have seen qlog show up in various profiles, especially its usage of `Rc`. I would expect the reference counting to be relatively cheap, compared to say our crypto operations. It is counterintuitive to me. Does someone have a good theory? Is there an alternative to disabling it?

larseggert:
> I am hesitant to disable qlog in Firefox at compile time.

We'd need to carefully explain to a user how to generate a qlog anyway. So I think @martinthomson's suggestion to ship them a custom build is therefore workable. Do you see a reason that wouldn't work?

mxinden:
Would work, yes. That said, I think we should make recording a qlog as low-friction as possible. Currently one has to [simply set a pref](https://firefox-source-docs.mozilla.org/networking/http/logging.html#how-to-enable-quic-logging). Shipping a custom build instead, of their specific Firefox version, for their specific OS version, is significantly more involved than that.

Yes qlog has an overhead. But compared to the many other low-hanging fruits we have in Firefox, is the above complexity worth the gain?

I am still puzzled why it shows up in profiles at all. We use `Rc` in various other hot paths. How is the `Rc` usage of qlog different?

larseggert:
One way forward would be to merge this with a default to "on", so we can more easily flip that later.

larseggert:
> One way forward would be to merge this with a default to "on", so we can more easily flip that later.

I did this now.

README.md
```diff
@@ -26,8 +26,9 @@ is suited to a live deployment.
 To build Neqo:
 
 ```shell
-cargo build
+cargo build --features qlog
```
mxinden: It is a default feature, thus no requirement to mention it here?

README.md
```diff
@@ -65,7 +66,7 @@ Note: If you did not already compile NSS separately, you need to have
 
 ### QUIC logging
 
-Enable generation of [QLOG][QLOG] logs with:
+When compiled with `--features qlog`, enable generation of [QLOG][QLOG] logs with:
```
mxinden: Needed given that it is a default feature?

mxinden (commented):
I am in favor of merging the changes around explicitly passing `now`.

I am indifferent to the introduction of the `qlog` feature. Given that it is on by default, and given that we never disable it, the disabled code path is effectively dead code without a near-term use case. (Please correct me if I am missing a use-case.) I am all for making qlog more performant. I don't think that performance exploration justifies a dead feature flag in `main`.

Feel free to merge here in case you think the `qlog ` feature is indeed useful.

README.md
larseggert: Correct. I left it in if we ever decide to flip it out of `default`, but can also remove it now,

README.md
larseggert: Same as above.

mxinden:
In regards to feature flagging, @larseggert what do you think of https://github.com/mozilla/neqo/pull/3129/ instead? It is less intrusive, and should provide the same performance characteristics as this pull request.

If you agree, I suggest merging the changes related to passing `now` here, but not the feature flagging. I would then clean up #3129.

larseggert:
Closing in favor of #3129.

PR #3006, opened 2025-09-24
feat: Remove use of `add_event_data_now`

larseggert:
So `qlog` always logs the time we maintain. Broken out of #3005.

### DIFF
--- a/neqo-bin/src/client/http3.rs
+++ b/neqo-bin/src/client/http3.rs
@@ -231,7 +231,7 @@ impl super::Handler for Handler {
                             qwarn!("Data on unexpected stream: {stream_id}");
                         }
                         Some(handler) => {
-                            handler.process_data_writable(client, stream_id);
+                            handler.process_data_writable(client, stream_id, Instant::now());
                         }
                     }
                 }
@@ -269,7 +269,12 @@ trait StreamHandler {
         data: &[u8],
         output_read_data: bool,
     ) -> Res<()>;
-    fn process_data_writable(&mut self, client: &mut Http3Client, stream_id: StreamId);
+    fn process_data_writable(
+        &mut self,
+        client: &mut Http3Client,
+        stream_id: StreamId,
+        now: Instant,
+    );
 }
 
 struct DownloadStreamHandler {
@@ -310,7 +315,13 @@ impl StreamHandler for DownloadStreamHandler {
         Ok(())
     }
 
-    fn process_data_writable(&mut self, _client: &mut Http3Client, _stream_id: StreamId) {}
+    fn process_data_writable(
+        &mut self,
+        _client: &mut Http3Client,
+        _stream_id: StreamId,
+        _now: Instant,
+    ) {
+    }
 }
 
 struct UploadStreamHandler {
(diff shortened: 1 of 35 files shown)


neqo-bin/src/client/http3.rs
```diff
@@ -342,12 +353,17 @@ impl StreamHandler for UploadStreamHandler {
...
+        now: Instant,
+    ) {
         let done = self
             .data
-            .send(|chunk| client.send_data(stream_id, chunk).unwrap());
+            .send(|chunk| client.send_data(stream_id, chunk, now).unwrap());
         if done {
-            client.stream_close_send(stream_id).unwrap();
+            client.stream_close_send(stream_id, now).unwrap();
         }
```
martinthomson, line 367: Can we disable mutants for this crate?

martinthomson (commented):
It's a big change here.

Is this all because of the `h3_data_moved_{up|down}` qlog calls in the -http3 crate?  Because it seems to me like those logs aren't particularly useful.  They only talk about local actions, not things that are observable on the wire.  And the cost is a big increase in API surface area.  All the stream-related actions now get times attached.

How necessary do you think these qlog traces are?  Because I might argue for their removal instead.  Then all the local actions that add data to streams and move stuff around within the stack don't need to know the time.

I certainly agree that `add_event_data_now` needs to go.  We should not be invoking `Instant::now()` anywhere in -transport, -http3, or -qpack.

larseggert:
> Is this all because of the `h3_data_moved_{up|down}` qlog calls in the -http3 crate? Because it seems to me like those logs aren't particularly useful.

It's those two and `qpack_read_insert_count_increment_instruction`. Which is probably also not super important to log.

I'd be fine with seeing those three events go, but would like to hear from @KershawChang and @mxinden. Not sure if those qlog events aren't helpful in debugging MASQUE.

larseggert:
We do actually (according to Claude) have calls to `Instant::now()` that are not in test code:

>⏺ Based on my analysis, here are the Instant::now() calls that are NOT in test code:
>
>  1. neqo-crypto/src/time.rs:49 - In the baseline() function (non-test code)
>  2. neqo-crypto/src/time.rs:75 - In the get_base() function initialization (non-test code)
>  3. neqo-common/src/qlog.rs:61 - In qlog initialization (non-test code)
>  4. neqo-http3/src/lib.rs:74 - In example/documentation code (non-test code)
>  5. neqo-http3/src/lib.rs:96 - In example/documentation code (non-test code)

The third looks like something we should fix.

martinthomson:
I agree that the qlog setup is probably worth fixing.  The challenge with initialization generally is that it is often done without reference to time.  But this one seems to have a time in every place it is invoked (only Server::setup_connection doesn't, but it's caller does, so that's a small fix).

mxinden (approved):
Noisy change, but in my eyes very much worth it. Thanks!

larseggert:
> I'd be fine with seeing those three events go, but would like to hear from @KershawChang and @mxinden. Not sure if those qlog events aren't helpful in debugging MASQUE.

See #3009 for how that would look like. CC @mxinden @KershawChang

martinthomson:
I have a slight preference for the alternative (#3009), but this is also fine.

mxinden (approved):
Preference for this option.

mxinden:
> Not sure if those qlog events aren't helpful in debugging MASQUE.

The more features we add to `neqo-http3` the more helpful qlog might become.

> I have a slight preference for the alternative (#3009), but this is also fine.

I think it is worth the additional complexity. That said, I will leave it up to you all.

larseggert:
Thanks for the feedback. I'm merging this one for now. We can always rip this out later.

PR #3008, opened 2025-09-24
fix: Don't call `Instant::now()` in `neqo-common/src/qlog.rs`

larseggert:
Pass it in instead.


martinthomson (approved):

martinthomson:
Lots of build failures.  The semver check looks like it is the most challenging.

larseggert:
SemVer check is not an issue, since we don't really "release". It's mainly there for the `mtu` sub-crate.

Codecov is confused again, because the uncovered lines were also previously uncovered.

PR #3009, opened 2025-09-25
feat: Remove use of `add_event_data_now` (alternative)

larseggert:
Remove the function and the code using it. Alternative to #3006.


martinthomson (approved):
For what it's worth, this is pretty nice.  It's not a huge loss in fidelity of the logs either.

larseggert:
@KershawChang @mxinden do you have a preference between this and #3006?

larseggert:
We went with #3006.

PR #3010, opened 2025-09-25
perf(http3): eliminate allocations with zero-copy `Bytes`

larseggert:
Fixes issue #2854 by implementing a zero-copy alternative to the closed PR #2988. 
Alternative to #2988

CC @mxinden

### DIFF
--- a/neqo-common/src/bytes.rs
+++ b/neqo-common/src/bytes.rs
@@ -62,6 +62,12 @@ impl AsRef<[u8]> for Bytes {
     }
 }
 
+impl AsMut<[u8]> for Bytes {
+    fn as_mut(&mut self) -> &mut [u8] {
+        &mut self.data[self.offset..]
+    }
+}
+
 impl PartialEq for Bytes {
     fn eq(&self, other: &Self) -> bool {
         self.as_ref() == other.as_ref()
@@ -69,14 +75,6 @@ impl PartialEq for Bytes {
 }
 
 impl Eq for Bytes {}
-
-impl From<Bytes> for Vec<u8> {
-    fn from(bytes: Bytes) -> Self {
-        // FIXME: This copies data, but the function is only used in tests.
-        // Tagging with `#[cfg(test)]` does not work.
-        bytes.as_ref().to_vec()
-    }
-}
 
 impl From<Vec<u8>> for Bytes {
     fn from(data: Vec<u8>) -> Self {
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -10,7 +10,7 @@ use std::{
     ops::{Deref, DerefMut},
 };
 
-use crate::{hex_with_len, Tos};
+use crate::{hex_with_len, Bytes, Tos};
 
 #[derive(Clone, PartialEq, Eq)]
 pub struct Datagram<D = Vec<u8>> {
@@ -126,6 +126,13 @@ impl<D: AsRef<[u8]>> Debug for Datagram<D> {
 impl<'a> Datagram<&'a mut [u8]> {
     #[must_use]
     pub fn from_slice(src: SocketAddr, dst: SocketAddr, tos: Tos, d: &'a mut [u8]) -> Self {
+        Self { src, dst, tos, d }
+    }
+}
+
+impl Datagram<Bytes> {
+    #[must_use]
+    pub const fn from_bytes(src: SocketAddr, dst: SocketAddr, tos: Tos, d: Bytes) -> Self {
         Self { src, dst, tos, d }
     }
 }
--- a/neqo-http3/tests/connect_udp.rs
+++ b/neqo-http3/tests/connect_udp.rs
@@ -139,7 +139,7 @@ fn exchange_packets_through_proxy(
     let server_dgrams = proxy.events().filter_map(|event| match event {
         Http3ServerEvent::ConnectUdp(ConnectUdpServerEvent::Datagram { datagram, session }) => {
             assert_eq!(session.stream_id(), connect_udp_session_id);
-            Some(Datagram::new(
+            Some(Datagram::from_bytes(
                 DEFAULT_ADDR,
                 DEFAULT_ADDR,
                 Tos::default(),
@@ -180,7 +180,7 @@ fn exchange_packets_through_proxy(
         }) = event
         {
             assert_eq!(session_id, connect_udp_session_id);
-            Some(Datagram::new(
+            Some(Datagram::from_bytes(
                 DEFAULT_ADDR,
                 DEFAULT_ADDR,
                 Tos::default(),


neqo-http3/src/lib.rs
```diff
@@ -184,6 +184,59 @@ use thiserror::Error;
...
+    }
+}
+
+impl AsRef<[u8]> for DatagramPayload {
+    fn as_ref(&self) -> &[u8] {
+        self.payload()
+    }
+}
+
+impl From<DatagramPayload> for Vec<u8> {
```
martinthomson: Want to implement a `From` in the other direction as well?  A `Vec<u8>` is trivially a `DatagramPayload` with zero offset.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -391,20 +391,28 @@ impl Session {
...
         }
-        let datagram = match self.protocol.dgram_context_id(datagram) {
-            Ok(datagram) => datagram,
+
+        // Validate the payload portion based on protocol requirements
+        match self.protocol.dgram_context_id(&datagram[payload_offset..]) {
+            Ok(slice) => {
+                // Calculate total offset: session_id varint + context_id (if any)
+                // If the returned slice is shorter than the input, a context identifier is present.
+                let context_offset = usize::from(slice.len() != datagram[payload_offset..].len());
```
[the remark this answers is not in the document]
martinthomson: Copilot might be more correct than you think.  

I think that there is a bug in the implementation of `dgram_context_id` in the UDP session.  I could be wrong, so check my work.

For webtransport, this is implemented as a stub.  It does not have a context ID.  So the function returns the whole payload. That's OK, because you check the lengths (which, being a bit of an odd way to check, was the red flag for me).

For UDP, the function splits the first byte.  But [context IDs are a varint](https://www.rfc-editor.org/rfc/rfc9298.html#section-4).  So that will fail.  Maybe that's OK, because we only support a zero-valued context ID for now.  But 0x4000 is a valid encoding of a zero.  So that function is subtly incorrect.  It really needs to do a varint decode, even if the only value we accept is zero.

neqo-http3/src/server_events.rs
```diff
@@ -501,7 +501,7 @@ pub enum WebTransportServerEvent {
     NewStream(Http3OrWebTransportStream),
     Datagram {
         session: WebTransportRequest,
-        datagram: Vec<u8>,
+        datagram: crate::DatagramPayload,
```
martinthomson: Use `use crate::DatagramPayload`

neqo-http3/src/lib.rs
```diff
@@ -184,6 +184,59 @@ use thiserror::Error;
 
 use crate::{features::extended_connect, priority::PriorityHandler};
 
+#[derive(Debug, Clone, Eq, PartialEq)]
```
martinthomson: Might be worth a doc comment here.

neqo-http3/src/lib.rs
```diff
@@ -184,6 +184,59 @@ use thiserror::Error;
...
+pub struct DatagramPayload {
```
martinthomson: This seems generally useful.  Any thought about a different name?

neqo-http3/src/lib.rs
```diff
@@ -184,6 +184,59 @@ use thiserror::Error;
...
+        self.payload()
+    }
+}
+
+impl From<DatagramPayload> for Vec<u8> {
+    fn from(payload: DatagramPayload) -> Self {
+        if payload.payload_offset == 0 {
+            payload.data
+        } else {
+            payload.data[payload.payload_offset..].to_vec()
```
larseggert: There is still a `to_vec()` here...

neqo-http3/src/features/extended_connect/session.rs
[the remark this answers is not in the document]
mxinden: Good catch. Thanks @martinthomson. I will fix this. Tracked in https://github.com/mozilla/neqo/issues/3017.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -391,20 +391,28 @@ impl Session {
         Ok(())
     }
 
-    pub(crate) fn datagram(&self, datagram: &[u8]) {
+    pub(crate) fn datagram_owned(&self, datagram: Vec<u8>, payload_offset: usize) {
```
```suggestion
    pub(crate) fn datagram(&self, datagram: Vec<u8>, payload_offset: usize) {
```

Given that there is no `datagram`, I don't think we need the `_owned` suffix.

neqo-http3/src/lib.rs
mxinden: How about `Bytes` in accordance with https://docs.rs/bytes/latest/bytes/ ?

Would as well allow it to move into `neqo-common` and thus find other applications.

neqo-http3/src/lib.rs
```diff
@@ -184,6 +184,59 @@ use thiserror::Error;
...
+}
+
+impl<const N: usize> PartialEq<&[u8; N]> for DatagramPayload {
+    fn eq(&self, other: &&[u8; N]) -> bool {
+        self.payload() == other.as_slice()
+    }
+}
+
+impl PartialEq<&[u8]> for DatagramPayload {
+    fn eq(&self, other: &&[u8]) -> bool {
```
mxinden:
```suggestion
impl PartialEq<[u8]> for DatagramPayload {
    fn eq(&self, other: &[u8]) -> bool {
```

Would this work as well?

mxinden (commented):
In favor.

larseggert:
@mxinden is that what you had in mind regarding `Bytes`?

mxinden:
Sorry for not being very clear in my last review. Your initial implementation (not importing the `Bytes` crate) was surprisingly simple. My suggestion in my review was to rename `DatagramPayload` to `Bytes` (not to be confused with the `Bytes` crate), and move `Bytes` to `neqo-common`.

If you say instead of writing our own `Bytes`, we should just use the `Bytes` crate, as you do in the latest PR version here, that is fine by me as well.

(Historic context: I initially suggested to use the `Bytes` crate. That said, your initial implementation without the crate, looked so simple, might actually be worth to have our own. Sorry for the confusion.)

No matter what we do, the type we are using here won't leak beyond `neqo_glue`, as we will copy it into a `ThinVec` there anyways:

https://github.com/mozilla-firefox/firefox/blob/1c6a8b56e6f5f5393ea5ea745ab63ecb67613c93/netwerk/socket/neqo_glue/src/lib.rs#L1648-L1656

larseggert:
The original implementation (done with Claude), still had a `to_vec()`, which is why I discounted it: https://github.com/mozilla/neqo/pull/3010#discussion_r2381281060

mxinden:
Why did we need the `to_vec()` with the custom implementation and not with `Bytes`? If I delete it, `neqo-*` still compiles:


``` diff
diff --git a/neqo-http3/src/lib.rs b/neqo-http3/src/lib.rs
index ae0bbee2..0bdbbfbb 100644
--- a/neqo-http3/src/lib.rs
+++ b/neqo-http3/src/lib.rs
@@ -215,16 +215,6 @@ impl AsRef<[u8]> for DatagramPayload {
     }
 }
 
-impl From<DatagramPayload> for Vec<u8> {
-    fn from(payload: DatagramPayload) -> Self {
-        if payload.payload_offset == 0 {
-            payload.data
-        } else {
-            payload.data[payload.payload_offset..].to_vec()
-        }
-    }
-}
-
```

larseggert:
Heh. Claude. I didn't think to check whether that function was actually used...

In that case, the original would IMO be best, since it avoids the dep. I'll revert the PR.

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl AsRef<[u8]> for Bytes {
+    fn as_ref(&self) -> &[u8] {
+        self.data()
+    }
+}
+
+impl From<Bytes> for Vec<u8> {
+    fn from(payload: Bytes) -> Self {
+        payload.data().to_vec()
```
larseggert: @mxinden I don't see a way around this `to_vec()`, do you?

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
mxinden:
```suggestion
/// Owned contiguous byte array with an optional offset.
///
/// Inspired by `bytes` crate's `Bytes`.
#[derive(Debug, Clone, PartialEq, Eq)]
```

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[must_use]
+    pub const fn new(data: Vec<u8>, offset: usize) -> Self {
+        Self { data, offset }
+    }
+
+    #[must_use]
+    pub fn data(&self) -> &[u8] {
+        &self.data[self.offset..]
+    }
+}
```
mxinden, line 57: Given that the type implements `AsRef<[u8]>`, is `fn data` necessary?

neqo-common/src/bytes.rs
mxinden: Why is it needed? The following compiles fine for me:

``` diff
mxinden@mxinden-framework ~/c/g/m/neqo (fix-2854-v2)> git diff
diff --git a/neqo-common/src/bytes.rs b/neqo-common/src/bytes.rs
index eebccfe1..615b7b89 100644
--- a/neqo-common/src/bytes.rs
+++ b/neqo-common/src/bytes.rs
@@ -28,12 +28,6 @@ impl AsRef<[u8]> for Bytes {
     }
 }
 
-impl From<Bytes> for Vec<u8> {
-    fn from(payload: Bytes) -> Self {
-        payload.data().to_vec()
-    }
-}
-
 impl<const N: usize> PartialEq<[u8; N]> for Bytes {
     fn eq(&self, other: &[u8; N]) -> bool {
         self.data() == other.as_slice()
```

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl From<Bytes> for Vec<u8> {
+    fn from(payload: Bytes) -> Self {
+        payload.data().to_vec()
+    }
+}
+
+impl<const N: usize> PartialEq<[u8; N]> for Bytes {
```
mxinden, line 85: Where is this needed?

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn eq(&self, other: &[u8; N]) -> bool {
+        self.data() == other.as_slice()
+    }
+}
+
+impl<const N: usize> PartialEq<&[u8; N]> for Bytes {
```
mxinden: Where is this needed?

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn eq(&self, other: &&[u8; N]) -> bool {
+        self.data() == other.as_slice()
+    }
+}
+
+impl PartialEq<[u8]> for Bytes {
```
mxinden, line 91: Where is this needed?

neqo-common/src/bytes.rs
mxinden:
```suggestion
#[derive(Debug, Clone)]
```

Are the derived `PartialEq` and `Eq` implementations intuitive? I assume we would consider the following two `Bytes` equal?

- `Bytes { data: [1, 2, 3, 4], offset: 1}`
- `Bytes { data: [2, 3, 4], offset: 0 }`

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -398,20 +398,28 @@ impl Session {
         Ok(())
     }
 
-    pub(crate) fn datagram(&self, datagram: &[u8]) {
+    pub(crate) fn datagram(&self, datagram: Vec<u8>, payload_offset: usize) {
```
mxinden: Why provide a `Vec` and an offset here? Why not provide the `Bytes` right away?

neqo-common/src/bytes.rs
larseggert: Huh. I get this error:
```
error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<neqo_common::Bytes>` is not satisfied
   --> neqo-http3/tests/connect_udp.rs:146:17
    |
142 |             Some(Datagram::new(
    |                  ------------- required by a bound introduced by this call
...
146 |                 datagram,
    |                 ^^^^^^^^ the trait `std::convert::From<neqo_common::Bytes>` is not implemented for `std::vec::Vec<u8>`
    |
    = help: the following other types implement trait `std::convert::From<T>`:
              `std::vec::Vec<u8>` implements `std::convert::From<&str>`
              `std::vec::Vec<u8>` implements `std::convert::From<neqo_common::Encoder>`
              `std::vec::Vec<u8>` implements `std::convert::From<std::bstr::ByteString>`
              `std::vec::Vec<u8>` implements `std::convert::From<std::ffi::CString>`
              `std::vec::Vec<u8>` implements `std::convert::From<std::string::String>`
    = note: required for `neqo_common::Bytes` to implement `std::convert::Into<std::vec::Vec<u8>>`
note: required by a bound in `neqo_common::Datagram::new`
   --> /Volumes/mozdrone/Users/mozdrone/Documents/Code/neqo/main/neqo-common/src/datagram.rs:90:19
    |
 90 |     pub fn new<V: Into<Vec<u8>>>(src: SocketAddr, dst: SocketAddr, tos: Tos, d: V) -> Self {
    |                   ^^^^^^^^^^^^^ required by this bound in `Datagram::new`

```

neqo-common/src/bytes.rs
larseggert:
```
error[E0277]: can't compare `neqo_common::Bytes` with `[u8; 4]`
   --> neqo-http3/tests/webtransport.rs:493:76
    |
493 |             )) if session_id == wt_server_session.stream_id() && &datagram == b"PING",
    |                                                                            ^^ no implementation for `neqo_common::Bytes == [u8; 4]`
    |
    = help: the trait `std::cmp::PartialEq<[u8; 4]>` is not implemented for `neqo_common::Bytes`
    = help: the following other types implement trait `std::cmp::PartialEq<Rhs>`:
              `neqo_common::Bytes` implements `std::cmp::PartialEq<[u8]>`
              `neqo_common::Bytes` implements `std::cmp::PartialEq`
    = note: required for `&neqo_common::Bytes` to implement `std::cmp::PartialEq<&[u8; 4]>`

```

neqo-common/src/bytes.rs
larseggert:
```
error[E0277]: can't compare `neqo_common::Bytes` with `[u8]`
   --> neqo-http3/tests/connect_udp.rs:223:5
    |
223 |     assert_eq!(&datagram, PING);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `neqo_common::Bytes == [u8]`
    |
    = help: the trait `std::cmp::PartialEq<[u8]>` is not implemented for `neqo_common::Bytes`
    = help: the following other types implement trait `std::cmp::PartialEq<Rhs>`:
              `neqo_common::Bytes` implements `std::cmp::PartialEq<[u8; N]>`
              `neqo_common::Bytes` implements `std::cmp::PartialEq`
    = note: required for `&neqo_common::Bytes` to implement `std::cmp::PartialEq<&[u8]>`
    = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)

```

neqo-common/src/bytes.rs
larseggert: Good catch!

neqo-common/src/bytes.rs
mxinden: Given that this is test code only, how about something along the lines of:

`datagram.as_ref() == b"PING",`

Not a strong opinion.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -398,20 +398,23 @@ impl Session {
...
+    pub(crate) fn datagram(&self, datagram: Bytes) {
         if self.state != State::Active {
             qdebug!("[{self}]: received datagram on {:?} session.", self.state);
             return;
         }
-        let datagram = match self.protocol.dgram_context_id(datagram) {
-            Ok(datagram) => datagram,
+
+        // dgram_context_id returns the payload after stripping any context ID; length difference
+        // indicates context ID presence.
```
mxinden: We don't check the length difference below. Why is this comment needed?

neqo-common/src/bytes.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl Eq for Bytes {}
+
+impl From<Bytes> for Vec<u8> {
+    fn from(bytes: Bytes) -> Self {
+        // FIXME: This copies data, but the function is only used in tests.
+        // Tagging with with `#[cfg(test)]` does not work.
+        bytes.as_ref().to_vec()
+    }
+}
```
larseggert: @mxinden Removing this causes a bunch of breakage in `datagram.rs` - would you mind taking a look? I can't seem to figure out a fix.

mxinden: @larseggert what do you think of https://github.com/larseggert/neqo/pull/38?

larseggert: Awesome! I tried to change Datagram in a similar way but couldn't get it to work

Issue #2544, opened 2025-03-31
Consider implementing connection level flow control


mxinden:
We currently do not implement connection level flow control, but instead simply grant the remote an excessively large connection level send window:

https://github.com/mozilla/neqo/blob/fab0046d2556eb6603c2354552a025d8ee0b782f/neqo-transport/src/connection/params.rs#L30

If we had connection level flow control, we could use its bandwidth-delay product estimate to bootstrap the receive window of a new stream flow control. The sender would thus be able to use the full bandwidth of the connection on that single stream early on.

Originally raised in: 
 
 > As for the follow-up idea, having connection-level flow control would be nice because it allows us to estimate actual BDP *usage*, not availability with overheads.  If we had that, then it might be easier to produce estimates for things like WebTransport throughput availability estimates.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/1868#pullrequestreview-2728125376_

abhiram-shaji:
Hi, I’d like to take a stab at addressing this issue.

I Plan to Implement Connection-Level Flow Control:
Replace the Static LOCAL_MAX_DATA Constant

Remove the hardcoded 0x3FFF_FFFF_FFFF_FFFF value and replace it with a dynamic estimate based on the bandwidth-delay product (BDP).

Add a BDP Estimator

Introduce a basic BDP estimator that calculates max_data = RTT × bandwidth, with some headroom to account for variations.

Use Estimated max_data in ConnectionParameters::default()

Replace the static value in the default implementation with the dynamically estimated BDP window.

Enable Runtime Adjustment (Optional for Future Work)

Optionally expose a method to update max_data based on live RTT/bandwidth feedback from the congestion controller.

Ensure Transport Parameters Use the Updated Value

InitialMaxData is already populated from self.max_data, so no change is needed there.

larseggert:
> Hi, I’d like to take a stab at addressing this issue.

Thanks for you interest in contributing. I have a few questions on your plan; I think I must be misunderstanding something fundamental:

> Add a BDP Estimator
> 
> Introduce a basic BDP estimator that calculates max_data = RTT × bandwidth, with some headroom to account for variations.
> 
> Use Estimated max_data in ConnectionParameters::default()

`ConnectionParameters::default` is used before we have RTT and bandwidth information. I don't quite see how this would work.

> Enable Runtime Adjustment (Optional for Future Work)
> 
[...]

PR #3011, opened 2025-09-25
fix: Increase stream limits to 100

### LINKED ISSUE #2921: Consider increasing default local stream limit
Neqo currently sets a default local stream limit of `16`.

https://github.com/mozilla/neqo/blob/ea6cde436bcf67e71deba0809978312fbf3ee734/neqo-transport/src/connection/params.rs#L33-L34

This value seems very conservative to me. For reference, Google Quiche seems to use a default of `100`:

https://github.com/google/quiche/blob/4f1f0fcea045cd71410c2c318773fc24c3523ed7/quiche/quic/core/quic_constants.h#L113-L114

What is the rational for the limit of `16`?

With upcoming WebTransport and MoQ work, and potentially connection pooling, i.e. HTTP/3 and WebTransport connection sharing, would it make sense to increase it?

larseggert:
Fixes #2921

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -30,8 +30,11 @@ use crate::{
 };
 
 const LOCAL_MAX_DATA: u64 = MAX_VARINT;
-const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
-const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
+
+// Same as Google: https://github.com/google/quiche/blob/4f1f0fcea045cd71410c2c318773fc24c3523ed7/quiche/quic/core/quic_constants.h#L113-L114
+const LOCAL_STREAM_LIMIT_BIDI: u64 = 100;
+const LOCAL_STREAM_LIMIT_UNI: u64 = 100;
+
 const MAX_QUEUED_DATAGRAMS_DEFAULT: usize = 10;
 
 /// What to do with preferred addresses.


mxinden (approved):

mxinden:
Pausing here for a second. Something that came to mind. We currently don't do connection flow control (https://github.com/mozilla/neqo/issues/2544), but only flow-control on number of streams and number of bytes on streams. Bumping the limit to 100 significantly increases the amount of bytes we grant to the remote. Is that a security concern?

mxinden:
Currently we grant an initial of 1 MB of flow control window to each stream.

https://github.com/mozilla/neqo/blob/d46f3f4a9454bb50fc40f6251b43d43442a5db00/neqo-transport/src/recv_stream.rs#L36

A limit of 16 streams allows the remote to allocate 16 MB per connection. A limit of 100 streams allows the remote to allocate 100 MB per connection.

larseggert:
Yeah, that's not great. We might need to think here.

martinthomson:
We have a connection-level flow control limit, right?  Are we auto-tuning that too?  Because that's the real safeguard on total allocation allowance.  Maybe we should look more closely at that.

Of course, because we auto-tune, that is essentially unbounded, provided that we are able to service the throughput.  That's the real limit.

mxinden:
> We have a connection-level flow control limit, right?

We currently do not have connection-level flow control. Tracked in https://github.com/mozilla/neqo/issues/2544.

mxinden:
Closing here in favor of https://github.com/mozilla/neqo/pull/3107.

PR #3012, opened 2025-09-25
fix: Bump `max_datagram_frame_size` TP to 64K

larseggert:
Per the RFC.

### DIFF
--- a/neqo-transport/src/connection/tests/datagram.rs
+++ b/neqo-transport/src/connection/tests/datagram.rs
@@ -10,8 +10,7 @@ use neqo_common::event::Provider as _;
 use static_assertions::const_assert;
 
 use super::{
-    assert_error, connect_force_idle, default_client, default_server, new_client, new_server, now,
-    AT_LEAST_PTO,
+    assert_error, connect_force_idle, default_server, new_client, new_server, now, AT_LEAST_PTO,
 };
 use crate::{
     connection::tests::DEFAULT_ADDR,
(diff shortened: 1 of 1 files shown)


larseggert:
Is the failing `dgram_no_allowed` test still useful with this change? Should it simply be removed?

neqo-transport/src/connection/params.rs
```diff
@@ -32,6 +32,8 @@ use crate::{
 const LOCAL_MAX_DATA: u64 = MAX_VARINT;
 const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
 const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
+// Maximum size of a QUIC DATAGRAM frame, as specified in RFC 9000 Section 7.4 (64KB).
+const MAX_DATAGRAM_FRAME_SIZE: u64 = 65536;
```
mxinden:
```suggestion
const MAX_DATAGRAM_FRAME_SIZE: u64 = 65535;
```

Note that the RFC suggests 65535.

> For most uses of DATAGRAM frames, it is RECOMMENDED to send a value of 65535 in the max_datagram_frame_size transport parameter to indicate that this endpoint will accept any DATAGRAM frame that fits inside a QUIC packet.

https://www.rfc-editor.org/rfc/rfc9221.html#section-3

larseggert:
> Needs a fix in `dgram_no_allowed`, i.e. setting the transport parameter lower for the client, such that the server's datagram leads to a protocol violation on the client side.

This is what the `dgram_too_big` test already does, hence my suggestion that this test may not be needed anymore.

mxinden:
> > Needs a fix in `dgram_no_allowed`, i.e. setting the transport parameter lower for the client, such that the server's datagram leads to a protocol violation on the client side.
> 
> This is what the `dgram_too_big` test already does, hence my suggestion that this test may not be needed anymore.

I am fine removing it.

PR #3013, opened 2025-09-25
fix: Replace `qpack::Data` with trait

### LINKED ISSUE #2754: Replace `qpack::Data` with trait.
This name is awful.  What is it?  From the implementation, it looks like a buffer, not a unit of data.

Not for this PR, but this looks like it could be replaced by a trait that defines the encode_prefixed_encoded_int (also a name that could be better) or maybe just the encode_literal methods and attaches them to `neqo_common::Encoder`.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2753#discussion_r2163059672_

larseggert:
Fixes #2754

### DIFF
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -409,6 +409,14 @@ impl Encoder<Vec<u8>> {
         Self::default()
     }
 
+    /// Create a new [`Encoder`] that takes ownership of the provided [`Vec<u8>`].
+    /// The start position is set to the current length of the vector.
+    #[must_use]
+    pub fn new_with_vec(buf: Vec<u8>) -> Self {
+        let start = buf.len();
+        Self { buf, start }
+    }
+
     /// Static helper function for previewing the results of encoding without doing it.
     ///
     /// # Panics
--- a/neqo-qpack/src/decoder.rs
+++ b/neqo-qpack/src/decoder.rs
@@ -6,14 +6,13 @@
 
 use std::fmt::{self, Display, Formatter};
 
-use neqo_common::{qdebug, Header};
+use neqo_common::{qdebug, Encoder, Header};
 use neqo_transport::{Connection, StreamId};
 
 use crate::{
     decoder_instructions::DecoderInstruction,
     encoder_instructions::{DecodedEncoderInstruction, EncoderInstructionReader},
     header_block::{HeaderDecoder, HeaderDecoderResult},
-    qpack_send_buf::Data,
     reader::ReceiverConnWrapper,
     stats::Stats,
     table::HeaderTable,
@@ -28,7 +27,7 @@ pub struct Decoder {
     table: HeaderTable,
     acked_inserts: u64,
     max_entries: u64,
-    send_buf: Data,
+    send_buf: Encoder,
     local_stream_id: Option<StreamId>,
     max_table_size: u64,
     max_blocked_streams: usize,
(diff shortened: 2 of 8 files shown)


neqo-common/src/codec.rs
```diff
@@ -409,6 +409,32 @@ impl Encoder<Vec<u8>> {
...
+    ///
+    /// The encoder will treat any existing data in the vector as "pre-existing"
+    /// and will append new encoded data after it. The start position is set to
+    /// the current length of the vector, so only newly encoded data will be
+    /// considered part of the encoder's content.
+    ///
+    /// This is useful when you want to preserve existing buffer contents while
+    /// continuing to encode additional data.
+    #[must_use]
+    pub fn new_with_vec(buf: Vec<u8>) -> Self {
```
mxinden: Mind adding a small test?

neqo-qpack/src/decoder.rs
```diff
@@ -175,11 +174,16 @@ impl Decoder {
...
                     self.local_stream_id.ok_or(Error::Internal)?,
-                    &self.send_buf[..],
+                    self.send_buf.as_ref(),
                 )
                 .map_err(|_| Error::DecoderStream)?;
             qdebug!("[{self}] {r} bytes sent");
-            self.send_buf.read(r);
+            if r < self.send_buf.len() {
+                // Efficiently drain the sent bytes from the buffer
+                self.send_buf.drain(r).for_each(drop);
```
mxinden: Note that this shifts all remaining bytes in `self.send_buf` by `r` to the left. Does that still count as "efficient"?

If we always drain from the front, your [newly introduced `Bytes` structure](https://github.com/mozilla/neqo/pull/3010) might be better, as it can `skip` without shifting any bytes.

neqo-qpack/src/qpack_send_buf.rs
```diff
@@ -4,27 +4,28 @@
...
-use std::ops::Deref;
-
 use neqo_common::Encoder;
 
 use crate::{huffman, prefix::Prefix};
 
-#[derive(Default, Debug, PartialEq, Eq)]
-pub struct Data {
-    buf: Vec<u8>,
+/// Trait providing QPACK-specific encoding methods for `Encoder`.
```
mxinden:
```suggestion
/// Extension trait providing QPACK-specific encoding methods for `Encoder`.
```

Generally referred to as extension traits. Feel free to ignore.

https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

neqo-qpack/src/qpack_send_buf.rs
```diff
@@ -4,27 +4,28 @@
...
+///
+/// This trait extends the standard `neqo_common::Encoder` with QPACK-specific
```
mxinden:
```suggestion
/// This trait extends the standard [`neqo_common::Encoder`] with QPACK-specific
```

neqo-qpack/src/decoder.rs
larseggert: I ended up skipping rather than draining, because switching to `Bytes` has a lot of ripple effects.

neqo-common/src/codec.rs
```diff
@@ -409,6 +409,17 @@ impl Encoder<Vec<u8>> {
...
+    /// Skip the first `n` bytes from the encoder buffer without copying.
+    /// This advances the internal offset, making those bytes inaccessible.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `n` is greater than the current length of the encoder.
+    pub fn skip(&mut self, n: usize) {
+        assert!(n <= self.len(), "Cannot skip beyond buffer length");
+        self.start += n;
+    }
```
mxinden, line 437: While this does make the skipped bytes inaccessible through most of `Encoder`'s methods, for example this `From` impl still allows to access those bytes:

https://github.com/mozilla/neqo/blob/e94d8c6c0f6ecf9df594eeb6fb56136f0d841343/neqo-common/src/codec.rs#L506-L510

How do we ensure users of `Encoder::skip` don't also make use of `From<Encoder> for Vec<u8>>`?

larseggert: @mxinden I tried to address this in the current version.

PR #3015, opened 2025-09-26
feat(http3): optimize stream sending to avoid blocked streams

### LINKED ISSUE #261: Optimise sending data in http3 code

larseggert:
Implements optimizations to address wasteful attempts to send data on flow-control blocked streams and fixes iteration patterns that could lose streams during re-addition.

Key improvements:
- Add blocked_streams HashSet to track flow-control blocked streams separately
- Modify stream_has_pending_data() to skip adding blocked streams to pending queue
- Add check_blocked_streams() method to automatically move unblocked streams back

Benefits:
- Reduces CPU usage by eliminating repeated attempts on blocked streams
- Improves responsiveness by automatically detecting when streams become unblocked

Closes #261

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1882,13 +1882,163 @@ impl Http3Connection {
 #[cfg(test)]
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
+    use neqo_common::Role;
+    use neqo_transport::StreamId;
+    use test_fixture::fixture_init;
     use url::Url;
 
     use crate::{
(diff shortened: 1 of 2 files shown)


mxinden (commented):
This increases the amount of state tracking, i.e. does not come for free both in terms of complexity, and potentially in terms of compute. Unless we have a benchmark showing this improves performance, i.e. shows that the additional tracking is cheaper than the continuous attempt at sending on a blocked stream, I don't think we should merge.

PR #3016, opened 2025-09-28
chore(transport/pmtud): improve logging

mxinden:
Add two new log events:

- on reaching the end of the search table
- on hitting the interface MTU limit

### DIFF
--- a/neqo-transport/src/pmtud.rs
+++ b/neqo-transport/src/pmtud.rs
@@ -315,21 +315,32 @@ impl Pmtud {
 
     /// Starts the next upward PMTUD probe.
     pub fn start(&mut self, now: Instant, stats: &mut Stats) {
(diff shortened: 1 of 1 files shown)


neqo-transport/src/pmtud.rs
```diff
@@ -315,21 +315,32 @@ impl Pmtud {
 
     /// Starts the next upward PMTUD probe.
     pub fn start(&mut self, now: Instant, stats: &mut Stats) {
-        if self.probe_index < SEARCH_TABLE_LEN - 1 // Not at the end of the search table
-        // Next size is <= iface MTU
-            && self.search_table[self.probe_index + 1] <= self.iface_mtu
-        {
```
mxinden, line 321: Instead of an `if` `else` block, this pull request turns the `else` into early returns before the `if` arm.

PR #3024, opened 2025-10-01
fix: Don't condition `neqo_common::log::init` on `cfg(test)`

### LINKED ISSUE #2368: Don't condition `neqo_common::log::init` on `cfg(test)`
Turns out we cannot make the initialization conditioned on `cfg(test)`, because before a test uses one of the logging macros, the log init won't happen and there will no output :-(

_Originally posted by @larseggert in https://github.com/mozilla/neqo/pull/2291#discussion_r1918257346_

larseggert:
Turns out we cannot make the initialization conditioned on `cfg(test)`, because before a test uses one of the logging macros, the log init won't happen and there will no output :-(

Fixes #2368

### DIFF
--- a/neqo-common/src/log.rs
+++ b/neqo-common/src/log.rs
@@ -52,6 +52,12 @@ pub fn init(level_filter: Option<log::LevelFilter>) {
 // TODO: Enable `#[clippy::format_args]` once our MSRV is >= 1.84
 macro_rules! qerror {
     ($($arg:tt)*) => ( {
+<<<<<<< Updated upstream
+||||||| Stash base
+        #[cfg(any(test, feature = "bench"))]
+=======
+        #[cfg(debug_assertions)]
+>>>>>>> Stashed changes
         ::neqo_common::log::init(None);
         ::log::error!($($arg)*);
     } );
@@ -60,6 +66,12 @@ macro_rules! qerror {
 // TODO: Enable `#[clippy::format_args]` once our MSRV is >= 1.84
 macro_rules! qwarn {
     ($($arg:tt)*) => ( {
+<<<<<<< Updated upstream
+||||||| Stash base
+        #[cfg(any(test, feature = "bench"))]
+=======
+        #[cfg(debug_assertions)]
+>>>>>>> Stashed changes
         ::neqo_common::log::init(None);
         ::log::warn!($($arg)*);
     } );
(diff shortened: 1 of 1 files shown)


mxinden:
> Turns out we cannot make the initialization conditioned on `cfg(test)`, because before a test uses one of the logging macros, the log init won't happen and there will no output :-(

Despite what I said #2368, I don't fully understand the above.

Are you saying that e.g. a `qdebug` in non-test code won't trigger unless there is a `qdebug` in test code as well? Why is that? `cfg(test)` is a compile time option, thus should apply to the whole binary, i.e. both non-test and test code.

Without `cfg(test)` we would as well initialize in Firefox, correct? Is that safe?

larseggert:
The issue is that the log init call in the `qdebug` macro does not execute in test code.

mxinden:
Why is that? The `cfg(test)` should be true in that case, no?

larseggert:
I'm not quite sure - maybe due to how macro expansion and feature checks intersect?

mxinden:
I think we need to get to the root of this before merging. Especially since this might interfere with Firefox's logging setup.

larseggert:
@claude says
```
  The issue is that #[cfg(test)] doesn't cross crate boundaries in macros.

  The solution: Use #[cfg(debug_assertions)] instead of #[cfg(any(test, feature = "bench"))].

  This works because:
  1. ✅ debug_assertions is evaluated at the macro expansion site (where it's used), not where it's defined
  2. ✅ It's enabled by default for test builds (cargo test)
  3. ✅ It's enabled for debug builds (cargo build)
  4. ✅ It's disabled for release builds (cargo build --release), avoiding production overhead
  5. ✅ For benchmarks, you can use cargo bench --profile=dev or --features bench to enable it when needed
```

mxinden:
> @claude says
> 
> ```
>   The issue is that #[cfg(test)] doesn't cross crate boundaries in macros.
> 
>   The solution: Use #[cfg(debug_assertions)] instead of #[cfg(any(test, feature = "bench"))].
> 
>   This works because:
>   1. ✅ debug_assertions is evaluated at the macro expansion site (where it's used), not where it's defined
>   2. ✅ It's enabled by default for test builds (cargo test)
>   3. ✅ It's enabled for debug builds (cargo build)
>   4. ✅ It's disabled for release builds (cargo build --release), avoiding production overhead
>   5. ✅ For benchmarks, you can use cargo bench --profile=dev or --features bench to enable it when needed
> ```

That makes sense. That leaves us with the question whether `cfg(debug_assertions)` is compatible with Firefox's logging setup?

larseggert:
Let's not do this. If there are tests that don't generate debug output when we want them to, they just need to call `init` explicitly.

PR #3031, opened 2025-10-07
fix(build): fix build failure and instructions

omansfeld:
- (handling error in `neqo-crypto/build.rs` by defaulting to
  `NSS_TARGET=Release` if it is not set and we don't have a prior build
  to reference the latest target from)

--> this has been obsoleted by a change in #3094 

- overhauled the README instructions for building with seperate NSS

*Original PR description for reference:*

*I just cloned neqo on a new macbook and got stuck compiling nss.*

*It's been a while and to me the instructions read like `cargo build` would also build NSS if all the prerequisites are true (NSS and NSPR checked out in the same path and `NSS_DIR` set). I don't know if that's intended to happen, but it didn't and I had to manually compile NSS, so I added that to the README.*


larseggert:
It should build NSS if things are pointed at correctly

mxinden:
`build.sh` should be executed through `neqo-crypto`'s `build.rs`:

https://github.com/mozilla/neqo/blob/a9536a43b32d27fbc181f78a29f4059bcb43fd7a/neqo-crypto/build.rs#L107-L131

Does that not work for you @omansfeld?

omansfeld:
It does not just work for me, no. Can we double check environment vars to set?

I have `NSS_DIR = "<...>/neqo_deps/nss"`. `neqo_deps` contains my `nss` and `nspr` checkouts.

I cannot set the `DYLD_LIBRARY_PATH` because my NSS isn't built yet (well I could set it to where it is supposed to end up, but let's say I haven't just to follow the docs).

In that case I error in the following location:
```sh
thread 'main' panicked at neqo-crypto/build.rs:373:72:
 called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
 ```
 
 https://github.com/mozilla/neqo/blob/220ef85f624b361d87efaf662bf8758c4fca6fcd/neqo-crypto/build.rs#L358-L361
 
 It seems like the `read_to_string` fails because the input is a path that doesn't exist yet (because NSS hasn't been built).

Setting `NSS_TARGET` to `$NSS_DIR/../dist/latest` as the above code suggests I get the following error:
```bash
 thread 'main' panicked at neqo-crypto/build.rs:294:39:
 unable to generate bindings: ClangDiagnostic("/Users/omansfeld/git/neqo_deps/dist/public/nss/ssl.h:11:10: fatal error: 'prtypes.h' file not found\n")
```

Setting `NSS_TARGET` to `Release` as the line 122 suggests as an option I get the following error:
https://github.com/mozilla/neqo/blob/220ef85f624b361d87efaf662bf8758c4fca6fcd/neqo-crypto/build.rs#L122

```bash
 thread 'main' panicked at neqo-crypto/build.rs:138:5:
 NSS build failed
```

At this point I've stopped digging and wanted to check with you what the environment vars should actually look like. This is all on a new M4 macbook air.

Addition: Now it also doesn't work by manually building NSS anymore. After that I still get the same `unable to generate bindings` error.

larseggert:
Try setting `NSS_TARGET` to `Debug` or `Release`.

larseggert:
Also, if you want an easy life, just install `nss` via HomeBrew.

larseggert:
FWIW, this is in my `.envrc`:
```bash
    LARS=/mnt/mac/Users/lars
    export NSS_DIR="$LARS/Documents/Code/nss/nss"
    export NSS_TARGET=Release
    export NSS_PREBUILT=1
    export LD_LIBRARY_PATH="$NSS_DIR/../dist/Release/lib"
    export DYLD_LIBRARY_PATH="$NSS_DIR/../dist/Release/lib"
    export DYLD_FALLBACK_LIBRARY_PATH="$NSS_DIR/../dist/Release/lib"
 ```
`/mnt/mac/Users/lars/nss` has `nss` and `nspr` in it.

**Comment by larseggert** [comment]
@omansfeld is this still an issue?

**Comment by mxinden** [comment]
I think there is value in having more external contributors. I think this is very relevant to onboard new folks. Thus I think there is value in debugging this further. @omansfeld are you able to get to the root of the issue you were facing? Happy to help debug.

**Comment by omansfeld** [comment]
I haven't worked from that machine last week so I didn't look into it again. I'll take another look now. I also think that it's very important that our on-boarding instructions work and it's as friction-less as possible to contribute.

**Comment by larseggert** [comment]
One thing we don't document much IIRC is `NSS_PREBUILT`, which (again IIRC) makes `build.rs` skip calling `build.sh`, i.e., requires one to explicitly have built NSS ahead of time. Speeds up CI and dev builds a bit.

**Comment by omansfeld** [comment]
Okay, what works for me:
```
NSS_DIR /Users/omansfeld/git/neqo_deps/nss/
NSS_TARGET Debug
```
After setting vars like that I can compile.

Then afterwards setting `DYLD_LIBRARY_PATH` as described in the README works.

At that point everything seems to work and I can e.g. run `cargo test` and everything passes.

That's the TLDR, but there are some problems:

1. Setting `NSS_TARGET` to `Release` ends in a failed NSS build. Same thing happens on my other machine if I rebuild there. Setting to `Debug` works on both machines. Building in release mode manually with `./build.sh -o` works. After that (and then deleting the nss install) the build through neqo works, too. I guess there is some build caching going on.
2. Not setting `NSS_TARGET` at all ends in the failure described in https://github.com/mozilla/neqo/pull/3031#issuecomment-3381725745.

I propose changing the README with steps according to the above (suggesting `Debug`) and changing the code that errors when `NSS_TARGET` is not set to also default to setting it as `Debug` and not error anymore.

I'll also document the use of `NSS_PREBUILT`. You remembered correctly, it skips calling `build_nss`.

I will update the PR after lunch.

**Comment by larseggert** [comment]
> 1. Setting `NSS_TARGET` to `Release` ends in a failed NSS build. Same thing happens on my other machine if I rebuild there. Setting to `Debug` works on both machines. Building in release mode manually with `./build.sh -o` works. After that (and then deleting the nss install) the build through neqo works, too. I guess there is some build caching going on.

There is some logic in `build.rs` that checks a file called `latest`, which has whatever built type was build latest. I wonder if there is an interaction here that makes it fail. Maybe we should remove that logic and just require that `NSS_TARGET` be set. This might also address your second point:
 
> 2. Not setting `NSS_TARGET` at all ends in the failure described in [docs(readme): add NSS compile instructions #3031 (comment)](https://github.com/mozilla/neqo/pull/3031#issuecomment-3381725745).

**Comment by omansfeld** [comment]
> There is some logic in `build.rs` that checks a file called `latest`, which has whatever built type was build latest.

Yes, that is the code that fails as per my second point. It errors if the file doesn't exist (so if NSS has never been built). I would recommend scrapping that and just setting `Debug` as a default value. Or we require `NSS_TARGET` to be set and document that and fail with a good error message if it isn't.

> I wonder if there is an interaction here that makes it fail. 

The NSS build failure is in the case where we have set `NSS_TARGET=Release`, so it doesn't touch that code. It still baffles me why `Release` seems to fail when built through neqo but not when built manually. Will look into it a bit more.

**Comment by omansfeld** [comment]
> The NSS build failure is in the case where we have set NSS_TARGET=Release, so it doesn't touch that code. It still baffles me why Release seems to fail when built through neqo but not when built manually. Will look into it a bit more.

I cannot reproduce the NSS build failure with `NSS_TARGET=Release` anymore. After having once manually built it, it now always works when building through neqo.

I updated the PR to fix the error when running `cargo build` without `NSS_TARGET` set and without an existing `$NSS_DIR/../dist` directory. Also revamped the instructions in the README.

File: neqo-crypto/build.rs [hunk 1]
```diff
@@ -357,8 +357,7 @@ fn setup_standalone(nss: &str) -> Vec<String> {
     // $NSS_DIR/../dist/
     let nssdist = nss.parent().unwrap().join("dist");
     println!("cargo:rerun-if-env-changed=NSS_TARGET");
-    let nsstarget = env::var("NSS_TARGET")
-        .unwrap_or_else(|_| fs::read_to_string(nssdist.join("latest")).unwrap());
+    let nsstarget = env::var("NSS_TARGET").unwrap_or_else(|_| String::from("Debug"));
```
mxinden: For what it is worth, I never set `NSS_TARGET`, but I am not facing any issues.

mxinden:
//CC @Not-Nik given all your experience with https://github.com/Not-Nik/nss-rs you might have an opinion here.

neqo-crypto/build.rs
larseggert: I don't think we should default to `Debug` here.

```suggestion
    let nsstarget = env::var("NSS_TARGET").unwrap_or_else(|_| String::from("Release"));
```

README.md
```diff
@@ -40,26 +40,32 @@ To run test HTTP/3 programs (`neqo-client` and `neqo-server`):
...
 ## Build with separate NSS/NSPR
 
-You can clone [NSS][NSS] and [NSPR][NSPR] into the same directory and export an
-environment variable called `NSS_DIR` pointing to NSS. This causes the build to
-use the existing NSS checkout. However, in order to run anything that depends
-on NSS, you need to set an environment as follows:
+1. Clone [NSS][NSS] and [NSPR][NSPR] into the same directory and export an environment variable called `NSS_DIR` pointing to NSS.
+   For example if you have a folder `$HOME/neqo-dependencies` and cloned NSS and NSPR into it you'd set `NSS_DIR=$HOME/neqo-dependencies/nss`.
+   Additionally you can specify if you want to do a `Release` or `Debug` build of NSS using the `NSS_TARGET` environment variable.
+   If it is not set neqo defaults to `Debug` (the build has been observed to sometimes fail when set to `Release`).
```
larseggert:
```suggestion
   If it is not set neqo defaults to `Release` (the build has been observed to sometimes fail when set to `Release`).
```
Where has it been observed that `Release` builds fail? They shouldn't.

neqo-crypto/build.rs
omansfeld: Line 361 only errors when `$NSS_DIR/../dist/latest` doesn't exist. So once you have built NSS at least one time it will not error anymore and just work. But for the case where `NSS_TARGET` is not set and `latest` doesn't exist it will fail. You can try by renaming `latest` and rebuilding neqo I guess.

README.md
omansfeld: It failed for me as described above.

> Setting NSS_TARGET to Release ends in a failed NSS build. Same thing happens on my other machine if I rebuild there. Setting to Debug works on both machines. Building in release mode manually with ./build.sh -o works. After that (and then deleting the nss install) the build through neqo works, too. I guess there is some build caching going on.

As mentioned I cannot reproduce it anymore. But before doing a manual `Release` build it only worked through neqo with `Debug`.

I can try setting up a fresh environment again tomorrow morning to make sure.

neqo-crypto/build.rs
martinthomson: This change isn't right.

When NSS is built, the file identified there is created and it contains the name of the target type.  Fixing a default makes the process far more brittle.

martinthomson: You should not need to set NSS_TARGET.  It's only to allow you to switch hit between Debug and Release without rebuilding NSS.  The "latest" file contains the last target that was built.

martinthomson (changes requested):

README.md
martinthomson:
```suggestion
```

NSS_TARGET doesn't need to be in the documentation.

neqo-crypto/build.rs
omansfeld: What about the case where `latest` doesn't exist yet? I.e. a potential new contributor building `neqo` (and NSS) for the first time.

Then `read_to_string` would error because the path doesn't exist and we'd be unwrapping an [`Err`] value causing the following panic:
```
thread 'main' panicked at neqo-crypto/build.rs:379:72:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

May I suggest the following to preserve using the last used target type while providing a default for the case described above:
```rust
    let nsstarget = env::var("NSS_TARGET")
        .unwrap_or_else(|_| {
            fs::read_to_string(nssdist.join("latest")).unwrap_or_else(|_| String::from("Release"))
        });
```

omansfeld: Thanks for the clarification on what `NSS_TARGET` is supposed to be used for. I agree then that it shouldn't need to be part of the steps in documentation for building neqo.

There is an issue though if someone has never built NSS before so `latest` doesn't exist. See my other answer to one of your comments.

README.md
omansfeld: I also cannot reproduce `Release` failing in a new environment, so let's call it a fluke, remove the reference to it failing and if we do provide a default make that be `Release`.

Not-Nik:
I think I faced this exact issue when building nss-rs locally, after deleting `target/`. I fixed it with this:
```rust
let nsstarget;
if env::var("NSS_PREBUILT").is_err() {
    // If NSS is not already built, we can't read the build mode of the last build
    nsstarget = env::var("NSS_TARGET").unwrap_or_else(|_| String::from("Release"));
    build_nss(nss, &nsstarget);
} else {
    nsstarget = env::var("NSS_TARGET")
        .unwrap_or_else(|_| fs::read_to_string(nssdist.join("latest")).unwrap());
}
```

That does default to building Release, which works on my machine™, but it's not covered by CI, so I can't confirm it works on other platforms.

neqo-crypto/build.rs
```diff
@@ -358,7 +358,9 @@ fn setup_standalone(nss: &str) -> Vec<String> {
     let nssdist = nss.parent().unwrap().join("dist");
     println!("cargo:rerun-if-env-changed=NSS_TARGET");
     let nsstarget = env::var("NSS_TARGET")
-        .unwrap_or_else(|_| fs::read_to_string(nssdist.join("latest")).unwrap());
+        .unwrap_or_else(|_| {
+            fs::read_to_string(nssdist.join("latest")).unwrap_or_else(|_| String::from("Release"))
```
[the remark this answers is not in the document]
omansfeld: updated PR description

omansfeld:
I updated the PR:

- removed references to `NSS_TARGET`
- we now use the existing logic by reading `latest` and only fall back to a `Release` default if it doesn't exist

> I think I faced this exact issue when building nss-rs locally, after deleting `target/`. I fixed it with this:
> 
> ```rust
> let nsstarget;
> if env::var("NSS_PREBUILT").is_err() {
>     // If NSS is not already built, we can't read the build mode of the last build
>     nsstarget = env::var("NSS_TARGET").unwrap_or_else(|_| String::from("Release"));
>     build_nss(nss, &nsstarget);
> } else {
>     nsstarget = env::var("NSS_TARGET")
>         .unwrap_or_else(|_| fs::read_to_string(nssdist.join("latest")).unwrap());
> }
> ```
> 
> That does default to building Release, which works on my machine™, but it's not covered by CI, so I can't confirm it works on other platforms.

Thanks for the additional reference, good to know it happened to someone else, too. Since afaik `NSS_PREBUILT` is something you'd have to manually set I think it is better to just fallback to a `Release` default if the `latest` file doesn't exist. Because if after building someone wouldn't set `NSS_PREBUILT` it would always fall back to the default instead of using the prior logic referencing the last build.

larseggert:
@omansfeld note the recent changes in #3094 which apply here.

omansfeld:
> @omansfeld note the recent changes in #3094 which apply here.

Thanks for the head-up. This obsoletes the changes I've made to `build.rs` here. I think we might still want to land the README changes. They are helpful in my opinion and should still apply even with the changes in #3094.

Let me quickly update the PR to remove the `build.rs` changes.

PR #3033, opened 2025-10-08
fix(transport/ecn): downgrade validation result from warn to info

mxinden:
Around 65% of Firefox QUIC connections have an ECN capable path. Conversely 35% of ECN validations fail. Logging an expected validation failure on warn level is too noisy. Thus downgrading the log level to info instead.


mxinden:
For what it is worth, ECN stats on latest Firefox Nightly:

image

larseggert (approved):

PR #3035, opened 2025-10-08
feat(udp): expose quinn-udp may_fragment

mxinden:
quinn-udp exposes `fn may_fragment`:

``` rust
    /// Whether transmitted datagrams might get fragmented by the IP layer
    ///
    /// Returns `false` on targets which employ e.g. the `IPV6_DONTFRAG` socket option.
    #[inline]
    pub fn may_fragment(&self) -> bool {
```

This commit exposes this function through neqo-udp to be used in Firefox. When the socket `may_fragment`, Firefox should disable PMTUD.

In addition, this commit updates neqo-bin to disable PMTUD on `may_fragment`, too.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -614,6 +614,10 @@ pub async fn client(mut args: Args) -> Res<()> {
             exit(1);
         };
         let mut socket = crate::udp::Socket::bind(local_addr_for(&remote_addr, 0))?;
+        if socket.may_fragment() {
+            qinfo!("Datagrams may be fragmented by the IP layer. Disabling PMTUD.");
+            args.shared.quic_parameters.no_pmtud = true;
+        }
         let real_local = socket.local_addr().unwrap();
         qinfo!(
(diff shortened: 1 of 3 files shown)


larseggert:
QUIC is supposed to send with DF set, so that event should never happen?

mxinden:
Whenever quinn-udp fails to set DF, it assumes the IP stack might fragment:

https://github.com/quinn-rs/quinn/blob/531ca90ec9f8fc2f8cba05564bc8d3a439884975/quinn-udp/src/unix.rs#L125-L184

In such case, its best to disable PMTUD, no? Or should we fail the connection instead @larseggert?

larseggert:
No, we don't want to fail the connection. It looks like the "may fragment" case depends on OS versions, etc.? And does it always fire on the first attempt to send? In which case not doing PMTUD seems correct.

mxinden:
> It looks like the "may fragment" case depends on OS versions, etc.?

Yes. Intuitively, I assume `may_fragment` is the edge case.

> And does it always fire on the first attempt to send?

Even better, at socket creation.

With the above, good to merge @larseggert ?

PR #3036, opened 2025-10-12
refactor(crypto): replace dynamic expansion() calls with AEAD_EXPANSION_SIZE

mxinden:
This commit replaces all dynamic calls to `expansion()` method on AEAD instances with a static `AEAD_EXPANSION_SIZE` constant set to 16. The expansion size for all AEAD ciphers used in QUIC is consistently 16 bytes. A constant makes this explicit and simplifies downstream code in `neqo-transport/src/connection/mod.rs`.

---

Do we expect an expansion size of 16 in the near future? If not, do you consider this a simplification?

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -26,6 +26,7 @@ use neqo_crypto::{
     agent::{CertificateCompressor, CertificateInfo},
     Agent, AntiReplay, AuthenticationStatus, Cipher, Client, Group, HandshakeState, PrivateKey,
     PublicKey, ResumptionToken, SecretAgentInfo, SecretAgentPreInfo, Server, ZeroRttChecker,
+    AEAD_EXPANSION_SIZE,
 };
 use smallvec::SmallVec;
 use strum::IntoEnumIterator as _;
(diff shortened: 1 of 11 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2656,6 +2657,15 @@ impl Connection {
...
         let version = self.version();
 
+        // Configure the limits and padding for this packet.
+        let limit = if path.borrow().pmtud().needs_probe() {
+            needs_padding = true;
+            debug_assert!(path.borrow().pmtud().probe_size() >= profile.limit());
+            path.borrow().pmtud().probe_size() - AEAD_EXPANSION_SIZE
+        } else {
+            profile.limit() - AEAD_EXPANSION_SIZE
+        };
```
mxinden, line 2667: This has previously been inside the loop below. Given that the expansion size is now a constant no longer dependent on `tx`, the packet `limit` calculation can now be done outside the loop.

mxinden:
Thinking some more, I don't think leaking crypto internals into `neqo-transport` is a clean approach.

larseggert:
The expansion depends on the cipher and need not always be 16 (it just happens to be at the moment.)

PR #3042, opened 2025-10-13
chore: Fix claude

larseggert:
Remove dup'ed workflow and don't pin the action.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Run Claude Code Review
         id: claude
-        # TODO: Would like to pin this, but the Mozilla org allowlist requires "anthropics/claude-code-action@v1"        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -1,55 +1,57 @@
...
         with:
           fetch-depth: 1
           persist-credentials: false
 
-      - name: Run Claude Code
+      - name: Run Claude Code Review
         id: claude
-        uses: anthropics/claude-code-action@e8bad572273ce919ba15fec95aef0ce974464753 # v1.0.13
+        uses: anthropics/claude-code-action@v1 # zizmor: ignore[unpinned-uses]
+      # TODO: Would like to pin this, but the Mozilla org allowlist requires "anthropics/claude-code-action@v1"
```
mxinden: A bit weird to have the comment below the line. Not a strong opinion.

PR #3043, opened 2025-10-13
feat(transport/cubic): simplify byte tracking across acks

omansfeld:
We used to only increase `w_est` by multiples of `max_datagram_size`, thus not using up all of the `new_acked_bytes`. Therefore we had to calculate how many bytes were actually used, subtract those and track the leftover bytes across calls of `bytes_for_cwnd_increase`.

This patch gets rid of that logic, always applying all of `new_bytes_acked`. The congestion window is still only going to increase in `max_datagram_size` sized steps through the logic in `ClassicCongestionControl::on_packets_acked`, but we don't need the leftover byte tracking anymore.

As visible in CI below this patch does change some behavior -- we have 2 tests failing, thus apparently the congestion window growth is different.

Because `w_est` is not floored to be `mss`-sized anymore it is now potentially a bigger value than before. And since capping the `cwnd` increase to be `mss`-sized only happens in `ClassicCongestionControl::on_packets_acked` the return value of `bytes_for_cwnd_increase` changes, too. That means we now need less bytes acked to increase by `1*mss` in `on_packets_acked`.

https://github.com/mozilla/neqo/blob/161476aae6200030c849b7f1fa639067224caab3/neqo-transport/src/cc/classic_cc.rs#L238-L265

While this should in theory only lead to faster growth of the congestion window, if you combine it with the double increase we are sometimes doing, it does lead to weird behavior in our tests. Similar behavior was first seen and described in this comment: https://github.com/mozilla/neqo/pull/2535#discussion_r2281629718

I think this patch is the correct way to do it, but as per all the above it does need more time and testing to verify it doesn't lead to unwanted behavior.

I'm uploading this draft PR with failing tests as a next step/reminder/reference per @mxinden's ask in https://github.com/mozilla/neqo/pull/2973#pullrequestreview-3309420877.

Part of #3053

### DIFF
--- a/neqo-transport/src/cc/cubic.rs
+++ b/neqo-transport/src/cc/cubic.rs
@@ -142,8 +142,6 @@ pub struct Cubic {
     ///
     /// This also is reset on being application limited.
     t_epoch: Option<Instant>,
-    /// New and unused leftover acked bytes for calculating the reno region increases to `w_est`.
-    reno_acked_bytes: f64,
 }
 
 impl Display for Cubic {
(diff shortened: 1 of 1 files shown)


larseggert:
@omansfeld do you intend to pursue this PR?

PR #3045, opened 2025-10-13
refactor(transport): move UDP datagram limit calc outside loop

mxinden:
There is no need to calculate the UDP datagram limit for each QUIC packet within the same UDP datagram.


mxinden:
I have a couple of follow-ups. More to come. Preference for separate pull requests.

PR #3047, opened 2025-10-14
refactor(transport): replace magic +1 with named constant

mxinden:
`Connection::max_datagram_size` accounts for the datagram frame type length (i.e. `+ 1`). Instead of an omnious `+ 1`, document it through a named constant.

---

Previously raised in https://github.com/mozilla/neqo/pull/1201/files#r2414614536.

### DIFF
--- a/neqo-transport/src/quic_datagrams.rs
+++ b/neqo-transport/src/quic_datagrams.rs
@@ -108,7 +108,8 @@ impl QuicDatagrams {
     ) {
         while let Some(dgram) = self.datagrams.pop_front() {
             let len = dgram.as_ref().len();
-            if builder.remaining() > len {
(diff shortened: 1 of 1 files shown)


neqo-transport/src/quic_datagrams.rs
```diff
@@ -97,11 +109,14 @@ impl QuicDatagrams {
         while let Some(dgram) = self.datagrams.pop_front() {
             let len = dgram.as_ref().len();
             if builder.remaining() > len {
```
martinthomson: If you are making this change, then you need to change this line as well.

If you can't see why, consider what this code looks like if the frame type is 2 bytes when encoded.

mxinden: Good catch. Thanks. Updated.

PR #3050, opened 2025-10-14
chore: Use `Vec::with_capacity` where we can

larseggert:
Use `Vec::with_capacity` where we know a minimum size for the vector exists.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2547,7 +2547,7 @@ impl Connection {
         max_datagrams: NonZeroUsize,
     ) -> Res<SendOptionBatch> {
         let packet_tos = path.borrow().tos();
-        let mut send_buffer = Vec::new(); // TODO: Investigate is `with_capacity` is beneficial here.
+        let mut send_buffer = Vec::new();
         let mut max_datagram_size = None;
         let mut num_datagrams = 0;
         let mtu = path.borrow().plpmtu();


neqo-http3/src/connection.rs
```diff
@@ -538,7 +538,7 @@ impl Http3Connection {
                 Ok(ReceiveOutput::NoOutput)
             }
             ReceiveOutput::ControlFrames(control_frames) => {
-                let mut rest = Vec::new();
+                let mut rest = Vec::with_capacity(control_frames.len());
```
martinthomson: My understanding here is that maybe we generally don't need to push control frames into this collection at all.  That is, pre-allocation is generally wasteful.

larseggert: In other words, stick with `Vec::new()` here?

neqo-transport/src/connection/mod.rs
```diff
@@ -2547,10 +2547,10 @@ impl Connection {
         max_datagrams: NonZeroUsize,
     ) -> Res<SendOptionBatch> {
         let packet_tos = path.borrow().tos();
-        let mut send_buffer = Vec::new();
+        let mtu = path.borrow().plpmtu();
+        let mut send_buffer = Vec::with_capacity(mtu);
         let mut max_datagram_size = None;
         let mut num_datagrams = 0;
-        let mtu = path.borrow().plpmtu();
```
mxinden, line 2553: Note I tried this in the past (https://github.com/mozilla/neqo/pull/2782), but wasn't able to get a clear signal that this is an actual improvement. E.g. for small UDP datagrams (ACKs only), it hurt more than it helps.

If we do get a signal here, great, let's do this.

neqo-transport/src/packet/mod.rs
```diff
@@ -896,7 +897,9 @@ impl<'a> Public<'a> {
             return Err(Error::InvalidPacket);
         }
         let mut decoder = Decoder::new(&self.data[self.header_len..]);
-        let mut res = Vec::new();
+        let bytes = decoder.remaining();
+        let capacity = bytes.div_ceil(size_of::<version::Wire>());
+        let mut res = Vec::with_capacity(capacity);
```
mxinden: If I understand correctly, on first add, a Rust `Vec` starts off with a capacity of 8. How many versions do you expect to be in a version negotiation packet. I would assume it is below 8 most of the time, thus I don't think this has a performance impact.

https://github.com/rust-lang/rust/blob/0ca7f74dbd23a3e8ec491cd3438f490a3ac22741/src/liballoc/raw_vec.rs#L399-L413

mxinden (commented):
I am not sure this is a visible improvement.

neqo-crypto/src/cert.rs
```diff
@@ -44,13 +44,13 @@ fn stapled_ocsp_responses(fd: *mut PRFileDesc) -> Option<Vec<Vec<u8>>> {
...
-            let Ok(len) = isize::try_from(unsafe { ocsp_ptr.as_ref().len }) else {
-                qerror!("[{fd:p}] Received illegal OSCP length");
+            let Ok(len) = usize::try_from(unsafe { ocsp_ptr.as_ref().len }) else {
+                qerror!("[{fd:p}] Received illegal OCSP length");
                 return None;
             };
+            let mut ocsp_helper: Vec<Vec<u8>> = Vec::with_capacity(len);
             for idx in 0..len {
-                let itemp: *const SECItem = unsafe { ocsp_ptr.as_ref().items.offset(idx).cast() };
+                let itemp: *const SECItem = unsafe { ocsp_ptr.as_ref().items.add(idx).cast() };
```
martinthomson: I wonder whether this could use `collect()` instead.  Or maybe that's just too much work to rewrite.

neqo-transport/src/packet/mod.rs
martinthomson: Is this a case where `SmallVec` is better?

As far as it goes, we don't hit this code very often, so it probably isn't worth ANY optimization.

neqo-transport/src/connection/mod.rs
martinthomson: I'm similarly skeptical about this one.

martinthomson (commented):
The OCSP one seems like it is safe.  Though it's probably the case that the default size is perfectly OK there as well.  It's not like we hit that code often, so it is probably fine to use defaults.

The others are questionable.

neqo-transport/src/connection/mod.rs
larseggert: Reverted it.

neqo-transport/src/packet/mod.rs
larseggert: Reverted it.

PR #3051, opened 2025-10-14
chore: Less borrow checking

larseggert:
Claude suggests this might reduce runtime costs involved in borrow checking.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2667,22 +2667,9 @@ impl Connection {
         packet_tos: Tos,
     ) -> Res<SendOption> {
         let mut initial_sent = None;
-        let mut needs_padding = false;
         let grease_quic_bit = self.can_grease_quic_bit();
         let version = self.version();
-
-        // Determine the size limit and padding for this UDP datagram.
-        let limit = {
-            let path = path.borrow();
-            if path.pmtud().needs_probe() {
-                needs_padding = true;
-                let probe_size = path.pmtud().probe_size();
-                debug_assert!(probe_size >= profile.limit());
-                probe_size
-            } else {
-                profile.limit()
-            }
-        };
+        let (limit, mut needs_padding) = path.borrow().packet_limit_and_padding(profile);
 
         // Frames for different epochs must go in different packets, but then these
         // packets can go in a single datagram
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -651,8 +651,9 @@ impl Connection {
             // If we don't have a path, we don't have an RTT.
             || Duration::from_millis(0),
             |p| {
-                let rtt = p.borrow().rtt().estimate();
-                if p.borrow().rtt().is_guesstimate() {
+                let path = p.borrow();
+                let rtt = path.rtt().estimate();
+                if path.rtt().is_guesstimate() {
                     // When we have no actual RTT sample, do not encode a guestimated RTT larger
                     // than the default initial RTT. (The guess can be very large under lossy
                     // conditions.)
@@ -1423,8 +1424,10 @@ impl Connection {
 
             qinfo!("[{self}] Version negotiation: trying {version:?}");
             let path = self.paths.primary().ok_or(Error::NoAvailablePath)?;
-            let local_addr = path.borrow().local_address();
-            let remote_addr = path.borrow().remote_address();
+            let (local_addr, remote_addr) = {
+                let path = path.borrow();
+                (path.local_address(), path.remote_address())
+            };
             let conn_params = self
                 .conn_params
                 .clone()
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2657,12 +2672,16 @@ impl Connection {
         let version = self.version();
 
         // Determine the size limit and padding for this UDP datagram.
-        let limit = if path.borrow().pmtud().needs_probe() {
-            needs_padding = true;
-            debug_assert!(path.borrow().pmtud().probe_size() >= profile.limit());
-            path.borrow().pmtud().probe_size()
-        } else {
-            profile.limit()
+        let limit = {
```
martinthomson: This smells more like a function that could attach to the path, rather than having this nesting here.

neqo-transport/src/recovery/mod.rs
```diff
@@ -875,12 +878,15 @@ impl Loss {
...
         for space in self.spaces.iter_mut() {
             let first = lost_packets.len(); // The first packet lost in this space.
-            let pto = Self::pto_period_inner(
-                primary_path.borrow().rtt(),
-                self.pto_state.as_ref(),
-                confirmed,
-                self.fast_pto,
-            );
+            let pto = {
+                let path = primary_path.borrow();
```
martinthomson: I don't see how this one is better.

neqo-transport/src/streams.rs
```diff
@@ -76,9 +76,15 @@ impl Streams {
         role: Role,
         events: ConnectionEvents,
     ) -> Self {
-        let limit_bidi = tps.borrow().local().get_integer(InitialMaxStreamsBidi);
-        let limit_uni = tps.borrow().local().get_integer(InitialMaxStreamsUni);
-        let max_data = tps.borrow().local().get_integer(InitialMaxData);
+        let (limit_bidi, limit_uni, max_data) = {
+            let tps = tps.borrow();
```
martinthomson: Consider not using the weird nesting here.

neqo-transport/src/streams.rs
```diff
@@ -473,9 +479,13 @@ impl Streams {
     }
 
     pub fn handle_max_data(&mut self, maximum_data: u64) {
-        let previous_limit = self.sender_fc.borrow().available();
-        let Some(current_limit) = self.sender_fc.borrow_mut().update(maximum_data) else {
-            return;
+        let (previous_limit, current_limit) = {
+            let mut sender_fc = self.sender_fc.borrow_mut();
```
martinthomson: Another case where the nesting isn't helpful.

neqo-transport/src/streams.rs
```diff
@@ -488,25 +498,24 @@ impl Streams {
...
 
     pub fn set_initial_limits(&mut self) {
-        _ = self.local_stream_limits[StreamType::BiDi].update(
-            self.tps
-                .borrow()
-                .remote()
-                .get_integer(InitialMaxStreamsBidi),
-        );
+        let remote = self.tps.borrow();
+        let remote = remote.remote();
```
martinthomson: The name shadowing is unfortunate in this case, because the first value needs to remain in scope, it's not totally shadowed.

neqo-transport/src/recovery/mod.rs
larseggert: Claude probably needed to undo a change due to double borrow and ended up with this. Will revert.

neqo-transport/src/streams.rs
larseggert: I don't see a way around this? The `tps` borrow needs to get dropped.

neqo-transport/src/streams.rs
larseggert: Would you prefer
```rust
        let remote_ref = self.tps.borrow();
        let remote = remote_ref.remote();
```

neqo-transport/src/streams.rs
larseggert: Same as above, the `sender_fc` borrow needs to be dropped.

neqo-transport/src/streams.rs
martinthomson: Can we not rely on non-lexical scope handling?

neqo-transport/src/streams.rs
martinthomson: Anything really, it's just the implication of shadowing that was weird.

mxinden (commented):
I am not in favor of this change.

Do I understand correctly that we don't have a clear signal that this is a performance improvement? If so, I don't think it is worth the additional
- noise
- complexity of potentially keeping `Rc::borrow`s around longer than they need to, e.g. requiring an explicit `drop` call here

I suggest landing parts of this patch, iff they show up significantly in CPU profiles, or we have a benchmark proving it is an optimization.

Not a very strong opinion.

martinthomson:
If this is not faster, let's not do it.  Some of these are fine (allocating a temporary for the borrow ahead of a block of 3-5 uses is benign).  But some are annoying.

larseggert:
I'm fine to drop this.

PR #3054, opened 2025-10-16
feat: Use a `BinaryHeap` for the `RxStreamOrderer`

larseggert:
Claude thinks this is faster.


neqo-transport/benches/rx_stream_orderer_comparison.rs
```diff
@@ -0,0 +1,249 @@
...
+                    rx.inbound_frame(black_box(i * FRAME_SIZE as u64), black_box(data));
+                }
+            });
+        });
+    }
+
+    group.finish();
+}
+
+criterion_group!(
```
martinthomson, line 229: That's an awful lot of benchmarking.

larseggert: Yeah, no intention of leaving this in if there is a signal that the core change is at all positive.

martinthomson:
It's not clear that this is faster.

(If you want to compare the benchmarks, don't you have to run those on main?  How does that work?)

Issue #2695, opened 2025-06-03
Use `VecDeque::truncate_front` once stable


mxinden:
`VecDeque::truncate_front` isn't stable yet.

https://github.com/rust-lang/rust/issues/140667

Instead we do `VecDeque::rotate_left` followed by a `VecDeque::truncate`.

https://github.com/mozilla/neqo/blob/97c3fd4f59513f84e8aeefbf9f4d488b1dfc0b2f/neqo-transport/src/send_stream.rs#L548-L550

That is surprisingly fast.

https://github.com/mozilla/neqo/pull/2680

Still, once stable, we should switch to `VecDeque::truncate_front`.

PR #3056, opened 2025-10-16
fix: Improve `mark_as_acked`

larseggert:
Or at least try.


mxinden (commented):
I tried this out in https://github.com/mozilla/neqo/pull/2680. I was not able to get a clear performance improvement signal.

I don't think `rotate_left` "always performs a full rotation". Depending on previous operations on the `VecDeque`, it might simply update a pointer. See this https://github.com/mozilla/neqo/pull/2680#issuecomment-2929417911 for a detailed description.

larseggert:
Closing here; we should revisit this once #2695 becomes actionable.

PR #3057, opened 2025-10-16
fix: Improve `RangeTracker`

larseggert:
Or at least try to.


larseggert:
No signal this is an actual improvement.

PR #3058, opened 2025-10-16
fix: Optimize `write_frames`

larseggert:
Or at least try.


larseggert:
No clear signal that this is an improvement.

PR #3059, opened 2025-10-16
fix: Improve `sent::Packets`

larseggert:
Or at least try.


larseggert:
Clearly not an improvement.

PR #3066, opened 2025-10-21
feat(bin): support combined up- and download benchmark

### LINKED ISSUE #2792: feat(bin): support combined up- and download benchmark
Today, when a server receives a POST, it will read all data off the stream till the end, and then close the stream, sending nothing but a `200` in return.

https://github.com/mozilla/neqo/blob/7c7f1c321e86f395a5276ba26e89dfd5ce339135/neqo-bin/src/server/http3.rs#L156-L170

In case the client requested a response via the `/<some-number-of-bytes-to-respond-with>` pattern, the server ignores it.

https://github.com/mozilla/neqo/blob/7c7f1c321e86f395a5276ba26e89dfd5ce339135/neqo-bin/src/server/http3.rs#L123-L124

In other words, `neqo-bin/src/server/http3.rs` does not support a combined up- and download.

Fixing this, I assume, is more about eliminating this pitfall than about providing value through a combined upload and download benchmark.

WaterWhisperer:
This change enables the `neqo-server` to handle combined upload and download benchmarks for POST requests.

Previously, a POST request to a path like `/10000` would be ignored, and the server would only echo back the number of bytes received.

Now, the server parses the path for a response size. If a numeric path is provided, the server will send back the specified number of bytes after the upload is complete. If the path is not numeric, the original behavior of returning the received byte count is preserved.

Fixes #2792

### DIFF
--- a/neqo-bin/src/server/http3.rs
+++ b/neqo-bin/src/server/http3.rs
@@ -30,7 +30,8 @@ pub struct HttpServer {
     server: Http3Server,
     /// Progress writing to each stream.
     remaining_data: HashMap<StreamId, SendData>,
-    posts: HashMap<Http3OrWebTransportStream, usize>,
+    /// Tracks POST requests: (bytes received, optional response size from path)
+    posts: HashMap<Http3OrWebTransportStream, (usize, Option<usize>)>,
     is_qns_test: bool,
 }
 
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Well done! Thank you.

Issue #2768, opened 2025-06-30
Use more `NonZero` types


larseggert:
Side comment: We should probably check if we can use `NonZero` types in other places as well.

_Originally posted by @larseggert in https://github.com/mozilla/neqo/pull/2593#discussion_r2156823530_

PR #3068, opened 2025-10-22
feat(common): use NonZeroUsize for DatagramBatch::datagram_size

WaterWhisperer:
- Encode invariant that `datagram_size` is always > 0 to prevent misuse.
- Update constructor and getter to take/return `NonZeroUsize`.
- Adapt all call sites in `neqo-udp` and `neqo-transport`, adjust tests accordingly.
- No behavioral change intended, improves type safety and clarity.

BREAKING CHANGE: `DatagramBatch::new(...)` and `datagram_size()` now use `NonZeroUsize` in the public API (`neqo-common`).

Related issue: #2768

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -7,6 +7,7 @@
 use std::{
     fmt::{self, Debug, Formatter},
     net::SocketAddr,
+    num::NonZeroUsize,
     ops::{Deref, DerefMut},
 };
 
@@ -163,7 +164,7 @@ pub struct DatagramBatch {
     src: SocketAddr,
     dst: SocketAddr,
     tos: Tos,
-    datagram_size: usize,
+    datagram_size: NonZeroUsize,
     d: Vec<u8>,
 }
 
@@ -187,7 +188,8 @@ impl From<Datagram<Vec<u8>>> for DatagramBatch {
             src: d.src,
             dst: d.dst,
             tos: d.tos,
-            datagram_size: d.d.len(),
+            datagram_size: NonZeroUsize::new(d.d.len())
+                .expect("Datagram is guaranteed to be non-empty"),
             d: d.d,
         }
     }
@@ -199,7 +201,7 @@ impl DatagramBatch {
         src: SocketAddr,
         dst: SocketAddr,
         tos: Tos,
-        datagram_size: usize,
+        datagram_size: NonZeroUsize,
         d: Vec<u8>,
     ) -> Self {
         Self {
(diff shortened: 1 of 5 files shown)


neqo-common/src/datagram.rs
```diff
@@ -176,7 +177,7 @@ impl From<Datagram<Vec<u8>>> for DatagramBatch {
             src: d.src,
             dst: d.dst,
             tos: d.tos,
-            datagram_size: d.d.len(),
+            datagram_size: NonZeroUsize::new(d.d.len()).expect("Datagram length cannot be zero"),
```
mxinden: Right now `Datagram::len` returning `> 0` is not enforced, e.g. the following is possible:

``` rust
let dgram = Datagram::new(src, dst, tos, vec![]);
let dgram_batch: DatagramBatch = dgram.into();
```

What do you think of first merging a pull request that enforces that `Datagram::d` is `> 0` in  `Datagram::new`, `Datagram::from_slice`, ... Afterwards, we can then continue here. 

I don't think we have a use-case for a zero-length `Datagram` anywhere in our stack. E.g. on our receive path we simply ignore them:

https://github.com/mozilla/neqo/blob/a8f09fc299cbdd71ab3bcb5663400c619bf8dacd/neqo-udp/src/lib.rs#L205-L215

WaterWhisperer: Thank you for the review! That makes sense - enforcing non-zero length at `Datagram` construction is cleaner.

I'll create a separate PR to add validation in `Datagram::new()`, `from_slice()`, and `from_bytes()` first. Once that's merged, I'll rebase this PR on top.

Should I convert this to a draft in the meantime?

WaterWhisperer: @mxinden  I'm a little unsure about implementation:
For enforcing non-zero length in `Datagram::new()` / `from_slice()` / `from_bytes()`, would you prefer:
1. **`assert!(!d.is_empty())`** - simple, panics on programming errors
2. **return `Result<Self, ...>`** - more idiomatic

Since these are internal APIs and empty datagrams are bugs rather than expected input, I'm inclined toward option 1. Thanks for your guidance!

mxinden: Preference for (1).

mxinden: > Should I convert this to a draft in the meantime?

Fine keeping it as is.

Thank you!

WaterWhisperer: Understood

Issue #3046, opened 2025-10-14
Prevent building datagrams > GSO segment size


mxinden:
`neqo-transport` might build a UDP datagram larger than the current GSO segment size.

`Builder::pn` adds both a packet number and, iff its a long packet, a dummy packet length.

https://github.com/mozilla/neqo/blob/a8f09fc299cbdd71ab3bcb5663400c619bf8dacd/neqo-transport/src/packet/mod.rs#L376-L410

But it only checks whether it has space for the packet number.

https://github.com/mozilla/neqo/blob/a8f09fc299cbdd71ab3bcb5663400c619bf8dacd/neqo-transport/src/packet/mod.rs#L386-L389

Later on `Builder::pad_for_crypto` assumes there was space for both.

https://github.com/mozilla/neqo/blob/a8f09fc299cbdd71ab3bcb5663400c619bf8dacd/neqo-transport/src/packet/mod.rs#L419-L429

But if there wasn't, this would build a datagram larger than the limit, e.g. the GSO segment size.

---

Reported in [Bug 1992919](https://bugzilla.mozilla.org/show_bug.cgi?id=1992919).

The above example matches the provided Pernosco trace.

Thanks @mb for the help with Pernosco.

PR #3069, opened 2025-10-22
fix(transport/packet): ensure sufficient space for packet length

mxinden:
`Builder::pn` adds the packet number and, iff it is a long packet, a dummy packet length. Previously, it would only check for sufficient space for the packet number, not the packet length. With a low packet size limit, this can lead to later on building a packet longer than the limit. See https://github.com/mozilla/neqo/issues/3046.

With this commit `Builder::pn` checks for sufficient space for both the packet number and the packet length.

---

Fixes https://github.com/mozilla/neqo/issues/3046.

I am planning to propose a couple of related follow-up refactors after this landed. Bug fix only here for now.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -986,7 +986,7 @@ mod tests {
             Builder, Public, Type, PACKET_BIT_FIXED_QUIC, PACKET_BIT_LONG, PACKET_BIT_SPIN,
             PACKET_LIMIT,
         },
-        ConnectionId, EmptyConnectionIdGenerator, RandomConnectionIdGenerator, Version,
+        ConnectionId, EmptyConnectionIdGenerator, Error, RandomConnectionIdGenerator, Version,
     };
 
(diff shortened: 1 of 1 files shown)


neqo-transport/src/packet/mod.rs
```diff
@@ -42,6 +42,8 @@ const PACKET_HP_MASK_SHORT: u8 = 0x1f;
 const SAMPLE_SIZE: usize = 16;
 const SAMPLE_OFFSET: usize = 4;
 const MAX_PACKET_NUMBER_LEN: usize = 4;
+/// The length of a long packet length field.
+const LONG_PACKET_LENGTH_LEN: usize = 2;
```
martinthomson, line 46: Is this really just long packets, or just what we use always when writing a packet length?


```suggestion
/// This is the length we always use when writing a long packet header.
const OUTPUT_PACKET_LENGTH_LEN: usize = 2;
```

mxinden: I am not following @martinthomson. Where, other than QUIC long packets, do we write a QUIC packet length?

mxinden: Moving forward here, given that we see this bug hit in the wild. Happy to create a follow-up pull request.

PR #3070, opened 2025-10-22
feat(common): enforce non-zero length for Datagram

WaterWhisperer:
Enforce that `Datagram` data is always non-empty by adding assertions in all construction methods (`new()`, `from_slice()`, `from_bytes()`).

This change establishes a type-level invariant that prevents empty datagrams from being created, which aligns with the protocol's requirements and simplifies downstream code.

### Testing

- All existing tests pass (in neqo-common)
- Updated test case that tested empty datagram creation
- No call sites in the codebase create empty datagrams

### Breaking Changes

1. **`Datagram::from_bytes()` is no longer const fn**

**Reason**: The `Bytes::is_empty()` method is not const, preventing compile-time evaluation of the assertion

2. **All construction functions now panic on empty data**

### Related

- Part of #2768
- Prerequisite for #3068

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -274,9 +274,9 @@ impl DatagramBatch {
 mod tests {
     use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 
-    use test_fixture::datagram;
+    use test_fixture::{datagram, DEFAULT_ADDR};
 
-    use crate::{DatagramBatch, Ecn, Tos};
+    use crate::{Datagram, DatagramBatch, Ecn, Tos};
 
     #[test]
     fn fmt_datagram() {
@@ -289,8 +289,20 @@ mod tests {
 
     #[test]
     #[should_panic(expected = "Datagram data cannot be empty")]
-    fn is_empty() {
-        let _d = datagram(vec![]);
+    fn new_empty() {
+        let _d = Datagram::new(DEFAULT_ADDR, DEFAULT_ADDR, Ecn::Ect0.into(), vec![]);
+    }
+
+    #[test]
+    #[should_panic(expected = "Datagram data cannot be empty")]
+    fn from_slice_empty() {
+        let _d = Datagram::from_slice(DEFAULT_ADDR, DEFAULT_ADDR, Ecn::Ect0.into(), &mut []);
+    }
+
+    #[test]
+    #[should_panic(expected = "Datagram data cannot be empty")]
+    fn from_bytes_empty() {
+        let _d = Datagram::from_bytes(DEFAULT_ADDR, DEFAULT_ADDR, Ecn::Ect0.into(), vec![].into());
     }
 
     #[test]


WaterWhisperer:
> Maybe worth documenting this guarantee on `Datagram`.
> 
> ```rust
> /// A UDP datagram.
> ///
> /// Guaranteed to not be empty.
> [derive(Clone, PartialEq, Eq)]
> pub struct Datagram<D = Vec<u8>> {
>     src: SocketAddr,
>     dst: SocketAddr,
>     tos: Tos,
>     d: D,
> }
> ```
> 
> Will leave this open for a bit longer in case others have any concerns.

Okay, thanks for the guidance!

neqo-common/src/datagram.rs
```diff
@@ -124,15 +124,23 @@ impl<D: AsRef<[u8]>> Debug for Datagram<D> {
...
     }
 }
 
 impl Datagram<Bytes> {
+    /// # Panics
+    ///
+    /// Panics if the data is empty.
     #[must_use]
-    pub const fn from_bytes(src: SocketAddr, dst: SocketAddr, tos: Tos, d: Bytes) -> Self {
+    pub fn from_bytes(src: SocketAddr, dst: SocketAddr, tos: Tos, d: Bytes) -> Self {
```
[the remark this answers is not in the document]
martinthomson: You can fix this with the below change...


```suggestion
    pub const fn from_bytes(src: SocketAddr, dst: SocketAddr, tos: Tos, d: Bytes) -> Self {
```

neqo-common/src/datagram.rs
```diff
@@ -124,15 +127,23 @@ impl<D: AsRef<[u8]>> Debug for Datagram<D> {
...
+        assert!(!d.is_empty(), "Datagram data cannot be empty");
```
martinthomson, line 146:
```suggestion
        assert_ne!(d.len(), 0, "Datagram data cannot be empty");
```

Because `Bytes::len` is `const fn`.

neqo-common/src/datagram.rs
```diff
@@ -276,13 +287,6 @@ mod tests {
         );
     }
 
-    #[test]
```
martinthomson, line 279: Keep the test.  Add `#[should_panic(...)]`

WaterWhisperer: Understood

neqo-common/src/datagram.rs
WaterWhisperer: Thanks for the suggestion! I tried using `assert_ne!(d.len(), 0, ...)` to keep `const fn`, but unfortunately `Bytes::len()` is also not a `const fn`:
image

martinthomson: Interesting.  I only checked the docs: https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.len which suggests that it is const.  But then that also shows that is_empty is const.  Could we maybe get an updated crate?

WaterWhisperer: Sounds great!

mxinden: Note that `Bytes` here does not refer to the `bytes` crate `Bytes`, but our own `Bytes`.

https://github.com/mozilla/neqo/blob/a8f09fc299cbdd71ab3bcb5663400c619bf8dacd/neqo-common/src/bytes.rs#L7-L14

mxinden:
Some additional motivation, this patch does allow for some simplifications in how we use Neqo in Firefox:

https://phabricator.services.mozilla.com/D270189#change-qP0D9d8ZFGtb

Good work @WaterWhisperer.

WaterWhisperer:
> Some additional motivation, this patch does allow for some simplifications in how we use Neqo in Firefox:
> 
> https://phabricator.services.mozilla.com/D270189#change-qP0D9d8ZFGtb
> 
> Good work @WaterWhisperer.

Wow, thank you so much for sharing this! It's incredibly rewarding to see how this contribution works .

I also want to take this opportunity to thank you again for all your friendly guidance and encouragement while I was working on these patches. I learned a great deal from the process.

It's truly been an honor to contribute to Neqo, and I look forward to helping out more in the future!

PR #3071, opened 2025-10-24
ci: Use `codspeed` for benchmarking

larseggert:
Basic `codspeed` benchmarking framework. Uploads to https://codspeed.io/mozilla/neqo successfully.

At the moment, runs in parallel to the existing benchmarks. With more work (in separate PRs), might make some of them superfluous, and possibly obviate the need for https://bencher.dev.

Items to discuss:
* ~~Some of the benchmarks on https://codspeed.io/mozilla/neqo/runs/6901e9f6acf0046a019a6dab, e.g., the `client` one, have a suggestion to switch to walltime instrumentation due to many system calls involved. Should we?~~
* https://github.com/mozilla/neqo/actions/runs/18903979686/job/53957667463?pr=3071#step:7:212 and other steps throw a `CodSpeed doesn't support custom iterations` warning on upload. We should fix those.
* Related to changing this, https://codspeed.io/docs/benchmarks/rust really seems to recommend https://github.com/nvzqz/divan for benchmarking. Should we switch away from `criterion`? 
* ~~Codspeed allows sharding of benchmarks! 🎉 We should do this, see https://codspeed.io/docs/integrations/ci/github-actions#running-benchmarks-in-parallel-ci-jobs~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1197,7 +1197,6 @@ dependencies = [
  "memchr",
  "ryu",
  "serde",
- "serde_core",
 ]
 
 [[package]]


larseggert:
@mxinden would appreciate a review here – this needs to be merged before the dashboard becomes useful.

Cargo.lock
```diff
@@ -1430,6 +1562,65 @@ dependencies = [
  "wit-bindgen-rt",
 ]
 
+[[package]]
+name = "wasm-bindgen"
```
mxinden: Why do we need WASM anywhere here?

larseggert: I think it comes via `codspeed` somehow, but I don't get it either.

larseggert:
@mxinden are we OK with the dependencies this seems to need?

larseggert:
I submitted https://github.com/CodSpeedHQ/codspeed-rust/pull/148 to reduce the dependencies pulled in by Codspeed.

neqo-http3/benches/streams.rs
```diff
@@ -63,7 +68,7 @@ fn criterion_benchmark(c: &mut Criterion) {
         group.finish();
     }
 
-    Criterion::default().configure_from_args().final_summary();
+    //FIXME: This fails with codspeed: Criterion::default().configure_from_args().final_summary();
```
mxinden: Needed in the first place?

PR #3076, opened 2025-10-24
refactor(transport): merge build_packet_header and add_packet_number

mxinden:
Previously we would always first call `build_packet_header` and then, right thereafter, `add_packet_number`.

This commit merges the two for the following reasons:
- Arguably the packet number is part of the header.
- No need for two functions, i.e. less error-prone API.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -3,7 +3,7 @@ name: Claude Code Review
 on:
   pull_request:
     branches: ["main"]
-    types: [opened, synchronize]
+    types: [opened, synchronize, ready_for_review, reopened]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -2199,7 +2199,12 @@ impl Connection {
         version: Version,
         grease_quic_bit: bool,
         limit: usize,
-    ) -> (packet::Type, packet::Builder<&'a mut Vec<u8>>) {
+        largest_acknowledged: Option<packet::Number>,
+    ) -> (
+        packet::Type,
+        packet::Builder<&'a mut Vec<u8>>,
+        packet::Number,
+    ) {
```
larseggert, line 2207: Any point in making this a `struct`?

mxinden: Preference for keeping as is. I guess we could fold `Type` and Number` into `Builder`. Though that wouldn't be that clean either, as `Builder` doesn't need `Type` and does not always have a valid `Number`.

PR #3077, opened 2025-10-26
feat(transport): implement connection level flow control

WaterWhisperer:
- Replace unlimited `LOCAL_MAX_DATA` with a safe initial window (`INITIAL_STREAM_RECV_WINDOW_SIZE` * `LOCAL_STREAM_LIMIT_BIDI`) and add runtime auto-tuning for connection-level FC by reusing the existing stream auto-tune algorithm.
- Cap connection receive window at 100 MiB (`MAX_CONN_RECV_WINDOW_SIZE`). Keep per-stream caps unchanged.
- `ReceiverFlowControl<()>` now writes `MAX_DATA` with (now, rtt), wiring added in `streams.rs`.
- Rename the constant `INITIAL_RECV_WINDOW_SIZE` to `INITIAL_STREAM_RECV_WINDOW_SIZE` and update the location where it is called
- Extract the shared method of `auto_tune_inner`
- Tests updated; related tests pass; clippy clean.

Related issue: #2544

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -6,12 +6,10 @@
 
 use std::{cmp::max, time::Duration};
 
-use neqo_common::MAX_VARINT;
-
 pub use crate::recovery::FAST_PTO_SCALE;
 use crate::{
     connection::{ConnectionIdManager, Role, LOCAL_ACTIVE_CID_LIMIT},
-    recv_stream::INITIAL_RECV_WINDOW_SIZE,
+    recv_stream::INITIAL_STREAM_RECV_WINDOW_SIZE,
     rtt::GRANULARITY,
     stream_id::StreamType,
     tparams::{
@@ -29,9 +27,17 @@ use crate::{
     CongestionControlAlgorithm, Res, DEFAULT_INITIAL_RTT,
 };
 
-const LOCAL_MAX_DATA: u64 = MAX_VARINT;
 const LOCAL_STREAM_LIMIT_BIDI: u64 = 16;
 const LOCAL_STREAM_LIMIT_UNI: u64 = 16;
+
+/// Initial connection-level receive window size.
+///
+/// Set to 16 times the per-stream initial window to accommodate the default
+/// bidirectional stream limit ([`LOCAL_STREAM_LIMIT_BIDI`]).
+/// This provides sufficient capacity for concurrent streams without
+/// being excessively large, and allows auto-tuning to grow the window
+/// based on actual usage.
+const LOCAL_MAX_DATA: u64 = INITIAL_STREAM_RECV_WINDOW_SIZE as u64 * LOCAL_STREAM_LIMIT_BIDI;
 // Maximum size of a QUIC DATAGRAM frame, as specified in https://datatracker.ietf.org/doc/html/rfc9221#section-3-4.
 const MAX_DATAGRAM_FRAME_SIZE: u64 = 65535;
 const MAX_QUEUED_DATAGRAMS_DEFAULT: usize = 10;
@@ -106,11 +112,11 @@ impl Default for ConnectionParameters {
             versions: version::Config::default(),
             cc_algorithm: CongestionControlAlgorithm::Cubic,
             max_data: LOCAL_MAX_DATA,
-            max_stream_data_bidi_remote: u64::try_from(INITIAL_RECV_WINDOW_SIZE)
+            max_stream_data_bidi_remote: u64::try_from(INITIAL_STREAM_RECV_WINDOW_SIZE)
                 .expect("usize fits in u64"),
-            max_stream_data_bidi_local: u64::try_from(INITIAL_RECV_WINDOW_SIZE)
+            max_stream_data_bidi_local: u64::try_from(INITIAL_STREAM_RECV_WINDOW_SIZE)
                 .expect("usize fits in u64"),
-            max_stream_data_uni: u64::try_from(INITIAL_RECV_WINDOW_SIZE)
+            max_stream_data_uni: u64::try_from(INITIAL_STREAM_RECV_WINDOW_SIZE)
                 .expect("usize fits in u64"),
             max_streams_bidi: LOCAL_STREAM_LIMIT_BIDI,
             max_streams_uni: LOCAL_STREAM_LIMIT_UNI,
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -9,7 +9,7 @@
 
 use std::{
     cmp::min,
-    fmt::Debug,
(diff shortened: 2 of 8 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -29,7 +27,7 @@ use crate::{
     CongestionControlAlgorithm, Res, DEFAULT_INITIAL_RTT,
 };
 
-const LOCAL_MAX_DATA: u64 = MAX_VARINT;
+const LOCAL_MAX_DATA: u64 = (INITIAL_RECV_WINDOW_SIZE * 16) as u64;
```
mxinden:
```suggestion
const LOCAL_MAX_DATA: u64 = (INITIAL_STREAM_RECV_WINDOW_SIZE * 16) as u64;
```

Given that we would now have both stream and connection level flow control, I suggest renaming the constant.

neqo-transport/src/fc.rs
```diff
@@ -330,20 +332,82 @@ impl ReceiverFlowControl<()> {
...
     }
 
+    /// Auto-tune [`ReceiverFlowControl::max_active`], i.e. the connection flow
+    /// control window.
+    ///
+    /// If the sending rate (`window_bytes_used`) exceeds the rate allowed by
+    /// the maximum flow control window and the current rtt
+    /// (`window_bytes_expected`), try to increase the maximum flow control
+    /// window ([`ReceiverFlowControl::max_active`]).
+    fn auto_tune(&mut self, now: Instant, rtt: Duration) {
```
mxinden, line 443: Can (some of) this be de-duplicated with the existing auto tuning implementation?

https://github.com/mozilla/neqo/blob/51fe8fc35c079802c93dbde79f59cfcc899af920/neqo-transport/src/fc.rs#L402-L458

neqo-transport/src/connection/params.rs
mxinden: Worth documenting that the `16` here is based on the `LOCAL_STREAM_LIMIT_BIDI`?

WaterWhisperer: Understood, thanks for the review!

neqo-transport/src/connection/params.rs
```diff
@@ -29,7 +27,14 @@ use crate::{
...
 
-const LOCAL_MAX_DATA: u64 = MAX_VARINT;
+/// Initial connection-level receive window size.
+///
+/// Set to 16 times the per-stream initial window to accommodate the default
+/// bidirectional stream limit ([`LOCAL_STREAM_LIMIT_BIDI`]).
+/// This provides sufficient capacity for concurrent streams without
+/// being excessively large, and allows auto-tuning to grow the window
+/// based on actual usage.
+const LOCAL_MAX_DATA: u64 = (INITIAL_STREAM_RECV_WINDOW_SIZE * 16) as u64;
```
mxinden:
```suggestion
const LOCAL_MAX_DATA: u64 = (INITIAL_STREAM_RECV_WINDOW_SIZE * LOCAL_STREAM_LIMIT_BIDI) as u64;
```

neqo-transport/src/fc.rs
```diff
@@ -322,6 +324,52 @@ where
...
     }
+
+    /// Core auto-tuning logic for adjusting the maximum flow control window.
+    ///
+    /// This method is called by both connection-level and stream-level
+    /// implementations. It increases `max_active` when the sending rate exceeds
+    /// what the current window and RTT would allow, capping at `max_window`.
+    ///
+    /// Returns the amount by which `max_active` was increased (0 if unchanged).
+    fn auto_tune_impl(&mut self, now: Instant, rtt: Duration, max_window: u64) -> u64 {
```
mxinden:
```suggestion
    fn auto_tune_inner(&mut self, now: Instant, rtt: Duration, max_window: u64) -> u64 {
```

Nit pick, more often do we use the `_inner` suffix instead of `_impl`.

https://github.com/search?q=repo%3Amozilla%2Fneqo%20_inner&type=code

neqo-transport/src/fc.rs
```diff
@@ -330,17 +378,51 @@ impl ReceiverFlowControl<()> {
...
+                new max_active: {} MiB, \
+                last update: {:?}, \
+                rtt: {rtt:?}",
+                increase,
+                prev_max_active / 1024 / 1024,
+                self.max_active / 1024 / 1024,
+                now - self
+                    .last_update
+                    .expect("last_update must be set by auto_tune_impl"),
+            );
```
mxinden: This only differs between connection and stream auto-tuning in the template string "Increasing max connection receive window", right? What do you think of passing `"connection"` and `"stream"` to `auto_tune_inner` and then add it to the template string:

``` rust
            qdebug!(
                "Increasing max {} receive window by {} B, \

File: neqo-transport/src/fc.rs [hunk 6]
```diff
@@ -1242,24 +1299,51 @@ mod test {
...
+        assert_eq!(max_data, (INITIAL_STREAM_RECV_WINDOW_SIZE * 16) as u64);
+    }
 
-        // Say that the remote consumes 1 byte of that connection flow control
-        // limit and then requests a connection flow control update.
-        fc.consume(1).unwrap();
-        fc.add_retired(1);
-        fc.send_flowc_update();
+    #[test]
+    fn connection_flow_control_auto_tune() -> Res<()> {
```
mxinden, line 1319: Great. Thank you. Could you extend this test, or add another test, that asserts that we stop at `LOCAL_MAX_DATA`?

mxinden (commented):
Thanks for the follow-ups. Minor comments.

Once this is merged, https://github.com/mozilla/neqo/pull/3011 will be unblocked.

neqo-transport/src/fc.rs
```diff
@@ -407,54 +485,8 @@ impl ReceiverFlowControl<StreamId> {
...
-            last update: {:?}, \
-            rtt: {rtt:?}, \
-            stream_id: {}",
-            self.max_active - prev_max_active,
-            prev_max_active / 1024 / 1024,
-            self.max_active / 1024 / 1024,
-            now - max_allowed_sent_at,
-            self.subject,
-        );
+        let fc_type = format!("stream {}", self.subject);
```
mxinden: Ah, I didn't think of the stream id in the case of stream receive window autotuning. Sorry about that.

Any simple idea how we can get around the heap allocation on each call to `auto_tune` here? Otherwise this looks great to me.

Maybe an `enum AutoTuneSubject { Connection, Stream(StreamId) }` that has a `Display` implementation?

martinthomson: I like the new enum idea.

mxinden:
CI failure is unrelated and should be fixed with https://github.com/mozilla/neqo/pull/3092.

PR #3083, opened 2025-10-27
feat: Try `RustCrypto`

larseggert:
WIP


larseggert:
Obviously quite a bit slower than NSS :-( The bencher only has SSE2, so maybe it's faster on other/newer silicon, but this is a pretty clear signal we shouldn't use this.

larseggert:
#3084 is much more promising.

PR #3084, opened 2025-10-27
feat: Try `aws-lc-rs`

larseggert:


larseggert:
This was a useful exercise, but we were never going to go that route.

larseggert:
Opening this to get @dennisjackson  some new numbers

larseggert:
See #3601.

Issue #1472, opened 2023-10-19
Out of order packets cause spurious retransmit


mb:
While debugging upload for [Bug 1852924](https://bugzilla.mozilla.org/show_bug.cgi?id=1852924) I noticed that all the detected losses during my testing are spurious. An ack arrives shortly after. The loss detection seems to trigger after an out-of-order packet E with slightly shorter RTT. Probably because we update the `last_rtt` and use it to determine whether the packets before the packet E were lost. Packets before E were send earlier, are therefore in flight for longer and are all marked as lost.

I will investigate further, but wanted to note down this observation.

mb:
This is the kind of packet loss I see: In this graph you can see the calls to [cc_on_packet_acked](https://github.com/mozilla/neqo/blob/6c05ac330dd4850e2088d9d36bc0db5ac4fda864/neqo-transport/src/cc/classic_cc.rs#L151) [on_packets_lost](https://github.com/mozilla/neqo/blob/6c05ac330dd4850e2088d9d36bc0db5ac4fda864/neqo-transport/src/cc/classic_cc.rs#L241) [on_packet_sent](https://github.com/mozilla/neqo/blob/6c05ac330dd4850e2088d9d36bc0db5ac4fda864/neqo-transport/src/cc/classic_cc.rs#L293) graphed over time. All the lost packets receive an ack slightly later (around 0.35ms). In this example we even sent out a recovery packet (=> still need to investigate what the conditions are when we don't sent out recovery packets https://github.com/mozilla/neqo/issues/1473)

![packet_loss](https://github.com/mozilla/neqo/assets/36711370/1e6c16fa-2121-4d6b-a670-78ca2de10e9c)

In this example it is probably the `PACKET_THRESHOLD = 3`[[1]](https://github.com/mozilla/neqo/blob/6c05ac330dd4850e2088d9d36bc0db5ac4fda864/neqo-transport/src/recovery.rs#L35)[[2]](https://github.com/mozilla/neqo/blob/6c05ac330dd4850e2088d9d36bc0db5ac4fda864/neqo-transport/src/recovery.rs#L421-L427) (from https://datatracker.ietf.org/doc/html/rfc9002#name-packet-threshold) causes the packets to be marked as lost.

mb:
[...]

PR #3086, opened 2025-10-28
feat(transport/cc): spurious congestion event detection

omansfeld:
- Adds `CongestionControlStats` struct with `congestion_events` and `spurious_congestion_events` stats.
- Passes `CongestionControlStats` down to the congestion controller.
- Adds mechanism to detect spurious congestion events by logging lost packet numbers in `ClassicCongestionControl::maybe_lost_packets` and cleaning them out as acks come in. Once all lost packets have been (late) acked a spurious congestion event is detected. The data structure is regularly purged from old packets where no late acks are expected anymore.
- Logs those events in the added stats.

This is the first part of #2694 taking care of the detection mechanism and exposing stats. I will link a PR on the Firefox side to expose the stats as telemetry later.

The next step after this will be to implement the congestion event undo mechanism described in [RFC 9438](https://datatracker.ietf.org/doc/html/rfc9438#name-spurious-fast-retransmits).

Some resources and other implementations of the same mechanisms:
- [quinn](https://github.com/quinn-rs/quinn/pull/2289) has very promising numbers in their PR. The graphs also nicely show the congestion window recovery.
- [msquic](https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L934-L960): This is where I got the `2*PTO` maximum age for lost packets from. Packets declared longer than `2 * PTO` ago are purged from the list.
- [cloudflare quiche](https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L934-L960) uses a different detection mechanism where they compare lost packet count to lost packet count before the congestion event and undo if it goes below a certain threshold (due to late acks). This is more aggressive, as it doesn't make sure all lost packets were actually spurious losses.
- it's hard to compare to what the linux kernel is doing as the detection mechanisms for TCP look very different than in QUIC, as is also described in [RFC 9438 4.9.2](https://datatracker.ietf.org/doc/html/rfc9438#section-4.9.2-1).

Using this patch and running upload tests against [h3.speed.cloudflare.com] (`RUST_LOG=neqo_bin=info cargo run --bin neqo-client -- -m=POST --stats --upload-size=300000000 https://h3.speed.cloudflare.com/`) on my home network (which can generally be described as very flaky) I can actually already observe a good amount of spurious congestion events. Take e.g. this example output:
```
53.453 INFO stats for Client ...
  rx: 147040 drop 3 dup 0 saved 3
  tx: 215429 lost 47 lateack 22 ptoack 0 unackdrop 0            <--------- late acks observed here
  cc: congestion_events 24 spurious_congestion_events 5          <-------- spurious congestion events here
  pmtud: 4 sent 4 acked 0 lost 0 change 1500 iface_mtu 1500 pmtu
  resumed: false
  frames rx:
    crypto 6 done 1 token 0 close 0
    ack 147026 (max 215411) ping 5745 padding 0
    stream 5 reset 0 stop 0
    max: stream 0 data 224 stream_data 251
    blocked: stream 0 data 0 stream_data 0
    datagram 0
    ncid 0 rcid 0 pchallenge 0 presponse 0
    ack_frequency 0
  frames tx:
    crypto 4 done 0 token 0 close 1
    ack 5787 (max 147906) ping 4 padding 4
    stream 215250 reset 0 stop 0
    max: stream 2 data 0 stream_data 0
    blocked: stream 0 data 0 stream_data 250
    datagram 0
    ncid 0 rcid 0 pchallenge 0 presponse 0
    ack_frequency 0
  ecn:
    tx:
      Initial Count({NotEct: 4, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 4, Ect1: 0, Ect0: 0, Ce: 0})
      Short Count({NotEct: 215411, Ect1: 0, Ect0: 10, Ce: 0})
    acked:
    rx:
      Initial Count({NotEct: 4, Ect1: 0, Ect0: 0, Ce: 0})
      Handshake Count({NotEct: 3, Ect1: 0, Ect0: 0, Ce: 0})
      Short Count({NotEct: 147030, Ect1: 0, Ect0: 0, Ce: 0})
    path validation outcomes: ValidationCount({Capable: 0, NotCapable(BlackHole): 0, NotCapable(Bleaching): 1, NotCapable(ReceivedUnsentECT1): 0})
    mark transitions:
  dscp: Cs0: 147043
```

@mb had also observed similar spurious losses in the past when uploading to google drive and had documented his findings in #1472.

I think overall this points to the undo mechanism having big potential wins to our upload. Looking forward to see what the stats will show in the wild!

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -1425,8 +1425,6 @@ mod tests {
         assert_eq!(cc_stats.congestion_events_loss, 0);
         assert_eq!(cc_stats.congestion_events_spurious, 0);
 
-        now += RTT;
-
         let mut lost_pkt1 = pkt1.clone();
         let mut lost_pkt2 = pkt2.clone();
         lost_pkt1.declare_lost(now);
@@ -1482,5 +1480,51 @@ mod tests {
         assert_eq!(cc.state, State::CongestionAvoidance);
         assert_eq!(cc_stats.congestion_events_loss, 1);
         assert_eq!(cc_stats.congestion_events_spurious, 1);
+    }
+
+    #[test]
+    fn spurious_congestion_event_detection_cleanup() {
+        let mut cc = ClassicCongestionControl::new(NewReno::default(), Pmtud::new(IP_ADDR, MTU));
+        let mut now = now();
+        let mut cc_stats = CongestionControlStats::default();
+        let rtt_estimate = RttEstimate::new(crate::DEFAULT_INITIAL_RTT);
+
+        let pkt1 = sent::make_packet(1, now, 1000);
+        cc.on_packet_sent(&pkt1, now);
+
+        cc.on_packets_lost(
+            Some(now),
+            None,
+            rtt_estimate.pto(true),
+            &[pkt1],
+            now,
+            &mut cc_stats,
+        );
+
+        // The lost should be added now.
+        assert!(!cc.maybe_lost_packets.is_empty());
+
+        // Packets older than 2 * PTO are removed, so we increase by exactly that.
+        now += 2 * rtt_estimate.pto(true);
+
+        // The cleanup is called when we ack packets, so we send and ack a new one.
+        let pkt2 = sent::make_packet(2, now, 1000);
+        cc.on_packet_sent(&pkt2, now);
+        cc.on_packets_acked(&[pkt2], &rtt_estimate, now, &mut cc_stats);
+
+        // The packet is exactly the maximum age, so it shouldn't be removed yet. This assert makes
+        // sure we don't clean up too early.
+        assert!(!cc.maybe_lost_packets.is_empty());
+
+        // Increase by 1ms to get over the maximum age.
+        now += Duration::from_millis(1);
+
+        // Send and ack another packet to trigger cleanup.
+        let pkt3 = sent::make_packet(3, now, 1000);
+        cc.on_packet_sent(&pkt3, now);
+        cc.on_packets_acked(&[pkt3], &rtt_estimate, now, &mut cc_stats);
+
+        // Now the packet should be removed.
+        assert!(cc.maybe_lost_packets.is_empty());
     }
 }


neqo-transport/src/stats.rs
```diff
@@ -136,6 +136,15 @@ pub struct DatagramStats {
     pub dropped_queue_full: usize,
 }
 
+/// Congestion Control stats
+#[derive(Default, Clone, PartialEq, Eq)]
+pub struct CongestionControlStats {
+    /// Total number of congestion events
+    pub congestion_events: usize,
+    /// Number of spurious congestion events (congestion events due to lost packets that were later
+    /// acked)
```
mxinden: Suggestion by Copilot:

```suggestion
/// Number of spurious congestion events, where congestion was incorrectly inferred due to packets
/// initially considered lost but subsequently acknowledged. This indicates instances where the
/// congestion control algorithm overreacted to perceived losses.
```

neqo-transport/src/stats.rs
```diff
@@ -136,6 +136,15 @@ pub struct DatagramStats {
     pub dropped_queue_full: usize,
 }
 
+/// Congestion Control stats
+#[derive(Default, Clone, PartialEq, Eq)]
+pub struct CongestionControlStats {
+    /// Total number of congestion events
```
mxinden:
```suggestion
    /// Total number of congestion events.
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -304,6 +304,13 @@ impl Packets {
     }
 }
 
+/// A packet that has supposedly been lost. Is used for spurious congestion event detection.
+#[derive(Debug)]
+pub struct MaybeLostPacket {
+    pub pn: packet::Number,
+    pub time_sent: Instant,
+}
```
mxinden: Given that this is only used in `classic_cc.rs`, why define it here?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
     }
 
+    fn detect_spurious_congestion_event(&mut self, acked_packets: &[sent::Packet]) -> bool {
+        if self.maybe_lost_packets.is_empty() {
+            return false;
+        }
+
+        // Build HashSet for O(1) lookup
+        let acked_pns: std::collections::HashSet<_> =
+            acked_packets.iter().map(&sent::Packet::pn).collect();
```
mxinden: Instead of allocating here, couldn't `maybe_lost_packets` instead be a `HashSet` and then we would do:

```
for acked_packet in acked_packets {
  self.maybe_lost_packets.remove(acked_packet.pn());
}
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+
+        // Removes all newly acked packets from `maybe_lost_packets`
+        self.maybe_lost_packets
+            .retain(|packet| !acked_pns.contains(&packet.pn));
+
+        // If all of them have been removed we detected a spurious congestion event
+        self.maybe_lost_packets.is_empty()
+    }
+
+    /// Cleanup lost packets that we are fairly sure will never be getting a late acknowledgment
```
```suggestion
    /// Cleanup lost packets that we are fairly sure will never be getting a late acknowledgment for.
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -275,21 +314,32 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
-        for pkt in lost_packets.iter().filter(|pkt| pkt.cc_in_flight()) {
-            qdebug!(
-                "packet_lost this={self:p}, pn={}, ps={}",
-                pkt.pn(),
-                pkt.len()
-            );
-            // BIF is set to 0 on a path change, but in case that was because of a simple rebinding
-            // event, we may still declare packets lost that were sent before the rebinding.
-            self.bytes_in_flight = self.bytes_in_flight.saturating_sub(pkt.len());
+        self.cleanup_maybe_lost_packets(now, pto);
```
mxinden: Say we  get a million losses early on, but no additional one. Then `cleanup_maybe_lost_packets` would never be called again, and thus the packets would never be cleaned up. Is that a problem? Would it make more sense to do this in `on_packets_acked`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         rtt_est: &RttEstimate,
         now: Instant,
+        cc_stats: &mut CongestionControlStats,
     ) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
+
+        if self.detect_spurious_congestion_event(acked_pkts) {
+            cc_stats.spurious_congestion_events += 1;
+            // Could send signal to CUBIC for undo here
```
mxinden: I suggest a `TODO` with a link to the issue. Example:

https://github.com/mozilla/neqo/blob/51fe8fc35c079802c93dbde79f59cfcc899af920/neqo-http3/src/frames/connect_udp_frame.rs#L12

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         acked_pkts: &[sent::Packet],
         rtt_est: &RttEstimate,
         now: Instant,
+        cc_stats: &mut CongestionControlStats,
     ) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
+
+        if self.detect_spurious_congestion_event(acked_pkts) {
+            cc_stats.spurious_congestion_events += 1;
```
mxinden: Why not move this into `detect_spurious_congestion_events`? That would nicely encapsulate all spurious congestion event related logic into one function.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         &mut self,
         acked_pkts: &[sent::Packet],
         rtt_est: &RttEstimate,
         now: Instant,
+        cc_stats: &mut CongestionControlStats,
     ) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
+
+        if self.detect_spurious_congestion_event(acked_pkts) {
```
mxinden: Shouldn't we call `cleanup_maybe_lost_packets` right before, to make sure old packets are ignored before `detect_spurious_congestion_event`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +176,46 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+        self.maybe_lost_packets
+            .retain(|packet| !acked_pns.contains(&packet.pn));
+
+        // If all of them have been removed we detected a spurious congestion event
+        self.maybe_lost_packets.is_empty()
+    }
+
+    /// Cleanup lost packets that we are fairly sure will never be getting a late acknowledgment
+    fn cleanup_maybe_lost_packets(&mut self, now: Instant, pto: Duration) {
+        // The `pto * 2` maximum age of the lost packets is taken from msquic's implementation
```
mxinden: Link to code

neqo-transport/src/recovery/sent.rs
omansfeld: Initially had it there because that's where we also define `Packet` etc. But no strong opinion, I will move it to `classic_cc.rs`.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Had to use a `HashMap<u64, MaybeLostPacket>` so I can compare to the packet number as key while still having the `time_sent` as part of the struct/value. Another option would be to ditch the `MaybeLostPacket` struct and just do `HashMap<u64, Instant>` which would save some lines of code but I think the `MaybeLostPacket` struct nicely documents what the `Instant` in that `HashMap` should be.

neqo-transport/src/cc/classic_cc.rs
omansfeld: As per offline discussion:
`on_packets_lost` is called even if there are no lost packets, so the cleanup would still be called. This is fairly hidden though, so I agree we could use the call somewhere else, as the important thing is it getting called before we evaluate and call `detect_spurious_congestion_events` the next time. So yes, we could move it prior to that call in `on_packets_acked`.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +183,52 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         &mut self,
         acked_pkts: &[sent::Packet],
         rtt_est: &RttEstimate,
         now: Instant,
+        cc_stats: &mut CongestionControlStats,
     ) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
+
+        self.cleanup_maybe_lost_packets(now, rtt_est.pto(true));
```
omansfeld, line 201: Moving the cleanup function into `on_packets_acked` meant  we didn't have the pto available as a parameter anymore. It can be calculated from `rtt_est` though, but we need to supply a bool (`confirmed`). If it is true the maximum ack delay is added to the PTO. 

https://github.com/mozilla/neqo/blob/ab40a794d38306cc6430e779b3e3db1ab0e578e7/neqo-transport/src/rtt.rs#L158-L164

I'm not sure if we can just supply `true` here or if it is better to pass down the pto or the `confirmed` value from `recovery/mod.rs`.

For now I opted with just supplying `true` as e.g. msquic is also adding the ack delay (https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L242).

neqo-transport/src/cc/classic_cc.rs
mxinden: Instead of a `u64`, maybe use this type alias:

https://github.com/mozilla/neqo/blob/ab40a794d38306cc6430e779b3e3db1ab0e578e7/neqo-transport/src/packet/mod.rs#L53

neqo-transport/src/cc/classic_cc.rs
mxinden: Not great, but I guess fine. Can you add a comment, e.g. along the lines of: this is best effort, too large pto does no harm.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -275,21 +327,32 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
-            qdebug!(
-                "packet_lost this={self:p}, pn={}, ps={}",
+        for pkt in lost_packets {
+            if pkt.cc_in_flight() {
+                qdebug!(
+                    "packet_lost this={self:p}, pn={}, ps={}",
+                    pkt.pn(),
+                    pkt.len()
+                );
+                // BIF is set to 0 on a path change, but in case that was because of a simple
```
```suggestion
                // bytes_in_flight is set to 0 on a path change, but in case that was because of a simple
```

Nit pick, in my eyes, `bytes_in_flight` is not long enough to warrant a abbreviation.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -95,14 +99,24 @@ pub trait WindowAdjustment: Display + Debug {
...
 #[derive(Debug)]
 pub struct ClassicCongestionControl<T> {
     cc_algorithm: T,
     state: State,
     congestion_window: usize, // = kInitialWindow
     bytes_in_flight: usize,
     acked_bytes: usize,
+    /// Packets that have supposedly been lost. These are used for spurious congestion event
+    /// detection.
+    maybe_lost_packets: HashMap<u64, MaybeLostPacket>,
```
```suggestion
    maybe_lost_packets: HashMap<packet::Number, MaybeLostPacket>,
```

Can you use the type alias here?

https://github.com/mozilla/neqo/blob/ab40a794d38306cc6430e779b3e3db1ab0e578e7/neqo-transport/src/packet/mod.rs#L53

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +183,52 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         &mut self.pmtud
     }
 
+    fn detect_spurious_congestion_event(
```
mxinden: Given that `detect_spurious_congestion_event` is only called from `on_packets_acked`, does it need to be part of the `CongestionControl` trait, or would a simple `fn` on `ClassicCongestionControl` suffice?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -169,14 +183,52 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+
+        if self.maybe_lost_packets.is_empty() {
+            cc_stats.spurious_congestion_events += 1;
+            // TODO: Implement spurious congestion event handling: <https://github.com/mozilla/neqo/issues/2694>
+        }
+    }
+
+    /// Cleanup lost packets that we are fairly sure will never be getting a late acknowledgment
+    /// for.
+    fn cleanup_maybe_lost_packets(&mut self, now: Instant, pto: Duration) {
```
mxinden: Same here. Needs to be part of the trait?

mxinden (commented):
Minor comments.

Once you have been able to integrate this in Firefox and we see it working, I suggest we merge here. Sounds good?

neqo-transport/src/cc/classic_cc.rs
larseggert: Can we use FxHashMap unless it's test code? (Can't tell if this is on the phone.)

neqo-transport/src/cc/classic_cc.rs
omansfeld: This was just copied over into the `if` from before this change, but agreed, I'll change the abbreviation.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -8,6 +8,7 @@
 
 use std::{
     cmp::{max, min},
+    collections::HashMap,
```
```suggestion
use rustc_hash::FxHashMap as HashMap;
```

omansfeld:
> Minor comments.
> 
> Once you have been able to integrate this in Firefox and we see it working, I suggest we merge here. Sounds good?

Sounds good, I will get to the Firefox integration next week!

neqo-transport/src/cc/classic_cc.rs
omansfeld: Wasn't aware of `FxHashMap`'s existence, thanks! :)

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -466,6 +508,38 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
+        }
+
+        // Removes all newly acked packets from `maybe_lost_packets`
+        for acked_packet in acked_packets {
+            self.maybe_lost_packets.remove(&acked_packet.pn());
+        }
+
+        // If all of them have been removed we detected a spurious congestion event
+
+        if self.maybe_lost_packets.is_empty() {
```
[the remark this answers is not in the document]
omansfeld: Copilot actually has a minor point here, because it could happen that we have a second spurious congestion event before having recovered from the first. In that case we'd have to clear all late acks before detecting ONE spurious congestion event because we don't track per congestion event. We will never declare an event spurious too early though because very lost packet has to be specifically recovered for `maybe_lost_packets` to be empty.

Had briefly discussed this offline and doing tracking per congestion event (especially if we also think about the future undo mechanism) would be complicated and introduce a whole lot more state that has to be saved. Other implementations also don't track per congestion event, as is linked in the PR description.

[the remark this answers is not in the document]
mxinden: Impressive find by Copilot. That said, agreed with @omansfeld. Let's keep it simple (for now).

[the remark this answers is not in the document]
larseggert: Want to add a TODO?

omansfeld:
Added a TODO for potentially tracking spurious congestion events per congestion epoch.

I have this integrated in firefox now, but not sure how to best upload the patch because it includes vendoring in the neqo changes. If it's wanted I can upload a patch to Phabricator as WIP, otherwise I'll just do the proper patch once the neqo change is in nightly.

Anyways, I can see the following in <https://debug-ping-preview.firebaseapp.com/pings/> after running a few simultaneous <https://h3.speed.cloudflare.com> speedtests:

  ```
  "networking.http_3_spurious_congestion_event_ratio": {
    "values": {
      "0": 10,
      "5000": 1 // Note the 10k scaling factor for glean metrics, so this is 50% spurious events observed for one connection
    },
    "sum": 5000
  },
  ```
  
  @mxinden can you do another review and approve if there is nothing left on your side so we can merge?

omansfeld:
#3111 got me thinking:

Should we include ECN congestion events here? They cannot be spurious, so I think it is correct to only include congestion events due to loss in the ratio recorded in glean.

Should we then also expose another stat that counts congestion events due to ECN-CE? I assume just as `lost` only correlates to `congestion_events` the amount of ECN-CE marks that we observe is also not equal to the number of congestion events caused by it. This would give us the biggest granularity and enable us to record all kind of things to glean in `neqo_glue` (ECN-CE vs loss ratio, spurious vs all cc events, spurious vs loss cc events, ...).

I think that would be best, but not sure if it should happen in this PR. I don't think it is much more complexity regarding review, but it's only barely related...

---

And while thinking about this I found an actual bug in this code: I recorded congestion events twice (in `on_packets_lost` and `on_congestion_event`). Will wait with a fix for the above discussion as it depends on what we decide here.

My suggestion would be to record loss based cc events here:

https://github.com/mozilla/neqo/blob/6d56ec96fc97b488b0d67df1e72807386fdb23dd/neqo-transport/src/cc/classic_cc.rs#L311

And ECN-CE based cc events here:

https://github.com/mozilla/neqo/blob/6d56ec96fc97b488b0d67df1e72807386fdb23dd/neqo-transport/src/cc/classic_cc.rs#L332-L334

@mxinden So hold off with the review I asked for above for now, but please share your thoughts on this when you get to it.

larseggert:
> Should we include ECN congestion events here? They cannot be spurious

Good observation! I would not include ECN when considering whether congestion events were spurious - they are a **confirmed** congestion event, albeit possibly at lower intensity (so might warrant a special reaction).

> Should we then also expose another stat that counts congestion events due to ECN-CE?

Yes.

mxinden:
Above sounds good!

For ECN CE, we currently only have:

https://glam.telemetry.mozilla.org/fog/probe/networking_http_3_ecn_ce_ect0_ratio_received/explore?

omansfeld:
I implemented the above in the latest commit.

We now have the stats `congestion_events_due_to_ecn` and `congestion_events_due_to_loss` and can then decide in `neqo_glue` on the Firefox side how exactly we want to collect those in metrics.

Also fixed the double count I had before and added a check for `congestion_events_due_to_ecn` counting in the existing `classic_cc::tests::ecn_ce` test.

---

If you're interested, another data point from my weird home network that sees a lot of spurious loss:

```
RUST_LOG=neqo_bin=info cargo run --bin neqo-client -- -m=POST --stats --upload-size=100000000 https://h3.speed.cloudflare.com/ --qlog-dir .
```

```
 tx: 72112 lost 69 lateack 41 ptoack 15 unackdrop 0
 cc: loss_congestion_events 11 ecn_congestion_events 0 spurious_congestion_events 5
```

And the following qvis congestion graph:

Screenshot 2025-11-06 at 16 36 32

Can really see the impact spurious congestion events have here. We basically don't spend any time in slowstart because we immediately have some spurious loss and then we're never able to really converge to a congestion point over the whole 100MB upload.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -79,7 +85,8 @@ fn packet_lost(cc: &mut ClassicCongestionControl<Cubic>, pn: u64) {
         recovery::Tokens::new(),
         cc.max_datagram_size(),
     );
-    cc.on_packets_lost(None, None, PTO, &[p_lost], now());
+    let mut cc_stats = CongestionControlStats::default();
+    cc.on_packets_lost(None, None, PTO, &[p_lost], now(), &mut cc_stats);
```
larseggert:
```suggestion
    cc.on_packets_lost(None, None, PTO, &[p_lost], now(), &mut CongestionControlStats::default());
```
Since `cc_stats` is not used afterwards. Or is there actually something we can assert for `cc_stats` after the call?

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -254,9 +263,10 @@ fn assert_within<T: Sub<Output = T> + PartialOrd + Copy>(value: T, expected: T,
 #[test]
 fn congestion_event_slow_start() {
     let mut cubic = ClassicCongestionControl::new(Cubic::default(), Pmtud::new(IP_ADDR, MTU));
+    let mut cc_stats = CongestionControlStats::default();
 
     _ = fill_cwnd(&mut cubic, 0, now());
-    ack_packet(&mut cubic, 0, now());
+    ack_packet(&mut cubic, 0, now(), &mut cc_stats);
```
larseggert, line 269: Can we assert something for `cc_stats`?

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -297,7 +308,7 @@ fn congestion_event_congestion_avoidance() {
         .set_w_max(3.0 * max_datagram_size_f64);
 
     _ = fill_cwnd(&mut cubic, 0, now());
-    ack_packet(&mut cubic, 0, now());
+    ack_packet(&mut cubic, 0, now(), &mut cc_stats);
```
larseggert, line 312: Can we assert something for `cc_stats`?

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -321,7 +333,7 @@ fn congestion_event_congestion_avoidance_fast_convergence() {
     cubic.cc_algorithm_mut().set_w_max(cwnd_initial_f64 * 10.0);
 
     _ = fill_cwnd(&mut cubic, 0, now());
-    ack_packet(&mut cubic, 0, now());
+    ack_packet(&mut cubic, 0, now(), &mut cc_stats);
```
larseggert, line 338: Ditto.

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -364,5 +377,10 @@ fn congestion_event_congestion_avoidance_no_overflow() {
     assert_eq!(cubic.cwnd(), cubic.cwnd_initial());
 
     // Now ack packet that was send earlier.
-    ack_packet(&mut cubic, 0, now().checked_sub(PTO).unwrap());
+    ack_packet(
+        &mut cubic,
+        0,
+        now().checked_sub(PTO).unwrap(),
+        &mut cc_stats,
+    );
```
larseggert, line 388: Ditto.

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -31,9 +32,14 @@ fn cwnd_is_halved(cc: &ClassicCongestionControl<NewReno>) {
     assert_eq!(cc.ssthresh(), cc.cwnd_initial() / 2);
 }
 
+#[expect(
+    clippy::too_many_lines,
+    reason = "A lot of multiline function calls due to formatting"
+)]
```
larseggert: Could also disable the lint for the entire file, since it is just tests.

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -217,6 +217,7 @@ fn debug() {
         "stats for\u{0020}
   rx: 0 drop 0 dup 0 saved 0
   tx: 0 lost 0 lateack 0 ptoack 0 unackdrop 0
+  cc: loss_congestion_events 0 ecn_congestion_events 0 spurious_congestion_events 0
```
larseggert: I'd shorten the stats output a lot, maybe:

```suggestion
  cc: loss_ce 0 ecn_ce 0 spurious_ce 0
```

neqo-transport/src/path.rs
```diff
@@ -1009,9 +1009,11 @@ impl Path {
 
         let ecn_ce_received = self.ecn_info.on_packets_acked(acked_pkts, ack_ecn, stats);
         if ecn_ce_received {
-            let cwnd_reduced = self
-                .sender
-                .on_ecn_ce_received(acked_pkts.first().expect("must be there"), now);
+            let cwnd_reduced = self.sender.on_ecn_ce_received(
+                acked_pkts.first().expect("must be there"),
```
larseggert, line 1015: Is this `expect` safe?

neqo-transport/src/stats.rs
```diff
@@ -136,6 +136,18 @@ pub struct DatagramStats {
...
+#[derive(Default, Clone, PartialEq, Eq)]
+pub struct CongestionControlStats {
+    /// Total number of congestion events caused by packet loss.
+    pub congestion_events_due_to_loss: usize,
+    /// Total number of congestion events caused by ECN-CE marked packets.
+    pub congestion_events_due_to_ecn: usize,
+    /// Number of spurious congestion events, where congestion was incorrectly inferred due to
+    /// packets initially considered lost but subsequently acknowledged. This indicates
+    /// instances where the congestion control algorithm overreacted to perceived losses.
+    pub spurious_congestion_events: usize,
```
larseggert:
```suggestion
    pub congestion_events_loss: usize,
    /// Total number of congestion events caused by ECN-CE marked packets.
    pub congestion_events_ecn: usize,
    /// Number of spurious congestion events, where congestion was incorrectly inferred due to
    /// packets initially considered lost but subsequently acknowledged. This indicates
    /// instances where the congestion control algorithm overreacted to perceived losses.
    pub congestion_events_spurious: usize,
```
And maybe think whether we can shorten `congestion_events_`, too.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -466,6 +515,42 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
         }
     }
 
+    // TODO: Maybe do tracking of lost packets per congestion epoch. Right now if we get a spurious
+    // event and then before the first was recovered get another (or even a real congestion event
+    // because of random loss, path change, ...), it will only be detected as spurious once the old
+    // and new lost packets are recovered. This means we'd have two spurious events counted as one
+    // and would also only be able to recover to the cwnd prior to the second event.
```
larseggert: For when is this `TODO`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1271,10 +1387,109 @@ mod tests {
...
 
         // Signal congestion (ECN CE) and thus change state to recovery start.
-        cc.on_ecn_ce_received(&p_ce, now);
+        cc.on_ecn_ce_received(&p_ce, now, &mut cc_stats);
         cwnd_is_halved(&cc);
         assert_eq!(cc.state, State::RecoveryStart);
+        assert_eq!(cc_stats.congestion_events_due_to_ecn, 1);
+    }
+
+    fn make_packet_by_pn(pn: u64, sent_time: Instant) -> sent::Packet {
```
larseggert: There is also `make_sent_packet` in `pmtud.rs`. Wonder if these can be consolidated.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -174,9 +189,19 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
         now: Instant,
+        cc_stats: &mut CongestionControlStats,
     ) {
         let mut is_app_limited = true;
         let mut new_acked = 0;
+
+        // Supplying `rtt_est.pto(true)` here is best effort not to have to track
+        // `recovery::Loss::confirmed()` all the way down to the congestion controller. Having too
+        // big a PTO does no harm here and usually the connection should be confirmed at this point
+        // anyway.
```
```suggestion
        // big a PTO does no harm here.
```

In an 1-RTT handshake, the client will send its initial, then receive the server's initial including an ACK from the server. That moves the connection into `State::Connected`. The ACK should call `on_packets_acked`. Only on the second round-trip does the client receive the `HANDSHAKE_DONE` frame from the client, subsequently going into `State::Confirmed`.

Long story short, I think the above is confusing. I suggest removing it.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -275,21 +300,32 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+                // were sent before the rebinding.
+                self.bytes_in_flight = self.bytes_in_flight.saturating_sub(pkt.len());
+            }
+            self.maybe_lost_packets.insert(
                 pkt.pn(),
-                pkt.len()
+                MaybeLostPacket {
+                    time_sent: pkt.time_sent(),
+                },
             );
```
mxinden: You could encode the assumption, that no packet is declared lost twice here:


``` rust
            let present = self.maybe_lost_packets.insert(
                pkt.pn(),
                MaybeLostPacket {
                    time_sent: pkt.time_sent(),
                },
            );
            debug_assert!(!present);
```

Up to you. I don't feel strongly about it.

neqo-transport/src/cc/classic_cc.rs
mxinden: Given that this is a fairly small test function only, I am fine with the duplication. Fine either way.

neqo-transport/src/cc/tests/cubic.rs
omansfeld: We could be asserting that the counter for congestion events due to loss increased on the `congestion_event_*` tests. For that we'd need to also pass `cc_stats` that is created in each test down to the `packet_lost` helper and use it there.

Alternatively we can just do what you suggested here for `ack_packet` as well and remove the creating of `cc_stats` in the test functions all together.

I'd prefer the first way as it is a nice way to also test the stat counting when we're running the congestion event tests.

neqo-transport/src/connection/tests/ecn.rs
omansfeld: `ecn_ce` is ambigous with the `ecn-ce` mark. I suggest doing `ce_loss`, `ce_ecn`, `ce_spurious` instead.

neqo-transport/src/path.rs
omansfeld: I'm not quite sure to be honest. This isn't my code though, I just added `&mut stats.cc` to the parameters of that function call, the `expect` was already there. Max added it a year ago in https://github.com/mozilla/neqo/commit/17c45d830469c19d60c876fb5e0393d65d9302e4, so I guess it is save if we haven't seen any reports on it.

neqo-transport/src/stats.rs
omansfeld: We could shorten to `ce_loss` and so forth mirroring the stats output. What is the rationale for going for short variable names here instead of the longer version that is more descriptive? I don't have a strong opinion, just curious.

neqo-transport/src/cc/classic_cc.rs
omansfeld: This was added because of the brief discussion around https://github.com/mozilla/neqo/pull/3086#discussion_r2478364593. I'm not sure if we will ever do this. As far as I can tell no other major implementation does. It might make more sense to either be a `NOTE` or be tracked as a GH issue.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Do we already have a way to make shared test helpers between modules? I couldn't find anything. Apart from making one function (and module) public and importing in the other place, which imo wouldn't be so pretty.

mxinden: There is https://github.com/mozilla/neqo/tree/main/test-fixture/src to share between crates.

There are various shared modules within crates, e.g.:

https://github.com/mozilla/neqo/blob/main/neqo-transport/tests/common/mod.rs

neqo-transport/src/path.rs
mxinden: Yes it is safe. We never call `on_packets_acked` with an empty list of acked packets.

https://github.com/mozilla/neqo/blob/4aebca85325bcb8c98ee16480c3cc263d18c606c/neqo-transport/src/recovery/mod.rs#L639-L642

That said, the `expect` being in a different file than the code upholding this invariant is less than ideal.

neqo-transport/src/cc/classic_cc.rs
omansfeld: This fired on the following test:

https://github.com/mozilla/neqo/blob/4dab7e69b5af3f38075420269ad63efd57f941f4/neqo-transport/src/recovery/mod.rs#L1441-L1442

In the current implementation we put packets from all packet number spaces into the `maybe_lost_packets` structure, meaning we could get the same packet number multiple times, but from different spaces, like it happens in that test.

As discussed offline I added `packet::Type` to the hash map key to be able to id packets across packet number spaces.

omansfeld:
> Exciting to see this happening in a real network, i.e. your home network.
> 
> Did you already start the integration into Firefox? I am curious whether you see this behavior when doing normal web browsing.

I had already done the integration into firefox and tested it locally, but just from the office and not with normal web browsing. With this new version I will compile again and use it for a bit the next time I'm working from home (which might not be this week though...).

That said I addressed the comments in https://github.com/mozilla/neqo/pull/3086/commits/16dae83efe4859216648e09cb1a8e00224909e23, so I think this should be ready for another round of reviews :)

neqo-transport/src/cc/classic_cc.rs
martinthomson: Serious question: could we bit-pack the type and packet number into 64 bits?  Would it help performance?  Note that we only send and track packets from four of the types, so we probably could do that packing in the container.  (I'd suggest that if you try, do that separately; in fact, it might even be a good idea to split the tracking piece from this work.)

larseggert: I think we can. Varints only use 62 bit, so if we always encode them as `u64` we can use the two extra upper bits for the packet type. Whether it helps performance? 🤷

mxinden: > Whether it helps performance? 🤷

Fine exploring it. That said, I would like to only merge such optimization if we have a flamegraph or benchmark proving it is worth it.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -466,6 +516,43 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
+
+    /// Cleanup lost packets that we are fairly sure will never be getting a late acknowledgment
+    /// for.
+    fn cleanup_maybe_lost_packets(&mut self, now: Instant, pto: Duration) {
+        // The `pto * 2` maximum age of the lost packets is taken from msquic's implementation:
+        // <https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L939-L943>
+        let max_age = pto * 2;
+        self.maybe_lost_packets
+            .retain(|_, packet| now.saturating_duration_since(packet.time_sent) <= max_age);
+    }
```
larseggert, line 554: There are some mutants in this function, which usually indicates missing test cases.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -95,14 +99,26 @@ pub trait WindowAdjustment: Display + Debug {
...
     cc_algorithm: T,
     state: State,
     congestion_window: usize, // = kInitialWindow
     bytes_in_flight: usize,
     acked_bytes: usize,
+    /// Packets that have supposedly been lost. These are used for spurious congestion event
+    /// detection. Gets drained when the same packets are later acked and regularly purged from too
+    /// old packets in [`Self::cleanup_maybe_lost_packets`]. Needs a tuple of `(packet::Number,
+    /// packet::Type)` to identify packets across packet number spaces.
+    maybe_lost_packets: HashMap<(packet::Number, packet::Type), MaybeLostPacket>,
```
larseggert, line 118: Didn't @martinthomson suggest to use a binary encoding of `(packet::Number, packet::Type)` here? Did  I miss how that was resolved?

omansfeld: It was suggested to do that seperately and @mxinden added to only do such optimizations if it shows up in flamegraphs. I haven't investigated it further because I was busy with other things, but I'll create an issue to track it for later.

mxinden: For reference, the past conversation: https://github.com/mozilla/neqo/pull/3086#discussion_r2540298810

neqo-transport/src/cc/classic_cc.rs
omansfeld: good catch, I added a test for the cleanup function in https://github.com/mozilla/neqo/pull/3086/commits/f4776b5a4f85ddfeaa479351f82b6c1030919d3b.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -275,21 +300,33 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+            let present = self.maybe_lost_packets.insert(
+                (pkt.pn(), pkt.packet_type()),
+                MaybeLostPacket {
+                    time_sent: pkt.time_sent(),
+                },
             );
-            // BIF is set to 0 on a path change, but in case that was because of a simple rebinding
-            // event, we may still declare packets lost that were sent before the rebinding.
-            self.bytes_in_flight = self.bytes_in_flight.saturating_sub(pkt.len());
+            debug_assert!(present.is_none());
```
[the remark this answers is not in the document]
omansfeld: The comment copilot refers to doesn't say that we'd have packets being declared lost multiple times.

PR #3090, opened 2025-10-29
ci: Run `pkgin update` for netbsd

larseggert:
Maybe that fixes the `curl` issue seen in CI.


larseggert:
Doesn't help. Seems like the NetBSD package repo is messed up.

PR #3089, opened 2025-10-29
chore: Disable Claude

larseggert:
Disable Claude while we're waiting for https://github.com/anthropics/claude-code-action/issues/649

PR #3091, opened 2025-10-29
build: update `windows` 0.58.0 → 0.62.0

ErichDonGubler:


mtu/Cargo.toml
```diff
@@ -23,7 +23,7 @@ libc = { workspace = true }
 static_assertions = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
-windows = { version = ">=0.58,<0.60", features = [
+windows = { version = ">=0.58,<0.63", features = [
```
mxinden: Related: @larseggert should this actually be a workspace dependency? As far as I recall, one can still do the specific feature selection here.

Cargo.toml
```diff
@@ -45,7 +45,7 @@ static_assertions = { version = "1.1", default-features = false }
 strum = { version = "0.27", default-features = false, features = ["derive"] }
 thiserror = { version = "2.0.12", default-features = false }
 url = { version = "2.5", default-features = false, features = ["std"] }
-windows = { version = "0.58", default-features = false }
+windows = { version = "0.62", default-features = false }
```
mxinden: How about adding a range here as well? Would that make your work easier, updating `windows` across the various Firefox repositories?

mtu/Cargo.toml
larseggert: Yes it should, if it's used by multiple crates in the workspace. I probably missed it when I moved mtu over.

Cargo.toml
ErichDonGubler: At first glance, it seems the most workable range is `0.60..=0.62`; 0.59.0 seems API-compatible, but brings in extra `windows-sys` duplicate dependencies. Because Firefox is on 0.58, we don't have an intersection for incremental upgrade, boo. 😐. But I really appreciate the thought! ❤️

ErichDonGubler:
> CI failure due to missing `Cargo.lock` update.

Fixed with 2a92bbff!

larseggert:
@ErichDonGubler can we make `mtu` use the workspace version as well as part of this?

Cargo.toml
mxinden: >  but brings in extra windows-sys duplicate dependencies.

Can you expand why this is the case?

> we don't have an intersection for incremental upgrade, boo. 

We thus have to coordinate the update across all dependencies? Do I understand correct, that, if we merge here today, we can not land a new version of Neqo in https://github.com/mozilla-firefox/firefox in the meantime?

larseggert (approved):

Cargo.toml
ErichDonGubler: > Can you expand why this is the case?

If you run `cargo update -p windows --precise 0.59.0`, you'll observe that `windows-sys` starts to have duplicate versions. This is because `windows` 0.60.0 is the first release to break away from `windows-sys` as a dependency; `windows` 0.59.0 uses a newer version than what's currently being consumed.

> We thus have to coordinate the update across all dependencies? Do I understand correct, that, if we merge here today, we can not land a new version of Neqo in https://github.com/mozilla-firefox/firefox in the meantime?

Yes, we need to coordinate merging this PR if you want to remain free to integrate with `firefox`. Good news on that front: I've already done most of the work; I have a patch stack ready for review at [bug 1990905](https://bugzilla.mozilla.org/show_bug.cgi?id=1990905). If you let me know what commit I should backport this PR to, then that should let you roll forward later.

mxinden: > If you run `cargo update -p windows --precise 0.59.0`, you'll observe that `windows-sys` starts to have duplicate versions.

@ErichDonGubler in Neqo or in Firefox? For Neqo, after executing the command, I don't see duplicate `windows-sys` versions.

Given that we only depend on `windows` in Neqo, I don't yet follow why a range wouldn't suffice for Firefox.

ErichDonGubler:
Heads-up, I now have a patch open for review in Firefox to consume this. I don't expect the form of consuming `neqo` to be final. Once this PR is handled, and I know what revision to target for backporting this PR (or just moving forward? 🫣), I will fix that portion up.

ErichDonGubler:
CI on Windows is running into an issue ([example link](https://github.com/mozilla/neqo/actions/runs/19057203321/job/54429954383?pr=3091#step:7:106)):

```
   = note: libneqo_crypto-7910f964666e40a9.rlib(freebl_static.sysrand.obj) : error LNK2019: unresolved external symbol __imp_SystemFunction036 referenced in function RNG_SystemRNG␍
```

I'm not sure if this is related to this PR, but I suspect it is. I'll investigate a bit, but somebody with understanding of `neqo`'s deps. might be able to figure this out much more effectively than me.

Cargo.toml
ErichDonGubler: Oh, sorry, you're right, it's `windows-targets` that are duped (0.52.6 and 0.53.5) and now aren't used in `windows` upstream, rather than `windows-sys`. That's what I get for trying to remember details 10 minutes after I had dumped them outta my head. 😅

This was with a Neqo checkout, BTW.

larseggert:
CC'ing some of the NSS folks regarding the error above: @dennisjackson @mozkeeler @jschanck

ErichDonGubler:
FYI: I've landed wgpu's end for an upgrade in Firefox: https://github.com/gfx-rs/wgpu/pull/6876

We won't feel any pressure on integrating this for another week or so, so no (immediate) pressure here, just want to facilitate getting this integrated, too.

ErichDonGubler:
The `Run check(s) for VM-only platforms (freebsd)` job in CI also appears to be failing with a network error when attempting to download VM images:

```
  ssh: Could not resolve hostname freebsd: Name or service not known
```

larseggert:
@ErichDonGubler I can't seem to push to this PR. Would you try adding this to `neqo-crypto/Cargo.toml` to see if that fixes the issue?
```toml
[target."cfg(windows)".dependencies]
windows = { workspace = true, features = ["Win32_Security_Authentication_Identity"] }
```

ErichDonGubler:
@larseggert: Sigh, this is a weird problem with submitting a PR from my `erichdongubler-mozilla` organization…edits from maintainers aren't allowed. 😐 Sorry about that.

Your suggested fix appears to work! It also needed to update the `cargo-machete` configuration to ignore `windows` being detected as unused in that crate, but 🙌🏻 yay! Fix found! Force-pushed with [bf2afe6](https://github.com/mozilla/neqo/pull/3091/commits/bf2afe6100efeb9e419f25cb8384f4bb5ee757f3).

larseggert:
@mxinden is this good to go?

ErichDonGubler:
The only failure in CI now is `Claude Code Review`, which has the following apparently auth.-related error:

```
Error: Failed to setup GitHub token: Error: Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?
```

larseggert:
Yeah I disabled Claude code while they fix the issue; see #3089.

ErichDonGubler:
@mxinden: Are you opposed to a `git` dependency with this PR backported on top of the existing release being consumed in Firefox? If not, that would make this a pretty easy `[patch]` entry in the workspace `Cargo.toml` to review there. It's how I've currently structured [D266878](https://phabricator.services.mozilla.com/D266878).

mxinden:
If this is fine with Mozilla supply-chain team, works for me @ErichDonGubler. Thanks.

PR #3093, opened 2025-11-01
fix(test-fixture): avoid server qlog filename collision

### LINKED ISSUE #2744: `AlreadyExists` error when running multiple benches with QLOG
When running a benchmark with QLOG, e.g. `QLOGDIR=/tmp/qlog cargo bench --bench min_bandwidth --features bench` and then re-running e.g. on another branch it fails with the following error:

```sh
thread 'main' panicked at test-fixture/src/lib.rs:250:14:
called `Result::unwrap()` on an `Err` value: IoError(Os { code: 17, kind: AlreadyExists, message: "File exists" })
```

This is because the server qlog filename is always just `server.sqlog` without the connection id suffix the client qlog filename has.

https://github.com/mozilla/neqo/blob/8173862d658a622f61f3ee1a075b1247eab7e5cd/test-fixture/src/lib.rs#L241-L250

I briefly tried fixing it but realized we can't call `Connection::odcid()` yet in `test_fixture::new_server(...)`, so it does not seem like an easy fix (which is probably why it's implemented as is).

At that point I stopped looking at it because it's really just a small annoyance (can just manually rename files between runs), but I still wanted to note down the issue here for the record.

WaterWhisperer:
## Description

Fixes #2744 

When running benchmarks with `QLOGDIR` set, the server qlog filename was hardcoded as `"server"`, causing `AlreadyExists` errors when running benchmarks multiple times in the same directory.

This PR uses `tempfile::Builder` to generate unique server qlog filenames (e.g., `server-abc123.sqlog`) similar to how client qlog files already include the connection ID suffix.

## Changes

- **Added** `tempfile = { version = "3.16.0" }` to workspace dependencies (matching Firefox's dependency tree)
- **Added** `tempfile = { workspace = true }` to test-fixture dependencies
- **Modified** `test-fixture/src/lib.rs` to use `tempfile::Builder` for generating unique server qlog filenames

## Testing

- Compiles successfully: `cargo check --locked --all-targets`
- Tests pass: `cargo test --locked -p test-fixture`
- The unique filename generation prevents file collisions when creating multiple servers

## Notes

The `tempfile` version (3.16.0) matches [what Firefox uses](https://github.com/mozilla-firefox/firefox/blob/3e5b2c1c5372335ef623ceccbac4cc92462b94d6/Cargo.lock#L6506-L6518), as suggested by @mxinden in the issue discussion

### DIFF
--- a/test-fixture/src/lib.rs
+++ b/test-fixture/src/lib.rs
@@ -248,13 +248,15 @@ where
     )
     .expect("create a server");
     if let Ok(dir) = std::env::var("QLOGDIR") {
+        // Use random bytes to generate a unique name
+        let unique_name = format!("server-{}", hex(random::<10>()));
         c.set_qlog(
             Qlog::enabled_with_file(
                 dir.parse().unwrap(),
                 Role::Server,
                 Some("Neqo server qlog".to_string()),
                 Some("Neqo server qlog".to_string()),
-                "server".to_string(),
+                unique_name,
                 now(),
             )
             .unwrap(),


Cargo.toml
```diff
@@ -43,6 +43,7 @@ regex = { version = "1.9", default-features = false }
 rustc-hash = { version = "2.1", default-features = false, features = [ "std" ]}
 static_assertions = { version = "1.1", default-features = false }
 strum = { version = "0.27", default-features = false, features = ["derive"] }
+tempfile = { version = "3.16.0", default-features = false }
```
larseggert:
```suggestion
tempfile = { version = "3.16", default-features = false }
```
Might be sufficient here to use this, since `Cargo.lock` then locks it to the version used in Firefox.

test-fixture/src/lib.rs
```diff
@@ -248,13 +250,25 @@ where
...
+        let unique_name = Builder::new()
+            .prefix("server-")
+            .suffix("")
+            .tempfile_in(&dir)
+            .unwrap()
+            .path()
+            .file_stem()
+            .unwrap()
+            .to_string_lossy()
+            .to_string();
```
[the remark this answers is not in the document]
larseggert: @WaterWhisperer I think Copilot has a point here, and it would allow us to drop the `tempfile` dependency. We might not need `rand` either if we used `as_micros()`.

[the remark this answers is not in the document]
WaterWhisperer: Okay

test-fixture/src/lib.rs
```diff
@@ -248,13 +248,19 @@ where
     )
     .expect("create a server");
     if let Ok(dir) = std::env::var("QLOGDIR") {
+        // Use system time in microseconds to generate a unique name
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_micros();
```
martinthomson: I think that I would prefer that you use `hex(neqo_crypto::random::<10>())`.  That will address copilot's concern here.

WaterWhisperer: Thanks for the review!

PR #3094, opened 2025-11-01
chore: Always build an optimized nss

larseggert:
And always use all cores to do so.


mxinden (approved):

//CC @Not-Nik in case you have opinions.

PR #3095, opened 2025-11-01
fix: prevent TP caching from affecting greasing behavior

larseggert:
The fix correctly implements RFC 9287 Section 3.1's requirement that greasing not be based on remembered parameters from previous connections. For Initial and 0-RTT packets, we now require confirmation in the current handshake before allowing the fixed bit to be potentially cleared.

This fixes the handshakeloss interop test failure where picoquic correctly discarded Initial packets with the fixed bit cleared.


neqo-transport/src/connection/mod.rs
```diff
@@ -2246,15 +2246,41 @@ impl Connection {
         (pt, builder, pn)
     }
 
-    fn can_grease_quic_bit(&self) -> bool {
+    fn can_grease_quic_bit(&self, epoch: Epoch) -> bool {
```
martinthomson: I think that this is a LOT easier:

```rust
fn can_grease_quic_bit(&self) -> bool {
  let tph = self.tps.borrow();
  tph.remote_handshake().as_ref().is_some_and(|r| r.get_empty(GreaseQuicBit))
}
```

neqo-transport/src/connection/mod.rs
```diff
@@ -2664,7 +2689,17 @@ impl Connection {
         // Frames for different epochs must go in different packets, but then these
         // packets can go in a single datagram
         for space in PacketNumberSpace::iter() {
-            // Ensure we have tx crypto state for this epoch, or skip it.
+            // First peek at the epoch to determine greasing (requires immutable borrow).
```
martinthomson: Given the above, this change is also very wrong.

PR #3096, opened 2025-11-02
events: fix datagram culling index; drop oldest

MegaManSec:
check_datagram_queued used rev().enumerate() and passed the reverse index to VecDeque::remove(), which expects a forward index. This could remove the wrong element (including non-datagram events). Use position() from the front to remove the oldest Datagram, matching the comment and preventing accidental event loss/DoS.

This bug was found with [ZeroPath](https://zeropath.com).

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2199,7 +2199,12 @@ impl Connection {
         version: Version,
         grease_quic_bit: bool,
         limit: usize,
-    ) -> (packet::Type, packet::Builder<&'a mut Vec<u8>>) {
+        largest_acknowledged: Option<packet::Number>,
+    ) -> (
+        packet::Type,
+        packet::Builder<&'a mut Vec<u8>>,
+        packet::Number,
+    ) {
         let pt = packet::Type::from(epoch);
         let mut builder = if pt == packet::Type::Short {
             qdebug!("Building Short dcid {:?}", path.remote_cid());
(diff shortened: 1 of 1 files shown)


larseggert:
Thanks for the PR! Would you please also add a testcase that exhibits the bug that you are proposing a fix for here?

MegaManSec:
I do not have a test case, I'm sorry. This was noticed because the code simply did not match the comment:

    // The number of datagrams in the events queue is limited to max_queued_datagrams.
    // This function ensure this and deletes the oldest datagrams if needed.

mxinden (commented):
You are onto something here. Well done. I am curious how you found this.

Before we can merge here, we would need a test that fails without your patch and succeeds with your patch. Would you be willing to add that. You can find an example of a related test below.

https://github.com/mozilla/neqo/blob/0dc00b457f46659500ccba392b9b7bb06ef28826/neqo-transport/src/events.rs#L261-L262

For what it is worth, we also head-drop on the send path:

https://github.com/mozilla/neqo/blob/0dc00b457f46659500ccba392b9b7bb06ef28826/neqo-transport/src/quic_datagrams.rs#L171-L181

MegaManSec:
>You are onto something here. Well done. I am curious how you found this.

I scanned `neqo` with [ZeroPath](https://zeropath.com/?utm_source=joshua.hu), and AI-assisted static analysis tool, with a "bugs rule" ([sic](https://joshua.hu/llm-engineer-review-sast-security-ai-tools-pentesters#policies--rules)) and it found this automatically. I'm using [ZeroPath](https://zeropath.com/?utm_source=joshua.hu) at the moment on some interesting (for me) codebases. Basically, it's an [AI-powered vulnerability/bug scanner](https://joshua.hu/llm-engineer-review-sast-security-ai-tools-pentesters). For example in curl, I've reported (+ submitted patches) over 200 valid bugs (see [this post](https://mastodon.social/@bagder/115348489377432847) or [this article](https://www.theregister.com/2025/10/02/curl_project_swamped_with_ai/) or [this digest](https://lists.haxx.se/pipermail/daniel/2025-September/000126.html) or [this blog post](https://daniel.haxx.se/blog/2025/10/10/a-new-breed-of-analyzers/)).

MegaManSec:
i will look at rust tests later; but it may take some time.

larseggert:
I added a test in https://github.com/MegaManSec/neqo/pull/1. @mxinden note the `FIXME` in it.

mxinden:
Oh, I somehow closed here. Sorry for the trouble @MegaManSec. I will propose the patches in a new pull request. You will still be credited in the commit.

MegaManSec:
all g. thx for the patch

PR #3097, opened 2025-11-02
fix(transport): arm Handshake PTO timer when server flight is lost

larseggert:
When a client receives the Server Hello (enabling Handshake key derivation) but the server's Handshake flight (Certificate, CertificateVerify, Finished) is completely lost or corrupted, the client would timeout instead of probing for retransmission.

The root cause: `LossRecoverySpace::pto_base_time()` returned `None` for the Handshake packet number space when:
- Handshake keys were installed (from Server Hello in Initial packets)
- No Handshake packets had been sent yet (client waiting for server's Certificate)
- No Handshake packets had been received (server's flight lost)

This caused `last_ack_eliciting` to remain `None` for Handshake space, preventing the PTO timer from arming, which violated RFC 9002 Section 6.2.4: *"When a PTO timer expires, a sender MUST send at least one ack-eliciting packet in the packet number space as a probe."*

When processing Initial ACKs after multiple PTOs (count > 1), prime the Handshake space PTO baseline if no Handshake packets have been sent or received yet. This sets `last_ack_eliciting` to the current time, allowing `pto_base_time()` to return a valid baseline and enabling Handshake PTO probes.

The `count > 1` condition prevents false positives in normal handshake flow where Handshake packets arrive before the second PTO.

Added `pto_handshake_space_when_server_flight_lost` regression test that verifies Handshake PING probes are sent when server's Handshake flight is lost.

### DIFF
--- a/neqo-transport/src/recovery/mod.rs
+++ b/neqo-transport/src/recovery/mod.rs
@@ -1008,13 +1008,14 @@ mod tests {
     use neqo_common::qlog::Qlog;
     use test_fixture::{now, DEFAULT_ADDR};
 
-    use super::{LossRecoverySpace, PacketNumberSpace, SendProfile, FAST_PTO_SCALE};
+    use super::{LossRecoverySpace, PacketNumberSpace, PtoState, SendProfile, FAST_PTO_SCALE};
     use crate::{
         cid::{ConnectionId, ConnectionIdEntry},
         ecn, packet,
         path::{Path, PathRef},
         recovery::{self, sent, MAX_PTO_PACKET_COUNT},
         stats::{Stats, StatsCell},
+        tracking::PacketNumberSpaceSet,
         ConnectionParameters,
     };
 
@@ -1780,5 +1781,54 @@ mod tests {
         assert_eq!(2, MAX_PTO_PACKET_COUNT); // because we're relying on that...
         let profile = lr.send_profile(now);
         assert!(profile.pto.is_none());
+    }
+
+    fn assert_no_handshake_last_ack_eliciting(lr: &Fixture) {
+        assert!(lr
+            .spaces
+            .get(PacketNumberSpace::Handshake)
+            .and_then(|s| s.last_ack_eliciting)
+            .is_none());
+    }
+
+    #[test]
+    fn maybe_prime_handshake_pto_no_pto_state() {
+        let mut lr = Fixture::default();
+        assert!(lr.pto_state.is_none());
+
+        // Verify nothing changes - the Handshake space should not be primed afterwards.
+        lr.maybe_prime_handshake_pto(now());
+        assert_no_handshake_last_ack_eliciting(&lr);
+    }
+
+    #[test]
+    fn maybe_prime_handshake_pto_wrong_space() {
+        // Create a PTO state in Handshake space.
+        let mut lr = Fixture::default();
+        let probe_set = PacketNumberSpaceSet::only(PacketNumberSpace::Handshake);
+        lr.pto_state = Some(PtoState::new(PacketNumberSpace::Handshake, probe_set));
+
+        // Verify nothing changes - the Handshake space should not be primed afterwards.
+        lr.maybe_prime_handshake_pto(now());
+        assert_no_handshake_last_ack_eliciting(&lr);
+    }
+
+    #[test]
+    fn maybe_prime_handshake_pto_no_handshake_space() {
+        // Create a PTO state in Initial space.
+        let mut lr = Fixture::default();
+        let probe_set = PacketNumberSpaceSet::only(PacketNumberSpace::Initial);
+        lr.pto_state = Some(PtoState::new(PacketNumberSpace::Initial, probe_set));
+
+        // Set up Initial space with an ACK and drop Handshake space.
+        lr.spaces
+            .get_mut(PacketNumberSpace::Initial)
+            .unwrap()
+            .largest_acked = Some(0);
+        lr.spaces.drop_space(PacketNumberSpace::Handshake);
+
+        // Verify Handshake space still doesn't exist afterwards.
+        lr.maybe_prime_handshake_pto(now());
+        assert!(lr.spaces.get(PacketNumberSpace::Handshake).is_none());
     }
 }


neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+/// This reproduces the haproxy handshakeloss bug where:
+/// 1. Client receives Server Hello → Handshake keys installed
+/// 2. Server's Handshake flight (Certificate, etc.) is lost/corrupted
+/// 3. Client never sends any Handshake packets (has nothing to send yet)
+/// 4. Client's Handshake space has last_ack_eliciting = None
+/// 5. PTO timer never arms for Handshake space
+/// 6. Connection times out
+///
+/// RFC 9002 Section 6.2.4: "When a PTO timer expires, a sender MUST send at least
+/// one ack-eliciting packet in the packet number space as a probe."
```
martinthomson: This is not a useful addition to the comment.


```suggestion
```

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+    now += RTT / 2;
+    let s1 = server.process(c1, now).dgram();
+    assert!(s1.is_some());
+    let s2 = server.process_output(now).dgram();
+    assert!(s2.is_some());
+
+    // Now let the client have the Initial, but drop the first coalesced Handshake packet.
+    now += RTT / 2;
+    let (initial, _) = split_datagram(&s1.unwrap());
+    client.process_input(initial, now);
```
martinthomson: So this is going to cause problems when (not if) the server Initial gets too big to fit into a single datagram.

The problem here is that we are on the boundary here and while the first datagram contains Handshake packets NOW, that is not going to remain the case for long.

The complicated part is that you need to take the second packet, TEST whether it is an initial packet, and then split it conditionally.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+
+    // Process s2 to provide remaining CRYPTO frames (Server Hello completion + Handshake data)
+    // This will install Handshake keys. We drop the client's response.
+    let c2 = client.process(s2, now).dgram();
+    assert!(c2.is_some()); // This is an ACK.  Drop it.
+
+    // Verify client has reached Handshaking state (which implies Handshake keys are installed).
+    // Handshake keys are derived after processing Server Hello.
```
martinthomson: We don't need all these comments.  The test is sufficient.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+    assert_eq!(
+        *client.state(),
+        State::Handshaking,
+        "Test setup: Client must be in Handshaking state (Handshake keys installed)"
```
martinthomson:
```suggestion
```

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+    );
+
+    let delay = client.process_output(now).callback();
+
+    // Record stats before PTO
+    let initial_pto_count = client.stats.borrow().pto_counts[0];
+
+    // Advance time to trigger PTO
+    now += delay;
```
martinthomson:
```suggestion
    let pto = client.process_output(now).callback();
    now += pto;

    let initial_pto_count = client.stats.borrow().pto_counts[0];
```

martinthomson: Also, the stats here should be fetched once rather than borrowed once and retrieved again (for ping count before, below).  Just run `let before_stats = client.stats();`

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+    }
+
+    let ping_count_after = client.stats().frame_tx.ping;
+
+    // Verify at least one datagram was sent for PTO
+    assert!(!pto_dgrams.is_empty(), "Client must send PTO probe");
+
+    // Verify PTO count incremented
+    assert!(
+        client.stats.borrow().pto_counts[0] > initial_pto_count,
```
martinthomson: This should have increased by the number of packets sent.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,99 @@ fn ack_for_unsent() {
...
+    // - RFC 9002 Section 6.2.4 requires probing packet number spaces
+
+    // Verify a Handshake PING probe was sent by checking stats
+    // We can't easily parse the encrypted/header-protected packet, so check frame stats
+    let ping_count_before = client.stats().frame_tx.ping;
+
+    // Drain all PTO output
+    let mut pto_dgrams = Vec::new();
+    while let Some(dgram) = client.process_output(now).dgram() {
+        pto_dgrams.push(dgram);
```
martinthomson: This doesn't need to collect the PTO datagrams.  Just count them.  But then, you don't use that count below either...

neqo-transport/src/recovery/mod.rs
```diff
@@ -191,17 +194,25 @@ impl LossRecoverySpace {
         } else if self.space == PacketNumberSpace::ApplicationData {
             None
         } else {
-            // Nasty special case to prevent handshake deadlocks.
+            // Special case to prevent handshake deadlocks.
```
martinthomson: OK, this is a gratuitous change.

neqo-transport/src/recovery/mod.rs
```diff
@@ -654,6 +665,41 @@ impl Loss {
...
+        // Without this, pto_base_time() returns None for Handshake space and the PTO
+        // timer never arms, causing a deadlock/timeout.
+        //
+        // We need both conditions to avoid false positives:
+        // 1. largest_acked.is_none() - no Handshake packets received yet
+        // 2. Multiple PTOs fired (count > 1) - clearly stuck, not just first retry
+        //
+        // Without count > 1, we'd prime on first Initial ACK even when Handshake
+        // packet is about to arrive, affecting PTO timing in normal flow.
+        if pn_space == PacketNumberSpace::Initial {
```
martinthomson: This should be factored out if it's going to be this much.  It's only a few lines, but it's so, so much.

Also, do we need to prime the PTO on ApplicationData when Handshake ACKs are received?

neqo-transport/src/recovery/mod.rs
```diff
@@ -654,6 +665,41 @@ impl Loss {
...
+                if needs_hs_priming {
+                    if let Some(hs_space) = self.spaces.get_mut(PacketNumberSpace::Handshake) {
+                        if hs_space.last_ack_eliciting.is_none() && hs_space.largest_acked.is_none()
+                        {
+                            // Use current time as baseline - PTO will fire at now + pto_period
+                            qtrace!(
+                                "Priming Handshake PTO baseline (no HS packets after {} PTOs)",
+                                pto.count()
+                            );
+                            hs_space.last_ack_eliciting = Some(now);
```
martinthomson: This is fine, though I think we need this to be made more generic and maybe do something about the six! levels of if statement.

neqo-transport/src/recovery/mod.rs
```diff
@@ -654,6 +665,41 @@ impl Loss {
...
+        //
+        // Without count > 1, we'd prime on first Initial ACK even when Handshake
+        // packet is about to arrive, affecting PTO timing in normal flow.
+        if pn_space == PacketNumberSpace::Initial {
+            if let Some(pto) = &self.pto_state {
+                let needs_hs_priming = pto.space == PacketNumberSpace::Initial && pto.count() > 1;
+                if needs_hs_priming {
+                    if let Some(hs_space) = self.spaces.get_mut(PacketNumberSpace::Handshake) {
+                        if hs_space.last_ack_eliciting.is_none() && hs_space.largest_acked.is_none()
+                        {
```
martinthomson: Cargo mutants is having a field day here.  That suggests that maybe we want to test this a little bit.

neqo-transport/src/recovery/mod.rs
larseggert: > Also, do we need to prime the PTO on ApplicationData when Handshake ACKs are received?

IIRC we must only do PTO in the app-space once the handshake is confirmed.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+    let mut server_dgrams = Vec::new();
+    server_dgrams.push(server.process(c1, now).dgram().unwrap());
+    while let Some(dgram) = server.process_output(now).dgram() {
+        server_dgrams.push(dgram);
+    }
+    assert!(!server_dgrams.is_empty());
+
+    // Verify packet structure. The server sends:
+    // - Zero or more Initial-only datagrams (if Initial is too big)
+    // - Exactly one Initial+Handshake coalesced datagram
```
martinthomson: This is not right.  There are cases where you do not send a datagram that has both Initial and Handshake (that's true for most normal handshakes today).

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+        let (first, second) = split_datagram(dgram);
+
+        // Check if first packet is Initial or Handshake
+        let first_is_initial = is_initial(&first, false);
+        let first_is_handshake = is_handshake(&first);
+        assert!(first_is_initial || first_is_handshake);
+
+        // Check if there's a coalesced second packet
+        if let Some(s) = &second {
+            let second_is_handshake = is_handshake(s);
```
martinthomson: This temporary is unnecessary.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+        // Verify all datagrams after Initial+Handshake are Handshake-only
+        if let Some(init_hs_idx) = initial_plus_handshake_idx {
+            if i > init_hs_idx {
+                assert!(is_handshake(&first) && second.is_none());
+            }
+        }
+    }
+
+    let initial_plus_handshake_idx = initial_plus_handshake_idx
+        .expect("Expected at least one Initial+Handshake coalesced datagram");
```
martinthomson: See above.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+    while let Some(dgram) = server.process_output(now).dgram() {
+        server_dgrams.push(dgram);
+    }
+    assert!(!server_dgrams.is_empty());
+
+    // Verify packet structure. The server sends:
+    // - Zero or more Initial-only datagrams (if Initial is too big)
+    // - Exactly one Initial+Handshake coalesced datagram
+    // - Zero or more Handshake-only datagrams (rest of the flight)
+    let mut initial_plus_handshake_idx = None;
```
martinthomson: This code is a bit long.  I think that you can do something like this:

```rust
let mut found_hs = false;
let filtered_dgrams = dgrams.into_iter().filter_map(|d| {
  let (d, h) = split_datagram(&d);
  if is_initial(d) {
    assert!(!found_hs, "got Initial after Handshake");
    found_hs &&= h.is_some_and(is_handshake);
    Some(d)
  } else if is_handshake(d) {
    if !found_hs {
      found_hs = true;
      None
    } else {
      Some(d)
    }
  } else {
    None
  }
}).flatten();
```
You can then feed those datagrams to `process_input()`

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+    }
+
+    // Client processes the Initial packets (which install Handshake keys).
+    // We drop the client's ACK response.
+    let c2 = client.process_output(now).dgram();
+    assert!(c2.is_some()); // This is an ACK.  Drop it.
+    assert_eq!(*client.state(), State::Handshaking);
+
+    let pto = client.process_output(now).callback();
+    now += pto;
```
martinthomson, line 1009: You need to check that this is non-zero.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+
+    // Fire Initial PTO - this should prime Handshake space
+    while client.process_output(now).dgram().is_some() {}
+
+    // Advance time - the next PTO should be for Handshake space (if primed).
+    let next_timeout = client.process_output(now).callback();
```
martinthomson: This is a duplicative call.  The call above that results in loop termination should provide this timeout.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,126 @@ fn ack_for_unsent() {
...
+    now += next_timeout;
+
+    // Collect all packets from this timeout
+    let mut pto_packets = Vec::new();
+    while let Some(dgram) = client.process_output(now).dgram() {
+        pto_packets.push(dgram);
+    }
+
+    // Check if any Handshake PING probes packets were sent.
```
martinthomson, line 1027: This doesn't check that PING was sent at all.

martinthomson (changes requested):

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,97 @@ fn ack_for_unsent() {
...
+    assert_ne!(pto, Duration::ZERO);
+    now += pto;
+
+    // Drain all packets and get the next timeout.
+    let next_timeout = loop {
+        let out = client.process_output(now);
+        let callback = out.callback();
+        if out.dgram().is_none() {
+            break callback;
+        }
```
martinthomson:
```suggestion
        if let Some(callback) = client.process_output(now).callback();
            break callback;
        }
```
So much simpler is possible.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,97 @@ fn ack_for_unsent() {
...
+    };
+    assert_ne!(next_timeout, Duration::ZERO);
```
martinthomson: This cannot fail.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -944,3 +944,97 @@ fn ack_for_unsent() {
...
+    }
+
+    // Check if any Handshake PING probes packets were sent.
+    let mut has_handshake = false;
+    for dgram in &pto_packets {
+        let (first, second) = split_datagram(dgram);
+        has_handshake |= is_handshake(&first);
+        if let Some(s) = &second {
+            has_handshake |= is_handshake(s);
+        }
```
martinthomson:
```suggestion
        has_handshake |= is_handshake(&first) ||
        second.as_ref().is_some_and(is_handshake);
```

neqo-transport/src/recovery/mod.rs
```diff
@@ -570,6 +581,39 @@ impl Loss {
...
+    /// Prime the Handshake space PTO timer when stuck in Initial space.
+    fn maybe_prime_handshake_pto(&mut self, now: Instant) {
+        let Some(pto) = &self.pto_state else {
+            return;
+        };
+
+        // Only prime if we're in Initial space (fire on first PTO).
+        if pto.space != PacketNumberSpace::Initial {
+            return;
+        }
```
martinthomson: You can maybe use is_some_and here.  And for the next.

larseggert:
> Is Claude making you lazy @larseggert ?

No Claude involved in the recent changes :-(

neqo-transport/src/connection/tests/recovery.rs
larseggert: IIRC you asked me to check that is not zero earlier?

neqo-transport/src/connection/tests/recovery.rs
larseggert: Doesn't quite work, because `callback()` returns `Duration`, but let me see what I can do.

neqo-transport/src/recovery/mod.rs
```diff
@@ -863,6 +908,11 @@ impl Loss {
         if let Some(pn_space) = pto_space {
             qtrace!("[{self}] PTO {pn_space}, probing {allow_probes:?}");
             self.fire_pto(pn_space, allow_probes, now);
+
+            // Maybe prime the Handshake PTO when PTO fires in Initial space.
+            if pn_space == PacketNumberSpace::Initial {
+                self.maybe_prime_handshake_pto(now);
+            }
```
mxinden, line 915: This might prime the handshake PTO even if we have no Handshake keys available, right?

Sample scenario:

1. Client sends ClientHello split across two Initial packets
2. Server receives only the first packet, cannot complete TLS, sends a bare ACK.
3. Client receives the ACK for second packet.
4. Client Initial PTO fires
5. `maybe_prime_handshake_pto` primes handshake PTO even though the client has no Handshake TX keys (ServerHello was never received).
6. Handshake PTO fires (potentially in the past), but we can not act on it.

Is the above correct @larseggert? What do you think of only calling `maybe_prime_handshake_pto` if we have handshake space keys to actually send anything.

This would fix https://github.com/mozilla/neqo/pull/3408 and thus also fix (at least one case of) https://bugzilla.mozilla.org/show_bug.cgi?id=2001901.

larseggert: I think you are correct, that change should be conditional on that. Want to do a PR?

PR #3101, opened 2025-11-03
fix(transport): fix Initial+Handshake padding

larseggert:
Fixes a bug where PTO retransmissions containing both Initial and Handshake packets were not padded to the minimum 1200 bytes required by RFC 9000 § 14.1, causing servers to reject undersized datagrams.

### DIFF
--- a/neqo-transport/src/connection/tests/keys.rs
+++ b/neqo-transport/src/connection/tests/keys.rs
@@ -362,6 +362,8 @@ fn automatic_update_write_keys_blocked() {
 
 /// Test that when both Initial and Handshake packets are sent together due to PTO,
 /// the resulting datagram is properly padded to `MIN_INITIAL_PACKET_SIZE` (1200 bytes).
+///
+/// See RFC 9000 14.1 <https://www.rfc-editor.org/rfc/rfc9000.html#name-initial-datagram-size>.
 #[test]
 fn initial_handshake_pto_padding() {
     let mut client = default_client();


neqo-transport/src/connection/tests/keys.rs
```diff
@@ -356,3 +359,46 @@ fn automatic_update_write_keys_blocked() {
         State::Closed(CloseReason::Transport(Error::KeysExhausted))
     ));
 }
+
+/// Test that when both Initial and Handshake packets are sent together due to PTO,
+/// the resulting datagram is properly padded to `MIN_INITIAL_PACKET_SIZE` (1200 bytes).
```
mxinden, line 364:
```suggestion
/// the resulting datagram is properly padded to `MIN_INITIAL_PACKET_SIZE` (1200 bytes).
///
/// See RFC 9000 14.1 <https://www.rfc-editor.org/rfc/rfc9000.html#name-initial-datagram-size>.
```

mxinden (approved):
Good catch. This looks good to me.

neqo-transport/src/connection/mod.rs
```diff
@@ -2765,7 +2765,11 @@ impl Connection {
                 initial_sent = Some(sent);
                 needs_padding = true;
             } else {
-                if pt == packet::Type::Handshake && self.role == Role::Client {
+                if pt == packet::Type::Handshake
```
martinthomson, line 2768: I realize, looking at this, that we might want this instead.


```suggestion
                if pt.is_long()
```

neqo-transport/src/connection/tests/keys.rs
```diff
@@ -356,3 +359,48 @@ fn automatic_update_write_keys_blocked() {
...
+///
+/// See RFC 9000 14.1 <https://www.rfc-editor.org/rfc/rfc9000.html#name-initial-datagram-size>.
+#[test]
+fn initial_handshake_pto_padding() {
```
martinthomson, line 368: It's not clear to me that this test case will reliably produce the conditions for the error.  That is because the server flight does not always include a coalesced Initial and Handshake packet.  We're betting here that there is a packet with small Handshake and small Initial such that you need to pad to the MTU.

martinthomson:
OK, I didn't notice that this was merged.  I don't think that this work is done.  See my review.

neqo-transport/src/connection/mod.rs
martinthomson: For clarity, this would be a problem with 0-RTT as well.

PR #3102, opened 2025-11-03
feat(transport): implement RESET_STREAM_AT frame

mxinden:
Implementation of https://datatracker.ietf.org/doc/draft-ietf-quic-reliable-stream-reset/.

---

Draft only for now.

### DIFF
--- a/neqo-transport/src/tparams.rs
+++ b/neqo-transport/src/tparams.rs
@@ -6,6 +6,11 @@
 
 // Transport parameters. See -transport section 7.3.
 
+// TODO: When using 0-RTT, both endpoints MUST remember the value of this
+// transport parameter. This allows use of this extension in 0-RTT packets. When
+// the server accepts 0-RTT data, the server MUST NOT disable this extension on
+// the resumed connection.
+
 use std::{
     cell::RefCell,
     fmt::{self, Display, Formatter},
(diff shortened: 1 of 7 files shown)


neqo-transport/src/tparams.rs
```diff
@@ -6,6 +6,11 @@
 
 // Transport parameters. See -transport section 7.3.
 
+// TODO: When using 0-RTT, both endpoints MUST remember the value of this
+// transport parameter. This allows use of this extension in 0-RTT packets. When
+// the server accepts 0-RTT data, the server MUST NOT disable this extension on
+// the resumed connection.
```
martinthomson, line 12: This is already handled.  By "default", our check in `ok_for_0rtt()` checks if the new connection includes the transport parameter when the old connection did.

neqo-transport/src/tparams.rs
```diff
@@ -54,6 +59,7 @@ pub enum TransportParameterId {
     GreaseQuicBit = 0x2ab2,
     MinAckDelay = 0xff02_de1a,
     MaxDatagramFrameSize = 0x0020,
+    ResetStreamAt = 0x17f7586d2cb571,
```
martinthomson, line 62:
```suggestion
    ResetStreamAt = 0x24,
```

martinthomson (commented):
Following a link, I find this.  No substance here yet, but a reasonable start.

larseggert:
Closing in favor of #3754

PR #3110, opened 2025-11-05
chore: `idna-adapter@1.2.1`

larseggert:


larseggert:
Needs MSRV >= 1.82.

larseggert:
OBE with #3339 or #3315

PR #3117, opened 2025-11-07
fix(transport/pace): accumulate debt when below granularity

MarcoPolo:
While evaluating neqo’s flow control autotuning, I found an issue with pacing within the simulator. This mostly affects the simulator, but may also behave poorly in the real world when sending many packets at once within a millisecond. Remember that a 50 mbps connection will send ~6,000 packets a millisecond.

If a packet needs to wait less than 1ms (the granularity constant), the pacer code tells it to wait no time (this makes sense due to system timer resolution). The issue is that, in the absence of something else that moves time forward, an unlimited number of packets can be sent in the same instant. This issue is particularly visible in the simulator, where time only advances when we tell it to.

For example, say we had ~ 12,000 packets to send out on a 50 mbps connection. Every individual packet is below the granularity, but overall we should expect to advance time by around 2ms. The current code would not do this, and instead send all 12k packets at the current time.

I’ve included a test case to illustrate the issue.

I’ve also included a small patch that allows the pacer code to accrue debt that will be paid off when we are above the granularity.

For the existing min_bandwidth simulation:

Before: 327.30050331683793 Mbit/s After: 481.3917733802056 Mbit/s

Looking at the qlog+qvis (for a smaller version of this test, the gigabyte version is too big for qvis), the pace looks better.


neqo-transport/src/pace.rs
```diff
@@ -192,4 +204,25 @@ mod tests {
             "Expect packet to be sent immediately, instead of being paced below timer granularity"
         );
     }
+
+    #[test]
+    fn have_many_packets_large_cwnd() {
+        const SHORT_RTT: Duration = Duration::from_millis(100);
```
mxinden: nit: not sure I would consider 100ms a _short_ RTT.

neqo-transport/src/pace.rs
```diff
@@ -67,23 +67,28 @@ impl Pacer {
...
+        self.p = min(mtu, isize::MAX as usize);
     }
 
     /// Determine when the next packet will be available based on the provided RTT
     /// and congestion window.  This doesn't update state.
     /// This returns a time, which could be in the past (this object doesn't know what
     /// the current time is).
     pub fn next(&self, rtt: Duration, cwnd: usize) -> Instant {
-        if self.c >= self.p {
+        let credit = i128::try_from(self.c).expect("isize fits into i64");
```
mxinden: The following would make the `credit` variable obsolete. What do you think?

``` diff
diff --git a/neqo-transport/src/pace.rs b/neqo-transport/src/pace.rs
index c30f8edc..e004dd3a 100644
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -75,10 +75,9 @@ impl Pacer {
     /// This returns a time, which could be in the past (this object doesn't know what
     /// the current time is).
     pub fn next(&self, rtt: Duration, cwnd: usize) -> Instant {
-        let credit = i128::try_from(self.c).expect("isize fits into i64");
-        let packet = i128::try_from(self.p).expect("packet size fits into i64");
+        let packet = isize::try_from(self.p).expect("packet size fits into isize");
 
-        if credit >= packet {
+        if self.c >= packet {
             qtrace!("[{self}] next {cwnd}/{rtt:?} no wait = {:?}", self.t);
             return self.t;
         }
@@ -87,7 +86,7 @@ impl Pacer {
         // self.t + rtt * (self.p - self.c) / (PACER_SPEEDUP * cwnd)
         let r = rtt.as_nanos();
         let deficit =
-            u128::try_from(packet - credit).expect("packet is larger than current credit");
+            u128::try_from(packet - self.c).expect("packet is larger than current credit");
         let d = r.saturating_mul(deficit);
         let add = d / u128::try_from(cwnd * PACER_SPEEDUP).expect("usize fits into u128");
         let w = u64::try_from(add).map(Duration::from_nanos).unwrap_or(rtt);
```

neqo-transport/src/pace.rs
```diff
@@ -121,7 +126,14 @@ impl Pacer {
             .unwrap_or(self.m);
 
         // Add the capacity up to a limit of `self.m`, then subtract `count`.
-        self.c = min(self.m, (self.c + incr).saturating_sub(count));
+        let incr_isize = isize::try_from(incr).unwrap_or(isize::MAX);
+        let count_isize = isize::try_from(count).unwrap_or(isize::MAX);
```
mxinden: Nit pick: I would inline these.

``` diff
@@ -126,13 +125,11 @@ impl Pacer {
             .unwrap_or(self.m);
 
         // Add the capacity up to a limit of `self.m`, then subtract `count`.
-        let incr_isize = isize::try_from(incr).unwrap_or(isize::MAX);
-        let count_isize = isize::try_from(count).unwrap_or(isize::MAX);
         self.c = min(
             isize::try_from(self.m).unwrap_or(isize::MAX),
             self.c
-                .saturating_add(incr_isize)
-                .saturating_sub(count_isize),
+                .saturating_add(isize::try_from(incr).unwrap_or(isize::MAX))
+                .saturating_sub(isize::try_from(count).unwrap_or(isize::MAX)),
         );
         self.t = now;
```

neqo-transport/src/pace.rs
```diff
@@ -67,23 +67,28 @@ impl Pacer {
...
     pub fn set_mtu(&mut self, mtu: usize) {
-        self.p = mtu;
+        self.p = min(mtu, isize::MAX as usize);
     }
 
     /// Determine when the next packet will be available based on the provided RTT
     /// and congestion window.  This doesn't update state.
     /// This returns a time, which could be in the past (this object doesn't know what
     /// the current time is).
     pub fn next(&self, rtt: Duration, cwnd: usize) -> Instant {
```
mxinden: How about documenting the fact, that `Pacer` might accumulate dept?

```suggestion
    /// Determine when the next packet will be available based on the provided
    /// RTT, provided congestion window and acccumulated credit or debt.  This
    /// doesn't update state.  This returns a time, which could be in the past
    /// (this object doesn't know what the current time is).
    pub fn next(&self, rtt: Duration, cwnd: usize) -> Instant {
```

neqo-transport/src/pace.rs
```diff
@@ -67,23 +67,28 @@ impl Pacer {
     }
 
     pub fn set_mtu(&mut self, mtu: usize) {
-        self.p = mtu;
+        self.p = min(mtu, isize::MAX as usize);
```
mxinden:
```suggestion
        self.p = mtu;
```

Needed?

MarcoPolo:
thank you for the review @mxinden. I've amended the commits with your changes. As a reminder you can view just the difference with `git range-diff  af02a79...5632e900`

neqo-transport/src/pace.rs
```diff
@@ -57,8 +57,8 @@ impl Pacer {
             enabled,
             t: now,
             m,
-            c: m,
-            p,
+            c: isize::try_from(m).unwrap_or(isize::MAX),
+            p: min(p, isize::MAX as usize),
```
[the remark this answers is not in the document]
mxinden: Comment is wrong. `usize` can exceed `isize::MAX`.

mxinden (approved):
Pull request looks good to me.

I will give @martinthomson another day, in case he wants to review. Will merge after that.

Thank you Marco!

MarcoPolo:
In case it's helpful for future context, here is the PR that identifies and fixes the flow control sizing issue I mentioned in the description of this PR: https://github.com/mozilla/neqo/pull/3125

Issue #2843, opened 2025-08-11
Should Neqo support the HTTP DATAGRAM Capsule?


mxinden:
> [3.5. ](https://www.rfc-editor.org/rfc/rfc9297.html#section-3.5)[The DATAGRAM Capsule](https://www.rfc-editor.org/rfc/rfc9297.html#name-the-datagram-capsule)
>
> This document defines the DATAGRAM (0x00) Capsule Type. This Capsule allows HTTP Datagrams to be sent on a stream using the Capsule Protocol. This is particularly useful when HTTP is running over a transport that does not support the QUIC DATAGRAM frame.

https://www.rfc-editor.org/rfc/rfc9297.html#section-3.5

One use-case would be proxying an HTTP/3 connection over a MASQUE connect-udp HTTP/3 proxy connection (https://github.com/mozilla/neqo/pull/2796), where the proxy connection does not support the [QUIC datagram extension](https://www.rfc-editor.org/rfc/rfc9221.html).

Asked differently, do we expect wide deployments of MASQUE connect-udp proxy servers without the QUIC datagram extension?

Note that this is not relevant for WebTransport datagrams, as WebTransport requires WebTransport datagrams to be sent via QUIC datagrams:

> WebTransport over HTTP/3 also requires support for QUIC datagrams. To indicate support, both the client and the server MUST send a max_datagram_frame_size transport parameter with a value greater than 0 (see [Section 3](https://rfc-editor.org/rfc/rfc9221#section-3) of [[QUIC-DATAGRAM](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#QUIC-DATAGRAM)]).
>
> Any WebTransport requests sent by the client without enabling QUIC and HTTP datagrams MUST be treated as malformed by the server, as described in [Section 4.1.2](https://rfc-editor.org/rfc/rfc9114#section-4.1.2) of [[HTTP3](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#HTTP3)].

https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#section-3.1

larseggert:
> Asked differently, do we expect wide deployments of MASQUE connect-udp proxy servers without the QUIC datagram extension?

Do we know what Fastly supports? I'd prioritize it accordingly.

martinthomson:
[...]

PR #3122, opened 2025-11-10
feat(http3): support the HTTP DATAGRAM Capsule

WaterWhisperer:
Related issue: #2843


neqo-http3/src/frames/connect_udp_frame.rs
```diff
@@ -4,42 +4,113 @@
...
+
+use super::{capsule::Capsule, hframe::HFrameType};
 use crate::{frames::reader::FrameDecoder, Res};
 
+pub const CAPSULE_TYPE_DATAGRAM: HFrameType = HFrameType(0x00);
+
 #[derive(PartialEq, Eq, Debug)]
 pub enum Frame {
-    // TODO: Implement HTTP Datagram <https://github.com/mozilla/neqo/issues/2843>.
+    Datagram { payload: Vec<u8> },
```
larseggert: Would there be a benefit in using `neqo_common::Bytes` here?

WaterWhisperer: I tried doing it this way, but it seems it would require a new conversion. Perhaps I can try it again.

WaterWhisperer: This time it's okay. Thanks for the review!

neqo-http3/src/frames/capsule.rs
```diff
@@ -0,0 +1,229 @@
...
+                .decode(capsule_length_usize)
+                .ok_or(crate::Error::HttpFrame)?
+                .to_vec();
+            Ok(Some(Self::Datagram {
+                payload: Bytes::from(payload),
+            }))
+        } else {
+            decoder.skip(capsule_length_usize);
+            Ok(None)
+        }
```
mxinden: Preference for this being an early return, as you do above, instead of an `else`.

In addition, mind adding a `debug!` log?

neqo-http3/src/frames/capsule.rs
```diff
@@ -0,0 +1,229 @@
...
+    #[test]
+    fn encode_datagram_capsule_empty_payload() {
+        let capsule = Capsule::Datagram {
+            payload: Bytes::from(Vec::new()),
+        };
+        let mut enc = Encoder::default();
+        capsule.encode(&mut enc);
+        let encoded = enc.as_ref();
+
+        assert!(encoded.len() > 1);
```
mxinden: I suggest moving the `        assert_eq!(encoded.len(), 2);` check from below here.

neqo-http3/src/frames/capsule.rs
```diff
@@ -0,0 +1,229 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[cfg(test)]
+use neqo_common::Encoder;
+use neqo_common::{Bytes, Decoder};
+
+use crate::Res;
+
+pub const CAPSULE_TYPE_DATAGRAM: u64 = 0x00;
```
mxinden: would it make sense to re-use the constant from `connect_udp_frame.rs` here?

mxinden (commented):
Well done!

I believe we would now as well need to send the `capsule-protocol` HTTP header on the EXTENDED_CONNECT stream.

https://www.rfc-editor.org/rfc/rfc9297#section-3.4

As far as I understand, this allows a user to send a capsule to us.

In addition, could you add an end-to-end test showcasing how a neqo-client and a neqo-server (neither supporting QUIC datagrams) will exchange (proxy) UDP datagrams using connect-udp and HTTP DATAGRAM capsules? `session_lifecycle` in the file below should be a good example.

https://github.com/mozilla/neqo/blob/681a9b33fbb7a4c9868c42b2ea05da00022d320b/neqo-http3/tests/connect_udp.rs#L286-L289

mxinden:
@WaterWhisperer let me know if you need any help with the above. I saw you resolved all comments but didn't make any changes yet.

WaterWhisperer:
> @WaterWhisperer let me know if you need any help with the above. I saw you resolved all comments but didn't make any changes yet.

@mxinden  Thanks for your attention! Yeah, I've resolved all the comments. However, when I was adding 

> an end-to-end test showing how a neqo-client and a neqo-server (neither supporting QUIC datagrams) will exchange (proxy) UDP datagrams using connect-udp and HTTP DATAGRAM capsules

, I found the test wasn't working. I then realized that my implementation is incomplete

Currently, my changes support:
- Decoding HTTP DATAGRAM Capsules from the stream (receiving side works)
- The capsule-protocol header in responses

However, the sending side still only uses QUIC datagrams. I haven't implemented sending via HTTP DATAGRAM Capsules on the stream yet. I'm trying my best on completing this to fully meet the RFC 9297 requirements.

I'll update the PR once I have a working end-to-end test. Thanks again for your follow-up!

mxinden:
Great work. I appreciate the help!

WaterWhisperer:
I've encountered a problem: I'm trying to send a Capsule using SendMessage::send_data(), but this method automatically wraps it in a DATA frame.
The final data structure becomes [DATA frame][Capsule], and the receiving end's FrameReader will parse the DATA frame and treat the contents of the Capsule as the frame payload.
My solution was to bypass the HTTP/3 framing layer and read/write raw bytes directly on the stream, manually handling Capsule encoding/decoding. However, this made the `connect_udp_frame` dead_code, which intuitively tells me is wrong, but I can't think of a better way.

larseggert:
@mxinden is busy at the moment, so please don't take his silence as us having lost interest here. Will take us a few more days to respond. Thanks again for your help!

mxinden:
@WaterWhisperer sorry for the delay here. Good catch. I didn't think of this before.

Instead of using `SendMessage::send_data`, how about using `<SendMessage as SendStream>::send_data_atomic`?

https://github.com/mozilla/neqo/blob/c2caaa235be91e98f2b1cae90e66fcec7f6d341b/neqo-http3/src/send_message.rs#L298-L306

Our WebTransport implementation has the same use-case, namely to send its custom-framed control messages. Here is how WebTransport sends a closing frame:

https://github.com/mozilla/neqo/blob/c2caaa235be91e98f2b1cae90e66fcec7f6d341b/neqo-http3/src/features/extended_connect/session.rs#L360-L363

Does that help @WaterWhisperer?

neqo-http3/src/frames/connect_udp_frame.rs
mxinden, line 1: Note @WaterWhisperer that I added this file back then as a placeholder.

In case you think there is a better way to structure this, e.g. implement `FrameDecoder` in `capsule.rs` instead, that is totally fine. I was just mirroring our WebTransport implementation back then, which in hindsight, might not have been the best design.

WaterWhisperer:
> @mxinden is busy at the moment, so please don't take his silence as us having lost interest here. Will take us a few more days to respond. Thanks again for your help!

@larseggert Thank you so much for your reassurance! I completely understand that everyone is busy with work and life (Just like I am busy with final exams :) ), and I really appreciate the time you and the team dedicate to maintaining this project. I've learned a lot from contributing to **neqo** and interacting with you all. Please take your time, and no worries at all about the delay.

neqo-http3/src/frames/connect_udp_frame.rs
WaterWhisperer: @mxinden Thank you for the pointer! `send_data_atomic` looks like a promising direction.

Regarding `connect_udp_frame.rs`, knowing it was intended as a placeholder clears up my confusion. My initial struggle was that `FrameReader` interprets the `0x00` type byte as a DATA frame (which is correct for HTTP/3), but I was trying to make it parse `ConnectUdpFrame` directly.

I will try to refactor the implementation to:
1. Use `send_data_atomic` to properly send Capsules wrapped in DATA frames (compliant with RFC 9297).
2. On the receiving side, handle the DATA frame payload and decode the Capsules from it.
3. See if I can repurpose or restructure `connect_udp_frame.rs` / `capsule.rs` to handle this decoding logic cleanly, as you suggested.

I'll update the PR once I have this working. Thanks again for the guidance!

neqo-http3/src/connection.rs
```diff
@@ -1565,7 +1572,7 @@ impl Http3Connection {
             .extended_connect_session()
             .ok_or(Error::InvalidStreamId)?
             .borrow_mut()
-            .send_datagram(conn, buf, id)
+            .send_datagram(conn, buf, id, Instant::now())
```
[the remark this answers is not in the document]
larseggert: > The now timestamp should be passed through from the caller

Copilot is correct here.

neqo-http3/src/frames/connect_udp_frame.rs
mxinden: > On the receiving side, handle the DATA frame payload and decode the Capsules from it.

@WaterWhisperer here is an example how our webtransport reads from the stream. I don't think it will use Data frames.

https://github.com/mozilla/neqo/blob/c2caaa235be91e98f2b1cae90e66fcec7f6d341b/neqo-http3/src/features/extended_connect/webtransport_session.rs#L94-L135

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -377,26 +377,46 @@ impl Session {
...
+        dgram_data.encode_varint(self.id.as_u64() / 4);
+        self.protocol.write_datagram_prefix(&mut dgram_data);
+        dgram_data.encode(buf);
+
+        match conn.send_datagram(dgram_data.into(), id) {
+            Ok(()) => {
+                qtrace!("[{self}] sent datagram via QUIC datagram");
+                Ok(())
+            }
+            Err(e) => {
```
mxinden: I don't think it is safe to fallback to capsules on all errors. I think we should only send capsules, if the remote did not negotiate the QUIC datagram extension, or if the remote set the max datagram size to `0`.

https://www.rfc-editor.org/rfc/rfc9221.html#section-3

Note also this paragraph from the connect-udp draft:

> If a UDP proxy is using QUIC DATAGRAM frames and it receives a UDP payload from the target that will not fit inside a QUIC DATAGRAM frame, the UDP proxy SHOULD NOT send the UDP payload in a DATAGRAM capsule, as that defeats the end-to-end unreliability characteristic that methods such as Datagram Packetization Layer PMTU Discovery (DPLPMTUD) depend on [[DPLPMTUD](https://www.rfc-editor.org/rfc/rfc9298.html#RFC8899)].

https://www.rfc-editor.org/rfc/rfc9298.html#section-6.1

You might have to expose the `remote_datagram_size` function to `neqo-http3` in order to check here whether it is `0` or not.

https://github.com/mozilla/neqo/blob/c2caaa235be91e98f2b1cae90e66fcec7f6d341b/neqo-transport/src/quic_datagrams.rs#L93-L95

neqo-http3/tests/connect_udp.rs
```diff
@@ -493,3 +494,176 @@ fn connect_udp_operation_on_fetch_stream() {
...
 }
+
+#[test]
+fn session_lifecycle_with_http_datagram_capsule() {
+    fixture_init();
+    neqo_common::log::init(None);
+
+    let conn_params = ConnectionParameters::default()
+        .pmtud(true)
+        .datagram_size(1)
```
mxinden: I suggest setting it to `0`, thus indicating to the remote that this node does not support the QUIC datagram extension and thus needs to communicate via capsules instead.

neqo-http3/src/frames/capsule.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use neqo_common::{qdebug, Bytes, Encoder};
+
+use super::{hframe::HFrameType, reader::FrameDecoder};
+use crate::Res;
+
+pub const CAPSULE_TYPE_DATAGRAM: HFrameType = HFrameType(0x00);
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum Capsule {
```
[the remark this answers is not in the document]
WaterWhisperer: I'm not sure if the comments for this part are necessary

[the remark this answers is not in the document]
larseggert: Feel free to ignore any Copilot comments that don't make sense to you. Copilot is overly ambitious and wrong a lot.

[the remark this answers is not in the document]
WaterWhisperer: > Feel free to ignore any Copilot comments that don't make sense to you. Copilot is overly ambitious and wrong a lot.

Understood. Thanks!

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -104,6 +111,23 @@ impl Protocol for Session {
...
+
+    fn write_datagram_capsule(
+        &self,
+        control_stream_send: &mut Box<dyn SendStream>,
+        conn: &mut Connection,
+        buf: &[u8],
+        now: Instant,
+    ) -> Res<bool> {
+        let capsule = Capsule::Datagram {
+            payload: Bytes::from(buf.to_vec()),
```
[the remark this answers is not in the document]
WaterWhisperer: the `Bytes` here is not the crate `Bytes` so it doesn't have `copy_from_slice` method, do we need to create a `copy_from_slice` method for `Bytes` in the codebase?

WaterWhisperer:
Sorry for my late follow-up, I was busy with a few final exams some time ago

neqo-http3/src/connection_client.rs
```diff
@@ -832,8 +832,13 @@ impl Http3Client {
...
     ) -> Res<()> {
         qtrace!("webtransport_send_datagram session:{session_id:?}");
-        self.base_handler
-            .webtransport_send_datagram(session_id, &mut self.conn, buf, id)
+        self.base_handler.webtransport_send_datagram(
+            session_id,
+            &mut self.conn,
+            buf,
+            id,
+            Instant::now(),
```
mxinden: We try to reduce side effects within `neqo_*`. Ideally we do all side effects at the outskirts of the code. We consider time a side-effect. Instead of calling `Instant::now()` here, can you pass in time, i.e. have `webtransport_send_datagram` take a parameter `now: Instant`? See other functions in this file doing the same.

neqo-http3/src/connection_client.rs
```diff
@@ -850,8 +855,13 @@ impl Http3Client {
...
     ) -> Res<()> {
         qtrace!("connect_udp_send_datagram session:{session_id:?}");
-        self.base_handler
-            .connect_udp_send_datagram(session_id, &mut self.conn, buf, id)
+        self.base_handler.connect_udp_send_datagram(
+            session_id,
+            &mut self.conn,
+            buf,
+            id,
+            Instant::now(),
```
mxinden: Same here, can this be provided to the function?

neqo-http3/src/connection_server.rs
```diff
@@ -263,7 +263,7 @@ impl Http3ServerHandler {
     ) -> Res<()> {
         self.needs_processing = true;
         self.base_handler
-            .webtransport_send_datagram(session_id, conn, buf, id)
+            .webtransport_send_datagram(session_id, conn, buf, id, Instant::now())
```
mxinden: Same here. Can this be provided to the function?

neqo-http3/src/connection_server.rs
```diff
@@ -275,7 +275,7 @@ impl Http3ServerHandler {
     ) -> Res<()> {
         self.needs_processing = true;
         self.base_handler
-            .connect_udp_send_datagram(session_id, conn, buf, id)
+            .connect_udp_send_datagram(session_id, conn, buf, id, Instant::now())
```
mxinden: Same here. Can this be provided to the function?

neqo-http3/src/frames/capsule.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+impl FrameDecoder<Self> for Capsule {
+    fn decode(frame_type: HFrameType, _frame_len: u64, data: Option<&[u8]>) -> Res<Option<Self>> {
+        if frame_type == CAPSULE_TYPE_DATAGRAM {
+            if let Some(payload) = data {
+                qdebug!("Decoded Datagram Capsule len={}", payload.len());
+                return Ok(Some(Self::Datagram {
+                    payload: Bytes::from(payload.to_vec()),
```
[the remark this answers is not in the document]
WaterWhisperer: ditto

neqo-http3/src/connection_client.rs
WaterWhisperer: Okay

neqo-http3/tests/connect_udp.rs
```diff
@@ -493,3 +496,176 @@ fn connect_udp_operation_on_fetch_stream() {
...
+#[test]
+fn session_lifecycle_with_http_datagram_capsule() {
+    fixture_init();
+    neqo_common::log::init(None);
+
+    let conn_params = ConnectionParameters::default()
+        .pmtud(true)
+        .datagram_size(0)
+        .outgoing_datagram_queue(0)
+        .incoming_datagram_queue(0);
```
mxinden: I don't think setting the queue limits is necessary. `.datagram_size(0)` should be enough.

neqo-http3/tests/connect_udp.rs
```diff
@@ -493,3 +496,176 @@ fn connect_udp_operation_on_fetch_stream() {
...
     );
 }
+
+#[test]
+fn session_lifecycle_with_http_datagram_capsule() {
+    fixture_init();
+    neqo_common::log::init(None);
+
+    let conn_params = ConnectionParameters::default()
+        .pmtud(true)
```
mxinden: Is this necessary? Given that we use capsules in this test, the PMTUD shouldn't matter?

neqo-http3/tests/connect_udp.rs
```diff
@@ -493,3 +496,176 @@ fn connect_udp_operation_on_fetch_stream() {
...
+            }) = event
+            {
+                Some((session.stream_id(), datagram))
+            } else {
+                None
+            }
+        })
+        .unwrap();
+    assert_eq!(session_id, id);
+    assert_eq!(&datagram, PING);
```
mxinden, line 595: How about asserting that no QUIC datagram has been sent, i.e. `FrameStats::datagram` is `0`?

https://github.com/mozilla/neqo/blob/e28dde17d25e12e0ea9a954700e0cfb3560f19ad/neqo-transport/src/stats.rs#L54

neqo-http3/src/conn_params.rs
```diff
@@ -147,8 +147,7 @@ impl Http3Parameters {
     #[must_use]
     pub fn get_http3_datagram(&self) -> bool {
         if self.http3_datagram && self.conn_params.get_datagram_size() == 0 {
-            qwarn!("HTTP/3 setting SETTINGS_HTTP3_DATAGRAM is enabled but QUIC transport parameter max_datagram_frame_size is 0.");
-            debug_assert!(false);
+            qwarn!("HTTP/3 setting SETTINGS_H3_DATAGRAM is enabled but QUIC transport parameter max_datagram_frame_size is 0. Datagrams will be sent via HTTP DATAGRAM Capsules.");
```
mxinden:
```suggestion
            qdebug!("HTTP/3 setting SETTINGS_H3_DATAGRAM is enabled but QUIC transport parameter max_datagram_frame_size is 0. Datagrams will be sent via HTTP DATAGRAM Capsules.");
```

With capsule support, I think it is fine to downgrade this to debug level.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -55,32 +55,39 @@ impl Protocol for Session {
...
-                CloseReason::Clean {
-                    error: 0,
-                    message: String::new(),
-                },
-                None,
-            );
-            Ok(Some(State::Done))
-        } else {
-            Ok(None)
+            if let Some(f) = f {
```
mxinden:
```suggestion
            if let Some(Capsule::Datagram { payload }) = f {
```

Would this compile as well? Saves you one `match` statement.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -104,6 +111,23 @@ impl Protocol for Session {
...
     }
+
+    fn write_datagram_capsule(
+        &self,
+        control_stream_send: &mut Box<dyn SendStream>,
+        conn: &mut Connection,
+        buf: &[u8],
+        now: Instant,
+    ) -> Res<bool> {
+        let capsule = Capsule::Datagram {
```
mxinden, line 142:
```suggestion
        if conn.stream_avail_send_space(self.session_id)? < buf.len() {
            qdebug!("Not enough space to send datagram capsule, dropping it.");
            return Ok(false);
        }
        let capsule = Capsule::Datagram {
```

When there is no space in the buffer, we should drop the datagram. Long term, we might want to have a mechanism to tell the producer how many more datagram (capsules) we are willing to accept.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
mxinden: This allocation is unfortunate. Long term I think we should make `Capsule` abstract over either an owned (e.g. `Bytes`) or borrowed (e.g. `&[u8]`) type. That said, I am fine adding a `TODO` here only.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -565,6 +587,16 @@ pub(crate) trait Protocol: Debug + Display {
...
+
+    fn write_datagram_capsule(
+        &self,
+        _control_stream_send: &mut Box<dyn SendStream>,
+        _conn: &mut Connection,
+        _buf: &[u8],
+        _now: Instant,
+    ) -> Res<bool> {
+        Ok(false)
+    }
```
mxinden: I would like to somehow document that WebTransport never sends datagram capsules. How about not providing a default implementation here, and instead adding a `write_datagram_capsule` implementation in `webtransport_session.rs` with the following comment?

```suggestion
    fn write_datagram_capsule(
        &self,
        _control_stream_send: &mut Box<dyn SendStream>,
        _conn: &mut Connection,
        _buf: &[u8],
        _now: Instant,
    ) -> Res<bool>;
```

`webtransport_session.rs`:

``` rust
    fn write_datagram_capsule(
        &self,
        _control_stream_send: &mut Box<dyn SendStream>,
        _conn: &mut Connection,
        _buf: &[u8],
        _now: Instant,
    ) -> Res<bool> {
        // WebTransport never sends datagram capsules.
        Ok(false)
    }
```

neqo-http3/src/recv_message.rs
```diff
@@ -441,6 +441,14 @@ impl RecvStream for RecvMessage {
                     self.set_closed();
                     break Ok((written, true));
                 }
+                RecvMessageState::ExtendedConnect => {
```
mxinden: Is this needed? Removing it still has the tests pass for me locally.

neqo-http3/tests/connect_udp.rs
WaterWhisperer: Understood

neqo-http3/src/features/extended_connect/session.rs
WaterWhisperer: Ohh, that's great!

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -55,32 +55,31 @@ impl Protocol for Session {
...
-                    error: 0,
-                    message: String::new(),
-                },
-                None,
-            );
-            Ok(Some(State::Done))
-        } else {
-            Ok(None)
+            if let Some(Capsule::Datagram { payload }) = f {
+                events.new_datagram(self.session_id, payload, ExtendedConnectType::ConnectUdp);
```
mxinden: I believe there is one missing piece, namely that `payload` here is still prefixed with the [connect-udp context identifier](https://datatracker.ietf.org/doc/html/rfc9298#name-context-identifiers). The same identifier exists on the non-capsule path. The following change would fix the issue, namely to leverage `fn dgram_context_id`:

``` diff
diff --git a/neqo-http3/src/features/extended_connect/connect_udp_session.rs b/neqo-http3/src/features/extended_connect/connect_udp_session.rs
index 4f41b44e..9b6ded93 100644
--- a/neqo-http3/src/features/extended_connect/connect_udp_session.rs
+++ b/neqo-http3/src/features/extended_connect/connect_udp_session.rs
@@ -65,7 +65,15 @@ impl Protocol for Session {
                 .map_err(|_| Error::HttpGeneralProtocolStream)?;
 
             if let Some(Capsule::Datagram { payload }) = f {
-                events.new_datagram(self.session_id, payload, ExtendedConnectType::ConnectUdp);
+                match self.dgram_context_id(payload) {
+                    Ok(slice) => {
+                        events
+                            .new_datagram(self.session_id, slice, self.connect_type());
+                    }
+                    Err(e) => {
+                        qdebug!("[{self}]: received datagram with invalid context identifier: {e}");
+                    }
+                }
```

(Maybe you find a prettier version than the one I suggested above.)

You can find the function here:

https://github.com/mozilla/neqo/blob/adcd4bddddce0fb83bbf3f6c01c3c740f6edc680/neqo-http3/src/features/extended_connect/connect_udp_session.rs#L91-L106

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -377,25 +377,47 @@ impl Session {
...
             debug_assert!(false);
             return Err(Error::Unavailable);
         }
+
+        if conn.remote_datagram_size() == 0 {
+            qtrace!("[{self}] remote_datagram_size is 0, trying HTTP DATAGRAM Capsule");
+            if self.protocol.write_datagram_capsule(
+                &mut self.control_stream_send,
+                conn,
+                buf,
```
mxinden, line 402: Analogues to the receive path (see comment before), we will need to prefix the context id (effectively `0`), before putting `buf` into the capsule. For reference, the non-capsule path below calls `self.protocol.write_datagram_prefix(&mut dgram_data);`.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
WaterWhisperer: Thanks for the review very much! This gave me a deeper understanding

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -55,32 +55,38 @@ impl Protocol for Session {
...
+                    CloseReason::Clean {
+                        error: 0,
+                        message: String::new(),
+                    },
+                    None,
+                );
+                return Ok(Some(State::Done));
+            } else {
+                return Ok(None);
+            }
```
mxinden: If `frame_reader.receive` above returns both a capsule in `f` and `fin == true`, then this code would properly handle the capsule in `f`, but ignore the `fin`. How about the following code that reacts to both.

``` diff
diff --cc neqo-http3/src/features/extended_connect/connect_udp_session.rs
index 4f41b44e,ba9249d6..00000000
--- a/neqo-http3/src/features/extended_connect/connect_udp_session.rs
+++ b/neqo-http3/src/features/extended_connect/connect_udp_session.rs
@@@ -56,7 -56,7 +56,7 @@@ impl Protocol for Session 
          now: Instant,
      ) -> Res<Option<State>> {
          loop {
--            let (f, fin) = self
++            let (capsule, fin) = self
                  .frame_reader
                  .receive::<Capsule>(
                      &mut StreamReaderRecvStreamWrapper::new(conn, control_stream_recv),
@@@ -64,9 -64,16 +64,21 @@@
                  )
                  .map_err(|_| Error::HttpGeneralProtocolStream)?;
  
--            if let Some(Capsule::Datagram { payload }) = f {
-                 events.new_datagram(self.session_id, payload, ExtendedConnectType::ConnectUdp);
-             } else if fin {
 -                match self.dgram_context_id(payload) {
++            let capsule_is_some = capsule.is_some();
++
++            match capsule {
++                Some(Capsule::Datagram { payload }) => match self.dgram_context_id(payload) {
+                     Ok(slice) => {
+                         events.new_datagram(self.session_id, slice, self.connect_type());
+                     }
+                     Err(e) => {
+                         qdebug!("[{self}]: received capsule with invalid context identifier: {e}");
+                     }
 -                }
 -            } else if fin {
++                },
++                None => {}
++            }
++
++            if fin {
                  events.session_end(
                      ExtendedConnectType::ConnectUdp,
                      self.session_id,
@@@ -77,7 -84,7 +89,9 @@@
                      None,
                  );
                  return Ok(Some(State::Done));
--            } else {
++            }
++
++            if !capsule_is_some {
                  return Ok(None);
              }
          }
```

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -104,6 +110,33 @@ impl Protocol for Session {
...
+        buf: &[u8],
+        now: Instant,
+    ) -> Res<bool> {
+        let mut dgram_data = Encoder::default();
+        self.write_datagram_prefix(&mut dgram_data);
+        dgram_data.encode(buf);
+
+        if conn.stream_avail_send_space(self.session_id)? < dgram_data.len() {
+            qdebug!("Not enough space to send datagram capsule, dropping it.");
+            return Ok(false);
```
mxinden: `false` here will have `session.rs` send the connect-udp datagram as a QUIC datagram instead, even though the peer does not support QUIC datagrams.

``` rust
        if conn.remote_datagram_size() == 0 {
            qtrace!("[{self}] remote_datagram_size is 0, trying HTTP DATAGRAM Capsule");
            if self.protocol.write_datagram_capsule(
                &mut self.control_stream_send,
                conn,
                buf,
                now,
            )? {
                qtrace!("[{self}] sent datagram via HTTP DATAGRAM Capsule");
                return Ok(());
            }
            qtrace!("[{self}] Capsule not supported, falling back to QUIC datagram");
        }
```

I don't think Neqo should send a QUIC datagram when the remote does not support QUIC datagrams. Instead, I believe we should really be dropping it here.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -104,6 +110,33 @@ impl Protocol for Session {
...
+        }
+        // TODO: Make Capsule abstract over either an owned (Bytes) or borrowed (&[u8]) type
+        // to avoid this allocation.
+        let capsule = Capsule::Datagram {
+            payload: Bytes::from(Vec::from(dgram_data)),
+        };
+        let mut enc = Encoder::default();
+        capsule.encode(&mut enc);
+        let encoded = enc.as_ref();
+        control_stream_send.send_data_atomic(conn, encoded, now)?;
```
mxinden:
```suggestion
        control_stream_send.send_data_atomic(conn, enc.as_ref(), now)?;
```

Compiles on my machine.

WaterWhisperer: Okay

neqo-http3/src/features/extended_connect/connect_udp_session.rs
WaterWhisperer: Ohh... this was my oversight. Thanks!

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -377,25 +377,47 @@ impl Session {
...
+                now,
+            )? {
+                qtrace!("[{self}] sent datagram via HTTP DATAGRAM Capsule");
+                return Ok(());
+            }
+            qtrace!("[{self}] Capsule not supported, falling back to QUIC datagram");
+        }
```
```suggestion
        if conn.remote_datagram_size() == 0 {
            qtrace!("[{self}] remote_datagram_size is 0, trying HTTP DATAGRAM Capsule");
            self.protocol.write_datagram_capsule(
                &mut self.control_stream_send,
                conn,
                buf,
                now,
            )?;
            return Ok(());
        }
```

I don't think it makes sense to fallback to QUIC datagrams here.

In the case of connect-udp if `self.protocol.write_datagram_capsule` would fail due to no send space in the stream, it would log:

``` rust
            qdebug!("Not enough space to send datagram capsule, dropping it.");
```

directly followed by the log line here:

``` rust
                qtrace!("[{self}] sent datagram via HTTP DATAGRAM Capsule");
```

I think that is confusing.

In the case of webtransport, if `remote_datagram_size() == 0`, there is no value in attempting to send a QUIC datagram thereafter, as it will fail here:

https://github.com/mozilla/neqo/blob/41ca450edbe52280208e42d95ddd95b997c0db29/neqo-transport/src/quic_datagrams.rs#L166-L173

How about `self.protocol.write_datagram_capsule` to return `Res<()>` only and moving the above log lines into the two `write_datagram_capsule` functions? In the case of connect-udp, we would log if the datagram capsule is dropped due to not enough space in the stream (already implemented). In the case of webtransport, we would log that webtransport never supports datagram capsules in the first place, thus dropping it.

WaterWhisperer: Great! Sorry for taking the time to understand, thanks for the review, and, Happy New Year!

larseggert:
@mxinden does this PR need anything other than a rebase? We should really get this merged. (And thanks @WaterWhisperer for your patience here.)

mxinden:
Thanks @WaterWhisperer for your help!

WaterWhisperer:
> Thanks @WaterWhisperer for your help!

@mxinden  It should be me thanking you! I learned a lot of knowledge, and also received a lot of your help in the process. To be honest, this PR is almost entirely your achievement, and perhaps I even held up the progress

PR #3125, opened 2025-11-10
fix(transport/fc): account for WINDOW_UPDATE_FRACTION when auto-tuning

MarcoPolo:
Two commits. 

The first updates the existing test to include RTT delays. The test now fails with the existing code. This is done to highlight the issue with the existing autotune logic.

The second commit fixes the test by sizing up the flow control window by a factor of 4/3. Refer to the commit message for details.

cc @mxinden as we talked about this

### DIFF
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -769,8 +769,7 @@ mod test {
     use crate::{
         connection::params::{MAX_LOCAL_MAX_DATA, MAX_LOCAL_MAX_STREAM_DATA},
         fc::WINDOW_UPDATE_FRACTION,
-        packet::{self, PACKET_LIMIT},
-        recovery,
+        packet, recovery,
         stats::FrameStats,
         stream_id::{StreamId, StreamType},
         ConnectionParameters, Error, Res, INITIAL_LOCAL_MAX_DATA, INITIAL_LOCAL_MAX_STREAM_DATA,
@@ -1013,7 +1012,7 @@ mod test {
         fc[StreamType::BiDi].send_flowc_update();
         // consume the frame
         let mut builder =
-            packet::Builder::short(Encoder::new(), false, None::<&[u8]>, PACKET_LIMIT);
+            packet::Builder::short(Encoder::new(), false, None::<&[u8]>, packet::LIMIT);
         let mut tokens = recovery::Tokens::new();
         fc[StreamType::BiDi].write_frames(&mut builder, &mut tokens, &mut FrameStats::default());
         assert_eq!(tokens.len(), 1);
@@ -1121,7 +1120,7 @@ mod test {
 
     fn write_frames(fc: &mut ReceiverFlowControl<StreamId>, rtt: Duration, now: Instant) -> usize {
         let mut builder =
-            packet::Builder::short(Encoder::new(), false, None::<&[u8]>, PACKET_LIMIT);
+            packet::Builder::short(Encoder::new(), false, None::<&[u8]>, packet::LIMIT);
         let mut tokens = recovery::Tokens::new();
         fc.write_frames(
             &mut builder,
@@ -1382,7 +1381,7 @@ mod test {
         // Helper to write frames
         let write_conn_frames = |fc: &mut ReceiverFlowControl<()>, now: Instant| {
             let mut builder =
-                packet::Builder::short(Encoder::new(), false, None::<&[u8]>, PACKET_LIMIT);
+                packet::Builder::short(Encoder::new(), false, None::<&[u8]>, packet::LIMIT);
             let mut tokens = recovery::Tokens::new();
             fc.write_frames(
                 &mut builder,
@@ -1422,7 +1421,7 @@ mod test {
         // Helper to write frames
         let write_conn_frames = |fc: &mut ReceiverFlowControl<()>| {
             let mut builder =
-                packet::Builder::short(Encoder::new(), false, None::<&[u8]>, PACKET_LIMIT);
+                packet::Builder::short(Encoder::new(), false, None::<&[u8]>, packet::LIMIT);
             let mut tokens = recovery::Tokens::new();
             fc.write_frames(
                 &mut builder,
--- a/.github/actions/check-vm/action.yml
+++ b/.github/actions/check-vm/action.yml
@@ -102,7 +102,7 @@ runs:
         } >> "$GITHUB_OUTPUT"
 
         curl -o "$WD/rustup.sh" --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs
-        echo "envs=CARGO_TERM_COLOR RUST_BACKTRACE RUST_LOG GITHUB_ACTIONS RUST_TEST_TIME_UNIT RUST_TEST_TIME_INTEGRATION RUST_TEST_TIME_DOCTEST WD" >> "$GITHUB_OUTPUT"
+        echo "envs=CARGO_TERM_COLOR RUST_BACKTRACE RUST_LOG RUST_TEST_TIME_UNIT RUST_TEST_TIME_INTEGRATION RUST_TEST_TIME_DOCTEST WD" >> "$GITHUB_OUTPUT"
 
     - if: ${{ inputs.platform == 'freebsd' }}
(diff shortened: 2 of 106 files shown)


martinthomson:
I'm not sure that I had the right intuition for why this would (or wouldn't) help.  But I think that this is right.

Let's assume consistent throughput, where we have perfectly even transmission.  In an ideal case, every byte received is matched with an update that arrives 1RTT later.  If the window is perfectly matched to the BDP and transmission is perfectly even, the sender receives an increase just in time for them to send the next byte.  If we coalesce transmissions, we need to account for bytes that are not yet allocated for credit[^1].  Otherwise, the receiver needs to burst those transmissions when it receives the update.

[^1]: Note that in practice, we also need to account for increases in the congestion window that might occur during the time we don't send updates as well. That is why we multiply the expansion by a factor larger than the slow start multiplier.  If we want to allow for careful resume or things that would exceed the combination of our starting value, plus this rate of increase, we'd need something to detect that mode.  That would motivate signaling of careful resume, maybe, but that's an aside.

If you update just once per RTT for the entire window, then the available window at the receiver jumps up to the BDP if the window has been filled.  But the window won't fill if the sender doesn't have the necessary credit.  In a 1-per-RTT mode, the receiver would have to produce an instantaneous burst of BDP size when it receives the update for that data to arrive at the receiver in time for the next update.  Anything short of that would mean that the increase would not be enough and the throughput would drop.  A 1-per-RTT mode therefore really needs 2*BDP to be happy.

The formula here is right.  We need enough credit available for there to be one window in flight from sender to receiver, plus 1/F of a window in credit in flight from receiver to sender.

Note that I suspect that the reason we're not seeing `*_BLOCKED` frames for our algorithm is that we are sending enough updates to keep most senders happy.  I suspect that a truly high throughput, long BDP link will see blocking with this setup.

martinthomson:
@mxinden can you confirm that @MarcoPolo and I are correct about the logic of this?

MarcoPolo:
Thanks for taking the time to think about this @martinthomson (and answering my questions at IETF!).

> If you update just once per RTT for the entire window, then the available window at the receiver jumps up to the BDP if the window has been filled. But the window won't fill if the sender doesn't have the necessary credit. In a 1-per-RTT mode, the receiver would have to produce an instantaneous burst of BDP size when it receives the update for that data to arrive at the receiver in time for the next update. Anything short of that would mean that the increase would not be enough and the throughput would drop. A 1-per-RTT mode therefore really needs 2*BDP to be happy.

I'm finding this paragraph hard to parse. I think I get the gist that you need a 2*BDP window if you're only sending an update after receiving 1 BDP of data. Is there a missing "sender" noun somewhere?

> Note that I suspect that the reason we're not seeing *_BLOCKED frames for our algorithm is that we are sending enough updates to keep most senders happy. I suspect that a truly high throughput, long BDP link will see blocking with this setup.

If you run your simulator after the changes here: https://github.com/mozilla/neqo/pull/3117, you'll see many blocked frames without the changes from this PR.

neqo-transport/src/fc.rs
```diff
@@ -375,9 +375,19 @@ where
             return;
         };
 
+        // Scale the max_active window down by
+        // [(F-1) / F]; where F=WINDOW_UPDATE_FRACTION.
+        // This is because, when not application limited, in a steady state
+        // transfer it takes 1 RTT after sending 1 / F bytes for the sender to
+        // receive the next update. The sender is effectively limited to
+        // [(F-1) / F] bytes per RTT.
```
mxinden:
```suggestion
        // Scale the max_active window down by
        // [(F-1) / F]; where F=WINDOW_UPDATE_FRACTION.
        //
        // In the ideal case, each byte sent would trigger a flow control
        // update.  However, in practice we only send updates every
        // WINDOW_UPDATE_FRACTION of the window.  Thus, when not application
        // limited, in a steady state transfer it takes 1 RTT after sending 1 /
        // F bytes for the sender to receive the next update. The sender is
        // effectively limited to [(F-1) / F] bytes per RTT.
        //
        // By calculating with this effective window instead of the full
        // max_active, we account for the inherent delay between when the sender
        // would ideally receive flow control updates and when they actually
        // arrive due to our batched update strategy.
        //
        // Example with F=4 without adjustment:
        //
        // t=0         start sending
        // t=RTT/4     sent 1/4 of window total
        // t=RTT       sent 1 window total
        //             sender blocked for RTT/4
        // t=RTT+RTT/4 receive update for 1/4 of window
        //
        // Example with F=4 with adjustment:
        //
        // t=0         start sending
        // t=RTT/4     sent 1/4 of window total
        // t=RTT       sent 1 window total
        // t=RTT+RTT/4 sent 1+1/4 window total; receive update for 1/4 of window (just in time)
        //
```

This would help me come back to this in a couple of months. Thoughts?

neqo-transport/src/fc.rs
```diff
@@ -1248,13 +1271,14 @@ mod test {
...
                     let (_, data) = send_to_recv.pop_front().unwrap();
+                    bytes_received += data;
                     let consumed = fc.set_consumed(fc.retired() + data)?;
                     fc.add_retired(consumed);
 
                     // Receiver sends window updates.
                     let prev_max_allowed = fc.max_allowed;
                     if write_frames(&mut fc, rtt, now) == 1 {
-                        recv_to_send.push_front((now, fc.max_allowed - prev_max_allowed));
+                        recv_to_send.push_back((now + half_rtt, fc.max_allowed - prev_max_allowed));
```
mxinden, line 1302: Thanks for the catch!

neqo-transport/src/fc.rs
```diff
@@ -1226,7 +1238,15 @@ mod test {
             let mut fc =
                 ReceiverFlowControl::new(StreamId::new(0), INITIAL_LOCAL_MAX_STREAM_DATA as u64);
 
+            let mut bytes_received: u64 = 0;
+            let start_time = now;
+
+            let mut accrued_latency = Duration::ZERO;
+            let mut last_accrued = now;
```
mxinden: The logic around these two variables was difficult for me to understand. What do you think of the below instead?

``` diff
diff --git a/neqo-transport/src/fc.rs b/neqo-transport/src/fc.rs
index 48e19ff6..05eef13b 100644
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -1240,12 +1264,10 @@ mod test {
             let mut bytes_received: u64 = 0;
             let start_time = now;
 
-            let mut accrued_latency = Duration::ZERO;
-            let mut last_accrued = now;
-            loop {
-                accrued_latency = accrued_latency.saturating_sub(now.duration_since(last_accrued));
-                last_accrued = now;
+            // Track when sender can next send.
+            let mut next_send_time = now;
 
+            loop {
                 // Sender receives window updates.
                 if recv_to_send.front().is_some_and(|(at, _)| *at <= now) {
                     let (_, update) = recv_to_send.pop_front().unwrap();
@@ -1258,9 +1280,13 @@ mod test {
                     sender_window -= to_send;
                     let time_to_send =
                         Duration::from_secs_f64(to_send as f64 * 8.0 / bandwidth as f64);
-                    // Every time we send data we accrue some latency and push back later sends.
-                    accrued_latency += time_to_send;
-                    send_to_recv.push_back((now + accrued_latency + half_rtt, to_send));
+
+                    let send_start = next_send_time.max(now);
+                    next_send_time = send_start + time_to_send;
+
+                    send_to_recv.push_back((send_start + time_to_send + half_rtt, to_send));
                     true
                 } else {
                     false
```

neqo-transport/src/fc.rs
MarcoPolo: The comment is good, but I'm having a hard time understanding the difference between the examples with and without adjustments. Should you use the "effective window" term somewhere?

neqo-transport/src/fc.rs
MarcoPolo: this is a nice way to think about it. thanks!

neqo-transport/src/fc.rs
MarcoPolo: ah, nevermind. Rereading it again makes sense.

PR #3128, opened 2025-11-12
feat: Introduce a type for `StatelessResetToken`

larseggert:
Rather than using a raw `[u8; 16]` array everywhere.

### DIFF
--- a/neqo-transport/src/cid.rs
+++ b/neqo-transport/src/cid.rs
@@ -598,9 +598,15 @@ impl ConnectionIdManager {
 #[cfg(test)]
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
+    use neqo_common::Encoder;
     use test_fixture::fixture_init;
 
-    use crate::{cid::MAX_CONNECTION_ID_LEN, ConnectionId};
+    use crate::{
+        cid::{ConnectionIdEntry, MAX_CONNECTION_ID_LEN},
+        packet,
+        stats::FrameStats,
+        ConnectionId, Token as Srt,
+    };
 
     #[test]
     fn generate_initial_cid() {
@@ -612,5 +618,28 @@ mod tests {
                 "connection ID length {cid:?}",
             );
         }
+    }
+
+    #[test]
+    fn write_checks_length_correctly() {
+        fixture_init();
+        let entry = ConnectionIdEntry::new(1, ConnectionId::from(&[]), Srt::random());
+        let limit = 1
+            + Encoder::varint_len(entry.sequence_number())
+            + 1
+            + 1
+            + entry.connection_id().len()
+            + Srt::LEN;
+        let enc = Encoder::with_capacity(limit);
+        let mut builder = packet::Builder::short(enc, false, Some(&[]), limit);
+        assert_eq!(
+            builder.remaining(),
+            limit - 1,
+            "Builder::short consumed one byte"
+        );
+        assert!(
+            !entry.write(&mut builder, &mut FrameStats::default()),
+            "couldn't write frame into too-short builder",
+        );
     }
 }


neqo-transport/src/cid.rs
```diff
@@ -19,7 +19,10 @@ use neqo_common::{hex, hex_with_len, qdebug, qinfo, Buffer, Decoder, Encoder};
 use neqo_crypto::{random, randomize};
 use smallvec::{smallvec, SmallVec};
 
-use crate::{frame::FrameType, packet, recovery, stats::FrameStats, Error, Res};
+use crate::{
+    frame::FrameType, packet, recovery, srt::StatelessResetToken as SRT, stats::FrameStats, Error,
```
mxinden:
```suggestion
    frame::FrameType, packet, recovery, srt::StatelessResetToken as Srt, stats::FrameStats, Error,
```

I believe Rust convention would be `Srt`:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

https://rust-lang.github.io/api-guidelines/naming.html

neqo-transport/src/cid.rs
```diff
@@ -270,20 +273,10 @@ impl ConnectionIdEntry<[u8; 16]> {
         )
     }
 
-    fn token_equal(a: &[u8; 16], b: &[u8; 16]) -> bool {
-        // rustc might decide to optimize this and make this non-constant-time
-        // with respect to `t`, but it doesn't appear to currently.
```
mxinden, line 275: Worth copying these lines to the new implementation?

neqo-transport/src/srt.rs
```diff
@@ -0,0 +1,118 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Stateless Reset Token implementation.
+
+use neqo_crypto::random;
+
+use crate::Error;
+
+pub const SRT_LEN: usize = 16;
```
mxinden:
```suggestion
pub const LEN: usize = 16;
```

Already in the `srt` module. No need to prefix with the module name.

neqo-transport/src/srt.rs
mxinden, line 1: I believe most of the time, we use full words, instead of abbreviations, for module names. Could as well call this file `stateless_reset.rs` and the main type just `struct Token`. Not a strong opinion.

neqo-transport/src/connection/tests/close.rs
```diff
@@ -211,7 +211,10 @@ fn stateless_reset_client() {
     let mut client = default_client();
     let mut server = default_server();
     server
-        .set_local_tparam(StatelessResetToken, TransportParameter::Bytes(vec![77; 16]))
+        .set_local_tparam(
+            StatelessResetToken,
+            TransportParameter::Bytes(vec![77; crate::stateless_reset::TOKEN_LEN]),
```
larseggert: Is there any way to avoid the `crate::` bit here?

neqo-transport/src/stateless_reset.rs
```diff
@@ -0,0 +1,120 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Stateless Reset Token implementation.
+
+use neqo_crypto::random;
+
+use crate::Error;
+
+pub const TOKEN_LEN: usize = 16;
```
martinthomson: Should this be `Token::LEN` ?  The compiler seems fine with this sort of loop and it would allow you to replace `crate::stateless_reset::TOKEN_LEN` with `Srt::LEN` in a couple of places.

neqo-transport/src/stateless_reset.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub const fn as_bytes(&self) -> &[u8; TOKEN_LEN] {
+        &self.0
+    }
+}
+
+/// Compare two tokens in constant time to prevent timing attacks.
+impl PartialEq for Token {
+    fn eq(&self, other: &Self) -> bool {
+        // rustc might decide to optimize this and make this non-constant-time
+        // with respect to `t`, but it doesn't appear to currently.
```
martinthomson:
```suggestion
        // It doesn't appear to currently.
```

neqo-transport/src/stateless_reset.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[must_use]
+    pub const fn as_bytes(&self) -> &[u8; TOKEN_LEN] {
+        &self.0
+    }
+}
+
+/// Compare two tokens in constant time to prevent timing attacks.
+impl PartialEq for Token {
+    fn eq(&self, other: &Self) -> bool {
+        // rustc might decide to optimize this and make this non-constant-time
```
martinthomson:
```suggestion
        // rustc might decide to optimize this and make this non-constant-time.
```

neqo-transport/src/tparams.rs
```diff
@@ -238,8 +239,10 @@ impl TransportParameter {
         }
 
         // Stateless reset token
-        let srtbuf = d.decode(16).ok_or(Error::NoMoreData)?;
-        let srt = <[u8; 16]>::try_from(srtbuf)?;
+        let srtbuf = d
+            .decode(crate::stateless_reset::TOKEN_LEN)
+            .ok_or(Error::NoMoreData)?;
+        let srt = Srt::try_from(srtbuf)?;
```
martinthomson: Should the decode be moved to the Srt implementation?

neqo-transport/src/cid.rs
```diff
@@ -312,7 +304,7 @@ impl ConnectionIdEntry<[u8; 16]> {
         builder.encode_varint(self.seqno);
         builder.encode_varint(0u64);
         builder.encode_vec(1, &self.cid);
-        builder.encode(&self.srt);
+        builder.encode(self.srt.as_bytes());
```
martinthomson: If `builder.encode` were to take `impl AsRef<[u8]>`, this wouldn't have needed to change.

martinthomson:
```suggestion
        builder.encode(&self.srt);
```

Assuming that `builder.encode()` takes `impl AsRef<[u8]>`, you can keep this the same.

neqo-transport/src/cid.rs
```diff
@@ -303,7 +290,12 @@ impl ConnectionIdEntry<[u8; 16]> {
         builder: &mut packet::Builder<B>,
         stats: &mut FrameStats,
     ) -> bool {
-        let len = 1 + Encoder::varint_len(self.seqno) + 1 + 1 + self.cid.len() + 16;
+        let len = 1
```
martinthomson: All of these mutants calls are worth paying attention to.  You should be able to write a test that asks this to be written to a packet builder that is full enough that there is no room to add the frame.  You'll have to test across a range of values, but I think that should clear up the whole mess with variants.  I suspect that this will need a `seqno` that is larger than 63 though.

neqo-transport/src/cid.rs
larseggert: I'm opening https://github.com/mozilla/neqo/issues/3135, since that change (which is nice) has some other consequences.

neqo-transport/src/cid.rs
```diff
@@ -312,7 +299,7 @@ impl ConnectionIdEntry<[u8; 16]> {
         builder.encode_varint(self.seqno);
         builder.encode_varint(0u64);
         builder.encode_vec(1, &self.cid);
-        builder.encode(self.srt);
+        builder.encode(&self.srt);
```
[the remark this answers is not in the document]
martinthomson: A foolish consistency in this case.

PR #3129, opened 2025-11-12
perf(qlog): introduce outer `Option`

mxinden:
Currently qlog is defined as:

```rust
pub struct Qlog {
    inner: Rc<RefCell<Option<SharedStreamer>>>,
}
```

https://github.com/mozilla/neqo/issues/1894 documents our finding, that dereferencing the `Rc<RefCell<_>>` is costly, i.e. showing up in our CPU profiles.

When writing to the qlog fails, we set the (inner) `Option` to `None`. Thus on each consecutive logging atttempt we dereference the `Rc<RefCell<_>>` only to find out that logging is disabled.

This patch still sets the inner `Option` to `None` to inform all other clones of the `Rc` that logging is disabled. In addition, it introduces an outer `Option`, which as well is set to `None`, thus avoiding the dereferencing of the `Rc<RefCell<_>>` on each logging attempt once logging has failed.

``` rust
pub struct Qlog {
    inner: Option<Rc<RefCell<Option<SharedStreamer>>>>,
}
```

---

Fixes https://github.com/mozilla/neqo/issues/1894.
Alternative for https://github.com/mozilla/neqo/pull/3005.

@martinthomson's idea, shared at IETF 124. Let's see what our benchmarks say. Not ready for actual review.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -135,13 +135,13 @@ impl Qlog {
 
         let mut borrow = inner.borrow_mut();
 
-        let Some(inner_inner) = borrow.as_mut() else {
+        let Some(shared_streamer) = borrow.as_mut() else {
             drop(borrow);
             self.inner = None;
             return;
         };
 
-        if let Err(e) = f(&mut inner_inner.streamer) {
+        if let Err(e) = f(&mut shared_streamer.streamer) {
             log::error!("Qlog event generation failed with error {e}; closing qlog.");
 
             // Explicitly drop the RefCell borrow to release the mutable borrow.


larseggert:
Do we need both `Option`s? Otherwise, I prefer this to #3005!

mxinden:
I don't see a way around the two `Option`s. The inner `Option` will notify all other owners of the qlog, that the log has errored, the outer `Option` prevents the current owner from checking the shared reference again.

neqo-common/src/qlog.rs
```diff
@@ -117,15 +117,27 @@ impl Qlog {
...
-                *self.inner.borrow_mut() = None;
-            }
+        let Some(inner) = self.inner.as_mut() else {
+            return;
+        };
+
+        let mut borrow = inner.borrow_mut();
+
+        let Some(inner_inner) = borrow.as_mut() else {
+            return;
```
martinthomson, line 141: You *could* do this here, which would mean that subsequent calls would hit the first check:

```suggestion
            self.inner = None;
            return;
```

... At least as far as the clone that this is operating on.  Other clones will do the same thing when they first touch a disabled log.

After all, the point of using `RefCell` is to avoid needing a mutable reference.  But since you made the reference mutable, we have the luxury of updating this.

The drawback is that this is a ratchet.  If the inner is ever set to `None`, it can never be set back to `Some`, because any clone that calls this function will disconnect from the `Rc`.  That means that you need to ensure that you initialize logging before using the connection.  If you don't, you can't enable.

larseggert: IMO that's fine _if_ we document it properly. It's not like there are a gazillion consumers of this API.

larseggert:
@mxinden given that you are busy, would you like me to apply @martinthomson's suggestion and merge this? It'd unblock #2989.

mxinden:
Yes please @larseggert. Thanks.

neqo-common/src/qlog.rs
```diff
@@ -69,17 +69,20 @@ impl Qlog {
 
     /// Create an enabled `Qlog` configuration.
     ///
+    /// This needs to be called before the connection is used, because otherwise `Qlog`-logging will
+    /// remain disabled (for performance reasons).
```
mxinden: Note that `enabled` is a constructor. With that in mind, I don't understand this comment. How would one call this method after creating the connection?

neqo-common/src/qlog.rs
```diff
@@ -22,7 +22,7 @@ use crate::Role;
 
 #[derive(Debug, Clone, Default)]
 pub struct Qlog {
-    inner: Rc<RefCell<Option<SharedStreamer>>>,
+    inner: Option<Rc<RefCell<Option<SharedStreamer>>>>,
```
mxinden, line 30:
```suggestion
    /// Both the inner and the outer `Option` are set to `None`
    /// on failure. The inner `None` will disable qlog for all other
    /// references (correctness). The outer `None` will prevent
    /// the local instance from de-referencing the `Rc` again
    /// (performance). 
    inner: Option<Rc<RefCell<Option<SharedStreamer>>>>,
```

neqo-common/src/qlog.rs
```diff
@@ -117,15 +120,29 @@ impl Qlog {
...
-            }
+        let Some(inner) = self.inner.as_mut() else {
+            return;
+        };
+
+        let mut borrow = inner.borrow_mut();
+
+        let Some(inner_inner) = borrow.as_mut() else {
+            drop(borrow);
+            self.inner = None;
```
mxinden, line 140: This is crucial. Thanks for adding it.

neqo-common/src/qlog.rs
```diff
@@ -117,15 +120,29 @@ impl Qlog {
...
+            return;
+        };
+
+        if let Err(e) = f(&mut inner_inner.streamer) {
+            log::error!("Qlog event generation failed with error {e}; closing qlog.");
+
+            // Remove the shared QLOG.
+            borrow.take();
+            // Remove the reference to the shared QLOG.
+            inner.take();
```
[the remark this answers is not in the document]
larseggert: @mxinden this is BS, right?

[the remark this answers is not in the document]
mxinden: > Line 145 calls inner.take() on an Rc

There is no `Rc::take`. In other words this is wrong. That said, I guess the suggestion `self.inner = None` is easier to read. @larseggert your call.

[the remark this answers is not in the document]
larseggert: Did that. Also needed to change `borrow.take()` to `drop(borrow)` to make the compiler happy.

neqo-common/src/qlog.rs
```diff
@@ -117,15 +125,29 @@ impl Qlog {
...
+        };
+
+        if let Err(e) = f(&mut shared_streamer.streamer) {
+            log::error!("Qlog event generation failed with error {e}; closing qlog.");
+
+            // Explicitly drop the RefCell borrow to release the mutable borrow.
+            drop(borrow);
+            // Set the outer Option to None to prevent future dereferences.
+            self.inner = None;
         }
```
mxinden, line 151: I am confused. Shouldn't this set `borrow = None;`? How else would other reference holders be notified that an error occurred @larseggert? This was removed in https://github.com/mozilla/neqo/pull/3129/commits/27624c504a1ea638b6c0091aaab627ff3191f1c6.

larseggert: Doing `take` caused a compiler error of some sort. I thought this was correct but probably messed up. Do a follow-on?

martinthomson: `borrow` isn't an `Option`, so setting `borrow = None` wouldn't have worked.  `*borrow = None` would be a noop, but wouldn't have ended the lifetime of the borrow.  You have to use `drop()`.

mxinden: I believe there is a confusion, I am not suggesting to remote the `drop(borrow)`, but instead add `*borrow = None;` above it. See suggestion in https://github.com/mozilla/neqo/pull/3267.

PR #3131, opened 2025-11-13
ci: Expose workflows to external repositories

Not-Nik:
This PR is the first step in replacing neqo-crypto with [nss-rs](https://github.com/mozilla/nss-rs). It allows using applicable workflows in that project without having to copy the workflow and action files.

The `uses` directives no longer use paths local to the repository the workflow is in, instead always directing to this repository. This requires silencing a security check in zizmor which disallows branch or tag pins on actions outside of official GitHub namespaces. Since we control this repository, we can be sure that there are no security implications. As the [zizmor docs on this issue](https://docs.zizmor.sh/audits/#unpinned-uses) say:

> If the upstream repository is trusted, then symbolic references are often suitable. However, if the upstream repository is not trusted, then actions should be pinned by SHA reference.

Additionally, this removes the "Run client/server transfer" job, so that all of `check.yml` can be used without dependencies on this repository. The test in that job is also checked in other connectivity tests.

### DIFF
--- a/.github/actions/check-android/action.yml
+++ b/.github/actions/check-android/action.yml
@@ -43,14 +43,14 @@ runs:
         WD: ${{ inputs.working-directory }}
       run: cd "$WD" && sdkmanager --install "ndk;$NDK_VERSION"
 
-    - uses: ./.github/actions/rust
+    - uses: mozilla/neqo/.github/actions/rust@main
       with:
         version: stable
         targets: ${{ inputs.target }}
         tools: cargo-ndk@^4
         token: ${{ inputs.github-token }}
 
-    - uses: ./.github/actions/nss
+    - uses: mozilla/neqo/.github/actions/nss@main
       if: ${{ inputs.minimum-nss-version != '' }}
       with:
         minimum-version: ${{ inputs.minimum-nss-version }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -102,7 +102,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: ./.github/actions/rust
+      - uses: mozilla/neqo/.github/actions/rust@main
         with:
           version: ${{ matrix.rust-toolchain }}
           components: ${{ matrix.rust-toolchain == 'stable' && 'llvm-tools' || '' }}
@@ -112,7 +112,7 @@ jobs:
       - id: nss-version
         run: echo "minimum=$(cat neqo-crypto/min_version.txt)" >> "$GITHUB_OUTPUT"
 
-      - uses: ./.github/actions/nss
+      - uses: mozilla/neqo/.github/actions/nss@main
         with:
           minimum-version: ${{ steps.nss-version.outputs.minimum }}
 
@@ -171,7 +171,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
-      - uses: ./.github/actions/rust
+      - uses: mozilla/neqo/.github/actions/rust@main
         with:
           version: stable
           tools: cargo-hack
@@ -193,7 +193,7 @@ jobs:
           persist-credentials: false
       - id: nss-version
         run: echo "minimum=$(cat neqo-crypto/min_version.txt)" >> "$GITHUB_OUTPUT"
-      - uses: ./.github/actions/check-android
+      - uses: mozilla/neqo/.github/actions/check-android@main
         with:
           target: ${{ matrix.target }}
           minimum-nss-version: ${{ steps.nss-version.outputs.minimum }}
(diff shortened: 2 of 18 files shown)


.github/workflows/check.yml
```diff
@@ -138,23 +139,6 @@ jobs:
...
-          # Give the server time to start.
-          sleep 1
-          "target/$BUILD_DIR/neqo-client" --output-dir . "https://$HOST:4433/$SIZE"
-          kill $PID
-          [ "$(wc -c <"$SIZE")" -eq "$SIZE" ] || exit 1
-        env:
-          HOST: localhost
-          SIZE: 54321
-          RUST_LOG: warn
-          BUILD_DIR: ${{ matrix.type == 'release' && 'release' || 'debug' }}
```
larseggert, line 156: @mxinden is it worth sticking this into a different workflow?

.github/actions/check-android/action.yml
```diff
@@ -43,14 +43,14 @@ runs:
         WD: ${{ inputs.working-directory }}
       run: cd "$WD" && sdkmanager --install "ndk;$NDK_VERSION"
 
-    - uses: ./.github/actions/rust
+    - uses: mozilla/neqo/.github/actions/rust@main
```
mxinden, line 46: Why check against a specific branch? How would I make a fix to this action and validating the fix, before merging into `main`?

.github/workflows/check.yml
mxinden: No strong opinion. Separate workflow sounds good to me.

.github/actions/check-android/action.yml
Not-Nik: Action references are either to a specific repository and branch, or to a file in the same repository as the called ([ref](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-a-public-action-in-a-subdirectory)). Without this trying to call neqo workflows from another repository would fail, because that repository doesn't have the required action files.

> How would I make a fix to this action and validating the fix, before merging into `main`?

Lars mentioned that this was already an issue with the current approach, but if it does make development harder, I'll need to find another fix.

mxinden: I am confused, how are you fetching this file? Wouldn't you thus fetch all files of the `.github/` folder?

Can we do it as we did for `mtu` in the past?

https://github.com/mozilla/mtu/blob/cb57a96013ea3b7c9e9d4fc0bdc089f5bcc3362b/.github/workflows/check.yml#L83-L95

Say that I cut a release off of a past branch, e.g. `v0.17`, that pull request would use workflows from `main` which might no longer be compatible with the `v0.17` code.

Not-Nik: I don't fetch the `.github` folder at all. Instead I use [workflow calling](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows), which avoids having to manually checkout this repository.

mxinden: Are there any alternatives @Not-Nik? How do other projects solve this? I don't think the current approach, i.e. pinning to `@main`, is worth the downsides (e.g. harder to iterate, no stable release processs off of non-`main`).

Not-Nik: I did a bit of testing and I don't think there are any viable alternatives that don't require keeping copies of workflow or action files around, but we can fix the development issues.

The obvious solutions of trying to checkout neqo's action files won't work. Workflow calling doesn't happen in a job's steps, so we can't give it anything that isn't already in the repository. Similarly, we can't checkout all the workflow and action files and then call the local workflow file.

A relatively easy fix is to keep both the current local references, and the "new" repository reference, and then set a flag that switches between the two, depending on the caller. Here is what that would roughly look like:

```
on:
  # ...
  workflow_call:
    inputs:
      is-workflow-call:
        type: boolean
        required: false
        default: true

jobs:
  check:
    steps:
      - if: ${{ !inputs.is-workflow-call }}
        uses: ./.github/actions/rust
        with:
          # ...

      - if: ${{ inputs.is-workflow-call }}
        uses: mozilla/neqo/.github/actions/rust@main
        with:
          # ...
```

It does require a few more lines, but it pretty much solves all our problems

larseggert: Can't you just use `github.event_name == 'workflow_call'` in the `if:` condition, rather than introducing a new parameter?

But duplicating the steps is still ugly.

larseggert: Would this issue go away if we created, say, `mozilla/actions` and placed the actions there?

mxinden: I believe the take-away from an in-person conversation with @Not-Nik was, that whatever we share, we should move into actions first. I don't think it is idiomatic to share workflows across repositories.

Not-Nik: > Can't you just use github.event_name == 'workflow_call'

Unfortunately no. The event name [isn't updated](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call) for workflow calls. As Max mentioned, moving any shared CI into actions is the best way to move forward, even though it takes some more effort.

Not-Nik:
Closing for now, since we're going a different way for the workflows, as discussed

PR #3133, opened 2025-11-14
fix(http3): allow non-utf8 HTTP header values

valenting:
See [1999659 - HTTP3 doesn't allow non-ascii header values](https://bugzilla.mozilla.org/show_bug.cgi?id=1999659)

It sees H3 request with a non-utf8 header would fallback and succeed with H2
```js
headers = {}
headers['x-panel-title'] = "ä";
res = await fetch(location.href, {
	method: 'GET',
    headers,
})
```

According to [RFC 9110 HTTP Semantics > 5.5 Field values](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5) 

```
  field-value    = *field-content
  field-content  = field-vchar
                   [ 1*( SP / HTAB / field-vchar ) field-vchar ]
  field-vchar    = VCHAR / obs-text
  obs-text       = %x80-FF
```
> A recipient SHOULD treat other allowed octets in field content (i.e., [obs-text](https://www.rfc-editor.org/rfc/rfc9110.html#fields.values)) as opaque data.

However, neqo's Header value is a rust String which doesn't allow non-utf8 characters.
To support this we need to turn value into a Vec<u8>

### DIFF
--- a/neqo-http3/src/headers_checks.rs
+++ b/neqo-http3/src/headers_checks.rs
@@ -176,7 +176,7 @@ pub fn trailers_valid(headers: &[Header]) -> Res<()> {
 mod tests {
     use neqo_common::Header;
 
-    use super::headers_valid;
+    use super::{headers_valid, is_interim};
     use crate::MessageType;
 
     fn create_connect_headers() -> Vec<Header> {
@@ -239,5 +239,20 @@ mod tests {
             MessageType::Request
         )
         .is_err());
+    }
+
+    #[test]
+    fn is_interim_invalid_utf8() {
+        // Create a header with invalid UTF-8 bytes in the status value
+        let invalid_utf8_bytes = vec![0xFF, 0xFE, 0xFD];
+        let header = Header::new(":status", invalid_utf8_bytes.as_slice());
+        let headers = vec![header];
+        assert!(is_interim(&headers).is_err());
+    }
+
+    #[test]
+    fn is_interim_not_a_number() {
+        let headers = vec![Header::new(":status", "not-a-number")];
+        assert!(is_interim(&headers).is_err());
     }
 }


neqo-common/src/header.rs
```diff
@@ -11,18 +11,18 @@ use thiserror::Error;
...
-    value: String,
+    value: Vec<u8>,
 }
 
 impl Header {
     pub fn new<N, V>(name: N, value: V) -> Self
     where
         N: Into<String>,
-        V: Into<String>,
+        V: AsRef<[u8]>,
```
mxinden: Preference for `Into<Vec<u8>>`. That way, if someone already provides a `Vec`, it is not re-allocated through the `to_vec` call below.

neqo-common/src/header.rs
```diff
@@ -11,18 +11,18 @@ use thiserror::Error;
 #[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone)]
 pub struct Header {
     name: String,
-    value: String,
+    value: Vec<u8>,
```
mxinden, line 21:
```suggestion
    /// The raw header field value as bytes.
    ///
    /// HTTP allows field values to contain any visible ASCII characters and
    /// arbitrary 0x80–0xFF bytes (`obs-text`).  Unlike field *names*, field
    /// values are not guaranteed to be valid UTF-8.
    ///
    /// See also <https://www.rfc-editor.org/rfc/rfc9110#section-5.5>.
    value: Vec<u8>,
```

neqo-common/src/header.rs
```diff
@@ -110,7 +110,7 @@ impl FromStr for Header {
             return Err(FromStrError::MissingName);
         }
 
-        let value = s[seperator + 1..].trim();
+        let value = s[seperator + 1..].trim().as_bytes().to_vec();
```
mxinden: Why is the `to_vec` needed here? Why not have `Self::new` below decide how to allocate?

neqo-common/src/header.rs
```diff
@@ -154,4 +154,42 @@ mod tests {
...
+    fn non_utf8_header_value() {
+        // Create a header with non-UTF-8 bytes in the value
+        let non_utf8_bytes: Vec<u8> = vec![0xFF, 0xFE, 0xFD, 0x80, 0x81];
+        let header = Header::new("custom-header", &non_utf8_bytes);
+
+        assert_eq!(header.name(), "custom-header");
+        assert_eq!(header.value(), non_utf8_bytes.as_slice());
+
+        // Verify that the value is indeed not valid UTF-8
+        assert!(std::str::from_utf8(header.value()).is_err());
```
mxinden: Add a `use std::str` or `use std::str::from_utf8` in the `mod tests` above?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -261,7 +261,7 @@ impl Session {
                     .iter()
                     .find_map(|h| {
                         if h.name() == ":status" {
-                            h.value().parse::<u16>().ok()
+                            std::str::from_utf8(h.value()).ok()?.parse::<u16>().ok()
```
mxinden: Same nit here, preference for a `use` above.

neqo-http3/tests/non_ascii_headers.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    let headers = response_headers.expect("No response headers received");
+
+    // Verify the echoed header contains the original non-UTF-8 byte
+    let echoed_header = headers
+        .find_header("x-echoed-data")
+        .expect("Echoed header not found");
+    assert_eq!(echoed_header.value(), test_bytes.as_slice());
+    assert_eq!(echoed_header.value(), &[0xE4]);
```
mxinden: Nit: preference for defining `0xE4` once as a constant above.

neqo-http3/tests/non_ascii_headers.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+#[test]
+fn non_ascii_emoji_header_echo() {
```
mxinden: Can this be de-duplicated with the function above? How about:

``` rust
fn echo_header(header: Header) {
  // most of the test logic
}

#[test]
fn non_ascii_non_utf8_header() {
  echo_header(todo!());
}

#[test]
fn non_ascii_emoji_header() {
  echo_header(todo!());
}

File: neqo-qpack/src/reader.rs [hunk 8]
```diff
@@ -130,9 +130,9 @@ impl<'a> ReceiverBufferWrapper<'a> {
             .try_into()
             .or(Err(Error::Decompression))?;
         if use_huffman {
-            Ok(parse_utf8(&huffman::decode(self.slice(length)?)?)?.to_string())
+            huffman::decode(self.slice(length)?)
         } else {
-            Ok(parse_utf8(self.slice(length)?)?.to_string())
+            Ok(self.slice(length)?.to_vec())
```
mxinden, line 135: Consumers of `read_literal_from_buffer` might not actually need an owned `Vec`, but would be fine with a `&[u8]`, e.g. when parsing into a `String` right away. Would returning a `Cow<'a, [u8]>` here make sense? For the `if`, we would return an owned `Cow::Owned` with a `Vec<u8>`. For the `else` we would return a `Cow::Borowed` `[u8]`.

(might be overly optimized)

mxinden (commented):
Thank you for tackling this!

Got a couple of comments, mostly around memory management.

Unfortunate that we can't use `String`. That said, rational makes sense.

neqo-qpack/src/reader.rs
valenting: I think that's a nice optimization, but all of the callers right now seem to be putting the return value into a new Header struct. The only exception is encode_header_block where it previously allocated a Vec by doing `iter.value().to_vec();`. So this wouldn't be a perf regression at least. I'd leave it for a follow-up if that's OK.

neqo-http3/tests/non_ascii_headers.rs
```diff
@@ -84,104 +81,23 @@ fn non_ascii_utf8_header_echo() {
...
-    // Connect a client and a server.
-    let mut client = default_http3_client();
-    let mut server = default_http3_server();
-    let out = test_fixture::connect_peers(&mut client, &mut server);
-    assert_eq!(server.process(out, now()).dgram(), None);
-
-    // Ignore all events so far.
-    drop(server.events());
-    drop(client.events());
+fn non_ascii_utf8_header_echo() {
```
mxinden:
```suggestion
fn extended_ascii_non_utf8_header_echo() {
```

Wouldn't this be correct?

neqo-qpack/src/reader.rs
mxinden: :+1: works for me.

neqo-bin/src/server/http3.rs
```diff
@@ -104,10 +104,14 @@ impl super::HttpServer for HttpServer {
                 } => {
                     qdebug!("Headers (request={stream} fin={fin}): {headers:?}");
 
-                    if headers.contains_header(":method", "POST") {
-                        let response_size = headers
-                            .find_header(":path")
-                            .and_then(|path| path.value().trim_matches('/').parse::<usize>().ok());
+                    if headers.contains_header(":method", b"POST") {
+                        let response_size = headers.find_header(":path").and_then(|path| {
+                            std::str::from_utf8(path.value())
```
larseggert: You import this in other files, might as well do so here.
```suggestion
                            from_utf8(path.value())
```

mxinden: Suggestion: This pull request introduces 3x `from_utf8(path.value())`. How about introducing a method on `path`, e.g. `path.value_utf8` de-duplicating the calls to `from_utf8`.

larseggert:
Looking at https://app.codecov.io/gh/mozilla/neqo/pull/3133?dropdown=coverage&src=pr&el=h1, this might use some tests where the headers aren't UTF8.

PR #3138, opened 2025-11-17
chore: `MAX_CONNECTION_ID_LEN` -> `ConnectionId::MAX_LEN`

larseggert:
Similar to what we did for `Token::LEN` in #3128.


mxinden (approved):
Neat!

PR #3140, opened 2025-11-18
chore: Make CID-related associated `const`s

larseggert:
Where that makes sense.

### DIFF
--- a/neqo-transport/src/cid.rs
+++ b/neqo-transport/src/cid.rs
@@ -247,7 +247,7 @@ pub struct ConnectionIdEntry<SRT: Clone + PartialEq> {
 impl ConnectionIdEntry<Srt> {
     /// Create the first entry, which won't have a stateless reset token.
     pub fn initial_remote(cid: ConnectionId) -> Self {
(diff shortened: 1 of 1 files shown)


neqo-transport/src/cid.rs
```diff
@@ -306,11 +298,13 @@ impl ConnectionIdEntry<Srt> {
     }
 
     pub fn is_empty(&self) -> bool {
-        self.seqno == CONNECTION_ID_SEQNO_EMPTY || self.cid.is_empty()
+        self.seqno == ConnectionIdManager::SEQNO_EMPTY || self.cid.is_empty()
     }
 }
 
 impl ConnectionIdEntry<()> {
+    const SEQNO_INITIAL: u64 = 0;
```
martinthomson, line 306: I think that you need to move this to an `impl<T> ConnectionIdEntry<T> {` with no constraints on `T`.  Otherwise, it won't be visible to other implementations.

larseggert: Oh, neat!

PR #3141, opened 2025-11-18
chore: Make `HFrame`-related associated consts

larseggert:
Where that makes sense.


martinthomson (approved):
Copilot's documentation is nice.

PR #3148, opened 2025-11-21
feat: Add initial fuzzing corpora

larseggert:
And a script to generate and minimize them. Next step would be to use the corpora as part of `oss-fuzz`.

(I was considering to use a separate repo for the corpora, but after minimization there are not so many files, so it should be fine to keep them here.)

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -220,7 +220,7 @@ jobs:
           key: qns-${{ github.sha }}
           restore-keys: qns-
 
-      - run: sudo apt-get install -y --no-install-recommends wdiff
+      - run: sudo apt-get install -y --no-install-recommends wdiff cmark-gfm
       - env:
           PAIRS: ${{ needs.implementations.outputs.pairs }}
         run: |
@@ -293,10 +293,16 @@ jobs:
                 echo ", differences relative to $SHA."
               fi
               echo
-              echo "#### $LATEST as client"
-              cat client.failed.md || echo
-              echo "#### $LATEST as server"
-              cat server.failed.md || echo
+              echo "<table><tr><th align=\"left\">$LATEST as client</th><th align=\"left\">$LATEST as server</th></tr>"
+              echo -n "<tr><td valign=\"top\">"
+              if [ -e client.failed.md ]; then
+                cmark-gfm -e strikethrough -e autolink client.failed.md | tr -d '\n' | sed 's/<\/*ul>//g; s/<li>//g; s/<\/li>/<br>/g'
+              fi
+              echo -n "</td><td valign=\"top\">"
+              if [ -e server.failed.md ]; then
+                cmark-gfm -e strikethrough -e autolink server.failed.md | tr -d '\n' | sed 's/<\/*ul>//g; s/<li>//g; s/<\/li>/<br>/g'
+              fi
+              echo "</td></tr></table>"
             else
               echo -n "None "
               if [ -e "client.succeeded.md" ] || [ -e "server.succeeded.md" ] || [ -e "client.unsupported.md" ] || [ -e "server.unsupported.md" ]; then
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -62,3 +62,10 @@ path = "fuzz_targets/find_sni.rs"
 test = false
 doc = false
 bench = false
+
+[[bin]]
+name = "tparams"
+path = "fuzz_targets/tparams.rs"
+test = false
+doc = false
+bench = false
--- a/fuzz/fuzz_targets/tparams.rs
+++ b/fuzz/fuzz_targets/tparams.rs
@@ -1,0 +1,13 @@
+#![cfg_attr(all(fuzzing, not(windows)), no_main)]
+
+#[cfg(all(fuzzing, not(windows)))]
+use libfuzzer_sys::fuzz_target;
+
+#[cfg(all(fuzzing, not(windows)))]
+fuzz_target!(|data: &[u8]| {
+    // Run the fuzzer
+    _ = neqo_transport::tparams::TransportParameters::decode(&mut neqo_common::Decoder::new(data));
+});
+
+#[cfg(any(not(fuzzing), windows))]
+fn main() {}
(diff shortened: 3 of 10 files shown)


mxinden (commented):
I don't have experience with how to handle these.

Note that some of them are > 1KB. Would we update them regularly? How well does git handle this?

mxinden:
> For better OSS-Fuzz integration, the seed corpus should be available in revision control (it can be the same as or different from the source code). It should be regularly extended with the inputs that (used to) trigger bugs and/or touch new parts of the code.

https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#seed-corpus

This does suggest using the same repository.

martinthomson:
Should we add a CI run to regenerate these?

larseggert:
It's not clear whether the hashes are stable, could lead to quite a bit of churn. I will check this next week.

martinthomson:
I doubt that anything will be stable here.  Too much randomness in our implementation, if nothing else.  However, we should have something on a schedule that ensures that any change in our stack is caught in new corpora.  An action can use `gh pr create` to create a pull request that we can review, rather than churn the repo directly.

larseggert:
> However, we should have something on a schedule that ensures that any change in our stack is caught in new corpora. An action can use `gh pr create` to create a pull request that we can review, rather than churn the repo directly.

I added this. However, we may want to exclude the `client_initial` and `server_initial` fuzzers from this, since there is still inherent randomization in how they are generated (see #2855), which generates new samples every time.

.github/workflows/fuzz-corpus.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: echo "minimum=$(cat neqo-crypto/min_version.txt)" >> "$GITHUB_OUTPUT"
+
+      - uses: ./.github/actions/nss
+        with:
+          minimum-version: ${{ steps.nss-version.outputs.minimum }}
+
+      - env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          HEAD_REF: ${{ github.ref }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
martinthomson: Isn't this `${{ github.token }}` instead?

.github/workflows/fuzz-corpus.yml
```diff
@@ -0,0 +1,66 @@
+name: Check for new fuzzing samples
+on:
+  pull_request:
```
martinthomson: Don't you want this on a schedule instead?  Then you only need to worry about updating main.

I would also add manual triggering.  That means workflow_dispatch, so that you can trigger a build if you know that something has significantly changed in the corpus.

Then, the branch that you make could be named for the revision that you are targeting.  So you could use `${{ github.sha }}` (maybe truncated `$(echo ${{ github.sha }} | cut -c 1-10 -)`) in the branch name.

.github/workflows/fuzz-corpus.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - id: nss-version
+        run: echo "minimum=$(cat neqo-crypto/min_version.txt)" >> "$GITHUB_OUTPUT"
+
+      - uses: ./.github/actions/nss
+        with:
+          minimum-version: ${{ steps.nss-version.outputs.minimum }}
+
+      - env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
```
martinthomson: See above.  I don't think that you want this to trigger on PRs.

larseggert: Ah, I had a different mental model: I was trying to get it to propose PRs for PRs that changed things so the corpora changed. In other words, the idea was to have each PR include any corpora samples it caused to be generated.

larseggert: I'm switching to your suggestion, since it avoids the "push from foreign clone" permissions issue.

.github/workflows/fuzz-corpus.yml
```diff
@@ -0,0 +1,67 @@
+name: Check for new fuzzing samples
+on:
+  workflow_dispatch:
+  schedule:
+    # Run at 1 AM each day, so there is a `main`-branch baseline in the cache.
+    - cron: '0 1 * * *'
```
martinthomson: I'd dial this back further.  Running weekly would be fine.

Keep in mind that you'll need to review and approve these PRs.

test/make-fuzz-corpus.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# First, let's run the test suite to generate initial fuzzing corpora.
+# This will have failing tests, ignore them.
+echo "Generating new fuzzing corpora under $TMP..."
+NEQO_CORPUS=$TMP cargo test --quiet --locked --features build-fuzzing-corpus --no-fail-fast > /dev/null 2>&1 || true
+
+# Do dev builds, since we just merge corpora and don't actually run fuzzers here.
+# LTO-linking otherwise takes longer than the merging.
+echo "Building all fuzzers..."
+cargo fuzz build --dev > /dev/null 2>&1
```
martinthomson: Do we really need to send all the output to `/dev/null`?

test/make-fuzz-corpus.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Now, only merge in those newly-generated samples that increase coverage.
+# "cargo fuzz" cannot do this, so use the underlying LLVM fuzzer binary directly.
+TRIPLE="$(basename "$(dirname "$(rustc --print target-libdir)")")"
```
martinthomson:
```suggestion
TRIPLE="$(rustc --print host-tuple)"
```

test/make-fuzz-corpus.sh
```diff
@@ -0,0 +1,34 @@
...
+for fuzzer in $(cargo fuzz list); do
+    generated="$TMP/$fuzzer"
+    if [ ! -d "$generated" ]; then
+        echo "$fuzzer fuzzer: WARNING, test suite generated no corpus"
+        continue
+    fi
+    corpus="fuzz/corpus/$fuzzer"
+    before=$(find "$corpus" -type f | wc -l | tr -d ' ')
+    echo "$fuzzer fuzzer: Merging new unique samples into corpus ($before samples currently)..."
+    "target/$TRIPLE/debug/$fuzzer" -merge=1 "$corpus" "$generated" 2> /dev/null
```
martinthomson: I'm not sure that you want to hide errors, just like above.  I know that those errors will be distracting, but it's worth having the record, on balance.

PR #3160, opened 2025-11-24
feat: Fuzz QPACK decoding

larseggert:
~~And also fix an off-by-one error that was quickly found while fuzzing.~~ Now broken out into #3174.

### DIFF
--- a/fuzz/fuzz_targets/qpack.rs
+++ b/fuzz/fuzz_targets/qpack.rs
@@ -27,11 +27,11 @@ fuzz_target!(|data: &[u8]| {
     let (encoder_stream, header_block) = if encoder_stream_len <= data.len() {
         data.split_at(encoder_stream_len)
     } else {
-        (&[], data)
+        (&[][..], data)
     };
 
-    let settings = neqo_http3::Http3Parameters::default().get_qpack_settings();
-    let mut decoder = neqo_qpack::Decoder::new(settings);
+    let mut decoder =
+        neqo_qpack::Decoder::new(neqo_http3::Http3Parameters::default().get_qpack_settings());
 
     // Process encoder stream data to populate the dynamic table.
     _ = decoder.receive_encoder_stream(encoder_stream);
--- a/neqo-qpack/src/fuzz.rs
+++ b/neqo-qpack/src/fuzz.rs
@@ -6,8 +6,6 @@
 
 //! Fuzzing support for QPACK.
 
-#[cfg(fuzzing)]
-pub use fuzzing::*;
 #[cfg(feature = "build-fuzzing-corpus")]
 pub use write_corpus::write_item_to_fuzzing_corpus;
 
@@ -69,7 +67,7 @@ mod fuzzing {
     }
 
     fn map_error(err: Error) -> Error {
-        if *err == Error::ClosedCriticalStream {
+        if err == Error::ClosedCriticalStream {
             Error::ClosedCriticalStream
         } else {
             Error::EncoderStream


neqo-qpack/src/decoder.rs
```diff
@@ -206,6 +206,15 @@ impl Decoder {
         buf: &[u8],
         stream_id: StreamId,
     ) -> Res<Option<Vec<Header>>> {
+        #[cfg(all(feature = "build-fuzzing-corpus", test))]
+        {
+            // Copy `stream_id` into the first 64 bits of corpus data.
+            let mut data = Vec::new();
```
martinthomson:
```suggestion
            let mut data = Vec::with_capacity(8 + buf.len());
```

fuzz/fuzz_targets/qpack.rs
```diff
@@ -0,0 +1,28 @@
...
+            right,
+        )
+    } else {
+        (0, data)
+    };
+
+    // Run the fuzzer
+    let mut decoder =
+        neqo_qpack::Decoder::new(neqo_http3::Http3Parameters::default().get_qpack_settings());
+    _ = decoder.decode_header_block(data, stream_id.into());
```
martinthomson: It might be worth having data from the qpack streams involved in this test as well.  Otherwise, we won't get all the interactions between the table updates and header decoding.

larseggert: Check the current version, is that what you had in mind?

neqo-qpack/src/table.rs
```diff
@@ -134,7 +134,7 @@ impl HeaderTable {
     /// `HeaderLookup` if the index does not exist in the static table.
     pub fn get_static(index: u64) -> Res<&'static StaticTableEntry> {
         let inx = usize::try_from(index).or(Err(Error::HeaderLookup))?;
-        if inx > HEADER_STATIC_TABLE.len() {
+        if inx >= HEADER_STATIC_TABLE.len() {
```
mxinden: Preference for a separate pull request, especially since this might be severe, i.e. shouldn't hide in this non-fix pull request.

fuzz/fuzz_targets/qpack.rs
```diff
@@ -0,0 +1,44 @@
...
+        (u16::from_le_bytes(left.try_into().unwrap()) as usize, right)
+    } else {
+        (0, data)
+    };
+
+    // Split remaining data into encoder stream and header block.
+    let (encoder_stream, header_block) = if encoder_stream_len <= data.len() {
+        data.split_at(encoder_stream_len)
+    } else {
+        (data, &[][..])
```
martinthomson: WDYT?
```suggestion
        (&[][..], data)
```

fuzz/fuzz_targets/qpack.rs
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    let mut decoder =
+        neqo_qpack::Decoder::new(neqo_http3::Http3Parameters::default().get_qpack_settings());
```
martinthomson:
```suggestion
    let settings = neqo_http3::Http3Parameters::default().get_qpack_settings();
    let mut decoder = neqo_qpack::Decoder::new(settings);
```

neqo-qpack/src/fuzz.rs
```diff
@@ -0,0 +1,88 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Fuzzing support for QPACK.
+
+#[cfg(feature = "build-fuzzing-corpus")]
```
martinthomson: The conditional compilation here hurts a little.  There are two things here, could they be added to modules (and re-exported) to clarify it a bit?

```rust
#[one]
pub(crate) use one::{...};
#[two]
pub(crate) use two::{...};

#[one]
mod one {

}

#[two]
mod two {
}
```

PR #3161, opened 2025-11-24
feat: Fuzz `HFrame` decoding

larseggert:


neqo-http3/src/frames/reader.rs
```diff
@@ -261,6 +262,8 @@ impl FrameReader {
             self.frame_len,
             if len > 0 { None } else { Some(&[]) },
         )? {
+            #[cfg(feature = "build-fuzzing-corpus")]
```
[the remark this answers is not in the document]
martinthomson: This is unrelated, but you can probably use `(len == 0).then_some(&[])` to simplify.

neqo-http3/src/frames/reader.rs
```diff
@@ -282,8 +286,33 @@ impl FrameReader {
...
+    /// - `data`: The frame payload as a byte slice, or `None` for frames without a payload.
+    ///
+    /// # Output format
+    /// The output consists of the varint-encoded frame type and length, followed by the optional payload data.
+    fn write_item_to_fuzzing_corpus(&self, data: Option<&[u8]>) {
+        // We need to include the frame type and length varints before the data
+        // to create a complete frame that the fuzzer can process.
+        let mut encoder = neqo_common::Encoder::default();
+        encoder.encode_varint(self.frame_type.0);
+        encoder.encode_varint(self.frame_len);
```
martinthomson, line 317: When `data` is `None`, this writes a length that isn't subsequently followed by that much data.  Is that going to work out in the fuzzer?

larseggert: Not really, since it's just a corrupt frame that should be handled gracefully. Or do you think we should skip writing those?

martinthomson: Probably not.  The fuzzers are all capable of producing variants that won't be valid.  That's basically their core function.

PR #3163, opened 2025-11-25
feat: Also save fuzzing samples on encode

larseggert:
Broken out of #3148. This gets us a few more unique samples for the corpora.

### DIFF
--- a/neqo-transport/src/tracking.rs
+++ b/neqo-transport/src/tracking.rs
@@ -457,11 +457,6 @@ impl RecvdPackets {
             return;
         }
 
-        let frame_type = if self.ecn_count.is_some() {
-            FrameType::AckEcn
-        } else {
-            FrameType::Ack
-        };
         let mut iter = ranges.iter();
         let Some(first) = iter.next() else { return };
         stats.largest_acknowledged = first.largest;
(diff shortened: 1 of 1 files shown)


neqo-transport/src/addr_valid.rs
```diff
@@ -419,9 +419,18 @@ impl NewTokenSender {
...
+                let frame_start = builder.len();
+
                 builder.encode_varint(FrameType::NewToken);
                 builder.encode_vvec(&t.token);
 
+                #[cfg(feature = "build-fuzzing-corpus")]
+                neqo_common::write_item_to_fuzzing_corpus(
+                    "frame",
+                    &builder.as_ref()[frame_start..],
+                );
```
martinthomson: Given the general shape of this, would it make sense to define a function that you could call like this?

```rust
builder.encode_frame(FrameType::NewToken, |b| {
  b.encode_vvec(&t.token);
});
```

That wrapper could do the special fuzzing corpus logging, rather than have it spread everywhere.

You might need to make a crate-local trait to add the function implementation to the builder, rather than have that code in `neqo-common`.

neqo-transport/src/tracking.rs
```diff
@@ -475,27 +474,31 @@ impl RecvdPackets {
...
-            builder.encode_varint(r.len() - 1); // Range
-            last = r.smallest;
-        }
 
-        if self.ecn_count.is_some() {
-            builder.encode_varint(self.ecn_count[Ecn::Ect0]);
-            builder.encode_varint(self.ecn_count[Ecn::Ect1]);
-            builder.encode_varint(self.ecn_count[Ecn::Ce]);
-        }
+        builder.encode_frame(frame_type, |b| {
```
mxinden: Nit: `frame_type` could be inlined, or definition could be moved down. No need for it to be defined so early.

Issue #3112, opened 2025-11-06
transport/cc: Expose stats for slow start phase


omansfeld:
For #2858 it would make sense to introduce some metrics for the slow start phase that can be tracked in Firefox telemetry.

I'm thinking of the following:

- `slow_start_exit_ratio`: ratio of connections that exit slow start at all
  - algorithm improvements kinda only apply to those, because the exit point is the biggest differentiator between algos
  - I think this is also generally useful to inform impact of changes to congestion avoidance
- `slow_start_exit_reason`: loss vs non-loss 
  - we want to avoid exit by loss, though for now 100% of exits will be due to loss because that's the only thing we implement right now
  - modern slow start algos like HyStart++ or SEARCH will change that
- `slow_start_distance_to_ideal_exit`: measure the `ssthresh` when exiting slow start and compare to the `cwnd` value right before the first congestion event after slow start
  - this is working on the assumption that an ideal slow start exit would be without packet loss, but as close to the actual congestion point as possible, i.e. spending as much time in slow start as possible without inducing loss
  - being very close means we exited close the the potential congestion point, i.e. don't undershoot
  - this is kind of best effort, because of course the first congestion event after slow start isn't necessarily close to the actual optimal exit point, but I think it is the best indicator we have for slow start efficiency in metrics
  - **I very much welcome alternative design approaches for this metric specifically**

I'll try to write a patch soon. I think it would make sense if this lands closely after #3086 so we can get a bunch of stats into Firefox at once without having to do multiple releases.

mxinden:
> I think it would make sense if this lands closely after [#3086](https://github.com/mozilla/neqo/pull/3086) so we can get a bunch of stats into Firefox at once without having to do multiple releases.

Whatever works best for you. Doing many small releases is fine as well.

> this is kind of best effort, because of course the first congestion event after slow start isn't necessarily close to the actual optimal exit point, but I think it is the best indicator we have for slow start efficiency in metrics

I can not think of a better indicator. That said, not sure it is a good indicator. @mwelzl mentioned at IETF 124 that the cwnd after slow start (i.e. half the slow start exit cwnd) is still usually too high. See example in slide 5 below.

https://datatracker.ietf.org/meeting/120/materials/slides-120-iccrg-ssthresh-after-slow-start-overshoot-01.pdf

Maybe _number of RTT till first loss after slow start_ is better. I have to give this more thought. What does the literature say?

mwelzl:
[...]

Issue #3111, opened 2025-11-05
Lower CWND reduction on ECN CE vs loss


mxinden:
Today our Cubic implementation reacts to ECN CE the same way it reacts to inferred packet loss.

`on_packets_loss` calls `on_congestion_event` the same way that `on_ecn_ce_received`.

https://github.com/mozilla/neqo/blob/6d56ec96fc97b488b0d67df1e72807386fdb23dd/neqo-transport/src/cc/classic_cc.rs#L270-L326

https://github.com/mozilla/neqo/blob/6d56ec96fc97b488b0d67df1e72807386fdb23dd/neqo-transport/src/cc/classic_cc.rs#L328-L334

[RFC 8511](https://www.rfc-editor.org/rfc/rfc8511.html) suggests to react less to an ECN CE than to an inferred packet loss.

Early testing (with a value of `0.9`, RFC suggests `0.8`) via our simulator is promising, note the very small cwnd reductions on CE:

<img width="2825" height="800" alt="Image" src="https://github.com/user-attachments/assets/1fd50c7e-77f1-4f7b-8174-ddc49020a35c" />

//CC @mwelzl

omansfeld:
I think this is really interesting! From the RFC:

> Modern AQM mechanisms can use ECN to signal the early signs of
   impending queue buildup long before a tail-drop queue would be forced
   to resort to dropping packets.  It is therefore appropriate for the
   transport protocol congestion control algorithm to have a more
   measured response when it receives an indication with an early
   warning of congestion after the remote endpoint receives an ECN
   CE-marked packet.

That argumentation makes total sense to me, receiving ECN CE is not as drastic a sign for queue buildup as dropped packets.

> The results of these tests indicate that
   CUBIC connections benefit from beta_{ecn} of 0.85 (cf.  beta_{loss} =
   0.7), and NewReno connections see improvements with beta_{ecn} in the
   range 0.7 to 0.85 (cf. beta_{loss} = 0.5).

FYI the suggested CUBIC value of `0.85`. It's supposed to be higher as the decrease factor for CUBIC is inherently already higher (= less decrese) than NewReno.

> If the method is only deployed by some senders, and not by others,
   the senders using it can gain some advantage, possibly at the expense
   of other flows that do not use this updated method.  Because this
   advantage applies only to ECN-marked packets and not to packet-loss
   indications, an ECN-Capable bottleneck will still fall back to
   dropping packets if a TCP sender using ABE is too aggressive.  The
[...]

PR #3165, opened 2025-11-25
feat(transport/cc): expose whether connection exited slow start as stat

omansfeld:
A first PR exposing the simplest stat discussed in #3112. Other stats like exit reason or approximations as to how close a heuristic was to the ideal exit point don't really apply as long as we only exit through loss/ecn and haven't implemented any heuristics.

But this is pretty useful already and would allow us to get some insight into how many connections actually exit slow start.

I hope with #3086 just merged we can also get this simple patch in quickly and then make a release so I can integrate both into Firefox glean metrics.

In addition to the unit tests below I also manually tested the changes using the `min_bandwidth` simulation with very small transfer amounts.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -345,16 +345,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             return false;
         };
 
-        // State needs to be saved before `on_congestion_event` is called where it changes to
-        // `State::RecoveryStart`.
-        let in_slow_start = self.state.in_slow_start();
-        let congestion = self.on_congestion_event(last_lost_packet, now);
-        if congestion {
-            if in_slow_start {
-                cc_stats.slow_start_exited = true;
-            }
-            cc_stats.congestion_events_loss += 1;
-        }
+        let congestion = self.on_congestion_event(last_lost_packet, false, now, cc_stats);
         let persistent_congestion = self.detect_persistent_congestion(
             first_rtt_sample_time,
             prev_largest_acked_sent,
@@ -381,17 +372,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         now: Instant,
         cc_stats: &mut CongestionControlStats,
     ) -> bool {
-        // State needs to be saved before `on_congestion_event` is called where it changes to
-        // `State::RecoveryStart`.
-        let in_slow_start = self.state.in_slow_start();
-        let cwnd_reduced = self.on_congestion_event(largest_acked_pkt, now);
-        if cwnd_reduced {
-            if in_slow_start {
-                cc_stats.slow_start_exited = true;
-            }
-            cc_stats.congestion_events_ecn += 1;
-        }
-        cwnd_reduced
+        self.on_congestion_event(largest_acked_pkt, true, now, cc_stats)
     }
 
     fn discard(&mut self, pkt: &sent::Packet, now: Instant) {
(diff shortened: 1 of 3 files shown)


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1411,7 +1420,7 @@ mod tests {
     #[test]
     fn spurious_congestion_event_detection() {
         let mut cc = ClassicCongestionControl::new(NewReno::default(), Pmtud::new(IP_ADDR, MTU));
-        let mut now = now();
```
omansfeld, line 1414: small drive-by to make clippy happy, this slipped through #3086

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -345,8 +345,14 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+        // State needs to be saved before `on_congestion_event` is called where it changes to
+        // `State::RecoveryStart`.
+        let in_slow_start = self.state.in_slow_start();
         let congestion = self.on_congestion_event(last_lost_packet, now);
         if congestion {
+            if in_slow_start {
+                cc_stats.slow_start_exited = true;
+            }
             cc_stats.congestion_events_loss += 1;
         }
```
```suggestion
        let was_in_slow_start = self.state.in_slow_start();
        let congestion = self.on_congestion_event(last_lost_packet, now);
        if congestion {
                cc_stats.slow_start_exited |= was_in_slow_start;
                cc_stats.congestion_events_loss += 1;
        }
```
You also have a very similar pattern below. Factor out into helper?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -375,8 +381,14 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+        // State needs to be saved before `on_congestion_event` is called where it changes to
+        // `State::RecoveryStart`.
+        let in_slow_start = self.state.in_slow_start();
         let cwnd_reduced = self.on_congestion_event(largest_acked_pkt, now);
         if cwnd_reduced {
+            if in_slow_start {
+                cc_stats.slow_start_exited = true;
+            }
             cc_stats.congestion_events_ecn += 1;
         }
```
```suggestion
        let was_in_slow_start = self.state.in_slow_start();
        let cwnd_reduced = self.on_congestion_event(largest_acked_pkt, now);
        if cwnd_reduced {
                cc_stats.slow_start_exited |= was_in_slow_start;
                cc_stats.congestion_events_ecn += 1;
        }
```
This is the "similar pattern" mentioned earlier.

neqo-transport/src/cc/classic_cc.rs
larseggert, line 379:
```suggestion
        if !pkt.cc_outstanding() {
                return;
        }
```
Early return would remove indentation.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1527,4 +1536,38 @@ mod tests {
...
+        assert!(cc.state.in_slow_start());
+        assert!(!cc_stats.slow_start_exited);
+
+        let pkt1 = sent::make_packet(1, now, 1000);
+        cc.on_packet_sent(&pkt1, now);
+        cc.on_ecn_ce_received(&pkt1, now, &mut cc_stats);
+
+        assert!(!cc.state.in_slow_start());
+        assert!(cc_stats.slow_start_exited);
+    }
```
larseggert: Can we extract the similarities between these tests into a helper?

neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -218,6 +218,7 @@ fn debug() {
   rx: 0 drop 0 dup 0 saved 0
   tx: 0 lost 0 lateack 0 ptoack 0 unackdrop 0
   cc: ce_loss 0 ce_ecn 0 ce_spurious 0
+  ss_exited: false
```
larseggert:
```suggestion
  ss_exit: false
```
Save two chars :-)

neqo-transport/src/stats.rs
```diff
@@ -392,8 +390,9 @@ impl Debug for Stats {
             "  cc: ce_loss {} ce_ecn {} ce_spurious {}",
             self.cc.congestion_events_loss,
             self.cc.congestion_events_ecn,
-            self.cc.congestion_events_spurious
+            self.cc.congestion_events_spurious,
         )?;
+        writeln!(f, "  ss_exited: {}", self.cc.slow_start_exited)?;
```
larseggert:
```suggestion
        writeln!(f, "  ss_exit: {}", self.cc.slow_start_exited)?;
```

neqo-transport/src/cc/classic_cc.rs
omansfeld: Helper not needed. It's possible to do this in `on_congestion_event` by passing in `ecn: bool` and thus avoid the duplication. We also need said bool if we ever want to do #3111.

PR #3166, opened 2025-11-26
chore: Reformat QNS results

larseggert:
Into two columns to save space.
Screenshot 2025-11-26 at 17 46 34

### DIFF
--- a/.github/workflows/qns.yml
+++ b/.github/workflows/qns.yml
@@ -296,11 +296,11 @@ jobs:
               echo "<table><tr><th align=\"left\">$LATEST as client</th><th align=\"left\">$LATEST as server</th></tr>"
               echo -n "<tr><td valign=\"top\">"
               if [ -e client.failed.md ]; then
-                cmark-gfm -e strikethrough -e autolink client.failed.md | tr -d '\n'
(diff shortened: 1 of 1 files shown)


martinthomson:
Shorter = good, but I still find the QNS results mostly unhelpful.  The level of noise is so high that I've trained myself to ignore them.  That's the thing that I'd like to see improved somehow.

martinthomson:
Oh, one thing you could do is remove "neqo-latest" from all the labels.

larseggert:
I'm trying to reduce the noise, but a bunch of it is due to bugs in other stacks (filed issues).

larseggert:
> Oh, one thing you could do is remove "neqo-latest" from all the labels.

Oh, there's actually a bug then! Because we also run `neqo-latest` (= the PR code) against `neqo`, which is the latest interop runner image we publish. But the results are not part of the table. Let me fix that.

FWIW, #3123 renames `neqo-latest` to `neqo-pr` for clarity, among other changes.

larseggert:
> Oh, there's actually a bug then!

No bug - all tests for `neqo-latest` vs. `neqo` for both the client and server cases succeed, so the "failure" table shouldn't include them 🎉

PR #3167, opened 2025-11-26
chore: Reformat `perfcompare` results

larseggert:
Only show significant changes by default, hide all results in collapsible.
Screenshot 2025-11-26 at 17 47 21


larseggert:
Merging this now, since it only affects CI output.

PR #3168, opened 2025-11-26
chore: Reformat `bench` output

larseggert:
Only show significant changes by default.
Screenshot 2025-11-27 at 11 54 36

### DIFF
--- a/.github/scripts/format-bench-results.py
+++ b/.github/scripts/format-bench-results.py
@@ -29,9 +29,8 @@ def bold_middle_pct(line: str) -> str:
         bracket = line[start + 1 : end - 1]  # Content inside brackets
         parts = bracket.split("%")
         if len(parts) >= 3:
-            # Reconstruct with middle percentage bolded using f-string formatting
             new_bracket = f"{parts[0]}%<b>{parts[1]}</b>%{parts[2]}"
-            return line[:start + 1] + new_bracket + line[end - 1 :]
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -169,22 +169,131 @@ jobs:
         run: |
           SHA=$(cd neqo && git log origin/main -1 --format=%H)
           echo "$SHA" > main-sha.txt
+          # Parse benchmark results into collapsibles using awk.
+          # A benchmark block starts with a non-indented, non-empty line (not "Found"),
+          # and includes all following indented lines plus any "Found" lines.
+          awk '
```
martinthomson: ETOOMUCHAWK

Want to move this to a file?

martinthomson (commented):
The results look good to me.  Not sure that I like the cost, but it's readable code.

(I'd have used python...)

.github/workflows/bench.yml
larseggert: I kinda prefer having workflow-related scripts in the workflows. But if we want to change that for the project, I'd be OK with it, but we should do it in a separate PR and also move other scripts into their own files then. @mxinden thoughts?

mxinden: No opinion on separate-file-vs-in-workflow-file. Preference for more Python and less bash and less awk.

larseggert: I can have Claude write the Python :-)

larseggert:

PR #3170, opened 2025-11-26
chore: Update `toml` to 0.9

larseggert:
Same as Gecko.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1484,8 +1484,11 @@ version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f0dc8b1fb61449e27716ec0e1bdf0f6b8f3e8f6b05391e8497b8b6d7804ea6d8"
 dependencies = [
+ "serde_core",
  "serde_spanned",
  "toml_datetime",
+ "toml_parser",
+ "winnow",
 ]
 
 [[package]]
@@ -1502,6 +1505,15 @@ name = "typenum"
 version = "1.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "562d481066bde0658276a35467c4af00bdc6ee726305698a55b86e61d7ad82bb"
+
+[[package]]
+name = "toml_parser"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c0cbe268d35bdb4bb5a56a2de88d0ad0eb70af5384a99d648cd4b3d04039800e"
+dependencies = [
+ "winnow",
+]
 
 [[package]]
 name = "unicode-ident"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1056,8 +1056,8 @@ dependencies = [
  "libc",
  "log",
  "once_cell",
- "socket2 0.5.7",
- "windows-sys 0.52.0",
+ "socket2",
+ "windows-sys 0.60.2",
```
mxinden, line 1110: Unrelated? Also note that Firefox uses `v0.52.0`:

https://github.com/mozilla-firefox/firefox/blob/main/Cargo.lock#L8242-L8249

neqo-crypto/Cargo.toml
```diff
@@ -26,9 +26,9 @@ thiserror = { workspace = true }
 bindgen = { version = "0.72", default-features = false, features = ["runtime"] }
 mozbuild = { version = "0.1", default-features = false, optional = true }
 semver = { version = "1.0", default-features = false }
-serde = { version = "1.0", default-features = false }
+serde = { version = "1.0", default-features = false, features = ["std"] }
```
mxinden, line 29: Why do we need the `std` feature?

larseggert: https://serde.rs/feature-flags.html#std

Without it 
> the trait `_::_serde::Deserialize<'_>` is not implemented for `std::vec::Vec<std::string::String>`

Cargo.lock
larseggert: Not sure. It came in via `cargo update -p toml --precise 0.9.8`.

larseggert: I think the newer `quinn-udp` in `neqo` is causing this:
```
cargo update windows-sys@0.60.2 --precise 0.52.0
    Updating crates.io index
error: failed to select a version for the requirement `windows-sys = "^0.60"`
candidate versions found which didn't match: 0.52.0
location searched: crates.io index
required by package `socket2 v0.6.1`
    ... which satisfies dependency `socket2 = ">=0.5, <0.7"` (locked to 0.6.1) of package `quinn-udp v0.5.14`
    ... which satisfies dependency `quinn-udp = "^0.5.13"` (locked to 0.5.14) of package `neqo-udp v0.19.0 (/Volumes/mozdrone/Users/mozdrone/Documents/Code/neqo/main/neqo-udp)`
```

neqo-crypto/Cargo.toml
larseggert: And if I just do `features = ["derive"]`
> the trait `_::_serde::Deserialize<'_>` is not implemented for `std::collections::HashMap<std::string::String, Bindings>`

larseggert: @mxinden we good here?

Cargo.lock
mxinden: `quinn-udp` should support `v0.52.0` of `windows-sys`:

https://github.com/quinn-rs/quinn/blob/67fcd79020e0d80387c79279cada3eec7a22458b/Cargo.toml#L55

neqo-crypto/Cargo.toml
mxinden: Yes, fine. Thanks.

Cargo.lock
larseggert: Then I don't know.

larseggert: This seems to have resolved itself after a rebase 🤷

PR #3174, opened 2025-11-27
fix: Off-by-one error in QPACK table size calc

larseggert:
Broken out of and found by #3160.

### DIFF
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -134,7 +134,7 @@ impl HeaderTable {
     /// `HeaderLookup` if the index does not exist in the static table.
     pub fn get_static(index: u64) -> Res<&'static StaticTableEntry> {
         let inx = usize::try_from(index).or(Err(Error::HeaderLookup))?;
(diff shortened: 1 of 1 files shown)


larseggert:
This was approved as part of PR #3160, merging.

neqo-qpack/src/table.rs
```diff
@@ -134,7 +134,7 @@ impl HeaderTable {
     /// `HeaderLookup` if the index does not exist in the static table.
     pub fn get_static(index: u64) -> Res<&'static StaticTableEntry> {
         let inx = usize::try_from(index).or(Err(Error::HeaderLookup))?;
-        if inx > HEADER_STATIC_TABLE.len() {
+        if inx >= HEADER_STATIC_TABLE.len() {
             return Err(Error::HeaderLookup);
         }
         Ok(&HEADER_STATIC_TABLE[inx])
```
martinthomson, line 140: FWIW, a better way to manage this is to do:

```rust
HEADER_STATIC_TABLE.get().ok_or(Error::HeaderLookup)
```

PR #3175, opened 2025-11-27
chore: Fuzz `WebTransportFrame` decoding

larseggert:


neqo-http3/src/frames/wtframe.rs
```diff
@@ -28,17 +28,36 @@ impl WebTransportFrame {
...
+        encoder.encode(data);
+        neqo_common::write_item_to_fuzzing_corpus("wtframe", encoder.as_ref());
     }
 }
 
 impl FrameDecoder<Self> for WebTransportFrame {
     fn decode(frame_type: HFrameType, frame_len: u64, data: Option<&[u8]>) -> Res<Option<Self>> {
         if let Some(payload) = data {
+            #[cfg(feature = "build-fuzzing-corpus")]
+            Self::write_to_fuzzing_corpus(frame_type, frame_len, payload);
```
mxinden: `write_to_fuzzing_corpus` re-encodes `frame_type` and `frame_len`. Couldn't we grab it at a higher level? In case encoding ever changes, `write_to_fuzzing_corpus` wouldn't need to change. Not sure what the best practice is.

larseggert: Check the last change? Not quite what you asked for, but at least we're factoring the writing out for all decided frames now.

PR #3178, opened 2025-11-27
fix: Handle empty directory in `make-fuzz-corpus.sh`

larseggert:
When we add a new fuzz target, its corresponding corpus directory may initially be empty. Support that.

Also don't show Rust backtraces for test suite runs (because there will be failures), and turn off the leak sanitizer while merging corpora. This is not where we care about those issues.


larseggert:

PR #3180, opened 2025-11-28
chore: Use `python` for `perfcompare`

larseggert:
Since we bit the bullet on Python in #3168, this is the other complicated shell script we have in the CI suite.


larseggert:

PR #3181, opened 2025-11-28
fix: qpack idiomization

### LINKED ISSUE #3179: Use `HEADER_STATIC_TABLE.get().ok_or()`
FWIW, a better way to manage this is to do:

```rust
HEADER_STATIC_TABLE.get().ok_or(Error::HeaderLookup)
```

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/3174#discussion_r2570120722_

Also needs a test for the error condition.

larseggert:
Fixes #3179 and rolls in one similar change.

### DIFF
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -134,10 +134,7 @@ impl HeaderTable {
     /// `HeaderLookup` if the index does not exist in the static table.
     pub fn get_static(index: u64) -> Res<&'static StaticTableEntry> {
         let inx = usize::try_from(index).or(Err(Error::HeaderLookup))?;
-        if inx >= HEADER_STATIC_TABLE.len() {
-            return Err(Error::HeaderLookup);
-        }
-        Ok(&HEADER_STATIC_TABLE[inx])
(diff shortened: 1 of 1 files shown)


neqo-qpack/src/table.rs
```diff
@@ -254,14 +245,15 @@ impl HeaderTable {
...
+        while let Some(e) = self.dynamic.back() {
+            if self.used <= reduce {
+                break;
+            }
+            if !e.can_reduce(self.acked_inserts_cnt) {
+                return false;
             }
+            self.used -= u64::try_from(e.size()).expect("usize fits in u64");
+            self.dynamic.pop_back();
         }
```
martinthomson, line 257: We need rust 1.93...

```suggestion
        while self.used > target { // note renamed from reduce
            let Some(e) = self.dynamic.pop_back_if(|e| e.can_reduce(self.acked_inserts_cnt)) else {
                return false;
            };
            self.used -= u64::try_from(e.size()).expect("usize fits in u64");
        }
```

As it is, I think that what you have is fine.  It's just not as idiomatic as you might like.

PR #3194, opened 2025-12-01
chore: Use Codspeed runners for wallclock benches

larseggert:
Since we're considering their paid plan.

CC @sylvestre

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,5 +1,7 @@
 name: CodSpeed
 on:
+  push:
+    branches: ["main"] # To generate a performance baseline.
   pull_request:
     branches: ["main"]
   merge_group:


.github/workflows/codspeed.yml
```diff
@@ -38,36 +38,25 @@ jobs:
...
+                  mode: ($pkg.metadata.bench[.name].codspeed.mode // "instrumentation")
+                }
+              ]'
           } >> "${GITHUB_OUTPUT}"
 
   benchmarks:
     name: Run bench
-    runs-on: ubuntu-24.04
+    runs-on:
+      group: codspeed # zizmor: ignore[self-hosted-runner]
```
adriencaccia: Hey @larseggert, I made the changes on our side, the following config will work:

```suggestion
    runs-on:
      group: codspeed # zizmor: ignore[self-hosted-runner]
      labels: codspeed-macro
```

larseggert: Thank you!

Any hints for how to keep the regular `instrumentation` runners on GitHub? I think as soon as I specify a group, `self-hosted` is implied for runner type.

larseggert: @adriencaccia The jobs still don't seem to be picked up, or does this PR need to be merged to `main` first?
https://github.com/mozilla/neqo/actions/runs/20238767203/job/58100667647?pr=3194

adriencaccia: There was a slight concurrency issue on my side, sorry 😓 
If you re-trigger the workflows, they will work now.

adriencaccia: > Thank you!
> 
> Any hints for how to keep the regular `instrumentation` runners on GitHub? I think as soon as I specify a group, `self-hosted` is implied for runner type.

I had the same issue when testing out the feature. It indeed seems that specifying `group` will enforce using a self-hosted runner. The only workaround I found to keep using default GH runners was to completely omit the `group` key.

I did the following:

```yaml
strategy:
      matrix:
        include:
          - mode: "simulation"
            runs-on: ubuntu-latest
          - mode: "walltime"
            runs-on:
              group: codspeed
              labels: codspeed-macro
    name: Run ${{ matrix.mode }} benchmarks
    runs-on: ${{ matrix.runs-on }}
```

larseggert: Getting linker errors: https://github.com/mozilla/neqo/actions/runs/20239528874/job/58103446593?pr=3194
It's pulling in a cached NSS build, but our cache key does include `runner.arch`...

adriencaccia: It seems that the cached NSS build was built with glibc 2.38.
Our macro runners are using Ubuntu 22.04, which uses glibc 2.35.

Is it possible to add the ubuntu version to the NSS cache key? Since regardless of the runner, the current builds will only work with 24.04

larseggert:
Merging this to debug it.

PR #3196, opened 2025-12-02
feat: Fuzz address validation tokens

larseggert:


neqo-transport/src/server.rs
```diff
@@ -211,6 +211,25 @@ impl Server {
...
+                std::net::IpAddr::V6(ip) => {
+                    let bytes = ip.octets();
+                    d.push(u8::try_from(bytes.len()).expect("IP address len fits in u8"));
+                    d.extend_from_slice(&bytes);
+                }
+            }
+            d.extend_from_slice(&dgram.source().port().to_be_bytes());
+            d.extend_from_slice(&initial.token);
+            neqo_common::write_item_to_fuzzing_corpus("addr_valid", &d);
+        }
```
martinthomson: Given the length of this chunk and the length of the function it is added to, this is worth a new function.

neqo-transport/src/addr_valid.rs
```diff
@@ -131,9 +131,20 @@ impl AddressValidation {
         // Include the token identifier ("Retry"/~) in the AAD, then keep it for plaintext.
         let mut buf = Self::encode_aad(peer_address, retry);
         let encrypted = self.self_encrypt.seal(buf.as_ref(), data.as_ref())?;
+        #[cfg(feature = "build-fuzzing-corpus")]
+        let mut corpus_data = buf.as_ref()[TOKEN_IDENTIFIER_RETRY.len()..].to_vec();
         buf.truncate(TOKEN_IDENTIFIER_RETRY.len());
```
martinthomson: This should really have a static assertion that says that the two token identifiers are identical in length.  This code is hard to parse otherwise.

Or define a new constant from one length and assert the other length matches.

neqo-transport/src/addr_valid.rs
```diff
@@ -131,9 +131,20 @@ impl AddressValidation {
...
         buf.encode(&encrypted);
-        Ok(buf.into())
+        let token: Vec<u8> = buf.into();
+        #[cfg(feature = "build-fuzzing-corpus")]
+        {
+            if !retry {
+                corpus_data.extend_from_slice(&peer_address.port().to_be_bytes());
```
martinthomson: Why are you adding the port number here?  The identifier ensures that it isn't parsed out unless it was encoded.  If it is just to ensure that the decoding is simpler, then maybe set the value to 0.

fuzz/fuzz_targets/addr_valid.rs
```diff
@@ -0,0 +1,37 @@
...
+    use test_fixture::now;
+
+    // Binary format: [4 or 6] [ip octets: 4 or 16 bytes] [port: 2 bytes BE] [token...]
+    let (peer, token) = match data.first() {
+        Some(4) if data.len() >= 7 => {
+            let ip = Ipv4Addr::from([data[1], data[2], data[3], data[4]]);
+            let port = u16::from_be_bytes([data[5], data[6]]);
+            (SocketAddr::new(IpAddr::V4(ip), port), &data[7..])
+        }
+        Some(6) if data.len() >= 19 => {
```
martinthomson: I think that this is 16, not 6.

fuzz/fuzz_targets/addr_valid.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[cfg(all(fuzzing, not(windows)))]
+fuzz_target!(|data: &[u8]| {
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+
+    use neqo_transport::{addr_valid::AddressValidation, server::ValidateAddress};
+    use test_fixture::now;
+
+    // Binary format: [4 or 6] [ip octets: 4 or 16 bytes] [port: 2 bytes BE] [token...]
+    let (peer, token) = match data.first() {
```
martinthomson: For this, I would use a `Decoder`.  `Decoder::decode_vec(1, ...)` would get you the IP address, ...

This manual parsing is sort of OK, but it's quite manual.  It would be easier with help.

fuzz/fuzz_targets/addr_valid.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    use neqo_common::Decoder;
+    use neqo_transport::{addr_valid::AddressValidation, server::ValidateAddress};
+    use test_fixture::now;
+
+    // Binary format: [1-byte length] [ip octets: 4 or 16 bytes] [port: 2 bytes BE] [token...]
+    let mut dec = Decoder::new(data);
+    let Some(ip_bytes) = dec.decode_vec(1) else {
+        return;
+    };
```
martinthomson: Oh, if you want less code, you could define an inner function with a signature of:

```rust
fn decode(data: &[u8]) -> Option<(peer, token)> {
}
```
The nice thing about that sort of function is that you can do this:

```suggestion
    let ip_bytes = dec.decode_vec(1)?;
```

... for all of the `if Some(foo) = blah else { return; };` code you have.  Not a lot in this case, but it might be a useful pattern in other contexts.

PR #3200, opened 2025-12-03
fix: Stop loss monitoring for PMTUD triggering

larseggert:
The current method of retriggering PMTUD based on loss monitorig turns out to trigger very frequently on even somewhat lossy paths, such as some VPNs. This change disables loss monitoring as a trigger for PMTUD. Instead, we trigger PMTUD probing when we think the path has changed, i.e., on migration or when receiving a path challenge.

Also adds a test for the "VPN" case.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -441,11 +441,11 @@ impl Connection {
             events.clone(),
         );
 
-        let mut c = Self {
+        let c = Self {
             role,
             version: conn_params.get_versions().initial(),
             state: State::Init,
-            paths: Paths::default(),
+            paths: Paths::new(conn_params.pmtud_enabled()),
             cid_manager,
             tps: Rc::clone(&tphandler),
             zero_rtt_state: ZeroRttState::Init,
@@ -473,7 +473,6 @@ impl Connection {
             #[cfg(any(test, feature = "build-fuzzing-corpus"))]
             test_frame_writer: None,
         };
-        c.paths.set_pmtud(c.conn_params.pmtud_enabled());
         c.stats.borrow_mut().init(format!("{c}"));
         Ok(c)
     }
--- a/neqo-transport/src/path.rs
+++ b/neqo-transport/src/path.rs
@@ -44,7 +44,7 @@ pub type PathRef = Rc<RefCell<Path>>;
 /// processing a packet.
 /// This structure limits its storage and will forget about paths if it
 /// is exposed to too many paths.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Paths {
     /// All of the paths.  All of these paths will be permanent.
     #[expect(clippy::struct_field_names, reason = "This is the best name.")]
@@ -68,8 +68,16 @@ pub struct Paths {
 }
 
 impl Paths {
-    pub fn set_pmtud(&mut self, pmtud: bool) {
-        self.pmtud = pmtud;
+    #[must_use]
+    pub fn new(pmtud: bool) -> Self {
+        Self {
+            paths: Vec::new(),
+            primary: None,
+            migration_target: None,
+            to_retire: Vec::new(),
+            qlog: Qlog::disabled(),
+            pmtud,
+        }
     }
 
     /// Find the path for the given addresses.


neqo-transport/src/connection/tests/pmtud.rs
```diff
@@ -55,3 +63,164 @@ fn gso_with_max_mtu() {
...
+    let after_challenge = server.stats().frame_tx.path_challenge;
+    assert_eq!(
+        after_challenge,
+        before_challenge + 1,
+        "Server should send PATH_CHALLENGE"
+    );
+
+    // Client receives the PATH_CHALLENGE. This triggers PMTUD restart on its path.
+    let s1 = s1.unwrap();
+    let s1_to_client = Datagram::new(s1.source(), c1.source(), s1.tos(), &s1[..]);
```
mxinden, line 165: Why rebuild the `Datagram`?

neqo-transport/src/connection/tests/pmtud.rs
```diff
@@ -55,3 +63,164 @@ fn gso_with_max_mtu() {
...
+    let server_pmtu_on_vpn = server.plpmtu();
+    let client_pmtu_on_vpn = client.plpmtu();
+    assert!(
+        server_pmtu_on_vpn <= VPN_PATH_MTU,
+        "Server PMTU {server_pmtu_on_vpn} should be <= VPN path MTU {VPN_PATH_MTU}"
+    );
+    assert!(
+        client_pmtu_on_vpn <= VPN_PATH_MTU,
+        "Client PMTU {client_pmtu_on_vpn} should be <= VPN path MTU {VPN_PATH_MTU}"
+    );
```
mxinden: How about checking that `client_pmtu_on_vpn` is larger than `client_pmtu_after_initial`? Otherwise PMTUD might settle e.g. for 1399 on the first path, take the same value on the second, and this test would still succeed, right?

neqo-transport/src/path.rs
```diff
@@ -288,6 +296,10 @@ impl Paths {
             old_path.borrow_mut().probe(stats);
             // TODO(mt) - suppress probing if the path was valid within 3PTO.
         }
+
+        if self.pmtud {
+            path.borrow_mut().pmtud_mut().start(now, stats);
+        }
```
mxinden, line 310: I assume ECN and PMTUD are similar in when we start it for a new path, right? `start_ecn` is called in `migrate`. Could the same be done for PMTUD?

neqo-transport/src/connection/tests/pmtud.rs
larseggert: So the server triggers a path challenge.

mxinden: Ah, yes, of course. Thanks.

neqo-transport/src/path.rs
larseggert: So we could, but there are with this patch some differences with when ECN and PMTUD trigger, and I think ECN needs changes, too. We should do those in a different patch.

larseggert:
@martinthomson the mechanism this removes was originally added based on a suggestion by you - any thoughts?

neqo-transport/src/connection/mod.rs
```diff
@@ -473,6 +473,7 @@ impl Connection {
             #[cfg(any(test, feature = "build-fuzzing-corpus"))]
             test_frame_writer: None,
         };
+        c.paths.set_pmtud(c.conn_params.pmtud_enabled());
```
martinthomson: You could make this an argument to `Paths::new()` instead.  That might be sensible.

neqo-transport/src/path.rs
```diff
@@ -62,9 +62,16 @@ pub struct Paths {
...
     qlog: Qlog,
+
+    /// Whether PMTUD is enabled for this connection.
+    pmtud: bool,
 }
 
 impl Paths {
+    pub fn set_pmtud(&mut self, pmtud: bool) {
+        self.pmtud = pmtud;
+    }
```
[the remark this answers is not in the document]
martinthomson: Oh look, great minds thinking alike?  Who am I kidding.

martinthomson (commented):
This all seems reasonable.  Though it seems like a pretty dramatic change.  I guess that the previous change wasn't complete in that we were probably receiving packets at the current PMTU, but we weren't tracking that, so we were restarting PMTUD too often.  Fixing that can be a separate task.  The trigger condition should have been consistent loss of packets above an MTU size that exceeds 1200, not just random less, but I guess we didn't quite nail that OR there are enough cases where we don't send at those sizes often enough to allow ambient "probing" to keep the MTU high.

The cost of this change is that a reduction in path MTU will be fatal, right?

larseggert:
One that is undetected by either end, I.e., doesn't cause a path challenge, yes. Hence the addition of the "VPN" case.

larseggert:
Plan is to keep this in Nightly while we see if that is an issue.

mxinden:
Another case where the previous logic would fail:

Say that Firefox does a large upload on a high bandwidth-delay-product link. Say that PMTUD discovered an MTU of 1500. Firefox would reset its PMTUD after a loss of [3 packets](https://github.com/mozilla/neqo/blob/18306622d6f446d19736f336e8e707656b63e7fe/neqo-transport/src/pmtud.rs#L39) at the current MTU (i.e. 1500). It is likely for Cubic to overshoot the link bandwidth by more than 3 * 1500. In other words it is likely that more than 3 1500 bytes datagrams are lost repeatedly. Thus PMTUD will reset repeatedly, without any change in the path's MTU.

> The cost of this change is that a reduction in path MTU will be fatal, right?

Yes. Given the above, whatever mechanism we come up with, it would need to account for the link characteristics, e.g. a constant of 3 won't do on a high bandwidth-delay link.

That said, agreed with @larseggert above that we should validate on Firefox Nightly whether those reductions in path MTU are frequent enough to warrant such a (potentially complex) mechanism.

PR #3202, opened 2025-12-03
chore: Reduce randomization in benches

larseggert:
In the hopes of stabilizing the results some.


larseggert:
Don't know how this helps until we land it.

larseggert:
The benches aren't involved in corpus generation – I was hoping this would reduce variances in the benchmark comparisons.

PR #3207, opened 2025-12-05
feat(transport): make max_udp_payload_size tparam configurable

mxinden:
Some QUIC servers use an initial MTU larger than 1280 bytes. This is problematic when connecting to such servers via a proxy (e.g. MASQUE connect-udp). While most unproxied paths can handle >1280 bytes, some proxied paths cannot.

Firefox will need to restrict the max_udp_payload_size to 1232 bytes (i.e. 1232 + 40 (v6) + 8 (UDP)) on proxied connections to support such restrictive paths.


neqo-transport/src/connection/params.rs
```diff
@@ -122,6 +122,13 @@ pub struct ConnectionParameters {
     max_streams_bidi: u64,
     /// Initial limit on unidirectional streams that this endpoint creates.
     max_streams_uni: u64,
+    /// > The maximum UDP payload size parameter is an integer value that limits
+    /// > the size of UDP payloads that the endpoint is willing to receive. UDP
+    /// > datagrams with payloads larger than this limit are not likely to be
+    /// > processed by the receiver.
+    ///
+    /// <https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2>
+    max_udp_payload_size: Option<u64>,
```
larseggert, line 131: No other `Option`s in this struct. I guess we need/want it here though?

larseggert (approved):

neqo-transport/src/connection/params.rs
mxinden: I believe `Option` is the right way to go here.

- We only need `max_udp_payload_size` for MASQUE inner connections.
- The default value is valid in all other cases.
    > The default for this parameter is the maximum permitted UDP payload of 65527.
- By making it `Option`al, we can omit it in all but the MASQUE use-case.
- I wish we wouldn't even need it for MASQUE inner connections, as it disabled PMTUD on the server side.

Now there is an argument to be made, that inner MASQUE connections should not differ in QUIC transport parameters for the sake of not sticking out. I am indifferent.

martinthomson:
Thinking about this more, how does this really help with connect-udp?  If the proxy has limits on the size of things it can handle in the server-to-client direction, that might be cause for us to let the server know of those limits, but that's not a limit we are made aware of, is it?  Any limit we set should be based on our own limits, which we only have by virtue of knowing the local interface MTU (maybe), because we don't have any real limits on receiving ourselves.

In the client-to-server direction, we might benefit from knowing about both server and proxy limits, but those are just things we can use to restrict our MTU; we don't need configuration for that.

larseggert:
@mxinden see @martinthomson's comments – anything you want to address in response, or should I click the merge button?

mxinden:
> If the proxy has limits on the size of things it can handle in the server-to-client direction, that might be cause for us to let the server know of those limits, but that's not a limit we are made aware of, is it?

Correct. Not a limit we are aware of. Thus far, we have simply applied this conservative limit to all **proxied** client<->server connections.

This is a trade-off. We can:
- either not set `max_udp_payload_size`, thus using > 1280 MTU on paths (server -> client) that support it,
- set `max_udp_payload_size` and thus support paths (server -> client) with an 1280 MTU AND servers that start with an > 1280 MTU (unless `max_udp_payload_size` is set).

Alternatively we could fallback to HTTP CONNECT on paths with an 1280 MTU AND servers that start with an > 1280 MTU.

@martinthomson can you think of any other approaches?

Would it make sense to advocate major players (e.g. Fastly) to start with an MTU of 1280 and using PMTUD thereafter, instead of starting with a larger initial MTU?

mxinden:
Friendly ping @martinthomson. Do you have thoughts on the above?

Propheticus:
> Would it make sense to advocate major players (e.g. Fastly) to start with an MTU of 1280

(First of, please let me know if this is annoying and I'll remove and refrain from commenting)

Reading the [RFC9000 spec section on datagram size](https://datatracker.ietf.org/doc/html/rfc9000#name-datagram-size), this looks like the right approach.  

Specifically section 14.2 states

>An endpoint **SHOULD** use DPLPMTUD ([Section 14.3](https://datatracker.ietf.org/doc/html/rfc9000#dplpmtud)) or PMTUD ([Section 14.2.1](https://datatracker.ietf.org/doc/html/rfc9000#pmtud)) to determine whether the path to a destination will support a desired maximum datagram size without fragmentation. In the absence of these mechanisms, QUIC endpoints **SHOULD NOT** send datagrams larger than the smallest allowed maximum datagram size.

and the section 14.3 explaining DPLPMTUD states

>Endpoints **SHOULD** set the initial value of BASE_PLPMTU ([Section 5.1](https://www.rfc-editor.org/rfc/rfc8899#section-5.1) of [[DPLPMTUD](https://datatracker.ietf.org/doc/html/rfc8899)]) to be consistent with QUIC's smallest allowed maximum datagram size. The MIN_PLPMTU is the same as the BASE_PLPMTU.

The smallest allowed maximum is 1200. (which is also the minimum, so anything smaller needs padding to this size.)

But then again 14.1 on initial size contains
> Datagrams containing Initial packets **MAY** exceed 1200 bytes if the sender believes that the network path and peer both support the size that it chooses.

To me it's unclear how or why the sender would assume / believe anything without negotiating or discovering first though.
All you can safely assume is
> QUIC assumes a minimum IP packet size of at least 1280 bytes

----
So yes it would make sense to advocate senders, like Fastly, to start at an MTU of 1280.

Using `max_udp_payload_size` which is meant to (and is documented to) convey limits in the capabilities of the _endpoint_ to solve limitations encountered in the _path_ feels hacky. It could work but introduces deviations that in 5 years time will be a nice headache for an engineer to troubleshoot because it's unexpected behaviour.
Fixing someone else's problems like this is friendly but also adds debt that can keep growing with other workarounds needed to fix side-effects of the out-of-spec usage.

mxinden:
> (First of, please let me know if this is annoying and I'll remove and refrain from commenting)

High quality contributions, e.g. yours above, are always very welcome. Thanks @Propheticus!

martinthomson:
The real challenge here is that servers that start with a too-large MTU will simply fail to connect in many cases.  Transport parameters are almost deliberately not available to servers in the first round trip.  This is because we take steps to shuffle the TLS ClientHello in ways that make it likely that transport parameters don't appear in the first packet.  In that case, anything we say in transport parameters isn't going to help much.

In the case where the proxy has symmetric limits (same size limit on up- and down-stream -- a very reasonable assumption, even if not 100% guaranteed), there might be some value in us changing our actions, even if signaling doesn't work.  But that is limited to helping servers that echo the client MTU size.

Those servers will already work best, so there is some value in encouraging servers to mirror the client packet size in their responses.  That would allow us to use information we have about the proxy MTU size limits to affect connection success rates.  It also lets us potentially probe for the handshake (first transmission with a high MTU, subsequent/retransmission packets with lower; ACKs of the first confirm that the higher MTU works).

I guess that my conclusion is that the transport parameter isn't much help, except in communications with the proxy.  It's analogous to knowing the MTU of the associated network interface.  It's not everything, but if you make some assumptions you can do a little better than the naive 1200.

larseggert:
@mxinden @martinthomson anything left to discuss here? Can this be merged?

martinthomson:
I'm of a view that this isn't going to help anything.  We don't have constraints on our own ability to handle different MTUs and that is all this signaling can communicate.  I'm interested in where @mxinden stands.

mxinden:
This is a (hacky) solution to a problem we are facing with websites behind Fastly's CDN through the Fastly proxy, namely for the CDN to start with a too-high MTU.

Agreed with @Propheticus and @martinthomson that this is a hack.

Once we resume the MASQUE project I will give this more thought. For now, marking as draft. Thanks for the input everyone.

Issue #3208, opened 2025-12-05
transport/fc.rs: Hit MOZ_CRASH(attempt to subtract with overflow)


mxinden:
@KershawChang hit an underflow in:

https://github.com/mozilla/neqo/blob/d07039370f55e424777c2b24407d0fb01f2d5a6d/neqo-transport/src/fc.rs#L426

I am unable to reproduce this thus far. I will add a `debug_assert!` to crash with more context next time.

mxinden:
Good news. @tysmith found a reproducer in https://bugzilla.mozilla.org/show_bug.cgi?id=2007159. I will take a look.

PR #3209, opened 2025-12-05
chore(transport/fc): add debug_assert to catch underflow

mxinden:
Needed for https://github.com/mozilla/neqo/issues/3208.

### DIFF
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -423,6 +423,12 @@ where
             max_window,
         );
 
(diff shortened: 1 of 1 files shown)


larseggert (approved):

mxinden:
> [codecov/patch](https://github.com/mozilla/neqo/pull/3209/checks?check_run_id=57248314815)
codecov/patch — 66.66% of diff hit (target 93.49%)

If I only knew how to hit it, I wouldn't need the `debug_assert!`.

@larseggert would you mind merging manually? I don't have the permissions.

PR #3216, opened 2025-12-08
chore: Bump action dependencies

larseggert:
Since Dependabot only does workflows.


larseggert:
Closing, since #3215 worked.

PR #3115, opened 2025-11-06
feat(transport/cc): lower CWND reduction on ECN CE vs loss

mxinden:
Work in progress.

Fixes https://github.com/mozilla/neqo/issues/3111.

PR #3233, opened 2025-12-08
feat(transport/cc): lower cwnd reduction on ECN-CE vs loss

### LINKED ISSUE #3111: Lower CWND reduction on ECN CE vs loss

omansfeld:
This changes the cubic multiplicative decrease factor to `0.85` when the congestion event was triggered by ECN according to [RFC 8511: TCP Alternative Backoff with ECN](https://www.rfc-editor.org/rfc/rfc8511.html):

> The results of these tests indicate that CUBIC connections benefit from beta_{ecn} of 0.85

This PR builds on @mxinden's draft in #3115 making tests pass, adding a constant and changing the factor to the recommended `0.85` value. I will delete that draft PR if there are no concerns.

Max and I discussed just landing this to nightly without the overhead of experimentation. Opinions? 

Closes #3111.


neqo-transport/src/cc/cubic.rs
```diff
@@ -135,7 +135,15 @@ impl Cubic {
     ///
     /// For implementation reasons neqo uses a dividend and divisor approach with `usize` typing to
     /// construct `CUBIC_BETA = 0.7`.
-    pub const BETA_USIZE_DIVIDEND: usize = 7;
+    pub const BETA_USIZE_DIVIDEND: usize = 70;
```
larseggert: Why this change? (And also for `BETA_USIZE_DIVISOR` below?)

omansfeld: Because otherwise it's not possible to build the new `0.85` factor for ECN from a `usize`. I preferred changing the `BETA_USIZE_DIVISOR` to `100` and `BETA_USIZE_DIVIDEND` to `70` instead of making 2 new variables for ECN (divisor and dividend).

omansfeld: I will add comments explaining this (as copilot also suggested above).

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -94,6 +94,7 @@ pub trait WindowAdjustment: Display + Debug {
         curr_cwnd: usize,
         acked_bytes: usize,
         max_datagram_size: usize,
+        ecn: bool,
```
mxinden: Slight preference for a new enum:

```suggestion
        congestion_event: CongestionEvent,
```

``` rust
enum CongestionEvent {
  Loss,
  Ecn,
}
```

larseggert: Yes! Also in `slow_start_exit_stats` and `on_congestion_event` please.

larseggert: Could then make `CongestionControlStats::congestion_events` an `EnumMap<CongestionEvent>` instead of having two `bool`s. (Add `Spurious` to @mxinden's `CongestionEvent`.)

neqo-transport/src/cc/cubic.rs
```diff
@@ -411,10 +423,17 @@ impl WindowAdjustment for Cubic {
 
         // Reducing the congestion window and resetting time
         self.t_epoch = None;
-        (
-            curr_cwnd * Self::BETA_USIZE_DIVIDEND / Self::BETA_USIZE_DIVISOR,
-            acked_bytes * Self::BETA_USIZE_DIVIDEND / Self::BETA_USIZE_DIVISOR,
-        )
+        if ecn {
```
martinthomson:
```rust
let m = if ecn { Self::BETA_USIZE_DIVIDEND_ECN } else { Self::BETA_USIZE_DIVIDEND };
(curr_cwnd * m / Self::BETA_USIZE_DIVISOR, acked_bytes * m / Self::BETA_USIZE_DIVISOR)
```

neqo-transport/src/cc/cubic.rs
martinthomson: You could use a dividend of 20 and beta values of 14 and 17.  Then you are 5x further from overflow, but I doubt we'll get that far.

omansfeld: If my math is correct for 64-bit systems we'd need a CWND of `217020 TB` to hit overflow.

I guess it could maybe happen on 32-bit systems. I think there we'd hit overflow around `50 MB` which doesn't seem too unrealistic. So going for 20/14/17 might actually be the safer option.

Could also use checked math to make sure it doesn't just silently wrap around.

`cwnd.checked_mul(85).unwrap_or(usize::MAX) / 100`

Then we'd even be able to add an assert.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1379,7 +1379,7 @@ mod tests {
     #[test]
     fn ecn_ce() {
         let now = now();
-        let mut cc = ClassicCongestionControl::new(NewReno::default(), Pmtud::new(IP_ADDR, MTU));
+        let mut cc = ClassicCongestionControl::new(Cubic::default(), Pmtud::new(IP_ADDR, MTU));
```
larseggert, line 1381: @omansfeld is this change deliberate?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1542,10 +1544,14 @@ mod tests {
...
-        } else {
-            cc.on_packets_lost(Some(now), None, PTO, &[pkt1], now, &mut cc_stats);
+        match congestion_event {
+            CongestionEvent::Ecn => {
+                cc.on_ecn_ce_received(&pkt1, now, &mut cc_stats);
+            }
+            CongestionEvent::Loss => {
+                cc.on_packets_lost(Some(now), None, PTO, &[pkt1], now, &mut cc_stats);
+            }
+            CongestionEvent::Spurious => panic!("unsupported congestion event"),
```
larseggert, line 1553: Or do we want to silently ignore this?

larseggert: Ah, this is test code.

neqo-transport/src/cc/classic_cc.rs
larseggert: @mxinden do you know? Otherwise I will revert this and merge the PR now. We can always make the change in  different PR if it was deliberate.

larseggert: I see it is deliberate, because the test otherwise fails.

larseggert:
@omansfeld sorry for stepping in here, I'm just trying to get this merged.

omansfeld:
> @omansfeld sorry for stepping in here, I'm just trying to get this merged.

No worries, I was busy with other things and on PTO. I can take another look and also review your additions tomorrow.

PR #3234, opened 2025-12-08
feat: emit event on QUIC datagram sent

mxinden:
Work-in-progress

Needed for MASQUE to notify inner connection that it can add more datagrams to the outer connection QUIC datagram queue.


mxinden:
Replaced by https://github.com/mozilla/neqo/pull/3859.

PR #3238, opened 2025-12-09
feat: `send_atomic_with` to avoid `Encoder` alloc

larseggert:
This might remove a few allocations.

### DIFF
--- a/neqo-http3/src/buffered_send_stream.rs
+++ b/neqo-http3/src/buffered_send_stream.rs
@@ -111,11 +111,11 @@ impl BufferedStream {
         let Some((stream_id, _)) = self.prepare_atomic_send(conn, now)? else {
             return Ok(false);
         };
-        let res = conn.stream_send_atomic(stream_id, to_send)?;
-        if res {
+        let sent = conn.stream_send_atomic(stream_id, to_send)?;
+        if sent {
             qlog::h3_data_moved_down(conn.qlog_mut(), stream_id, to_send.len(), now);
         }
-        Ok(res)
+        Ok(sent)
     }
 
     /// Encode data using the provided closure and send it atomically.


neqo-http3/src/buffered_send_stream.rs
```diff
@@ -100,9 +123,13 @@ impl BufferedStream {
         if !buf.is_empty() {
             return Ok(false);
         }
-        let res = conn.stream_send_atomic(*stream_id, to_send)?;
+        // Use the internal buffer as scratch space for encoding.
+        f(&mut Encoder::new_borrowed_vec(buf));
```
martinthomson, line 138: So for the variant where we are given a `&[u8]`, this will create a copy unnecessarily.  The underlying stream functions take `&[u8]` happily.

The trick is working out what to do in the case where you don't want to encode in order to get that `&[u8]`.  It seems to me that you want a think that produces a `&[u8]`, either because it has one already, or because it has an encoder function (`f` here) and you can combine that with the `buf` that you are lending it.

All of that sounds like a trait to me.  Basically, the old function was fine, but you want to create a new form that shares almost all of the implementation, apart from where you call `f()`.

martinthomson (changes requested):

neqo-http3/src/buffered_send_stream.rs
```diff
@@ -92,21 +108,44 @@ impl BufferedStream {
...
-        if !buf.is_empty() {
-            return Ok(false);
-        }
-        let res = conn.stream_send_atomic(*stream_id, to_send)?;
+        let res = conn.stream_send_atomic(stream_id, to_send)?;
         if res {
-            qlog::h3_data_moved_down(conn.qlog_mut(), *stream_id, to_send.len(), now);
+            qlog::h3_data_moved_down(conn.qlog_mut(), stream_id, to_send.len(), now);
         }
         Ok(res)
```
martinthomson:
```suggestion
        let sent = conn.stream_send_atomic(stream_id, to_send)?;
        if sent {
            qlog::h3_data_moved_down(conn.qlog_mut(), stream_id, to_send.len(), now);
        }
        Ok(sent)
```

Rename to match below...sort of.

PR #3171, opened 2025-11-26
fix(transport/quic-datagram): remove inbound queue limit

mxinden:
work-in-progress

Fixes https://github.com/mozilla/neqo/issues/1373.

PR #3239, opened 2025-12-09
fix: Iterate once in `add_datagram`

larseggert:


mxinden (commented):
Check would be removed in https://github.com/mozilla/neqo/pull/3171. I don't think we need to move forward here.

larseggert:
Oh, nice

larseggert:
Closing in favor of #3171.

PR #3241, opened 2025-12-09
chore: Run less CI on merge to `main`

larseggert:
Run only those workflows that are either required checks or that create `main`-branch cached state for others.

Especially skipping benchmarking should help reduce load on the self-hosted runners.

Also run a fuller set of CI on Dependabot updates, to catch more issues.


larseggert:
Merging, since this is just CI. (Let me know if you prefer me to not just merge CI things – I want to keep the review load down.)

mxinden (commented):
:+1:

PR #3243, opened 2025-12-10
chore: Use matrix for `clippy`

larseggert:
Instead of just `nightly`, since we need to make sure our lints/suppressions also work for MSRV.

### DIFF
--- a/neqo-common/src/datagram.rs
+++ b/neqo-common/src/datagram.rs
@@ -149,7 +149,7 @@ impl<D: AsRef<[u8]>> AsRef<[u8]> for Datagram<D> {
 /// batch have the same size. The last datagram may be equal or smaller.
 #[derive(Clone, PartialEq, Eq, derive_more::Debug)]
 #[debug(
-    "DatagramBatch {:?} {:?}->{:?} {:?}: {}",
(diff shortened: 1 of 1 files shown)


larseggert:
Put this on the back burner until we're on 1.87; too many changes otherwise.

neqo-common/src/datagram.rs
```diff
@@ -160,19 +160,19 @@ impl<D: AsRef<[u8]>> AsRef<[u8]> for Datagram<D> {
...
     d: Vec<u8>,
 }
 
-impl Debug for DatagramBatch {
+impl Debug for Batch {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(
             f,
-            "DatagramBatch {:?} {:?}->{:?} {:?}: {}",
+            "Batch {:?} {:?}->{:?} {:?}: {}",
```
martinthomson: Is this a change you want?  Debug statements don't have module context, so it might be better to say "Datagrams" or similar.

larseggert: This is one of the things that 1.81 complains about but later versions don't. I expect to roll this back when we are on 1.87 in O(days).

mxinden (approved):

PR #3253, opened 2025-12-12
chore: remove `Qlog::add_event_with_instant`

larseggert:
Remove dead code `add_event_with_instant` which was only used in its own test. Update the test to use `add_event_data_with_instant` instead, which is the function actually used throughout the codebase (14 call sites).

This also fixes a missed mutant for `add_event_data_with_instant`.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -56,8 +56,7 @@ impl Qlog {
             // As a server, the original DCID is chosen by the client. Using
             // create_new() prevents attackers from overwriting existing logs.
             .create_new(true)
-            .open(&qlog_path)
-            .map_err(qlog::Error::IoError)?;
+            .open(&qlog_path)?;
 
         let streamer = QlogStreamer::new(
             qlog::QLOG_VERSION.to_string(),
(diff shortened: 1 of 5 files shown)


neqo-common/src/qlog.rs
martinthomson, line 60: I can comment outside the diff!

Might be worth looking into `thiserror` for mapping this error with `#[from]`.

neqo-common/src/qlog.rs
```diff
@@ -97,19 +97,6 @@ impl Qlog {
...
-        self.add_event_with_stream(|s| {
-            if let Some(evt) = f() {
-                s.add_event_with_instant(evt, now)?;
-            }
-            Ok(())
-        });
-    }
-
     /// If logging enabled, closure may generate an event to be logged.
     pub fn add_event_data_with_instant<F>(&mut self, f: F, now: Instant)
```
martinthomson: Given that we don't have an `add_event()` or any other variant, I think that you want to name this `add_event_at()` and the one below `add_event()`.  I think that users will be able to handle the difference and being more succinct is good.

martinthomson (approved):

neqo-common/src/qlog.rs
martinthomson: Note that GitHub has a bug here.  The review finalization dialog did not show this comment at all.  But I was able to make it.

larseggert: Don't need `thiserror` even.

PR #3255, opened 2025-12-12
chore: improve mutation coverage for `neqo-udp`

larseggert:
Add unit tests to eliminate missed mutants identified by `cargo mutants`:

- `is_emsgsize`: Tests for Unix (`EMSGSIZE`, `EAGAIN`) and Windows (`WSAEMSGSIZE`, `WSAEINVAL`, `WSAEWOULDBLOCK`) error codes
- `max_gso_segments`: Assert returns at least 1
- `may_fragment`: Exercise the method (platform-dependent return value)

Also use `max_gso_segments()` in test to address `TODO` item.

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -398,14 +398,6 @@ mod tests {
         Ok(())
     }
 
-    #[test]
-    fn may_fragment_returns_bool() -> Result<(), io::Error> {
-        let s = socket()?;
-        // Just exercise the method - the return value is platform-dependent.
-        let _: bool = s.may_fragment();
-        Ok(())
-    }
-
     /// Expect [`Socket::recv`] to handle multiple [`Datagram`]s on GRO read.
     #[test]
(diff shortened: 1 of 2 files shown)


neqo-udp/src/lib.rs
```diff
@@ -349,6 +349,63 @@ mod tests {
...
+        Ok(())
+    }
+
+    #[test]
+    fn may_fragment_returns_bool() -> Result<(), io::Error> {
+        let s = socket()?;
+        // Just exercise the method - the return value is platform-dependent.
+        let _: bool = s.may_fragment();
+        Ok(())
+    }
```
mxinden: This is already enforced at the type system level, i.e. the function can by definition not return anything other than `bool`. Is it worth the noise?

larseggert: Your call, am OK to remove.

mxinden: Preference to remove.

PR #3257, opened 2025-12-12
chore: add `http3` tests to eliminate missed mutants

larseggert:
Add unit tests to catch some mutations identified by `cargo mutants`:

- `features/mod.rs`: Test `NegotiationState::locally_enabled()` returns false when disabled
- `conn_params.rs`: Test QPACK table size limit boundary validation
- `priority.rs`: Test `priority_update_sent()` state update and invalid urgency handling in `from_bytes()`
- `headers_checks.rs`: Test `:protocol` requires CONNECT method, classic CONNECT validation, request/response header requirements
- `lib.rs`: Test `Http3StreamInfo::is_http()` for different stream types
- `frames/wtframe.rs`: Test `is_known_type()` and frame size limit validation
- `client_events.rs`: Test `has_events()` empty and non-empty states
- `server_connection_events.rs`: Test `has_events()` empty and non-empty states

### DIFF
--- a/neqo-http3/src/headers_checks.rs
+++ b/neqo-http3/src/headers_checks.rs
@@ -209,17 +209,9 @@ mod tests {
 
     #[test]
     fn invalid_scheme_webtransport_connect() {
-        assert!(headers_valid(
-            &[
-                Header::new(":method", "CONNECT"),
-                Header::new(":protocol", "webtransport"),
-                Header::new(":authority", "something.com"),
-                Header::new(":scheme", "http"),
-                Header::new(":path", "/here"),
-            ],
-            MessageType::Request
-        )
-        .is_err());
+        let mut headers = create_connect_headers();
+        headers[2] = Header::new(":scheme", "http");
+        assert!(headers_valid(&headers, MessageType::Request).is_err());
     }
 
     #[test]
@@ -259,13 +251,8 @@ mod tests {
     #[test]
     fn protocol_requires_connect_method() {
         // :protocol is only valid with CONNECT method.
-        let headers = vec![
-            Header::new(":method", "GET"),
-            Header::new(":protocol", "webtransport"),
-            Header::new(":scheme", "https"),
-            Header::new(":authority", "something.com"),
-            Header::new(":path", "/here"),
-        ];
+        let mut headers = create_connect_headers();
+        headers[0] = Header::new(":method", "GET");
         assert!(headers_valid(&headers, MessageType::Request).is_err());
     }
 
--- a/neqo-http3/src/settings.rs
+++ b/neqo-http3/src/settings.rs
@@ -406,13 +406,17 @@ mod tests {
         }
 
         // Duplicate: first wins.
-        let mut enc = Encoder::new();
-        enc.encode_varint(SETTINGS_H3_DATAGRAM).encode_varint(1u64);
-        enc.encode_varint(SETTINGS_H3_DATAGRAM_DRAFT04)
-            .encode_varint(0u64);
-        let mut s = HSettings::new(&[]);
-        s.decode_frame_contents(&mut enc.as_decoder()).unwrap();
-        assert_eq!(s.get(HSettingType::EnableH3Datagram), 1);
+        for (first, second, expected) in [
+            (SETTINGS_H3_DATAGRAM, SETTINGS_H3_DATAGRAM_DRAFT04, 1),
+            (SETTINGS_H3_DATAGRAM_DRAFT04, SETTINGS_H3_DATAGRAM, 0),
+        ] {
+            let mut enc = Encoder::new();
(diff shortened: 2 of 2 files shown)


neqo-http3/src/headers_checks.rs
```diff
@@ -255,4 +255,76 @@ mod tests {
...
+    #[test]
+    fn protocol_requires_connect_method() {
+        // :protocol is only valid with CONNECT method.
+        let headers = vec![
+            Header::new(":method", "GET"),
+            Header::new(":protocol", "webtransport"),
+            Header::new(":scheme", "https"),
+            Header::new(":authority", "something.com"),
+            Header::new(":path", "/here"),
+        ];
```
mxinden: Could you use `create_connect_headers` here?

neqo-http3/src/headers_checks.rs
```diff
@@ -255,4 +255,76 @@ mod tests {
         let headers = vec![Header::new(":status", "not-a-number")];
         assert!(is_interim(&headers).is_err());
     }
+
+    #[test]
+    fn protocol_requires_connect_method() {
+        // :protocol is only valid with CONNECT method.
```
mxinden, line 253: Do we have a test testing the inverse, i.e. a `:protocol` without `CONNECT`?

neqo-http3/src/lib.rs
```diff
@@ -652,3 +652,87 @@ impl CloseType {
...
+            (Error::HttpMissingSettings, 0x10a),
+            (Error::HttpRequestRejected, 0x10b),
+            (Error::HttpRequestCancelled, 0x10c),
+            (Error::HttpRequestIncomplete, 0x10d),
+            (Error::HttpMessage, 0x10e),
+            (Error::HttpConnect, 0x10f),
+            (Error::HttpVersionFallback, 0x110),
+        ] {
+            assert_eq!(error.code(), expected);
+        }
```
mxinden, line 702: Isn't this basically testing every constant? Do we need to test the values of constants?

neqo-http3/src/settings.rs
```diff
@@ -384,4 +384,78 @@ mod tests {
...
+        }
+
+        // Duplicate: first wins.
+        let mut enc = Encoder::new();
+        enc.encode_varint(SETTINGS_H3_DATAGRAM).encode_varint(1u64);
+        enc.encode_varint(SETTINGS_H3_DATAGRAM_DRAFT04)
+            .encode_varint(0u64);
+        let mut s = HSettings::new(&[]);
+        s.decode_frame_contents(&mut enc.as_decoder()).unwrap();
+        assert_eq!(s.get(HSettingType::EnableH3Datagram), 1);
```
martinthomson: So, if this were to put the DRAFT04 thing first, would that win?  Why not test that?

neqo-http3/src/settings.rs
```diff
@@ -384,4 +384,78 @@ mod tests {
...
+    }
+
+    fn make_0rtt_token(settings: &[(u64, u64)]) -> Vec<u8> {
+        let mut enc = Encoder::new();
+        enc.encode_varint(SETTINGS_ZERO_RTT_VERSION);
+        for (k, v) in settings {
+            enc.encode_varint(*k).encode_varint(*v);
+        }
+        enc.into()
+    }
```
martinthomson, line 429: This could move inside the function.

neqo-http3/src/settings.rs
```diff
@@ -384,4 +384,78 @@ mod tests {
...
+
+        // Token requests connect=1 but server doesn't have it: rejected.
+        let token = make_0rtt_token(&[(SETTINGS_ENABLE_CONNECT_PROTOCOL, 1)]);
+        assert_eq!(checker.check(&token), ZeroRttCheckResult::Reject);
+
+        // Server with connect enabled.
+        let params = Http3Parameters::default().connect(true);
+        let checker = HttpZeroRttChecker::new(params);
+        let token = make_0rtt_token(&[(SETTINGS_ENABLE_CONNECT_PROTOCOL, 1)]);
+        assert_eq!(checker.check(&token), ZeroRttCheckResult::Accept);
```
martinthomson, line 463: You should add a test with an invalid token, right?  What happens if you send in `&token[1..]` ?

neqo-http3/src/lib.rs
larseggert: There are mutants that change the values associated with the variants that are otherwise not caught. Do we care? :shrug:

martinthomson: For enums where the value is something that appears in specifications, it probably isn't a bad idea to have tests.

The value of a test is primarily that you have two places in the code that have the codepoint mapping.  A sort of double-check.  That's of pretty marginal value for such a direct mapping like this.

PR #3259, opened 2025-12-15
chore: Address new clippy warnings

larseggert:
And apparently something for reordered in `Cargo.lock`.


larseggert:
We can of course suppress. I thought that because we don't care about the value, it doesn't matter which representation it's in.

PR #3267, opened 2025-12-15
fix(common/qlog): disable qlog on error on inner (shared) option

mxinden:
Without setting the inner Option to None, other references will attempt to write logs to a previously errored qlog stream. By setting the inner Option to None on error, the current reference notifies all other references that an error occured.

See conversation in https://github.com/mozilla/neqo/pull/3129#discussion_r2607855548.


larseggert:
Thanks for the fix!

PR #3276, opened 2025-12-16
chore: Some simple additions to drive up coverage

larseggert:
On the fence about the Dispay/Debug checks.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -226,11 +226,12 @@ mod test {
 
     #[test]
     fn add_event_with_stream_error_disables_logging() {
-        let (mut log, _contents) = test_fixture::new_neqo_qlog();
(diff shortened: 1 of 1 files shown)


neqo-common/src/qlog.rs
```diff
@@ -217,4 +217,20 @@ mod test {
...
     }
+
+    #[test]
+    fn shared_streamer_debug() {
+        let (log, _contents) = test_fixture::new_neqo_qlog();
+        assert!(format!("{log:?}").contains("Qlog writing to"));
+    }
+
+    #[test]
+    fn add_event_with_stream_error_disables_logging() {
```
martinthomson, line 228: Any way to test whether something was added to the log?  I know that might be more work, but it isn't like this test actually tests what it purports to.

PR #3277, opened 2025-12-16
chore: Use strum instead of manual `FromStr`

larseggert:


neqo-bin/src/lib.rs
```diff
@@ -122,7 +123,9 @@ pub struct QuicParameters {
     /// The initial round-trip time, in milliseconds.
     pub initial_rtt_ms: u64,
 
-    #[arg(long = "cc", default_value = "cubic")]
+    #[arg(long = "cc", default_value = "cubic",
+        value_parser = clap::builder::PossibleValuesParser::new(CongestionControlAlgorithm::VARIANTS)
+            .map(|s| s.parse::<CongestionControlAlgorithm>().unwrap()))]
```
martinthomson, line 128: Is there a standard formula for teaching clap the variants, so that it can present useful help?  Ideally, the help text lists the options...

larseggert: The current implementation already shows the possible values in the help text :-)
```
  --cc <CONGESTION_CONTROL>
      The congestion controller to use

      [default: cubic]
      [possible values: newreno, cubic]
```

PR #3278, opened 2025-12-16
feat(transport/cc): make slow start pluggable and add hystart++

omansfeld:
[fdc7597](https://github.com/mozilla/neqo/pull/3278/commits/fdc7597df7d5c45ba5e2a4b74dc7b056a42b6d35) introduces the scaffolding to make slow start configurable by introducing a `SlowStart` trait and using a generic parameter with that trait bound in the `ClassicCongestionControl` struct. It also rewrites the current slow start implementation in `classic_slow_start.rs`.

[df81bdb](https://github.com/mozilla/neqo/pull/3278/commits/df81bdb76966a5e4fcbaa6ec61c639ed608ca0d4) makes the slow start algorithm configurable at run-time via the CLI/ConnectionParams, similar to how it works for Cubic/NewReno.

As a reminder: [fdc7597..df81bdb](https://github.com/mozilla/neqo/pull/3278/changes/BASE..df81bdb76966a5e4fcbaa6ec61c639ed608ca0d4) had already had a review cycle before I started working on HyStart++ and made the PR a draft again.

[41e8c04](https://github.com/mozilla/neqo/pull/3278/commits/41e8c04577bb69aa7947e0d00fbde6159ca5becb) implements HyStart++ as per [RFC 9406](https://datatracker.ietf.org/doc/html/rfc9406) with extensive tests added in [86b496d](https://github.com/mozilla/neqo/pull/3278/commits/86b496d41fc2256c9cdda01ee4137ca413d32e28).

With the latest commit [ae92b97](https://github.com/mozilla/neqo/pull/3278/commits/ae92b975b27770deedb4ea87a2683eb896b8b95f) I re-architected the `SlowStart` trait and the order in which slow start functionality is called from the congestion controller. Hystart functionality is unchanged between that commit and the former implementation, this just improves architecture and makes it more extensible for future work. 

It now separates functionality into the correct order of operations for slow start (`check for exit` -> `do the exit` -> `apply growth if not exiting` -> `do congestion avoidance if not in slow start anymore`) which also makes it easier to add things like [SUSS](https://dl.acm.org/doi/10.1145/3651890.3672234) or [rapid-start](https://datatracker.ietf.org/doc/html/draft-kazuho-ccwg-rapid-start-00) that accelerate slow start growth (but not touch the exit heuristics) in the future.

To maybe make review easier, the main parts are:

- `neqo-transport/src/cc/hystart.rs` contains the HyStart++ implementation.
- `neqo-transport/src/cc/tests/hystart.rs` contains tests for the HyStart++ implementation.
- `neqo-transport/src/cc/classic_slow_start.rs` has the standard slow start implementation that was previously hardcoded into `classic_cc.rs`.
- `neqo-transport/src/cc/classic_cc.rs` has the core `on_packets_acked` flow that calls everything else and the trait definition for `SlowStart`. There is also a small change to `on_packet_sent` for round-tracking.
- ... and then there is a lot of changing function calls in tests, scaffolding, etc.

I think it might make it the easiest to look at `classic_cc.rs` first and then follow the code flow from there into the HyStart++ implementation. I'm not sure it makes a lot of sense to look at commits one-by-one because of the re-architecture in the latest commit.

I also have some qlogs from validating the implementation against `h3.speed.cloudflare.com` where one can see HyStart++ working as intended. See this qvis screenshot where it exits at the same window size that is later converged on during congestion avoidance:

hystart_textbook-exit

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -126,7 +126,7 @@ pub trait SlowStart: Display + Debug {
 
     /// Calculates the congestion window increase in bytes during slow start. The default
     /// implementation returns `new_acked`, i.e. classic exponential slow start growth.
-    fn calc_cwnd_increase(&mut self, new_acked: usize, _max_datagram_size: usize) -> usize {
+    fn calc_cwnd_increase(&self, new_acked: usize, _max_datagram_size: usize) -> usize {
         new_acked
     }
 
--- a/neqo-transport/src/cc/hystart.rs
+++ b/neqo-transport/src/cc/hystart.rs
@@ -290,7 +290,7 @@ impl SlowStart for HyStart {
         Some(curr_cwnd)
     }
 
-    fn calc_cwnd_increase(&mut self, new_acked: usize, max_datagram_size: usize) -> usize {
+    fn calc_cwnd_increase(&self, new_acked: usize, max_datagram_size: usize) -> usize {
         // > For each arriving ACK in slow start, where N is the number of previously unacknowledged
         // > bytes acknowledged in the arriving ACK:
         // >


neqo-transport/src/cc/tests/cubic.rs
omansfeld, line 1: just lots of function call changes here

neqo-transport/src/cc/tests/new_reno.rs
omansfeld, line 1: same here

omansfeld:
> In favor of the design itself.
> 
> I suggest only merging here once we have more than one slow start algorithm. Each pull request should improve the state of `main`. Merging here now only adds a dead abstraction, i.e. `SlowStart`, only implemented once. (Similar to `trait CongestionControl` today.)

I very much get that point. That said, what would you say is the best point to move forward with this work then?

Having a bunch of dependent PRs seems like it could easily become a rebasing nightmare and one big PR seems like it could become very cumbersome to review. I would also very much like to have the along-the-way feedback of incremental PRs.

Should this be reviewed and merged onto a feature-branch that then eventually gets merged onto main?

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -44,8 +44,15 @@ const fn cwnd_after_loss_slow_start(cwnd: usize, mtu: usize) -> usize {
 /// If false, sets `w_max` lower than cwnd to prevent fast convergence.
 fn setup_congestion_avoidance(
     fast_convergence: bool,
-) -> (ClassicCongestionControl<Cubic>, CongestionControlStats) {
-    let mut cc = ClassicCongestionControl::new(Cubic::default(), Pmtud::new(IP_ADDR, MTU));
+) -> (
+    ClassicCongestionControl<ClassicSlowStart, Cubic>,
```
larseggert:
```suggestion
    cc:Classic<ss:Classic, Cubic>,
```
The module is called `cc`, so the `ClassicCongestionControl` -> `cc::Classic` rename should be straightforward. In order to be able to use `ss`, we need a module. (Also wonder if `cc` and `ss` are to terse and we should find better module names.)

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -983,13 +1024,21 @@ mod tests {
     fn persistent_congestion_no_lost() {
         let lost = make_lost(&[]);
         assert!(!persistent_congestion_by_pto(
-            ClassicCongestionControl::new(NewReno::default(), Pmtud::new(IP_ADDR, MTU)),
+            ClassicCongestionControl::new(
+                ClassicSlowStart::default(),
+                NewReno::default(),
+                Pmtud::new(IP_ADDR, MTU)
+            ),
```
larseggert: Make a helper. This is repeated a lot.

neqo-transport/src/cc/classic_slow_start.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// <https://datatracker.ietf.org/doc/html/rfc9002#section-7.3.1-2>
+#[derive(Debug, Default)]
+pub struct ClassicSlowStart {}
+
+impl Display for ClassicSlowStart {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ClassicSlowStart")
+    }
+}
```
larseggert, line 25: Is `Debug` not enough?

neqo-transport/src/cc/classic_slow_start.rs
```diff
@@ -0,0 +1,50 @@
...
+        let cwnd_increase = min(ssthresh - curr_cwnd, acked_bytes + new_acked);
+        let unused_acked_bytes = acked_bytes + new_acked - cwnd_increase;
+
+        if curr_cwnd + cwnd_increase == ssthresh {
+            // This doesn't look like it is necessary, but it can happen
+            // after persistent congestion.
+            exit_slow_start = true;
+        }
+
+        (cwnd_increase, unused_acked_bytes, exit_slow_start)
```
larseggert:
```suggestion
        let cwnd_increase = min(ssthresh - curr_cwnd, acked_bytes + new_acked);
        let unused_acked_bytes = acked_bytes + new_acked - cwnd_increase;

        // This doesn't look like it is necessary, but it can happen
        // after persistent congestion.
        let exit_slow_start = curr_cwnd + cwnd_increase == ssthresh;

        (cwnd_increase, unused_acked_bytes, exit_slow_start)
```

neqo-transport/src/cc/classic_slow_start.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+impl SlowStart for ClassicSlowStart {
+    fn on_packets_acked(
+        &mut self,
+        curr_cwnd: usize,
+        ssthresh: usize,
+        acked_bytes: usize,
+        new_acked: usize,
+    ) -> (usize, usize, bool) {
```
larseggert: Does this return triple warrant a type?

omansfeld:
> I suggest only merging here once we have more than one slow start algorithm. Each pull request should improve the state of `main`. Merging here now only adds a dead abstraction, i.e. `SlowStart`, only implemented once. (Similar to `trait CongestionControl` today.)

As per @mxinden's comment I made this PR a draft again and will also add HyStart++ as part of it. I might still ask for review on single commits along the way to have some incremental feedback.

neqo-transport/src/cc/classic_slow_start.rs
omansfeld: It's used in the `impl Display` for `ClassicCongestionControl` to display which algorithms are in use:

```rust
impl<S, T> Display for ClassicCongestionControl<S, T>
where
    S: SlowStart,
    T: WindowAdjustment,
{
        write!(
            f,
            "{}/{} CongCtrl: bif {} cwnd {} ssthresh {}",
            self.ss_algorithm,
            self.cc_algorithm,
            self.bytes_in_flight,
            self.congestion_window,
            self.ssthresh,
        )?;
        Ok(())
    }
}
```

neqo-transport/src/cc/tests/cubic.rs
omansfeld: I don't think `ss` should be it's own module. For me it's the same category as `Cubic` or `NewReno`. I think this just looks very weird because `ClassicSlowStart` is not the best name (because it includes "SlowStart"). 

If it was `ClassicCongestionControl<HyStart, Cubic>` this would look much more sensible imo. I'm not sure what a better name for the standard slow start is though. I'm reluctant towards just naming it `SlowStart` or `Classic` because that might get confusing.

As for `ClassicCongestionControl` I think we should refactor the structure there a bit anyways because right now we don't have anything else then `Classic` and are not planning to implement BBR afaik. So it seems a bit unnecessary to have the `CongestionControl` trait that is only implemented by `ClassicCongestionControl`. I think it could just be `CongestionControl` without an extra trait. But that is a seperate PR and I'd like to focus on getting the functionality in before doing that.

But yes, ideally this would become something like `CongestionController<HyStart, Cubic>` (or any other combination).

omansfeld:
I added another commit to make slow start configurable via connection parameters and the CLI. Would welcome another review once you have some time @mxinden!

Also a question, just to make sure I understood the parameter code structure correctly:

`ConnectionParameters` is what actually sets the cc parameters for the sender and is also what we would set via prefs from Firefox for e.g. experimentation in [neqo_glue](https://searchfox.org/firefox-main/rev/14c08f0368ead8bfdddec62f43e0bb5c8fd61289/netwerk/socket/neqo_glue/src/lib.rs#432-444), correct?

https://github.com/mozilla/neqo/blob/adcd4bddddce0fb83bbf3f6c01c3c740f6edc680/neqo-transport/src/sender.rs#L36-L44

But `QuicParameters` is used to set CLI flags in which case it passes them on to `ConnectionParams` when a client/server is created (of course it also contains more than just connection parameters).

https://github.com/mozilla/neqo/blob/adcd4bddddce0fb83bbf3f6c01c3c740f6edc680/neqo-bin/src/lib.rs#L234-L243

But from Firefox we [directly set our `ConnectionParameters`](https://searchfox.org/firefox-main/rev/14c08f0368ead8bfdddec62f43e0bb5c8fd61289/netwerk/socket/neqo_glue/src/lib.rs#432,457,468), so any defaults we have in Neqo basically don't matter if the consumer is Firefox, correct?

mxinden:
> Also a question, just to make sure I understood the parameter code structure correctly:
> 
> `ConnectionParameters` is what actually sets the cc parameters for the sender and is also what we would set via prefs from Firefox for e.g. experimentation in [neqo_glue](https://searchfox.org/firefox-main/rev/14c08f0368ead8bfdddec62f43e0bb5c8fd61289/netwerk/socket/neqo_glue/src/lib.rs#432-444), correct?

correct
 
> But `QuicParameters` is used to set CLI flags in which case it passes them on to `ConnectionParams` when a client/server is created (of course it also contains more than just connection parameters).

correct
 
> But from Firefox we [directly set our `ConnectionParameters`](https://searchfox.org/firefox-main/rev/14c08f0368ead8bfdddec62f43e0bb5c8fd61289/netwerk/socket/neqo_glue/src/lib.rs#432,457,468), so any defaults we have in Neqo basically don't matter if the consumer is Firefox, correct?

`neqo_glue` does not use `QuicParameters`. `neqo_glue` does use [`ConnectionParameters::default`](https://github.com/mozilla-firefox/firefox/blob/20a1fb35a4d5c2f2ea6c865ecebc8e4bee6f86c9/netwerk/socket/neqo_glue/src/lib.rs#L432).

mxinden (commented):
Changes here thus far look good to me.

omansfeld:
The HyStart++ implementation is done now, so I'm opening up for review! I updated the main PR description to resemble the current state.

neqo-bin/src/lib.rs
```diff
@@ -126,8 +126,14 @@ pub struct QuicParameters {
     #[arg(long = "cc", default_value = "cubic",
         value_parser = clap::builder::PossibleValuesParser::new(CongestionControlAlgorithm::VARIANTS)
             .map(|s| s.parse::<CongestionControlAlgorithm>().unwrap()))]
-    /// The congestion controller to use.
-    pub congestion_control: CongestionControlAlgorithm,
+    /// The congestion control algorithm to use.
+    pub cc_algorithm: CongestionControlAlgorithm,
+
```
larseggert, line 131: This seems like a gratuitous change? But I guess the intent is to align with `ss_algorithm`.

The `_algorithm` bit is longest, but has the least amount of useful information. How about `cong_ctrl_alg` and `slow_start_alg` or something like that?

neqo-bin/src/lib.rs
```diff
@@ -160,7 +166,8 @@ impl Default for QuicParameters {
             idle_timeout: 30,
             initial_rtt_ms: u64::try_from(DEFAULT_INITIAL_RTT.as_millis())
                 .expect("this value will always be less than u64::MAX"),
-            congestion_control: CongestionControlAlgorithm::Cubic,
+            cc_algorithm: CongestionControlAlgorithm::Cubic,
+            ss_algorithm: SlowStartAlgorithm::Classic,
```
larseggert: Is the `Algorithm` part of the name really needed? Seems to just waste characters.

neqo-transport/src/cc/tests/hystart.rs
```diff
@@ -0,0 +1,600 @@
...
+
+use super::make_cc_hystart;
+use crate::{
+    cc::{CongestionControl as _, classic_cc::SlowStart as _, hystart::HyStart},
+    recovery::sent,
+    rtt::RttEstimate,
+    stats::CongestionControlStats,
+};
+
+const SMSS: usize = 1200;
```
larseggert: Can we not reuse `MIN_INITIAL_PACKET_SIZE`?

neqo-transport/src/cc/tests/hystart.rs
```diff
@@ -0,0 +1,600 @@
...
+        "New round should start with new window_end"
+    );
+}
+
+/// Tests that `current_round_min_rtt` is tracked correctly when packets are acked.
+#[test]
+fn rtt_sample_collection_tracks_minimum() {
+    const BASE_RTT: Duration = Duration::from_millis(100);
+    const HIGH_RTT: Duration = Duration::from_millis(120);
+    const LOW_RTT: Duration = Duration::from_millis(80);
```
larseggert: In other tests, you use literals. Suggest to use these `const`s more widely for consistency.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,37 @@ pub trait WindowAdjustment: Display + Debug {
...
+/// slow start threshold (`ssthresh`) is set on exit and they can influence how fast the exponential
+/// congestion window growth rate during slow start is.
+pub trait SlowStart: Display + Debug {
+    /// Handle tracking RTT rounds through having the next packet number that is to be sent out
+    /// passed in.
+    fn on_packet_sent(&mut self, sent_pn: packet::Number);
+    /// Handle packets being acknowledged during initial slow start and do computations to decide if
+    /// initial slow start should be exited to congestion avoidance.
+    ///
+    /// Returns `exit_to_ca`.
```
larseggert: What is `exit_to_ca`?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+    cmp::{max, min},
+    fmt::Display,
+    time::Duration,
+};
+
+use neqo_common::{qdebug, qinfo};
+
+use crate::{cc::classic_cc::SlowStart, packet, rtt::RttEstimate};
+
+#[derive(Debug, Default, Clone, Copy)]
```
larseggert: `Copy` seems costly. Do you even need it? And do you need `Default`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -291,22 +339,59 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
 
-        // Slow start, up to the slow start threshold.
+        // Initial slow start exit heuristics.
+        if self.in_initial_slow_start() {
+            let exit_to_ca = self
+                .ss_algorithm
+                .on_packets_acked(rtt_est, largest_packet_acked.pn());
+
+            if exit_to_ca {
+                qinfo!("Exited slow start by algorithm");
```
```suggestion
                qdebug!!("Exited slow start by algorithm");
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -291,22 +339,59 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+                    .maybe_change_cwnd_increase(cwnd_increase, self.max_datagram_size());
+            } else {
+                cwnd_increase = min(
+                    cwnd_increase,
+                    self.current.ssthresh - self.current.congestion_window,
+                );
+            }
+
+            self.current.congestion_window += cwnd_increase;
+            qdebug!("[{self}] slow start += {cwnd_increase}");
```
```suggestion
            qtrace!("[{self}] slow start += {cwnd_increase}");
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -291,22 +339,59 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+
+            // This can only happen after persistent congestion when we re-enter slow start while
+            // having a previously established `ssthresh` which has now been reached.
             if self.current.congestion_window == self.current.ssthresh {
-                // This doesn't look like it is necessary, but it can happen
-                // after persistent congestion.
+                qinfo!(
```
```suggestion
                qdebug!(
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -916,11 +1004,11 @@ mod tests {
     }
 
     fn persistent_congestion(lost_packets: &[sent::Packet], persistent_expected: bool) {
-        let cc = congestion_control(CongestionControlAlgorithm::NewReno);
+        let cc: Box<dyn CongestionControl> = Box::new(make_cc_newreno());
```
larseggert: I don't think this needs a `Box` if you make `persistent_congestion_by_algorithm` take an `impl CongestionControl`.

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+            current_round_min_rtt: Duration::MAX,
+            rtt_sample_count: 0,
+            window_end: None,
+            css_baseline_min_rtt: Duration::MAX,
+            css_round_count: 0,
+        }
+    }
+}
+
+#[derive(Debug, Default)]
```
larseggert: Do you even need `Default`?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+        // > cssBaselineMinRtt = infinity
+        // > resume slow start including HyStart++
+        // > ```
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9406#section-4.2-20>
+        if self.in_css()
+            && self.enough_samples()
+            && self.current.current_round_min_rtt < self.current.css_baseline_min_rtt
+        {
+            qinfo!(
```
larseggert:
```suggestion
            qdebug!(
```

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+            let rtt_thresh = max(
+                Self::MIN_RTT_THRESH,
+                min(
+                    self.current.last_round_min_rtt / Self::MIN_RTT_DIVISOR,
+                    Self::MAX_RTT_THRESH,
+                ),
+            );
+            if self.current.current_round_min_rtt >= self.current.last_round_min_rtt + rtt_thresh {
+                self.current.css_baseline_min_rtt = self.current.current_round_min_rtt;
+                qinfo!(
```
larseggert:
```suggestion
                qdebug!(
```

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+                exit_to_ca = self.current.css_round_count >= Self::CSS_ROUNDS;
+                qinfo!(
+                    "HyStart: on_packets_acked -> exit={exit_to_ca} because css_rounds={} >= {}",
+                    self.current.css_round_count,
+                    Self::CSS_ROUNDS
+                );
+            }
+        }
+
+        exit_to_ca
```
larseggert:
```suggestion
        false
```
And then `return self.current.css_round_count >= Self::CSS_ROUNDS;` in the conditional above. No need for a `mut exit_to_ca`.

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    fn maybe_change_cwnd_increase(
+        &mut self,
+        cwnd_increase: usize,
+        max_datagram_size: usize,
+    ) -> usize {
+        self.calc_cwnd_increase(cwnd_increase, max_datagram_size)
+    }
```
larseggert: Can we not just use `calc_cwnd_increase` directly?

neqo-transport/src/rtt.rs
```diff
@@ -177,7 +177,6 @@ impl RttEstimate {
         self.first_sample_time
     }
 
-    #[cfg(test)]
     pub const fn latest(&self) -> Duration {
```
larseggert: If this isn't test-only anymore, maybe rename to `latest_rtt()`?

neqo-bin/src/lib.rs
omansfeld: The intent was to align `ss_algorithm` with the `cc_algorithm` we already had on the `ClassicCongestionControl` struct and on `ConnectionParameters`. Then I renamed `congestion_control` here to align it everywhere. I agree that the naming could be better. 

I will check if it makes sense to just have `congestion_control` and `slow_start` in all places. Or if that doesn't work out go with your suggestion.

neqo-bin/src/lib.rs
omansfeld: I think we can shorten both `SlowStart` and `CongestionControl` to omit the `-Algorithm`.

neqo-transport/src/cc/classic_cc.rs
omansfeld: It is the return value to decide if we want to exit to congestion avoidance (`ca`). I renamed it from the `exit_slow_start` I had initially because I felt `exit_slow_start` was ambiguous as the HyStart RFC also uses "exiting slow start" when referencing going from slow start to conservative slow start.

I don't have a strong opinion here. If you think `exit_to_ca` is not clear enough I can fall back to `exit_slow_start` or another suggestion.

neqo-transport/src/cc/hystart.rs
omansfeld: I think I needed `Copy` at some (bad) previous state of the implementation, then forgot to take it out. Good catch, thanks! Also taking out `Default` here and for the one below as it is not needed.

This also prompted me to look further into what the different `derive` macros do, I've just been doing what other code does without thinking much about it tbh, so thanks for nudging me here.

neqo-transport/src/cc/hystart.rs
omansfeld: No, see above.

neqo-transport/src/cc/hystart.rs
omansfeld: Nice way to get rid of the `let mut`, thanks!

neqo-transport/src/cc/classic_cc.rs
larseggert: I think `exit_slow_start` is nicer (also because maybe some future SS algorithm would not enter CA on exit.)

neqo-transport/src/cc/hystart.rs
omansfeld: Other way around, I will get rid of `calc_cwnd_increase` and put the code directly into the `maybe_change_cwnd_increase` trait function implementation instead. This is the only consumer of `calc_cwnd_increase`.

neqo-transport/src/cc/tests/hystart.rs
omansfeld: Forgot that exists, thanks.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -146,8 +177,9 @@ impl State {
 }
 
 #[derive(Debug)]
-pub struct ClassicCongestionControl<T> {
-    cc_algorithm: T,
+pub struct ClassicCongestionControl<S, T> {
```
```suggestion
pub struct ClassicCongestionControl<T, S = ClassicSlowStart> {
```

Would a default help here?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -177,23 +209,32 @@ pub struct ClassicCongestionControl<T> {
...
     }
+
+    /// Return if the Congestion Controller currently is in initial slow start.
+    pub const fn in_initial_slow_start(&self) -> bool {
+        self.current.ssthresh == usize::MAX
+    }
 }
 
-impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
+impl<S, T> CongestionController for ClassicCongestionControl<S, T>
```
mxinden: Previously both the trait and the struct ended in `Control`. Now the trait ends in `Controller` and the struct ends in `Control`. Is that mismatch necessary?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,37 @@ pub trait WindowAdjustment: Display + Debug {
...
 }
 
+/// Trait for slow start exit algorithms.
+///
+/// Implementations define when and if to exit from slow start, how the slow start threshold
+/// (`ssthresh`) is set on exit and they can influence how fast the exponential congestion window
+/// growth rate during slow start is.
+pub trait SlowStart: Display + Debug {
+    /// Handle tracking RTT rounds through having the next packet number that is to be sent out
+    /// passed in.
```
mxinden: Not sure I understand. Would the below be correct, too?


```suggestion
    /// Enables a trait implementor to track RTT rounds via the next packet number that is to be sent out.
```

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,37 @@ pub trait WindowAdjustment: Display + Debug {
...
+    fn on_packet_sent(&mut self, sent_pn: packet::Number);
+    /// Handle packets being acknowledged during initial slow start and do computations to decide if
+    /// initial slow start should be exited.
```
```suggestion
    /// Handle packets being acknowledged during initial slow start. Returns whether
    /// initial slow start should be exited.
```

Feel free to ignore.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,37 @@ pub trait WindowAdjustment: Display + Debug {
...
+    ///
+    /// Returns `exit_slow_start`.
+    fn on_packets_acked(&mut self, rtt_est: &RttEstimate, largest_acked: packet::Number) -> bool;
+
+    /// Can apply changes to the exponential congestion window increase during initial slow start.
+    /// One example is the reduced growth in Conservative Slow Start (CSS) from HyStart++.
+    ///
+    /// Returns `cwnd_increase`.
+    fn maybe_change_cwnd_increase(
```
mxinden: Nit pick: `maybe_` implies returning an `Option`, yet an implementor that doesn't alter the cwnd (classic slow start) just returns `cwnd_increase`. Is that intuitive?

If you return an `Option` instead, you could add a default implementation here:

``` rust
fn maybe_change_cwnd_increase(
        &mut self,
        cwnd_increase: usize,
        max_datagram_size: usize,
    ) -> Option<usize> {
    None
}
```

and then not implement it in classic slow start at all.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -469,6 +554,11 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
     }
 
     fn on_packet_sent(&mut self, pkt: &sent::Packet, now: Instant) {
+        // Pass next packet number to send into slow start algorithm for initial slow start phase.
+        if self.in_initial_slow_start() {
+            self.slow_start.on_packet_sent(pkt.pn());
+        }
```
mxinden: Software design wise I suggest encapsulating all slow start logic in `self.slow_start`. (See also John Ousterhout "modules should be deep".)

Instead of the `if self.in_initial_slow_start()` here and in other places, couldn't `self.slow_start` do the check? We could tell `self.slow_start` at construction time, whether it is an initial slow start or a non-initial, right?

neqo-transport/src/cc/classic_slow_start.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ClassicSlowStart")
+    }
+}
+
+impl SlowStart for ClassicSlowStart {
+    fn on_packet_sent(&mut self, _sent_pn: packet::Number) {}
+
+    fn on_packets_acked(&mut self, _rtt_est: &RttEstimate, _largest_acked: packet::Number) -> bool {
+        // Standard slow start does not have any heuristic for exiting initial slow start. Always
```
mxinden:
```suggestion
        // Classic slow start does not have any heuristic for exiting initial slow start. Always
```

Either _Standard_ or _Classic_ in `struct` name and comments, but not both. Also below.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -291,27 +339,64 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
...
+        if self.in_initial_slow_start() {
+            let exit_slow_start = self
+                .slow_start
+                .on_packets_acked(rtt_est, largest_packet_acked.pn());
+
+            if exit_slow_start {
+                qdebug!("Exited slow start by algorithm");
+                let exit_cwnd = self
+                    .slow_start
+                    .on_slow_start_exit(self.current.congestion_window);
```
mxinden: Is `on_slow_start_exit` needed? Above `self.slow_start_on_packets_acked` decides whether to exit slow start. Based on that decision, it is called back directly, providing it with a value that hasn't changed since (i.e. `self.current.congestion_window`). Why not provide `self.current.congestion_window` with `self.slow_start.on_packets_acked` right away?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+    fmt::Display,
+    time::Duration,
+};
+
+use neqo_common::{qdebug, qinfo, qtrace};
+
+use crate::{cc::classic_cc::SlowStart, packet, rtt::RttEstimate};
+
+#[derive(Debug, Clone)]
+pub struct State {
```
mxinden: Why is `State` `pub`? For testing? If so, can you document it please?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// >
+    /// > ```pseudo
+    /// > lastRoundMinRTT = currentRoundMinRTT
+    /// > currentRoundMinRTT = infinity
+    /// > rttSampleCount = 0
+    /// > ```
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9406#section-4.2-4>
+    ///
+    /// Neqo sets `window_end` to `None` when it is acked and calls this function when packets are
```
mxinden:
```suggestion
    /// [`<HyStart as SlowStart>::on_packets_acked`] sets `window_end` to `None` when it is acked and calls this function when packets are
```

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        // > For rounds where at least N_RTT_SAMPLE RTT samples have been obtained and
+        // > currentRoundMinRTT and lastRoundMinRTT are valid, check to see if delay increase
+        // > triggers slow start exit.
+        //
+        // <https://datatracker.ietf.org/doc/html/rfc9406#section-4.2-13>
+        if !self.in_css()
+            && self.enough_samples()
+            && self.current.current_round_min_rtt != Duration::MAX
+            && self.current.last_round_min_rtt != Duration::MAX
```
mxinden: `Duration::Max` here represents the no-value value, overloading the meaning of a `usize`. I suggest leveraging Rust's `Option` here. In other words, I suggest making `current_round_min_rtt` and `last_round_min_rtt` an `Option`, then check for `is_none`.

Feel free to ignore in case it significantly complicates the implementation.

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+                "HyStart: on_packets_acked -> round ended because largest_acked={largest_acked} >= window_end={window_end}"
+            );
+            self.current.window_end = None;
+
+            // If a round ends while in CSS increase the counter and do a check if enough rounds
+            // to exit to congestion avoidance have been completed.
+            if self.in_css() {
+                self.current.css_round_count += 1;
+                let exit_slow_start = self.current.css_round_count >= Self::CSS_ROUNDS;
+                qinfo!(
```
mxinden: Isn't `info` a bit high?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+                    "HyStart: on_packets_acked -> exit={exit_slow_start} because css_rounds={} >= {}",
+                    self.current.css_round_count,
+                    Self::CSS_ROUNDS
+                );
+                return exit_slow_start;
+            }
+        }
+
+        false
```
mxinden: How about using early returns above, thus reducing indentation?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    pub const MIN_RTT_DIVISOR: u32 = 8;
+
+    pub const N_RTT_SAMPLE: usize = 8;
+
+    pub const CSS_GROWTH_DIVISOR: usize = 4;
+
+    pub const CSS_ROUNDS: usize = 5;
+
+    pub const NON_PACED_L: usize = 8;
```
mxinden, line 57: Do any other implementations deviate from the suggested constants in the RFC?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+            css_baseline_min_rtt: Duration::MAX,
+            css_round_count: 0,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct HyStart {
+    limit: usize,
+    current: State,
```
mxinden: I don't understand why the `State` struct is needed and why it is called `current`? Why is `limit` in `HyStart` while all other variables are in `State`? Next to the `current: State` I can not find a `next: State` or a `previous: State`.

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+            window_end: None,
+            css_baseline_min_rtt: Duration::MAX,
+            css_round_count: 0,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct HyStart {
+    limit: usize,
```
mxinden, line 29: Worth documenting that this is the cwnd increase limit factor with the MSS?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,301 @@
...
+            cwnd_increase /= Self::CSS_GROWTH_DIVISOR;
+        }
+        cwnd_increase
+    }
+
+    // > If CSS_ROUNDS rounds are complete, enter congestion avoidance by setting the ssthresh to
+    // > the current cwnd.
+    //
+    // <https://datatracker.ietf.org/doc/html/rfc9406#section-4.2-23>
+    fn on_slow_start_exit(&mut self, curr_cwnd: usize) -> usize {
```
mxinden: As mentioned elsewhere, is this function needed?

neqo-transport/src/cc/classic_cc.rs
omansfeld: We set the default in `ConnectionParameters::default()`. So I don't think this is necessary, as we also don't set a default for `CongestionControl::Cubic` here, but in `ConnectionParameters::default()` instead.

https://github.com/mozilla/neqo/blob/3c7e8d40faa9556baf28d305fcd1c8f1e7475a84/neqo-transport/src/connection/params.rs#L155-L187

neqo-transport/src/cc/classic_cc.rs
omansfeld: Correct. I think `Controller` makes more sense, as that struct is what holds both slow start and the congestion control algorithm. I will rename `ClassicCongestionControl` accordingly, so the naming follows the trait it implements.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Yes, your wording is more concise.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Returning an `Option` wasn't occurring to me intuitively, but if it is the rust idiomatic way and I just didn't know about it then I see the point. It is intuitive to me that `maybe_change...` returns the unchanged value if it doesn't change it.

Doing the `Option` thing would mean we'd have to handle that on the caller side, which imo would make that code harder to read.

The default implementation is a good idea, I'd prefer for it to just return the passed in `cwnd_increase`, though. I will specify the return values and default implementation in the doc comment so it is more easily understandable.

neqo-transport/src/cc/classic_slow_start.rs
omansfeld: Do you have a preference? To be honest I don't really like either name, but I also don't have a good alternative in mind. I guess historically this is just where we're at and there is no "official" name for it. Could also call it `TCPSlowStart` which I really don't like ....

mxinden: Preference for `ClassicSlowStart`. Most intuitive to me.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Very interesting read, thanks for the reference.

I'm not sure I 100% understand your suggestion though, or if it makes sense to me. Maybe lets chat tomorrow in the office, I think that's easiest.

neqo-transport/src/cc/classic_cc.rs
omansfeld: The abstraction is because other algorithms might not just return the current congestion window. I.e. SEARCH has this:

>  When exiting slow start at the chokepoint, SEARCH is delayed in its prediction by almost exactly two RTTs. SEARCH can compute exactly how many extra bytes have been added to the congestion window (cwnd) beyond this point. The cwnd is reduced by this amount, before setting ssthresh to the cwnd (line 55). This exits slow start.

https://datatracker.ietf.org/doc/html/draft-chung-ccwg-search-07#section-3.2-20

I guess it could fundamentally be a part of `on_packets_acked` in either case, but imo the code in `classic_cc::on_packets_acked` is more readable by having those two seperate blocks of A: deciding if to exit slow start, and B: exiting slow start by setting the `ssthresh`.

neqo-transport/src/cc/hystart.rs
omansfeld: I think I did that because I had just worked on spurious event recovery where this pattern was used for cc parameters, but it isn't needed here. Thanks for pointing it out.

neqo-transport/src/cc/hystart.rs
omansfeld: msquic is using the default

https://github.com/microsoft/msquic/blob/56fe046c1038ddbeb760ecffbddc4fe7d4547627/src/core/quicdef.h#L571-L592

so is quiche

https://github.com/cloudflare/quiche/blob/12a5de0701ec98f174b6c84c916dbe117b7e4c50/quiche/src/recovery/congestion/hystart.rs#L39-L48

on a first glance linux, too, but I'll have to look deeper to check for the other values that are not explicitly listed as a constant in the linux implementation

https://github.com/torvalds/linux/blob/6fab32bb6508abbb8b7b1c5498e44f0c32320ed5/net/ipv4/tcp_cubic.c#L43-L47

neqo-transport/src/cc/hystart.rs
omansfeld: I got rid of `State` so this solved itself.

neqo-transport/src/cc/classic_cc.rs
mxinden: I would prefer not introducing an abstraction that is only needed in a follow-up pull request (e.g. here the SEARCH pull request).

> I guess it could fundamentally be a part of `on_packets_acked` in either case, but imo the code in `classic_cc::on_packets_acked` is more readable by having those two seperate blocks of A: deciding if to exit slow start, and B: exiting slow start by setting the `ssthresh`.

I prefer a single function, preventing misuse, e.g. calling `on_slow_start_exit` without calling `self.slow_start_on_packets_acked`. There more restrictive the interface, the less misuse it allows.

neqo-transport/src/cc/classic_cc.rs
omansfeld: After in person discussions Max and I decided to not restrict HyStart++ to initial slow start only for the following reasons:

- simplifies the code in `classic_cc::on_packets_acked`
- RFC is only stating a SHOULD, so we are not in direct conflict
- after persistent congestion the previously established `ssthresh` might still be too high, so there might be loss before it is reached, thus HyStart++ still makes sense

omansfeld:
I addressed @mxinden's review now. Also found a bug in the main HyStart++ logic that is addressed in https://github.com/mozilla/neqo/pull/3278/commits/1c5a29879284ec5d1245d04522840879a5eac6ed (see the commit message for details) and now in line with the RFC and other implementations.

neqo-transport/src/cc/tests/mod.rs
```diff
@@ -9,18 +9,45 @@ use std::{
...
+pub fn make_cc_cubic() -> ClassicCongestionController<ClassicSlowStart, Cubic> {
+    ClassicCongestionController::new(
+        ClassicSlowStart::default(),
+        Cubic::default(),
+        Pmtud::new(IP_ADDR, MTU),
+    )
+}
+
+/// Helper to create `ClassicCongestionControl` with HyStart++ for tests.
+pub fn make_cc_hystart(paced: bool) -> ClassicCongestionController<HyStart, NewReno> {
```
mxinden: Why test HyStart with NewReno? We don't deploy NewReno in Firefox.

omansfeld: I used NewReno because it is supposedly easier to assert what it does than with Cubic just in case I need to assert what Congestion Avoidance does to confirm i.e. having left Slow Start. In the end it didn't make a difference with how the tests have turned out. I see your point of putting Cubic just to future-proof and make sure the function isn't abused later.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,31 @@ pub trait WindowAdjustment: Display + Debug {
...
+        largest_acked: packet::Number,
+        curr_cwnd: usize,
+    ) -> Option<usize>;
+
+    /// Calculates the congestion window increase in bytes during slow start. The default
+    /// implementation returns `new_acked`, i.e. classic exponential slow start growth.
+    fn calc_cwnd_increase(&mut self, new_acked: usize, _max_datagram_size: usize) -> usize {
+        new_acked
+    }
+}
```
mxinden, line 136: Given that we use a concrete `SlowStart` implementation instance both for the initial slow start as well as the slow start when re-entered through persistent congestion, should there be a `SlowStart::reset` function which is called before re-entering slow start? Otherwise e.g. `HyStart` would continue off of its state from the initial slow start.

neqo-transport/src/sender.rs
```diff
@@ -34,13 +37,31 @@ impl PacketSender {
...
         let mtu = pmtud.plpmtu();
         Self {
-            cc: match conn_params.get_cc_algorithm() {
-                CongestionControlAlgorithm::NewReno => {
-                    Box::new(ClassicCongestionControl::new(NewReno::default(), pmtud))
-                }
-                CongestionControlAlgorithm::Cubic => {
-                    Box::new(ClassicCongestionControl::new(Cubic::default(), pmtud))
-                }
+            cc: match conn_params.get_congestion_control() {
```
mxinden: You could do this in one `match`:

``` rust
match (conn_params.get_congestion_control(), conn_params.get_slow_start()) {
```

Not a strong opinion.

neqo-transport/src/cc/tests/hystart.rs
```diff
@@ -0,0 +1,540 @@
...
+    );
+}
+
+#[test]
+fn l_limit_unpaced_is_capped() {
+    let mut hystart = make_hystart_unpaced(); // L = NON_PACED_L
+    hystart.on_packet_sent(0);
+
+    // Try to increase by more than NON_PACED_L * SMSS
+    // Try to increase by more than NON_PACED_L * SMSS
```
mxinden:
```suggestion
    // Try to increase by more than NON_PACED_L * SMSS
```

mxinden:
> After in person discussions Max and I decided to not restrict HyStart++ to initial slow start only for the following reasons:
> 
>     * simplifies the code in `classic_cc::on_packets_acked`
> 
>     * RFC is only stating a SHOULD, so we are not in direct conflict
> 
>     * after persistent congestion the previously established `ssthresh` might still be too high, so there might be loss before it is reached, thus HyStart++ still makes sense

Can you document this decision in the code? Sorry in case it is already documented and I missed it.

neqo-transport/src/cc/tests/hystart.rs
omansfeld: whoops.

neqo-transport/src/cc/classic_cc.rs
omansfeld: That is a good catch. I will add it.

Then there is the question of spurious recovery, because if we reset the parameters on slow start exit (loss and algorithm based) then we won't have them if the exit was spurious. I think it would make sense not call the `reset` on slow start exit, but when we re-enter slow start after persistent congestion, so e.g. here if `persistent_congestion` is `true`:

https://github.com/mozilla/neqo/blob/586ab2255a13c3bfedf77fca07bdd934375b91cd/neqo-transport/src/cc/classic_cc.rs#L417-L426

neqo-transport/src/sender.rs
omansfeld: Thanks, I prefer that to what I had before actually. Makes it simpler to see what exactly we match on.

omansfeld:
Updated again in https://github.com/mozilla/neqo/pull/3278/commits/89a516977e45430a9bf77f972e1b33bc44f20ae0. Addressed nits from @mxinden, added comment about the non-initial slow start decision and added `SlowStart::reset()` to erase state after persistent congestion. Also added another test that asserts this.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -107,6 +107,35 @@ pub trait WindowAdjustment: Display + Debug {
...
+    fn on_packets_acked(
+        &mut self,
+        rtt_est: &RttEstimate,
+        largest_acked: packet::Number,
+        curr_cwnd: usize,
+    ) -> Option<usize>;
+
+    /// Calculates the congestion window increase in bytes during slow start. The default
+    /// implementation returns `new_acked`, i.e. classic exponential slow start growth.
+    fn calc_cwnd_increase(&mut self, new_acked: usize, _max_datagram_size: usize) -> usize {
```
```suggestion
    fn calc_cwnd_increase(&self, new_acked: usize, _max_datagram_size: usize) -> usize {
```

Hystart is the only implementor and it doesn't mutate `self`, right?

omansfeld:
@larseggert do you want to take another look or can I merge? The plan is to land this pref'd off in firefox with a new release next week and then run experiments on it (classic is still the default in neqo).

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -145,8 +174,9 @@ impl State {
 }
 
 #[derive(Debug)]
-pub struct ClassicCongestionControl<T> {
-    cc_algorithm: T,
+pub struct ClassicCongestionController<S, T> {
+    slow_start: S,
+    congestion_control: T,
```
larseggert, line 179: Any reason for the renaming? (It's longer and makes the diff more noisy.)

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -231,6 +265,10 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
         &mut self.pmtud
     }
 
+    #[expect(
+        clippy::too_many_lines,
+        reason = "The main congestion control function contains a lot of logic."
+    )]
```
larseggert, line 271: Can we break some of it out?

neqo-transport/src/cc/hystart.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// > A paced TCP implementation SHOULD use `L = infinity`. Burst concerns are mitigated by
+    /// > pacing, and this setting allows for optimal cwnd growth on modern networks.
+    ///
+    /// <https://datatracker.ietf.org/doc/html/rfc9406#section-4.3-9>
+    limit: usize,
+    last_round_min_rtt: Option<Duration>,
+    current_round_min_rtt: Option<Duration>,
+    rtt_sample_count: usize,
+    window_end: Option<packet::Number>,
+    css_baseline_min_rtt: Option<Duration>,
```
larseggert, line 34: Are any of these `Option`s algorithmically coupled? In other words, would be be clearer to have `Option<(X, Y, ...)>` tuples instead of separate standalone ones?

omansfeld: No, they can all be `None` or `Some` on their own and different things can be derived from them being `None`.

neqo-transport/src/cc/classic_cc.rs
omansfeld: I shortened `CongestionControlAlgorithm` to `CongestionControl` and thus had to change the `CongestionControl` trait to `CongestionController`. `ClassicCongestionController` then just followed that renaming because Max had pointed out the mismatch. I think semantically `Controller` also makes sense, because that is the struct that holds the different algorithms that are part of the controller, i.e. slow start and congestion control (cubic).

martinthomson: What's wrong with `CongestionAlgorithm` ?  Or just `Algorithm` given the context.

omansfeld: We were also changing `SlowStartAlgorithm` to `SlowStart` and I prefered having them aligned (i.e. not `CongestionAlgorithm` and `SlowStart` (without `Algorithm`).  This is the `enum`s in `cc/mod.rs`, so it cannot be just `Algorithm` as both the slow start and the congestion control `enum` co-exist in the same scope.

neqo-transport/src/cc/classic_cc.rs
omansfeld: I'm sure there is opportunity there to modularize/break out. I will open an issue for it as it touches more than slow start.

neqo-transport/src/cc/classic_cc.rs
larseggert: OK, let's land it as-is. We can always tweak things in follow-up PRs.

neqo-transport/src/cc/classic_cc.rs
omansfeld: for reference: https://github.com/mozilla/neqo/issues/3453?reload=1

PR #3279, opened 2025-12-16
chore: Split QNS workflows

larseggert:
Into one generating a Docker image for the interop runner, and one running on PRs. The old file had both, and they had zero overlap.


.github/workflows/qns-pr.yml
```diff
@@ -0,0 +1,345 @@
+name: QNS PR
```
mxinden, line 4: Mind adding a comment here and in the other file differentiating the two? Will be helpful in a couple of months once we forgot.

PR #3280, opened 2025-12-16
chore: Add tests for Gecko-specific functions

larseggert:


neqo-crypto/tests/agent.rs
```diff
@@ -134,6 +142,8 @@ fn raw() {
     // The client should have one certificate for the server.
     let certs = client.peer_certificate().unwrap();
     assert_eq!(1, certs.into_iter().count());
+    assert!(certs.stapled_ocsp_responses().unwrap().is_empty());
+    assert!(certs.signed_cert_timestamp().unwrap().is_empty());
```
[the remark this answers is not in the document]
martinthomson: The comment is good, though I think we'll need a bug on file for these.  Configuring an SCT is easy (it's a blob of bytes if you don't care about validating it, which we don't) but setting up the plumbing requires a bunch of poking at NSS internals.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1605,3 +1605,24 @@ fn grease_quic_bit_respects_current_handshake() {
...
+            Ok(())
+        }
+        fn encode(input: &[u8], output: &mut [u8]) -> neqo_crypto::Res<usize> {
+            output[..input.len()].copy_from_slice(input);
+            Ok(input.len())
+        }
+    }
+    let mut client = default_client();
+    client.set_certificate_compression::<Copy>().unwrap();
+}
```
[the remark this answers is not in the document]
martinthomson: This comment is good.  One thing you might consider doing is move the certificate to a global variable to compress it.  It's a bit more work, but you can defer that to a new issue.  It's a small fun programming exercise to do that safely (consider the possibility that the test is run twice concurrently and how you would make the global state safe to use).

martinthomson (approved):

neqo-crypto/tests/agent.rs
```diff
@@ -134,11 +142,37 @@ fn raw() {
...
 }
 
+#[test]
+fn ocsp_stapling_and_signed_cert_timestamps() {
+    use neqo_crypto::Opt;
+
+    fixture_init();
+    let mut client = Client::new("server.example", true).expect("should create client");
+    client
+        .set_option(Opt::SignedCertificateTimestamps, true)
```
martinthomson:
```suggestion
    fixture_init();
    let mut client = Client::new("server.example", true).expect("should create client");
    client
        .set_option(neqo_crypto::Opt::SignedCertificateTimestamps, true)
```

neqo-crypto/src/agent.rs
```diff
@@ -1264,22 +1274,59 @@ impl Server {
...
+            let ocsp_array = ssl::SECItemArrayStr {
+                items: ocsp_items.as_ptr().cast::<ssl::SECItem>().cast_mut(),
+                len: c_uint::try_from(ocsp_items.len())?,
             };
-            let key_ptr = unsafe { p11::PK11_FindKeyByAnyCert(*cert, null_mut()) };
-            let Ok(key) = PrivateKey::from_ptr(key_ptr) else {
-                return Err(Error::CertificateLoading);
+            let sct_item = p11::Item::wrap(scts)?;
+            let extra = ssl::SSLExtraServerCertDataStr {
+                authType: ssl::SSLAuthType::ssl_auth_null,
```
martinthomson, line 1317: I checked, and null is the "I don't care what sort of certificate this is" option.  It might be worth noting that.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1605,3 +1605,45 @@ fn grease_quic_bit_respects_current_handshake() {
...
 }
+
+#[test]
+fn certificate_compression() {
+    use std::sync::Mutex;
+
+    use neqo_crypto::agent::CertificateCompressor;
+
+    static ORIGINAL: Mutex<Vec<u8>> = Mutex::new(Vec::new());
+    static DECODED: Mutex<Vec<u8>> = Mutex::new(Vec::new());
```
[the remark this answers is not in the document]
martinthomson: This isn't quite what I had in mind, though it will do.  Copilot is correct in that running the test concurrently - in theory at least - is exposed to failure.  But the tests we produce don't do that because the certificate is effectively a fixed value.  I would document that assumption here though, in case someone comes back to try this again and encounters problems.

A more robust test would hash the certificate, using a strong hash, on decode and put (hash, certificate) in a global/thread-local map and the hash into the output.  Decoding would then do a map lookup.  It's a memory leak, but a finite one.

neqo-crypto/src/agent.rs
```diff
@@ -1264,22 +1274,59 @@ impl Server {
...
+            zero_rtt_check: None,
+        })
+    }
 
+    /// Create a server with OCSP responses and SCTs configured.
+    ///
+    /// # Errors
+    ///
+    /// Errors returned when NSS fails.
+    pub fn new_with_ocsp_and_scts<A: AsRef<str>>(
```
martinthomson, line 1298:
```suggestion
    #[cfg(test)]
    pub fn new_with_ocsp_and_scts<A: AsRef<str>>(
```

?

neqo-crypto/src/agent.rs
```diff
@@ -1264,22 +1274,59 @@ impl Server {
...
+            })?;
+        }
+        agent.ready(true, true)?;
+        Ok(Self {
+            agent,
+            zero_rtt_check: None,
+        })
+    }
 
+    /// Create a server with OCSP responses and SCTs configured.
```
martinthomson, line 1290:
```suggestion
    /// Create a server with OCSP responses and SCTs configured.
    /// Not suitable for multiple certificates, because it configures the same OCSP/SCT for all certificates.
    /// In other words, this is good for testing that the plumbing works, not for a real server.
```

neqo-crypto/src/auth.rs
```diff
@@ -50,11 +50,25 @@ impl From<PRErrorCode> for AuthenticationStatus {
...
+        assert_eq!(
+            AuthenticationStatus::from(12345),
+            AuthenticationStatus::Unknown
+        );
+        assert_eq!(
+            AuthenticationStatus::from(i32::MIN),
+            AuthenticationStatus::Unknown
+        );
+    }
+}
```
martinthomson:
```suggestion
}

```

neqo-crypto/src/agent.rs
larseggert: The `#[cfg(test)]` attribute doesn't work, because tests in `tests/` are compiled as a separate crate.

PR #3286, opened 2025-12-18
chore(mtu): prepare v0.3.0 release

mxinden:
Would allow https://github.com/quinn-rs/quinn/pull/2465/ to depend on the latest `mtu` crate version.


mtu/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "mtu"
 authors.workspace = true
 homepage.workspace = true
 repository.workspace = true
-version = "0.2.9"
+version = "0.3.0"
```
mxinden, line 6: Being conservative here. @larseggert in case you say we haven't done any breaking changes, we could as well go with `0.2.10`. I am not aware of any other users of `mtu`, thus I don't think it is worth the risk.

PR #3287, opened 2025-12-19
fix: integer overflow when parsing QPACK headers

### LINKED ISSUE #3285: OSS-Fuzz issue 470034085
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5492990622892032 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=470034085 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

larseggert:
The crash was caused by an integer overflow when parsing malicious QPACK headers. The fuzzer provided a test case with an extremely large length value (close to `usize::MAX`) which caused a panic when adding `self.offset + len` in the `slice()` method at `neqo-qpack/src/reader.rs:140`.

- Use `checked_add()` in `ReceiverBufferWrapper::slice()` to detect integer overflow instead of panicking
- Add `MAX_LEN` validation to `ReceiverBufferWrapper::read_literal_from_buffer()` matching `LiteralReader`'s existing protection
- Add tests to verify overflow detection and length limit enforcement

Fixes #3285

### DIFF
--- a/.github/actions/nss/action.yml
+++ b/.github/actions/nss/action.yml
@@ -88,32 +88,28 @@ runs:
           echo "SCCACHE_GHA_ENABLED=true" >> "$GITHUB_ENV"
         fi
 
-    - name: Checkout NSS
+    - name: Retrieve NSS
+      id: nss
       if: ${{ !steps.system_nss.outputs.suitable }}
-      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-      with:
-        repository: nss-dev/nss
-        path: nss
-        persist-credentials: false
+      shell: bash
+      env:
+        NSS_VERSION: 3.119.1 # TODO: Bump this periodically; also in qns/Dockerfile.
+      run: |
+        NSS_TAG=${NSS_VERSION//./_}
+        curl -L https://ftp.mozilla.org/pub/security/nss/releases/NSS_${NSS_TAG}_RTM/src/nss-${NSS_TAG}.tar.gz |
+          tar xz --strip-components=1
+        echo "version=$NSS_VERSION" >> "$GITHUB_OUTPUT"
 
     - name: Retrieve NSPR
       id: nspr
       if: ${{ !steps.system_nss.outputs.suitable }}
       shell: bash
       env:
-        NSPR_VERSION: 4.38.2 # This changes so rarely that we can hardcode it.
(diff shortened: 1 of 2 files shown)


neqo-qpack/src/reader.rs
```diff
@@ -137,11 +139,12 @@ impl<'a> ReceiverBufferWrapper<'a> {
     }
 
     fn slice(&mut self, len: usize) -> Res<&[u8]> {
-        if self.offset + len > self.buf.len() {
+        let end = self.offset.checked_add(len).ok_or(Error::Decompression)?;
+        if end > self.buf.len() {
```
martinthomson, line 143: I see that `bool::ok_or()` is coming, which will make this even nicer.


```rust
        (end <= self.buf.len()).ok_or(Error::Decompression)?;
```

neqo-qpack/src/reader.rs
```diff
@@ -688,4 +691,28 @@ mod tests {
...
     }
+
+    #[test]
+    fn buffer_wrapper_rejects_oversized_literal() {
+        const PREFIX_LEN: u8 = 3;
+        let mut data = Encoder::default();
+        data.encode_literal(
+            false,
+            Prefix::new(0x00, PREFIX_LEN),
+            &vec![b'a'; LiteralReader::MAX_LEN + 1],
```
martinthomson: Does this need to be a vec?  Or is it large enough that we need the allocation on the heap?

A better to do this is to encode the length directly, without the allocation for the literal at all.  That would make the test a whole lot simpler.

You would still call `read_literal_from_buffer()`, but you would expect it to fail with the right message.

neqo-qpack/src/reader.rs
```diff
@@ -688,4 +691,28 @@ mod tests {
...
+            Err(Error::Decompression)
+        );
+    }
+
+    #[test]
+    fn buffer_wrapper_slice_detects_overflow() {
+        let buf = [0u8; 10];
+        let mut wrapper = ReceiverBufferWrapper::new(&buf);
+        wrapper.offset = 5;
+        assert_eq!(wrapper.slice(usize::MAX), Err(Error::Decompression));
```
martinthomson, line 718:
```suggestion
        assert_eq!(wrapper.slice(7), Err(Error::Decompression));
        assert_eq!(wrapper.slice(usize::MAX), Err(Error::Decompression));
```

Test with something smaller as well.

martinthomson (approved):

PR #3290, opened 2025-12-19
chore: Use `derive_more` to reduce boilerplate

larseggert:
Already a dependency in Gecko. Let's us reduce boilerplate quite a bit.

### DIFF
--- a/neqo-crypto/src/agent.rs
+++ b/neqo-crypto/src/agent.rs
@@ -13,7 +13,7 @@ use std::{
     cell::RefCell,
     ffi::{CStr, CString},
     fmt::{self, Debug, Formatter},
-    mem::{size_of, MaybeUninit},
+    mem::MaybeUninit,
     ops::{Deref, DerefMut},
     os::raw::{c_uint, c_void},
     pin::Pin,


Cargo.lock
```diff
@@ -1522,6 +1551,12 @@ version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "84a22b9f218b40614adcb3f4ff08b703773ad44fa9423e4e0d346d5db86e4ebc"
 
+[[package]]
+name = "unicode-xid"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ebc1c04c71510c7f702b52b7c350734c9ff1295c464a03335b00bb84fc54f853"
+
```
mxinden, line 1559: I don't think this is included in mozilla-central today. Is it necessary?

neqo-crypto/src/p11.rs
```diff
@@ -41,11 +34,29 @@ pub use nss_p11::*;
...
 #[macro_export]
 macro_rules! scoped_ptr {
+    // With custom debug method
+    ($scoped:ident, $target:ty, $dtor:path, $debug_method:ident) => {
+        #[derive(derive_more::Debug)]
+        #[debug("{}", self.$debug_method().map_or_else(|_| format!("Opaque {}", stringify!($scoped)), |b| format!("{} {}", stringify!($scoped), neqo_common::hex_with_len(b))))]
+        pub struct $scoped {
+            ptr: *mut $target,
+        }
+
```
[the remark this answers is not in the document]
martinthomson: This seems mostly good, though I would maybe use `concat!()` rather than `format!()` to place the type name; the second arm still needs `format!()`, but it can compile the type name down easier with `concat!()`.

Cargo.lock
larseggert: It's required for deriving `Debug` and `Display`, which is where a lot of the savings come from.

https://github.com/JelteF/derive_more/blob/f7bb41ac054c060caaf5ff3212e74e42794cb4b4/impl/Cargo.toml#L56C35-L59

mxinden: Unfortunately `unicode-xid` isn't currently allowed in mozilla-central. See below for context. We might have to roll back here.

https://bugzilla.mozilla.org/show_bug.cgi?id=1986247#c8

More to come.

PR #3291, opened 2025-12-19
chore: Replace `url` crate with `http` crate

larseggert:
`http` is much more lightweight, but does not do IDNA. **Do we care in neqo?**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1342,12 +1342,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "84a22b9f218b40614adcb3f4ff08b703773ad44fa9423e4e0d346d5db86e4ebc"
 
 [[package]]
-name = "unicode-xid"
-version = "0.2.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ebc1c04c71510c7f702b52b7c350734c9ff1295c464a03335b00bb84fc54f853"
-
-[[package]]
 name = "version_check"
 version = "0.9.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"


mxinden:
@larseggert what do you mean with "lightweight" here? Compile times? Binary size? Code size? Complexity? Asked differently, what do we gain from switching? Are the switching costs worth it?

larseggert:
Compile times. And probably binary size, but Gecko uses url and so that doesn't really matter.

mxinden:
How much faster does Neqo compile with `http` vs `url`?

How well is each crate maintained? I assume we (Mozilla) have connections to maintainers of `url`, given that it is a servo project. Is the same true for `http`?

martinthomson:
The http crate is pretty central to a lot of stuff, I wouldn't worry about it going neglected.  If we can avoid pulling in more dependencies by using it (we are only doing this for tests, so full URL spec support is probably overkill), that's good.  (You could feature-swap, but that's the antithesis of the goal here.)

neqo-bin/src/client/mod.rs
```diff
@@ -161,15 +161,19 @@ impl Args {
...
+        use std::iter::repeat_with;
 
         let addr =
             server_addr.map_or_else(|| "[::1]:12345".into(), |a| format!("[::1]:{}", a.port()));
         Self {
             shared: SharedArgs::default(),
-            urls: repeat_with(|| Url::from_str(&format!("http://{addr}/{download_size}")).unwrap())
-                .take(num_requests)
-                .collect(),
+            urls: repeat_with(|| {
```
martinthomson: You can use `repeat(format!(...).parse().unwrap())`: https://docs.rs/http/latest/http/uri/struct.Uri.html#impl-Clone-for-Uri

neqo-bin/src/client/mod.rs
```diff
@@ -536,23 +540,20 @@ const fn local_addr_for(remote_addr: &SocketAddr, local_port: u16) -> SocketAddr
...
+fn urls_by_origin(urls: &[Uri]) -> impl Iterator<Item = ((String, u16), VecDeque<Uri>)> {
     urls.iter()
         .fold(
-            HashMap::<Origin, VecDeque<Url>>::default(),
-            |mut urls, url| {
-                urls.entry(url.origin()).or_default().push_back(url.clone());
-                urls
+            HashMap::<(String, u16), VecDeque<Uri>>::default(),
+            |mut map, uri| {
+                if let Some(authority) = uri.authority() {
```
martinthomson: I think that I'd prefer `.expect()` here, rather than the warning or the new silent discard.  It's not like this isn't something users can't control.

larseggert:
*Is* this only used in tests though? This changes `RequestTarget` in `neqo-http3/src/request_target.rs`, and I'm not sure that is only used in tests or for `neqo-bin`...

@mxinden any clues?

Cargo.lock
```diff
@@ -627,121 +613,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
- "yoke",
- "zerofrom",
- "zerovec",
-]
-
-[[package]]
-name = "icu_provider_macros"
-version = "1.5.0"
+name = "http"
+version = "0.2.12"
```
mxinden: Firefox uses `0.2.19`:

https://github.com/mozilla-firefox/firefox/blob/131497bb1b747587b2b21b1abf14f44ecffad805/Cargo.lock#L3048

Cargo.lock
```diff
@@ -89,6 +89,12 @@ dependencies = [
  "generic-array",
 ]
 
+[[package]]
+name = "bytes"
+version = "1.11.0"
```
mxinden: Firefox uses `1.4.0`:

https://github.com/mozilla-firefox/firefox/blob/main/Cargo.lock#L686

mxinden:
> _Is_ this only used in tests though? This changes `RequestTarget` in `neqo-http3/src/request_target.rs`, and I'm not sure that is only used in tests or for `neqo-bin`...
> 
> @mxinden any clues?

Yes, only used in `neqo-bin`. I assume we could do the parsing ourselves in `neqo-bin`.

That said, I am still not sure what the goal is here. If it is compile time of Neqo, how much does it really safe us? Note that Firefox already depends on both crates, thus I assume this pull request will only affect Neqo as a standalone. 

Is this worth the engineering effort?

larseggert:
It's up to us to decide if this is worth it. It reduces compile times by about 10% (1-2sec) for me. It also reduces the deps quite a bit, which makes keeping them synced with Gecko a bit more manageable maybe.

mxinden:
For what it is worth, I can reproduce the 2 second build time improvement. Each with `cargo clean`:

main:

```
Executed in   24.70 secs    fish           external
   usr time  351.47 secs    0.00 micros  351.47 secs
   sys time   17.42 secs  981.00 micros   17.42 secs
```

this pull request:

```
Executed in   22.26 secs    fish           external
   usr time  301.72 secs    0.05 millis  301.72 secs
   sys time   15.67 secs    1.10 millis   15.67 secs
```

PR #3292, opened 2025-12-19
chore: Remove `regex` crate

larseggert:
It's heavyweight and custom code is sufficient for our simple needs.


mxinden:
What is the goal here? Reduce compile time? Reduce binary size?

Cargo.lock
mxinden, line 1: @larseggert note that this did not actually remove the `regex` crate from the project.

larseggert: It's needed by two dev-dependencies (`codspeed-criterion-compat` and `bindgen`).

PR #3311, opened 2026-01-12
fix: Pin `NSS`

larseggert:
So that CI doesn't break when NSS changes unexpectedly.


mxinden (approved):

martinthomson:
FWIW, I tend to appreciate notice of NSS changes.  I don't appreciate the opportunity to drop everything and fix bugs.  Is there a canary we can use for this?

PR #3318, opened 2026-01-14
fix: Fix QNS reporting in CI

larseggert:


larseggert:
Merging, since this needs to be on `main` to test.

PR #3315, opened 2026-01-13
chore: Update deps to Gecko versions

larseggert:
Except `itertools`, which fails with
```
cargo update -p itertools --precise 0.14.0
    Updating crates.io index
error: failed to select a version for the requirement `itertools = ">=0.10, <0.14"`
candidate versions found which didn't match: 0.14.0
location searched: crates.io index
required by package `bindgen v0.72.0`
    ... which satisfies dependency `bindgen = "^0.72"` (locked to 0.72.0) of package `neqo-crypto v0.21.0 (/Volumes/mozdrone/Users/mozdrone/Documents/Code/neqo/main/neqo-crypto)`
    ... which satisfies path dependency `neqo-crypto` (locked to 0.21.0) of package `fuzz v0.21.0 (/Volumes/mozdrone/Users/mozdrone/Documents/Code/neqo/main/fuzz)`
```

PR #3320, opened 2026-01-14
chore: Transition to Rust 2024 edition

### LINKED ISSUE #2479: Update to Rust 2024 edition
[Rust 2024 edition](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html) released.

> each crate can decide when to migrate to a new edition independently. This decision is 'private' - it won't affect other crates in the ecosystem.

https://doc.rust-lang.org/edition-guide/editions/index.html#editions-do-not-split-the-ecosystem

In other words, we can update `neqo-*` at any time without having to wait for dependencies nor consumers to update.

I am not aware of a Mozilla Rust edition policy.

larseggert:
This was **super easy** following https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html and required only two trivial manual code changes. Very impressed.

Fixes #2479

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -328,8 +328,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             self.current.acked_bytes += new_acked;
             if self.current.acked_bytes >= bytes_for_increase {
                 self.current.acked_bytes -= bytes_for_increase;
-                self.current.congestion_window += self.max_datagram_size(); // or is this the
-                // current MTU?
+                self.current.congestion_window += self.max_datagram_size(); // or is this the current MTU?
             }
             // The number of bytes we require can go down over time with Cubic.
             // That might result in an excessive rate of increase, so limit the number of unused


mtu/src/lib.rs
```diff
@@ -54,7 +54,7 @@ use std::{
 
 #[cfg(not(target_os = "windows"))]
 macro_rules! asserted_const_with_type {
-    ($name:ident, $t1:ty, $e:expr, $t2:ty) => {
+    ($name:ident, $t1:ty, $e:expr_2021, $t2:ty) => {
```
mxinden: Looking at the docs below, is this intermediate migration step actually needed?

https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html

mtu/Cargo.toml
```diff
@@ -4,7 +4,7 @@ authors.workspace = true
 homepage.workspace = true
 repository.workspace = true
 version = "0.3.0"
-edition = "2021"
+edition = "2024"
```
mxinden: Why not `edition.workspace = true`?

neqo-bin/src/client/http09.rs
```diff
@@ -340,10 +340,13 @@ impl<'b> Handler<'b> {
...
                 if fin_recvd {
-                    if let Some(mut out_file) = maybe_out_file.take() {
-                        out_file.flush()?;
-                    } else {
-                        qinfo!("<FIN[{stream_id}]>");
+                    match maybe_out_file.take() {
+                        Some(mut out_file) => {
+                            out_file.flush()?;
+                        }
+                        _ => {
```
mxinden: Why not an exhaustive match, i.e. `None`?

neqo-bin/src/client/http3.rs
```diff
@@ -305,10 +305,13 @@ impl StreamHandler for DownloadStreamHandler {
...
         if fin {
-            if let Some(mut out_file) = self.out_file.take() {
-                out_file.flush()?;
-            } else {
-                qdebug!("<FIN[{stream_id}]>");
+            match self.out_file.take() {
+                Some(mut out_file) => {
+                    out_file.flush()?;
+                }
+                _ => {
```
mxinden: Same here, preference for an exhaustive match.

neqo-bin/src/client/mod.rs
```diff
@@ -536,7 +539,7 @@ const fn local_addr_for(remote_addr: &SocketAddr, local_port: u16) -> SocketAddr
     }
 }
 
-fn urls_by_origin(urls: &[Url]) -> impl Iterator<Item = ((Host, u16), VecDeque<Url>)> {
+fn urls_by_origin(urls: &[Url]) -> impl Iterator<Item = ((Host, u16), VecDeque<Url>)> + use<> {
```
mxinden: Is the `use<>` necessary here?

neqo-common/src/codec.rs
```diff
@@ -981,7 +981,7 @@ mod tests {
     }
 
     macro_rules! uint_tc {
-        [$( $v:expr => $b:expr ),+ $(,)?] => {
+        [$( $v:expr_2021 => $b:expr_2021 ),+ $(,)?] => {
```
mxinden: Same. Given that the macro has a single arm, is the `expr_2021` necessary?

neqo-common/src/incrdecoder.rs
```diff
@@ -202,7 +202,7 @@ mod tests {
     }
 
     macro_rules! uint_tc {
-        [$( $b:expr => $v:expr ),+ $(,)?] => {
+        [$( $b:expr_2021 => $v:expr_2021 ),+ $(,)?] => {
```
mxinden: Needed?

neqo-crypto/src/aead.rs
```diff
@@ -123,19 +123,21 @@ impl RealAead {
...
-            cipher,
-            secret,
-            p.as_ptr().cast(),
-            c_uint::try_from(p.len())?,
-            &raw mut ctx,
-        )?;
-        Ok(Self {
-            ctx: AeadContext::from_ptr(ctx)?,
-        })
+        unsafe {
```
mxinden, line 127: Does all of it need the `unsafe`? Can the scope be reduced?

neqo-http3/src/frames/reader.rs
```diff
@@ -262,33 +262,38 @@ impl FrameReader {
...
+                #[cfg(feature = "build-fuzzing-corpus")]
+                if let Some(corpus) = T::FUZZING_CORPUS {
+                    // Write zero-length frames to the fuzzing corpus to test parsing of frames with
+                    // only type and length fields.
+                    self.write_item_to_fuzzing_corpus(corpus, None);
+                }
+                self.reset();
+                return Ok(Some(f));
+            }
+            _ => {
```
mxinden: Exhaustive match instead?

neqo-http3/src/server_events.rs
```diff
@@ -535,7 +535,7 @@ impl Http3ServerEvents {
     }
 
     /// Take all events
-    pub fn events(&self) -> impl Iterator<Item = Http3ServerEvent> {
+    pub fn events(&self) -> impl Iterator<Item = Http3ServerEvent> + use<> {
```
mxinden, line 582: Needed?

neqo-qpack/src/huffman_table.rs
```diff
@@ -11,7 +11,7 @@ pub struct Entry {
 }
 
 macro_rules! huffman_table {
-    [$($v:expr => $l:expr),+ $(,)?] => {
+    [$($v:expr_2021 => $l:expr_2021),+ $(,)?] => {
```
mxinden: Needed?

neqo-qpack/src/prefix.rs
```diff
@@ -54,7 +54,7 @@ macro_rules! create_prefix {
             mask: 0xFF,
         };
     };
-    ($n:ident, $v:expr, $l:expr) => {
+    ($n:ident, $v:expr_2021, $l:expr_2021) => {
```
mxinden: Needed?

neqo-qpack/src/prefix.rs
```diff
@@ -63,7 +63,7 @@ macro_rules! create_prefix {
             mask: ((1 << $l) - 1) << (8 - $l),
         };
     };
-    ($n:ident, $v:expr, $l:expr, $m:expr) => {
+    ($n:ident, $v:expr_2021, $l:expr_2021, $m:expr_2021) => {
```
mxinden: Needed?

neqo-qpack/src/static_table.rs
```diff
@@ -26,7 +26,7 @@ impl StaticTableEntry {
 }
 
 macro_rules! static_table_entries {
-    [$($i:expr, $n:expr, $v:expr);+ $(;)?] => {
+    [$($i:expr_2021, $n:expr_2021, $v:expr_2021);+ $(;)?] => {
```
mxinden: Needed?

larseggert:
All of these changes came from the clippy auto-update step. Can check if they are really needed.

mtu/src/windows.rs
```diff
@@ -14,13 +14,13 @@ use std::{
 use windows::Win32::{
     NetworkManagement::{
         IpHelper::{
-            if_indextoname, FreeMibTable, GetBestInterfaceEx, GetIpInterfaceTable,
-            MIB_IPINTERFACE_ROW, MIB_IPINTERFACE_TABLE,
+            FreeMibTable, GetBestInterfaceEx, GetIpInterfaceTable, MIB_IPINTERFACE_ROW,
+            MIB_IPINTERFACE_TABLE, if_indextoname,
```
martinthomson, line 18: This is just a change in sort order.  Is that edition-specific?

neqo-bin/src/client/http09.rs
martinthomson: Oh, wow, the change to the `if let` scoping is not something I like.  I think that the right way to do this is to do it differently...

neqo-bin/src/client/http09.rs
```diff
@@ -340,10 +340,13 @@ impl<'b> Handler<'b> {
...
+                            qinfo!("<FIN[{stream_id}]>");
+                        }
```
martinthomson:
```suggestion
                    maybe_out_file.take().map_or_else(|| {
                        qinfo("<FIN[{stream_id}]>");
                        Ok
                    }, |mut out_file| out_file.flush())?;
```

neqo-bin/src/client/http3.rs
martinthomson: See above.

neqo-bin/src/client/mod.rs
martinthomson: The iterator needs to pick up the lifetime of the `urls` argument somehow.  The other way to do this is:


```suggestion
fn urls_by_origin<'a>(urls: &'a [Url]) -> impl Iterator<Item = ((Host, u16), VecDeque<Url>)> + 'a {
```

I think.  I had to deal with similar things elsewhere.

neqo-crypto/src/aead.rs
martinthomson: I  think that only the MakeAead invocation needs it.

neqo-crypto/src/agent.rs
```diff
@@ -501,26 +501,30 @@ impl SecretAgent {
         _check_sig: PRBool,
         _is_server: PRBool,
     ) -> ssl::SECStatus {
-        let auth_required_ptr = arg.cast::<bool>();
-        *auth_required_ptr = true;
-        // NSS insists on getting SECWouldBlock here rather than accepting
-        // the usual combination of PR_WOULD_BLOCK_ERROR and SECFailure.
-        ssl::_SECStatus_SECWouldBlock
+        unsafe {
```
martinthomson, line 504: This could be more narrowly scoped to just the pointer deref.

neqo-crypto/src/agent.rs
```diff
@@ -501,26 +501,30 @@ impl SecretAgent {
...
     ) {
-        let alert = alert.as_ref().unwrap();
-        if alert.level == 2 {
-            // Fatal alerts demand attention.
-            let st = arg.cast::<Option<Alert>>().as_mut().unwrap();
-            if st.is_none() {
-                *st = Some(alert.description);
-            } else {
-                qwarn!("[{fd:p}] duplicate alert {}", alert.description);
+        unsafe {
```
martinthomson: I think that this needs two unsafe blocks, not a big one.

neqo-crypto/src/agent.rs
```diff
@@ -1062,37 +1066,40 @@ impl Client {
...
-        let Some(resumption) = arg.cast::<Vec<ResumptionToken>>().as_mut() else {
-            return ssl::SECFailure;
-        };
-        let Ok(len) = usize::try_from(len) else {
-            return ssl::SECFailure;
-        };
-        let mut v = Vec::with_capacity(len);
-        v.extend_from_slice(null_safe_slice(token, len));
-        qdebug!("[{fd:p}] Got resumption token {}", hex_snip_middle(&v));
+        unsafe {
```
martinthomson: This is more complicated (and more unsafe), but it might still be worth trying to narrow the scope some.

neqo-crypto/src/agentio.rs
```diff
@@ -87,18 +87,20 @@ impl RecordList {
...
+            let Ok(epoch) = Epoch::try_from(epoch) else {
+                return ssl::SECFailure;
+            };
+            let Ok(ct) = ContentType::try_from(ct) else {
+                return ssl::SECFailure;
+            };
+            let Some(records) = arg.cast::<Self>().as_mut() else {
+                return ssl::SECFailure;
+            };
+            let slice = null_safe_slice(data, len);
```
martinthomson: I think that this is the only unsafe line here.

martinthomson (commented):
Formatting changes are making this patch longer than I'd like.  Is there a way to split those changes out?  I agree with most of what Max has highlighted though.  Limiting the scope of `unsafe` wraps is probably the biggest one (we want narrow scope, so that reviewers can spot where the danger is more easily).

mtu/src/windows.rs
larseggert: Yes: https://doc.rust-lang.org/nightly/style-guide/editions.html

> Change “ASCIIbetical” sort to Unicode-aware “non-lowercase before lowercase”.

neqo-bin/src/client/mod.rs
larseggert: That particular one (and one other instance) seems to be removable, but some others aren't.

neqo-crypto/build.rs
```diff
@@ -87,7 +87,10 @@ fn setup_clang() {
     };
     let libclang_dir = mozbuild_root.join("clang").join("lib");
     if libclang_dir.is_dir() {
-        env::set_var("LIBCLANG_PATH", libclang_dir.to_str().unwrap());
+        // TODO: Audit that the environment access only happens in single-threaded code.
```
martinthomson: Um, this is a build script that only runs in a single thread.  I think we're OK.
```suggestion
```

martinthomson: Or maybe
```suggestion
        // SAFETY: env::set_var only works reliably in single-threaded contexts.
        // Good news: this is one of those.
```

neqo-http3/src/frames/reader.rs
mxinden: Why Is this change necessary? Why is a `match` with an `_ =>` better than the previous exhaustive `if let Some(corpus)`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -306,7 +313,7 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             if self.current.acked_bytes >= bytes_for_increase {
                 self.current.acked_bytes -= bytes_for_increase;
                 self.current.congestion_window += self.max_datagram_size(); // or is this the
-                                                                            // current MTU?
+                // current MTU?
```
mxinden: Bad formatting. Can this be on top of the `self.current.congestion_window += self.max_datagram_size();` line instead?

neqo-transport/src/connection/mod.rs
```diff
@@ -859,16 +859,17 @@ impl Connection {
...
+                    .generate_new_token(path.borrow().remote_address(), now)
+                {
+                    self.new_token.send_new_token(token);
+                }
+                Ok(())
             }
-            Ok(())
-        } else {
-            Err(Error::NotConnected)
+            _ => Err(Error::NotConnected),
```
mxinden: Why is this better than an `if let Some`? I really value Rust's exhaustive matching. Why move away from it?

neqo-transport/src/connection/mod.rs
```diff
@@ -989,10 +990,11 @@ impl Connection {
                                 now,
                             );
                         }
-                    } else {
+                    }
+                    _ => {
```
mxinden: Same. Why is this change needed?

neqo-transport/src/connection/mod.rs
```diff
@@ -1933,21 +1936,30 @@ impl Connection {
...
+                    self.paths.make_permanent(path, None, cid, now);
                     Ok(())
-                } else {
-                    qtrace!("[{self}] Unable to make path permanent: {}", path.borrow());
-                    Err(Error::InvalidMigration)
                 }
-            } else {
-                qtrace!("[{self}] Unable to make path permanent: {}", path.borrow());
-                Err(Error::InvalidMigration)
+                _ => {
```
mxinden: Why the `_`? Isn't it always `None`?

neqo-transport/src/connection/mod.rs
```diff
@@ -2862,12 +2874,15 @@ impl Connection {
...
-            self.set_state(State::Closing { error, timeout }, now);
-        } else {
-            self.set_state(State::Closed(error), now);
+        match self.paths.primary() {
+            Some(path) => {
+                self.state_signaling
+                    .close(path, error.clone(), FrameType::Padding, msg);
+                self.set_state(State::Closing { error, timeout }, now);
+            }
+            _ => {
```
mxinden: Preference for `None` or the previous `if let Some`.

neqo-transport/src/addr_valid.rs
```diff
@@ -238,43 +238,46 @@ impl AddressValidation {
...
                 }
-            } else if cid.is_empty() {
-                // An empty connection ID means NEW_TOKEN.
-                if self.validation == ValidateAddress::Always {
-                    qinfo!("AddressValidation: valid NEW_TOKEN token; validating again");
-                    AddressValidationResult::Validate
-                } else {
-                    qinfo!("AddressValidation: valid NEW_TOKEN token; accepting");
+            }
+            _ => {
```
mxinden: `None` or the previous `if let Some`. Unless there is a good reason for the `_`?

test/update-lockfile.sh
```diff
@@ -14,21 +14,15 @@ fi
 
 check-lockfile-intersection \
     https://raw.githubusercontent.com/mozilla-firefox/firefox/refs/heads/main/Cargo.lock \
-    Cargo.lock 2> /dev/null | while read -r d ; do
```
mxinden, line 17: Related? Does this fix the issues seen in https://github.com/mozilla/neqo/pull/3315/?

mxinden:
//CC @not-nik as I assume this patch will result in many conflicts with https://github.com/mozilla/nss-rs/.

Not-Nik:
Looks like it's mostly formatting and a few extra `unsafe`s, right. Shouldn't take much longer than a day to integrate.

 ~ famous last words

PR #3330, opened 2026-01-19
Build changes for NSS 3.121

martinthomson:
This isn't released yet, but as NSS tip it is going to be soon.

Draft because we'll want to be careful when we require a bump to NSS.

Note: The code this changes was clearly not reviewed.  That's not good.

### DIFF
--- a/neqo-crypto/build.rs
+++ b/neqo-crypto/build.rs
@@ -192,7 +192,6 @@ fn static_link() {
     }
     // Hardware specific libs.
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
-    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
     if target_arch == "arm" || target_arch == "aarch64" {
         static_libs.push("armv8_c_lib");
     }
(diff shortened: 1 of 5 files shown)


neqo-crypto/build.rs
```diff
@@ -188,33 +189,26 @@ fn static_link() {
...
     }
     if target_arch == "x86_64" {
         static_libs.push("hw-acc-crypto-avx");
         static_libs.push("hw-acc-crypto-avx2");
-    }
-    // https://searchfox.org/nss/rev/08c4d05078d00089f8d7540651b0717a9d66f87e/lib/freebl/freebl.gyp#315-324
-    if (target_os == "android" || target_os == "linux") && target_arch == "x86_64" {
-        static_libs.push("intel-gcm-wrap_c_lib");
-        // https://searchfox.org/nss/rev/08c4d05078d00089f8d7540651b0717a9d66f87e/lib/freebl/freebl.gyp#43-47
-        if (target_os == "android" || target_os == "linux") && target_arch == "x86_64" {
```
martinthomson, line 218: Take a look at line 213 above...

larseggert:
@martinthomson we also have a patch from @jschanck here, FYI: https://mozilla.slack.com/archives/CD2LCAGJD/p1768239009024569?thread_ts=1768207491.435279&cid=CD2LCAGJD

martinthomson (commented):
John's patch wasn't complete.  I think he missed the arm/aarch64 changes, but he definitely missed the version number changes elsewhere.  Thanks for catching the Dockerfile tweak.

larseggert:
@martinthomson NSS 3.121 is out

martinthomson:
OK, I gave it a shot, but the windows and android builds are messed up in ways I might not have time to debug.

larseggert:
This seems to now work. The NetBSD and OpenBSD failures are due to the packaged NSS there being too old (and NSS not being buildable from source there).

Also, we're dropping `i686-linux-android`, because `application-services` did, too, and we reuse their patch scripts in CI. I added `aarch64-linux-android`, but there is no emulator for that, so it's build-only.

PR #3334, opened 2026-01-20
chore: Tests for `mtu` using Linux namespaces

larseggert:


mtu/tests/netns.rs
```diff
@@ -0,0 +1,396 @@
...
+            "add",
+            "10.30.0.0/24",
+            "via",
+            "10.0.1.2",
+            "dev",
+            &vi_main,
+        ],
+    );
+
+    // Route in table 100 via VPN interface.
```
mxinden, line 355: Doesn't our implementation always use `RT_TABLE_MAIN`? Would it find routes in table 100?

https://github.com/mozilla/neqo/blob/26711da50ad60136b627155dc5d1a9d8971a5f94/mtu/src/linux.rs#L122

larseggert: That field doesn't matter unless we also give it `RTM_F_LOOKUP_TABLE`, in which case the output would be filtered to the given table. (The test checks exactly that, two routes with different MTUs in different tables + policy routing.)

Would it be clearer if we used `RT_TABLE_UNSPEC` in the code?

mxinden: I am fine either way.

PR #3335, opened 2026-01-20
fix: Restart PMTUD when full-sized packets are lost

larseggert:
Turning on PMTUD in Firefox causes loss rates to go up. My current theory is that there are VPNs or other things in the path that lower the MTU below what the outgoing interface MTU is but that don't cause IP address changes that would trigger path challenges.

This mechanism adds a heuristic to restart PMTUD when we see loss of full-sized packets while smaller packets get ACKed, or when a PTO happens.

We had a version of this in the past that #3200 removed due to it triggering during normal bulk uploads.

### DIFF
--- a/neqo-transport/src/connection/tests/datagram.rs
+++ b/neqo-transport/src/connection/tests/datagram.rs
@@ -30,8 +30,7 @@ use crate::{
 /// - 1 byte: DATAGRAM frame type
 /// - 16 bytes: AEAD authentication tag
 const MIN_DATAGRAM_PACKET_OVERHEAD: usize = 8 + 1 + 1 + 1 + 16;
-const DATAGRAM_LEN_MTU: usize =
-    default_plpmtu(DEFAULT_ADDR.ip()) - MIN_DATAGRAM_PACKET_OVERHEAD;
+const DATAGRAM_LEN_MTU: usize = default_plpmtu(DEFAULT_ADDR.ip()) - MIN_DATAGRAM_PACKET_OVERHEAD;
 const DATA_MTU: &[u8] = &[1; DATAGRAM_LEN_MTU];
 const DATA_BIGGER_THAN_MTU: &[u8] = &[0; 2 * DATAGRAM_LEN_MTU];
 const_assert!(DATA_BIGGER_THAN_MTU.len() > DATAGRAM_LEN_MTU);


martinthomson:
I'm going to defer to @mxinden on this one.  At least for initial passes.  Happy to jump in as needed.

neqo-transport/src/pmtud.rs
```diff
@@ -76,18 +80,25 @@ impl Pmtud {
...
         let search_table = Self::search_table(remote_ip);
-        let probe_index = 0;
+        let header_size = Self::header_size(remote_ip);
+        let base_plpmtu = search_table[0] - header_size;
         Self {
             search_table,
-            header_size: Self::header_size(remote_ip),
-            mtu: search_table[probe_index],
+            header_size,
+            mtu: search_table[0],
```
martinthomson: Swapping `probe_index` for a 0 here loses some of the intentionality of the code.  I think that you probably mean `probe_index` on this line as well as the added assignment to `base_plpmtu`.  Maybe you can call `min_mtu()` instead?

neqo-transport/src/pmtud.rs
```diff
@@ -267,6 +304,133 @@ impl Pmtud {
...
+    }
+
+    /// Handle ACK of packets. If a large packet was acknowledged,
+    /// the path is working for that size, so reset detection.
+    fn on_ack(&mut self, acked_pkts: &[sent::Packet]) {
+        if self.min_lost_size.is_none() {
+            return;
+        }
+
+        let max_acked = acked_pkts
```
martinthomson: You use the `let Some(_) = ... else { return; }` patter below, but not here.  The other might be more readable.

neqo-transport/src/pmtud.rs
```diff
@@ -267,6 +304,133 @@ impl Pmtud {
...
+            .iter()
+            .filter(|p| p.on_primary_path())
```
martinthomson, line 406: Would it make sense to use `self.is_large_data_packet()` here as well?  At least you want to ensure that the packet was sent on the active path.  The difference I see is that including probes is fine here, though you don't want that for loss (because loss of probes is fine).

test-fixture/src/sim/connection.rs
```diff
@@ -343,3 +343,106 @@ impl Goal for ReceiveData {
...
+        e: &ConnectionEvent,
+        now: Instant,
+    ) -> GoalStatus {
+        self.0.handle_event(c, e, now)
+    }
+}
+
+/// Sends data and verifies that PMTUD black hole detection did NOT trigger.
+#[derive(Debug)]
+pub struct SendDataExpectNoPmtudChange(SendDataCheckPmtud);
```
martinthomson: Why did you choose to wrap these types this way?  Using the bool parameter seems easier than having an entirely new type for each value.

test-fixture/src/sim/mtu.rs
```diff
@@ -31,13 +31,57 @@ impl Node for Mtu {
...
+    }
+}
+
+/// Drops all datagrams larger than the configured MTU, with the ability to
+/// change MTU after a specified number of packets have passed.
+#[derive(Debug)]
+pub struct DynamicMtu {
+    initial_mtu: usize,
+    reduced_mtu: usize,
+    switch_after: usize,
```
martinthomson, line 46: Documentation for this should indicate that it is measured in units of packets.

test-fixture/src/sim/mtu.rs
```diff
@@ -31,13 +31,57 @@ impl Node for Mtu {
...
+impl DynamicMtu {
+    /// Creates a new [`DynamicMtu`] that starts at `initial_mtu` and switches
+    /// to `reduced_mtu` after `switch_after` packets have passed.
+    ///
+    /// Both MTU values include IP and UDP header size.
+    #[must_use]
+    pub const fn new(initial_mtu: usize, reduced_mtu: usize, switch_after: usize) -> Self {
+        Self {
+            initial_mtu,
+            reduced_mtu,
```
martinthomson: The name implies that the change is a reduction.  But it's just another value to change to.  Consider renaming to match.

(Our code won't handle MTU increases, but that's OK.)

test-fixture/src/sim/mtu.rs
```diff
@@ -31,13 +31,57 @@ impl Node for Mtu {
     fn init(&mut self, _rng: Rng, _now: Instant) {}
 
     fn process(&mut self, d: Option<Datagram>, _now: Instant) -> Output {
-        d.filter(|dgram| {
-            let header = match dgram.destination().ip() {
-                IpAddr::V4(_) => 20 + 8,
-                IpAddr::V6(_) => 40 + 8,
-            };
+        d.filter(|dgram| Pmtud::header_size(dgram.destination().ip()) + dgram.len() <= self.mtu)
```
martinthomson: Having Pmtud publicly exposed is a bit unfortunate.  Is this value something that the MTU crate can expose for us instead?

neqo-transport/tests/network.rs
```diff
@@ -210,3 +213,61 @@ fn transfer_fixed_seed() {
     sim.seed_str("117f65d90ee5c1a7fb685f3af502c7730ba5d31866b758d98f5e3c2117cf9b86");
     sim.run();
 }
+
+/// Tests PMTUD black hole detection when PMTU decreases mid-transfer.
+/// The `DynamicMtu` node reduces MTU after initial PMTUD completes, simulating
+/// a VPN coming up. PMTUD should detect the black hole and reduce MTU.
+#[test]
+fn pmtud_mtu_decrease() {
```
martinthomson, line 218: I like these tests.  They are pretty easy to read.

neqo-transport/src/recovery/mod.rs
```diff
@@ -906,6 +908,13 @@ impl Loss {
             qtrace!("[{self}] PTO {pn_space}, probing {allow_probes:?}");
             self.fire_pto(pn_space, allow_probes, now);
 
+            // Notify PMTUD about PTO.
+            primary_path.borrow_mut().on_pto(
+                &lost[pto_packets_start..],
```
martinthomson: I'd like to ask a little about why you chose to only consider the "PTO packets" in this.  I don't follow the reasoning.  The reason we mark a few outstanding packets as "lost" during PTO is so that we have something constructive to send, not because they were genuinely marked as lost.  This elevates their status to being lost because of the PTO, which might be OK, but I'm curious about why you think that might be the right balance (as opposed to passing all lost packets; what other packets are included in this slice?)

neqo-transport/src/pmtud.rs
```diff
@@ -182,35 +205,42 @@ impl Pmtud {
...
     /// times, the PMTUD process is stopped at the current MTU.
     pub fn on_packets_lost(
         &mut self,
         lost_packets: &[sent::Packet],
         stats: &mut Stats,
         now: Instant,
     ) {
+        // Only restart PMTUD if we aren't already at the minimum MTU.
+        if self.mtu > self.min_mtu() && self.black_hole.on_loss(lost_packets, now) {
+            stats.pmtud_change += 1;
```
martinthomson: Is "change" the right name for this?  It seems like this is a reset count.  Or a count of the number of black holes detected.

neqo-transport/src/pmtud.rs
larseggert: We can't use `min_mtu()` since this is the constructor, but I can go back to `probe_index`.

neqo-transport/src/pmtud.rs
larseggert: We do want to include ACKs of probe packets here. I'll see if adding a Boolean argument to `is_large_data_packet` is cleaner.

larseggert: Actually, this is uglier. Propose to leave as-is.

neqo-transport/src/pmtud.rs
larseggert: pmtud_count?

test-fixture/src/sim/mtu.rs
larseggert: In order to not make `Pmtud` public, we also need to make helpers for `search_table` and `default_plpmtu`, which is a bit noisy. Doing this now, but we might want to rethink.

neqo-transport/src/recovery/mod.rs
larseggert: So the PTO-based trigger is meant as a fallback for when the ACK-based trigger is prevented from firing (because we got no ACKs). IMO it doesn't really matter whether these packets were actually lost or just not ACKed for some longish amount of time - what matters is that the PTO is a signal that something is messed up with the connection.

neqo-transport/src/pmtud.rs
martinthomson: It might be worth some commentary: we ignore probes on loss, but pay attention on ack.  But otherwise, we only care about whether something is on the primary path.

test-fixture/src/sim/mtu.rs
martinthomson: What code outside of `Pmtud` needs the table?

neqo-transport/src/recovery/mod.rs
martinthomson: That might be worth commenting.  The key point is not that these particular packets are lost, but that the PTO is a trigger.

You should be careful here: we mark the very last packets in a flight as "lost" when we hit a PTO.  If we're done sending (and not blocked on other things), those packets might be small.  And small packets won't trigger the black hole logic, I think.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -13,9 +13,9 @@ use super::{
     AT_LEAST_PTO, assert_error, connect_force_idle, default_server, new_client, new_server, now,
 };
 use crate::{
-    CloseReason, Connection, ConnectionParameters, Error, MIN_INITIAL_PACKET_SIZE, Pmtud,
-    StreamType,
+    CloseReason, Connection, ConnectionParameters, Error, MIN_INITIAL_PACKET_SIZE, StreamType,
     connection::tests::DEFAULT_ADDR,
+    default_plpmtu,
```
martinthomson, line 18: The uses I see of this function are all internal to the crate, so you might get away with `pub(crate)` for this sort of access.  It's mostly OK to have that sort of coupling between things internally.

I am still a little concerned about adding these internals to the public interface.

neqo-transport/src/stats.rs
```diff
@@ -392,8 +394,13 @@ impl Debug for Stats {
         writeln!(f, "  ss_exit: {}", self.cc.slow_start_exited)?;
         writeln!(
             f,
-            "  pmtud: {} sent {} acked {} lost {} iface_mtu {} pmtu",
-            self.pmtud_tx, self.pmtud_ack, self.pmtud_lost, self.pmtud_iface_mtu, self.pmtud_pmtu
+            "  pmtud: {} sent {} acked {} lost {} count {} iface_mtu {} pmtu",
+            self.pmtud_tx,
+            self.pmtud_ack,
+            self.pmtud_lost,
+            self.pmtud_count,
```
martinthomson: I'm not sure that I understand from just reading this code (including the formatted output) that "count" refers to the total number of times that PMTUD was restarted.  I would go with either:

* pmtud_restarts
* pmtud_resets

To indicate that this is about the times that you had to start over.

test-fixture/src/sim/mtu.rs
```diff
@@ -31,13 +31,58 @@ impl Node for Mtu {
     fn init(&mut self, _rng: Rng, _now: Instant) {}
 
     fn process(&mut self, d: Option<Datagram>, _now: Instant) -> Output {
-        d.filter(|dgram| {
-            let header = match dgram.destination().ip() {
-                IpAddr::V4(_) => 20 + 8,
-                IpAddr::V6(_) => 40 + 8,
-            };
+        d.filter(|dgram| header_size(dgram.destination().ip()) + dgram.len() <= self.mtu)
```
martinthomson, line 34: The other place you could move this sort of function is to the common crate, where we already have IP ToS (DSCP & ECN).

test-fixture/src/assertions.rs
```diff
@@ -176,7 +176,7 @@ pub fn assert_path(dgram: &Datagram, path_addr: SocketAddr) {
 pub fn assert_v4_path(dgram: &Datagram, padded: bool) {
     assert_path(dgram, DEFAULT_ADDR_V4);
     if padded {
-        assert_eq!(dgram.len(), Pmtud::default_plpmtu(DEFAULT_ADDR_V4.ip()));
+        assert_eq!(dgram.len(), default_plpmtu(DEFAULT_ADDR_V4.ip()));
     }
```
martinthomson, line 180: I love the way that cargo mutants complains about these assert_foo functions.  Of course a passing test suite will never generate an assertion...

Should we be testing the assertion functions?  After how many turtles should we stop?

Is there a way to disable cargo mutants on a per-file basis?  https://mutants.rs/skip_files.html suggests that the answer is a command-line option, which isn't great.

neqo-transport/src/connection/tests/datagram.rs
larseggert: `test-fixture` and `neqo-http3` need some of them, too?

test-fixture/src/sim/mtu.rs
larseggert: My mistake. It's also not exported.

test-fixture/src/assertions.rs
larseggert: https://github.com/mozilla/neqo/issues/3367

test-fixture/src/sim/mtu.rs
larseggert: I thought about it, but then `search_table` also has to move there, because `default_plpmtu` needs it.

neqo-transport/src/recovery/mod.rs
larseggert: We mark the **oldest** packets as lost AFAICT.

neqo-transport/src/pmtud.rs
```diff
@@ -253,17 +299,135 @@ impl Pmtud {
...
+    /// Count of consecutive loss events for large packets (ACK-based detection).
+    loss_count: usize,
+    /// Ignore losses/PTOs of packets sent before this time.
+    /// Set when we restart PMTUD to avoid counting stale events.
+    ignore_before: Option<Instant>,
+}
+
+impl BlackHoleDetector {
+    /// Number of consecutive ACK-based loss events before declaring a black hole.
+    const THRESHOLD: usize = 3;
```
mxinden, line 347: Why does a static constant (here 3) work for both low and high bandwidth-delay connections?

Given a high bandwidth-delay connection and given a sender doing a large upload, say that the sender overshoots the bandwidth by 10 high MTU UDP datagrams, thus causing the loss of those 10 UDP datagrams.

I acknowledge that each call to `on_loss` at most increases `loss_count` by `1`, no matter how many packets have been declared lost in that call. That said, would the above scenario result in one `on_loss` call? If so, what guarantees a single `on_loss` call only? Would this result in potentially multiple `on_loss` calls, if so, wouldn't the `BlackHoleDetector` do a lot of false positive detection?

neqo-transport/src/pmtud.rs
```diff
@@ -182,35 +211,42 @@ impl Pmtud {
...
+        self.raise_timer = Some(now + Self::RAISE_TIMER);
         qinfo!(
             "PMTUD stopped, PLPMTU is now {}, raise timer {:?}",
             self.mtu,
             self.raise_timer
         );
     }
 
-    /// Checks whether a PMTUD probe has been lost. If it has been lost more than `MAX_PROBES`
+    /// Checks whether a PMTUD probe has been lost. If it has been lost more than `Self::MAX_PROBES`
```
mxinden, line 223: Needs updating. Also checks for black holes now.

neqo-transport/src/pmtud.rs
larseggert: I was operating under the assumption it would generate one call, but we should add a test for this.

larseggert:
> Alternative theory that would explain the high packet loss, but is not fixed by the `BlackHoleDetector`:
> 
>     1. The DONTFRAGMENT bit is not properly propagated on IPv4 connections, e.g. a VPN not properly configured.
> 
>     2. PMTUD increases the MTU beyond the path MTU.
> 
>     3. Lower level IP stack fragments.
> 
>     4. Fragmented packets are more likely to be lost intermittently, as the whole is declared lost even if actually only a single fragment went missing.

Why is this not fixed by this patch? If sufficient fragments get lost to trigger the ACK-based case, it will.

neqo-transport/src/pmtud.rs
larseggert: I added `burst_loss_counts_as_single_event`.

mxinden:
> > Alternative theory that would explain the high packet loss, but is not fixed by the `BlackHoleDetector`:
> > ```
> > 1. The DONTFRAGMENT bit is not properly propagated on IPv4 connections, e.g. a VPN not properly configured.
> > 
> > 2. PMTUD increases the MTU beyond the path MTU.
> > 
> > 3. Lower level IP stack fragments.
> > 
> > 4. Fragmented packets are more likely to be lost intermittently, as the whole is declared lost even if actually only a single fragment went missing.
> > ```
> 
> Why is this not fixed by this patch? If sufficient fragments get lost to trigger the ACK-based case, it will.

It would run into an infinite cycle:
1. Start PMTUD.
2. Overshoot path MTU, thus causing fragmentation.
3. Black hole detection kicking in.
4. Go back to (1).

(I might be missing something in the current implementation.)

larseggert:
> It would run into an infinite cycle:

I think you're on to something. How should we deal with this? We could stick to the minimum MTU after a black hole on a connection, or we could tie into the raise timer (10 min) to try again? Anything smarter?

larseggert:
@mxinden I have a suggestion for this in 3762a93baa4a68ae02bbae65d3616742cb99c7b8

martinthomson:
I'm glad to see that the performance tests show that wall clock time regresses with this change.  That's expected, especially since our tests are highly sensitive to PMTU.  It's disappointing to see that the noise on take_ranges continues unabated.

mxinden:
> > > Alternative theory that would explain the high packet loss, but is not fixed by the `BlackHoleDetector`:
> > > ```
> > > 1. The DONTFRAGMENT bit is not properly propagated on IPv4 connections, e.g. a VPN not properly configured.
> > > 
> > > 2. PMTUD increases the MTU beyond the path MTU.
> > > 
> > > 3. Lower level IP stack fragments.
> > > 
> > > 4. Fragmented packets are more likely to be lost intermittently, as the whole is declared lost even if actually only a single fragment went missing.
> > > ```
> > 
> > 
> > Why is this not fixed by this patch? If sufficient fragments get lost to trigger the ACK-based case, it will.
> 
> It would run into an infinite cycle:
> 
>     1. Start PMTUD.
> 
>     2. Overshoot path MTU, thus causing fragmentation.
> 
>     3. Black hole detection kicking in.
> 
>     4. Go back to (1).
> 
> 
> (I might be missing something in the current implementation.)

@larseggert can you explain how to the current version of this pull request addresses the issue above? Say that the actual path MTU is `1380`. Say that the path does not propagate the DONTFRAGMENT bit, thus leading to the above issue. Say that PMTUD "discovered" a MTU of `1500`. Wouldn't your current implementation just fall back to `1472`, which is still wrong?

larseggert:
This is handled - we repeatedly lower the MTU. See the `black_hole_limits_probing` test.

mxinden:
Thanks. That makes sense. If I understand correctly, this will require 3 packet train losses per search table entry (assuming no positive `on_ack` resets the blackhole detection in between). I would expect this to translate into significant delay before arriving at the correct MTU.

I am still against merging here.

- I don't think we understand the actual issue, i.e. the reason for high packet loss on the 75th percentile. Thus we don't know whether this patch actually fixes it.
- I am having a hard time reasoning about the black-hole-detection state machine, finding it difficult to predict how it behaves on the Internet, especially with ACKs and losses interleaved.
- We don't have a signal that PMTUD has a significant positive performance impact on Firefox users.

I think the best next step is to find a setup that actually suffers from the high packet loss we see in telemetry, to understand (one of) the root cause(s).

I don't mean to be the gate keeper here. If you want to proceed, feel free to overrule.

larseggert:
**Claude encountered an error** —— [View job](https://github.com/mozilla/neqo/actions/runs/22488294117)

---

larseggert:
Nothing against finding a reproducer, I just think we won't be able to...

larseggert:
@mxinden what do you think about putting this behind a pref that we can set to Nightly-only, to gather some data?

mxinden:
I suggest testing #3414 in Nightly first. If that doesn't fix the issue, we can revisit this change. I don't think we need an additional pref, as we already have the PMTUD pref, which is currently disabled on all Firefox channels.

mxinden:
> I suggest testing #3414 in Nightly first. If that doesn't fix the issue, we can revisit this change. I don't think we need an additional pref, as we already have the PMTUD pref, which is currently disabled on all Firefox channels.

I suggest not testing this pull request and #3414 together in Firefox Nightly as we couldn't correlate any changes in metrics to a specific pull request. @larseggert do you mind if we revert this pull request for now?

larseggert:
No, go ahead. (Sorry, didn't recall we wanted to land these separately.)

PR #3337, opened 2026-01-20
feat: Remove `bindgen` dependency for MTU crate

larseggert:
Fixes https://github.com/quinn-rs/quinn/pull/2465#discussion_r2597690789

### DIFF
--- a/.github/workflows/check-mtu.yml
+++ b/.github/workflows/check-mtu.yml
@@ -135,10 +135,20 @@ jobs:
 
       - name: Check for binding changes
         id: check
+        env:
+          EVENT_NAME: ${{ github.event_name }}
         run: |
+          MISSING=0
           for p in freebsd linux macos netbsd openbsd solaris; do
-            [ -d "artifacts/bindings-$p" ] || echo "::warning::Missing bindings artifact for $p — upstream job may have failed."
+            if [ ! -d "artifacts/bindings-$p" ]; then
+              echo "::warning::Missing bindings artifact for $p — upstream job may have failed."
+              MISSING=1
+            fi
           done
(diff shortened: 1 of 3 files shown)


.github/workflows/check-mtu.yml
```diff
@@ -12,6 +12,41 @@ permissions:
...
+      - name: Install bindgen-cli
+        run: cargo install bindgen-cli --locked
+      - name: Generate bindings
+        env:
+          OS: ${{ matrix.os }}
+          ARGS: ${{ matrix.args }}
+          HEADER: ${{ matrix.header }}
+        run: |
+          # shellcheck disable=SC2086
+          bindgen $ARGS --generate-cstr --explicit-padding --with-derive-default "mtu/src/bindings/$HEADER" > "$OS.rs"
```
martinthomson: When $ARGS is expanded, do the `|` characters turn into pipes?  It's a real shame that you can't define variables in YAML that are forwarded as as bash arrays properly.  Then you could use `"${ARGS[@]}"` and get proper hygiene.  The other way to manage this is to use:

```sh
read -a args < <(echo "$ARGS")
bindgen "${args[@]}"
```

Which is a horrid kludge, but it avoids the shellcheck warning.

larseggert:
> But then you might also have CI propose a PR to update what is in-tree on the affected branch.

That is probably overkill. I don't really expect these bindings to ever change, since OSs have ABI promises.

Edit: But Rust version changes may cause bindgen to generate slightly different bindings. Let me look into the PR thing.

.github/workflows/check-mtu.yml
larseggert: They don't. There I actually had quotes first, but _that_ broke things.

PR #3338, opened 2026-01-21
Revert "chore: Use derive_more to reduce boilerplate (#3290)"

mxinden:
This reverts commit https://github.com/mozilla/neqo/commit/6c04e92db59bfc0dd739e77686ddb2e2fc7fd794.

The commit introduced `unicode-xid` crate, which Firefox tries to avoid as a dependency. See https://bugzilla.mozilla.org/show_bug.cgi?id=1986247#c8 for details.

We can re-introduce `derive_more` with a fix at a later time. That said, I don't think it should block all other changes in the Neqo `v0.22` release to land in Firefox.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ description = "Neqo, the Mozilla implementation of QUIC in Rust."
 keywords = ["quic", "http3", "neqo", "mozilla", "ietf", "firefox"]
 categories = ["network-programming", "web-programming"]
 readme = "README.md"
-version = "0.22.0"
+version = "0.22.1"
 # Keep in sync with `.rustfmt.toml` `edition`.
 edition = "2021"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ description = "Neqo, the Mozilla implementation of QUIC in Rust."
 keywords = ["quic", "http3", "neqo", "mozilla", "ietf", "firefox"]
 categories = ["network-programming", "web-programming"]
 readme = "README.md"
-version = "0.22.0"
+version = "0.22.1"
```
mxinden, line 24: Note that this pull request bumps the workspace patch version in order to cut a new release right away.

mxinden:
image

Given this being a revert, does it have to adhere to our code coverage rule? If not, @larseggert would you mind merging manually?

mxinden:
Thanks for the quick help!

PR #3339, opened 2026-01-21
chore: Update deps to Gecko versions (v2)

larseggert:
Extend the script that aligns our lockfile with Gecko's. Convert the script from bash to python. 

Alternative to #3315

### DIFF
--- a/.github/workflows/firefox.yml
+++ b/.github/workflows/firefox.yml
@@ -1,10 +1,27 @@
-name: Firefox
+name: Vendor
 on:
+  pull_request:
+    branches: ["main"]
+    types: [opened, synchronize, reopened]
+  merge_group:
   workflow_dispatch:
-  # Don't run this as part of CI anymore; takes too many resources and fails too often.
-  # pull_request:
-  #   branches: ["main"]
-  #   types: [opened, synchronize, reopened, ready_for_review]
+    inputs:
+      build:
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -147,3 +147,14 @@ debug = true
 [profile.mutants]
 inherits = "test"
 debug = "none"
+
+[patch.crates-io]
+backtrace = { path = "build/rust/backtrace" }
+hermit-abi = { path = "build/rust/hermit-abi" }
+r-efi = { path = "build/rust/r-efi" }
```
mxinden: I am not sure I understand the pattern, nor have I seen it anywhere else. What problem is this trying to solve? We didn't depend on `backtrace` before. Why do we need to depend on it (an empty stub really) now?

larseggert: The script is copying the stubs over from Gecko.

larseggert: backtrace seems to come in via tokio 1.45.1. This is the same on Gecko.

larseggert:
@mxinden I figured out a way to do this without copying the stub crates over from Gecko.

larseggert:
@mxinden mind taking another look? Just the lockfile, the python is sort of a work in progress.

larseggert:
Merging this now to get rid of the `cargo audit` CI failures. We can follow-up with any needed changes separately.

PR #3340, opened 2026-01-22
chore: Make CodSpeed show better benchmark names

larseggert:
Should now be identical to what criterion shows.

### DIFF
--- a/neqo-bin/benches/main.rs
+++ b/neqo-bin/benches/main.rs
@@ -17,7 +17,7 @@ use neqo_bin::{client, server};
 use tokio::runtime::Builder;
 
 struct Benchmark {
-    name: String,
+    name: &'static str,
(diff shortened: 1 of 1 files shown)


neqo-bin/benches/main.rs
```diff
@@ -58,13 +58,13 @@ fn transfer(c: &mut Criterion) {
             download_size: 0,
         },
     ] {
-        let mut group = c.benchmark_group(name);
+        let mut group = c.benchmark_group(&name);
```
martinthomson: Looking at this, it seems like we aren't really using `format!()` in at least this test.  Maybe we can switch the name field to `&'static str` from `String`.

larseggert: @martinthomson check the latest commit, is that what you had in mind?

PR #3341, opened 2026-01-22
fix: Remove trailing newline from REQUESTS URLs

o8vm:
### Problem

The interop test runner passes URLs via the `$REQUESTS` environment variable.
When parsing this in `qns/interop.sh`, we use:

```bash
mapfile -d ' ' -t URLS <<<"$REQUESTS"
```

However, bash's here-string (`<<<`) appends a trailing newline to the input.
This causes the last URL to contain a newline character, resulting in:

```
error: invalid value 'https://server46:443/some-file
' for '[URLS]...': invalid uri character
```

This breaks specific interop tests (e.g.,   connectionmigration).

### Solution

Replace `mapfile` with `read -ra` which doesn't have this issue:

```bash
read -ra URLS <<<"$REQUESTS"
```

### Why this works

- `read -ra`: Splits on IFS (whitespace by default). The trailing newline from here-string is consumed as a delimiter.
- `mapfile -d ' '`: Splits only on space. The trailing newline remains attached to the last element.

### Testing

Tested with quic-interop-runner against hibana server:
- Before: `error: invalid value ... invalid uri character`
- After: Test runs successfully

### DIFF
--- a/qns/interop.sh
+++ b/qns/interop.sh
@@ -14,7 +14,7 @@ client)
   /wait-for-it.sh sim:57832 -s -t 30
   OPTIONS=(--cc cubic --qns-test "$TESTCASE" --qlog-dir "$QLOGDIR" --output-dir /downloads)
   if [ "$REQUESTS" ]; then
-    mapfile -d ' ' -t URLS <<<"$REQUESTS"
+    read -ra URLS <<<"$REQUESTS"
     OPTIONS+=("${URLS[@]}")
   fi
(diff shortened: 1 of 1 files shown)


larseggert:
~~Thanks! Do you have a theory why the failure is spurious, i.e., why the error only occurs somtimes?~~

#3291 broke this. The `http` crate is more strict than the `url` crate.

larseggert:
> Tested with quic-interop-runner against hibana server:

Oh, a new server? Is there a docker image we can add to our interop matrix and maybe also https://interop.seemann.io/?

o8vm:
Thanks for the quick review!
hibana is still under private development and not ready for public release yet.
I've verified neqo ↔ neqo connectionmigration test passes with this fix applied.

PR #3343, opened 2026-01-22
 fix(*/Cargo.toml): use nested tables instead of dotted keys 

mxinden:
Change `package.metadata.bench.*.codspeed.mode` from dotted key notation
to nested table notation for compatibility with older TOML parsers.

The dotted key syntax (codspeed.mode = "...") is not properly handled
by some TOML parsers when combined with `[[bench]]` array definitions,
causing "implicitly defined table can't be an array" errors in Firefox's
vendoring process which uses Python's toml library.

See https://bugzilla.mozilla.org/show_bug.cgi?id=2010796#c5 for details.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -49,19 +49,20 @@ jobs:
                 .targets[] |
                 select(.kind[] == "bench") |
                 select(has("required-features")) |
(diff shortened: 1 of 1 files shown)


larseggert (approved):

mxinden:
@larseggert can you override once more given the codecov failure? I am puzzled why codecov fails on a non Rust change.

PR #3347, opened 2026-01-22
chore: Address some easy TODOs

larseggert:
- **neqo-qpack/src/decoder.rs** — Documented specific error types (`Error::Internal`, `Error::DecoderStream`) for `send()` function

- **neqo-transport/src/connection/mod.rs**
  - Replaced stale issue #47 TODO with RFC 9002 A.8 explanation
  - Simplified stale PR #2118 TODO (no longer needed after merge)

- **neqo-transport/src/connection/tests/datagram.rs** — Replaced magic constant 27 with documented `MIN_DATAGRAM_PACKET_OVERHEAD`

- **neqo-transport/src/connection/tests/handshake.rs** — Replaced 2 pacing delay questions with explanations

- **neqo-transport/src/connection/tests/vn.rs** — Replaced 5 pacing delay TODOs with brief explanations

- **neqo-transport/src/connection/tests/idle.rs** — Clarified comment about recovery timeout

- **neqo-http3/src/connection.rs** — Changed error from `InvalidStreamId` to `InvalidState` for inconsistent stream state

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1165,8 +1165,8 @@ impl Connection {
         // drive it.
 
         let earliest = delays.into_iter().min().expect("at least one delay");
-        // Per RFC 9002 A.8, timers set in the past should fire immediately.
-        // `saturating_duration_since` clamps to zero in that case.
+        // TODO(agrover, mt) - need to analyze and fix #47
+        // rather than just clamping to zero here.
         debug_assert!(earliest > now);
         let delay = earliest.saturating_duration_since(now);
         qdebug!("[{self}] delay duration {delay:?}");


neqo-qpack/src/decoder.rs
```diff
@@ -160,7 +160,8 @@ impl Decoder {
 
     /// # Errors
     ///
-    /// May return an error in case of any transport error. TODO: define transport errors.
+    /// May return `Error::Internal` if the decoder stream is not initialized,
+    /// or `Error::DecoderStream` if sending on the decoder stream fails.
```
mxinden:
```suggestion
    /// May return [`Error::Internal`] if the decoder stream is not initialized,
    /// or [`Error::DecoderStream`] if sending on the decoder stream fails.
```

In case we ever rename these, CI will tell us to update these docs.

neqo-transport/src/connection/mod.rs
```diff
@@ -1165,8 +1165,8 @@ impl Connection {
         // drive it.
 
         let earliest = delays.into_iter().min().expect("at least one delay");
-        // TODO(agrover, mt) - need to analyze and fix #47
-        // rather than just clamping to zero here.
+        // Per RFC 9002 A.8, timers set in the past should fire immediately.
+        // `saturating_duration_since` clamps to zero in that case.
```
mxinden: I am not sure this matches the intention of the initial comment. We still see timers from the past, e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=2001901. What does this have to do with RFC 9002?

larseggert: Are you saying it's not the loss recovery timer that's in the past?

mxinden: This is part of our output path. We first try to send a packet. In case that fails, we check whether we can emit a timer, in order for us to be woken up again in the future. If that timer is in the past, that implies that we could be sending something right now, in other words that our premise (failed to send sth) is not correct. I don't think this has anything to do with any RFC, but just our software architecture itself.

larseggert: ACK. Let me just revert this bit.

martinthomson (approved):

larseggert:
**Claude encountered an error** —— [View job](https://github.com/mozilla/neqo/actions/runs/22489015928)

---

PR #3350, opened 2026-01-23
fix: Allow Initial CRYPTO retransmission during Handshake PTO

larseggert:
When the client has lost Initial CRYPTO data and PTO fires for Handshake space (primed to prevent deadlocks), the client must still be able to retransmit the lost Initial CRYPTO frames per RFC 9002 Section 6.2.4.

Previously, `SendProfile::ack_only()` incorrectly returned `true` for Initial space when PTO was for Handshake space, blocking CRYPTO frame transmission. Additionally, Initial packets were not being marked for retransmission when Handshake PTO fired.

This fix:

- Simplifies `ack_only()` to only check congestion window limits, removing the incorrect space-based restriction
- Marks Initial packets for retransmission in `maybe_fire_pto()` when Handshake PTO fires, ensuring lost Initial CRYPTO data is resent
- Removes the redundant `SendProfile::pto` field, using `should_probe()` for equivalent assertions in tests
- Adds unit and integration tests that verify Initial CRYPTO can be retransmitted when Handshake PTO fires
- Fixes `reorder_handshake` test flakiness by disabling packet number randomization for deterministic behavior

Bug scenario from QNS L1/C1 test failures:
1. Client sends `ClientHello` split across Initial packets
2. Server receives first packet but second is lost/corrupted
3. Server ACKs first packet; client detects second as lost
4. PTO fires for Handshake (primed to prevent deadlocks)
5. BUG: `ack_only(Initial)` returned true, blocking CRYPTO retransmission
6. Client times out waiting for handshake completion


neqo-transport/src/recovery/mod.rs
```diff
@@ -902,15 +895,42 @@ impl Loss {
...
+        // Maybe prime the Handshake PTO when PTO fires in Initial space.
+        if pn_space == PacketNumberSpace::Initial {
+            self.maybe_prime_handshake_pto(now);
         }
+
+        // When Handshake PTO fires, also mark Initial packets for retransmission.
+        // This handles the case where Initial space has outstanding CRYPTO data
+        // but its PTO timer hasn't fired yet (because we keep sending Initial
+        // packets which push `last_ack_eliciting` forward). Without this, the
+        // client would send PING-only probes instead of retransmitting CRYPTO.
```
mxinden: At this point `lost` might not be empty, right? In other words, it would not be a PING-only probe.

neqo-transport/src/recovery/mod.rs
```diff
@@ -902,15 +895,42 @@ impl Loss {
...
+        if pn_space != PacketNumberSpace::Handshake {
+            return;
+        }
+
+        let Some(initial_space) = self.spaces.get_mut(PacketNumberSpace::Initial) else {
+            return;
+        };
+
+        let mtu = primary_path.borrow().plpmtu();
+        let mut size = 0;
```
mxinden: We might have added packets to `lost` in the loop above. Starting with `0` here does not uphold `MAX_PTO_PACKET_COUNT * mtu` across the entire function. Is that by design?

neqo-transport/src/recovery/mod.rs
```diff
@@ -902,15 +895,42 @@ impl Loss {
...
+        lost.extend(
```
mxinden: Instead of this bolt-on, first extending `lost` in the loop above (potentially including initial space), then this special case again for the initial space with various caveats above, how about restructuring the whole function into two phases.

1. Determine in which spaces to retransmit:
    ```` rust
    let mut retransmit = PacketNumberSpaceSet::default();

    for pn_space in PacketNumberSpace::iter() {
        let Some(t) = self.pto_time(primary_path.borrow().rtt(), pn_space) else {
            continue;
        };
        allow_probes.insert(pn_space);
        if t > now {
            continue;
        }
        qdebug!("[{self}] PTO timer fired for {pn_space:?}");
        retransmit.insert(pn_space);
        // When Handshake PTO fires, also retransmit Initial CRYPTO data.
        // This handles lost Initial CRYPTO that hasn't triggered its own PTO yet
        // because last_ack_eliciting keeps advancing with each new Initial send.
        if pn_space == PacketNumberSpace::Handshake {
            retransmit.insert(PacketNumberSpace::Initial);
        }
        pto_space = pto_space.or(Some(pn_space));
    }
    ```
2. Do the retransmission, i.e. extending `lost`:
    ``` rust
    let Some(pn_space) = pto_space else {
        return;
    };

    let mtu = primary_path.borrow().plpmtu();
    for space in PacketNumberSpace::iter().filter(|s| retransmit.contains(*s)) {
        let Some(s) = self.spaces.get_mut(space) else {
            continue;
        };
        let mut size = 0;
        lost.extend(
            s.pto_packets()
                .take_while(|p| {
                    size += p.len();
                    size <= MAX_PTO_PACKET_COUNT * mtu
                })
                .cloned(),
        );
    }

    if pn_space == PacketNumberSpace::Initial {
        self.maybe_prime_handshake_pto(now);
    }
```

mxinden (commented):
This makes sense to me. Thanks for the elaborate test. I think we can simplify `maybe_fire_pto`. What do you think?

neqo-transport/src/recovery/mod.rs
larseggert: Yup. I'll rephrase.

neqo-transport/src/recovery/mod.rs
larseggert: Good catch. Let me think what would be best.

larseggert: I moved it up.

larseggert:
**Claude encountered an error** —— [View job](https://github.com/mozilla/neqo/actions/runs/22486115561)

---

PR #3351, opened 2026-01-23
fix: resend HANDSHAKE_DONE when server receives undecryptable Handshake

larseggert:
When the server is in Confirmed state and receives Handshake packets it cannot decrypt (because Handshake keys were discarded per RFC 9001 Section 4.9.2), it now triggers resending `HANDSHAKE_DONE` immediately.

Previously, the server would simply drop these packets and wait for its PTO timer to fire before retransmitting, which could take 9+ seconds. This caused unnecessary delays in completing the handshake under packet loss conditions.

The fix mirrors the existing client-side logic that resends Initial CRYPTO when receiving undecryptable Handshake/Short packets.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -495,10 +495,7 @@ impl<'a, H: Handler> Runner<'a, H> {
     }
 
     async fn process_multiple_input(&mut self) -> Res<()> {
-        loop {
-            let Some(dgrams) = self.socket.recv(self.local_addr, &mut self.recv_buf)? else {
-                break;
-            };
+        while let Some(dgrams) = self.socket.recv(self.local_addr, &mut self.recv_buf)? {
             self.client.process_multiple_input(dgrams, Instant::now());
             self.process_output().await?;
         }
--- a/neqo-transport/src/connection/tests/idle.rs
+++ b/neqo-transport/src/connection/tests/idle.rs
@@ -681,7 +681,7 @@ fn keep_alive_with_ack_eliciting_packet_lost() {
     //    IDLE_TIMEOUT / 2)
     //  - PTO timer will trigger again. (at the start time + pto + 2*pto)
     //  - Idle time out  will trigger (at the timeout + IDLE_TIMEOUT)
-    const IDLE_TIMEOUT: Duration = Duration::from_millis(6000);
+    const IDLE_TIMEOUT: Duration = Duration::from_secs(6);
 
     // This test makes too many assumptions about single-packet flights and PTOs for multi-packet
     // MLKEM flights to work.
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -91,7 +91,7 @@ impl Pacer {
             u128::try_from(packet - self.c).expect("packet is larger than current credit");
         let d = r.saturating_mul(deficit);
         let add = d / u128::try_from(cwnd * Self::SPEEDUP).expect("usize fits into u128");
-        let w = u64::try_from(add).map(Duration::from_nanos).unwrap_or(rtt);
+        let w = u64::try_from(add).map_or(rtt, Duration::from_nanos);
 
         // If the increment is below the timer granularity, send immediately.
         if w < GRANULARITY {
@@ -161,7 +161,7 @@ mod tests {
 
     use super::Pacer;
 
-    const RTT: Duration = Duration::from_millis(1000);
+    const RTT: Duration = Duration::from_secs(1);
     const PACKET: usize = 1000;
     const CWND: usize = PACKET * 10;
 
--- a/test-fixture/src/lib.rs
+++ b/test-fixture/src/lib.rs
@@ -72,7 +72,7 @@ pub fn fixture_init() {
 // This needs to be > 2ms to avoid it being rounded to zero.
 // NSS operates in milliseconds and halves any value it is provided.
 // But make it a second, so that tests with reasonable RTTs don't fail.
-pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_millis(1000);
+pub const ANTI_REPLAY_WINDOW: Duration = Duration::from_secs(1);
 
 /// A baseline time for all tests.  This needs to be earlier than what `now()` produces
 /// because of the need to have a span of time elapse for anti-replay purposes.


neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // Server receives client Finished, enters Confirmed state.
+    // Server sends Handshake ACK + 1-RTT `HANDSHAKE_DONE`.
+    let handshake_done_before = server.stats().frame_tx.handshake_done;
+    let s_hs_done = server.process(client_finished, now).dgram();
+    assert!(s_hs_done.is_some());
+    assert_eq!(*server.state(), State::Confirmed);
+    assert_eq!(
+        server.stats().frame_tx.handshake_done,
+        handshake_done_before + 1
```
martinthomson: This should just be 1 here.  No reason to check the before.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    let mut client = new_client(params.clone());
+    let mut server = new_server(params);
+
+    // Complete the handshake up to the point where client is Connected.
+    let pkt = client.process_output(now).dgram();
+    let pkt2 = client.process_output(now).dgram();
+
+    now += HALF_RTT;
+    server.process_input(pkt.unwrap(), now);
+    let pkt = server.process(pkt2, now).dgram();
```
martinthomson, line 1056: Are you banking on the server flight being two packets?  Because that seems to be the assumption behind needing another couple of rounds of back and forth on the handshake.

Could you instead wrap this into a loop until the client gets the AuthNeeded event, then fulfill that and send Finished? That would mean wasting round trips whenever the flight is more than one packet, but it might be easier to read and less fragile.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    assert_eq!(*server.state(), State::Confirmed);
+    assert_eq!(
+        server.stats().frame_tx.handshake_done,
+        handshake_done_before + 1
+    );
+
+    // Verify the server has no additional immediate output.
+    assert!(server.process_output(now).dgram().is_none());
+
+    // DROP the server's response (simulating network loss).
```
martinthomson: I know the explicit drop is nice documentation, but you could just not use the variable.  If you drop it like this, that's fine, but you don't need to have a comment.


```suggestion
```

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    // Client never receives the ACK or `HANDSHAKE_DONE`.
+    drop(s_hs_done);
+
+    // Client is still in Connected state, waiting for `HANDSHAKE_DONE`.
+    assert_eq!(*client.state(), State::Connected);
```
martinthomson: This is pointless.  The client hasn't done anything since this was last checked.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // Save the time when server entered Confirmed state.
+    let server_confirmed_time = now;
```
martinthomson, line 1076: Move this up to when you check for confirmed and drop the comment.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // Get the client's PTO so we can generate a Handshake retransmission.
```
martinthomson: There are too many comments here.  The comment only needs to be "retransmit the Finished", on top of the callback(), now+=, process_output, assert(is_some())

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    let c_hs_retrans = client.process_output(now).dgram();
+    assert!(c_hs_retrans.is_some());
+
+    // The retransmission is coalesced (Handshake + 1-RTT).
+    // Split it and only send the Handshake portion to the server.
+    let (c_hs_only, _) = split_datagram(c_hs_retrans.as_ref().unwrap());
+    assert!(is_handshake(&c_hs_only));
+
+    // Server receives the Handshake retransmission shortly after it entered
+    // Confirmed state (before its own timer would fire).
```
martinthomson: Which timer?  PTO?

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    // Server has discarded Handshake keys, so it can't decrypt.
+    // Server SHOULD resend its 1-RTT data (`HANDSHAKE_DONE`).
+    // Use a very small delay to ensure no server timer fires.
+    let receive_time = server_confirmed_time + Duration::from_millis(1);
```
martinthomson: You should note that this is *back in time* relative to the time the client was given.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    let handshake_done_before = server.stats().frame_tx.handshake_done;
```
martinthomson: Again, you don't need to check the before stats for handshake_done.  (You will probably want the drop count though, so this is mostly OK.)

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    let dropped_before = server.stats().dropped_rx;
+
+    // Use process_input to receive the packet without triggering timers.
+    server.process_input(c_hs_only, receive_time);
+
+    // Verify the server dropped the undecryptable Handshake packet.
```
martinthomson:
```suggestion
```
E_TOO_MANY_COMMENTS

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+    assert!(
+        server.stats().dropped_rx > dropped_before,
+        "Server should have dropped the undecryptable Handshake packet"
+    );
```
martinthomson: It's a single packet in, so shouldn't you check before + 1 == after?

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1031,3 +1031,123 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // BUG: Server should resend 1-RTT data (`HANDSHAKE_DONE`) but currently doesn't.
+    // These assertions demonstrate the bug - they should pass after the fix.
+    assert!(
+        s_response.is_some(),
+        "Server should send 1-RTT response when receiving undecryptable Handshake"
+    );
+    assert_eq!(
+        server.stats().frame_tx.handshake_done,
+        handshake_done_before + 1,
```
martinthomson:
```suggestion
        2,
```

neqo-transport/src/connection/tests/recovery.rs
larseggert: I tried to avoid loops, since they can mask breakage if we make changes later.

neqo-transport/src/connection/tests/recovery.rs
larseggert: Yeah. I made if fix other relative checks, but I missed this one.

neqo-transport/src/connection/tests/recovery.rs
martinthomson: I can see that, but you have already made this brittle.  The full handshake should only take three flights to get to the point we're talking about, but this has 5.  That suggests that you already have a problem.  I'm guessing that it is because the server flight is two messages.  Sometimes.

So you are already masking breakage.  But you don't need to.  I think that the event is a more robust way to structure this test.

larseggert: See https://github.com/mozilla/neqo/pull/3366

PR #3353, opened 2026-01-24
chore: Fix new clippy warnings on nightly

larseggert:


larseggert:
Merging this now so that other PRs don't fail clippy.

PR #3364, opened 2026-01-26
feat: Run perfcompare on CodSpeed

larseggert:


larseggert:
Merging this now. We'll only be able to figure out if this works for us on `main`.

PR #3373, opened 2026-01-29
feat: Investigate using object pools

larseggert:
WIP, no idea if this will help performance.


larseggert:
No benefit whatsoever.

PR #3374, opened 2026-01-29
chore: Simplify NSS minimum version determination

larseggert:


.github/actions/nss/action.yml
```diff
@@ -3,15 +3,36 @@ description: Install NSS
...
   minimum-version:
-    description: "Minimum required version of NSS"
-    required: true
+    description: "Minimum required version of NSS. If not provided, reads from version-file."
+    required: false
+    default: ""
+  version-file:
+    description: "Path to file containing the minimum NSS version (relative to workspace root)"
+    required: false
+    default: "neqo-crypto/min_version.txt"
```
martinthomson, line 12: Do you really need this when you have `minimum-version`?

larseggert: Possibly not. It may be convenient to not do the `cat`, but probably minor.

PR #3375, opened 2026-01-30
fix: Re-enable Firefox vendoring validation in CI

larseggert:
The Firefox workflow was disabled because builds consumed too many resources and frequently failed. This change re-enables it with a different approach: CI validates that neqo can vendor into Gecko, but actually building Firefox is best-effort and non-blocking.

This ensures we catch vendoring issues early while not blocking PRs on resource-intensive Firefox builds that GitHub may terminate.

### DIFF
--- a/.github/workflows/firefox.yml
+++ b/.github/workflows/firefox.yml
@@ -113,6 +113,10 @@ jobs:
             echo ""
           done >> supply-chain/audits.toml
 
+          # Hide .git to prevent mach from running git operations
+          mv .git .git.bak
+          trap 'mv .git.bak .git' EXIT
+
           if ./mach vendor rust --ignore-modified 2>&1 | tee vendor.log; then
             echo "Vendoring succeeded"
             exit 0


.github/workflows/firefox.yml
```diff
@@ -13,192 +30,167 @@ concurrency:
...
+            echo "who = \"CI\""
+            echo "criteria = \"safe-to-deploy\""
+            echo "version = \"$version\""
+            echo "notes = \"Placeholder created by CI.\""
+            echo ""
+          done >> supply-chain/audits.toml
+
+          # Hide .git to prevent mach from running git operations
+          mv .git .git.bak
+          trap 'mv .git.bak .git' EXIT
```
mxinden, line 118: What does `mach` do here? Can this be disabled via a flag instead?

.github/workflows/firefox.yml
```diff
@@ -13,192 +30,167 @@ concurrency:
...
-            echo 'criteria = "safe-to-deploy"'
-            echo "version = \"$QLOG_VERSION\""
-            echo '[[audits.mtu]]'
-            echo 'who = "CI"'
-            echo 'criteria = "safe-to-deploy"'
-            echo "version = \"$MTU_VERSION\""
-          } >> supply-chain/audits.toml
-          sed -i'' -e "s/qlog =.*/qlog = \"$QLOG_VERSION\"/" netwerk/socket/neqo_glue/Cargo.toml
+            echo "ac_add_options --disable-tests"
+            echo "ac_add_options --enable-release"
```
mxinden, line 90: Are you building in debug mode? I usually set:

```
ac_add_options --enable-debug
ac_add_options --enable-debug-symbols
```

.github/workflows/firefox.yml
larseggert: `mach` tries to `git add` the vendored files, and that fails on Windows. I don't see an option to prevent that.

.github/workflows/firefox.yml
larseggert: I thought the main use case here would be to download the artifact to run locally, and debug builds are slow.

.github/workflows/firefox.yml
mxinden: Oh, why does it fail on Windows? Not a blocker.

.github/workflows/firefox.yml
mxinden: Ah, that makes sense. It is a trade off. Your call.

.github/workflows/firefox.yml
larseggert: Command line longer than 32K :-)

PR #3382, opened 2026-02-03
Small update to set_max_stream_data

jesup:
patch for Issue 3381

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -1189,9 +1189,10 @@ impl SendStream {
 
     pub fn set_max_stream_data(&mut self, limit: u64) {
         qdebug!("setting max_stream_data to {limit}");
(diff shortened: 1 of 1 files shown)


martinthomson:
Two things:
1. Can you split out the change?
2. Can you produce a test case that would have failed prior to the fix?  (Ideally, this operates at the connection level, but I understand that this might need to just work on internal objects.)

larseggert:
There is a **ton** of copy&paste repetition in the test code. Can we make this more DRY?

larseggert:
@jesup is the key export stuff here addressing https://bugzilla.mozilla.org/show_bug.cgi?id=2007201?

jesup:
Yes, the key export stuff is for bug 2007201.   

Sorry, I accidentally included the entire patch stack here.  I'll fix it.

jesup:
I'll abandon this and put up proper PRs

PR #3384, opened 2026-02-04
fix: Server should reject Retry w/short DCID

larseggert:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1997780

### DIFF
--- a/neqo-transport/src/server.rs
+++ b/neqo-transport/src/server.rs
@@ -251,7 +251,10 @@ impl Server {
                 //
                 // <https://www.rfc-editor.org/rfc/rfc9000.html#section-7.2>
                 if initial.dst_cid.len() < 8 {
-                    qerror!("[{self}] DCID too short ({} bytes), dropping packet", initial.dst_cid.len());
+                    qerror!(
+                        "[{self}] DCID too short ({} bytes), dropping packet",
+                        initial.dst_cid.len()
+                    );
                     return Output::None;
                 }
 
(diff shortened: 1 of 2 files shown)


neqo-transport/src/server.rs
```diff
@@ -247,6 +247,12 @@ impl Server {
             AddressValidationResult::Validate => {
                 qinfo!("[{self}] Send retry for {:?}", initial.dst_cid);
 
+                // RFC 9000: Initial packets MUST use a DCID of at least 8 bytes.
+                if initial.dst_cid.len() < 8 {
+                    qinfo!("[{self}] DCID too short ({} bytes), dropping packet", initial.dst_cid.len());
```
mxinden: All other logs here are `qerror`. Why use `qinfo` here?

neqo-transport/src/server.rs
```diff
@@ -247,6 +247,12 @@ impl Server {
             AddressValidationResult::Validate => {
                 qinfo!("[{self}] Send retry for {:?}", initial.dst_cid);
 
+                // RFC 9000: Initial packets MUST use a DCID of at least 8 bytes.
```
mxinden:
```suggestion
                // > This Destination Connection ID MUST be at least 8 bytes in length.
                //
                // <https://www.rfc-editor.org/rfc/rfc9000.html#section-7.2>
```

Feel free to ignore.

PR #3385, opened 2026-02-04
fix: Reject invalid extended CONNECT requests

larseggert:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1997780


neqo-http3/src/server_connection_events.rs
```diff
@@ -293,4 +314,35 @@ mod tests {
...
+        assert_eq!(rejected, vec![(StreamId::new(0), 501)]);
+    }
+
+    #[test]
+    fn extended_connect_missing_protocol() {
+        let events = Http3ServerConnEvents::default();
+        let headers = vec![
+            Header::new(":method", "CONNECT"),
+            Header::new(":authority", "example.com"),
+            Header::new(":path", "/"),
```
martinthomson, line 350: This is *almost* a valid request.  The inclusion of `:path` tips it over into being invalid, but a bare CONNECT is valid.  We don't support CONNECT, which is fine, but there are two cases to handle:

A CONNECT that is valid and unsupported: see https://datatracker.ietf.org/doc/html/rfc9114#section-4.4

A CONNECT that is not valid, because it includes `:path` or `:scheme`, for instance.

So I think your switch statement needs a tiny bit more sophistication.

neqo-http3/src/server_connection_events.rs
```diff
@@ -159,11 +161,22 @@ impl HttpRecvStreamEvents for Http3ServerConnEvents {
...
             }
-            Some(_) => {
-                unimplemented!("Extended connect other than webtransport or connect-udp")
+            Some(protocol) => {
+                // RFC 9220: Server SHOULD respond with 501 (Not Implemented)
+                qwarn!(
+                    "Unsupported extended CONNECT protocol: {:?}",
+                    String::from_utf8_lossy(protocol)
+                );
+                self.rejected_extended_connects
```
martinthomson: Consider naming this "rejected_connects" instead.

neqo-http3/src/server_connection_events.rs
```diff
@@ -159,11 +161,22 @@ impl HttpRecvStreamEvents for Http3ServerConnEvents {
...
+                    .borrow_mut()
+                    .push((stream_id, 501));
             }
             None => {
-                unimplemented!("connect without :protocol header");
+                // Missing :protocol - malformed request
```
martinthomson: Incorrect comment, see below.

neqo-http3/src/connection_server.rs
```diff
@@ -287,11 +287,22 @@ impl Http3ServerHandler {
...
         let res = self.check_connection_events(conn, now);
         if !self.check_result(conn, now, &res) && self.base_handler.state().active() {
+            self.handle_rejected_extended_connects(conn, now);
             let res = self.base_handler.process_sending(conn, now);
             self.check_result(conn, now, &res);
         }
     }
 
+    /// Handle rejected Extended CONNECT requests by sending automatic error responses.
+    fn handle_rejected_extended_connects(&mut self, conn: &mut Connection, now: Instant) {
```
martinthomson:
```suggestion
    fn handle_rejected_connects(&mut self, conn: &mut Connection, now: Instant) {
```

martinthomson (commented):
This is probably largely academic, but your focus on extended connect forgets the not-extended version.

neqo-http3/src/server_connection_events.rs
```diff
@@ -88,6 +88,8 @@ pub enum ConnectUdpEvent {
 #[derive(Debug, Default, Clone)]
 pub struct Http3ServerConnEvents {
     events: Rc<RefCell<VecDeque<Http3ServerConnEvent>>>,
+    /// Rejected Extended CONNECT requests: (`stream_id`, `status_code`).
+    rejected_extended_connects: Rc<RefCell<Vec<(StreamId, u16)>>>,
```
mxinden: Is the asynchronicity required here? Why split it into detection and rejection? Why not synchronously reject right after detection?

https://github.com/larseggert/neqo/blob/c0027faa9b344635e35f9ab27df268b4c61d3a4d/neqo-http3/src/recv_message.rs#L189-L190

mxinden (commented):
Do we need this server functionality for a test? If not, is this fix worth the effort and complexity?

neqo-http3/src/server_connection_events.rs
larseggert: `HttpRecvStreamEvents` trait doesn't pass `Connection` - should it?

neqo-http3/src/server_connection_events.rs
```diff
@@ -88,6 +88,10 @@ pub enum ConnectUdpEvent {
 #[derive(Debug, Default, Clone)]
 pub struct Http3ServerConnEvents {
     events: Rc<RefCell<VecDeque<Http3ServerConnEvent>>>,
+    /// Rejected CONNECT requests pending response: (`stream_id`, `http_status_code`).
+    /// Deferred because the `HttpRecvStreamEvents` trait doesn't pass `Connection`;
+    /// processed later in `Http3ServerHandler::handle_rejected_connects`.
+    rejected_connects: Rc<RefCell<Vec<(StreamId, u16)>>>,
```
martinthomson: Why did you decide not to add an event for this?  I'm not seeing a reason.

neqo-http3/src/connection_server.rs
```diff
@@ -289,11 +289,22 @@ impl Http3ServerHandler {
...
             let res = self.base_handler.process_sending(conn, now);
             self.check_result(conn, now, &res);
         }
     }
 
+    /// Handle rejected CONNECT requests by sending automatic error responses.
+    fn handle_rejected_connects(&mut self, conn: &mut Connection, now: Instant) {
+        for (stream_id, status) in self.events.take_rejected_connects() {
+            let headers = [Header::new(":status", status.to_string())];
+            if self.send_headers(stream_id, &headers, conn).is_ok() {
```
martinthomson: If this send doesn't work, you aren't going to try again.  The stream is going to stick around indefinitely.  That's not good.

You can test this by ensuring that flow control prevents the sending of a response.

neqo-http3/src/connection_server.rs
```diff
@@ -289,11 +289,22 @@ impl Http3ServerHandler {
...
+                _ = self.stream_close_send(stream_id, conn, now);
+            }
+        }
```
martinthomson: The missing mutants notice here is worth paying attention to.  You have a test for the events portion of the addition, but you don't have an end-to-end test for this.

martinthomson (changes requested):
These three comments are each separately problematic.

neqo-http3/src/server.rs
```diff
@@ -317,6 +317,23 @@ impl Http3Server {
...
+                    Http3ServerConnEvent::RejectedConnect { stream_id, status } => {
+                        let headers = [Header::new(":status", status.to_string())];
+                        let conn = &mut conn.borrow_mut();
+                        if handler_borrowed
+                            .send_headers(stream_id, &headers, conn)
+                            .is_ok()
+                        {
+                            _ = handler_borrowed.stream_close_send(stream_id, conn, now);
+                        } else {
+                            // Ensure the stream doesn't linger if we can't send the response.
```
martinthomson, line 329: I like this.  It's a very sensible way to avoid sitting on requests.

neqo-http3/src/server.rs
```diff
@@ -1097,6 +1147,44 @@ mod tests {
...
+        let (mut hconn, mut peer_conn) = connect();
+
+        let stream_id = peer_conn.stream_create(StreamType::BiDi).unwrap();
+        peer_conn.stream_send(stream_id, REQUEST_WITH_BODY).unwrap();
+        peer_conn.stream_close_send(stream_id).unwrap();
+
+        // Client sends STOP_SENDING in the same flight as the request.
+        peer_conn
+            .stream_stop_sending(stream_id, Error::HttpNone.code())
+            .unwrap();
```
martinthomson, line 1167: What happens if we change our code such that STOP_SENDING ends up causing other frames to not be sent?  I mean, that's something we might want to consider doing, isn't it?

(If we had RESET_STREAM_AT, you could `commit()`, which would make the data you sent relevant.  You wouldn't necessarily guarantee that the fin flag makes it through. The write would, so this is probably OK.)

The alternative approach here is to write the STOP_SENDING onto the wire directly, using the test frame writer.

neqo-http3/src/server.rs
```diff
@@ -1394,4 +1482,95 @@ mod tests {
...
+
+    // Client SETTINGS with ENABLE_CONNECT_PROTOCOL=1 appended.
+    const SETTINGS_WITH_CONNECT: &[u8] =
+        &[0x0, 0x4, 0x08, 0x1, 0x40, 0x64, 0x7, 0x40, 0x64, 0x08, 0x01];
+
+    /// Set up a server and peer with an extended CONNECT request using an
+    /// unknown `:protocol` queued on the stream. Packets have NOT been exchanged.
+    fn setup_rejected_extended_connect() -> (Http3Server, PeerConnection, StreamId) {
+        let mut server = create_server(http3params(DEFAULT_SETTINGS).connect(true));
+        let (mut peer, _token, mut encoder) =
```
martinthomson: Is "peer" a client?  Can it be called that?

neqo-http3/src/server.rs
```diff
@@ -1394,4 +1482,95 @@ mod tests {
...
+                got_response = true;
+            }
+        }
+        assert!(got_response, "client did not receive the 501 response");
+    }
+
+    /// When the client sends `STOP_SENDING` before the server processes the
+    /// rejection, `send_headers` fails and the server falls back to `cancel_fetch`.
+    #[test]
+    fn rejected_extended_connect_fallback_on_stop_sending() {
```
martinthomson, line 1570: I *think* that there is a missing test: where the server has to reject the stream rather than send a response.

martinthomson (commented):
Getting there.  Just a potential tweak to one test (for robustness) and a missing test.  Other than that, I think the design looks good.

neqo-http3/src/server.rs
larseggert: I'm not quite sure I understand what you mean here. I added a check that the client receives a `RESET_STREAM`; is this what you meant?

neqo-http3/src/server.rs
larseggert: The frame writer is AFAICT not easily accessible in `neqo-http3` without some `cargo` feature additions. Worth it?

(@mxinden started implementing `RESET_STREAM_AT` in #3102 - should I leave a TODO` to revisit this when we land this?)

neqo-http3/src/server.rs
martinthomson: Yeah.  For that you probably need to configure the client with a very small flow control buffer.

neqo-http3/src/server.rs
martinthomson: Waiting on RESET_STREAM_AT seems like a good plan.

larseggert:
Blocked until #3102 lands, see [comment above.](https://github.com/mozilla/neqo/pull/3385#discussion_r2810042568)

PR #3386, opened 2026-02-04
chore(transport/stats): expose cwnd stats

omansfeld:
This adds the following stats:

- `slow_start_exit_cwnd` -> cwnd after slow start was exited
- `cwnd` -> current cwnd (equals cwnd at connection end if called from glean in glue code)
- also added tests to enforce that those are recorded correctly

As a drive-by change:

- moved the `debug()` test for `Stats::default()` from `ecn.rs` to `stats.rs`

Plan is to record them as glean metrics once released into ff. This gives us a general insight into how those numbers look in the wild and also allows to make conclusions about slow start exit point accuracy if they are recorded as a per connection difference.

Part of #3112

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -657,6 +657,7 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
         );
         self.current = stored;
 
+        // If we are restoring back to slow start then we should undo the stat recording.
         if self.current.phase.in_slow_start() {
             cc_stats.slow_start_exit_cwnd = None;
         }
@@ -786,6 +787,8 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
 
         cc_stats.congestion_events[congestion_event] += 1;
         cc_stats.cwnd = self.current.congestion_window;
+        // If we were in slow start when `on_congestion_event` was called we will exit slow start
+        // and should record the exit congestion window.
         if self.current.phase.in_slow_start() {
             cc_stats.slow_start_exit_cwnd = Some(self.current.congestion_window);
         }


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -777,7 +786,11 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
         );
 
         cc_stats.congestion_events[congestion_event] += 1;
-        cc_stats.slow_start_exited |= self.current.phase.in_slow_start();
+        cc_stats.cwnd = self.current.congestion_window;
+        if self.current.phase.in_slow_start() {
+            cc_stats.slow_start_exited = true;
+            cc_stats.slow_start_exit_cwnd = Some(self.current.congestion_window);
+        }
```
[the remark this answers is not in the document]
omansfeld: The cwnd reduction has already been applied at this point, so this does in fact record cwnd AFTER reduction as it is supposed to. Copilot seems very confused here.

neqo-transport/src/stats.rs
```diff
@@ -146,6 +146,13 @@ pub struct CongestionControlStats {
     pub congestion_events: EnumMap<CongestionEvent, usize>,
     /// Whether this connection has exited slow start.
     pub slow_start_exited: bool,
+    /// The congestion window size (in bytes) when we exited slow start.
+    /// None if we haven't exited slow start or if we re-entered after spurious congestion.
+    /// When exiting via congestion event, this is the cwnd AFTER the reduction.
+    pub slow_start_exit_cwnd: Option<usize>,
```
larseggert, line 150: With this addition, do we still need a separate `slow_start_exited`? Because if `slow_start_exit_cwnd` is not `None`, that'd imply that `slow_start_exited` should be `true`, no?

neqo-transport/src/stats.rs
```diff
@@ -382,14 +389,20 @@ impl Debug for Stats {
             "  tx: {} lost {} lateack {} ptoack {} unackdrop {}",
             self.packets_tx, self.lost, self.late_ack, self.pto_ack, self.unacked_range_dropped
         )?;
+        writeln!(f, "  cc:")?;
         writeln!(
             f,
-            "  cc: ce_loss {} ce_ecn {} ce_spurious {}",
+            "    ce_loss {} ce_ecn {} ce_spurious {} final_cwnd {}",
```
larseggert: Removal of `cc:` is a spurious change?

neqo-transport/src/stats.rs
omansfeld: That's correct, I'll remove it. That would be a breaking change with the next neqo release then, as we're already collecting `slow_start_exited` into glean. Just to have in mind.

neqo-transport/src/stats.rs
omansfeld: No, I moved it into an extra line grouping all cc stats beneath it.

```
  cc:
    ce_loss 0 ce_ecn 0 ce_spurious 0 final_cwnd 0
    ss_exit false exit_cwnd None
```

neqo-transport/src/stats.rs
larseggert: You can keep the clean stat but feed it based on the "not None" check.

mxinden: > That's correct, I'll remove it. That would be a breaking change with the next neqo release then, as we're already collecting `slow_start_exited` into glean. Just to have in mind.

`neqo-glue` is the only known consumer today. We bump the minor version (pre 1.0) on each release. Thus breaking changes in Neqo are not a concern.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -777,7 +786,11 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
         );
 
         cc_stats.congestion_events[congestion_event] += 1;
-        cc_stats.slow_start_exited |= self.current.phase.in_slow_start();
+        cc_stats.cwnd = self.current.congestion_window;
+        if self.current.phase.in_slow_start() {
+            cc_stats.slow_start_exited = true;
```
```suggestion
            cc_stats.slow_start_exited = false;
```

Given that we revert back to slow start, shouldn't this be `false`?

omansfeld: This is where we exit slow start in `on_congestion_event`, so it should be set to `true` here. The code for reverting back to slow start (where we would set this to `false` and the new `slow_start_exit_cwnd` to `None`) is in `on_spurious_congestion_event`.

https://github.com/omansfeld/neqo/blob/783e8d443ea4e26f3fc675ea7aeea1ebfc1c2e95/neqo-transport/src/cc/classic_cc.rs#L660-L663

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -777,7 +785,10 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
         );
 
         cc_stats.congestion_events[congestion_event] += 1;
-        cc_stats.slow_start_exited |= self.current.phase.in_slow_start();
+        cc_stats.cwnd = self.current.congestion_window;
+        if self.current.phase.in_slow_start() {
+            cc_stats.slow_start_exit_cwnd = Some(self.current.congestion_window);
+        }
```
mxinden, line 794: It confused me, that we first check whether we are in slow start, and if so, then update the stat that we exited slow start. How about adding a comment?

PR #3389, opened 2026-02-09
feat: Add `neqo_udp::Socket::send_buffer()`

larseggert:
This allows sending from a `&[u8]` without copying into a `Vec<u8>`. (I have a pending Gecko patch to use `neqo-udp` for WebRTC where this avoids copying.)


neqo-udp/src/lib.rs
```diff
@@ -284,6 +305,29 @@ mod tests {
...
     }
 
+    fn recv_segments(
+        receiver: &Socket<std::net::UdpSocket>,
+        expected_count: usize,
+        expected_segment_size: usize,
+    ) {
+        let receiver_addr = receiver.inner.local_addr().unwrap();
+        let mut num_received = 0;
+        let mut recv_buf = RecvBuf::default();
```
martinthomson: Do you really want the allocation to be performed by this code, rather than having a buffer provided?

neqo-udp/src/lib.rs
```diff
@@ -284,6 +305,29 @@ mod tests {
         Ok(socket)
     }
 
+    fn recv_segments(
```
martinthomson, line 318: This addition doesn't seem to be used in the change, nor is it `pub`, and it isn't tested.  What's going on?

mxinden:
> I have a pending Gecko patch to use neqo-udp for WebRTC where this avoids copying.

Can you share the patch?

I suggest not merging here until the Firefox patch is in a reviewable state to reduce churn on the Neqo side.

larseggert:
> Can you share the patch?

https://github.com/larseggert/firefox/tree/feat-webrtc-neqo-udp

larseggert:
I'll leave this as draft per @mxinden's suggestion to first stabilize the Gecko parts.

PR #3390, opened 2026-02-09
feat: Add `datagram::Batch::extend_from_slice()`

larseggert:
Makes WebRTC reuse of `datagram::Batch` for GSO easier.


neqo-common/src/datagram.rs
```diff
@@ -256,6 +256,10 @@ impl Batch {
         })
     }
```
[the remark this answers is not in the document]
martinthomson: In addition to the suggestion from Copilot here, I think that you want to be very careful about the potential for a latent bug.

The use of `slice::chunks_mut` in the iterator creates a situation where you can have the last datagram in a batch be a different size from the others.  Appending bytes this way might lead people to believe that this adds one datagram or another batch.  Or that you could do `batch1.extend_from_slice(batch2.as_slice())` after just checking that the sizes match.  That's not going to be the case, so you probably want to a) document this limitation very clearly OR b) don't allow extension when the last datagram isn't full-sized.

Either way, you want some tests that demonstrate this corner case.

martinthomson (changes requested):

larseggert:
I'll leave this as draft per @mxinden's suggestion to first stabilize the Gecko parts.

neqo-common/src/datagram.rs
```diff
@@ -256,6 +256,20 @@ impl Batch {
...
 
+    /// Append a single datagram's payload to the batch.
+    ///
+    /// Returns `true` if appended successfully, `false` if rejected because:
+    /// - `data` is empty or larger than `datagram_size`, or
+    /// - the last datagram in the batch is partial (not yet full-sized).
+    pub fn push_datagram(&mut self, data: &[u8]) -> bool {
```
larseggert, line 264: Should this return `Result` instead?

neqo-common/src/datagram.rs
```diff
@@ -256,6 +256,20 @@ impl Batch {
         })
     }
 
+    /// Append a single datagram's payload to the batch.
+    ///
+    /// Returns `true` if appended successfully, `false` if rejected because:
+    /// - `data` is empty or larger than `datagram_size`, or
+    /// - the last datagram in the batch is partial (not yet full-sized).
```
mxinden, line 263: How can a datagram _not yet_ be full-sized? Do you expect a user to call `push_datagram` multiple times for a single on-the-wire datagram? If so, is that intuitive?

mxinden (commented):
Do I understand correctly, that this is a pure convenience function, in other words that a user could as well aggregate datagrams in its own `Vec<u8>` and only as a final step create a `DatagramBatch`? If so, is this helper function worth the potential foot guns it introduces?

larseggert:
Excellent catch @mxinden. No need for this.

Issue #3342, opened 2026-01-22
"Observed bandwidth is smaller than the link rate"


larseggert:
We're still seeing spurious CI failures with
```
thread 'fc::test::auto_tuning_approximates_bandwidth_delay_product' (41944) panicked at neqo-transport/src/fc.rs:1344:13:
Got receive window of 1024 KiB (effectively 768 KiB) on connection with observed bandwidth 1.4117647058823528 MBit/s. Expected: bandwidth 3 MBit/s (3000000 Bit/s), rtt 1ms, bdp 0 KiB. Observed bandwidth is smaller than the link rate.
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <neqo_transport::fc::test::auto_tuning_approximates_bandwidth_delay_product::{closure#0} as core::ops::function::FnOnce<()>>::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    fc::test::auto_tuning_approximates_bandwidth_delay_product
```

https://github.com/mozilla/neqo/actions/runs/21245821946/job/61134743352?pr=3339#step:7:1722

larseggert:
@mxinden Claude thinks it's due to this:

## Failure Details

```
thread 'fc::test::auto_tuning_approximates_bandwidth_delay_product' panicked at neqo-transport/src/fc.rs:1344:13:
Got receive window of 1024 KiB (effectively 768 KiB) on connection with observed bandwidth 0.96 MBit/s.
Expected: bandwidth 2 MBit/s (2000000 Bit/s), rtt 1ms, bdp 0 KiB.
Observed bandwidth is smaller than the link rate.
```

## Root Cause

The test fails under a specific combination of randomized parameters: **low bandwidth (2 Mbit/s) + very low RTT (1ms)**.

### Why this combination is problematic

1. **Tiny BDP**: With 2 Mbit/s and 1ms RTT, the bandwidth-delay product is only ~250 bytes
2. **Initial window already sufficient**: `INITIAL_LOCAL_MAX_STREAM_DATA` is 1 MiB, which is ~4000x larger than the BDP
3. **No window growth needed**: Since BDP << initial window, auto-tuning never increases the window
4. **Short measurement period**: The test exits when the window is stable for 8 RTT = **8ms**
5. **Slow transmission**: At 2 Mbit/s, transmitting a single 1500-byte frame takes **6ms**
[...]

PR #3392, opened 2026-02-10
test(transport/fc): skip small BDPs in auto_tune simulation

mxinden:
The `auto_tuning_approximates_bandwidth_delay_product` test generates random bandwidth and RTT values and simulates a network to test the auto tuning algorithm in. The test batches bytes sent in ~UDP datagram sizes (i.e. 1_500 bytes). If the BDP is lower than one UDP datagram, latency calculations are off and incorrect values are generated, leading to test failure.

With this commit, the test skips BDPs smaller than one UDP datagram. I don't think they are worth testing.

Fixes https://github.com/mozilla/neqo/issues/3342.

### DIFF
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -1226,6 +1226,7 @@ mod test {
     }
 
     #[expect(clippy::cast_precision_loss, reason = "This is test code.")]
+    #[expect(clippy::too_many_lines, reason = "This is test code.")]
     #[test]
     fn auto_tuning_approximates_bandwidth_delay_product() -> Res<()> {
         const DATA_FRAME_SIZE: u64 = 1_500;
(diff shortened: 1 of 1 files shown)


neqo-transport/src/fc.rs
```diff
@@ -1246,6 +1246,13 @@ mod test {
             let half_rtt = rtt / 2;
             let bdp = bandwidth * u64::try_from(rtt.as_millis()).unwrap() / 1_000 / 8;
 
+            if bdp <= DATA_FRAME_SIZE {
+                // Skip very low BDPs, i.e. BDPs smaller than one UDP datagram.
+                // Latency calculations in test assume it can transfer
+                // DATA_FRAME_SIZE bytes in 1 RTT.
+                continue;
```
martinthomson: Rather than skipping.  How about setting this as the floor on the randomness?  You could achieve that by setting the low value of the or bandwidth RTT to a higher value.  1Mbps / 8000 = 125 is the low end, so all you need to do is set the minimum RTT to 12 or the minimum bandwidth to 12Mbps given the current setup.

neqo-transport/src/fc.rs
```diff
@@ -1246,6 +1246,13 @@ mod test {
             let half_rtt = rtt / 2;
             let bdp = bandwidth * u64::try_from(rtt.as_millis()).unwrap() / 1_000 / 8;
```
martinthomson:
```suggestion
let rtt_int = u64::from(random::<1>()[0]) + 1;
let rtt = Duration::from_millis(rtt_int);
let half_rtt = rtt / 2;
let bdp = bandwidth * rtt_int / 1_000 / 8;
```

neqo-transport/src/fc.rs
larseggert: Agree. The test should pick parameters from ranges that avoid this issue.

larseggert (approved):

PR #3393, opened 2026-02-10
Fix: guard against invalid state in http09 client

bobsingh-dev:
This change adds a guard to prevent an invalid HTTP/0.9 client state during request handling.

This avoids unexpected behavior when the client encounters an unsupported or empty response.

### DIFF
--- a/neqo-bin/src/client/http09.rs
+++ b/neqo-bin/src/client/http09.rs
@@ -287,7 +287,9 @@ impl<'b> Handler<'b> {
                 false
             }
             Err(e) => {
-                return Err(e.into());
+                qwarn!("Unexpected error creating stream {e:?}");
+                self.url_queue.push_front(url);
+                false
             }
         }
     }


bobsingh-dev:
> This doesn't compile: the function has a return type of `bool`.
> 
> What situation are you seeking to address?

Thanks for the catch Martin— fixed the return type and clarified the behavior.

bobsingh-dev:
> This needs a test (demonstrating the panic).

Good point. I was aiming to make this more defensive, but I don’t have a concrete reproduction case for this error path.

Given that this is example code, I’m happy to revert the change if the panic is considered acceptable here.

larseggert:
Well, fixing it would be nice if it can be triggered. Hence the test.

bobsingh-dev:
> Well, fixing it would be nice if it can be triggered. Hence the test.

From reviewing stream_create(), it appears the only expected errors are StreamLimit and ConnectionState, which are already handled above. I don’t see a reproducible path that would trigger the “unexpected” branch. If there’s a scenario where another error can occur, I’d be happy to add a test case for it.

martinthomson:
It's your pull request; the onus is on you to motivate it.  Closing for now.

PR #3399, opened 2026-02-20
refactor: Use nss-rs instead of neqo-crypto

Not-Nik:
Replaces neqo-crypto with nss-rs. nss-rs is the new unified NSS binding for Rust. It merges functionality of nss-gk-api and neqo-crypto, which were originally based on the same code, but diverged into separate libraries. The new crate is located in its own repository at [mozilla/nss-rs](https://github.com/mozilla/nss-rs). Other Firefox components like mls-platform-api and authenticator-rs will also transition to nss-rs in the future.

Most of the API of both nss-gk-api and neqo-crypto was preserved except for minor changes. Since both libraries declared an `Aead` type, the neqo-crypto version is now named `RecordProtection` (mozilla/nss-rs#14), but both are exposed at the old path (i.e. `nss_rs::RecordProtection`, `nss_rs::aead::Aead`).

P.S.: This is still a draft, because some of neqo's CI jobs can't run on forks and I am unsure if they will pass here.


martinthomson:
Code coverage here is completely off.  We'll need a fresh baseline for the project after such a large change.

.github/actions/nss/action.yml
```diff
@@ -9,14 +9,21 @@ inputs:
   version-file:
     description: "Path to file containing the minimum NSS version (relative to workspace root)"
     required: false
-    default: "neqo-crypto/min_version.txt"
+    default: "nss-rs/min_version.txt"
```
larseggert: Should this move into nss-rs? How would you handle the case where two conumsers of nss-rs have different requirements for min versions?

.github/workflows/bench.yml
```diff
@@ -61,8 +61,13 @@ jobs:
...
           cd neqo
-          cat neqo-crypto/min_version.txt > versions.txt
-          git show origin/main:neqo-crypto/min_version.txt >> versions.txt
+          nss_rs_dir=$(cargo metadata --format-version=1 -q \
+            | jq -r ".packages[]
+                     | select(.name==\"nss-rs\")
+                     | .manifest_path" \
+            | xargs dirname)
+          cat "$nss_rs_dir/min_version.txt" > versions.txt
+          curl https://raw.githubusercontent.com/mozilla/nss-rs/refs/heads/main/min_version.txt >> versions.txt
```
larseggert: This is repeated several times. Can this be factored out?

neqo-transport/src/packet/retry.rs
```diff
@@ -7,17 +7,17 @@
...
 
 use neqo_common::qerror;
-use neqo_crypto::{Aead, AeadTrait as _, TLS_AES_128_GCM_SHA256, TLS_VERSION_1_3, hkdf};
+use nss_rs::{RecordProtection, TLS_AES_128_GCM_SHA256, TLS_VERSION_1_3, hkdf};
 
 use crate::{Error, Res, version::Version};
 
 /// The AEAD used for Retry is fixed, so use thread local storage.
-fn make_aead(version: Version) -> Aead {
+fn make_aead(version: Version) -> RecordProtection {
```
larseggert: Why the name change from `Aead` to `RecordProtection`?

.deny.toml
```diff
@@ -33,3 +33,6 @@ version = "0.4"
 unknown-registry = "deny"
 unknown-git = "deny"
 required-git-spec = "rev"
+allow-git = [
+    "https://github.com/mozilla/nss-rs.git",
+]
```
larseggert, line 42: I guess this is temporary - add a `TODO` about reverting it.

.github/actions/nss/action.yml
martinthomson: nss-rs is going to be the thing that has the NSS dependency, so it should be part of that crate, yes.

neqo-transport/src/packet/retry.rs
martinthomson: That's on me.  There are two AEADs in nss-rs that use different interfaces in NSS (one is hooked into TLS record protection).  I think we want to eliminate that one eventually, but we agreed to rename it for now.

Cargo.toml
```diff
@@ -36,6 +35,7 @@ hex = { version = "0.4", default-features = false }
 http = { version = "0.2.9", default-features = false }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
+nss-rs = { git = "https://github.com/mozilla/nss-rs", rev = "0b9edb5a11e7f3d26638ec4b190947d30611cc18" }
```
larseggert:
```suggestion
neqo-crypto = { package = "nss-rs", git = "https://github.com/mozilla/nss-rs", rev = "0b9edb5a11e7f3d26638ec4b190947d30611cc18" }
```
If you rename like this, can we avoid a bunch of `neqo-crypto` -> `nss-rs` changes? 

But we might actually _want_ those for clarity. What do others think?

neqo-transport/src/packet/retry.rs
larseggert: How about
```rust
use nss_rs::RecordProtection as Aead;
```
Would minimize the diff...

martinthomson: I don't mind the renaming.

.github/actions/nss/action.yml
Not-Nik: In my understanding the requirement for the version comes from nss-rs, not from it's consumers. Even if two consumers in the same project want different versions of nss-rs and we were to handle that, there's a good chance we would end up with linking issues. We probably just want to use one version that's compatible with all consumers.

.github/actions/nss/action.yml
larseggert, line 1: Should neqo not use the nss-rs version of this action, and remove this file?

.github/actions/minimum-version/action.yml
larseggert, line 1: Should this action move to nss-rs?

.deny.toml
larseggert: Should this change before merging the PR?

Cargo.lock
```diff
@@ -88,9 +88,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bindgen"
-version = "0.72.0"
+version = "0.72.1"
```
larseggert, line 91: Gratuitous change?

larseggert (commented):
Also needs a rebase.

Cargo.lock
Not-Nik: `pkcs11-bindings` fails to compile with =0.72.0
```
error[E0080]: attempt to compute `1_usize - 552_usize`, which would overflow
    --> /Users/nwipper/Documents/nss-rs/target/debug/build/pkcs11-bindings-363623b7b7711c45/out/bindings.rs:1033:34
     |
1033 |     ["Size of CK_FUNCTION_LIST"][::std::mem::size_of::<CK_FUNCTION_LIST>() - 552usize];
     |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `_` failed here

For more information about this error, try `rustc --explain E0080`.
error: could not compile `pkcs11-bindings` (lib) due to 1 previous error
```

.github/actions/nss/action.yml
Not-Nik: That's a possibility, but we'll have to think about how we get the action from nss-rs. Pinning to a commit might be extra hassle, because we'll have to update it every time the action changes. Using a branch name will cause trouble, if the action has a breaking change and neqo isn't updated accordingly

neqo-transport/src/connection/tests/null.rs
```diff
@@ -6,12 +6,14 @@
 
 #![cfg(feature = "disable-encryption")]
 
-use neqo_crypto::aead_null::AEAD_NULL_TAG;
+use nss_rs::aead::AEAD_NULL_TAG;
```
martinthomson: This isn't used, but should it be?

neqo-transport/src/connection/tests/null.rs
```diff
@@ -6,12 +6,14 @@
...
 use test_fixture::now;
 
 use super::{connect_force_idle, default_client, default_server};
 use crate::StreamType;
 
+const AEAD_NULL_TAG_LEN: usize = 16;
```
martinthomson:
```suggestion
const AEAD_NULL_TAG_LEN: usize = AEAD_NULL_TAG.len();
```

martinthomson (approved):
I'm ok with this, but I have a request about the name.

It will require talking to JC and Jed, but I think that you should wholesale replace https://crates.io/crates/nss

The "-rs" part of the nss-rs name is pure baggage.

Not-Nik:
"Vendor into Gecko" fails, because rc_crypto already has a crate named "nss". Obviously they are not the same crate, but we still get an error. Perhaps we can bypass that check? Otherwise we would likely have to go back to naming the crate nss-rs, although we could rename it in Cargo.toml so the import stays the same

larseggert:
We *will* need to resolve this, because it will hit us when actually vendoring into Gecko. Any ideas @mxinden?

Cargo.toml
```diff
@@ -37,6 +36,7 @@ hex = { version = "0.4", default-features = false }
 http = { version = "0.2.9", default-features = false }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
+nss = { git = "https://github.com/mozilla/nss-rs", rev = "3c99b2581d1b17a66d531db9d0aab5237c62ba59" }
```
larseggert: @Not-Nik should we start creating releases for `nss-rs` going forward, to make managing this dep a bit easier? (You need not publish the releases to crates.io, but at least create them on GitHub.)

Not-Nik: Sounds good

larseggert:
@Not-Nik once we have nss-rs@0.0.1, let's get this merged!

Not-Nik:
Version is 0.8.0, because the current version of the nss crate on crates.io is 0.7.1. Do we need a fix for NetBSD?

mxinden:
>> "Vendor into Gecko" fails, because rc_crypto already has a crate named "nss". Obviously they are not the same crate, but we still get an error. Perhaps we can bypass that check? Otherwise we would likely have to go back to naming the crate nss-rs, although we could rename it in Cargo.toml so the import stays the same
>
> We will need to resolve this, because it will hit us when actually vendoring into Gecko. Any ideas @mxinden?

I don't have a good solution. Bypassing that check is in my eyes not worth the complexity. Renaming to `nss-rs` is fine by me. Alternatively, by convention of a Rust crate wrapping a C library: `nss-sys`. Maybe even just `nss-wrapper`? I don't think any of this needs to be catchy.

larseggert:
@martinthomson given the vendoring issue, OK to rename this back to `nss-rs`?

martinthomson:
I can live with that, but I would strongly prefer that it be a temporary situation.  Is there something we can do to change the rc_crypto dependency?

larseggert:
@martinthomson we can try and contact the `application-services` folks about this. @Not-Nik in the meantime, let's stick with `nss-rs` so we can get this landed. Please open an issue in `nss-rs` and `application-services` to track the renaming.

Cargo.lock
mxinden: Note that Firefox uses `0.72.0`:

https://github.com/mozilla-firefox/firefox/blob/c34ebfaedf1a009f0bd13f1700e9b795a1b520c7/Cargo.lock#L503-L518

That said, if the change is needed, I am sure we can change the `bindgen` version used by Firefox.

Cargo.lock
```diff
@@ -965,6 +959,15 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
+[[package]]
+name = "pkcs11-bindings"
+version = "0.1.7"
```
mxinden, line 963: 👍 For the record, same as Firefox:

https://github.com/mozilla-firefox/firefox/blob/main/Cargo.lock#L5752

.deny.toml
mxinden: Not a strong opinion:

I am fine consuming nss-rs through git instead of crates.io, both here and in Firefox. It's just one less attack vector (i.e. crates.io) in our software supply chain.

mxinden (approved):
This is ready to merge from my end.

Given my low involvement in the effort, and minimal expertise around NSS, let's wait for Lars' and/or Martin's review.

Thanks @Not-Nik for this work!

larseggert:
@Not-Nik is there still something you want to do to this PR before merging?

Not-Nik:
Nope, all good from my end

larseggert:
At the very least, the PR needs a rebase.

Not-Nik:
Done! Let's get this merged

neqo-transport/src/connection/tests/null.rs
```diff
@@ -33,7 +35,7 @@ fn no_encryption() {
     assert_eq!(&buf[..len], DATA_CLIENT);
     server.stream_send(stream_id, DATA_SERVER).unwrap();
     let server_pkt = server.process_output(now()).dgram().unwrap();
-    assert!(server_pkt[..server_pkt.len() - AEAD_NULL_TAG.len()].ends_with(DATA_SERVER));
+    assert!(server_pkt[..server_pkt.len() - AEAD_NULL_TAG_LEN].ends_with(DATA_SERVER));
```
larseggert: Don't really need `AEAD_NULL_TAG_LEN` given that all of these are `const`.

qns/Dockerfile
```diff
@@ -23,17 +23,12 @@ ENV NSS_DIR=/nss \
     NSPR_DIR=/nspr \
     LD_LIBRARY_PATH=/dist/Release/lib
 
+ARG MOZILLA_ACTIONS_REF=main
```
larseggert:
```suggestion
ARG MOZILLA_ACTIONS_REF=v1
```
Does this work? `main` may be too new and/or have issues.

Cargo.lock
larseggert: I still wonder if we really need this.

larseggert (approved):

Cargo.lock
Not-Nik: 0.72.1 includes a [fix for exactly this issue](https://github.com/rust-lang/rust-bindgen/issues/3264). We definitely do need the newer version.

larseggert:
@Not-Nik thanks for getting this over the line! I'm rerunning the performance comparison and will then merge.

PR #3408, opened 2026-03-01
test(transport): trigger earliest > now via blocked Initial and HS

mxinden:
After ~12h Claude came up with a reproducer for https://bugzilla.mozilla.org/show_bug.cgi?id=2001901 (with minimal guidance from me and Pernosco).

Given that in this test the client is blocked on initial and handshake space, https://github.com/mozilla/neqo/pull/3350 is a potential fix, allowing an initial space transmission during handshake PTO.

---

This commit adds a test where the client sends a split ClientHello (two Initial packets) but only the first reaches the server. The server cannot complete the TLS handshake, so the client never obtains Handshake TX keys.

After two PTO iterations the Handshake PTO timer falls behind the Initial PTO timer. On the third PTO iteration only the Handshake timer fires. The resulting PING is ACKed by the server, which clears `pto_state`.

On the next `process_output` the Handshake PTO fires alone: Initial space is blocked (`ack_only`) and Handshake space has no TX keys, so nothing is sent. `next_delay` then finds the Handshake PTO deadline in the past and debug_assert!(earliest > now) fires.

Reproducer for https://bugzilla.mozilla.org/show_bug.cgi?id=2001901.

### DIFF
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -1657,3 +1657,64 @@ fn certificate_compression() {
     assert!(!ORIGINAL.lock().unwrap().is_empty());
     assert_eq!(*ORIGINAL.lock().unwrap(), *DECODED.lock().unwrap());
 }
+
+/// Regression test for <https://bugzilla.mozilla.org/show_bug.cgi?id=2001901>.
+#[test]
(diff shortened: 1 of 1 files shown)


mxinden:
Replaced by https://github.com/mozilla/neqo/pull/3411.

PR #3411, opened 2026-03-03
fix: Don't prime the handshake PTO w/o keys

### LINKED ISSUE #3410: Don't prime the handshake PTO w/o keys
This might prime the handshake PTO even if we have no Handshake keys available, right?

Sample scenario:

1. Client sends ClientHello split across two Initial packets
2. Server receives only the first packet, cannot complete TLS, sends a bare ACK.
3. Client receives the ACK for second packet.
4. Client Initial PTO fires
5. `maybe_prime_handshake_pto` primes handshake PTO even though the client has no Handshake TX keys (ServerHello was never received).
6. Handshake PTO fires (potentially in the past), but we can not act on it.

Is the above correct @larseggert? What do you think of only calling `maybe_prime_handshake_pto` if we have handshake space keys to actually send anything.

This would fix https://github.com/mozilla/neqo/pull/3408 and thus also fix (at least one case of) https://bugzilla.mozilla.org/show_bug.cgi?id=2001901.

_Originally posted by @mxinden in https://github.com/mozilla/neqo/pull/3097#discussion_r2874486815_

larseggert:
Fixes #3410

### DIFF
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -159,12 +159,17 @@ impl HeaderTable {
     /// `HeaderLookup` if entry does not exist.
     pub fn get_dynamic(&self, index: u64, base: u64, post: bool) -> Res<&DynamicTableEntry> {
         let inx = if post {
-            if self.base < (base + index + 1) {
+            if self.base
+                < base
+                    .checked_add(index)
+                    .and_then(|s| s.checked_add(1))
+                    .ok_or(Error::IntegerOverflow)?
+            {
                 return Err(Error::HeaderLookup);
             }
             self.base - (base + index + 1)
         } else {
-            if (self.base + index) < base {
+            if self.base.checked_add(index).ok_or(Error::IntegerOverflow)? < base {
                 return Err(Error::HeaderLookup);
             }
             (self.base + index) - base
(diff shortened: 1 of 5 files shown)


neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1032,6 +1032,50 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+/// `pto_state` and the stale Handshake PTO fires in the past.
+#[test]
+fn handshake_pto_not_primed_without_keys() {
+    let mut now = now();
+    let mut client = new_client(ConnectionParameters::default().pacing(false));
+    let mut server = new_server(ConnectionParameters::default().pacing(false));
+
+    // Split ClientHello across two Initial packets (MLKEM).
+    let c_init_1 = client.process_output(now).dgram().unwrap();
+    assert!(client.process_output(now).dgram().is_some()); // c_init_2, intentionally dropped
```
martinthomson: Copilot is sort of correct here.  This should assert that the second datagram indeed starts with an Initial.

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1032,6 +1032,50 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // Deliver only c_init_1.  Incomplete CH → server sends bare ACK.
+    now += DEFAULT_RTT / 2;
+    server.process_input(c_init_1, now);
+    let s_ack = server.process_output(now).dgram().unwrap();
```
martinthomson: collapse these two into a single "process()" call

neqo-transport/src/connection/tests/recovery.rs
```diff
@@ -1032,6 +1032,50 @@ fn pto_handshake_space_when_server_flight_lost() {
...
+
+    // Two PTO iterations (output dropped) push Initial baseline ahead of the
+    // Handshake baseline that was primed during iteration 1.
+    for _ in 0..2 {
+        now += client.process_output(now).callback();
+        while client.process_output(now).dgram().is_some() {}
+    }
+
+    // Third PTO: deliver the PING so the server ACKs it, clearing pto_state.
+    now += client.process_output(now).callback();
```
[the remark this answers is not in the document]
martinthomson: Copilot is right to ask that the value isn't zero.  I don't like it's suggested code.

martinthomson (commented):
The agents might not be quite ready to take our jobs, but this is a good find.

PR #3413, opened 2026-03-03
chore: Sync deps with Gecko

larseggert:
Via `test/update-lockfile.py`


Cargo.lock
```diff
@@ -1659,3 +1654,9 @@ checksum = "6f42320e61fe2cfd34354ecb597f86f413484a798ba44a8ca1165c58d42da6c1"
 dependencies = [
  "bitflags",
 ]
+
+[[package]]
+name = "zmij"
+version = "1.0.21"
```
mxinden: `1.0.20`

https://github.com/mozilla-firefox/firefox/blob/83d1a08db47b91a4d53341a799745caac9c38bde/Cargo.lock#L8807

PR #3428, opened 2026-03-09
Add WebTransport ExportKeyingMaterial backend

jesup:
Support exporting keying material to support WebTransport export API


larseggert:
@jesup can you rebase this?

jesup:
It's odd we're showing performance regressions here unless perhaps it's because it's based on an older revision...  This should have no impact on perf if the code isn't called, other than we have a small amount of extra code, changing code layout/page layout slightly (very slightly)

larseggert:
You can ignore these regressions, especially the Codspeed ones (Codspeed is a third-party vendor we're evaluating for @sylvestre to see if their offering eliminates the need for our own on-prem machines.)

mxinden:
@jesup is this pull request ready for review? If not, want to convert it back to _draft_ via the `Convert to draft` button in the right sidebar?

jesup:
This *should* be ready for review, but note I'm still hammering out workflows with jj and github (and patch stacks).

neqo-crypto/min_version.txt
```diff
@@ -1 +1 @@
-3.121
+3.120.1
```
jesup, line 1: This is not supposed to be part of this; I'll remove it

neqo-crypto/src/agent.rs
```diff
@@ -802,6 +802,50 @@ impl SecretAgent {
...
+    /// * `context` - Optional context data
+    /// * `out_len` - Length of output keying material
+    ///
+    /// # Errors
+    ///
+    /// Returns error if handshake is not complete or export fails.
+    pub fn export_keying_material(
+        &self,
+        label: &[u8],
+        context: Option<&[u8]>,
```
martinthomson, line 820: You can remove the Option here, which will simplify the interface.

https://tlswg.org/tls13-spec/auth48/rfc9846.html#section-7.5-5 explains that in TLS 1.3, there is no difference between having no context and having an empty context, which is a change from previous versions.  Because we don't have to support previous versions, the result is that we always have a context.

neqo-crypto/tests/agent.rs
```diff
@@ -829,3 +829,113 @@ fn connection_fails_encoder_returned_too_long() {
...
+#[test]
+fn export_keying_material_basic() {
+    fixture_init();
+    let mut client = Client::new("server.example", true).expect("should create client");
+    let mut server = Server::new(&["key"]).expect("should create server");
+
+    connect(&mut client, &mut server);
+
+    let label = b"EXPORTER-test";
+    let result = client.export_keying_material(label, None, 32);
```
martinthomson, line 842: A better test for this would be to check that client and server both produce the same value (you have that test, use it).  You might also confirm that a sufficiently large value is different for different connections.

neqo-http3/src/connection_client.rs
```diff
@@ -929,6 +929,30 @@ impl Http3Client {
...
+    ///
+    /// Returns `TransportError` if the connection is not ready or export fails.
+    pub fn export_keying_material(
+        &self,
+        label: &[u8],
+        context: Option<&[u8]>,
+        out_len: usize,
+    ) -> Res<Vec<u8>> {
+        self.conn
+            .export_keying_material(label, context, out_len)
```
martinthomson, line 952: I just realized that giving sites access to an unrestricted exporter is a big problem.  Let's talk about that in the specification discussion.

martinthomson (changes requested):
I've still got it.  LLMs aren't catching these problems...

neqo-http3/src/connection_client.rs
martinthomson: The spec considers this already (it just isn't in the overview document).  https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#section-4.8 has a set of extra pieces that need to be fed into the context.

A webtransport-specific function here should do that work.

jesup:
replaced by https://github.com/mozilla/neqo/pull/3464

Issue #3022, opened 2025-09-29
Examine usage and conversion of `f64` values in CUBIC code


omansfeld:
> There are a bunch of `f64` values in this code (not necessarily new with this PR, also before) that are inherently integers. For example, everything measured in bytes, etc. Could we re-examine this code to make sure we use integer types wherever possible, rather than `f64`? (At least if we can do so without more conversions.)

I looked at this when first starting the work on cubic. While all the byte values (cwnd, mss, acks etc.) are inherently `usize` they have to be multiplicated/divided with `f64` values (e.g. all the constants that are not whole numbers).

I don't know if there is a smarter rust way to do this, but the only thing I can think of is using a divisor/divident approach instead, like in the `reduce_cwnd` function:

https://github.com/mozilla/neqo/blob/a75f3b58c889917fd28ec4a7d69ead8573eb8567/neqo-transport/src/cc/cubic.rs#L345-L351

I believe doing the conversions is more readable in comparison. That's why I initially abandoned the idea of doing something about it -- happy to hear of another way though if there is one.

Anyways, I will open another issue to track this, because I don't think it has to be part of this specific PR.

_Originally posted by @omansfeld in https://github.com/mozilla/neqo/issues/2973#issuecomment-3346978910_

larseggert:
If all instances need to go into `f64` math, we are done :-)

(Still wonder what the math came out with if we used something like https://docs.rs/fixed/latest/fixed/ like the Linux code does...)

PR #3431, opened 2026-03-10
feat(stats): add a stat recording cubic's w_max

omansfeld:
I would like to use this as an approximation for the ideal congestion window size to compare it with `slow_start_exit_cwnd` and rate exit-point accuracy. 

We already compare with `cwnd` in the new `slow_start_exit_accuracy` glean probe being added in https://bugzilla.mozilla.org/show_bug.cgi?id=2021867 and I would like to have this as an alternative, to confirm if the two yield similar results.

~~Depends on #3278. I will rebase once that lands on main. Only commit belonging to this PR is dd20486ea6f871efd36d181dd089774a9a3058e1.~~

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -1626,7 +1626,7 @@ mod tests {
 
         // 2. Lose packets (1, 2) --> `RecoveryStart`, 1 event, reduced cwnd
         let cwnd_before_loss = cc.cwnd();
-        assert!(cc_stats.w_max.is_none());
+        assert_eq!(cc_stats.w_max, None);
         let mut lost_pkt1 = pkt1.clone();
         let mut lost_pkt2 = pkt2.clone();
         lost_pkt1.declare_lost(now, sent::LossTrigger::TimeThreshold);
@@ -1646,9 +1646,13 @@ mod tests {
             Some(SlowStartExitReason::CongestionEvent)
         );
         assert_eq!(cc_stats.congestion_events[CongestionEvent::Loss], 1);
-        #[expect(clippy::cast_precision_loss, reason = "cwnd_before_loss is small")]
-        let expected_w_max = cwnd_before_loss as f64;
-        assert_eq!(cc_stats.w_max, Some(expected_w_max));
+        #[expect(
+            clippy::cast_sign_loss,
+            clippy::cast_possible_truncation,
+            reason = "w_max is non-negative and represents whole bytes"
+        )]
+        let w_max_stat = cc_stats.w_max.unwrap() as usize;
+        assert_eq!(w_max_stat, cwnd_before_loss);
         assert_eq!(
             cc.cwnd(),
             cc.cwnd_initial() * Cubic::BETA_USIZE_DIVIDEND / Cubic::BETA_USIZE_DIVISOR
@@ -1696,7 +1700,7 @@ mod tests {
         assert_eq!(cc_stats.congestion_events[CongestionEvent::Loss], 1);
         assert_eq!(cc_stats.congestion_events[CongestionEvent::Spurious], 1);
         assert_eq!(cc.cwnd(), cc.cwnd_initial());
-        assert!(cc_stats.w_max.is_none());
+        assert_eq!(cc_stats.w_max, None);
     }
 
     /// This tests a scenario where spurious detection happens late, after cwnd has recovered and


neqo-transport/src/stats.rs
```diff
@@ -148,6 +148,11 @@ pub struct CongestionControlStats {
     /// None if we haven't exited slow start or if we re-entered after spurious congestion.
     /// When exiting via congestion event, this is the cwnd AFTER the reduction.
     pub slow_start_exit_cwnd: Option<usize>,
+    /// Cubic's `w_max`: the congestion window (in bytes) just before the most recent
+    /// congestion reduction (with fast convergence applied). Reports 0 (the initial value) if no
+    /// congestion event has occurred or Cubic is not in use. Recorded as a stat to approximate a
+    /// connection's ideal congestion window in metrics.
+    pub w_max: usize,
```
larseggert: Is it is an `f64` internally, why not export as such?

omansfeld: Glean cannot take floats, that was the initial reason. On second thought It makes more sense to do the conversion on the Firefox side though, and keep Neqo coherent with itself. Will change to f64.

neqo-transport/src/cc/cubic.rs
```diff
@@ -472,5 +477,7 @@ impl WindowAdjustment for Cubic {
             self.current
         );
         self.current = stored;
+        // The stat is `None` if it is unset, so set it to that instead of `Some(0.0)`.
+        cc_stats.w_max = (self.current.w_max != 0.0).then_some(self.current.w_max);
```
omansfeld: @larseggert I made this stat an `Option` too, as per your suggestion. As you can see the actual Cubic `w_max` is also using `0.0` as it's *has not been set*-value, prompting the above line of code to align stat and reality. Should `Cubic::w_max` also be changed to an `Option` or is that overkill (for this PR)?

larseggert: I think it would be cleaner if it were. Also allows distinguishing "never set" from "happens to be zero" (which isn't really possible I guess but anyway.)

omansfeld:
@larseggert can you take another now that I also made `Cubic::w_max` an `Option` in https://github.com/mozilla/neqo/pull/3431/commits/d84b8ea632ccef85b57256da874867f1af0b34a3?

neqo-transport/src/cc/tests/cubic.rs
```diff
@@ -304,7 +304,7 @@ fn congestion_event_slow_start() {
     // w_max is equal to cwnd before decrease.
     let cwnd_initial_f64 = convert_to_f64(cubic.cwnd_initial());
     assert_within(
-        cubic.congestion_control().w_max(),
+        cubic.congestion_control().w_max().unwrap(),
         cwnd_initial_f64 + convert_to_f64(cubic.max_datagram_size()),
```
larseggert, line 308: See https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/ for why using Epsilon in this way is flaky.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1623,6 +1625,8 @@ mod tests {
         assert_eq!(cc_stats.congestion_events[CongestionEvent::Spurious], 0);
 
         // 2. Lose packets (1, 2) --> `RecoveryStart`, 1 event, reduced cwnd
+        let cwnd_before_loss = cc.cwnd();
+        assert!(cc_stats.w_max.is_none());
```
```suggestion
        assert_eq!(cc_stats.w_max, None);
```
You used this elsewhere... (also in other places)

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1642,6 +1646,9 @@ mod tests {
             Some(SlowStartExitReason::CongestionEvent)
         );
         assert_eq!(cc_stats.congestion_events[CongestionEvent::Loss], 1);
+        #[expect(clippy::cast_precision_loss, reason = "cwnd_before_loss is small")]
+        let expected_w_max = cwnd_before_loss as f64;
+        assert_eq!(cc_stats.w_max, Some(expected_w_max));
```
larseggert: `f64` comparison

larseggert (approved):

neqo-transport/src/cc/tests/cubic.rs
omansfeld: Very interesting read, that part was amazing:

> If you haven’t memorized pi to 15+ digits then this may be lost on you, but the salient point is that float(pi) + sin(float(pi)) is a more accurate value of pi than float(pi). Or, alternately, sin(float(pi)) tells you nothing more than the error in float(pi).

I think here it just highlights the fundamental problem that we use `f64` for a bunch of values that represent bytes (so really should be `usize`) because they are used in floating point math in Cubic that then converts it's results back to `usize` when returning values to the outside. That is already tracked in #3022.

So really most of the comparisons in the cubic tests are comparing whole numbers (as `f64`) that have been assigned directly from `usize`. I guess that makes them less flaky. At least I have never noticed tests failing intermittently because of the comparisons in this file.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Thanks for pointing it out, I fixed that in #3432 and missed it here.

neqo-transport/src/cc/classic_cc.rs
omansfeld: Converted to `usize` instead and added more explanations in the clippy reason. Also see the comment above about the whole `f64` situation.

PR #3432, opened 2026-03-10
feat(stats): add HyStart++ stats

omansfeld:
~~Depends on #3278. I will rebase once that lands on main. Only commits belonging to this PR are the following:~~

add `slow_start_exit_reason` stat
- `CongestionEvent` if loss or ecn
- `Heuristic` if algorithm (e.g. HyStart++)

add `css_rounds_finished` stat
add `css_entries` stat to count how often css was entered

The two CSS stats can be used to judge how often HyStart++ spuriously enters CSS and thus spends (unnecessary) time in throttled growth. In a perfect world we'd have 1 css entry and 5 rounds finished. If values are wildly increased that might point to an inefficiency.

The exit reason stat can be used to judge how well HyStart++ (or another heuristic) performs at detecting an exit point before packet loss occurs. Can also be used as a filter when recording stats on the firefox side as glean probes, e.g. to differentiate `slow_start_exit_accuracy` (introduced in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2021867)) for heuristic and congestion event based exits.

### DIFF
--- a/neqo-transport/src/stats.rs
+++ b/neqo-transport/src/stats.rs
@@ -162,10 +162,10 @@ pub struct CongestionControlStats {
     /// Number of times HyStart++ entered CSS (Conservative Slow Start). Only meaningful when
     /// HyStart++ is enabled. Higher values indicate that HyStart++ had many spurious CSS
     /// entries, spending more time throttling slow start growth.
-    pub css_entries: usize,
+    pub hystart_css_entries: usize,
     /// Number of CSS (Conservative Slow Start) rounds completed. Only meaningful when HyStart++ is
     /// enabled. Higher values indicate the heuristic spent more time throttling slow start growth.
-    pub css_rounds_finished: usize,
+    pub hystart_css_rounds_finished: usize,
     /// The current congestion window size (in bytes). Updated throughout the connection
     /// lifetime.
     pub cwnd: Option<usize>,
--- a/neqo-transport/src/cc/hystart.rs
+++ b/neqo-transport/src/cc/hystart.rs
@@ -220,7 +220,7 @@ impl SlowStart for HyStart {
             );
             if current >= last + rtt_thresh {
                 self.css_baseline_min_rtt = Some(current);
-                cc_stats.css_entries += 1;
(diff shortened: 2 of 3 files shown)


neqo-transport/src/stats.rs
```diff
@@ -148,6 +156,16 @@ pub struct CongestionControlStats {
     /// None if we haven't exited slow start or if we re-entered after spurious congestion.
     /// When exiting via congestion event, this is the cwnd AFTER the reduction.
     pub slow_start_exit_cwnd: Option<usize>,
+    /// The reason slow start was exited. None if we haven't exited slow start or if we re-entered
+    /// after spurious congestion.
```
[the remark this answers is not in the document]
omansfeld: The `cwnd >= ssthresh` case Copilot is referencing can only be true if `ssthresh` is not it's default `MAX` value, i.e. has been set on a slow start exit. In that case the stat is also not `None` anymore.

neqo-transport/src/stats.rs
```diff
@@ -148,6 +156,16 @@ pub struct CongestionControlStats {
...
+    pub slow_start_exit_reason: Option<SlowStartExitReason>,
+    /// Number of times HyStart++ entered CSS (Conservative Slow Start). Only meaningful when
+    /// HyStart++ is enabled. Higher values indicate that HyStart++ had many spurious CSS
+    /// entries, spending more time throttling slow start growth.
+    pub css_entries: usize,
```
mxinden:
```suggestion
    pub hystart_css_entries: usize,
```

Why not prefix it?

neqo-transport/src/stats.rs
```diff
@@ -148,6 +156,16 @@ pub struct CongestionControlStats {
...
+    /// Number of CSS (Conservative Slow Start) rounds completed. Only meaningful when HyStart++ is
+    /// enabled. Higher values indicate the heuristic spent more time throttling slow start growth.
+    pub css_rounds_finished: usize,
```
mxinden:
```suggestion
    pub hystart_css_rounds_finished: usize,
```

PR #3434, opened 2026-03-11
feat: qlog@0.16.0 and fix silent metric drops

larseggert:
Update qlog to 0.16.0.

Remove the `_ => ()` catch-all from `metrics_updated` so the match is exhaustive.

That exposed that `Metric::InRecovery` and `Metric::MaxAckDelay` don't have corresponding fields in `MetricsUpdated`, so don't emit those.

### DIFF
--- a/.github/actions/check-vm/action.yml
+++ b/.github/actions/check-vm/action.yml
@@ -27,14 +27,14 @@ runs:
           set -ex
           pwd
           case "$PLATFORM" in
-            freebsd)    pkg install -y curl llvm nss pkgconf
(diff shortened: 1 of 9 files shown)


neqo-transport/src/qlog.rs
mxinden, line 295: Still valid?

larseggert: Yes, because clippy finds that we never construct these:
```rust
    PacketsInFlight(u64),
    PacingRate(u64),
```

PR #3435, opened 2026-03-11
feat: emit `pacing_rate` and `ssthresh` in qlog

larseggert:


neqo-transport/src/pace.rs
```diff
@@ -90,7 +90,10 @@ impl Pacer {
         let deficit =
             u128::try_from(packet - self.c).expect("packet is larger than current credit");
         let d = r.saturating_mul(deficit);
-        let add = d / u128::try_from(cwnd * Self::SPEEDUP).expect("usize fits into u128");
+        let divisor = u128::try_from(cwnd)
+            .expect("usize fits into u128")
+            .saturating_mul(u128::try_from(Self::SPEEDUP).expect("usize fits into u128"));
+        let add = d / divisor;
```
mxinden, line 96: Is this change related?

neqo-transport/src/pace.rs
```diff
@@ -104,6 +107,24 @@ impl Pacer {
...
+        let factor = (cwnd as u128).saturating_mul(Self::SPEEDUP as u128);
+        elapsed
+            .as_nanos()
+            .saturating_mul(factor)
+            .checked_div(rtt.as_nanos())
+    }
+
+    /// Compute the effective pacing rate in bytes per second.
+    ///
+    /// Returns `u64::MAX` if `rtt` is zero.
```
mxinden: Why not `None`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -341,6 +345,10 @@ impl<T: WindowAdjustment> CongestionControl for ClassicCongestionControl<T> {
             &[
                 qlog::Metric::CongestionWindow(self.current.congestion_window),
                 qlog::Metric::BytesInFlight(self.bytes_in_flight),
+                qlog::Metric::PacingRate(Pacer::rate(
+                    self.current.congestion_window,
+                    rtt_est.estimate(),
+                )),
```
mxinden: I find it confusing that we emit the pacing rate qlog in the congestion control module. Can this instead be done in `pace.rs`?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -718,13 +728,24 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
+                    if let Some(rtt) = self.last_rtt {
+                        qlog::metrics_updated(
+                            &mut self.qlog,
+                            &[
+                                cwnd,
+                                ssthresh,
+                                qlog::Metric::PacingRate(Pacer::rate(
+                                    self.current.congestion_window,
+                                    rtt,
+                                )),
```
mxinden: Why do this here and not in `pace.rs`? The information would be available in `Pacer::spend`.

neqo-transport/src/cc/classic_cc.rs
larseggert: The pacing rate (also) needs to be emitted when cwnd changes, not just on packet sends. Also, the pacer has currently no Qlog handle.

neqo-transport/src/pace.rs
larseggert: Fixes the same `usize` overflow-before-widening pattern that `new bytes_for()` fixes in `spend()`.

larseggert:
@mxinden could you take another look?

mxinden (commented):
I still don't think we should emit the `pacer.rs` qlogs in `classic_cc.rs`. I don't think our cc code should be aware of our pacer.

If moving these to `pacer.rs` isn't possible, how about instead moving them to `sender.rs`. `PacketSender` owns both the `pacer` as well as `cc` and thus has access to all relevant information in order to emit the pacing related qlogs.

larseggert:
Moving to `sender.rs` is a good idea!

neqo-transport/src/sender.rs
```diff
@@ -76,10 +80,15 @@ impl PacketSender {
...
             ),
+            qlog: Qlog::default(),
+            last_rtt: None,
+            last_pacing_rate: None,
         }
     }
 
     pub fn set_qlog(&mut self, qlog: Qlog) {
+        self.qlog = qlog.clone();
+        self.last_pacing_rate = None;
```
mxinden: Why is this needed? Isn't this already happening at initialization?

neqo-transport/src/sender.rs
```diff
@@ -30,6 +31,9 @@ pub const PACING_BURST_SIZE: usize = 2;
 pub struct PacketSender {
     cc: Box<dyn CongestionController>,
     pacer: Pacer,
+    qlog: Qlog,
+    last_rtt: Option<Duration>,
```
mxinden: Why do we need to store `last_rtt`? Just for `on_packets_lost` as the rtt measurement isn't available there? Why not either:

1. pass the rtt measurement to `on_packets_lost`
2. or don't emit a qlog in `on_packets_lost`, simply depending on the other call sites through `on_ack_received`.

neqo-transport/src/sender.rs
```diff
@@ -30,6 +31,9 @@ pub const PACING_BURST_SIZE: usize = 2;
...
+    last_pacing_rate: Option<u64>,
```
mxinden: Do I understand correctly that we don't want to emit a pacing qlog if the pace hasn't changed and thus need to track the last one? Is that required by the qlog spec? If not, is this optimization worth it? Can't this simply be de-duplicated on the visualization side?

larseggert: It can, but we'd log for every single packet, which is a lot.

mxinden: I assume we log on every single packet already anyways.

I am fine either way.

PR #3436, opened 2026-03-11
feat: add trigger to qlog `congestion_state_updated`

larseggert:


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -555,20 +558,20 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
         let old_state = self.current.phase;
-        // No need to tell qlog about exit from transient states.
-        if !old_state.transient() {
+        // Only emit a qlog event when a transition changes the qlog state.
+        if old_state.to_qlog() != phase.to_qlog() {
             self.qlog.add_event_at(
                 || {
-                    Some(EventData::CongestionStateUpdated(CongestionStateUpdated {
+                    Some(EventData::CongestionStateUpdated(CSU {
                         old: Some(old_state.to_qlog().to_owned()),
```
martinthomson: It looks like you can call the to_qlog() function one fewer time.  Or does that lead to borrow checker errors?  (I can't see how, as Phase should be Copy.)

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -555,20 +558,20 @@ impl<T: WindowAdjustment> ClassicCongestionControl<T> {
...
                         new: phase.to_qlog().to_owned(),
```
martinthomson: Same here.

larseggert:
@martinthomson I wrapped the `qlog` types like we do for other events.

PR #3437, opened 2026-03-11
feat: add `recovery:loss_timer_updated` qlog events

larseggert:


neqo-common/src/qlog.rs
```diff
@@ -128,14 +128,17 @@ impl Qlog {
...
-        if let Err(e) = f(&mut shared_streamer.streamer) {
-            log::error!("Qlog event generation failed with error {e}; closing qlog.");
-            // Set the inner Option to None to disable future logging for other references.
-            *borrow = None;
-            // Explicitly drop the RefCell borrow to release the mutable borrow.
-            drop(borrow);
-            // Set the outer Option to None to prevent future dereferences.
-            self.inner = None;
+        match f(&mut shared_streamer.streamer) {
+            Ok(()) | Err(qlog::Error::Done) => (),
```
mxinden, line 133: Why the special case on `qlog::Error::Done`? Why not set `borrow` and self.inner` to `None` in that case?

mxinden (approved):
Other than the seemingly unrelated change, this looks good to me.

larseggert:
> Other than the seemingly unrelated change, this looks good to me.

Sorry, which change is this?

neqo-common/src/qlog.rs
larseggert: Error::Done means "event was below the configured importance threshold". Terrible naming.

larseggert: We could leave the old code, since we now set the level to Extra by default, but it's more correct this way if we ever not set that.

mxinden: Filtering is fine by me. @larseggert mind adding a comment explaining what `qlog::Error::Done` implies?

mxinden:
> > Other than the seemingly unrelated change, this looks good to me.
> 
> Sorry, which change is this?

Was referring to the `qlog::Error::Done` case. Fine to merge with a comment explaining `Error::Done`.

PR #3439, opened 2026-03-11
fix: ConnectionClose error code, NewToken type, LH fields

larseggert:
`ConnectionClose` hardcoded `error_code_value` as 0; `NewToken` set token type to `retry`; long header packets omitted version/scid/dcid.

### DIFF
--- a/neqo-transport/src/qlog.rs
+++ b/neqo-transport/src/qlog.rs
@@ -478,7 +478,7 @@ impl From<Frame<'_>> for QuicFrame {
             },
             Frame::NewToken { token } => Self::NewToken {
                 token: qlog::Token {
-                    ty: Some(qlog::TokenType::Retry),
+                    ty: None,
                     details: None,
                     raw: Some(RawInfo {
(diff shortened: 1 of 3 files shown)


neqo-transport/src/qlog.rs
```diff
@@ -569,7 +569,7 @@ impl From<Frame<'_>> for QuicFrame {
                     CloseError::Application(_) => Some(ErrorSpace::ApplicationError),
                 },
                 error_code: Some(error_code.code()),
-                error_code_value: Some(0),
+                error_code_value: Some(error_code.code()),
```
mxinden, line 572: Would we ideally pass different values for `error_code` and `error_code_value` in the future?

mxinden (approved):

neqo-transport/src/qlog.rs
larseggert: I think `error_code_value` may not be even in the current draft anymore...

PR #3440, opened 2026-03-11
feat(qlog): add CC/performance events

larseggert:
- Add `recovery_parameters_set`, `connection_closed`, `packets_acked`, and `mtu_updated` events
- Populate `old` state in `connection_state_updated`; add `trigger` and `details` to `packet_dropped`; set `owner` in `connection_tparams_set`

### DIFF
--- a/neqo-transport/src/pmtud.rs
+++ b/neqo-transport/src/pmtud.rs
@@ -103,13 +103,13 @@ impl Pmtud {
     }
 
     fn set_mtu(&mut self, idx: usize, stats: &mut Stats, now: Instant) {
-        let old = self.plpmtu();
+        let old_mtu = self.plpmtu();
         self.mtu = self.search_table[idx];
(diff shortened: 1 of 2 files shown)


neqo-transport/src/pmtud.rs
```diff
@@ -93,6 +94,19 @@ impl Pmtud {
...
             raise_timer: None,
+            qlog: Qlog::disabled(),
+        }
+    }
+
+    pub fn set_qlog(&mut self, qlog: Qlog) {
+        self.qlog = qlog;
+    }
+
+    fn maybe_mtu_updated(&mut self, old: usize, now: Instant) {
```
mxinden: How about a `Self::set_mtu` function that also updates qlog? Easier than a helper function that only does the qlog tracking, no?

larseggert: Nice, that actually is more DRY.

PR #3443, opened 2026-03-12
Avoid sending datagrams in  PMTUD packets

jesup:


jesup:
> So copilot is right about this being a stack of changes, where only the last commit really addresses the problem.
> 
> This needs a test also.

Yeah, github sometimes gets confused and sets it up to be everything; I'll replace this

PR #3451, opened 2026-03-13
chore: Avoid `Box<dyn CongestionControl>`

larseggert:
For discussion - is this worth it?

(The extra dep could be avoided if we implement a dispatch macro ourselves.)

### DIFF
--- a/neqo-transport/src/cc/mod.rs
+++ b/neqo-transport/src/cc/mod.rs
@@ -128,21 +128,18 @@ pub enum CongestionControlImplementation {
     HyStartCubic(ClassicCongestionController<HyStart, Cubic>),
 }
 
-/// Expand a method call on `self` into a match that delegates to each variant's inner value.
 macro_rules! dispatch {
-    ($self:ident.$method:ident($($arg:expr),* $(,)?)) => {
-        match $self {
-            Self::ClassicNewReno(v) => v.$method($($arg),*),
-            Self::HyStartNewReno(v) => v.$method($($arg),*),
-            Self::ClassicCubic(v) => v.$method($($arg),*),
-            Self::HyStartCubic(v) => v.$method($($arg),*),
-        }
+    ($self:ident . $method:ident $args:tt) => {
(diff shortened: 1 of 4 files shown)


neqo-transport/src/cc/mod.rs
```diff
@@ -11,6 +11,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use enum_dispatch::enum_dispatch;
```
martinthomson: This seems like a fine thing to have here, but Firefox doesn't have this.  So you'd need to audit the code.

larseggert: Understood.

larseggert:
@mxinden worth it?

mxinden:
I assume the goal is an improvement in performance? Unless we have some indicator that this patch actually does improve performance, I don't think this patch is worth the added complexity.

larseggert:
Well, the transfer benches seem a bit faster. Let's see if the run after the rebase also shows +3%.

martinthomson (commented):
@larseggert do you want to push on this?  I've taken a look at enum_dispatch and it's pretty complicated, but I see no warning flags.  As it is only run at compile time, I think we're pretty safe, provided that someone spends a little more time than I did checking that the crate isn't poisoned somehow.

larseggert:
@martinthomson should we just define our own dispatch macro and not depend on `enum_dispatch`? I looked into it before and it was like ten lines or so.

martinthomson (approved):
Yeah, this is pretty reasonable.  A simpler proc macro than enum_dispatch would have made this easier, but we don't need to go to those lengths for just one of these.

mxinden:
> No significant performance differences relative to https://github.com/mozilla/neqo/commit/143849dbe01a6e2787a4cacb9161d566ec01cda4.

@larseggert is there a signal that this a performance improvement? Intuitively dynamic dispatch isn't always slower, e.g. stack size on static dispatch can have an impact. Is this worth the complexity it introduces?

PR #3464, opened 2026-03-16
Add WebTransport ExportKeyingMaterial backend

jesup:


neqo-crypto/tests/agent.rs
```diff
@@ -829,3 +829,113 @@ fn connection_fails_encoder_returned_too_long() {
...
+    let mut client = Client::new("server.example", true).expect("should create client");
+    let mut server = Server::new(&["key"]).expect("should create server");
+
+    connect(&mut client, &mut server);
+
+    let label = b"EXPORTER-test";
+    let result = client.export_keying_material(label, None, 32);
+    assert!(result.is_ok());
+    let material = result.unwrap();
+    assert_eq!(material.len(), 32);
```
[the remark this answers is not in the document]
larseggert: Copilot is correct here.

neqo-crypto/tests/agent.rs
```diff
@@ -829,3 +829,113 @@ fn connection_fails_encoder_returned_too_long() {
...
+    let mut server = Server::new(&["key"]).expect("should create server");
+
+    connect(&mut client, &mut server);
+
+    let label = b"EXPORTER-test";
+    let context = b"context-data";
+    let result = client.export_keying_material(label, Some(context), 32);
+    assert!(result.is_ok());
+    let material = result.unwrap();
+    assert_eq!(material.len(), 32);
```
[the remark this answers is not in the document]
larseggert: Copilot is correct here.

larseggert (approved):
LGTM, but has clippy and format issues. You could also be DRYer in the tests.

larseggert:
@jesup once this passes CI, can you rebase the stacked patches? I committed some changes to `main` (and rebased this PR) so that CI will then hopefully also run on the stacked patches.

neqo-http3/src/connection_client.rs
```diff
@@ -929,6 +929,30 @@ impl Http3Client {
...
+    pub fn export_keying_material(
+        &self,
+        label: &[u8],
+        context: Option<&[u8]>,
+        out_len: usize,
+    ) -> Res<Vec<u8>> {
+        self.conn
+            .export_keying_material(label, context, out_len)
+            .map_err(Error::Transport)
+    }
```
larseggert: This function has no test coverage.

neqo-transport/src/connection/mod.rs
```diff
@@ -902,6 +902,25 @@ impl Connection {
...
+    pub fn export_keying_material(
+        &self,
+        label: &[u8],
+        context: Option<&[u8]>,
+        out_len: usize,
+    ) -> Res<Vec<u8>> {
+        self.crypto
+            .tls()
+            .export_keying_material(label, context, out_len)
+            .map_err(Into::into)
```
larseggert, line 926: This delegates directly to the TLS agent without checking the QUIC connection state. The SecretAgent checks HandshakeState::is_connected(), but this only verifies the TLS handshake. Is that sufficient?

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -1824,3 +1824,44 @@ fn initial_crypto_retransmit_during_handshake_pto() {
         );
     }
 }
+
+#[test]
```
larseggert, line 1907: No test covers the edge case of requesting zero-length keying material.

neqo-http3/src/connection_client.rs
```diff
@@ -929,6 +929,30 @@ impl Http3Client {
...
+    /// the same for all WebTransport sessions on this connection.
+    ///
+    /// # Arguments
+    /// * `label` - The exporter label
+    /// * `context` - Optional context data
+    /// * `out_len` - Length of output keying material
+    ///
+    /// # Errors
+    ///
+    /// Returns `TransportError` if the connection is not ready or export fails.
```
larseggert:
```suggestion
    /// Returns `Error::Transport` if the connection is not ready or export fails.
```
An h3 Error.

jesup:
Note: I have some changes based on @martinthomson 's comments in PR #3428 which I'll be adding

larseggert:
@jesup last commits included `#agent.rs#`, which is probably a temp file?

jesup:
This PR is part of a stack of 17 bookmarks:

1. `trunk()`
1. **users/jesup/exportkey ← this PR**
1. [users/jesup/draining](https://github.com/mozilla/neqo/pull/3465)
1. [users/jesup/sub_protocol](https://github.com/mozilla/neqo/pull/3466)
1. [users/jesup/sendGroup](https://github.com/mozilla/neqo/pull/3467)

---

neqo-crypto/src/agent.rs
```diff
@@ -802,6 +802,47 @@ impl SecretAgent {
...
+        }
+        let mut out = vec![0u8; out_len];
+
+        let has_context = !context.is_empty();
+        secstatus_to_res(unsafe {
+            ssl::SSL_ExportKeyingMaterial(
+                self.fd,
+                label.as_ptr().cast(),
+                c_uint::try_from(label.len())?,
+                PRBool::from(has_context),
```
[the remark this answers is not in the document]
martinthomson: Please make the suggested change:


```suggestion
                PRBool::from(true),  // hasContext: In TLS 1.3 context is always present.
```

neqo-crypto/src/agent.rs
```diff
@@ -802,6 +802,47 @@ impl SecretAgent {
...
+        if !self.state.is_connected() {
+            return Err(Error::InvalidState);
+        }
+
+        if out_len == 0 {
+            return Err(Error::InvalidState);
+        }
+        let mut out = vec![0u8; out_len];
+
+        let has_context = !context.is_empty();
```
```suggestion
```

See below.

neqo-http3/src/connection_client.rs
```diff
@@ -937,6 +937,45 @@ impl Http3Client {
...
+    /// correspond to an active WebTransport session,
+    /// `Error::InvalidInput` if the label or context exceed 255
+    /// bytes, or `Error::Transport` if the connection is not ready or
+    /// the TLS export fails.
+    pub fn webtransport_export_keying_material(
+        &self,
+        session_id: StreamId,
+        label: &[u8],
+        context: &[u8],
+        out_len: usize,
```
martinthomson: Here's a thing for `out_len`.  I might prefer instead that this is passed `&mut [u8]` instead because then the memory allocation is done by the caller.

Mainly because I think that -- without that change -- this is the point where we would want to put a limit on the size of buffer that can be returned.  Obviously, the exporter interface in the browser needs that as well, but we can maybe move that responsibility to the API layer, closer to where the API is defined.

From my perspective, 4k of export is way more than anyone should be requesting, so we can cap at that (or lower; frankly, anything over 256 bits is probably a bit much, but maybe we can avoid having callers invoke their own KDF if we give them at least a little more than that).  I'll open a PR on the spec to recommend a limit and a minimum for the cap.

martinthomson (approved):
A few changes to make here (I also took a suggestion from the bot, watch for that), but none of them critical.

larseggert:
@jesup I think this needs to target https://github.com/mozilla/nss-rs now, right?

martinthomson:
Only partially, but yes.  See also https://github.com/w3c/webtransport/issues/756

jesup:
> @jesup I think this needs to target https://github.com/mozilla/nss-rs now, right?

Yes, I've moved it to that and created a pr there

jesup:
Feels like whack-a-mole with the copilot suggestions.  It would be nicer if they found all they're going to find on the first (or second) pass...

jesup:
Note that tests will fail until the nss-rs PR lands: https://github.com/mozilla/nss-rs/pull/41

neqo-http3/tests/webtransport.rs
```diff
@@ -497,3 +497,145 @@ fn wt_session_ok_and_wt_datagram_in_same_udp_datagram() {
...
+    let (mut client, mut server) = connect();
+    let wt_session = create_wt_session(&mut client, &mut server);
+    let session_id = wt_session.stream_id();
+
+    let long_label = vec![0u8; 256];
+    assert!(
+        client
+            .webtransport_export_keying_material(session_id, &long_label, b"ctx", &mut [0u8; 32])
+            .is_err()
+    );
```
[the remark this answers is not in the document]
martinthomson: There's a formatting error in this suggestion, but it's a good one.

neqo-http3/src/connection.rs
```diff
@@ -1440,6 +1434,20 @@ impl Http3Connection {
...
+        &mut self,
+        conn: &mut Connection,
+        session_id: StreamId,
+        stream_type: StreamType,
+        send_events: Box<dyn SendStreamEvents>,
+        recv_events: Box<dyn RecvStreamEvents>,
+    ) -> Res<StreamId> {
+        qtrace!("Create new WebTransport stream session={session_id} type={stream_type:?}");
+
+        let wt = self.validate_wt_session(session_id)?;
```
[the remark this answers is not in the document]
jesup: This code is already in the patch....???

jesup:
github says there's one requested change from github-actions bot.  I'm not sure what requested change it wants resolved; right now there are none obvious showing (though it may add some later).  All issues show as Resolved right now.

neqo-http3/src/connection_client.rs
```diff
@@ -937,6 +937,52 @@ impl Http3Client {
...
+        }
+        let label_len = u8::try_from(label.len()).map_err(|_| Error::InvalidInput)?;
+        let context_len = u8::try_from(context.len()).map_err(|_| Error::InvalidInput)?;
+
+        let mut wt_context = Encoder::with_capacity(
+            Encoder::varint_len(session_id.as_u64()) + 1 + label.len() + 1 + context.len(),
+        );
+        wt_context.encode_varint(session_id.as_u64());
+        wt_context.encode_byte(label_len);
+        wt_context.encode(label);
```
martinthomson: You should use `encode_vec()` for this.  To ensure that you don't panic, you can check lengths with:

```rust
if out.is_empty() || label.len() > 255 || context.len() > 255 {
  return Err(Error::InvalidInput);
}
```

neqo-http3/src/server_events.rs
```diff
@@ -380,6 +380,38 @@ impl WebTransportRequest {
...
+
+    /// Export keying material for this WebTransport session
+    /// (draft-ietf-webtrans-http3 §4.8).
+    ///
+    /// # Errors
+    ///
+    /// Returns `Error::InvalidInput` if `out` is empty or `label`/`context`
+    /// exceed 255 bytes, or `Error::Transport` if the connection is not ready
+    /// or the TLS export fails.
+    pub fn export_keying_material(&self, label: &[u8], context: &[u8], out: &mut [u8]) -> Res<()> {
```
martinthomson, line 396: I'm seeing double!

Given that this is identical code, I think that you can lift a `wt_export_keying_material()` function to the connection.

If that makes you feel awkward, you could do this:

```rust
trait WebtransportExporter {
  pub fn export_keying_material(&self, label: &[u8], context: &[u8], out: &mut [u8]) -> Res<()>;
}
impl WebtransportExporter for Connection {
  pub fn export_keying_material(&self, label: &[u8], context: &[u8], out: &mut [u8]) -> Res<()> {
    // code goes here
  }
}

File: neqo-transport/src/connection/mod.rs [hunk 14]
```diff
@@ -902,6 +902,32 @@ impl Connection {
...
+        if !(self.state.connected() || self.state.closing()) {
+            return Err(Error::NotConnected);
+        }
+        if out.is_empty() {
+            return Err(Error::InvalidInput);
+        }
+        // RFC 8446 §7.5: labels MUST start with "EXPORTER-".
+        if !label.starts_with(b"EXPORTER-") {
+            return Err(Error::InvalidInput);
+        }
```
martinthomson: IMPORTANT: This claim is completely false.  You've been hallucinated on.

RFC 5705 says: In order to prevent this, labels SHOULD begin with "EXPORTER".

Please remove this check.  It's enough to check that the label is non-empty.

martinthomson (dismissed):

neqo-http3/src/connection_client.rs
```diff
@@ -937,6 +937,52 @@ impl Http3Client {
...
+        wt_context.encode_byte(context_len);
+        wt_context.encode(context);
+
+        self.conn
+            .export_keying_material(b"EXPORTER-WebTransport", wt_context.as_ref(), out)
+            .map_err(|e| match e {
+                neqo_transport::Error::InvalidInput => Error::InvalidInput,
+                other => Error::Transport(other),
```
larseggert: These error cases don't have test coverage.

neqo-http3/src/server_events.rs
```diff
@@ -380,6 +380,38 @@ impl WebTransportRequest {
...
+        wt_context.encode_byte(context_len);
+        wt_context.encode(context);
+
+        self.stream_handler
+            .conn
+            .borrow()
+            .export_keying_material(b"EXPORTER-WebTransport", wt_context.as_ref(), out)
+            .map_err(|e| match e {
+                neqo_transport::Error::InvalidInput => Error::InvalidInput,
+                other => Error::Transport(other),
```
larseggert: Error cases don't have test coverage.

neqo-http3/src/connection.rs
```diff
@@ -1440,6 +1434,20 @@ impl Http3Connection {
         if !wt.borrow().is_active() {
             return Err(Error::InvalidStreamId);
```
larseggert, line 1441: No test coverage.

larseggert (commented):
Some text coverage gaps on error conditions. Tests could benefit from some helpers to make them more DRY.

neqo-transport/src/connection/mod.rs
```diff
@@ -902,6 +902,28 @@ impl Connection {
...
+    /// Export keying material per RFC 5705/8446.
+    ///
+    /// This can only be called after the handshake is complete.
+    /// Per RFC 8446 §7.5, labels SHOULD begin with `"EXPORTER-"`.
+    ///
+    /// # Errors
+    ///
+    /// Returns error if connection is not in a connected or closing state,
+    /// or export fails.
+    pub fn export_keying_material(&self, label: &str, context: &[u8], out: &mut [u8]) -> Res<()> {
```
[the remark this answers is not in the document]
jesup: github-actions is treating these parameters as if they're the same "label" passed through layers, but they're not. The HTTP/3 "label" becomes part of the context bytes, not the TLS EKM label (which is a string, and we pass a fixed string for it). No change is needed here — the current types correctly reflect the underlying semantics

jesup:
> **Main actionable item:** `webtransport_create_stream_remote` still inlines the session-lookup logic and skips the `is_active()` check. Since `validate_wt_session` was extracted to centralize this, it should be used there too (see inline comment).

This can't use validate_wt_session directly. The session has four states: Negotiating, Active, FinPending, Done. The race condition test covers a legitimate case: a remote stream arrives while the session is still Negotiating (stream data before session acceptance datagram). validate_wt_session only allows Active, so it rejects the stream.

  The asymmetry is intentional:
  - Local (_local): must require Active — we shouldn't initiate actions on unconfirmed sessions
  - Remote (_remote): must allow Negotiating too — the peer can already be sending streams

jesup:
@martinthomson all items had been resolved so re-requested review

neqo-http3/src/lib.rs
```diff
@@ -408,6 +408,31 @@ impl From<QpackError> for Error {
     }
 }
 
+fn webtransport_export_keying_material(
```
martinthomson: I'm going to suggest again that you look at moving this to a decorator trait.  Putting code like this at the top level of the crate is a bit of pollution of the namespace that we don't need.

The pattern to follow is:

```rust
trait WebTransportExport {
  fn webtransport_export_keying_material(&self, ... ) -> Res<()>;
}
impl WebTransportExport for Connection {
  fn webtransport_export_keying_material(&self, ... ) -> Res<()> { ... }
}
```

For this, I would implement that in neqo-http3/src/features/extended_connect/webtransport_session.rs and then import the trait where the implementation is needed.

martinthomson (approved):
Code is good, tests are good.  Just one piece of structural feedback.

mxinden:
Merge failed due to unrelated CI failure. I am looking into it.

mxinden:
> Merge failed due to unrelated CI failure. I am looking into it.

For the record, fixed by @larseggert in https://github.com/mozilla/actions/pull/87.

PR #3466, opened 2026-03-16
Get the negotiated protocol to support WebTransport

jesup:

PR #3465, opened 2026-03-16
Add support for GOAWAY draining events to support WebTransport

jesup:

PR #3467, opened 2026-03-16
WebTransport sendGroup support (and improve sendOrder)

jesup:


mxinden:
Given the large merge conflicts, can you do another rebase before review? Can also review now, though I am afraid comments might get lost.

larseggert:
It needs a rebase so CI runs

jesup:
This PR is part of a stack of 17 bookmarks:

1. `trunk()`
1. [users/jesup/exportkey](https://github.com/mozilla/neqo/pull/3464)
1. [users/jesup/draining](https://github.com/mozilla/neqo/pull/3465)
1. [users/jesup/sub_protocol](https://github.com/mozilla/neqo/pull/3466)
1. **users/jesup/sendGroup ← this PR**

---

neqo-http3/src/features/extended_connect/send_group.rs
```diff
@@ -0,0 +1,80 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use neqo_transport::StreamId;
+
+static NEXT_SEND_GROUP_ID: AtomicU64 = AtomicU64::new(1);
```
larseggert: Why does this need to be global?

neqo-http3/src/features/extended_connect/send_group.rs
```diff
@@ -0,0 +1,80 @@
...
+
+#[expect(
+    clippy::module_name_repetitions,
+    reason = "SendGroupId is the canonical name for this type"
+)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct SendGroupId(u64);
```
larseggert: You _could_ call it `Id` and use it as `send_group::Id`...

neqo-http3/src/features/extended_connect/send_group.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub const fn as_u64(self) -> u64 {
+        self.0
+    }
+}
+
+impl Default for SendGroupId {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
larseggert: Just implement `default()` instead of `new()` then?

neqo-http3/src/features/extended_connect/send_group.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl TryFrom<u64> for SendGroupId {
+    type Error = ();
+
+    fn try_from(id: u64) -> Result<Self, Self::Error> {
+        if id == 0 { Err(()) } else { Ok(Self(id)) }
+    }
+}
```
larseggert: `SendGroupId` wraps `u64`, how can this ever fail; why do we need `TryFrom`?

neqo-transport/src/send_stream.rs
```diff
@@ -1640,28 +1808,42 @@ impl SendStreams {
...
+        self.per_group.clear();
+        self.per_group_next = 0;
     }
 
     pub fn remove_terminal(&mut self) {
+        // Collect grouped streams that need removal from per-group queues separately,
+        // since we can't borrow self.per_group while self.map.retain is running.
+        let mut grouped_removals: Vec<(SendGroupId, StreamId, Option<SendOrder>)> = Vec::new();
+
         self.map.retain(|stream_id, stream| {
```
larseggert: Can you use `extract_if` instead of `retain` to get around the double loop?

neqo-transport/src/send_stream.rs
```diff
@@ -1713,24 +1895,53 @@ impl SendStreams {
...
+        // groups get equal bandwidth share regardless of differing sendOrder values between
+        // groups.  Within a group, the highest-sendOrder stream is served first (starving
+        // lower sendOrder within the same group), matching the spec starvation requirement.
+        let num_groups = self.per_group.len();
+        if num_groups > 0 {
+            if self.per_group_next >= num_groups {
+                self.per_group_next = 0;
+            }
+            let start = self.per_group_next;
+            for i in 0..num_groups {
```
larseggert: Two nested loops on every `write_frame` make me wonder about the performance impact of this.

larseggert (dismissed):
Lots of new code without test coverage. Tests would benefit from DRY helpers.

neqo-http3/src/connection_client.rs
```diff
@@ -896,6 +896,39 @@ impl Http3Client {
...
+        sendgroup: SendGroupId,
+    ) -> Res<()> {
+        // Update the HTTP3-layer stream record (validates ownership against the session).
+        self.base_handler
+            .stream_set_sendgroup(stream_id, sendgroup)?;
+        // Update the transport-layer scheduler so sendOrder is namespaced per group.
+        self.conn
+            .stream_sendgroup(stream_id, Some(sendgroup.as_u64()))
+            .map_err(|_| Error::InvalidStreamId)
+    }
```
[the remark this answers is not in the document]
jesup: There are issues no matter which you resolve first, but it's better to do it the way that was requested here and document.  It's not a practical problem.

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
     Http3State, Priority, SessionAcceptAction, WebTransportEvent, WebTransportServerEvent,
     features::extended_connect::{
         CloseReason,
+        send_group::Id as SendGroupId,
```
larseggert: I would just use leave it as `send_group::Id` in the code (clearer to me), but YMMV.

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -603,6 +604,67 @@ fn wt_session_protocol_malformed_unquoted_rejected() {
...
+    let mut wt = WtTest::new();
+    let wt_session = wt.create_wt_session();
+    let session_id = wt_session.stream_id();
+
+    let group = SendGroupId::default();
+    let result = wt
+        .client
+        .webtransport_register_send_group(session_id, group);
+    assert!(result.is_ok());
+    assert!(group.as_u64() > 0);
```
larseggert: Most of the tests below duplicate a variation of this, a helper would be more DRY. (Ask Claude to DRY up these tests for you.)

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -603,6 +604,67 @@ fn wt_session_protocol_malformed_unquoted_rejected() {
...
+    // Test that we can create a send group for a WebTransport session.
+    let mut wt = WtTest::new();
+    let wt_session = wt.create_wt_session();
+    let session_id = wt_session.stream_id();
+
+    let group = SendGroupId::default();
+    let result = wt
+        .client
+        .webtransport_register_send_group(session_id, group);
+    assert!(result.is_ok());
```
larseggert: You could also just `unwrap()` `result`.

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -634,3 +696,518 @@ fn wt_session_protocol_non_webtransport_session() {
...
+        .webtransport_register_send_group(session_id1, group1)
+        .unwrap();
+
+    // Try to create stream in session2 with group from session1 - should fail
+    let result = wt.client.webtransport_create_stream_with_send_group(
+        wt_session2_id,
+        StreamType::UniDi,
+        Some(group1),
+    );
+    assert!(result.is_err());
```
larseggert: Assert the actual expected error.

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
```diff
@@ -634,3 +696,518 @@ fn wt_session_protocol_non_webtransport_session() {
...
+    // Without group-level fairness the scheduler would prefer group_high streams
+    // globally and completely starve group_low.  With per-group round-robin, the
+    // server receives Data events from both groups interleaved throughout the transfer.
+    //
+    // We verify fairness by checking that the FIRST group_low Data event arrives
+    // before the LAST group_high Data event — proving they overlap in delivery
+    // (group_low is not held back until after group_high finishes).
+    //
+    // DATA_SIZE must produce multiple packets per stream so the interleaving is visible:
+    // 10 KB / ~1440 B MTU ≈ 7 packets per stream → 14+ Data events per group.
```
larseggert: Aren't we using the 1280-byte default packet size here?

neqo-http3/src/features/extended_connect/send_group.rs
```diff
@@ -0,0 +1,36 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::cell::Cell;
+
+thread_local! {
+    static NEXT_ID: Cell<u64> = const { Cell::new(1) };
+}
```
larseggert: We do we need a global variable here; does this need to be unique across all connections ever?

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -56,8 +59,29 @@ impl Session {
...
         }
     }
+    /// Register a send group with a caller-provided ID for this session.
+    ///
+    /// Returns an error if the ID is already in use.
+    pub(crate) fn register_send_group(&mut self, id: SendGroupId) -> Res<()> {
+        if !self.send_groups.insert(id) {
+            return Err(Error::InvalidState);
+        }
+        Ok(())
```
larseggert:
```suggestion
        self.send_groups.insert(id).then_some(()).ok_or(Error::InvalidState)
```

neqo-http3/src/features/extended_connect/webtransport_session.rs
larseggert: We use `rustc_hash::FxHashSet as HashSet` in neqo.

neqo-transport/src/send_stream.rs
```diff
@@ -1556,21 +1611,26 @@ pub struct SendStreams {
...
+    // path.  Lets that path iterate `map` by index (cache-friendly, no per-stream
+    // hash lookup) while still resuming after the last-served stream when the packet
+    // builder fills mid-pass, preserving fairness.
+    fair_rr_next: usize,
 }
 
+/// Key used in `per_group` to represent the null sendGroup (ungrouped fair streams).
+/// Real `SendGroupId` values start at 1 (see `neqo-http3` `send_group.rs`), so 0 is safe
+/// as a sentinel here.
+const NULL_GROUP_ID: SendGroupId = 0;
```
larseggert: If you used a separate field for ungrouped streams, you could save the useless map lookup for them and avoid the magic value.

neqo-transport/src/send_stream.rs
```diff
@@ -1600,11 +1660,88 @@ impl SendStreams {
...
+            if let Some(grp_queues) = self.per_group.get_mut(&NULL_GROUP_ID) {
+                grp_queues.remove_stream(stream_id, old_sendorder);
+                if grp_queues.is_empty() {
+                    self.per_group.shift_remove(&NULL_GROUP_ID);
+                    if self.per_group_next >= self.per_group.len() {
+                        self.per_group_next = 0;
+                    }
+                }
+            }
+        }
```
larseggert:
```suggestion
		if let Some(gid) = old_group.or_else(|| was_fair.then_some(NULL_GROUP_ID)) {
		    if let Some(grp_queues) = self.per_group.get_mut(&gid) {
        		grp_queues.remove_stream(stream_id, old_sendorder);
		        if grp_queues.is_empty() {
        		    self.per_group.shift_remove(&gid);
		            if self.per_group_next >= self.per_group.len() {
        		        self.per_group_next = 0;
		            }
		        }
		    }
		}
```

neqo-transport/src/send_stream.rs
```diff
@@ -1600,11 +1660,88 @@ impl SendStreams {
...
+                .group_mut(old_sendorder)
+                .insert(stream_id);
+        } else if was_fair {
+            // Move back to the null-group (ungrouped) slot.
+            self.per_group
+                .entry(NULL_GROUP_ID)
+                .or_default()
+                .group_mut(old_sendorder)
+                .insert(stream_id);
         }
```
larseggert:
```suggestion
		if let Some(gid) = group_id.or_else(|| was_fair.then_some(NULL_GROUP_ID)) {
		    self.per_group
		        .entry(gid)
		        .or_default()
		        .group_mut(old_sendorder)
		        .insert(stream_id);
		}

```

neqo-transport/src/send_stream.rs
```diff
@@ -1644,37 +1795,44 @@ impl SendStreams {
...
-            // To be safe we can try to insert at the end and if not
-            // fall back to binary-search insertion
-            if matches!(self.regular.stream_ids().last(), Some(last) if stream_id > *last) {
-                self.regular.push(stream_id);
-            } else {
-                self.regular.insert(stream_id);
+            // A newly fair stream with no send_group goes to the null-group slot in per_group,
+            // so it participates in round-robin alongside all explicit sendGroups (spec: equal).
+            // Streams with a send_group are managed by set_sendgroup; don't add here.
+            if send_group.is_none() {
```
larseggert, line 1778: Similar simplification as above.

neqo-transport/src/send_stream.rs
```diff
@@ -1644,37 +1795,44 @@ impl SendStreams {
...
+                }
+            } else if let Some(grp_queues) = self.per_group.get_mut(&NULL_GROUP_ID) {
+                grp_queues.remove_stream(stream_id, sendorder);
+                if grp_queues.is_empty() {
+                    self.per_group.shift_remove(&NULL_GROUP_ID);
+                    if self.per_group_next >= self.per_group.len() {
+                        self.per_group_next = 0;
+                    }
+                }
+            }
```
larseggert:
```suggestion
			let gid = send_group.unwrap_or(NULL_GROUP_ID);
			if let Some(grp_queues) = self.per_group.get_mut(&gid) {
			    grp_queues.remove_stream(stream_id, sendorder);
			    if send_group.is_none() && grp_queues.is_empty() {
			        self.per_group.shift_remove(&NULL_GROUP_ID);
			        if self.per_group_next >= self.per_group.len() {
			            self.per_group_next = 0;
			        }
			    }
			}
```

neqo-transport/src/send_stream.rs
larseggert, line 1881: Any chance we can take the opportunity to refactor this logic since we're touching it now? It's pretty complex.

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
jesup: I'm good with that

neqo-http3/src/features/extended_connect/tests/webtransport/sessions.rs
jesup: yes, thanks

neqo-http3/src/features/extended_connect/send_group.rs
jesup: It does not need to be globally unique.  However: per-session unique is not sufficient, the transport's group map is per-connection, keyed by the raw u64: conn.stream_sendgroup(stream_id, group_id.as_u64()) → SendStreams::per_group: IndexMap<u64, _> (send_stream.rs:1619). That map is shared by all WebTransport sessions on the same H3 connection.
It would need to be per-connection unique.  I can do that if you want by adding a create-and-register call that returns the id in the transport code.  The current impl is overly-unique, and there's no risk of running out of Ids, so my inclination here is to leave it as-is and just add a comment.

larseggert: So maybe handle it like we handle CIDs?

neqo-transport/src/send_stream.rs
jesup: There are some upsides cognitively, but:
The map lookup has no real impact perf-wise; the hot path uses index access.
The Fairness round-robin code gets noticeably more complex and there's some real risk of (re-)introducing some subtle starvation or fairness-cursor ordering bugs.
The later patch for datagrams uses the same pattern for similar reasons, which helps with maintenance.  
I'd like to leave it as-is: The "null group is just another entry" design is elegant because it avoids special-casing fairness - the one place this code is genuinely tricky.  The magic value is contained, documented, and symmetric across the stream and datagram paths. The payoff (a cold-path hash + slightly nicer types) doesn't justify the fairness-cursor risk, the test churn, and the stream/datagram divergence.

neqo-transport/src/send_stream.rs
jesup: I'm dealing with this slightly differently to also clean up freshly-empty groups

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -603,6 +616,15 @@ pub(crate) trait Protocol: Debug + Display {
         None
     }
 
+    fn register_send_group(&mut self, _id: SendGroupId) -> Res<()> {
+        Err(Error::InvalidStreamId)
```
mxinden, line 619:
```suggestion
        debug_assert!(false, "register_send_group called, but not supported by protocol");
        Err(Error::InvalidStreamId)
```

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -603,6 +616,15 @@ pub(crate) trait Protocol: Debug + Display {
...
+    }
+
+    fn validate_send_group(&self, _group_id: SendGroupId) -> bool {
+        // Default implementation returns false
+        false
```
mxinden, line 623:
```suggestion
        debug_assert!(false, "validate_send_group, but not supported by protocol");
        false
```

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -603,6 +616,15 @@ pub(crate) trait Protocol: Debug + Display {
         None
     }
 
+    fn register_send_group(&mut self, _id: SendGroupId) -> Res<()> {
+        Err(Error::InvalidStreamId)
+    }
+
+    fn validate_send_group(&self, _group_id: SendGroupId) -> bool {
+        // Default implementation returns false
```
```suggestion
```

Nit: fairly obvious from the implementation :)

neqo-http3/src/connection.rs
```diff
@@ -1514,9 +1576,52 @@ impl Http3Connection {
...
-            send_events,
-            recv_events,
+            (send_events, recv_events),
+            true,
+            None,
+        )?;
+        Ok(stream_id)
+    }
+
+    pub(crate) fn webtransport_create_stream_local_with_send_group(
```
mxinden, line 1464: Can some of this be de-duplicated with `webtransport_create_stream_local`? If not, can more of the two be moved into `webtransport_create_stream_internal`?

neqo-http3/src/connection_client.rs
```diff
@@ -896,6 +896,42 @@ impl Http3Client {
         Http3Connection::stream_set_sendorder(&mut self.conn, stream_id, sendorder)
     }
 
+    /// Sets the `SendGroup` for a given WebTransport stream
```
mxinden:
```suggestion
    /// Sets the [`SendGroupId`] for a given WebTransport stream
```

Is `SendGroup` a type? If not, why the backticks? 

Also, whereever possible, please use Rust doc links. Those are checked in CI, thus if docs ever get outdated, we get a failing CI.

neqo-http3/src/lib.rs
```diff
@@ -439,6 +440,14 @@ trait Stream: Debug {
     fn session_protocol(&self) -> Option<String> {
         None
     }
+
+    fn register_send_group(&mut self, _id: SendGroupId) -> Res<()> {
+        Err(Error::InvalidStreamId)
```
mxinden, line 451: `debug_assert!`? I assume this executing is always a bug? Same below?

neqo-transport/src/streams.rs
```diff
@@ -34,6 +34,7 @@ use crate::{
 };
 
 pub type SendOrder = i64;
+pub type SendGroupId = u64;
```
mxinden: Instead of a type alias, can this be a new type? That would prevent misuse at compile time.

neqo-http3/src/features/extended_connect/send_group.rs
mxinden: Strongly against the `thread_local` _global_. What if we ever decide to run the same Neqo connection on a thread pool?

E.g. `Http3Connection` could manage `Id`s.

neqo-transport/src/send_stream.rs
```diff
@@ -1600,41 +1659,118 @@ impl SendStreams {
...
+        // Extract the info we need before any other mutable borrows.
+        let (was_fair, old_sendorder, old_group) = {
+            let stream = self.map.get(&stream_id).ok_or(Error::InvalidStreamId)?;
+            (stream.is_fair(), stream.sendorder(), stream.send_group())
+        };
+
+        if old_group == group_id {
+            return Ok(());
+        }
+        debug_assert!(
```
mxinden: In release mode this would bring us into an inconsistent state, right? Why not `assert!`?


```suggestion
        assert!(
```

neqo-transport/src/send_stream.rs
```diff
@@ -1645,37 +1781,34 @@ impl SendStreams {
...
-                    .get_mut(&sendorder)
-                    .ok_or(Error::Internal)?
-            } else {
-                &mut self.regular
-            };
-            group.remove(stream_id);
+            // Remove from whichever queue currently owns this stream: an explicit
+            // group, or the null-group (ungrouped) slot.
+            let gid = send_group.unwrap_or(NULL_GROUP_ID);
+            self.remove_from_group(gid, stream_id, sendorder);
```
mxinden, line 1798: Does this need to update `stream.send_group` as well given that the stream just got removed from the group?

mxinden (dismissed):
Requesting changes for the use of `thread_local`. Otherwise my review is not blocking.

> transfer/1-conn/1-100mb-req (aka. Upload): 💔 Performance has regressed by +9.8790%.

Anything above 3% I consider a signal instead of noise. Worth looking into?

jesup:
> > transfer/1-conn/1-100mb-req (aka. Upload): 💔 Performance has regressed by +9.8790%.
> 
> Anything above 3% I consider a signal instead of noise. Worth looking into?

Testing locally via Claude: 
```
  ┌──────────────────────────────┬─────────────────────────────────────────────────────────────┐
  │            Build             │                         Upload time                         │
  ├──────────────────────────────┼─────────────────────────────────────────────────────────────┤
  │ HEAD-vs-HEAD (3 runs)        │ 113.3 / 114.3 / 113.1 ms — ±5% noise, stable point estimate │
  ├──────────────────────────────┼─────────────────────────────────────────────────────────────┤
  │ Pre-series baseline dd501143 │ ~115 ms                                                     │
  ├──────────────────────────────┼─────────────────────────────────────────────────────────────┤
  │ HEAD vs baseline (criterion) │ No change detected: [-4.83% −0.70% +3.37%], p=0.74          │
  └──────────────────────────────┴─────────────────────────────────────────────────────────────┘
```

  HEAD is statistically indistinguishable from (if anything slightly faster than) the code from before the entire WebTransport/sendGroup series.

Why the report showed +9.8%: the reported confidence interval was [+4.39%, +16.15%] with 21% high-severe outliers -- this reads as a noise signature. The neqo-bin upload benchmark spawns a real server thread and transfers 100 MB over localhost UDP per iteration, so it's inherently high-variance, and that CI run happened to land on the high side. Note the throughput numbers also differ wildly from mine (~558 MiB/s reported vs ~880 MiB/s here), confirming a much noisier environment.

jesup:
And just for confirmation a bunch more: 
```
  ┌─────────┬─────────────────────┬────────────────────────────┐
  │   Run   │ change (point est.) │          verdict           │
  ├─────────┼─────────────────────┼────────────────────────────┤
  │ initial │ −0.70%              │ No change                  │
  ├─────────┼─────────────────────┼────────────────────────────┤
  │ 1       │ −0.30%              │ No change                  │
  ├─────────┼─────────────────────┼────────────────────────────┤
  │ 2       │ −0.09%              │ No change                  │
  ├─────────┼─────────────────────┼────────────────────────────┤
  │ 3       │ +0.02%              │ No change                  │
  ├─────────┼─────────────────────┼────────────────────────────┤
  │ 4       │ −3.92%              │ Within noise (HEAD faster) │
  └─────────┴─────────────────────┴────────────────────────────┘
```

neqo-http3/src/connection_client.rs
jesup: Done, and for all the other patches too

neqo-transport/src/send_stream.rs
jesup: Instead of assert!(), it should return an error I think (InvalidInput).

neqo-transport/src/send_stream.rs
jesup: Added a fix and a test, verified the test

jesup: Note the bug here was latent; it couldn't happen in practice, but future changes could have exposed it

jesup:
>Strongly against the thread_local global. What if we ever decide to run the same Neqo connection on a thread pool?
>
>E.g. Http3Connection could manage Ids.

Done

neqo-transport/src/send_stream.rs
jesup: done

jesup:
The codecov % is slightly lower because there are some unreachable functions meant to catch future errors.  They're marked with codecov(off), but still apparently count.  There's also a dead-code empty function for use in later patches.  There's no functional code that's missed

larseggert:
I still think you can make the tests less repetitive via some more simple helpers, and by looping instead of repeating statements verbatim.

neqo-http3/src/connection_client.rs
```diff
@@ -896,6 +896,42 @@ impl Http3Client {
...
 
+    /// Sets the [`SendGroupId`] for a given WebTransport stream.
+    ///
+    /// # Errors
+    ///
+    /// It may return [`Error::InvalidStreamId`] if a stream does not exist anymore.
+    pub fn webtransport_set_sendgroup(
+        &mut self,
+        stream_id: StreamId,
+        sendgroup: SendGroupId,
```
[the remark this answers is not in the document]
jesup: There was a reason for ? - stream gone vs not a webtransport stream.  Added comments to document this.

neqo-transport/src/send_stream.rs
```diff
@@ -1796,27 +1936,126 @@ impl SendStreams {
...
                 }
             }
+        } else {
+            if self.per_group_next >= num_groups {
+                self.per_group_next = 0;
+            }
+            let start = self.per_group_next;
+            // Split borrows on disjoint fields so we can access both per_group (for
+            // priority ordering) and map (for the stream itself) in the same loop body
+            // without an intermediate Vec.
```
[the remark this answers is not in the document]
jesup: Ok, but no change is needed now

jesup:
I don't have permissions to acknowledge them, but the CodSpeed remove_expired and coalesce regressions aren't real; those didn't change - likely due to rebuild/LTO

larseggert:
@jesup needs a rebase.

mxinden:
@jesup do we need to prevent a regular stream to starve a high send order stream within the same send group? If so, is the following a valid test failure? Confirmed, fails on this branch.

https://github.com/mxinden-bot/neqo/commit/fb820ca754b7be81af29b8235488b0e885a6e82b

jesup:
> @jesup do we need to prevent a regular stream to starve a high send order stream within the same send group? If so, is the following a valid test failure? Confirmed, fails on this branch.
> 
> [mxinden-bot@fb820ca](https://github.com/mxinden-bot/neqo/commit/fb820ca754b7be81af29b8235488b0e885a6e82b)

That's a valid test; I also found that the existing starvation test (round_robin_serves_regular_and_sendordered_in_group) was testing an old spec version that has since changed, so the behavior is slightly different when you mix un-sendordered streams with sendorder.  But the new test checks the other side of the equation from the old test, so it's still useful.

I'm updating the scheduler to match the spec change, which fixes your test.  Thanks!!

jesup:
FYI, clippy found an issue in that test.  For now I've merged that test into my sendgroup PR; if yours lands (or has landed) it will get dropped.

● Update(neqo-transport/src/send_stream.rs)
  ⎿  Added 1 line, removed 6 lines
      2228  
      2229          // Higher-priority sendordered stream with only a few bytes: it easily fits
      2230          // alongside the regular stream if the scheduler gives it a turn.
      2231 -        let mut s = SendStream::new(                                                                                                                 
      2232 -            sendordered,                                                                                                                             
      2233 -            1 << 20,                                                                                                                                 
      2234 -            Rc::clone(&conn_fc),                                                                                                                     
      2235 -            conn_events.clone(),                                                                                                                     
      2236 -        );                                                                                                                                           
      2231 +        let mut s = SendStream::new(sendordered, 1 << 20, Rc::clone(&conn_fc), conn_events);                                                         
      2232          s.send(&[0; 8]).unwrap();
      2233          ss.insert(sendordered, s);
      2234          ss.set_fairness(sendordered, true).unwrap();

neqo-transport/src/streams.rs
```diff
@@ -35,6 +35,29 @@ use crate::{
...
+    #[must_use]
+    pub const fn new(id: u64) -> Self {
+        Self(id)
+    }
+}
+
+impl From<u64> for SendGroupId {
+    fn from(id: u64) -> Self {
+        Self(id)
+    }
```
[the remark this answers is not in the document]
jesup: This frames From<u64> as "the type-conversion boundary," but it's an internal-to-internal seam (http3 Id → transport SendGroupId), not a system boundary, and the invariant is already enforced at the only two places that matter: minting (≥1) and the public set_sendgroup. A debug_assert here guards an input that can't occur — which is exactly the "don't validate scenarios that can't happen, trust internal code" case. The documentation value is also already covered by the doc comments on SendGroupId::new and send_group::Id.

Let's leave the code as-is

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -158,9 +160,25 @@ impl WebTransportSendStream {
             events,
             session_id,
```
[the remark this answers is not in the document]
jesup: I'm leaving it for future use, with expect -- if it gets used, this will trip and force the expect to be removed

jesup:
The two bench regressions - transfer/1-conn/1-100mb-req and streams/walltime/1000-streams/each-1000-bytes - do not reproduce locally.  These are wall-clock times, and are subject to noise on the runners.  Code analysis says my changes affect fair streams; these are not fair streams - if anything, my code should be slightly faster due to an early return - I see each-1000-bytes as 1.8% faster than main, in fact.

These are codspeed.mode = "walltime" benchmarks with codspeed.memory = false  # FIXME: Disabled due to high variance — the repo itself documents these as high-variance. The upload's CI signature (7 high-severe outliers, [+1.6%, +8.9%]) is textbook wall-clock noise. Single non-fair-stream upload cannot be affected by a sendGroup change.

neqo-http3/src/features/extended_connect/send_group.rs
mxinden, line 1: Currently both `neqo-transport` and `neqo-http3` each have a send group ID type, representing the same underlying context. Given that `neqo-http3` depends on `neqo-transport`, why not use `neqo-transport`'s send group ID type in `neqo-http3`?

neqo-transport/src/streams.rs
```diff
@@ -35,6 +35,29 @@ use crate::{
 
 pub type SendOrder = i64;
 
+/// Identifier for a send group, unique within a connection.
+///
+/// A newtype around `u64` rather than a bare alias, so a raw integer (or a `SendOrder`)
+/// cannot be passed where a send-group id is expected.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct SendGroupId(u64);
```
mxinden, line 43: > Currently both `neqo-transport` and `neqo-http3` each have a send group ID type, representing the same underlying context. Given that `neqo-http3` depends on `neqo-transport`, why not use `neqo-transport`'s send group ID type in `neqo-http3`?

mxinden (commented):
Other than the 3 items below, this is ready to merge from my end. Thanks for the follow-up patch on the stream starvation @jesup.
- minor `SendGroupId` type clean-up suggested below
- rebase onto latest `main`
- investigation into the benchmark regressions (I consider everything <3% noise)

neqo-transport/src/streams.rs
jesup: Agreed - there's no good reason for the duplicate. At this point neqo_transport::SendGroupId is already a newtype, so this Id is just a parallel type for the same concept. 
  
neqo-http3 now uses neqo_transport::SendGroupId directly and this Id type is removed -- matching how SendOrder and StreamId are already used from the transport crate. The connection-scoped Generator stays here (per-connection minting is an http3 concern) but now mints neqo_transport::SendGroupId, and the as_u64().into() conversions at the call sites are gone. (Added SendGroupId::as_u64() on the transport side for wire encoding.)

The change is folded into the "Make transport SendGroupId a newtype…" commit.

larseggert:
@jesup this needs a rebase

larseggert:
@jesup anything still needing changes, e.g., based on the latest round of bot comments?

larseggert (approved):

larseggert:
@jesup #3747 is the issue for follow-on fixes based on the bot review above.

mxinden (approved):

PR #3468, opened 2026-03-16
WebTransport stats support

jesup:

PR #3469, opened 2026-03-16
Support handling queuing and highwatermarks/max-age

jesup:
Support handling queuing and highwatermarks/max-age datagrams per the WebTransport spec

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1638,10 +1638,81 @@ impl Http3Connection {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<(
+        bool,
+        Option<extended_connect::datagram_queue::DatagramOutcome>,
+    )> {
         self.validate_extended_connect_session(session_id)?
             .borrow_mut()
             .send_datagram(conn, buf, id, now)
+    }
+
+    pub fn webtransport_set_datagram_high_water_mark(
+        &mut self,
+        session_id: StreamId,
+        mark: f64,
+    ) -> Res<()> {
+        self.recv_streams
+            .get_mut(&session_id)
+            .ok_or(Error::InvalidStreamId)?
+            .extended_connect_session()
+            .ok_or(Error::InvalidStreamId)?
+            .borrow_mut()
+            .set_datagram_high_water_mark(mark);
+        Ok(())
+    }
+
+    pub fn webtransport_set_datagram_max_age(
+        &mut self,
+        session_id: StreamId,
+        age_ms: f64,
+        now: Instant,
+    ) -> Res<Vec<extended_connect::datagram_queue::DatagramOutcome>> {
+        Ok(self
+            .recv_streams
+            .get_mut(&session_id)
+            .ok_or(Error::InvalidStreamId)?
+            .extended_connect_session()
+            .ok_or(Error::InvalidStreamId)?
+            .borrow_mut()
+            .set_datagram_max_age(age_ms, now))
+    }
+
+    pub fn webtransport_process_datagram_queue(
+        &mut self,
+        session_id: StreamId,
+        conn: &mut Connection,
+        now: Instant,
+    ) -> Res<Vec<extended_connect::datagram_queue::DatagramOutcome>> {
+        Ok(self
+            .recv_streams
+            .get_mut(&session_id)
+            .ok_or(Error::InvalidStreamId)?
+            .extended_connect_session()
+            .ok_or(Error::InvalidStreamId)?
+            .borrow_mut()
+            .process_datagram_queue(conn, now))
+    }
+
+    pub(crate) fn process_all_datagram_queues(
+        &mut self,
+        conn: &mut Connection,
+        now: Instant,
+        mut on_outcome: impl FnMut(StreamId, extended_connect::datagram_queue::DatagramOutcome),
+    ) {
+        let session_ids: Vec<StreamId> = self
+            .recv_streams
+            .iter()
+            .filter_map(|(id, stream)| stream.extended_connect_session().is_some().then_some(*id))
+            .collect();
+
+        for session_id in session_ids {
+            if let Ok(outcomes) = self.webtransport_process_datagram_queue(session_id, conn, now) {
+                for outcome in outcomes {
+                    on_outcome(session_id, outcome);
+                }
+            }
+        }
     }
 
     /// If the control stream has received frames `MaxPushId`, `Goaway`, `PriorityUpdateRequest` or
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -802,6 +802,24 @@ impl Http3Client {
                     .maybe_send_max_push_id_frame(&mut self.base_handler);
                 let res = self.base_handler.process_sending(&mut self.conn, now);
                 self.check_result(now, &res);
+
+                // Drain per-session datagram queues into the QUIC layer.
+                // Datagrams are enqueued by webtransport_send_datagram() and only
+                // moved to the QUIC send queue here, during process_http3(). This
+                // is called from process_output()/process(), so datagrams are sent
+                // on the next outgoing packet after the caller triggers processing.
+                self.base_handler.process_all_datagram_queues(
+                    &mut self.conn,
+                    now,
+                    |session_id, outcome| {
+                        self.events.insert(Http3ClientEvent::WebTransport(
+                            WebTransportEvent::DatagramOutcome {
+                                session_id,
+                                outcome,
+                            },
+                        ));
+                    },
+                );
             }
             Http3State::Closed { .. } => {}
             _ => {
--- a/neqo-http3/src/features/extended_connect/connect_udp_session.rs
+++ b/neqo-http3/src/features/extended_connect/connect_udp_session.rs
@@ -16,15 +16,21 @@ use crate::{
     Error, RecvStream, Res, SendStream,
     features::extended_connect::{
         CloseReason, ExtendedConnectEvents, ExtendedConnectType, Protocol,
+        datagram_queue::{DatagramOutcome, WebTransportDatagramQueue},
         session::{DgramContextIdError, State},
     },
     frames::{FrameReader, StreamReaderRecvStreamWrapper, capsule::Capsule},
 };
 
+#[expect(
+    clippy::struct_field_names,
+    reason = "session_id is the most descriptive name here"
+)]
 #[derive(Debug)]
 pub struct Session {
     frame_reader: FrameReader,
     session_id: StreamId,
+    datagram_queue: WebTransportDatagramQueue,
 }
 
 impl Session {
@@ -33,6 +39,7 @@ impl Session {
         Self {
             session_id,
             frame_reader: FrameReader::new(),
+            datagram_queue: WebTransportDatagramQueue::new(),
         }
     }
 }
@@ -99,6 +106,34 @@ impl Protocol for Session {
 
     fn write_datagram_prefix(&self, encoder: &mut Encoder) {
         encoder.encode_varint(0u64);
+    }
+
+    fn set_datagram_high_water_mark(&mut self, mark: f64) {
+        self.datagram_queue.set_high_water_mark(mark);
+    }
+
+    fn set_datagram_max_age(&mut self, age_ms: f64, now: Instant) -> Vec<DatagramOutcome> {
+        self.datagram_queue.set_max_age(age_ms, now)
+    }
+
+    fn enqueue_datagram(
+        &mut self,
+        data: Bytes,
+        id: u64,
+        payload_len: usize,
+        now: Instant,
+    ) -> (bool, Option<DatagramOutcome>) {
+        self.datagram_queue.enqueue(data, id, payload_len, now)
+    }
+
+    fn drain_datagram_queue(
+        &mut self,
+        now: Instant,
+    ) -> (
+        Vec<DatagramOutcome>,
+        Vec<super::datagram_queue::QueuedDatagram>,
+    ) {
+        self.datagram_queue.drain(now)
     }
 
     fn dgram_context_id(&self, datagram: Bytes) -> Result<Bytes, DgramContextIdError> {
--- a/neqo-http3/src/features/extended_connect/datagram_queue.rs
+++ b/neqo-http3/src/features/extended_connect/datagram_queue.rs
@@ -1,0 +1,289 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::{
+    collections::VecDeque,
+    time::{Duration, Instant},
+};
+
+use neqo_common::{Bytes, qdebug, qtrace};
+
+const DEFAULT_HARD_LIMIT: usize = 1000;
+const DEFAULT_MAX_AGE: Duration = Duration::from_secs(u64::MAX);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DatagramOutcome {
+    Sent(u64),
+    Expired(u64),
+    Overflowed(u64),
+}
+
+impl DatagramOutcome {
+    #[must_use]
+    pub const fn id(&self) -> u64 {
+        match *self {
+            Self::Sent(id) | Self::Expired(id) | Self::Overflowed(id) => id,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct QueuedDatagram {
+    pub data: Bytes,
+    pub id: u64,
+    /// Length of the original payload before framing (varint session ID + protocol prefix).
+    pub payload_len: usize,
+    pub timestamp: Instant,
+}
+
+impl QueuedDatagram {
+    pub const fn new(data: Bytes, id: u64, payload_len: usize, now: Instant) -> Self {
+        Self {
+            data,
+            id,
+            payload_len,
+            timestamp: now,
+        }
+    }
+
+    pub fn age(&self, now: Instant) -> Duration {
+        now.duration_since(self.timestamp)
+    }
+}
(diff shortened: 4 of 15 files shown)


neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    pub fn age(&self) -> Duration {
+        self.timestamp.elapsed()
+    }
+}
+
+/// Per-session outgoing datagram queue with high water mark and max-age support.
+///
+/// Datagrams are enqueued here by `send_datagram()` and drained into the QUIC
```
mxinden:
```suggestion
/// Datagrams are enqueued here by [`WebTransportDatagramQueue::send_datagram()`] and drained into the QUIC
```

Please use Rust doc links  for each of these wherever possible. In case we ever rename a method, CI will fail, reminding us to update the documentation as well.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+use std::{
+    collections::VecDeque,
+    time::{Duration, Instant},
+};
+
+use neqo_common::{qdebug, qtrace, Bytes};
+
+const DEFAULT_HARD_LIMIT: usize = 1000;
+const DEFAULT_HIGH_WATER_MARK: f64 = f64::INFINITY;
+const DEFAULT_MAX_AGE: Duration = Duration::from_secs(u64::MAX / 1000);
```
mxinden: Why `/ 1000`? Asked differently, what is the constant `1000` based on?

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+/// layer by `process_queue()`, which is called during `process_http3()` as part
+/// of `process_output()`. The caller must call `process_output()` after enqueuing
+/// to ensure datagrams are actually transmitted. In Gecko, this happens via
+/// `StreamHasDataToWrite()` -> `ForceSend()`, which asynchronously dispatches
+/// `SendData()` -> `ProcessOutput()` on the next event loop cycle.
+#[derive(Debug)]
+pub struct WebTransportDatagramQueue {
+    queue: VecDeque<QueuedDatagram>,
+    hard_limit: usize,
+    high_water_mark: f64,
```
mxinden: Instead of overloading `f64::INIFINITY` with the semantic meaning of not-specified, I would prefer using an `Option` here.


```suggestion
    high_water_mark: Option<f64>,
```

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+        Self {
+            data,
+            id,
+            payload_len,
+            timestamp: Instant::now(),
+        }
+    }
+
+    pub fn age(&self) -> Duration {
+        self.timestamp.elapsed()
```
mxinden: `elapsed` calls `Instant::now`. Throughout the Neqo codebase we consider time a side effect, passed in explicitly. See the many `now: Instant` function parameters. That allows us to easily e.g. move in time in tests, or depend on deterministic execution.

Please do the same here.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+        (below_watermark, dropped)
+    }
+
+    /// Returns `(outcomes, payload_bytes_sent, overhead_bytes_sent)`.
+    ///
+    /// `payload_bytes_sent` counts only the application payload bytes for datagrams
+    /// that were successfully handed to the QUIC layer (excludes framing overhead).
+    /// `overhead_bytes_sent` is the framing overhead (varint session ID + protocol
+    /// prefix) for those same datagrams.
+    pub fn process_queue(&mut self, send_fn: &mut dyn FnMut(&[u8], u64) -> Result<(), ()>) -> (Vec<(u64, DatagramOutcome)>, u64, u64) {
```
mxinden: I wonder whether the indirection through the `send_fn` is necessary, or said differently, worth the complexity. I will think some more. Maybe you have alternative design ideas.

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -376,4 +394,20 @@ impl Protocol for Session {
...
+    fn set_datagram_max_age(&mut self, age_ms: f64) -> Vec<(u64, super::datagram_queue::DatagramOutcome)> {
+        self.set_datagram_max_age(age_ms)
+    }
+
+    fn enqueue_datagram(&mut self, data: Bytes, id: u64, payload_len: usize) -> (bool, Option<(u64, DatagramOutcome)>) {
+        self.enqueue_datagram(data, id, payload_len)
+    }
+
+    fn process_datagram_queue(&mut self, send_fn: &mut dyn FnMut(&[u8], u64) -> Result<(), ()>) -> (Vec<(u64, DatagramOutcome)>, u64, u64) {
+        self.process_datagram_queue(send_fn)
```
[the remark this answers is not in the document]
mxinden: Agreed

neqo-http3/src/connection.rs
```diff
@@ -1737,6 +1737,111 @@ impl Http3Connection {
             .send_datagram(conn, buf, id, now)
     }
 
+    fn extended_connect_set_datagram_high_water_mark(
```
mxinden: Do you foresee any other EXTENDED CONNECT protocols to have the concept of a high water mark? If not, I suggest inlining this function into `webtransport_set_datagram_high_water_mark`. Same with the functions below.

neqo-http3/src/connection.rs
```diff
@@ -1737,6 +1737,111 @@ impl Http3Connection {
...
+            .ok_or(Error::InvalidStreamId)?
+            .borrow_mut()
+            .process_datagram_queue(conn))
+    }
+
+    pub(crate) fn process_all_datagram_queues(
+        &mut self,
+        conn: &mut Connection,
+    ) -> Vec<(StreamId, u64, crate::features::extended_connect::datagram_queue::DatagramOutcome)> {
+        let session_ids: Vec<StreamId> = self
```
mxinden, line 1703: Allocating a `Vec` on each `process` call seems wasteful. Can we do better?

mxinden (commented):
Early review. I still need to acquire a better understanding of the queuing mechanism to give an in-depth review.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -25,6 +26,7 @@ use crate::{
 pub struct Session {
     frame_reader: FrameReader,
     session_id: StreamId,
+    datagram_queue: WebTransportDatagramQueue,
```
mxinden, line 33: I am a bit confused. Why does a ConnectUDP session need a WebTransport datagram queue?

neqo-http3/src/connection.rs
```diff
@@ -1737,6 +1737,111 @@ impl Http3Connection {
...
+            .extended_connect_session()
+            .ok_or(Error::InvalidStreamId)?
+            .borrow_mut()
+            .set_datagram_max_age(age_ms))
+    }
+
+    pub fn webtransport_set_datagram_high_water_mark(
+        &mut self,
+        session_id: StreamId,
+        mark: f64,
```
mxinden: Why a `f64`?

What does the W3C side use? A double?

Even if the W3C side use floating points, can we use a u64 here instead? Why infect more of our stack with this inprecission?

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: A value that's effectively infinite but was below the max_u64 value for safety out of habit.  Of course, with checked arithmetic or proper design there's no overflow worry, so we could just use max_u64

neqo-http3/src/connection_client.rs
```diff
@@ -831,10 +840,21 @@ impl Http3Client {
...
+        let (below_watermark, dropped) = self
+            .base_handler
+            .webtransport_send_datagram(session_id, &mut self.conn, buf, id, now)?;
+        if let Some((tracking_id, outcome)) = dropped {
+            self.events.insert(Http3ClientEvent::WebTransport(
+                WebTransportEvent::DatagramOutcome {
+                    session_id,
+                    tracking_id,
+                    outcome,
+                },
```
martinthomson: We should not be generating events when the event could be returned with an error code.

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: The alternatives would be to move the process_queue implementation out in to the callers (there are 2 plus tests) and have it call into a "get next ready datagram" (and expire old ones).  Stats keeping would have to be done in multiple places, as would the loop.  
Net: simpler interface and impl in datagram_queue, but more complex for the callers and more duplicate code to keep in sync.  At worst I see the current design as a wash, and maybe a little better.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -416,34 +430,79 @@ impl Session {
...
         dgram_data.encode_varint(self.id.as_u64() / 4);
         self.protocol.write_datagram_prefix(&mut dgram_data);
         dgram_data.encode(buf);
 
-        conn.send_datagram(dgram_data.into(), id)?;
-        self.protocol.record_datagram_sent();
-        self.protocol.record_bytes_sent(buf.len() as u64);
-        qtrace!("[{self}] sent datagram via QUIC datagram");
-        Ok(())
+        if dgram_data.len() as u64 > max_datagram_size {
```
martinthomson: I'd prefer something without the cast, like:

```rust
if u64::try_from(dgram_data.len()).map_or(true, |l| l > conn.remote_datagram_size()) {
    return Err(...);
}
```

neqo-http3/src/features/extended_connect/datagram_queue.rs
martinthomson: That and... why does this need to be f64, as opposed to a usize?

neqo-http3/src/connection.rs
jesup: done

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        while let Some(dgram) = self.queue.front() {
+            if dgram.age() > self.max_age {
+                let dgram = self.queue.pop_front().unwrap();
+                qdebug!("Datagram {} expired (age: {:?})", dgram.id, dgram.age());
+                expired.push(dgram.id);
+            } else {
+                break;
+            }
+        }
```
martinthomson: This is going to be inefficient.  You can use partition_point and split_off and then return a mapped, owning iterator over the part that you don't want any more.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+            .collect()
+    }
+
+    fn expire_old_datagrams(&mut self) -> Vec<u64> {
+        let mut expired = Vec::new();
+
+        while let Some(dgram) = self.queue.front() {
+            if dgram.age() > self.max_age {
+                let dgram = self.queue.pop_front().unwrap();
+                qdebug!("Datagram {} expired (age: {:?})", dgram.id, dgram.age());
```
martinthomson: This is solidly qtrace!() at best.

neqo-http3/src/connection.rs
jesup: By using a callback for results, yes

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    pub fn enqueue(&mut self, data: Bytes, id: u64, payload_len: usize) -> (bool, Option<(u64, DatagramOutcome)>) {
+        let dropped = if self.queue.len() >= self.hard_limit {
+            if let Some(oldest) = self.queue.pop_front() {
+                qdebug!(
+                    "Queue at hard limit ({}), dropping oldest datagram {}",
+                    self.hard_limit,
+                    oldest.id
+                );
+                Some((oldest.id, DatagramOutcome::DroppedQueueFull))
```
martinthomson: Rather than `Option<(u64, DatagramOutcome)>` could you change the enum to encompass the entirely of the returned result?

I realize that this is a case where you have just one thing, so Option is just a collection, but you could define the outcome to be:

```rust
enum DatagramOutcome {
  Ok,
  Expired(u64),
  Overflowed(u64)
}
```

Or even:

```rust
enum DatagramOutcome<C: IntoIterator<Item = u64>> {
  Ok,
  Expired(C),
  Overflowed(C)
}
```

... so that you could return a collection in the enum.  It's not like there are any cases where you overflow AND expire.

neqo-http3/src/connection.rs
jesup: Spec requirement: sendDatagramsHighWaterMark is an unrestricted double

neqo-http3/src/features/extended_connect/datagram_queue.rs
martinthomson: I see that this doesn't hold for the function below, where you want to record which datagrams were expired and which were sent.  But you can do that by returning a tuple of collections rather than use this enum.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// Returns `(outcomes, payload_bytes_sent, overhead_bytes_sent)`.
+    ///
+    /// `payload_bytes_sent` counts only the application payload bytes for datagrams
+    /// that were successfully handed to the QUIC layer (excludes framing overhead).
+    /// `overhead_bytes_sent` is the framing overhead (varint session ID + protocol
+    /// prefix) for those same datagrams.
+    pub fn process_queue(&mut self, send_fn: &mut dyn FnMut(&[u8], u64) -> Result<(), ()>) -> (Vec<(u64, DatagramOutcome)>, u64, u64) {
+        let mut outcomes = Vec::new();
+        let mut payload_bytes_sent: u64 = 0;
+        let mut overhead_bytes_sent: u64 = 0;
```
martinthomson: Does type inference not work?  0_u64 if absolutely necessary.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
jesup: It needs a datagram queue — CONNECT-UDP sends datagrams via QUIC DATAGRAM or HTTP DATAGRAM Capsules, and the send path enqueues in send_datagram then drains during process_output. That requires a queue. 

But it doesn't need the WebTransport-specific features on the queue — high water mark and max-age. I just removed the public APIs to set those on CONNECT-UDP sessions. The queue is really just being used as a plain FIFO for CONNECT-UDP. 

The naming is misleading -- WebTransportDatagramQueue suggests it's WebTransport-specific. It could be renamed to just DatagramQueue since it's a general-purpose outgoing datagram queue that optionally has high water mark and max-age support. Or the CONNECT-UDP session could use a simpler queue type without those features, but that's more code for no practical benefit since the defaults (None / u64::MAX) effectively disable them.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -416,34 +430,79 @@ impl Session {
...
+        qtrace!("[{self}] enqueued datagram for sending via QUIC datagram");
+        Ok((below_watermark, dropped))
+    }
+
+    pub(crate) fn process_datagram_queue(
+        &mut self,
+        conn: &mut Connection,
+    ) -> Vec<(u64, super::datagram_queue::DatagramOutcome)> {
+        let (outcomes, payload_bytes, _overhead_bytes) = self.protocol.process_datagram_queue(&mut |data, id| {
+            match conn.send_datagram(data.to_vec(), DatagramTracking::Id(id)) {
```
martinthomson: Like Max, I find this arrangement with send_fn undesirable.  Is this the only send_fn you have?  Because I think that you would be better off setting a cap on the bytes that can be sent down to the transport and have a function that returns those datagrams.

The current structure requires that an error in writing to the underlying transport results in the datagram being kept and retried.  Is that really what you want for this?  What errors are you seeking to avoid?  (Documentation in the form of comments seems really important for this sort of thing.)

martinthomson (commented):
In my mind, any queues for sending stuff in WebTransport would be outside of the stack.  So I found this a little surprising to see.

neqo-http3/src/connection.rs
martinthomson: We don't need to let the spec requirement pollute this code.  We want a positive integer value, plus maybe Infinity, but a called to this code can translate the (WebIDL-sanitized) f64 into a u64 with `u64::try_from(mark).unwrap_or(u64::MAX)` very easily.

jesup:

1. `trunk()`
1. **users/jesup/highwatermarks ← this PR**

---

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: The structure had changed, but I the same issue applied.  Also applied it to highest_valid_order()

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: If it's not a problem I'd like to leave this for now; perhaps address later - I have a lot of following patches this would impact.

neqo-http3/src/features/extended_connect/session.rs
jesup: Done

neqo-http3/src/features/extended_connect/datagram_queue.rs
larseggert: I'd have a mild preference for getting fundamentals into their idiomatic shape sooner rather than later, but will defer to @martinthomson. Claude will be able to fix these up during rebasing.

jesup: Ok, I made the change.

PR #3472, opened 2026-03-16
Don't send datagrams combined with PMTUD packets since lost datagrams…

jesup:
… can't be retried

PR #3471, opened 2026-03-16
Track bytesSentOverhead separately per WebTransport spec

jesup:

PR #3470, opened 2026-03-16
sendgroup and sendorder improved implementation & tracking

jesup:

PR #3473, opened 2026-03-16
Support WebTransport anticipatedStreams

jesup:


neqo-http3/src/features/extended_connect/tests/webtransport/anticipated_streams.rs
```diff
@@ -0,0 +1,266 @@
...
+use crate::{
+    Error,
+    features::extended_connect::tests::webtransport::{WtTest, wt_default_parameters},
+};
+
+/// The connection-level baseline for stream limits (from LOCAL_STREAM_LIMIT_*
+/// in neqo-transport).  Values at or below this require no MAX_STREAMS frame.
+const STREAM_LIMIT_DEFAULT: u64 = 100;
+
+// ── helpers ──────────────────────────────────────────────────────────────────
```
mxinden: I don't think we use these heading styles anywhere else in the codebase. Not opposed, but would like to be consistent.

mxinden (commented):
I am missing context. What is the goal of this feature? What does it enable applications to do?

Note that, in case we ever run multiple sessions and HTTP/3 requests on the same connection, a single session could choose the global limit, by setting its session limit, which propagates to a global limit.

martinthomson (commented):
This change is not what I would have expected from this one.  The goal of this is to ensure that the flow control allocation for streams is generous enough that the server can make streams without getting blocked.  To that end, this should route the message through to the `max_active` value on the relevant `ReceiverFlowControl` instance.

There is also the potential for this input to be fed to `ConnectionParameters::max_streams()` if the connection has not been created yet.  That is a matter for the neqo glue code in Firefox.

In terms of spelling, this can use a better name than the one in the WT spec.  `set_active_streams` or `increase_max_active_streams`...

jesup:

1. `trunk()`
1. **users/jesup/anticipated_streams ← this PR**

---

jesup:
Perhaps this is closer to what you expected

neqo-http3/src/features/extended_connect/tests/webtransport/anticipated_streams.rs
jesup: The point of Anticipated Streams is to reduce the chance we have to negotiate for more streams, especially with applications doing video using stream-per-frame.  When we add pooled connections we should take into account the anticipated streams of all pooled connections.

neqo-http3/src/connection.rs
```diff
@@ -1500,10 +1500,79 @@ impl Http3Connection {
...
+        // LOCAL_STREAM_LIMIT_UNI (100); only a total exceeding that will cause a
+        // MAX_STREAMS frame to be sent.
+        let total: u64 = self
+            .recv_streams
+            .values()
+            .filter_map(|s| s.extended_connect_session())
+            .filter(|s| s.borrow().connect_type() == ExtendedConnectType::WebTransport)
+            .map(|s| u64::from(s.borrow().anticipated_incoming_uni()))
+            .sum();
+        conn.set_remote_max_streams_uni(total);
```
martinthomson: This function should account for control streams.  There are three of those, but you probably want to make a constant that we can increase if we add more at some point.

neqo-transport/src/connection/params.rs
```diff
@@ -28,13 +28,23 @@ use crate::{
...
 const LOCAL_STREAM_LIMIT_BIDI: u64 = 100;
 /// Maximum number of unidirectional streams that the remote can open.
 ///
-/// Constant throughout the lifetime of the connection.
+/// This is the baseline sent in the `initial_max_streams_uni` transport
+/// parameter during the QUIC handshake. It can be increased later via
+/// `MAX_STREAMS` frames (see [`crate::streams::Streams::set_remote_max_streams_uni`]),
+/// for example when WebTransport sessions set anticipated concurrent incoming
+/// stream counts that exceed this baseline.  Per the QUIC spec, stream limits
+/// can only increase; values at or below the current limit have no effect.
```
martinthomson: These comment changes aren't really necessary.  (If anything, the original are a little out of date: the values are defaults, not baselines.)

neqo-transport/src/fc.rs
```diff
@@ -330,7 +330,6 @@ where
     }
 
     pub const fn set_max_active(&mut self, max: u64) {
-        // If max_active has been increased, send an update immediately.
```
martinthomson, line 349: Why delete this?  It's a useful note.

neqo-http3/src/connection.rs
```diff
@@ -1500,10 +1500,79 @@ impl Http3Connection {
...
+            .ok_or(Error::InvalidStreamId)?;
+        {
+            let session_ref = session.borrow();
+            if session_ref.connect_type() != ExtendedConnectType::WebTransport {
+                return Err(Error::InvalidStreamId);
+            }
+        }
+        session.borrow_mut().set_anticipated_incoming_bidi(value);
+        // Sum across all sessions: see comment in webtransport_increase_max_uni_streams.
+        // The connection's baseline is LOCAL_STREAM_LIMIT_BIDI (100).
```
martinthomson: This comment is not right.  The baseline is determined by `ConnectionParameters`.

neqo-http3/src/connection.rs
```diff
@@ -1500,10 +1500,79 @@ impl Http3Connection {
...
+            .ok_or(Error::InvalidStreamId)?;
+        if session.borrow().connect_type() != ExtendedConnectType::WebTransport {
+            return Err(Error::InvalidStreamId);
+        }
+        session.borrow_mut().set_anticipated_incoming_uni(value);
+        // Sum across all sessions: multiple WebTransport sessions may share a
+        // connection in the future, and each session's streams count against the
+        // same QUIC connection-level limit.  The connection's baseline is
+        // LOCAL_STREAM_LIMIT_UNI (100); only a total exceeding that will cause a
+        // MAX_STREAMS frame to be sent.
```
martinthomson: See below about `LOCAL_STREAM_LIMIT_UNI` and this being wrong.

martinthomson (dismissed):
I'm not sure about the logic here.  If we are giving a session streams up to the limit they request, shouldn't we be policing their use of that stream limit?

And what about defaults?  You are setting the default value to 0, which

PR #3474, opened 2026-03-16
Add support for waitUntilAvailable

jesup:

PR #3475, opened 2026-03-16
atomicWrite support

jesup:


neqo-http3/src/connection_client.rs
```diff
@@ -1035,6 +1035,36 @@ impl Http3Client {
...
+    ///
+    /// # Errors
+    ///
+    /// `InvalidStreamId` if the stream does not exist or other stream errors.
+    pub fn webtransport_send_stream_atomic(
+        &mut self,
+        stream_id: StreamId,
+        data: &[u8],
+    ) -> Res<bool> {
+        let now = std::time::Instant::now();
```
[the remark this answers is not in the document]
mxinden: Copilot is right. Please pass time as an input argument.

neqo-transport/src/send_stream.rs
```diff
@@ -1165,6 +1165,17 @@ impl SendStream {
         }
     }
 
+    /// Bytes currently buffered in the tx buffer awaiting transmission.
+    #[must_use]
+    pub fn buffered(&self) -> usize {
+        match &self.state {
+            State::Send { send_buf, .. } | State::DataSent { send_buf, .. } => {
+                send_buf.buffered()
```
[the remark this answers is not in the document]
mxinden: Agreed with Copilot.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -688,6 +688,10 @@ impl SendStream for Rc<RefCell<Session>> {
     fn handle_stop_sending(&mut self, close_type: CloseType) {
         self.borrow_mut().close(close_type);
     }
+
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+        self
+    }
```
mxinden: I don't understand the need for the downcast pattern through `std::any::Any`. I am hesitant to use it in this code base. As far as I am aware, this is the first time. Why not override `SendStream::send_data_atomic`?

https://github.com/larseggert/neqo/blob/ea6d40ded1706141ea0695bac97883ceed4964f4/neqo-http3/src/lib.rs#L597-L602

neqo-http3/src/connection_client.rs
```diff
@@ -1035,6 +1035,36 @@ impl Http3Client {
...
 
+    /// Returns available send window and buffered bytes for a WebTransport send stream.
+    ///
+    /// # Errors
+    ///
+    /// `InvalidStreamId` if the stream does not exist.
+    pub fn webtransport_send_stream_flow_control_info(
+        &self,
+        stream_id: StreamId,
+    ) -> Res<(usize, usize)> {
```
mxinden: Returning `(usize, usize)` is not descriptive. How about returning a struct?

``` rust
struct FlowControl {
  available: usize,
  buffered: usize,
}
```

jesup:

1. `trunk()`
1. **users/jesup/atomicWrite ← this PR**

---

neqo-http3/src/send_message.rs
```diff
@@ -293,14 +293,14 @@ impl SendStream for SendMessage {
...
-    fn send_data_atomic(&mut self, conn: &mut Connection, buf: &[u8], now: Instant) -> Res<()> {
+    fn send_data_atomic(&mut self, conn: &mut Connection, buf: &[u8], now: Instant) -> Res<bool> {
         let data_frame = HFrame::Data {
             len: buf.len() as u64,
         };
         self.stream.encode_with(|e| data_frame.encode(e));
         self.stream.buffer(buf);
         _ = self.stream.send_buffer(conn, now)?;
-        Ok(())
+        Ok(true)
```
martinthomson: Where is the false?

neqo-http3/src/features/extended_connect/webtransport_streams.rs
```diff
@@ -180,6 +180,24 @@ impl WebTransportSendStream {
...
+    ) -> Res<bool> {
+        // First ensure init buffer is sent
+        self.send(conn, now)?;
+
+        // Only send atomically if we're in SendingData state (init buffer sent)
+        if self.state == WebTransportSenderStreamState::SendingData {
+            Ok(conn.stream_send_atomic(self.stream_id, data)?)
+        } else {
+            // Can't send atomically yet, init buffer not fully sent
+            Ok(false)
```
martinthomson, line 198: So this is the only false return case?  What about when the stream hasn't enough space to send?

neqo-http3/src/send_message.rs
```diff
@@ -293,14 +293,14 @@ impl SendStream for SendMessage {
...
     }
 
-    fn send_data_atomic(&mut self, conn: &mut Connection, buf: &[u8], now: Instant) -> Res<()> {
+    fn send_data_atomic(&mut self, conn: &mut Connection, buf: &[u8], now: Instant) -> Res<bool> {
         let data_frame = HFrame::Data {
             len: buf.len() as u64,
         };
         self.stream.encode_with(|e| data_frame.encode(e));
         self.stream.buffer(buf);
         _ = self.stream.send_buffer(conn, now)?;
```
martinthomson: IMPORTANT: I don't think that this function is truly atomic.  The stream has a method called send_atomic, but this goes nowhere near that.  So I think we're into false promises here.  I DON'T think that it's enough to replace this line and the next with a simple call.  What you will need to do is ensure that the HTTP/3-layer buffer is passed down AND that a new buffer (including the length field) can be passed down atomically.

martinthomson (changes requested):
I believe that you have uncovered a bug to fix that is a bit deeper than this superficial fix.

neqo-http3/src/send_message.rs
martinthomson: I'm pretty sure that this comment wasn't addressed.

martinthomson (dismissed):
As far as I can tell, the bug I identified in the last review was dismissed, but not fixed.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    assert!(
+        wt.client
+            .webtransport_send_stream_atomic(stream_id, CHUNK_B, now())
+            .unwrap()
+    );
+
+    // The two atomic writes are batched in the tx buffer; server sees them concatenated.
+    let mut combined = Vec::with_capacity(CHUNK_A.len() + CHUNK_B.len());
+    combined.extend_from_slice(CHUNK_A);
+    combined.extend_from_slice(CHUNK_B);
```
martinthomson: This is overwrought.

```suggestion
    let combined = CHUNK_A.iter().chain(CHUNK_B).collect:<Vec<_>>();
```

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+        .unwrap();
+    assert!(sent);
+
+    // Before exchanging packets everything is in the tx buffer but not acked.
+    let fc = wt
+        .client
+        .webtransport_send_stream_flow_control_info(stream_id)
+        .unwrap();
+    assert!(
+        fc.buffered >= SMALL_DATA.len(),
```
martinthomson: Can you compute the size of the overhead?  It seems like the cost is fixed for a test this controlled.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    );
+
+    // After the full exchange the peer has acked all data.
+    wt.exchange_packets();
+    let fc = wt
+        .client
+        .webtransport_send_stream_flow_control_info(stream_id)
+        .unwrap();
+    assert_eq!(fc.buffered, 0, "all data acked, nothing buffered");
+    assert!(fc.available > 0);
```
martinthomson: Maybe you can check that `fc_after.available = fc_before.available - fc_before.buffered`.  This isn't perfectly robust against changes to our FC code, but it's OK to bet on a small message not triggering a flow control update.  (A bet that your next test already makes, knowingly.)

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    let after = wt
+        .client
+        .webtransport_send_stream_flow_control_info(stream_id)
+        .unwrap();
+
+    // The flow-control window shrinks by (at least) the data we sent.
+    // (The peer may or may not send a MAX_STREAM_DATA update, so we only
+    // assert the window did not grow beyond what it was before.)
+    assert!(
+        after.available <= before.available,
```
martinthomson: See above; I think that you can integrate this check into the above test.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    // Configure a tiny stream receive window so the client quickly runs out
+    // of send credits.
+    const TINY_STREAM_WINDOW: u64 = 64;
+    // The server advertises its receive-window for client-initiated (remote)
+    // UniDi streams.  Setting this small limits how much the client can send.
+    let server_params = wt_default_parameters().connection_parameters(
+        ConnectionParameters::default().max_stream_data(
+            StreamType::UniDi,
+            true,
+            TINY_STREAM_WINDOW,
```
martinthomson: If the constant were a `u8`, you could use `u64::from()` here and `usize::from()` below.  Haxx, I know, but it's less text.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+        .webtransport_send_stream_atomic(stream_id, &large, now())
+        .unwrap();
+    assert!(!sent, "atomic send larger than window should return false");
+
+    // The stream is still usable — a small write succeeds.
+    let fc = wt
+        .client
+        .webtransport_send_stream_flow_control_info(stream_id)
+        .unwrap();
+    if fc.available >= SMALL_DATA.len() {
```
martinthomson: I think that you should *assert* that `fc.available >= SMALL_DATA.len()` instead.  Otherwise, the test that this check guards won't run and that would not be good.

I know that you can't guarantee that 64 is enough larger than the size of the small write if you are pretending that you don't know what the overheads are.  But we know what the overheads really are and the absolute worst case as well.  And the worst case (25 bytes, I believe) is knowable.  Make sure that you have that much space spare, that's all.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    let session = wt.create_wt_session();
+    let stream_id = wt.create_wt_stream_client(session.stream_id(), StreamType::UniDi);
+    wt.exchange_packets();
+
+    // A write that exceeds the tiny window must fail atomically (return false).
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "TINY_STREAM_WINDOW=64 fits in usize"
+    )]
+    let large: Vec<u8> = vec![0xAB; TINY_STREAM_WINDOW as usize + 1];
```
martinthomson: Is this big enough?  We might hedge the computation of available space, so just one extra byte might end up working.  I'd choose a larger increment (the 25 from below might be wise; you can define a MAX_WT_STREAM_OVERHEAD constant, and then add one more to that).

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+}
+
+/// Calling atomic send with an unknown stream ID returns `InvalidStreamId`.
+#[test]
+fn wt_atomic_send_invalid_stream_id() {
+    let mut wt = WtTest::new();
+    drop(wt.create_wt_session());
+
+    let bogus = StreamId::from(9999);
```
martinthomson:
```suggestion
    let bogus = StreamId::from(u64::MAX);
```

This is always bogus, because it's not even a valid stream ID.  What you have is fine, of course.

martinthomson (approved):
Plenty of nits on the tests, but this is in good shape.

neqo-http3/src/features/extended_connect/tests/webtransport/atomic_write.rs
jesup: FC credit is consumed at write time, so `available` is unchanged after ack (a small message doesn't trigger MAX_STREAM_DATA), so I'm testing available against available_before

PR #3476, opened 2026-03-16
Support for the WT_DRAIN_SESSION frame to initiate draining

jesup:
Support for the WT_DRAIN_SESSION frame to initiate draining for a specific session

PR #3477, opened 2026-03-16
subtract out some overheads from the max datagram size

jesup:


neqo-http3/src/connection_client.rs
```diff
@@ -894,9 +894,17 @@ impl Http3Client {
     ///
     /// This cannot panic. The max varint length is 8.
     pub fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64> {
+        let session_id_len = u64::try_from(Encoder::varint_len(session_id.as_u64()))
+            .map_err(|_| Error::Internal)?;
+        // Reserve space for ACK frames and packet number growth beyond what is
+        // seen at session start (when max_datagram_size is sampled). A worst-case
+        // ACK frame is ~50 bytes (RecvdPackets::USEFUL_ACK_LEN) and packet number
+        // varints can grow by up to 3 bytes over the life of a connection, giving
+        // ~53 bytes of realistic overhead; 64 provides a small margin.
```
mxinden: I don't understand why ACK size is relevant on the HTTP3 level. Can you expand on that? If a pending ACK plus the current WebTransport datagram don't fit into a single UDP datagram, why don't we send two UDP datagrams?

> what is seen at session start (when max_datagram_size is sampled)

In what sense is the calculation only happening at session start? Below we are calling `self.conn.max_datagram_size()`, potentially during the session.

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
 };
 
 const DGRAM: &[u8] = &[0, 100];
+// Overhead subtracted by webtransport_max_datagram_size on the client side.
+const CLIENT_DATAGRAM_OVERHEAD: u64 = 64;
```
mxinden: Why is the constant re-defined here? Why not import the constant from `neqo-http3/src/connection_client.rs`?

neqo-http3/src/connection_client.rs
jesup: We sample max_datagram_size() in the DOM WebTransport code at session start and return that to the application when it asks.

Ack size is relevant in that if a datagram doesn't fit because there's an ACK in the buffer, the datagram can end up dropped (though not usually): if it's deferred due to an ACK it will be pushed back.  On retry, if the number of bytes for pn encoding has changed (say from <128 to >128 unacked packets in flight), the packet will be dropped.

More importantly, if there's significant incoming traffic, there may always be an ACK in the buffer, leading to repeated delays in sending datagrams, perhaps to the point where they end up dropped. Generally we want datagrams to go on the wire as soon as possible if other traffic doesn't have precedence.  (People will be using them for audio and video).

We could avoid the first problem with a 3 or 4 byte reduction, but not the second.  To reduce the impact of the second, we need to reduce the max size by ~53 bytes or more.

mxinden: I understand the concerns above, though if I understand correctly, they are based on intuition and not real-world experience. Thus I suggest only introducing these optimizations once we have proof that they actually occur.

While this might improve performance by allowing ACKs and datagrams to more likely share the same UDP datagram, it might also reduce performance as it reduces the size of WebTransport datagrams. In addition, it is a layer violation, having `neqo-http3` be aware of `neqo-transport`, i.e. WebTransport aware of QUIC ACKs.

jesup: Reasonable points.  If we have incoming HD video+audio (and maybe multiple incoming streams in a conference), we will have a lot of ACKs to send; this could delay sending datagrams.  OTOH, maybe what happens is it doesn't fit, we send the ACK, then we have an empty buffer and send the datagram.  We waste some bits by having a higher packet rate (close to 2x at the limit, but probably much less than that), but that may be ok.
IIRC I added this because some tests failed without it; I'll see if I can find them

martinthomson: When we invest in ack frequency, the odds that one of those large datagrams will have to share a UDP datagram with an ACK is still significant, but it will be low enough that we don't need to concern ourselves.  As far as I'm aware, we'll still get ample chance to squish smaller-than-MTU datagrams in with ACKs, but even if we have no packet sharing, that will still be less of an overhead than a fixed tax on every datagram frame.

jesup:

1. `trunk()`
1. **users/jesup/datagram_max_size ← this PR**

---

neqo-http3/src/connection_client.rs
```diff
@@ -939,9 +944,13 @@ impl Http3Client {
     ///
     /// This cannot panic. The max varint length is 8.
     pub fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64> {
-        Ok(self.conn.max_datagram_size()?
-            - u64::try_from(Encoder::varint_len(session_id.as_u64()))
-                .map_err(|_| Error::Internal)?)
+        let session_id_len =
+            u64::try_from(Encoder::varint_len(session_id.as_u64())).map_err(|_| Error::Internal)?;
```
martinthomson: This is wrong.  The length of the encoding is the encoded length of the session ID *divided by four*.

neqo-http3/src/connection_client.rs
```diff
@@ -54,6 +54,11 @@ where
     move |(id, _)| (*id >= base && !(id.is_bidi() ^ base.is_bidi())).then_some(*id)
 }
 
+/// Reserve space packet number growth beyond what is seen at session start
+/// (when max_datagram_size is sampled). Packet number varints can grow by
+/// up to 3 bytes over the life of a connection.
+pub(crate) const DATAGRAM_OVERHEAD: u64 = 3;
```
martinthomson: The number is maybe correct, but the explanation is completely wrong.  This is also implemented at the wrong layer.

The challenge here is that the `neqo_transport::Connection::max_datagram_size()` implementation doesn't account for the potential for the packet number encoding to end up large.  It computes what the datagram size of the very next packet is.  By its nature, that can change.  Send more packets and the header gets bigger.  Depending on how many packets are in flight, the packet number might be any value between 1 and 4.  Most likely, when you have tested this, the number is 1, because most tests don't end up with many outstanding packets.

If there is a cache for the value (as there probably should be), then the packet number encoding is going to be a problem if the situation changes.  At higher throughput, the number you got when throughput was lower will be wrong.

If you are planning to make an adjustment for this, which does make sense, then the fix involves tweaks to `max_datagram_size` at the connection layer.

Personally, I find the implementation of `max_datagram_size` at that layer to be a little overdone.  It should be possible to compute the value from knowledge of the short header, the current connection ID on the primary path, the datagram packet size, and the chosen AEAD expansion.  That computation can assume that the packet number is going to be 4 bytes long.  Anyone who is looking to push a ton of datagrams, such that the size matters, will hit that limit.

martinthomson (dismissed):
Two major bugs here.  Which is doing great given that there's really only two lines of code involved :)  One is pre-existing though.

mxinden (commented):
@jesup most of the changes in this pull request are unrelated formatting changes. Is this by mistake?

larseggert:
This might have been me running `cargo fmt` to fix the failing tests. I'm really starting to dislike stacked PRs.

larseggert:
I think we need to land this stack from the bottom up, and @jesup needs to rebase the rest of the stack every time.

jesup:
@codspeedbot explain why this is faster

PR #3482, opened 2026-03-17
fix: Handle `ENOBUFS` in UDP send

larseggert:
Based on a discussion with @ianswett yesterday, don't treat `ENOBUFS` as a fatal error when sending UDP packets.

### DIFF
--- a/neqo-udp/src/lib.rs
+++ b/neqo-udp/src/lib.rs
@@ -391,22 +391,21 @@ mod tests {
     #[test]
     #[cfg(windows)]
     fn is_emsgsize_true_for_wsaemsgsize() {
-        let err = io::Error::from_raw_os_error(windows::Win32::Networking::WinSock::WSAEMSGSIZE.0);
+        let err = io::Error::from_raw_os_error(WinSock::WSAEMSGSIZE.0);
         assert!(is_emsgsize(&err));
     }
 
     #[test]
     #[cfg(windows)]
     fn is_emsgsize_true_for_wsaeinval() {
-        let err = io::Error::from_raw_os_error(windows::Win32::Networking::WinSock::WSAEINVAL.0);
+        let err = io::Error::from_raw_os_error(WinSock::WSAEINVAL.0);
         assert!(is_emsgsize(&err));
     }
 
     #[test]
     #[cfg(windows)]
     fn is_emsgsize_false_for_other_windows_errors() {
-        let err =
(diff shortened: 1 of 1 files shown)


mxinden (commented):
I would prefer fixing this issue in `quinn-udp`, thus de-duplicating the logic across the various usages.

https://github.com/quinn-rs/quinn/issues/2293

larseggert:
No objections!

larseggert:
@mxinden https://github.com/quinn-rs/quinn/issues/2293 got closed - what now?

mxinden:
@larseggert let's try to get to the root of the discussion. I prepared https://github.com/quinn-rs/quinn/pull/2594. Can you help out?

neqo-udp/src/lib.rs
```diff
@@ -96,27 +104,27 @@ pub fn send_inner(
...
-        {
-            false
-        }
+        return false;
+    })
+}
+
+fn is_enobufs(e: &io::Error) -> bool {
+    e.raw_os_error().is_some_and(|e| {
+        #[cfg(unix)]
```
martinthomson: This fails with `if cfg!(foo)` doesn't it?  Because `libc::ENOBUFS` is only defined in one place.  Shame, because an if statement would be so much easier to read and validate.

martinthomson (approved):

larseggert:
This is on hold pending the discussion in https://github.com/quinn-rs/quinn/pull/2594

larseggert:
Note to self: Add a test that confirms this is basically packet loss and triggers an RTX.

larseggert:
@mxinden?

neqo-udp/src/lib.rs
```diff
@@ -81,6 +83,13 @@ pub fn send_inner(
             );
             return Ok(());
         }
+        Err(e) if is_enobufs(&e) => {
+            // The send queue is momentarily full. Don't map to WouldBlock: the
+            // socket IS writable, so edge-triggered epoll/kqueue won't re-signal
+            // and the send loop would hang. Drop the packet; QUIC will retransmit.
+            qdebug!("Interface send queue full (ENOBUFS), dropping packet: {e}");
```
mxinden, line 90: I assume this is very rare. I would be fine with `qinfo`.

mxinden (approved):
Thanks. Once this release lands in Firefox we can revert https://phabricator.services.mozilla.com/D288120.

PR #3495, opened 2026-03-20
feat: Increase default UDP send buffer size to 1MB

larseggert:
See https://bugzilla.mozilla.org/show_bug.cgi?id=2024900

### DIFF
--- a/neqo-bin/src/udp.rs
+++ b/neqo-bin/src/udp.rs
@@ -30,20 +30,19 @@ impl Socket {
         let socket = std::net::UdpSocket::bind(addr)?;
         let state = quinn_udp::UdpSocketState::new((&socket).into())?;
 
-        // FIXME: We need to experiment if increasing this actually improves performance.
-        // Also, on BSD and Apple targets, this seems to increase the `net.inet.udp.maxdgram`
-        // sysctl, which is not the same as the socket buffer.
-        // if send_buf_before < ONE_MB {
-        //     state.set_send_buffer_size((&socket).into(), ONE_MB)?;
-        //     let send_buf_after = state.send_buffer_size((&socket).into())?;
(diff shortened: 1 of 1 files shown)


mxinden:
Is there evidence that a larger send buffer improves performance? Intuitively, I can not tell whether it is a performance improvement, or leads to source buffer bloat.

See past discussion in https://github.com/mozilla/neqo/pull/2470#issuecomment-2772077732.

mxinden:
I also commented on Phabricator: https://phabricator.services.mozilla.com/D288864#10030860

larseggert:
Are you concerned 1MB is too much? The defaults are IIRC ~260KB on Linux and less on Mac, so I worry we may be buffer limited when uploading over reasonably fast paths.

mxinden:
Yes, I am worried that 1 MB is too much. In which scenario do you see Firefox exceeding the 260KB OS buffer?

An example:

- given a 50ms RTT connection
- given a 260 KB OS UDP send buffer
- with a single send per RTT we would thus be able to saturate a 40 Mbit/s link
- more realistically, but still very conservative, with 10 sends per RTT we would be able to saturate a 400 Mbit/s link

Given that we pace all sends across 0.5 of the RTT, I don't see us benefiting from a larger OS buffer. With the theoretical benefits of a larger buffer come drawbacks, i.e. buffer bloat, which I assume are far more negative than the benefits.

martinthomson:
Careful, you almost make a good argument for a larger buffer.  Lots of people are on gigabit links.  I don't know how often we'd be able to service timers over 50ms, but I suspect it is more like 3 times than 10 in some cases (if timer granularity isn't increased beyond the typical).

We definitely need to fix that pacing speedup.  It's doing us more harm than good right now.

larseggert:
If we do this, we should do #3898

PR #3533, opened 2026-04-09
feat: add contains_handshake and assert_contains_handshake helpers

### LINKED ISSUE #3447: Add and use `assert_contains_handshake()`
Given where things stand, it seems like it might pay to have a `assert_contains_handshake()` or similar, which would ignore a small Initial packet.  There's a lot of fragility in our current tests due to packet sizing.  

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/2814#discussion_r2928149971_

mvanhorn:
## Summary

Adds `contains_handshake()` and `assert_contains_handshake()` to the test fixture assertions module. These scan through coalesced long-header packets in a datagram to check if any is a Handshake packet.

## Why this matters

Tests that assert on Handshake packets are fragile when packet sizing causes a small Initial to be coalesced before the Handshake. The existing `assert_handshake()` only checks the first packet in a datagram. `assert_contains_handshake()` scans through all coalesced packets.

## Changes

- `test-fixture/src/assertions.rs`: Added `contains_handshake()` and `assert_contains_handshake()` following the same coalesced-packet scanning pattern used by `assert_no_1rtt()`.

## Testing

- Follows the established pattern of `assert_no_1rtt` for scanning coalesced packets
- Could not run `cargo check` locally (NSS system dependency not available); GitHub CI will verify

Fixes #3447

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/neqo-transport/src/connection/tests/recovery.rs
+++ b/neqo-transport/src/connection/tests/recovery.rs
@@ -9,7 +9,9 @@ use std::time::{Duration, Instant};
 use neqo_common::qdebug;
 use neqo_crypto::AuthenticationStatus;
 use test_fixture::{
-    assertions::{assert_handshake, assert_initial, is_handshake, is_initial},
+    assertions::{
+        assert_contains_handshake, assert_handshake, assert_initial, is_handshake, is_initial,
+    },
     now, split_datagram,
 };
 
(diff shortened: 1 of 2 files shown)


larseggert (commented):
Were you planning to actually use these new functions for something?

mvanhorn:
Yes - they're intended for tests where packet sizing causes a small Initial to be coalesced before the Handshake. The existing `assert_handshake()` only checks the first packet in a datagram, so those tests fail intermittently. I have a follow-up PR that uses `assert_contains_handshake()` to fix some of those flaky assertions, but wanted to land the helpers separately to keep the diff focused.

larseggert:
I'd suggest to combine those PRs (into this PR), so we can see whether those additions have the right shape overall.

mvanhorn:
Pushed a consumer in `pto_handshake_complete` (recovery.rs) that uses `assert_contains_handshake`.

That test had two `assert_handshake(...)` calls commented out at the point where the server hands back its second handshake burst, with the inline note that the datagram might have extra Initial data coalesced at the front. That is exactly what this helper is for, so I re-enabled both assertions against `assert_contains_handshake`.

The rest of the `assert_handshake` call sites in `recovery.rs` and `handshake.rs` run on packets that have already been extracted via `split_datagram`, or on datagrams where the first packet is deterministically the Handshake (e.g. after `client.authenticated(...)`), so migrating them would just obscure the assertion.

Ran `cargo test -p neqo-transport --lib pto_handshake_complete -- --test-threads=1` five times in a row, all green, and the full `connection::tests::recovery` (21 tests) and `connection::tests::handshake` (42 tests) modules also pass. `cargo fmt --check` (nightly) and `cargo clippy -p neqo-transport --tests -- -D warnings` are clean.

Happy to grow the migration if you have other specific tests in mind.

larseggert (approved):

PR #3551, opened 2026-04-10
fix: Don't serve h09 requests before stream FIN

larseggert:
The h09 server was processing requests as soon as any stream data arrived, without waiting for the stream FIN. During 0-RTT, stream data for a single request is often split across multiple packets, so the server would see a partial URL, fail to find the corresponding file, and respond with a 3-byte "404" — causing the zerortt interop test to fail. Fix by buffering incoming data until FIN is received before attempting to parse and serve the request. Also factor out the shared QNS file-lookup logic into response_for_path() and the common server setup into configure_server(), and add tests.

### DIFF
--- a/neqo-bin/src/server/http09.rs
+++ b/neqo-bin/src/server/http09.rs
@@ -5,7 +5,6 @@
 // except according to those terms.
 
 use std::{
-    borrow::Cow,
     cell::RefCell,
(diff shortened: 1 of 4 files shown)


larseggert:
@mxinden any chance for a review?

mxinden (approved):

PR #3552, opened 2026-04-10
fix: Make the server obey the amplification limit

larseggert:
The server was validating the client's address — and dropping the amplification limit — upon receiving a retransmitted client Initial whose DCID matched the server's chosen source CID. RFC 9000 Section 8.1 designates Handshake packets as the validation mechanism; the CID-based path (a MAY) is weaker here because Initial keys are publicly derivable, so the SCID is visible to any on-path observer.

In practice, with MLKEM the ClientHello spans two Initial packets. If only the first reaches the server (as in the QNS amplificationlimit test with drops_to_server=2,3,4,5,6,7), the server sends PTO probes that exhaust most of the amplification budget. When the client retransmits the second fragment using the server's SCID as DCID, the old code validated the path and let the server send the full ~12 KB Handshake flight unrestricted — over 4× amplification.

Fix: remove the DCID condition from postprocess_packet; only Handshake packets validate the path during setup.

Test: anti_amplification_initial_retransmit_no_validate simulates the exact scenario — first fragment only, server ACKs it, client PTO retransmits the second fragment with the server's SCID as DCID — and asserts the server never exceeds 3× amplification across the full paced burst.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1686,9 +1686,7 @@ impl Connection {
 
         if self.state.connected() {
             self.handle_migration(path, remote, migrate, now);
-        } else if self.role != Role::Client
-            && packet.packet_type() == packet::Type::Handshake
-        {
+        } else if self.role != Role::Client && packet.packet_type() == packet::Type::Handshake {
             // We only allow one path during setup, so apply handshake
             // path validation to this path.
             path.borrow_mut().set_valid(now);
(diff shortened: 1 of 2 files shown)


martinthomson (changes requested):
Removing a condition that provides path validation in order to make tests pass is backwards.  The fix is to ensure that the test correctly measures the thing we care about.

larseggert:
@marten-seemann has been sitting on https://github.com/quic-interop/quic-interop-runner/pull/442 for a while and I got impatient with the spurious failures.

PR #3558, opened 2026-04-13
fix: Improve build script idiomaticity per Cargo docs

larseggert:
- Add missing `cargo:` prefixes on `rerun-if-env-changed`, `rustc-env`, and `warning` directives (were silently ignored by Cargo before)
- Replace `env::consts::OS` with `CARGO_CFG_TARGET_OS` and `TARGET` string matching with `CARGO_CFG_TARGET_ARCH`/`CARGO_CFG_TARGET_OS` for correct cross-compilation behavior
- Add `rerun-if-changed` for `src/min_version.rs` and `min_version.txt` (needed since the build script uses explicit rerun tracking)
- Replace `eprintln!` with `cargo:warning=` for user-facing warnings
- Remove dead `cargo:include=` emissions (no `links` key, so metadata was never forwarded to dependents)
- Delete `fuzz/build.rs` and `neqo-transport/build.rs` (redundant with workspace `unexpected_cfgs` lint which already registers `fuzzing`)

### DIFF
--- a/fuzz/build.rs
+++ b/fuzz/build.rs
@@ -1,9 +1,0 @@
-// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
-// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
-// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
-// option. This file may not be copied, modified, or distributed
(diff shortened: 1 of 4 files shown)


larseggert:
@Not-Nik some of this will need to move to nss-rs

PR #3563, opened 2026-04-16
chore: mtu@0.4.1

larseggert:
~~Because apparently in a workspace you can't just `cargo publish` in a subdirectory...~~


mxinden:
> Because apparently in a workspace you can't just `cargo publish` in a subdirectory...

I am not following. Can you expand? You can publish individual crates of a workspace.

larseggert:
> I am not following. Can you expand? You can publish individual crates of a workspace.

I got confused because the `mtu` crate was published with the `neqo` README. Because of the `Cargo.toml` settings, not because of what I thought happened.

PR #3564, opened 2026-04-16
feat(cc): add parameter to disable spurious congestion recovery

omansfeld:
#3298 landed in neqo release `v0.22.0`, which got vendored into Firefox 149. In release telemetry we see an increase in [http/3 upload throughput](https://glam.telemetry.mozilla.org/fog/probe/networking_http_3_upload_throughput/explore?aggregationLevel=version&app_id=release&visiblePercentiles=%5B95%2C75%2C50%2C25%5D) across all percentiles and up to **+12.9%** in the 50pct going from 148 to 149.

Spurious congestion event recovery is the only performance impacting feature we landed in neqo in 149, so I would like to run an experiment to confirm impact.

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -1744,7 +1744,7 @@ mod tests {
         cc.on_packet_sent(&pkt2, now);
         assert_eq!(cc.current.phase, Phase::SlowStart);
 
-        // 2. Lose packets (1, 2) --> `RecoveryStart`, reduced cwnd, but no state saved
+        // 2. Lose packets (1, 2) --> `RecoveryStart` and reduced cwnd
         let mut lost_pkt1 = pkt1.clone();
         let mut lost_pkt2 = pkt2.clone();
         lost_pkt1.declare_lost(now, sent::LossTrigger::TimeThreshold);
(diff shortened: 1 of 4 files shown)


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -718,6 +726,12 @@ where
     }
 
     fn on_spurious_congestion_event(&mut self, cc_stats: &mut CongestionControlStats) {
+        if !self.spurious_recovery {
+            cc_stats.congestion_events.spurious += 1;
+            qinfo!("[{self}] Spurious cong event detected -> recovery disabled;");
+            return;
+        }
```
mxinden, line 733: If you were to always set the undo state, no matter whether `self.spurious_recovery` is enabled or not, could this function then:

1. Do the `self.stored.take()` check.
2. Do the `stored.congestion_window <= self.current.congestion_window` check.
3. Increase the metric (i.e. `        cc_stats.congestion_events.spurious += 1;`).
4. Return on `!self.spurious_recovery`.
5. Otherwise continue as is?

That would not duplicate ` cc_stats.congestion_events.spurious += 1;` and more generally, make the two code paths (`self.spurious_recovery` `true` or `false`) more similar.

mxinden (approved):
Very curious what the experiment will show.

neqo-transport/src/cc/classic_cc.rs
omansfeld: That’d work. I first had something like that but then went for not storing the prior state.

De-dublication vs. unnecessary state. No real preference for either on my end.

omansfeld: For the record: I went with your suggestion because it makes sure that the spurious detection metric is the exact same between `spurious_recovery` `true`/`false`.

PR #3573, opened 2026-04-24
 fix: gate SCONE Initial padding on connection parameter

mxinden:
The `scone` connection parameter (off by default since https://github.com/mozilla/neqo/pull/3492) only gated advertisement of the SCONE transport parameter. The Initial packet SCONE indication (`0xc8 0x13`) was still emitted unconditionally, so every default-config client was signaling SCONE on the wire without actually negotiating it.

This is suspected — not yet confirmed — to be the cause of a Firefox 150 regression where Facebook's edge reacts to the indicator and Bitdefender's HTTP/3 inspection then trips on the altered server-side
frames (bug 2034178). The QUIC handshake still completes, so the H3->H2 fallback never kicks in.

Gate both the build-time reservation and the pad-time indicator on `scone_enabled()`; when disabled, pad Initials with zeros as before SCONE was introduced.

---

Includes the version bump to `v0.24.2`.

Pull request targets the `v0.24` branch, branched off of the `v0.24.1` tag.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2689,7 +2689,9 @@ impl Connection {
                 path.borrow().pmtud().probe_size()
             } else {
                 profile.limit()
-                    - if space == PacketNumberSpace::Initial {
+                    - if space == PacketNumberSpace::Initial
+                        && self.conn_params.scone_enabled()
+                    {
                         // Reserve some space for the SCONE indication in an Initial.
                         // This reduces the amount available for building the packet,
                         // but we'll pad to `profile.limit()` when padding.
@@ -2857,16 +2859,20 @@ impl Connection {
         );
         let pad_amount = profile.limit() - encoder.len();
         initial.track_padding(pad_amount);
-        // This ensures that the last bytes are a SCONE indication, if there is enough space.
-        // This is not tracked, other than for congestion control (above)
-        if pad_amount >= Self::SCONE_INDICATION.len() {
-            encoder.pad_to(
-                profile.limit() - Self::SCONE_INDICATION.len() + 1,
-                Self::SCONE_INDICATION[0],
-            );
-            encoder.encode(&Self::SCONE_INDICATION[1..]);
+        if self.conn_params.scone_enabled() {
+            // This ensures that the last bytes are a SCONE indication, if there is enough space.
+            // This is not tracked, other than for congestion control (above)
+            if pad_amount >= Self::SCONE_INDICATION.len() {
+                encoder.pad_to(
+                    profile.limit() - Self::SCONE_INDICATION.len() + 1,
+                    Self::SCONE_INDICATION[0],
+                );
+                encoder.encode(&Self::SCONE_INDICATION[1..]);
+            } else {
+                encoder.pad_to(profile.limit(), Self::SCONE_INDICATION[0]);
+            }
         } else {
-            encoder.pad_to(profile.limit(), Self::SCONE_INDICATION[0]);
+            encoder.pad_to(profile.limit(), 0);
         }
     }
 
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -806,8 +806,7 @@ fn corrupted_initial() {
         .iter()
         .enumerate()
         .rev()
-        .skip(1) // Skip the last byte, which might be a SCONE indicator.
-        .find(|&(_, &v)| v != Connection::SCONE_INDICATION[0]) // The SCONE padding value.
+        .find(|&(_, &v)| v != 0)
         .unwrap();
     corrupted[idx] ^= 0x76;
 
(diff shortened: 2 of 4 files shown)


KershawChang (approved):
Looks good

larseggert:
CC @martinthomson since this was his change originally.

larseggert (approved):

larseggert (approved):
clippy needs fixing

mxinden:
> clippy needs fixing

@larseggert do you feel strongly about it? As far as I can tell, clippies complaints are unrelated. I would like to keep the change minimal:

```
  error: this `match` expression can be replaced with `?`
     --> neqo-transport/src/addr_valid.rs:188:9
      |
  188 | /         match dec.decode_uint::<u32>() {
  189 | |             Some(d) => {
  190 | |                 let end = self.start_time + Duration::from_millis(u64::from(d));
  191 | |                 if end < now {
  ...   |
  196 | |             None => return None,
  197 | |         }
      | |_________^
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#question_mark
      = note: `-D clippy::question-mark` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(clippy::question_mark)]`
  help: try instead
      |
  188 ~         {
  189 +             let d = dec.decode_uint::<u32>()?;
  190 +             let end = self.start_time + Duration::from_millis(u64::from(d));
  191 +             if end < now {
  192 +                 qtrace!("Expired token: {end:?} vs. {now:?}");
  193 +                 return None;
  194 +             }
  195 +         }
      |
```

larseggert:
If unrelated, I'm ok. Do a different PR for them (or I can after lunch)

mxinden:
@larseggert note that this is targeting the `v0.24` branch, branched off of `v0.24.1` tag. In other words, this is not our most recent `main` with all Clippy lint fixed.

larseggert:
Ah. Sorry. Jet lagged with 2h sleep and a cancelled plane in LHR. Am not fully coherent.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -806,8 +806,7 @@ fn corrupted_initial() {
         .iter()
         .enumerate()
         .rev()
-        .skip(1) // Skip the last byte, which might be a SCONE indicator.
-        .find(|&(_, &v)| v != Connection::SCONE_INDICATION[0]) // The SCONE padding value.
```
martinthomson, line 810: This change is no good.

martinthomson (commented):
If there were protocol police, I would be ringing them right now.  We should be free to pad packets with whatever noise we like.

neqo-transport/src/connection/mod.rs
```diff
@@ -2857,16 +2859,20 @@ impl Connection {
...
-        if pad_amount >= Self::SCONE_INDICATION.len() {
-            encoder.pad_to(
-                profile.limit() - Self::SCONE_INDICATION.len() + 1,
-                Self::SCONE_INDICATION[0],
-            );
-            encoder.encode(&Self::SCONE_INDICATION[1..]);
+        if self.conn_params.scone_enabled() {
+            // This ensures that the last bytes are a SCONE indication, if there is enough space.
+            // This is not tracked, other than for congestion control (above)
+            if pad_amount >= Self::SCONE_INDICATION.len() {
```
martinthomson, line 2865: FWIW, this is not at all how I would have "fixed" this.  I would have changed this to:


```suggestion
            if pad_amount >= Self::SCONE_INDICATION.len() && self.conn_params.scone_enabled() {
```

mxinden: At this point I don't know whether BitDefender trips over `Self::SCONE_INDICATION[0]` or `Self::SCONE_INDICATION[1]`, thus I opted for disabling both on `self.conn_params.scone_enabled()`, using `0` padding instead as we did before.

@martinthomson is there much value in skipping `Self::SCONE_INDICATION[0]` but including `Self::SCONE_INDICATION[1]`?

neqo-transport/src/connection/tests/handshake.rs
mxinden: Note that this is simply a revert of https://github.com/mozilla/neqo/pull/2814. Fine for now in tests?

mxinden (commented):
> If there were protocol police, I would be ringing them right now. We should be free to pad packets with whatever noise we like.

We are in touch with BitDefender. I assume they will fix their bug soon. Once fixed, we can re-enable SCONE. We expect a large amount of users to be affected by this BitDefender bug. I doubt the majority of users running into this bug associate it BitDefender instead of Firefox. Given that SCONE doesn't provide user value yet, I assumed temporarily disabling its padding is worth those users being able to access facebook.com.

Do you disagree with this strategy @martinthomson? What should we do instead?

martinthomson:
You've taken the conservative approach, which is probably sensible given the current circumstances.  As long as we revert the entire thing when we sort out the BitDefender bug, we'll be fine.  (I worry that Facebook is also reacting to our signal, even without the transport parameter.  That's worth digging into as well.)

PR #3582, opened 2026-04-30
fix: Does symbol localization actually help?

larseggert:
See slides 80+ of https://raw.githubusercontent.com/yugr/rust-slides/main/EN.pdf


larseggert:
Seems to make no difference and is super intrusive.

PR #3589, opened 2026-05-07
fix(cc): lower pacing speedup when not in slow start

### LINKED ISSUE #3114: transport: consider lower pacing rate in congestion avoidance
Neqo currently paces over half the RTT, i.e. 2x the current rate:

https://github.com/mozilla/neqo/blob/207d309e941d3ae9384c191c13463dc81fca4ce9/neqo-transport/src/pace.rs#L19-L27

This is independent of the congestion control phase, i.e. same in slow start and congestion avoidance.

In slow start, we likely want to keep the aggressive pacing rate to keep up with the cwnd doubling.

In congestion avoidance, to prevent overflowing buffers and thus causing packet loss, we might want to reduce the rate to e.g. 1.2x times the rate.

What do others do?

- [Linux TCP stack](https://github.com/torvalds/linux/blob/a1388fcb52fcad3e0b06e2cdd0ed757a82a5be30/net/ipv4/tcp_input.c#L1114-L1115) uses 2x in slow start and 1.2x in congestion avoidance.
- [quic-go](https://github.com/quic-go/quic-go/blob/eb7fcf5637f780fe1355002534430d95bb5b53c8/internal/congestion/pacer.go#L27-L31) uses 1.25x

//CC @MarcoPolo and @marten-seemann since we discussed this over the week.
//CC @mwelzl thanks for the input on Linux.

omansfeld:
Very ugly (claude generated) but working draft pull request to lower the pacing speedup factor to `1.2x` while not in slow start. Using this to test and get some data for now, will need eventual cleanup once we decide to ship.

The factor is just copied from the Linux kernel: https://github.com/torvalds/linux/blob/8ab992f815d6736b5c7a6f5fd7bfe7bc106bb3dc/net/ipv4/tcp_ipv4.c#L3491

This is getting some good initial results when running the `min_bandwidth` benchmark, as seen below.

main:
```bash
~/g/n/neqo (main|✔) $ RUST_LOG=info cargo bench --bench min_bandwidth --features bench &| grep "gbit-bandwidth.*achieved"
7.940 INFO gbit-bandwidth-noecn achieved 620.4358903732716 Mb/s bandwidth (link rate 1000)
17.762 INFO gbit-bandwidth-ecn achieved 554.4264457615302 Mb/s bandwidth (link rate 1000)
```
fix/pacing:
```bash
~/g/n/w/n/w/neqo (fix/pacing|✔) $ RUST_LOG=info cargo bench --bench min_bandwidth --features bench &| grep "gbit-bandwidth.*achieved"
9.208 INFO gbit-bandwidth-noecn achieved 778.3476704714313 Mb/s bandwidth (link rate 1000)
22.907 INFO gbit-bandwidth-ecn achieved 787.7023838948422 Mb/s bandwidth (link rate 1000)
```

I'm currently working on running some performance tests on real paths, trying to leverage and extend @mxinden's [fireflare](https://github.com/mxinden/fireflare) tooling.

fixes #3114 (eventually)

cc: @larseggert

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -16,6 +16,9 @@ use neqo_common::qtrace;
 
 use crate::rtt::GRANULARITY;
 
+/// Scale factor for [`Pacer::SPEEDUP_SLOW_START`] and [`Pacer::SPEEDUP`].
+pub(crate) const SPEEDUP_SCALE: usize = 10;
+
 /// A pacer that uses a leaky bucket.
 pub struct Pacer {
     /// Whether pacing is enabled.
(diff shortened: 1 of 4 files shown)


omansfeld:
Reminder for self about discussion with @larseggert:

- we could add the speedup to the `Phase` enum and use that instead of all the boilerplate passing things around in the current patch

Just to prompt myself to look into it again, I'm sure I'm not entirely getting this right currently, so feel free to add something Lars. Or I'll reach out again when I clean this up.

larseggert:
> we could add the speedup to the Phase

I meant you could use an `EnumMap<Phase>` for the various speedup `const`s.

neqo-transport/src/pace.rs
```diff
@@ -72,6 +82,37 @@ impl Pacer {
         self.p = mtu;
     }
 
+    pub(crate) fn set_speedup(&mut self, speedup: usize) {
+        self.speedup = speedup;
+    }
```
martinthomson, line 87: Please don't impose details of the internals of this class to its clients.  It might be enough to have a `set_slowstart(&mut self, on: bool)` method, where this function manages the extra noise.  See also the changes in #3594.  You'll want to rebase on that.

martinthomson (changes requested):

PR #3595, opened 2026-05-08
feat: skip terminal-stream scan when no streams have completed

larseggert:
`cleanup_closed_streams` — called on every timer tick and every input batch — unconditionally scanned all send and recv streams for terminal states. With many concurrent streams this is O(n) work per packet even when nothing has finished.

Add a `has_terminal` flag to `SendStreams` and `RecvStreams`, set precisely when a stream transitions to a terminal state. This allows `remove_terminal` and `clear_terminal` to return early when the flag is unset, reducing the common case from O(n) to O(1).


neqo-transport/src/streams.rs
```diff
@@ -316,9 +319,17 @@ impl Streams {
         self.recv.clear();
     }
 
+    /// Note that a recv stream has become terminal (e.g. the application read the FIN).
+    pub fn note_recv_terminal(&mut self) {
+        self.recv.note_terminal();
```
martinthomson: This name is a bit off.  How about "stream_ended()"?

mxinden (commented):
`RecvStream::read` is a `pub` API and the "must call  `stream_ended()` on fin" rule lives only in a comment. A future caller might forget. Can we do better? Is it worth the introduced complexity?

larseggert:
@mxinden would you take another look?

neqo-transport/src/recv_stream.rs
```diff
@@ -820,7 +877,8 @@ impl RecvStream {
         }
     }
 
-    pub fn stop_sending(&mut self, err: AppError) {
+    #[must_use]
+    pub fn stop_sending(&mut self, err: AppError) -> bool {
```
martinthomson, line 889: Could you document what the meaning of the boolean is?  I think that it indicate whether the stream was terminated.

neqo-transport/src/recv_stream.rs
```diff
@@ -673,7 +731,7 @@ impl RecvStream {
 
     /// # Errors
     /// When the reset occurs at an invalid point.
-    pub fn reset(&mut self, application_error_code: AppError, final_size: u64) -> Res<()> {
+    pub fn reset(&mut self, application_error_code: AppError, final_size: u64) -> Res<bool> {
```
martinthomson, line 738: Again, documentation of the return value would help.

neqo-transport/src/send_stream.rs
```diff
@@ -1642,11 +1646,20 @@ impl SendStreams {
...
-    pub fn remove_terminal(&mut self) {
+    /// Remove ended streams. Returns `true` if any were removed.
+    #[must_use]
+    pub fn remove_ended(&mut self) -> bool {
+        if !self.has_ended {
+            return false;
+        }
+        self.has_ended = false;
+        let mut removed = false;
         self.map.retain(|stream_id, stream| {
```
martinthomson: This looks like a grand candidate for `Map::extract_if` as a further optimization.

```rust
for s in self.map.extract_if(.., Stream::is_ended) {
  removed = true;
  if stream.is_fair() {  // ...  all that stuff
  }
}
removed
```

neqo-transport/src/streams.rs
```diff
@@ -316,11 +315,24 @@ impl Streams {
         self.recv.clear();
     }
 
+    /// # Errors
+    /// When the stream does not exist or has no more data.
+    pub fn recv(&mut self, stream_id: StreamId, data: &mut [u8]) -> Res<(usize, bool)> {
```
martinthomson, line 323: We don't do a great job of documenting return values, do we?

martinthomson (approved):

PR #3598, opened 2026-05-10
feat: Define a path migration event

martinthomson:


larseggert (approved):
LGTM; minor clippy nit.

PR #3600, opened 2026-05-11
chore: Benchmark "bypass SSL AEAD wrapper in RecordProtection"

larseggert:
Try of https://github.com/mozilla/nss-rs/pull/49 in neqo. #3601 is even faster, so I prefer landing that in neqo.


larseggert:
OBE by #3620

PR #3601, opened 2026-05-11
feat: use the `nss-rs` `blapi` feature

larseggert:
Once https://github.com/mozilla/nss-rs/pull/52 lands. This is significantly faster than the alternative in #3600, so I would prefer to use this for neqo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ hex = { version = "0.4", default-features = false }
 http = { version = "1", default-features = false, features = ["std"] }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
-nss = { rev = "af943cfecb083270beaeb8845f6aa5b79684212e", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
-nss-test-fixture = { rev = "af943cfecb083270beaeb8845f6aa5b79684212e", package = "test-fixture", git = "https://github.com/mozilla/nss-rs" }
+nss = { rev = "0.12.0", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
+nss-test-fixture = { rev = "0.12.0", package = "test-fixture", git = "https://github.com/mozilla/nss-rs" }
 qlog = { version = "0.16.0", default-features = false }
 quinn-udp = { version = "0.6", default-features = false, features = ["log", "fast-apple-datapath"] }
 rustc-hash = { version = "2.1", default-features = false, features = [ "std" ]}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ hex = { version = "0.4", default-features = false }
 http = { version = "1", default-features = false, features = ["std"] }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
-nss = { rev = "0.11.0", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
-nss-test-fixture = { rev = "0.11.0", package = "test-fixture", git = "https://github.com/mozilla/nss-rs" }
+nss = { rev = "af943cfecb083270beaeb8845f6aa5b79684212e", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
+nss-test-fixture = { rev = "af943cfecb083270beaeb8845f6aa5b79684212e", package = "test-fixture", git = "https://github.com/mozilla/nss-rs" }
 qlog = { version = "0.16.0", default-features = false }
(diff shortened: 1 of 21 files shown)


martinthomson (commented):
I'm sad to say that this looks really, really good.  About the same sort of performance lift as the aws-lc patch you put together; perhaps even better and more consistent.  So I'm going to suggest that we continue to explore this one.

Cargo.toml
```diff
@@ -36,7 +36,7 @@ hex = { version = "0.4", default-features = false }
 http = { version = "1", default-features = false, features = ["std"] }
 libc = { version = "0.2", default-features = false }
 log = { version = "0.4", default-features = false }
-nss = { rev = "0.9.0", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
+nss = { rev = "2f5bbfe4fe3bce69b9b73f6e3d9a115e30bfaeb2", package = "nss-rs", git = "https://github.com/mozilla/nss-rs" }
```
larseggert: This needs to be updated to the `nss-rs` release version that ships the `blapi` feature.

neqo-transport/src/crypto.rs
```diff
@@ -489,7 +498,13 @@ impl CryptoDxState {
             version,
             direction,
             epoch: usize::from(epoch),
-            aead: Aead::new(TLS_VERSION_1_3, cipher, secret, version.label_prefix())?,
+            aead: Aead::new(
+                TLS_VERSION_1_3,
+                cipher,
+                secret,
+                version.label_prefix(),
+                direction.into(),
```
martinthomson:
```suggestion
                Mode::from(direction),
```

Just a preference.

neqo-transport/src/crypto.rs
```diff
@@ -581,6 +596,7 @@ impl CryptoDxState {
                 cipher,
                 next_secret,
                 self.version.label_prefix(),
+                self.direction.into(),
```
martinthomson:
```suggestion
                Mode::from(self.direction),
```

neqo-transport/src/crypto.rs
```diff
@@ -734,21 +757,35 @@ impl CryptoDxState {
     #[cfg(not(feature = "disable-encryption"))]
     #[cfg(test)]
     pub(crate) fn test_default() -> Self {
```
martinthomson: Why not have this take a Read/Write arg?  Or at least call it `test_default_write()`?

neqo-transport/src/crypto.rs
```diff
@@ -708,6 +724,13 @@ impl CryptoDxState {
     }
 
     #[must_use]
+    #[cfg(not(feature = "disable-encryption"))]
+    pub const fn expansion(&self) -> usize {
+        self.aead.expansion()
+    }
+
+    #[must_use]
+    #[cfg(feature = "disable-encryption")]
```
martinthomson: Any way we can get the disable-encryption variant of this to be `const`?  This is awkward.

neqo-transport/src/packet/retry.rs
```diff
@@ -22,24 +22,33 @@ fn make_aead(version: Version) -> Aead {
         TLS_AES_128_GCM_SHA256,
         &secret,
         version.label_prefix(),
+        mode,
     )
     .expect("can create AEAD")
 }
 #[cfg(feature = "draft-29")]
```
martinthomson, line 32: When do you think we can get rid of draft-29?  It's been a little while now.

martinthomson (approved):
This is really two changes:

1. Tracking the API changes in the NSS crate
2. Switching to blapi

It might have been nicer to do them as two steps, but this is fine.

neqo-transport/src/packet/retry.rs
larseggert: https://github.com/mozilla/neqo/issues/3623

neqo-http3/Cargo.toml
```diff
@@ -45,6 +45,8 @@ bench = [
         "test-fixture/bench",
         "log/release_max_level_info",
 ]
+blapi = ["neqo-transport/blapi"]
+default = ["blapi"]
```
mxinden, line 49: @larseggert why is this needed? Does qpack make use of any of the `blapi` features? If not, since Rust unifies Cargo features, why explicitly enable it here?

larseggert: Might be a leftover - I was struggling for a bit to get the feature to propagate across benches, etc.

mxinden: See https://github.com/mozilla/neqo/pull/3629.

PR #3606, opened 2026-05-12
chore: Better `perf` sampling

larseggert:
Collect cycle-accurate attribution for user space, and capture kernel-mode execution during syscalls.

Record context switches as PERF_RECORD_SWITCH events. Samply imports these and shows them as markers.

Increase perf ring buffer to 8MB.


mxinden (approved):
🚀 the more we can pass to samply the better.

larseggert:
This needs https://github.com/mstange/samply/pull/816 to be in a `samply` release.

larseggert:
Think I found a way to make this work without a new samply release.

PR #3613, opened 2026-05-12
feat(cc/stats): Add stats for SEARCH slow start heuristic

omansfeld:
## Summary

Adds telemetry fields to `CongestionControlStats` for tuning and evaluating the SEARCH slow start exit heuristic. Each commit adds one metric with its own test. Uploading as one PR to avoid merge conflicts, as commits touch the same files. If there are concerns with a particular metric it should be easy to pull out so the rest can land swiftly.

Is probably easiest to review commit-by-commit.

I will work on the Gecko side of this now.

## Commits

1. [**record drain-phase target cwnd**](https://github.com/omansfeld/neqo/commit/1e29e0d4) — Records BDP estimates for empty and full buffers at slow start exit. The [SEARCH draft](https://datatracker.ietf.org/doc/html/draft-chung-ccwg-search-09#section-3.2-17) suggests a drain-phase using the empty-buffer target; we record them for evaluation without implementing the drain.

2. [**record EXTRA_BINS needed**](https://github.com/omansfeld/neqo/commit/4567d82e) — Tracks the running maximum of lookback bins needed when RTT inflation prevents SEARCH from evaluating. Helps determine if `EXTRA_BINS=15` is sufficient.

3. [**record norm_diff for THRESH tuning**](https://github.com/omansfeld/neqo/commit/2735ca44) — Tracks the running maximum of the normalized sent/acked difference during `Continue` (i.e. non-exiting) outcomes. Helps evaluate whether `THRESH=0.26` is well-calibrated.

4. [**record if a reset happened**](https://github.com/omansfeld/neqo/commit/e63632d1) — Tracks reset count and maximum bins skipped across all resets via a `SearchResetStats` struct. Helps evaluate whether the reset logic is well-calibrated.

5. [**record zero bytes sent count**](https://github.com/omansfeld/neqo/commit/3063e812) — Counts how often SEARCH observes zero sent bytes in the previous RTT. Provides insight into app-limited behavior during slow start.

6. [**record first and second RTT**](https://github.com/omansfeld/neqo/commit/04088da9) — Captures the `latest_rtt` from the first two ACKs to evaluate whether the initial RTT (which sets `bin_duration`) is inflated and whether a fix using `min(first, second)` would be viable.

7. [**expose min_rtt in connection stats**](https://github.com/omansfeld/neqo/commit/458a9912) — Adds `min_rtt` to `Stats` for comparison with the recorded first/second RTT samples. Helps determine if `initial_rtt` inflation is an issue.

### DIFF
--- a/neqo-transport/src/cc/search.rs
+++ b/neqo-transport/src/cc/search.rs
@@ -135,7 +135,11 @@ impl Search {
         clippy::cast_possible_truncation,
         reason = "casting small usize to u32 for use in Duration::saturating_mul"
     )]
-    fn update_bins(&mut self, now: Instant) -> Option<usize> {
+    fn update_bins(
+        &mut self,
+        now: Instant,
+        cc_stats: &mut CongestionControlStats,
+    ) -> Option<usize> {
         let mut curr_idx = self.curr_idx?;
         let mut bin_end = self.bin_end?;
 
@@ -159,6 +163,13 @@ impl Search {
             qdebug!(
                 "SEARCH: update_bins: resetting because we skipped {passed_bins} bins (limit {})",
                 Self::W
+            );
+            cc_stats.search_reset.count += 1;
+            cc_stats.search_reset.max_passed_bins = Some(
+                cc_stats
+                    .search_reset
+                    .max_passed_bins
+                    .map_or(passed_bins, |c| c.max(passed_bins)),
             );
             self.reset();
             return None;
@@ -390,7 +401,7 @@ impl SlowStart for Search {
             return None;
         }
 
-        let curr_idx = self.update_bins(now)?;
+        let curr_idx = self.update_bins(now, cc_stats)?;
 
         // NOTE: SEARCH draft-09 implements a drain-phase to gradually lower the congestion window
         // towards the approximated empty-buffer BDP. I think that could be counter-intuitive while
--- a/neqo-transport/src/cc/search.rs
+++ b/neqo-transport/src/cc/search.rs
@@ -27,7 +27,7 @@ pub enum Outcome {
     /// Not enough data to run SEARCH evaluation (expected early in the connection or after reset).
     WarmingUp,
     /// Can't run SEARCH evaluation because RTT inflated past the point where there isn't enough
-    /// data to look back one RTT. Provides the amount of bins that SEARCH tried to look back.
+    /// data to look back one RTT. Provides the number of bins that SEARCH tried to look back.
     RttInflated(usize),
     /// Haven't sent data for the last RTT thus can't evaluate.
     ZeroSent,
@@ -56,7 +56,7 @@ pub struct Search {
     /// Tracking amount of sent bytes on this connection. Is incremented on every sent packet.
     sent_bytes: usize,
     /// The RTT used to initialize SEARCH (set in [`Self::initialize`]).
-    initial_rtt: Duration,
+    initial_rtt: Option<Duration>,
 }
 
 impl Display for Search {
@@ -98,7 +98,7 @@ impl Search {
             bin_duration: Duration::from_millis(0),
             acked_bytes: 0,
             sent_bytes: 0,
-            initial_rtt: Duration::ZERO,
+            initial_rtt: None,
         }
     }
 
@@ -108,7 +108,7 @@ impl Search {
         reason = "casting small constant usize to u32 for use in Duration::div"
     )]
     fn initialize(&mut self, initial_rtt: Duration, now: Instant) {
-        self.initial_rtt = initial_rtt;
+        self.initial_rtt = Some(initial_rtt);
         // BIN_DURATION = WINDOW_SIZE / W = initial_rtt * WINDOW_SIZE_FACTOR / W
         self.bin_duration =
             initial_rtt * Self::WINDOW_SIZE_FACTOR / Self::SCALE as u32 / Self::W as u32;
@@ -165,12 +165,8 @@ impl Search {
                 Self::W
             );
             cc_stats.search_reset.count += 1;
-            cc_stats.search_reset.max_passed_bins = Some(
-                cc_stats
-                    .search_reset
-                    .max_passed_bins
-                    .map_or(passed_bins, |c| c.max(passed_bins)),
-            );
+            cc_stats.search_reset.max_passed_bins =
+                cc_stats.search_reset.max_passed_bins.max(Some(passed_bins));
             self.reset();
             return None;
         }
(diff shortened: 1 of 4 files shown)


neqo-transport/src/cc/search.rs
```diff
@@ -95,6 +99,7 @@ impl Search {
             bin_duration: Duration::from_millis(0),
             acked_bytes: 0,
             sent_bytes: 0,
+            initial_rtt: Duration::ZERO,
```
larseggert: Should this be an `Option`, to avoid special-casing `Duration::ZERO`?

neqo-transport/src/cc/search.rs
```diff
@@ -275,6 +292,38 @@ impl Search {
...
+    /// implementing this, but do record the calculated targets for evaluation in this function.
+    ///
+    /// The draft specifies using the initial rtt to approximate an 'empty-buffer BDP'.
+    ///
+    /// In addition also record an estimate using the current rtt to approximate BDP with full
+    /// buffers. This should estimate the upper end of the CUBIC curve during congestion avoidance.
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "casting small u128 division result to usize"
+    )]
```
larseggert: Can we avoid `u128` math here, esp. division? It's slow.

neqo-transport/src/cc/search.rs
```diff
@@ -21,13 +21,15 @@ use crate::{cc::classic_cc::SlowStart, packet, rtt::RttEstimate, stats::Congesti
...
+    /// Evaluation ran and slow start should be continued. Provides the normalized difference
+    /// between sent and acked bytes, which can be used in metrics to tune the exit threshold.
+    Continue(usize),
     /// Not enough data to run SEARCH evaluation (expected early in the connection or after reset).
     WarmingUp,
-    /// Can't run SEARCH evaluation because RTT inflated past the point where there is enough
-    /// data to look back one RTT.
-    RttInflated,
+    /// Can't run SEARCH evaluation because RTT inflated past the point where there isn't enough
+    /// data to look back one RTT. Provides the amount of bins that would've been needed to have
```
larseggert:
```suggestion
    /// Can't run SEARCH evaluation because RTT inflated past the point where there isn't enough
    /// data to look back one RTT. Provides the number of bins that would have needed to have
```

neqo-transport/src/cc/search.rs
```diff
@@ -358,12 +414,39 @@ impl SlowStart for Search {
...
+            Outcome::Exit(cwnd) => {
+                self.record_target_cwnd_estimates(curr_idx, rtt, cc_stats);
+                Some(cwnd)
+            }
+            Outcome::RttInflated(lookback_bins) => {
+                cc_stats.search_lookback_bins_needed = Some(
+                    cc_stats
+                        .search_lookback_bins_needed
+                        .map_or(lookback_bins, |c| c.max(lookback_bins)),
+                );
```
larseggert:
```suggestion
                cc_stats.search_lookback_bins_needed = cc_stats.search_lookback_bins_needed.max(Some(lookback_bins));
                );
```
And similar for wherever else you use this pattern.

neqo-transport/src/cc/search.rs
```diff
@@ -275,6 +292,38 @@ impl Search {
...
+    const fn record_target_cwnd_estimates(
+        &self,
+        curr_idx: usize,
+        rtt: Duration,
+        cc_stats: &mut CongestionControlStats,
+    ) {
+        let initial_rtt_bins = self
+            .initial_rtt
+            .as_nanos()
+            .div_ceil(self.bin_duration.as_nanos()) as usize;
```
larseggert: You do this twice, maybe add a helper?
```rust
fn rtt_to_bins(&self, rtt: Duration) -> usize {
      rtt.as_nanos().div_ceil(self.bin_duration.as_nanos()) as usize
  }
```

neqo-transport/src/cc/search.rs
```diff
@@ -155,6 +165,13 @@ impl Search {
                 "SEARCH: update_bins: resetting because we skipped {passed_bins} bins (limit {})",
                 Self::W
             );
+            cc_stats.search_reset.count += 1;
+            cc_stats.search_reset.max_passed_bins = Some(
+                cc_stats
+                    .search_reset
+                    .max_passed_bins
+                    .map_or(passed_bins, |c| c.max(passed_bins)),
+            );
```
larseggert: Make this a method on `CongestionControlStats`?

neqo-transport/src/cc/search.rs
```diff
@@ -331,11 +380,18 @@ impl SlowStart for Search {
...
         now: Instant,
     ) -> Option<usize> {
         let rtt = rtt_est.latest_rtt();
 
+        // Guard on the stats fields so post-reset ACKs don't overwrite the initial samples.
+        if cc_stats.search_first_rtt.is_none() {
+            cc_stats.search_first_rtt = Some(rtt);
+        } else if cc_stats.search_second_rtt.is_none() {
+            cc_stats.search_second_rtt = Some(rtt);
+        }
```
larseggert, line 397:
```suggestion
        if cc_stats.search_first_rtt.is_none() {
            cc_stats.search_first_rtt = Some(rtt);
        } else {
            cc_stats.search_second_rtt.get_or_insert(rtt);
        }
```

omansfeld:
Thanks for the review! 

I addressed co-pilot's feedback in the push just now and added some more test coverage. I'll implement your suggestions tomorrow morning.

neqo-transport/src/cc/search.rs
```diff
@@ -275,6 +292,38 @@ impl Search {
...
+            .div_ceil(self.bin_duration.as_nanos()) as usize;
+        let initial_rtt_cong_idx = curr_idx.saturating_sub(initial_rtt_bins);
+        let empty_buffer_target = self.compute_delv(initial_rtt_cong_idx, curr_idx);
+
+        let current_rtt_bins = rtt.as_nanos().div_ceil(self.bin_duration.as_nanos()) as usize;
+        let current_rtt_cong_idx = curr_idx.saturating_sub(current_rtt_bins);
+        let full_buffer_target = self.compute_delv(current_rtt_cong_idx, curr_idx);
+
+        cc_stats.search_empty_buffer_target = Some(empty_buffer_target as usize);
+        cc_stats.search_full_buffer_target = Some(full_buffer_target as usize);
```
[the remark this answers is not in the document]
omansfeld: Copilot found a real issue here. I added a guardrail similar to what it suggested.

neqo-transport/src/cc/search.rs
omansfeld: Yes it could be. Adds some minor boilerplate in `record_target_cwnd_estimates` to handle safely but is semantically correct (There is no `initial_rtt` when uninitialized).

neqo-transport/src/cc/search.rs
omansfeld: I'll handle it like you did in #3594

neqo-transport/src/cc/search.rs
omansfeld: That's much easier, thanks!

neqo-transport/src/cc/search.rs
omansfeld: Using your suggestion from further up (directly using `max(Some())` instead of `map_or`) this is now 3 lines of code instead of 7. With that I think it doesn't warrant an extra function anymore, since this is also just used here.

larseggert (approved):

PR #3617, opened 2026-05-14
chore: upgrade http@1.4.0 and bump version

### LINKED ISSUE #3597: Coordinating the next release
We have some stuff going on that warrants some extra coordination I think.

[`v0.27.0`](https://github.com/mozilla/neqo/releases/tag/v0.27.0) (released ~2weeks ago) is supposed to be vendor-ed into Firefox as part of [this](https://bugzilla.mozilla.org/show_bug.cgi?id=2033393) larger effort upgrading the `http` dependancy. I'm not sure how much longer that will take. If we want to get anything into Firefox prior to that we have to revert #3572 and coordinate with the larger effort.

#3590 updated `nss-rs` to `0.9.0` which is the version that will be vendor-ed into Firefox. Neqo in Firefox currently doesn't depend on `nss-rs` yet  and `v0.27.0` won't either, but other things in Firefox do (i.e. other things also need to update to `v0.9.0`). @Not-Nik afaik you could update the other consumers in Firefox? It would probably make sense to do that first before Neqo cuts another release including `nss-rs`, so we don't have to land at the same time. Do you have an ETA?

The next Nightly (153) starts May 18 and I would want to get another release in before that. Which means I'd ideally want to cut another release **early-ish next week** (as soon as #3518 is merged, I will push some final test-coverage today and fix remaining nits and I think then it should be ready). Is that realistic for having other `nss-rs` consumers updated to `0.9.0` @Not-Nik? 

I can coordinate with the `http` effort mentioned above if it hasn't landed by then.

Thoughts @larseggert and @Not-Nik?

omansfeld:
After `v0.28.0` got released into Firefox we can now upgrade `http` again and bump Neqo to a new `v0.28.1` release that can be used to upgrade `http` across Gecko.

closes #3597

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -531,7 +531,7 @@ dependencies = [
 
 [[package]]
 name = "fuzz"
-version = "0.28.0"
+version = "0.28.1"
 dependencies = [
  "libfuzzer-sys",
  "neqo-common",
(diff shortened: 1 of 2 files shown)


larseggert (approved):

mxinden:
For the record, Firefox now uses `http` `1.4.0`:

https://bugzilla.mozilla.org/show_bug.cgi?id=2033393

PR #3618, opened 2026-05-15
fix: limit QPACK header list size during decoding

### LINKED ISSUE #3616: OSS-Fuzz issue 512889418
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5123306985684992 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=512889418 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

larseggert:
A maliciously crafted header block with many indexed references to large dynamic table entries could use lots of memory, as `decode_header_block` had no bound on the number of headers it would accumulate. Cap the total uncompressed header list size to `LiteralReader::MAX_LEN` using a `checked_sub` budget, consistent with RFC 9113 §6.5.2.

Fixes #3616.

### DIFF
--- a/neqo-qpack/src/header_block.rs
+++ b/neqo-qpack/src/header_block.rs
@@ -22,8 +22,8 @@ use crate::{
         NO_PREFIX,
     },
     qpack_send_buf::Encoder as _,
-    reader::{ReceiverBufferWrapper, parse_utf8},
-    table::HeaderTable,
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
A good change.

PR #3621, opened 2026-05-15
chore: rely on nss-rs for NSS initialization and test DB

larseggert:
Delegate NSS init in `fixture_init` to `nss-rs`'s own `test-fixture` crate instead of duplicating the logic. Remove `test-fixture/db` — the NSS database now comes from `nss-rs`, which exposes `TEST_FIXTURE_DB` as the default path for `neqo-server --db`. Update CI and scripts accordingly.


larseggert:
Depends on https://github.com/mozilla/nss-rs/pull/56

martinthomson (approved):
I can't see a problem here, other than the challenge of finding the db from the NSS crate (which I hope the tests will sort out)

PR #3629, opened 2026-05-19
chore(http3,qpack): remove leftover blapi feature

mxinden:
The `blapi` feature and `default = ["blapi"]` in neqo-http3 and neqo-qpack aren't needed. `blapi` is enabled in `neqo-transport` and thus transitively enabled in all of the neqo crates.

Follow-up to https://github.com/mozilla/neqo/pull/3601#discussion_r3267787344.


larseggert:
Why are there perf regressions?

larseggert:
@mxinden so the perf regressions are noise?

larseggert (approved):

PR #3639, opened 2026-05-20
fix: increase `MAX_TRACKED_RANGES` from 32 to 128

larseggert:
On high-throughput paths, the receiver's ACK tracker overflowed at 32 ranges, dropping unacknowledged ranges. The sender then saw gaps in the ACK feedback and declared those packets lost, triggering unnecessary congestion events.

Observed at 200 Mbit/s with 40 ms RTT: hundreds of spurious reordering-threshold losses traced to "Dropping unacknowledged ACK range" warnings on the client.

Increase to 128 ranges. This eliminates the drops and the resulting false losses.


martinthomson (commented):
The bot is right. Also, does this show up in the simulator? A test would be very helpful.

PR #3640, opened 2026-05-20
fix: do not suppress cwnd growth when pacing limits sends

larseggert:
RFC 9002 §7.8 states:
> A sender SHOULD NOT consider itself application limited if it would have fully utilized the
  congestion window without pacing delay.

The `first_app_limited` marker in `on_packet_sent` only advanced when `bytes_in_flight` was within 2 MTU of `cwnd`. On paths where `cwnd` exceeded the BDP (common after Cubic overshoot), `bytes_in_flight` could never reach `cwnd` in steady state, so `first_app_limited` got stuck. Every subsequent ACK round then saw all acked packet numbers above the frozen marker and classified the round as app-limited, permanently suppressing cwnd growth.

After each packet send, check whether the pacer would block the next send (`pacer.next() > now`). If so, the pacer — not the application — is the reason `bytes_in_flight` hasn't reached `cwnd`. Advance `first_app_limited` so that ACKs for this packet allow cwnd growth.

When pacing is disabled, `pacer.next()` always returns the past, so `pacing_limited` is always false and the original `app_limited()` heuristic applies unchanged. neqo-transport/src/cc/tests/cubic.rs #	modified: neqo-transport/src/cc/tests/hystart.rs #	modified: neqo-transport/src/cc/tests/new_reno.rs #	modified: neqo-transport/src/sender.rs #

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -574,7 +574,7 @@ where
         // limited if it would have fully utilized the congestion window without
         // pacing delay."  When the pacer is the reason we cannot send more
         // right now, cwnd underutilization is from pacing, not the application.
-        if !self.app_limited() || pacing_limited {
+        if pacing_limited || !self.app_limited() {
             self.first_app_limited = pkt.pn() + 1;
         }
 
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -553,7 +553,7 @@ where
         );
     }
 
-    fn on_packet_sent(&mut self, pkt: &sent::Packet, now: Instant) {
+    fn on_packet_sent(&mut self, pkt: &sent::Packet, now: Instant, pacing_limited: bool) {
         // Record the recovery time and exit any transient phase.
         if self.current.phase.transient() {
             self.current.recovery_start = Some(pkt.pn());
@@ -570,11 +570,11 @@ where
             self.slow_start.on_packet_sent(pkt.pn(), pkt.len());
         }
 
-        if !self.app_limited() {
-            // Given the current non-app-limited condition, we're fully utilizing the congestion
-            // window. Assume that all in-flight packets up to this one are NOT app-limited.
-            // However, subsequent packets might be app-limited. Set `first_app_limited` to the
-            // next packet number.
+        // RFC 9002 §7.8: "A sender SHOULD NOT consider itself application
+        // limited if it would have fully utilized the congestion window without
+        // pacing delay."  When the pacer is the reason we cannot send more
+        // right now, cwnd underutilization is from pacing, not the application.
+        if !self.app_limited() || pacing_limited {
             self.first_app_limited = pkt.pn() + 1;
         }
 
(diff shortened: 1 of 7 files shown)


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -570,11 +570,11 @@ where
...
-        if !self.app_limited() {
-            // Given the current non-app-limited condition, we're fully utilizing the congestion
-            // window. Assume that all in-flight packets up to this one are NOT app-limited.
-            // However, subsequent packets might be app-limited. Set `first_app_limited` to the
-            // next packet number.
+        // RFC 9002 §7.8: "A sender SHOULD NOT consider itself application
+        // limited if it would have fully utilized the congestion window without
+        // pacing delay."  When the pacer is the reason we cannot send more
+        // right now, cwnd underutilization is from pacing, not the application.
+        if !self.app_limited() || pacing_limited {
```
martinthomson: Can you flip the tests?  pacing_limited will be cheaper, such that the app_limited call can be skipped.  (It's possible that the optimizer is on the case already, but performance discipline is never harmful.

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -225,7 +225,6 @@ fn issue_1465() {
 
     // send out recovery packet and get it acked to get out of recovery state
     let p4 = send_next(&mut cc, now);
-    cc.on_packet_sent(&p4, now);
```
martinthomson, line 228: I'm not sure that I understand the change in this test.  Can you explain it?

neqo-transport/src/sender.rs
```diff
@@ -225,7 +225,8 @@ impl PacketSender {
     pub fn on_packet_sent(&mut self, pkt: &sent::Packet, rtt: Duration, now: Instant) {
         self.pacer
             .spend(pkt.time_sent(), rtt, self.cc.cwnd(), pkt.len());
-        self.cc.on_packet_sent(pkt, now);
+        let pacing_limited = self.pacer.next(rtt, self.cc.cwnd()) > now;
```
martinthomson: Could you have the spend call return the value you need here?  I think that you can, because this is the only call site outside of tests.  And the pacer should already have computed a lot of the things it needs to answer the question of whether it is limited.

martinthomson (commented):
The apparent performance improvements here are pretty weak.  What other evidence do you have that this is the right call?

larseggert:
> The apparent performance improvements here are pretty weak. What other evidence do you have that this is the right call?

I'll make some plots that show the issue (horizontal cwnd line) pretty convincingly. But basically: Without this, if a slow-start ends up with a cwnd overshoot that is beyond the path capacity, the connection would **never** probe upward. So the "AI" part of "AI/MD" was missing. Whether a connection ends up in that state is somewhat random, and if it doesn't, you wouldn't see a difference.

Let me see if I can get the simulator to deterministically create such a case.

neqo-transport/src/cc/tests/new_reno.rs
larseggert: The change removed a duplicate `cc.on_packet_sent(&p4, now)` call (line 228 in the old code) that was double-counting `p4` in `bytes_in_flight`.

larseggert:
Before this change:
Screenshot 2026-05-21 at 11 15 15
After:
Screenshot 2026-05-21 at 11 21 33

martinthomson (approved):
Those charts are pretty compelling.

PR #3642, opened 2026-05-20
fix: don't count FC-blocked time against rwnd auto-tuning

larseggert:
When the peer reports being blocked or has used all its credit, count at most
one RTT of the time since the last window update as sending time. The rest was
spent waiting for the update, and counting it made the measured consumption rate
look too low to grow the window, so the peer stayed blocked.


larseggert:
Wow this did not happen in manual testing!

larseggert:
Further testing revealed the issue is actually #3963

PR #3643, opened 2026-05-20
fix: improve congestion control and loss timer logging

larseggert:
Three qlog fixes:

1. Emit initial congestion state: Connections now log a `congestion_state_updated` event with `"new": "slow_start"` at startup, so visualizations can shade the slow-start phase from the first packet rather than only showing transitions.

2. Fix loss timer type attribution: `timeout()` re-evaluated the timer type by checking `earliest_loss_time`, but by then the incoming ACK (processed in the same `process()` call) had already cleared the loss candidates. The check fell through to PTO for every expiry. Fix: `note_timeout_type()` snapshots the timer type before input processing; `timeout()` reads the snapshot. Guarded by `qlog.is_enabled()` to avoid overhead when not logging.

3. Log metrics more consistently: The app-limited early-return path in `on_packets_acked` now logs `BytesInFlight` (which changes on every ACK). The non-app-limited path includes `SsThresh` when it has been set (not `usize::MAX`), so visualizations can show the slow-start threshold throughout congestion avoidance. modified: neqo-transport/src/qlog.rs #	modified: neqo-transport/src/recovery/mod.rs #

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -440,10 +440,7 @@ where
                 qlog::Metric::BytesInFlight(self.bytes_in_flight),
             ]
             .into_iter()
-            .chain(
-                (self.current.ssthresh != usize::MAX)
-                    .then_some(qlog::Metric::SsThresh(self.current.ssthresh)),
-            ),
+            .chain(self.current.ssthresh.map(qlog::Metric::SsThresh)),
             now,
         );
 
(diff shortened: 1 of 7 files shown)


mxinden:
> **Main concern:** The snapshot only covers `process_multiple` (the combined input+output path). The split API (`process_input` / `process_multiple_input` followed by `process_multiple_output`) has the same misattribution scenario but doesn't take the snapshot. Not a regression, but worth fixing for completeness — see inline comment.

I think this is a valid concern. How about moving the logic further down, e.g. into `input` to make sure all `process_*` methods hit the code path?

Also //CC @omansfeld. Would be good to get your review.

larseggert:
@mxinden any chance for a review before 0.29.0?

mxinden (approved):

PR #3644, opened 2026-05-20
feat: Add a qlog plotter

larseggert:
WIP


larseggert:
server-c89617cd379fd59d

martinthomson:
Is this producing an image, or can you use an HTML charting package?  I find the interactive capabilities of typical HTML charting packages extremely useful in this context.  Especially for high information charts.

larseggert:
It's a PDF. Good idea to make it spit out HTML, let me look at that.

larseggert:
@martinthomson now HTML
[server-57250fa48529c6f3f6.html](https://github.com/user-attachments/files/28152632/server-57250fa48529c6f3f6.html)

test/qvis.py
```diff
@@ -0,0 +1,1482 @@
...
+@lru_cache
+def _fetch(url: str) -> str:
+    try:
+        with urllib.request.urlopen(url, timeout=10) as resp:  # noqa: S310
+            return resp.read().decode()
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        raise SystemExit(f"Failed to fetch {url}: {e}") from e
+
+
+HTML_TEMPLATE = r"""<!DOCTYPE html>
```
martinthomson: Does this need to be inline?  A separate HTML file would be easier to review.

test/qvis.py
```diff
@@ -0,0 +1,1482 @@
...
+  reordering_threshold:[dk?"#ff6666":"#cc0000","Reordering Threshold Loss"],
+};
+
+// ── Formatters (4 significant digits, trailing .0 stripped) ──────────
+const s4=v=>{if(v===0)return"0";return+(Math.abs(v).toPrecision(4))*(v<0?-1:1)+""};
+const fmtB=v=>{if(v==null)return"";const a=Math.abs(v);if(a>=1e9)return s4(v/1e9)+" GB";if(a>=1e6)return s4(v/1e6)+" MB";if(a>=1e3)return s4(v/1e3)+" KB";return s4(v)+" B"};
+const fmtBps=v=>v==null?"":fmtB(v)+"/s";
+const fmtPn=v=>v==null?"":Math.round(v).toLocaleString();
+const fmtMs=v=>{if(v==null)return"";const a=Math.abs(v);if(a>=1e3)return s4(v/1e3)+" s";if(a>=1)return s4(v)+" ms";if(a>=0.001)return s4(v*1e3)+" µs";return s4(v)+" ms"};
+const scaleFmt={sb:{f:fmtB,u:" bytes"},bytes:{f:fmtB,u:" bytes"},rate:{f:fmtBps,u:" bytes/s"},pn:{f:fmtPn,u:""},rtt:{f:fmtMs,u:" ms"},gap:{f:fmtMs,u:" ms"}};
```
martinthomson: Ow, my eyes!

martinthomson (approved):
I didn't review in detail.  This doesn't need to be hardened.  The code is absolute spaghetti, but maybe we don't need to care.

The result is pretty good.  The visualization is very helpful.  

The only usability improvement I would suggest is that, when the content is not zoomed, dragging could select an area to focus on.  Scrolling to zoom means that you zoom both x and y axes at the same time, which isn't always what you want.  See the graphs [here](https://canvasjs.com/javascript-charts/chart-zoom-pan/) as an example of how dragging to select a space can be useful.

The zones that are shown at the top are useful.  I did find that the highlight wasn't consistently right though.  Not a big deal, of course.

larseggert:
@martinthomson [client-0001020304050607.html](https://github.com/user-attachments/files/28211191/client-0001020304050607.html)

martinthomson:
It might pay to tweak the help text.  The panning modifier key is "Ctrl" on platforms where the option key doesn't exist (all of them, aside from Mac).

Also, I've managed to lock up the plotting tool multiple times by clicking and dragging with various modifier keys (it took me a while to work out that ctrl-drag was how you pan).  The hang is in uPlot, which is minified, so it might not be very useful:

```
$e (client-0001020304050607.html#252)
uPlot (client-0001020304050607.html#252)
Dn (client-0001020304050607.html#252)
qn (client-0001020304050607.html#252)
qn (client-0001020304050607.html#252)
```

The code in question is `for(let l=t=s?t:Al(Dl(t,i),u);n>=l;l=Al(l+i,u))r.push(Object.is(l,-0)?0:l);` and right now, the value of `n` is 601616 and so is `l`.  I think that the problem is that `i` is 1e-11 and `Al()` seems to be a function that rounds to an integer, so it's turned into an infinite loop.

Issue #2795, opened 2025-07-16
Update WebTransport implementation from draft 4 to latest


mxinden:
We currently implement WebTransport draft 4:

https://github.com/mozilla/neqo/blob/0f4a2060047594446c81113ea9c6a09dd5b35d34/neqo-http3/src/settings.rs#L23

See corresponding settings definition in [draft 4 IANA section](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-04.html#name-http-3-settings-parameter-r).

As of writing this issue, there exists [draft 13](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/13/).

Related: https://github.com/mozilla/neqo/issues/1287

jan-ivar:
The latest is [draft 15](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-15.html#name-protocol-overview) which specifies `SETTINGS_WT_ENABLED`.

PR #3478, opened 2026-03-16
remove 2 debug_asserts

jesup:
Remove a debug assert that could cause debug builds to fail in a race condition, plus one that should never happen

PR #3646, opened 2026-05-23
draft 15 support

jesup:
This should resolve issue  #2795


jesup:

1. `trunk()`
1. **users/jesup/ietf_draft_15_support ← this PR**

---

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
```diff
@@ -60,10 +60,9 @@ pub fn default_http3_server(server_params: Http3Parameters) -> Http3Server {
 }
 
 pub fn assert_wt(headers: &[Header]) {
-    assert!(
-        headers.contains_header(":method", "CONNECT")
-            && headers.contains_header(":protocol", "webtransport")
-    );
+    let has_wt_protocol = headers.contains_header(":protocol", "webtransport-h3")
+        || headers.contains_header(":protocol", "webtransport");
```
martinthomson, line 67: Do you REALLY want to include "webtransport" in this check?  Because the capsule-based variant of this has very different properties to "webtransport-h3" and I don't think that this patch is big enough to include all of that (I'm just starting, so I could be wrong, but if I am, this patch is far too much stuff for a single pull request).

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
martinthomson, line 200: How does this differ from `create_wt_session`?  It doesn't collect the server side of the session, which makes it less useful, but other than that, what distinguishes it?.

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
```diff
@@ -321,9 +320,17 @@ impl WtTest {
         wt_session_id: StreamId,
         stream_type: StreamType,
     ) -> StreamId {
+        self.try_create_wt_stream_client(wt_session_id, stream_type)
+            .unwrap()
```
martinthomson, line 321: It might pay to use an expectation here.


```suggestion
            .expect(format!("Unable to create {stream_type:?} stream {wt_session_id}"))
```

martinthomson (commented):
This patch set is too large.  I would like to see the changes for settings discrete from the send group changes, which are massive and complicated (and I really didn't want them in the spec either).

There are also some really small, but consequential changes in here that need to be more carefully thought out.  Like the "webtransport" protocol tag, which needs to be made conditional on the draft version that is available.  For that, we need to be very careful about how we manage that, because "webtransport" in the new draft means something very different to what it meant in previous drafts.

neqo-http3/src/connection.rs
```diff
@@ -1595,6 +1602,18 @@ impl Http3Connection {
...
+        if let Http3RemoteSettingsState::Received(settings) = &self.settings_state {
+            let max_streams = match stream_type {
+                StreamType::UniDi => settings.get(HSettingType::WtInitialMaxStreamsUni),
+                StreamType::BiDi => settings.get(HSettingType::WtInitialMaxStreamsBidi),
+            };
+            if max_streams != u64::MAX && wt.borrow().local_stream_count(stream_type) >= max_streams
+            {
+                return Err(Error::StreamLimit);
+            }
+        }
```
mxinden, line 1646: Stream count is cumulative:

> Note that, because Maximum Streams is a cumulative value representing the total allowed number of streams, including previously closed streams, endpoints repeatedly send new [WT_MAX_STREAMS](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#WT_MAX_STREAMS) capsules with increasing Maximum Streams values as streams are opened.

https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-13.html#section-5.6.2

I don't think looking at the number of active streams here is correct. Here is a test that fails with the current implementation.

https://github.com/mxinden-bot/neqo/pull/3/changes/571a82fe4a66b3a6942b3cc3868f9c1be94ed476

jesup: I expanded the fix to include bidi, and added tests for both.

neqo-http3/src/settings.rs
```diff
@@ -15,12 +15,19 @@ type SettingsType = u64;
...
 const SETTINGS_QPACK_BLOCKED_STREAMS: SettingsType = 0x7;
-const SETTINGS_ENABLE_WEB_TRANSPORT: SettingsType = 0x2b60_3742;
+// draft-ietf-webtrans-http3-07#section-8.2
+const SETTINGS_ENABLE_WEB_TRANSPORT_DRAFT07: SettingsType = 0xc671_706a;
+// draft-ietf-webtrans-http3-15#section-9.2
+const SETTINGS_ENABLE_WEB_TRANSPORT_DRAFT15: SettingsType = 0x2c7c_f000;
+// draft-ietf-webtrans-http3-15 / draft-ietf-webtrans-http2-11 shared settings
+const SETTINGS_WT_INITIAL_MAX_DATA: SettingsType = 0x2b61;
+const SETTINGS_WT_INITIAL_MAX_STREAMS_UNI: SettingsType = 0x2b64;
+const SETTINGS_WT_INITIAL_MAX_STREAMS_BIDI: SettingsType = 0x2b65;
```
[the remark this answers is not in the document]
jesup: I verified against draft-ietf-webtrans-http3-15 §9.2 (the IANA registration table). The spec explicitly assigns:
  - SETTINGS_WT_INITIAL_MAX_STREAMS_UNI = 0x2b64
  - SETTINGS_WT_INITIAL_MAX_STREAMS_BIDI = 0x2b65
  - SETTINGS_WT_INITIAL_MAX_DATA = 0x2b61

  The current code already matches this exactly. The WebTransport draft does not follow the QUIC transport-parameter convention (bidi-before-uni) that motivated the suggestion — it numbers UNI first. Applying the proposed swap would introduce the very interop bug the comment is trying to prevent. Left unchanged.

neqo-http3/src/settings.rs
```diff
@@ -307,13 +379,19 @@ impl ZeroRttChecker for HttpZeroRttChecker {
...
                 }
                 let value = setting.value == 1;
                 self.settings.get_webtransport() || !value
             }
+            // Draft-07 in a 0-RTT token: always accept, we handle both versions.
+            HSettingType::EnableWebTransportDraft07
+            | HSettingType::WtInitialMaxData
+            | HSettingType::WtInitialMaxStreamsUni
+            | HSettingType::WtInitialMaxStreamsBidi
+            | HSettingType::MaxHeaderListSize => true,
```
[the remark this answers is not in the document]
jesup: Replaced the unconditional => true for the three WT flow-control settings with current == 0 || current >= setting.value, matching the existing QPACK/BlockedStreams pattern. This rejects 0-RTT when the server has lowered a per-session limit since issuing the token (preventing replay overflow). The current == 0 guard preserves the "0 = not advertised = unlimited" semantics so loosening a limit doesn't needlessly reject 0-RTT.

neqo-http3/src/features/extended_connect/tests/webtransport/mod.rs
jesup: replaced with negotiate_wt_session(), which picks up a few extra useful bits

PR #3647, opened 2026-05-26
fix: remove duplicate word in two source comments

mvanhorn:
## Summary

Two small comment-only fixes:

- `neqo-transport/src/cid.rs:473` — "until until after" → "until after"
- `neqo-http3/src/push_controller.rs:342` — "does not not know" → "does not know"

No behavior changes.


martinthomson (approved):

mvanhorn:
Thanks @martinthomson for taking the duplicate-word comment fixes in cid.rs.

mvanhorn:
Cheers @martinthomson, thanks for taking the comment cleanup. Small fixes but the duplicated words were real.

PR #3648, opened 2026-05-26
fix: Allow >1 MSS of `cwnd` growth

### LINKED ISSUE #3113: cc/cubic: consider >2 MSS increase on ACK
Currently `on_packets_acked`, we at most increase by 2 MSS per ACK:

https://github.com/mozilla/neqo/blob/207d309e941d3ae9384c191c13463dc81fca4ce9/neqo-transport/src/cc/classic_cc.rs#L243-L257

I am not sure this artificial limit is needed, especially since outgoing packets will be paced.

This code was added in https://github.com/mozilla/neqo/pull/1074.

Note that the pacer was already in place (https://github.com/mozilla/neqo/pull/618) at this time, i.e. the 2 MSS limit was likely not due to missing a pacer.

I don't think the 2 MSS limit is specified anywhere. Please correct me if I am wrong. RFC 9002 [suggests an increase relative to the amount of bytes acked](https://datatracker.ietf.org/doc/html/rfc9002#section-b.5). If I understand [RFC 3465](https://datatracker.ietf.org/doc/html/rfc3465?utm_source=chatgpt.com#section-2.1) correctly, the suggested behavior is as well allowed.

//CC @ddragana and @mwelzl as discussed earlier

larseggert:
Fixes #3113

### DIFF
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -7,7 +7,7 @@
 // Congestion control
 
 use std::{
-    cmp::max,
+    cmp::{max, min},
     fmt::{Debug, Display},
     time::{Duration, Instant},
 };
@@ -418,7 +418,7 @@ where
             // Cap prior credit to one increment (in case bytes_for_increase dropped between
             // calls), then apply all increments earned, allowing >1 MSS of growth when a large
             // burst is acknowledged (RFC 9002 has no 2-MSS cap; RFC 3465 §2.3 is TCP-only).
-            let acked = self.current.acked_bytes.min(bytes_for_increase) + new_acked;
+            let acked = min(self.current.acked_bytes, bytes_for_increase) + new_acked;
             let n = acked / bytes_for_increase;
             self.current.acked_bytes = acked % bytes_for_increase;
             self.current.congestion_window += n * self.max_datagram_size();
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -7,7 +7,7 @@
 // Congestion control
 
 use std::{
-    cmp::{max, min},
+    cmp::max,
     fmt::{Debug, Display},
     time::{Duration, Instant},
 };
@@ -405,8 +405,8 @@ where
             .ssthresh
             .is_some_and(|s| self.current.congestion_window >= s)
         {
-            // The following function return the amount acked bytes a controller needs
-            // to collect to be allowed to increase its cwnd by MAX_DATAGRAM_SIZE.
+            // The following function returns the amount of acked bytes a controller
+            // needs to collect to be allowed to increase its cwnd by one datagram.
             let bytes_for_increase = self.congestion_control.bytes_for_cwnd_increase(
                 self.current.congestion_window,
                 new_acked,
@@ -415,21 +415,17 @@ where
                 now,
             );
             debug_assert!(bytes_for_increase > 0);
-            // If enough credit has been accumulated already, apply them gradually.
-            // If we have sudden increase in allowed rate we actually increase cwnd gently.
-            if self.current.acked_bytes >= bytes_for_increase {
-                self.current.acked_bytes = 0;
-                self.current.congestion_window += self.max_datagram_size();
-            }
-            self.current.acked_bytes += new_acked;
-            if self.current.acked_bytes >= bytes_for_increase {
-                self.current.acked_bytes -= bytes_for_increase;
-                self.current.congestion_window += self.max_datagram_size(); // or is this the current MTU?
-            }
-            // The number of bytes we require can go down over time with Cubic.
-            // That might result in an excessive rate of increase, so limit the number of unused
-            // acknowledged bytes after increasing the congestion window twice.
-            self.current.acked_bytes = min(bytes_for_increase, self.current.acked_bytes);
+            // If prior credit crossed the threshold (e.g., bytes_for_increase dropped
+            // between calls), consume one increment and discard the excess so it does
+            // not compound with the credit from the current ACK.
+            let acked = self.current.acked_bytes;
+            let carry = acked >= bytes_for_increase;
+            let acked = (if carry { 0 } else { acked }) + new_acked;
+            // Apply all increments earned by the current ACK, allowing >1 MSS of growth when a
+            // large burst is acknowledged (RFC 9002 has no 2-MSS cap; RFC 3465 §2.3 is TCP-only).
+            let n = usize::from(carry) + acked / bytes_for_increase;
+            self.current.acked_bytes = acked % bytes_for_increase;
+            self.current.congestion_window += n * self.max_datagram_size();
         }
 
         cc_stats.cwnd = Some(self.current.congestion_window);
--- a/neqo-transport/src/cc/classic_cc.rs
+++ b/neqo-transport/src/cc/classic_cc.rs
@@ -415,15 +415,11 @@ where
                 now,
             );
             debug_assert!(bytes_for_increase > 0);
-            // If prior credit crossed the threshold (e.g., bytes_for_increase dropped
-            // between calls), consume one increment and discard the excess so it does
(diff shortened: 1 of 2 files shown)


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -394,8 +394,8 @@ where
 
         // Congestion avoidance, above the slow start threshold.
         if self.current.congestion_window >= self.current.ssthresh {
-            // The following function return the amount acked bytes a controller needs
-            // to collect to be allowed to increase its cwnd by MAX_DATAGRAM_SIZE.
+            // The following function returns the amount of acked bytes a controller
+            // needs to collect to be allowed to increase its cwnd by one datagram.
             let bytes_for_increase = self.congestion_control.bytes_for_cwnd_increase(
```
mxinden, line 410: As an aside, and as well discussed with @omansfeld in the past, I think `self.congestion_control.bytes_for_cwnd_increase` should return the `cwnd` increase itself, not the number of bytes needed for a 1 MSS increase. I think that would significantly simplify the logic further below.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -404,21 +404,17 @@ where
...
+            // between calls), consume one increment and discard the excess so it does
+            // not compound with the credit from the current ACK.
+            let acked = self.current.acked_bytes;
+            let carry = acked >= bytes_for_increase;
+            let acked = (if carry { 0 } else { acked }) + new_acked;
+            // Apply all increments earned by the current ACK, allowing >1 MSS of growth when a
+            // large burst is acknowledged (RFC 9002 has no 2-MSS cap; RFC 3465 §2.3 is TCP-only).
+            let n = usize::from(carry) + acked / bytes_for_increase;
+            self.current.acked_bytes = acked % bytes_for_increase;
+            self.current.congestion_window += n * self.max_datagram_size();
```
mxinden: Wouldn't this be easier?

```suggestion
let acked = self.current.acked_bytes.min(bytes_for_increase) + new_acked;
let n = acked / bytes_for_increase;
self.current.acked_bytes = acked % bytes_for_increase;
self.current.congestion_window += n * self.max_datagram_size();
```

mxinden: Wouldn't this be easier?

```suggestion
let acked = self.current.acked_bytes.min(bytes_for_increase) + new_acked;
let n = acked / bytes_for_increase;
self.current.acked_bytes = acked % bytes_for_increase;
self.current.congestion_window += n * self.max_datagram_size();
```

mxinden (commented):
Very much in favor. Makes reasoning about this code path a lot easier.

neqo-transport/src/cc/classic_cc.rs
larseggert: https://github.com/mozilla/neqo/issues/3655

larseggert:
Thanks for the suggestion. I'm still trying to find any performance benefit from this change.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -404,21 +404,13 @@ where
...
-                self.current.congestion_window += self.max_datagram_size(); // or is this the current MTU?
-            }
-            // The number of bytes we require can go down over time with Cubic.
-            // That might result in an excessive rate of increase, so limit the number of unused
-            // acknowledged bytes after increasing the congestion window twice.
-            self.current.acked_bytes = min(bytes_for_increase, self.current.acked_bytes);
+            // Cap prior credit to one increment (in case bytes_for_increase dropped between
+            // calls), then apply all increments earned, allowing >1 MSS of growth when a large
+            // burst is acknowledged (RFC 9002 has no 2-MSS cap; RFC 3465 §2.3 is TCP-only).
+            let acked = self.current.acked_bytes.min(bytes_for_increase) + new_acked;
```
martinthomson: I find this easier to reason about:

```suggestion
            let acked = min(self.current.acked_bytes, bytes_for_increase) + new_acked;
```

which makes me ask why you would allow that and not:

```suggestion
            let acked = min(self.current.acked_bytes + new_acked, bytes_for_increase);
```

I think that this is because you don't want to lose any new values in excess of `bytes_for_increase`, but it's a bit opaque.  In that spirit, I don't know why this isn't the simpler:


```suggestion
            let acked = self.current.acked_bytes + new_acked;
```

That is, if the number of bytes needed for an MSS worth of increases reduces (as it always does), then why lose those bytes we already have had acknowledged?

martinthomson (commented):
I have a simpler change than even Max.  It will increase the CWND more, but not so much more that I think it's problematic.

neqo-transport/src/cc/classic_cc.rs
larseggert: I think with your change, there is a risk of a big increase, but I need to think about it more.

martinthomson: Let me make two points that might help clarify the suggestion:

We perform this update whenever there are new acknowledged bytes.  If we get more acknowledgments, those increments will be smaller and `new_acked` will be a smaller proportion of the acknowledged bytes that we act on in this code.  So, less frequent acknowledgments will result in less loss from the `min()` call here.  That seems to me to be backwards.  We should be increasing CWND the same, no matter how frequent or infrequent acknowledgments are.

We can still cap the number of MSS increases we execute at once.  However, I don't think we need to.  The value of `bytes_for_increase` should have a floor, which would bound the increase here.  And the entire point of cubic is to increase at a rate higher than 1MSS/RTT once you get past the previous estimate of the path throughput.  Capping increases like this only handicaps the CC.

larseggert:
@martinthomson I'd prefer to land this now and keep discussing your proposal in the linked issue - OK?

martinthomson:
WFM

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -256,3 +256,35 @@ fn issue_1465() {
...
+    // For NewReno, bytes_for_cwnd_increase returns cwnd, so one MSS of cwnd
+    // growth requires acknowledging a full cwnd worth of bytes.
+    let cwnd0 = cc.cwnd();
+    cc.set_ssthresh(cwnd0);
+
+    // Send 3 * cwnd / mtu packets. This keeps BIF well above the app-limited
+    // threshold so that is_app_limited is false when all packets are acked.
+    let n = 3 * (cwnd0 / mtu);
+    let mut pkts = Vec::with_capacity(n);
+    for pn in 0..n as u64 {
```
martinthomson:
```suggestion
    let n = 3 * u64::try_from(cwnd0 / mtu).unwrap();
    let mut pkts = Vec::with_capacity(n);
    for pn in 0..n {
```

It's a test, but we should still avoid casting.  Bad habit to get into.

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -256,3 +256,35 @@ fn issue_1465() {
...
+        cc.on_packet_sent(&p, now, false);
+        pkts.push(p);
+    }
+
+    // ACK all packets in one call: new_acked = 3 * cwnd.
+    cc.on_packets_acked(&pkts, &RttEstimate::new(RTT), now + RTT, &mut cc_stats);
+
+    // new_acked / bytes_for_increase = 3*cwnd0 / cwnd0 = 3 increments.
+    assert_eq!(cc.cwnd(), cwnd0 + 3 * mtu);
+    assert_eq!(cc.acked_bytes(), 0);
```
martinthomson: I think that I'd like to see if the carry is working as well.  Could you add one to n and check that you have a spare MTU left over?

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -415,21 +415,13 @@ where
...
```
martinthomson:
```suggestion
            let acked = min(self.current.acked_bytes, bytes_for_increase) + new_acked;
```

Am I the only one that finds the chained variant of this function harder to parse?

martinthomson (approved):

PR #3652, opened 2026-05-27
fix: Patches for Redox targets

willnode:
This add `target_os = "redox"` to the list of unsupported `mtu` implementation, so it can compile for [the platform](https://doc.rust-lang.org/rustc/platform-support/redox.html).

This is part of patches that's used to port Firefox for Redox OS, other patches such as [quinn](`https://github.com/quinn-rs/quinn/pull/2659`) and a number of C libraries (e.g. nspr) is needed to actually compile.

I'm looking for `neqo-crypto` which was patched for Firefox 150 but it seems disappear or moved recently, so it's not included here.

### DIFF
--- a/mtu/src/lib.rs
+++ b/mtu/src/lib.rs
@@ -110,7 +110,7 @@ const fn aligned_by(size: usize, align: usize) -> usize {
 // Platforms currently not supported.
 //
 // See <https://github.com/mozilla/mtu/issues/82>.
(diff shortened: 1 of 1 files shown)


larseggert:
Thanks! `neqo-crypto` got replaced by https://github.com/mozilla/nss-rs.

larseggert (approved):

PR #3653, opened 2026-05-27
feat: Add OS timer jitter/skew to simulator

larseggert:
Because it matters for high-rate simulations.

For discussion: 50usec seems much lower than our 1ms granularity, but it's the only number backed up by a reputable source.

### DIFF
--- a/test-fixture/src/sim/mod.rs
+++ b/test-fixture/src/sim/mod.rs
@@ -94,10 +94,9 @@ pub trait Node: Debug {
     }
     /// Print out a summary of the state of the node.
     fn print_summary(&self, _test_name: &str) {}
-    /// Whether this node models an application process subject to OS timer jitter.
-    /// Network nodes return `false` (the default); connection nodes return `true`.
-    fn has_timer_jitter(&self) -> bool {
-        false
+    /// Maximum additional delay the simulator applies to timer callbacks for this node.
+    fn timer_jitter_bound(&self) -> Duration {
+        Duration::ZERO
     }
 }
 
@@ -154,6 +153,11 @@ pub enum GoalStatus {
     Done,
 }
 
+/// Linux default timer slack: the maximum delay the kernel may apply to coalesce
+/// `epoll_wait`, `poll`, and `nanosleep` wakeups for power saving.
+/// <https://man7.org/linux/man-pages/man2/PR_SET_TIMERSLACK.2const.html>
+pub(crate) const LINUX_TIMER_SLACK: Duration = Duration::from_micros(50);
+
 pub struct Simulator {
     name: String,
     nodes: Vec<NodeHolder>,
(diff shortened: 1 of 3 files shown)


test-fixture/src/sim/mod.rs
```diff
@@ -202,6 +207,22 @@ impl Simulator {
...
 
+    /// Models Linux OS timer delivery jitter: up to 50µs, right-skewed.
+    ///
+    /// Uses min(U, U) where U ~ Uniform(0, `50_000` ns). Taking the minimum of
+    /// two independent uniform samples produces a right-skewed distribution.
+    /// The 50µs bound matches the Linux default timer slack (`PR_SET_TIMERSLACK`),
+    /// which is the maximum delay the kernel may apply to coalesce timer wakeups
+    /// for `epoll_wait`, `poll`, `nanosleep`, and similar calls:
+    /// <https://man7.org/linux/man-pages/man2/PR_SET_TIMERSLACK.2const.html>
+    fn os_timer_jitter(rng: &Rng, delay: Duration) -> Duration {
```
martinthomson: Isn't this just added latency, not jitter?

test-fixture/src/sim/mod.rs
```diff
@@ -94,6 +94,11 @@ pub trait Node: Debug {
     }
     /// Print out a summary of the state of the node.
     fn print_summary(&self, _test_name: &str) {}
+    /// Whether this node models an application process subject to OS timer jitter.
+    /// Network nodes return `false` (the default); connection nodes return `true`.
+    fn has_timer_jitter(&self) -> bool {
+        false
+    }
```
martinthomson: Rather than have this be a boolean, can this not be a time?  Then we could have varying jitter.

martinthomson (commented):
50 microseconds doesn't seem like a lot.  What evidence do you have that this is something that alters the outcome of tests?

If you change this to either

```rust
fn timer_jitter(&self, delay: Duration) -> Duration {
  delay
}
```

or

```rust
fn timer_delay(&self, _rng: &Rng) -> Duration {
  Duration::new(0, 0)
}
```

or maybe even

```rust
/* const ? */ fn max_timer_delay(&self) -> Duration {
  Duration::new(0, 0)
}
```

You get more mileage.

I think that this becomes more flexible without a lot extra work (a common `linux_timer_delay()` function that uses what you have implemented here, parameterized appropriately means no extra code, even).

test-fixture/src/sim/mod.rs
larseggert: It adds random delay of latency, so it jitters the series?

larseggert:
> 50 microseconds doesn't seem like a lot. What evidence do you have that this is something that alters the outcome of tests?

As I said; for discussion. But on high-bandwidth paths, the pacer delays are <50usec, so this has am effect.

PR #3654, opened 2026-05-27
feat: Switch from RED to CoDel

larseggert:
Our RED code was extremely aggressive in marking, and RED is not really deployed much. Let's try CoDel.


test-fixture/src/sim/taildrop.rs
```diff
@@ -17,11 +17,109 @@ use neqo_common::{Datagram, Dscp, Ecn, Tos, qinfo, qtrace};
...
+            // Enter dropping state.
+            self.dropping = true;
+            // Fast restart: if we re-enter quickly, ramp up from where we left off
+            // rather than starting from count=1 (RFC 8289 §4: 16×INTERVAL window).
+            let delta = self.count.saturating_sub(self.lastcount);
+            let recently_dropping = self
+                .drop_next
+                .is_some_and(|dn| now.saturating_duration_since(dn) < CODEL_INTERVAL * 16);
+            self.count = if delta > 1 && recently_dropping {
+                delta
```
martinthomson: Try this:


```suggestion
            let recently_dropping = self
                .drop_next
                .is_some_and(|dn| now.saturating_duration_since(dn) < CODEL_INTERVAL * 16);
            self.count = if recently_dropping {
                max(1, self.count.saturating_sub(self.lastcount))
```

Though I don't quite follow the logic here.  This counts the number of drops we had while in the dropping phase, but subtracting the lastcount value will ensure that the value is effectively reset every time we re-enter.  So we don't really "ramp up from where we left off".  We enter dropping, then drop some number of times, then exit, then re-enter with the number of drops retained (at least 1, but later re-entries don't continue indefinitely, they only pick up any count from the most recent time spent in the dropping state.

test-fixture/src/sim/taildrop.rs
```diff
@@ -17,11 +17,109 @@ use neqo_common::{Datagram, Dscp, Ecn, Tos, qinfo, qtrace};
...
+                dgram.len()
+            );
+            None
+        }
+    }
+
+    /// next = now + INTERVAL / sqrt(count)
+    fn control_law(&self, base: Instant) -> Instant {
+        base + Duration::from_secs_f64(
+            CODEL_INTERVAL.as_secs_f64() / f64::from(self.count.max(1)).sqrt(),
```
martinthomson: Are you familiar with 0x5F3759DF?

Seems like a golden opportunity to put some computer science in: https://en.wikipedia.org/wiki/Fast_inverse_square_root

test-fixture/src/sim/taildrop.rs
```diff
@@ -17,11 +17,109 @@ use neqo_common::{Datagram, Dscp, Ecn, Tos, qinfo, qtrace};
...
+            self.drop_next = Some(self.control_law(now));
+            return true;
+        }
+
+        false
+    }
+
+    /// Apply the `CoDel` congestion signal: CE-mark ECT(0) packets, drop others.
+    /// Returns `None` if the packet was dropped.
+    fn signal(dgram: &Datagram) -> Option<Datagram> {
```
martinthomson: This isn't CoDel-specific, but it's more generic ECT(0)->CE marking.  (You can tell, because this function doesn't take `self` in any form.)

test-fixture/src/sim/taildrop.rs
```diff
@@ -17,11 +17,109 @@ use neqo_common::{Datagram, Dscp, Ecn, Tos, qinfo, qtrace};
...
+    /// offset by INTERVAL. None when sojourn is below target or queue is empty.
+    first_above_time: Option<Instant>,
+    /// Whether we are currently in the "dropping" (signalling) state.
+    dropping: bool,
+    /// How many marks/drops have occurred in the current dropping interval.
+    count: u32,
+    /// `count` at entry to the last dropping period; used for fast restart.
+    lastcount: u32,
+    /// The time at which the next mark/drop is due (only valid when dropping).
+    drop_next: Option<Instant>,
```
martinthomson: Given that we are favoring ECN, I would prefer this be called `next_mark_time`.  Obviously, if you don't get ECT(0), this will be dropped, but that's not what we expect.

martinthomson: Given that we are favoring ECN, I would prefer this be called `next_mark_time`.  Obviously, if you don't get ECT(0), this will be dropped, but that's not what we expect.

test-fixture/src/sim/taildrop.rs
```diff
@@ -17,11 +17,109 @@ use neqo_common::{Datagram, Dscp, Ecn, Tos, qinfo, qtrace};
...
+/// `CoDel` (RFC 8289) algorithm state.
+#[derive(Clone, Default)]
+struct CodelState {
+    /// When sojourn time first exceeded TARGET in the current busy period,
+    /// offset by INTERVAL. None when sojourn is below target or queue is empty.
+    first_above_time: Option<Instant>,
+    /// Whether we are currently in the "dropping" (signalling) state.
+    dropping: bool,
+    /// How many marks/drops have occurred in the current dropping interval.
+    count: u32,
```
martinthomson: Shouldn't these be usize?  Unless you plan to do the crazy stuff I suggest.

martinthomson (commented):
Is there any value in having both modes available?

You could provide a CoDel or RED marking engine in place of the ECN flag.

Looking at how this is structured, it's not a big change to accept a marking engine as input to the TailDrop constructor.

test-fixture/src/sim/taildrop.rs
```diff
@@ -51,24 +207,8 @@ struct Stats {
...
-            capacity: 0,
-            start_time: None,
-            last_update: None,
-            cumulative_usage: 0,
-        }
-    }
-
     fn update_maxq(&mut self, used: usize) {
-        self.maxq = max(self.maxq, used);
+        self.maxq = self.maxq.max(used);
```
martinthomson: Gratuitous.  I'm not a fan of `x.max(y)` for readability.  At best, it saves a `use` statement, which I think is not a good economy (min/max should be in the preface in my view).

test-fixture/src/sim/taildrop.rs
```diff
@@ -16,13 +16,169 @@ use std::{
...
+    }
+}
+
+/// Congestion-signalling mode for a [`TailDrop`] queue.
+///
+/// The inner state types are opaque; use [`Aqm::codel()`] and [`Aqm::red()`] to create instances.
 #[derive(Clone)]
+pub enum Aqm {
+    /// No AQM; packets are dropped only on buffer overflow (pure tail-drop).
+    None,
```
martinthomson:
```suggestion
    #[default]
    None,
```

test-fixture/src/sim/taildrop.rs
```diff
@@ -16,13 +16,169 @@ use std::{
...
+            .borrow_mut()
+            .random_from(0..(1 << 20));
+        r < p
+    }
+}
+
+/// Congestion-signalling mode for a [`TailDrop`] queue.
+///
+/// The inner state types are opaque; use [`Aqm::codel()`] and [`Aqm::red()`] to create instances.
 #[derive(Clone)]
```
martinthomson:
```suggestion
#[derive(Clone, Default)]
```

test-fixture/src/sim/taildrop.rs
```diff
@@ -16,13 +16,169 @@ use std::{
...
 use neqo_transport::Output;
 
-use super::Node;
-use crate::sim::Rng;
+use super::{Node, Rng};
 
 /// One second in nanoseconds.
 const ONE_SECOND_NS: u128 = 1_000_000_000;
 
+const CODEL_TARGET: Duration = Duration::from_millis(5);
```
martinthomson: All this looks like an `aqm.rs` to me.

test-fixture/src/sim/taildrop.rs
```diff
@@ -242,69 +367,62 @@ impl TailDrop {
...
+        self.used -= self.size(&pkt);
+
+        let should_signal = match &mut self.aqm {
+            Aqm::CoDel(state) => state.update(
+                now.saturating_duration_since(enqueue_time),
+                self.used == 0,
+                now,
+            ),
+            Aqm::Red(state) => {
+                Ecn::from(pkt.tos()).is_ect() && state.should_mark(self.used, self.capacity)
```
martinthomson: It seems like you could make `Aqm::mark(d: Datagram, ...) -> Option<Datagram>` and encapsulate all this stuff, including the call to `mark_ce()`.

test-fixture/src/sim/taildrop.rs
```diff
@@ -16,13 +16,169 @@ use std::{
...
+    /// `count` at entry to the last dropping period; used for fast restart.
+    lastcount: u32,
+    /// The time at which the next mark/drop is due (only valid when dropping).
+    next_mark_time: Option<Instant>,
+}
+
+impl CodelState {
+    /// Update the `CoDel` state machine for the packet just dequeued.
+    /// Returns true if congestion should be signalled for this packet.
+    fn update(&mut self, sojourn: Duration, queue_empty: bool, now: Instant) -> bool {
```
martinthomson: I just realized that the soujorn in our simulator is strictly correlated with the amount of capacity that was used at the time that the datagram was enqueued.  Is that a fact you can use?

test-fixture/src/sim/aqm.rs
```diff
@@ -0,0 +1,210 @@
...
+        } else if over_interval {
+            // Enter dropping state.
+            self.dropping = true;
+            // Fast restart (RFC 8289 §4): if we re-enter dropping within 16×INTERVAL
+            // of the previous interval, start count at `count − lastcount` (the increment
+            // from the last dropping period) rather than 1.  This means the marking rate
+            // picks up where it left off rather than restarting from scratch each time.
+            let recently_dropping = self
+                .next_mark_time
+                .is_some_and(|dn| now.saturating_duration_since(dn) < CODEL_INTERVAL * 16);
```
martinthomson: Should the 16 be a constant as well.  `CODEL_FAST_RESTART_INTERVALS` or something like that.

test-fixture/src/sim/aqm.rs
```diff
@@ -0,0 +1,210 @@
...
+                // ok_to_drop became false (RFC 8289): leave dropping state.
+                self.dropping = false;
+            } else if let Some(dn) = self.next_mark_time.filter(|&dn| now >= dn) {
+                // Time for another mark/drop in the current dropping interval.
+                // RFC 8289: next drop is relative to the previous next_mark_time, not now.
+                self.count += 1;
+                self.next_mark_time = Some(self.control_law(dn));
+                return true;
+            }
+        } else if over_interval {
```
martinthomson: This might be easier to follow if you structure it like this:


```suggestion
            }
            return false;
        }
        if !over_interval {
            return false;
        }
        // ... enter dropping state
```

martinthomson (approved):
This looks pretty good.  It even looks like we get a tiny gain in the testing.

PR #3656, opened 2026-05-28
feat: Encode zoom parameter in qvis URL

larseggert:
So that one can share a view by sharing the URL.

### DIFF
--- a/test/qvis.js
+++ b/test/qvis.js
@@ -1181,7 +1181,6 @@ decompress("__DATA_B64GZ__").then((D) => {
                 syncing = true;
                 try {
                   for (const ch of charts) {
-                    ch.setSelect({ left: 0, width: 0, top: 0, height: 0 }, false);
                     ch.batch(() => {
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
This attaches hooks to every chart.  Then saves parameters for every chart.  But my understanding was that the zoom level was linked across all the charts.  Wouldn't it be sufficient to track just the first?

larseggert:
> But my understanding was that the zoom level was linked across all the charts. Wouldn't it be sufficient to track just the first?

It's possible to zoom/pan each plot individually, but holding (IIRC) the shift key.

martinthomson:
Huh, that doesn't sound like a very useful feature.  Still, no sense in blocking this if you find it helpful.

(I do wonder whether this isn't better hosted on a personal project, so that you don't have to bother with us reviewing the tool.  This is basically fine as vibe-coded slop.)

martinthomson (approved):

PR #3659, opened 2026-05-29
feat: More benches

larseggert:
Add a few more benches that might help with optimizing some data structures. Or at least safeguard from regressions.


larseggert:
@mxinden could you give this a look when you have the chance? Would like to use these to validate some other improvements the profiles suggest.

mxinden (approved):

PR #3660, opened 2026-06-01
chore: Use `Option` for CC integers with magic sentinel values

larseggert:
Be a bit more idiomatic.


mxinden (approved):
Generally in favor. Not a strong opinion.

PR #3661, opened 2026-06-01
chore: Make `blocked_at` an `Option`

larseggert:
Instead of using 0 as a sentinel and encoding values as "one greater than".

### DIFF
--- a/neqo-transport/src/fc.rs
+++ b/neqo-transport/src/fc.rs
@@ -73,13 +73,10 @@ where
     limit: u64,
     /// How much of that limit we've used.
     used: u64,
-    /// The point at which blocking occurred.  This is updated each time
-    /// the sender decides that it is blocked.  It only ever changes
-    /// when blocking occurs.  This ensures that blocking at any given limit
-    /// is only reported once.
-    /// Note: All values are one greater than the corresponding `limit` to
-    /// allow distinguishing between blocking at a limit of 0 and no blocking.
-    blocked_at: u64,
+    /// The limit at which blocking was last reported, or `None` if never blocked.
+    /// Updated each time the sender decides it is blocked, ensuring that blocking
+    /// at any given limit is only reported once.
+    blocked_at: Option<u64>,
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Thanks

neqo-transport/src/fc.rs
```diff
@@ -130,18 +126,22 @@ where
     /// Mark flow control as blocked.
     /// This only does something if the current limit exceeds the last reported blocking limit.
     pub const fn blocked(&mut self) {
-        if self.limit >= self.blocked_at {
-            self.blocked_at = self.limit + 1;
-            self.blocked_frame = true;
+        if let Some(l) = self.blocked_at
+            && self.limit <= l
```
martinthomson, line 130:
```suggestion
        if let Some(block) = self.blocked_at
            && self.limit <= block
```

neqo-transport/src/fc.rs
```diff
@@ -153,7 +153,9 @@ where
     /// Only send again if value of `self.blocked_at` hasn't increased since sending.
     /// That would imply that the limit has since increased.
     pub const fn frame_lost(&mut self, limit: u64) {
-        if self.blocked_at == limit + 1 {
+        if let Some(l) = self.blocked_at
+            && l == limit
```
martinthomson, line 157:
```suggestion
        if let Some(block) = self.blocked_at
            && block == limit
```

martinthomson (commented):
Try to use better variable names than `l`, which tends to look like a `1`.

neqo-transport/src/fc.rs
larseggert: You need a better font :-) ACK, will do.

larseggert: #3664

neqo-transport/src/fc.rs
larseggert: #3664

PR #3662, opened 2026-06-01
feat(transport): close connection after N consecutive PTOs

mxinden:
## Summary

Adds an **opt-in** `ConnectionParameters::max_pto(Some(count))`. When set, the connection is closed locally with a new, distinct reason `Error::TooManyPtos` once `count` consecutive PTOs have fired without any acknowledgement, i.e. the path has black-holed.

This is a safeguard for when the path breaks: if packets stop getting through entirely, the connection is closed promptly and with a reason the application can act on, instead of waiting out `max_idle_timeout` and reporting a generic timeout.

## Motivation: the `facebook-not-loading` incident

Recent incident (the same one behind #3573): with SCONE padding enabled, a client completed the QUIC  handshake to Facebook, exchanged some application data, and then an on-path antivirus middlebox (BitDefender) silently dropped the altered post-handshake server frames. From neqo's point of view: the client had requests outstanding, its packets stopped being acked, and nothing came back.

Per [RFC 9002 §6.2.1](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.1) the PTO backs off without bound, and per [RFC 9000 §10.1](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.1) the idle timeout is the designated backstop (and is held `>= 3*PTO` precisely so probes can be lost before it fires). The consequence for this incident: the only thing that ends a post-handshake black hole today is the idle timeout, 30 s, and HTTP/3 has no HTTP/2 fallback after the handshake. The user sees a spinner, and a reload just pins a fresh request onto the same dead connection for another 30 s.

The sooner the connection is declared broken, the better on two fronts. First, fewer requests are sent into the abyss: the application stops piling new streams onto a dead connection. Second, Firefox can surface an actionable error page sooner, for example one suggesting the user disable their antivirus, instead of leaving them on a spinner for 30 s per attempt.

## Worked example

With exponential backoff (`pto_period = base_PTO * 2^count`, default `fast_pto` factor = 1), the n-th consecutive PTO fires at `base_PTO * (2^n - 1)`, so `max_pto = 7` fires at `127 * base_PTO`.

Facebook's edge is close; take **RTT ~= 50 ms**, so `base_PTO = smoothed_rtt + 4*rttvar + max_ack_delay ~= 50 + ~25 + 25 = ~100 ms`.

- Dead today (idle timeout): **30 s**
- Dead with `max_pto = 7`: `127 * 100 ms` ~= **12.7 s**
- **~17 s earlier**, less than half the time.

How it scales (the virtue of a *count* over a fixed timer: it tracks RTT, so it fails fast on a quick path and waits longer on a genuinely slow one):

| base_PTO (~ RTT) | dead at 7 PTOs (127x) | vs. 30 s idle | earlier by |
| --- | --- | --- | --- |
| 50 ms (~25 ms RTT) | **6.4 s** | 30 s | ~23.6 s |
| 100 ms (~50 ms RTT) | **12.7 s** | 30 s | ~17.3 s |
| 150 ms (~75 ms RTT) | 19.0 s | 30 s | ~11 s |
| 200 ms (~100 ms RTT) | 25.4 s | 30 s | ~4.6 s |

Beyond "earlier", the close carries a **distinct** reason (`TooManyPtos`, not `IdleTimeout`), so the application can tell a black hole apart from an ordinary idle close and react (error page, disable HTTP/3 for the origin and retry over HTTP/2, ...). That consumption is out of scope here; this PR only produces the earlier, distinct signal.

## Prior art (this is not novel)

Closing or abandoning a path before the idle timeout, driven by unacked probes, is what the other major stacks already do, including the server in this incident:

- **mvfst** (Meta, i.e. Facebook's own server): `maxNumPTOs = 7` tears the connection down,   with a degrading/blackhole ladder at 4/6:   https://github.com/facebook/mvfst/blob/main/quic/QuicConstants.h
- **QUICHE** (Chrome): default "5-RTO" blackhole detection (2 TLP + 5 RTOs) closing with   `QUIC_TOO_MANY_RTOS`, on for clients:   https://github.com/google/quiche/blob/main/quiche/quic/core/quic_sent_packet_manager.cc
- **msquic** (Microsoft): a fixed-time variant, `DisconnectTimeoutMs = 16s` (RTT-independent, so it does not track the path the way a PTO count does): https://github.com/microsoft/msquic/blob/main/docs/Settings.md

`7` is chosen to match mvfst's `maxNumPTOs` and QUICHE's effective 2+5 cumulative RTO delays. We might want to be more aggressive, i.e. choose a value like `5`.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1067,6 +1067,27 @@ impl Connection {
             qlog::packets_lost(&mut self.qlog, &lost, now);
         }
 
+        // Declare the connection broken if too many consecutive PTOs have gone
+        // unacknowledged, i.e. the path is a black hole. This closes the connection
+        // sooner than, and with a distinct reason from, the idle timeout.
+        //
+        // We only do this once connected, i.e. after the handshake completes. A
+        // black hole during the handshake is the job of happy eyeballs, which races
+        // multiple connection attempts and does not need us to give up on any single
+        // one early. Requiring `connected()` also guarantees we have an RTT sample,
+        // so the PTO backoff we are counting is based on a real estimate.
+        if let Some(max_pto) = self.conn_params.get_max_pto()
+            && self.state.connected()
+            && self.loss_recovery.pto_count() >= max_pto.get()
+        {
+            qinfo!("[{self}] {max_pto} consecutive PTOs, declaring connection broken");
+            self.set_state(
+                State::Closed(CloseReason::Transport(Error::TooManyPtos)),
+                now,
+            );
+            return;
+        }
+
         if self.release_resumption_token_timer.is_some() {
             self.create_resumption_token(now);
         }
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -4,7 +4,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::{cmp::max, time::Duration};
+use std::{cmp::max, num::NonZeroUsize, time::Duration};
 
 pub use crate::recovery::FAST_PTO_SCALE;
 use crate::{
@@ -133,6 +133,11 @@ pub struct ConnectionParameters {
     ack_ratio: u8,
     /// The duration of the idle timeout for the connection.
     idle_timeout: Duration,
+    /// Maximum number of consecutive PTOs (probe timeouts) without an
+    /// acknowledgement before the connection is declared broken and closed
+    /// locally. `None` (the default) disables the check, leaving the idle timeout
+    /// as the only backstop.
+    max_pto: Option<NonZeroUsize>,
     preferred_address: PreferredAddressConfig,
     datagram_size: u64,
     outgoing_datagram_queue: usize,
@@ -177,6 +182,7 @@ impl Default for ConnectionParameters {
             max_streams_uni: LOCAL_STREAM_LIMIT_UNI,
             ack_ratio: Self::DEFAULT_ACK_RATIO,
             idle_timeout: Self::DEFAULT_IDLE_TIMEOUT,
+            max_pto: None,
             preferred_address: PreferredAddressConfig::Default,
             datagram_size: MAX_DATAGRAM_FRAME_SIZE,
             outgoing_datagram_queue: MAX_QUEUED_DATAGRAMS_DEFAULT,
@@ -363,6 +369,25 @@ impl ConnectionParameters {
     #[must_use]
     pub const fn get_idle_timeout(&self) -> Duration {
         self.idle_timeout
+    }
+
+    /// Close the connection once `count` consecutive PTOs have fired without any
+    /// acknowledgement, treating the path as a black hole. `None` disables the
+    /// check, leaving the idle timeout as the only backstop.
+    ///
+    /// A value of `7` matches mvfst's `maxNumPTOs` and Google QUICHE's default
+    /// "5-RTO" blackhole detection (2 tail-loss probes + 5 RTOs). msquic instead
+    /// uses a fixed-time `DisconnectTimeoutMs` of 16s, which, being independent of
+    /// the RTT, only lines up with a PTO count on one specific RTT.
+    #[must_use]
+    pub const fn max_pto(mut self, count: Option<NonZeroUsize>) -> Self {
+        self.max_pto = count;
+        self
+    }
+
+    #[must_use]
+    pub const fn get_max_pto(&self) -> Option<NonZeroUsize> {
+        self.max_pto
     }
 
     #[must_use]
--- a/neqo-transport/src/connection/tests/idle.rs
+++ b/neqo-transport/src/connection/tests/idle.rs
@@ -4,19 +4,22 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::time::{Duration, Instant};
+use std::{
+    num::NonZeroUsize,
+    time::{Duration, Instant},
+};
 
 use neqo_common::{Encoder, qtrace, qwarn};
 use test_fixture::now;
 
 use super::{
     super::{Connection, ConnectionParameters, IdleTimeout, Output, State},
-    AT_LEAST_PTO, DEFAULT_STREAM_DATA, connect, connect_force_idle, connect_rtt_idle,
+    AT_LEAST_PTO, DEFAULT_RTT, DEFAULT_STREAM_DATA, connect, connect_force_idle, connect_rtt_idle,
     connect_with_rtt, default_client, default_server, maybe_authenticate, new_client, new_server,
     send_and_receive, send_something,
 };
 use crate::{
-    packet, recovery,
+    CloseReason, Error, packet, recovery,
     stats::FrameStats,
     stream_id::{StreamId, StreamType},
     tparams::{TransportParameter, TransportParameterId},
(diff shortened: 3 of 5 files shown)


larseggert:
What if there is just a temporary drop in connectivity? Killing connections because the link went down for 10s seems harsh?

mxinden:
> What if there is just a temporary drop in connectivity? Killing connections because the link went down for 10s seems harsh?

We would kill the connection and establish a new one. I would argue after a 10s connectivity loss, the <1s connection re-establishment time (potentially using 0-RTT) is negligible in terms of user experience.

larseggert:
I'm all for automatically/transparently retrying sooner, but an earlier comments talks about "surface an actionable error page sooner", which is not that.

mxinden:
> I'm all for automatically/transparently retrying sooner, but an earlier comments talks about "surface an actionable error page sooner", which is not that.

@larseggert I am not sure I am following. Can you rephrase the above?

In the facebook incident case, requests were sent on a dead connection, but only failed after 30s, i.e. the connection idle timeout. The faster we kill the dead connection, the faster Firefox learns that the request fails, the faster Firefox can show a network error page, e.g. with a hint to disable ones anti virus.

larseggert:
From the first comment, it sounded like what this does is show an error page sooner, which I think is of questionable benefit. Your later comment then says this would let us retry sooner, which is obviously vastly better. I was reacting on the discrepancy.

martinthomson:
I'm supportive of this feature, though I think that we should move closer to the Chrome approach, where we enable the fast PTO as well.  The fast PTO isn't quite a a TLP, but it does accelerate loss detection far more than a PTO.  7 PTOs with a PTO speedup brings the times down even further.

Note that we should ensure that this feature is not engaged until we have an RTT estimate.  I need to check how you implemented this, but the super-long RTT cases will fail to connect if we let this act without a solid estimate.

neqo-transport/src/connection/mod.rs
```diff
@@ -1067,6 +1067,20 @@ impl Connection {
             qlog::packets_lost(&mut self.qlog, &lost, now);
         }
 
+        // Declare the connection broken if too many consecutive PTOs have gone
+        // unacknowledged, i.e. the path is a black hole. This closes the connection
+        // sooner than, and with a distinct reason from, the idle timeout.
+        if let Some(max_pto) = self.conn_params.get_max_pto()
+            && self.loss_recovery.pto_count() >= max_pto
```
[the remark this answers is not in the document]
martinthomson: The suggestion here from the bot is close to what I have suggested.  I don't mind which of these you use, but I would maybe prefer a check on the RTT.  Then the check will engage during a handshake as well.

```suggestion
        if let Some(max_pto) = self.conn_params.get_max_pto()
            && self.loss_recovery.pto_count() >= max_pto
            && let Some(p) = self.paths.primary()
            && p.borrow().rtt().first_sample_time().is_some()
```

The connected check does let the other happy eyeballs machinery work, so that is probably fine if you think the RTT check is a bit much.  (The code cost is all that would flip this.)

neqo-transport/src/connection/params.rs
```diff
@@ -133,6 +133,11 @@ pub struct ConnectionParameters {
     ack_ratio: u8,
     /// The duration of the idle timeout for the connection.
     idle_timeout: Duration,
+    /// Maximum number of consecutive PTOs (probe timeouts) without an
+    /// acknowledgement before the connection is declared broken and closed
+    /// locally. `None` (the default) disables the check, leaving the idle timeout
+    /// as the only backstop.
+    max_pto: Option<usize>,
```
martinthomson:
```suggestion
    max_pto: Option<NonZeroUsize>,
```

neqo-transport/src/recovery/mod.rs
```diff
@@ -823,6 +823,12 @@ impl Loss {
         )
     }
 
+    /// The number of consecutive PTOs that have fired without being acknowledged.
+    /// Reset to `0` whenever an acknowledgement is received.
```
martinthomson:
```suggestion
    /// The value is reset to `0` whenever an acknowledgement is received.
```

neqo-transport/src/lib.rs
```diff
@@ -294,6 +296,7 @@ mod tests {
             (Error::EchRetry(vec![]), 0x179),
             (Error::VersionNegotiation, 0x53f8),
             (Error::Internal, 1),
+            (Error::TooManyPtos, 1),
```
martinthomson: Is this a case where we could use NO_VIABLE_PATH (0x10)?

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -801,3 +801,51 @@ fn keep_alive_with_unresponsive_server() {
...
+            match client.process_output(now) {
+                // Drop the PTO probe(s) without advancing time.
+                Output::Datagram(_) => {}
+                Output::Callback(t) => break t,
+                Output::None => break Duration::ZERO,
+            }
+        };
+        if matches!(client.state(), State::Closed(_)) {
+            break;
+        }
```
martinthomson: Do you need this break?  The loop terminates after 7 PTOs and you check that.

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -801,3 +801,51 @@ fn keep_alive_with_unresponsive_server() {
...
+        client.state()
+    );
+    assert_eq!(client.loss_recovery.pto_count(), MAX_PTO);
+    // The whole point: we gave up well before the idle timeout would have fired.
+    assert!(
+        now - start < default_timeout(),
+        "closed after {:?}, which is not earlier than the {:?} idle timeout",
+        now - start,
+        default_timeout()
+    );
```
martinthomson, line 855: This check only works if you also check that (1<<MAX_PTO - 1)*3*DEFAULT_RTT < default_timeout(), right?

larseggert:
FWIW, we have been doing fast PTO for the DoH connections in Gecko for a while now, and I don't think we're seeing issues. Am OK with doing it for all.

larseggert:
> FWIW, we have been doing fast PTO for the DoH connections in Gecko for a while now, and I don't think we're seeing issues. Am OK with doing it for all.

https://bugzilla.mozilla.org/show_bug.cgi?id=2047616

neqo-transport/src/connection/tests/idle.rs
mxinden: 👍 added above:

```
    // For this test to prove anything, the whole PTO backoff ladder must fit inside
    // the idle timeout; otherwise "closed before the idle timeout" is vacuously true.
    assert!(DEFAULT_RTT * (1_u32 << (MAX_PTO as u32 - 1)) * 3 < default_timeout());
```

neqo-transport/src/connection/tests/idle.rs
mxinden: `break` removed.

neqo-transport/src/connection/mod.rs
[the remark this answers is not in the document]
mxinden: Thanks for the suggestions. As a big Happy Eyeballs fan, I opted for the bot suggestion, i.e. only act after `self.state.connected()`, thus delegating pre-connection black hole responsibility to Happy Eyeballs.

neqo-transport/src/connection/tests/idle.rs
```diff
@@ -801,3 +801,51 @@ fn keep_alive_with_unresponsive_server() {
...
+}
```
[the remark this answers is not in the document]
mxinden: Added.

neqo-transport/src/recovery/mod.rs
```diff
@@ -823,6 +823,12 @@ impl Loss {
...
+    pub fn pto_count(&self) -> usize {
```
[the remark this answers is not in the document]
mxinden: done

neqo-transport/src/connection/params.rs
```diff
@@ -365,6 +371,25 @@ impl ConnectionParameters {
...
+    ///
+    /// A value of `7` matches mvfst's `maxNumPTOs` and Google QUICHE's default
+    /// "5-RTO" blackhole detection (2 tail-loss probes + 5 RTOs). msquic instead
+    /// uses a fixed-time `DisconnectTimeoutMs` of 16s, which, being independent of
+    /// the RTT, only lines up with a PTO count on one specific RTT.
+    #[must_use]
+    pub const fn max_pto(mut self, count: Option<usize>) -> Self {
+        self.max_pto = count;
+        self
+    }
```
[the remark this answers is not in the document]
mxinden: Switched to NonZero type.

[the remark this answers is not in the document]
mxinden: switched to nonzero type

neqo-transport/src/lib.rs
mxinden: Done. Sending `NO_VIABLE_PATH` on the wire.

mxinden:
Thank you for the input.

I would like to move forward here, i.e. merge independent of Fast PTO. In my eyes this PR is an improvement in itself, only getting better with Fast PTO. The latter I would like to further discuss in https://github.com/mozilla/neqo/pull/1236#issuecomment-4751320180.

@martinthomson @larseggert any objections?

larseggert (approved):

mxinden:
Merging here. Discussed with Martin at IETF. Off by default anyways. Happy to follow-up on any additional review comments after merge.

PR #3674, opened 2026-06-07
fix: lift the 32-range cap on ACK frames

### LINKED ISSUE #3641: ACK frames are unnecessarily limited to 32 ranges
## Problem

`write_frame` in `tracking.rs` caps ACK frames at `MAX_ACKS_PER_FRAME` (32) ranges to "keep plenty of space for other stuff." This cap is overly conservative in two situations:

1. **ACK-only packets (cwnd < 256):** These are limited to 255 bytes by `ACK_ONLY_SIZE_LIMIT`, which fits only ~13 ranges. The cap of 32 never fires — the space constraint is always binding. These packets could use the full MTU instead, since ACK-only packets are exempt from congestion control per RFC 9002 Section 6.1.

2. **Non-ACK-only packets with nothing else to send:** When cwnd >= 256 but no data or control frames are queued, the packet is effectively ACK-only, yet the 32-range cap still applies. At a 1200-byte MTU, up to ~72 ranges could fit, but 40 of those slots go unused.

## Impact

After #3639 increased `MAX_TRACKED_RANGES` to 128, we can now track many more ranges — but we can only communicate 32 per frame. Under reordering, this means it takes more round trips to acknowledge all received packets, prolonging the window where the sender might falsely declare packets lost.

## Possible approaches

- **Remove `MAX_ACKS_PER_FRAME` entirely.** The space-based formula `1 + (avail / 16)` already uses a pessimistic 16-bytes-per-range estimate. In practice, most ranges encode much smaller, so even at the calculated maximum, the ACK frame leaves substantial room for other frames. The cap is belt-and-suspenders on top of an already conservative estimate.

- **Lift `ACK_ONLY_SIZE_LIMIT` to the full MTU.** Since ACK-only packets don't count against the congestion window (RFC 9002 Section 6.1), there's no reason to artificially constrain them to 255 bytes. This was an implementation choice from #618 (2020), not an RFC requirement.

- **Write other frames first, ACK last.** This would let the ACK frame fill whatever space remains, but is a more invasive change.

## Context

`ACK_ONLY_SIZE_LIMIT` (256) was introduced in #618 and `MAX_ACKS_PER_FRAME` (32) in #998, both without recorded discussion about the specific values.

mvanhorn:
## Summary

ACK frames are no longer capped at 32 ranges. `write_frame` in `neqo-transport/src/tracking.rs` previously truncated to `MAX_RANGES = 32` unconditionally; it now sizes the range count against the actual space available in the packet, so loss patterns producing many small holes get acknowledged in one frame instead of dropping older ranges, as requested in #3641.

## Why this matters

larseggert filed #3641 after observing the hard cap: with more than 32 ranges, the peer never learns about older holes and retransmits data the receiver already has. The frame builder already negotiates space byte by byte; the cap was a leftover safety bound. Sizing now accounts for the range-count varint growing to two bytes at 64+ ranges (review round caught that), so `ACK_FRAME_RESERVE` still guarantees room for a following useful frame in tight packets.

## Testing

Added tracking tests covering: many-range ACK in a roomy packet (all ranges encoded), tight-packet truncation (reserve preserved, frame still parses), and the 64+ range varint boundary. `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full `neqo-transport` suite (772 tests) pass locally.

AI was used for assistance.

Fixes #3641


larseggert (approved):
I confirmed this fixes the "dropped ranges" occurrences I saw over an emulated path (400ms RTT, 1Mbit/s). Performance wasn't improved however.

Can we extract some repeated code from the various new tests into helpers to be more DRY?

mvanhorn:
Thanks for testing it over the emulated path, that's great to hear. Done in 05b05c5: the repeated tracker setup, ACK frame write, token range assertion, and frame decode logic are now shared helpers, which removed about 60 duplicated lines from the new tests. `cargo test -p neqo-transport --lib tracking` (31 tests) and clippy both pass locally.

martinthomson (commented):
In looking at this changeset, I don't think that it approaches the problem from the right angle.  When things are an uncertain length and we have a cap, we should only need to rewind the encoding by a single item at the point that the frame overruns the limit.  The O(n^2) linear search for the right size is not the right way to build this.

mvanhorn:
@martinthomson thanks for looking. One correction and then I think we agree on the direction.

The sizing loop is not an O(n^2) search. `ack_len_for_range_count` is closed-form arithmetic (constant frame overhead, plus 16 bytes per extra range, plus the varint width of the extra-range count), so each iteration is O(1). The initial estimate is `1 + avail/16`, which only ever overshoots by the growth of the extra-range varint, so the loop decrements once or twice at most.

The real weakness is the one your approach fixes: 16 bytes per range is a worst case, so on typical ranges the frame is packed well under the limit and we send fewer ranges than would fit.

The wrinkle with rewinding a single item is that the frame writes the extra-range count as a varint before any of the ranges, and its width depends on the final count. Dropping the last range can shrink that field, so the rewind is not purely local to the tail. It is handleable, either by reserving the varint at its estimated width or by re-encoding the header after the count settles.

If you are happy with the latter, I will rework it to encode ranges until the frame overruns the limit, rewind the last one, and fix up the count, which drops the pre-computation entirely and packs tighter than this version. I will also rebase, the branch is behind main.

larseggert:
@codspeedbot fix this regression

mvanhorn:
Thanks @larseggert. CodSpeedBot looked into it and found no regression attributable to this PR: head 0eaaf2c and main at 9f9c801, measured on the same macro runners, give identical numbers (coalesce_acked_from_zero 1000 ranges at 655.386 us on both, peak memory 2,443,591 B on both). A main-vs-main comparison across the runner migration (fb2eb16 to 9f9c801) reproduces the same 10 regressions and 6 improvements with zero source changes in between, so the report's base was just measured in the old environment.

The remaining red checks are the Vendor into Gecko jobs, which are currently failing on every branch including the merge-queue run for #4016 on main, so those look unrelated too.

One thing CodSpeedBot noticed that is worth recording: the ACK path this PR changes is not exercised by the bench suite. transfer_simulated drops packets through tail-drop queues, so losses arrive as contiguous bursts and never reach the old 32-range cap. If you want CI coverage for the many-small-holes pattern from #3641, that needs a simulated path with sparse random loss. Happy to add one in a follow-up if you think it is worth having.

Anything else you want before this goes in?

larseggert:
@mvanhorn I think you wanted to rewrite this to implement @martinthomson's go-back-1-on-overrun design?

PR #3684, opened 2026-06-10
fix: guard qpack increment_acked against u64 overflow

uwezkhan:
Repro: on the QPACK decoder stream, send an Insert Count Increment whose value pushes the encoder's acknowledged-inserts count past u64::MAX (one valid increment, then a second of u64::MAX).
Cause: increment_acked adds the peer-supplied increment to acked_inserts_cnt before the range check, so the add itself overflows. Debug builds panic; release builds wrap and slip past the `base < acked_inserts_cnt` check, leaving the known-received count corrupted.
Fix: add with checked_add and map overflow to IncrementAck, the same error an over-large increment already produces. Valid increments behave exactly as before; only the overflow edge changes from a panic/wrap to a clean decoder-stream error.

### DIFF
--- a/neqo-qpack/src/table.rs
+++ b/neqo-qpack/src/table.rs
@@ -362,10 +362,14 @@ impl HeaderTable {
     /// `IncrementAck` if ack is greater than actual number of inserts.
     pub fn increment_acked(&mut self, increment: u64) -> Res<()> {
         qtrace!("[{self}] increment acked by {increment}");
-        self.acked_inserts_cnt += increment;
-        if self.base < self.acked_inserts_cnt {
+        let acked = self
+            .acked_inserts_cnt
+            .checked_add(increment)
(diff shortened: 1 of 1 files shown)


larseggert (approved):
LGTM but would like someone else with a better grasp on QPACK to take a look.

martinthomson (approved):
This is a fine fix, but I would have preferred that the test case be targeted at the messaging rather than just doing a narrow check on the operation of the function.  In this case, the narrower test is acceptable.

I had to double check the range of integers we accept from peers here.  If this were a QUIC varint, then this wouldn't have been a problem.  But we accept 2^64-1 values from our qpack integer encoder.

I considered suggesting a change to that cap, which might have made the code simpler there at the same time, but this fix is neater.

larseggert:
@uwezkhan thanks for your contribution!

PR #3686, opened 2026-06-10
feat: replace Packets `BTreeMap` with `VecDeque`

larseggert:
Packet numbers are monotonically increasing, so track() becomes an O(1)
push_back instead of O(log n) BTree insert with pointer-chasing and
potential node splits. ACKs remove the oldest (front) packets; VecDeque
drain moves only the short front side, avoiding BTree's per-node heap
allocations while keeping cache-friendly contiguous storage.

take_ranges uses partition_point + drain(...).rev() per range;
remove_expired front-drains after an O(1) fast-exit check.

### DIFF
--- a/neqo-transport/src/recovery/sent.rs
+++ b/neqo-transport/src/recovery/sent.rs
@@ -7,7 +7,7 @@
 // A collection for sent packets.
 
 use std::{
-    collections::BTreeMap,
+    collections::VecDeque,
     ops::RangeInclusive,
     rc::Rc,
     time::{Duration, Instant},
@@ -214,7 +214,7 @@ impl Packet {
 #[derive(Debug, Default)]
 pub struct Packets {
     /// The collection.
-    packets: BTreeMap<u64, Packet>,
+    packets: VecDeque<Packet>,
 }
 
 impl Packets {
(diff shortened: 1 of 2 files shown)


neqo-transport/src/recovery/sent.rs
```diff
@@ -243,93 +246,58 @@ impl Packets {
...
+                start_idx + self.packets[start_idx..].partition_point(|p| p.pn <= *range.end());
+            if start_idx >= end_idx {
+                continue;
             }
-
-            // Take the acked packets.
-            result.extend(acked_range.into_values().rev());
+            let before = result.len();
+            result.extend(self.packets.drain(start_idx..end_idx));
+            result[before..].reverse();
```
martinthomson: Is it faster to reverse the output of `drain()` as it is passed to `extend()`?  This is going to involve a bunch of memory swaps here, but it might not if you chain the iterator.  The drawback I see there is that you might not be able to access `extend()` optimizations.

This is less code:
```suggestion
            result.extend(self.packets.drain(start_idx..end_idx).rev());
```

neqo-transport/src/recovery/sent.rs
```diff
@@ -243,93 +246,58 @@ impl Packets {
...
-                // instead replace the empty one entirely.
-                self.packets = after_acked_range;
-            } else {
-                // Need to extend existing one. Not the first iteration, thus
-                // `after_acked_range` should be small.
-                self.packets.extend(after_acked_range);
+            let start_idx = self.packets.partition_point(|p| p.pn < *range.start());
+            let end_idx =
+                start_idx + self.packets[start_idx..].partition_point(|p| p.pn <= *range.end());
+            if start_idx >= end_idx {
```
martinthomson:
```suggestion
            if start_idx == end_idx {
```

martinthomson (approved):
Is this faster with `VecDeque`?  We constantly add at one end and remove from the other, which suggests that there is a performance gain to be had from not having to move stuff around on every drain.

Another potential improvement highlighted below as well, but I'm happy with this if the benchmarks show any sign of gains.  Or even if the performance is comparable.  The code is much easier to reason about.

larseggert:
Messed up a rebase, fixed.

PR #3687, opened 2026-06-10
feat: skip streams with no data in SendStreams::write_frames

larseggert:
Add `SendStream::has_data_at(priority)` to cheaply check whether any frame (RESET_STREAM, STREAM_DATA_BLOCKED, or STREAM) would be written, and gate both iteration loops on it. Eliminates the per-stream dispatch for idle streams on every priority pass.

Let's see what the benches say.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -768,8 +768,9 @@ impl SendStream {
         self.has_next_bytes(retransmission)
     }
 
-    // return false if the builder is full and the caller should stop iterating
-    // Any new frame type added here must also be reflected in [`Self::has_data_at`].
+    /// Return `false` if the builder is full and the caller should stop iterating.
+    ///
+    /// Any new frame type added here must also be reflected in [`Self::has_data_at`].
     pub fn write_frames<B: Buffer>(
         &mut self,
         priority: TransmissionPriority,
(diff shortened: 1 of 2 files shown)


neqo-transport/src/send_stream.rs
```diff
@@ -727,6 +727,36 @@ impl SendStream {
         ss
     }
 
+    /// Returns `true` if [`Self::write_frames`] at this priority could emit any frame.
+    ///
+    /// Conservative: may return `true` when nothing is ultimately written, but never returns
+    /// `false` when something needs to be written.
```
martinthomson: In what specific case would this report true when there is nothing to send?  I can't spot it.

neqo-transport/src/send_stream.rs
```diff
@@ -727,6 +727,36 @@ impl SendStream {
...
+        }
+        // STREAM pending?
+        let retransmission = if priority == self.priority {
+            false
+        } else if priority == self.effective_priority {
+            true
+        } else {
+            return false;
+        };
+        self.next_bytes(retransmission).is_some()
```
martinthomson: This line is the one I'd look at more carefully.  `next_bytes()` does work to identify exact ranges, but you only really need a check for whether bytes exist.  `has_next_bytes()` might be even faster.

martinthomson (approved):
The bot's comment about maintenance coupling is a good one.  Updating `write_frames()` with a note about how `has_data_at` needs to be updated in sync is a good one.

PR #3689, opened 2026-06-10
feat: inline Pacer::spend return condition

larseggert:
To avoid redundant next() call.

When self.t == now (which is always true at the end of spend()), the next() > now check reduces to a simple arithmetic comparison. Inline it to eliminate the function-call overhead and duplicate credit/deficit arithmetic.

### DIFF
--- a/neqo-transport/src/pace.rs
+++ b/neqo-transport/src/pace.rs
@@ -88,42 +88,48 @@ impl Pacer {
         self.p = mtu;
     }
 
+    /// Nanoseconds to wait after `self.t` before the next packet may be sent.
+    /// Returns `0` when a packet can be sent immediately: credit is available,
+    /// the deficit overflows, `cwnd` is zero, or the computed wait is below the
+    /// timer granularity.
+    #[inline]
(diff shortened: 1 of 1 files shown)


neqo-transport/src/pace.rs
```diff
@@ -172,7 +179,21 @@ impl Pacer {
...
+        }
+        let Ok(deficit) = u64::try_from(packet - self.c) else {
+            return false; // overflow: treat as no wait
+        };
+        let rtt_ns = u64::try_from(rtt.as_nanos()).unwrap_or(u64::MAX);
+        let divisor = (cwnd as u64).saturating_mul(Self::SPEEDUP);
+        if divisor == 0 {
+            return false;
+        }
+        rtt_ns.saturating_mul(deficit) / divisor >= Self::GRANULARITY_NS
```
martinthomson: That's a lot of inlined/copy-pasted code.  Is there any way to factor this stuff better?  And then mark that function as `#[inline]` or whatever it takes to get it back inline.

larseggert:
Not enough of an improvement.

PR #3690, opened 2026-06-10
feat: fast path in RxStreamOrderer::inbound_frame

larseggert:
Add `frontier_end` field tracking the end offset of the rightmost received byte. When new data arrives at or beyond the frontier (the common in-order case), skip the unconditional `range_mut(..=new_start).next_back()` BTree probe.

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -180,12 +180,24 @@ pub struct RxStreamOrderer {
     data_ranges: BTreeMap<u64, Vec<u8>>, // (start_offset, data)
     retired: u64,                        // Number of bytes the application has read
     received: u64,                       // The number of bytes stored in `data_ranges`
+    /// End offset of the rightmost received byte (the end of the last entry in
+    /// `data_ranges`, or `retired` if the map is empty).  Used to skip the
+    /// unconditional backwards `BTreeMap` probe in `inbound_frame` for the common
+    /// in-order case.
+    frontier_end: u64,
 }
 
 impl RxStreamOrderer {
     #[must_use]
     pub fn new() -> Self {
         Self::default()
+    }
+
+    /// Allocate a `Vec` for a new data-range entry, pre-sized with headroom.
+    fn make_buf(data: &[u8]) -> Vec<u8> {
+        let mut v = Vec::with_capacity(data.len().next_power_of_two().max(4096));
+        v.extend_from_slice(data);
+        v
     }
 
     /// Process an incoming stream frame off the wire. This may result in data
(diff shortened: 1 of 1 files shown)


larseggert:
Not worth it

neqo-transport/src/recv_stream.rs
```diff
@@ -218,6 +224,36 @@ impl RxStreamOrderer {
...
+            debug_assert_eq!(
+                self.end,
+                self.data_ranges
+                    .last_key_value()
+                    .map_or(self.retired, |(&k, v)| {
+                        k + u64::try_from(v.len()).expect("usize fits in u64")
+                    }),
+                "end must equal the end of the last range, or retired if empty"
+            );
+            self.received += u64::try_from(new_data.len()).expect("usize fits in u64");
```
martinthomson, line 241: I'm starting to think that we need a helper for this conversion.  It can be `#[inline(always)]`, but I don't like seeing this same code repeated everywhere.

neqo-transport/src/recv_stream.rs
```diff
@@ -218,6 +224,36 @@ impl RxStreamOrderer {
...
+            // Adjacent: try to extend the last entry to avoid a BTreeMap insert.
+            // Gap (new_start > end): short-circuits straight to insert.
+            if new_start != self.end
+                || self
+                    .data_ranges
+                    .last_entry()
+                    .filter(|e| e.get().len() < Self::RANGE_LEN)
+                    .map(|mut e| e.get_mut().extend_from_slice(new_data))
+                    .is_none()
+            {
```
martinthomson: I think that the bot has a suggestion to avoid the hidden side-effect of this map call.  You can definitely do better than this.

neqo-transport/src/recv_stream.rs
```diff
@@ -218,6 +224,36 @@ impl RxStreamOrderer {
...
+                "end must equal the end of the last range, or retired if empty"
+            );
+            self.received += u64::try_from(new_data.len()).expect("usize fits in u64");
+            // Adjacent: try to extend the last entry to avoid a BTreeMap insert.
+            // Gap (new_start > end): short-circuits straight to insert.
+            if new_start != self.end
+                || self
+                    .data_ranges
+                    .last_entry()
+                    .filter(|e| e.get().len() < Self::RANGE_LEN)
```
martinthomson: This test doesn't check the new length, it checks the existing length.  That means that the ultimate size of chunks we store could be up to one MTU larger than `RANGE_LEN`.   Is that what you want?  If it is, it might be worth a short note.

neqo-transport/src/recv_stream.rs
```diff
@@ -180,9 +180,15 @@ pub struct RxStreamOrderer {
...
     retired: u64,                        // Number of bytes the application has read
     received: u64,                       // The number of bytes stored in `data_ranges`
+    /// Offset of the rightmost received byte (the end of the last entry in
+    /// `data_ranges`, or `retired` if the map is empty).
+    end: u64,
 }
 
 impl RxStreamOrderer {
+    /// Maximum length of a buffered range before a new entry is created instead of extending.
+    const RANGE_LEN: usize = 4096;
```
martinthomson: Not sure about this name.  `MAX_CHUNK_SIZE` or something like that might be better.  Except that it's not a maximum in truth, more of a target size.

BTW, if the MTU exceeds this number, we're going to buffer chunks much larger than this. That's a good thing, I think, but it is worth documenting.

martinthomson: Separately, did you consider trying variations on this size in our benchmarks to see how that plays out?  It seems like this is in approximately the right spot, but a larger value could mean less BTreeMap traffic, which could be good, provided that the copying is cheap enough.

martinthomson (approved):
I have to confess that I don't trust the benchmarks much, but these seem good.

Please do fix the factoring of the extension.  There has to be a cleaner way to do what you want without hiding the operation like this does.

neqo-transport/src/recv_stream.rs
larseggert: `RANGE_TARGET`?

I did try 64KB and 256KB instead of 4K. Basically, what we're saving in BTreeMap operations were then spending on memcpy.

neqo-transport/src/recv_stream.rs
larseggert: I'll do this in a separate PR.

PR #3691, opened 2026-06-10
feat: reuse ACK scratch buffer in LossRecoverySpace

larseggert:
Change Packets::take_ranges to fill a caller-provided Vec rather than
allocating a fresh one. LossRecoverySpace gains an acked_scratch field
that persists across ACK frames, eliminating one Vec allocation per
ACK received.


larseggert:
I don't think the performance benefits here warrant the complexity.

larseggert:
Not worth it.

PR #3692, opened 2026-06-10
feat: skip SmallVec allocation in mark_acked/mark_sent

larseggert:
Use peekable() to check for covered entries before collecting into a SmallVec. The common case (sequential sends, typical gap-ACKs) has an empty covered set, so no SmallVec is constructed. The slow path is unchanged.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -184,6 +184,18 @@ impl RangeTracker {
         }
     }
 
+    /// Collect the keys of all `used` entries that start within `[start, end)`.
+    /// Returns an empty `SmallVec` — without heap allocation — when the range is clear,
+    /// which is the common case for both `mark_acked` and `mark_sent`.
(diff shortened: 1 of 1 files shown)


neqo-transport/src/send_stream.rs
```diff
@@ -309,12 +315,17 @@ impl RangeTracker {
...
+        // case (sequential sends and retransmissions after loss). With acked entries
+        // inside the range after out-of-order ACKs (rare), collect them for the body below.
+        let covered: Vec<u64> = {
+            let mut it = self.used.range(new_off..new_end).peekable();
+            if it.peek().is_some() {
+                it.map(|(&k, _)| k).collect()
+            } else {
+                Vec::new()
+            }
+        };
```
martinthomson: I'm seeing copied and pasted code here.

martinthomson (changes requested):

larseggert:
@martinthomson no need to review draft PRs - I push those just to see CI benchmark results.

larseggert:
Not resulting in conclosive gains.

PR #3694, opened 2026-06-10
move cwnd tracking from CC stats to connection-level stats snapshot

jesup:


jesup:

1. `trunk()`
1. **users/jesup/cc_stats ← this PR**

---

neqo-transport/src/stats.rs
```diff
@@ -518,7 +519,12 @@ impl Debug for Stats {
             "  bytes: rx {} lost {} acked {}",
             self.bytes_rx, self.bytes_lost, self.bytes_acked
         )?;
-        writeln!(f, "  min_rtt: {:?}", self.min_rtt)
+        writeln!(f, "  min_rtt: {:?}", self.min_rtt)?;
+        writeln!(
+            f,
+            "  cwnd: {} bytes_in_flight: {}",
```
martinthomson:
```suggestion
            "  cwnd: {} in_flight: {}",
```

neqo-transport/src/stats.rs
```diff
@@ -626,6 +632,7 @@ fn debug() {
     mark transitions:
   dscp: 
   bytes: rx 0 lost 0 acked 0
-  min_rtt: 0ns\n"
+  min_rtt: 0ns
+  cwnd: 0 bytes_in_flight: 0\n"
```
martinthomson:
```suggestion
  cwnd: 0 in_flight: 0\n"
```

neqo-transport/src/stats.rs
```diff
@@ -394,6 +391,10 @@ pub struct Stats {
     pub bytes_lost: usize,
     /// Total bytes in acknowledged packets.
     pub bytes_acked: usize,
+    /// Congestion window size.
+    pub cwnd: usize,
+    /// Bytes in flight (sent but not yet acked or declared lost).
+    pub bytes_in_flight: usize,
```
martinthomson: I don't follow why this change is being made.

The `Stats` object has a `cc` member that used to cover this.  I get that it might be preferable to pull the current value from the object when stats are pulled, rather than have it constantly updated, but that doesn't motivate the move.

Instead, I would add these two items under the CC stats and note that the values in that struct aren't live: they are only updated when stats are pulled from the connection.  This is something the bot notes as well.

neqo-transport/src/stats.rs
martinthomson, line 652: This is terrible formatting.  @larseggert @mxinden we can do better here.

neqo-transport/src/connection/mod.rs
```diff
@@ -974,6 +974,8 @@ impl Connection {
             v.rtt = p.rtt().estimate();
             v.rttvar = p.rtt().rttvar();
             v.min_rtt = p.rtt().minimum();
+            v.cwnd = p.sender().cwnd();
+            v.bytes_in_flight = p.sender().bytes_in_flight();
```
martinthomson: This is now at the point that it would make sense to have a function on the path that takes a snapshot of the state of these stats, rather than doing all the work here.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -796,7 +791,7 @@ where
         pto: Duration,
         lost_packets: impl IntoIterator<Item = &'a sent::Packet>,
         now: Instant,
-        cc_stats: &mut CongestionControlStats,
+        _cc_stats: &mut CongestionControlStats,
```
martinthomson: If you are going to update CWND when stats are pulled, then *remove* the argument.

neqo-transport/src/stats.rs
```diff
@@ -481,8 +482,8 @@ impl Debug for Stats {
         )?;
         writeln!(
             f,
-            "    final_cwnd {:?} ss_exit_cwnd {:?} ss_exit_reason {:?}",
-            self.cc.cwnd, self.cc.slow_start_exit_cwnd, self.cc.slow_start_exit_reason
+            "    ss_exit_cwnd {:?} ss_exit_reason {:?}",
+            self.cc.slow_start_exit_cwnd, self.cc.slow_start_exit_reason
         )?;
```
martinthomson: This should all be handled by the CC stats object.  That is, we should avoid playing with some else's toys.

neqo-transport/src/stats.rs
martinthomson: I would move the CWND/BIF reporting to the first line of this block.

martinthomson (changes requested):

PR #3696, opened 2026-06-11
feat: store `sent::Packet` tokens inline, removing `Rc`

larseggert:
Every sent packet called `Rc::new()` to wrap its recovery tokens — one extra heap allocation per packet on the hot send path. The `Rc` only served to make `Packet::clone()` cheap; clone only happens on the cold loss/PTO paths, where a real `Vec<Token>` copy may be fine.


martinthomson (approved):
I find the analysis convincing.  And the code change is straight up goodness.

PR #3702, opened 2026-06-12
feat: pre-allocate packet send buffer to MTU size to avoid realloc

larseggert:
We tried this before but Claude insists this is a good idea...


mxinden:
Which benchmark shows the reliable signal that this is an improvement?

> transfer/1-conn/1-100mb-req (aka. Upload): 💔 Performance has regressed by +55.900%.

I don't think this is a regression we are willing to accept?

mxinden (commented):
Firefox mostly sends ACKs and small requests past the handshake. Is this a good idea?

larseggert:
The regression is from before the latest changes, which should fix it. If not, obviously we can't accept.

I see no downside in using a 1500 byte buffer here?

mxinden:
Some historical context on this:

- #2086: dropped the up-front MTU allocation in `max_datagram_size` (`Encoder::with_capacity(mtu)` → `Encoder::default()`)
- #2670: discussion on not pre-allocating each UDP datagram in `output_path`
- #2747: groundwork for a single long-lived, reused send buffer

The long-lived send buffer in #2747 is in my eyes the best next step.

larseggert:
@mxinden this really looks performance positive?

mxinden:
Where do I find the https://github.com/mozilla/neqo/blob/main/neqo-bin/benches/main.rs benchmark results? I can't find them in https://github.com/mozilla/neqo/pull/3702#issuecomment-4922381072.

This is counterintuitive to me as:

- Firefox mostly sends ACKs, thus in a Download benchmark, we would always over-allocate.
- When Firefox uploads, I would expect it to use GSO, thus build >MTU sized datagrams, thus this patch would just like `main` under-allocate, granted a bit less than `main`.

larseggert:
> Where do I find the [`main`/neqo-bin/benches/main.rs](https://github.com/mozilla/neqo/blob/main/neqo-bin/benches/main.rs?rgh-link-date=2026-07-09T19%3A02%3A41Z) benchmark results? 

See #3804 - we haven't been generating them. (That PR was exactly motivated by me also wanting to compare profiles apples to apples...)

>     * Firefox mostly sends ACKs, thus in a Download benchmark, we would always over-allocate.

Well, only by 1500 bytes. I still don't see how that matters.

 >     * When Firefox uploads, I would expect it to use GSO, thus build >MTU sized datagrams, thus this patch would just like `main` under-allocate, granted a bit less than `main`.

Correct. But IIRC `Vec` allocates 16 entries by default, so for a full-sized packet, we save ~7 reallocs/copies with this. It is ~4-5 more reallocs/copies to go to GSO size, which we still need to pay (unless we can be smarter when we know we produce a GSO burst.)

larseggert:
Closing in favor of #3948

PR #3704, opened 2026-06-12
feat: cache has_data_at result in SendStream

larseggert:
Replace the per-call BTree probe with a bool maintained at every state-transition site.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -728,8 +728,7 @@ pub struct SendStream {
     fair: bool,
     send_group: Option<SendGroupId>,
     writable_event_low_watermark: NonZeroUsize,
-    /// Cached result: true iff `has_data_at(self.priority)` would return true.
-    /// Maintained at every state transition so the common path is an O(1) field read.
+    /// `has_data_at` cache: true iff `has_data_at(self.priority)` would return true.
     has_data: bool,
 }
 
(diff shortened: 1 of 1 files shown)


neqo-transport/src/send_stream.rs
```diff
@@ -1106,12 +1147,18 @@ impl SendStream {
...
+            self.update_has_data();
         }
+        written
     }
 
     pub fn blocked_lost(&mut self, limit: u64) {
         if let State::Ready { fc, .. } | State::Send { fc, .. } = &mut self.state {
             fc.frame_lost(limit);
+            // frame_lost re-arms blocked_frame, so the cache needs refreshing.
+            self.update_has_data();
```
martinthomson: Do you really need to track the blocked state?  You have a simple check in `has_data_at()` that seems to cover this path.  That would save you a bunch of these updates, reducing them to just those that involve stream data (and FIN and RESET_STREAM, of course).

neqo-transport/src/send_stream.rs
```diff
@@ -1331,7 +1386,11 @@ impl SendStream {
                 Ok(sent)
             }
             _ => Err(Error::FinalSize),
+        };
+        if result.is_ok() {
+            self.update_has_data();
         }
```
martinthomson, line 1543: Why is this here, rather than in `mark_as_sent()`?

It seems like you could use an unconditional call to `update_has_data()` to manage this more cleanly.  After all, that is always called when some data is sent.  A simple `len > 0 || fin` guard should work.

I say this because you have this in `mark_as_acked()` and `mark_as_lost()`, so it makes sense to parallelize.

larseggert: `has_data_at` controls whether to call `write_frames` at all, and `mark_as_sent` only runs after the stream is already selected. I don't see how it can be moved (or I am missing something)?

martinthomson: I'm not talking about `has_data_at`, but `update_has_data()`.  I think that `mark_as_sent()` is a more natural place to include it.  And it is easier to reason about as a result.

larseggert: But `mark_as_sent` is too late, no? We need to update `has_data` when we buffer data, not after we send it?

martinthomson: Oh, I see the bot response, which is helpful.  However, when data is buffered, can't this set `has_data` unconditionally?

larseggert: Ah, finally I understand what you mean! Yes! (And there may be other such cases...)

larseggert:
OK, I played with this a lot and I don't think there's a performance gain here.

PR #3706, opened 2026-06-13
fix: avoid push id capacity underflow in maybe_send_max_push_id_frame

uwezkhan:
When the peer completes every push it was permitted (push_id 0 through the advertised maximum) before the client sends a MAX_PUSH_ID update, `number_done()` reports one more than `current_max_push_id`. `maybe_send_max_push_id_frame` then evaluates `current_max_push_id - push_done` through `PushId`'s plain `Sub`, which underflows on that input.

Before: debug builds panic on the subtraction; release builds wrap to a near-`u64::MAX` value that fails the `<= max/2` comparison, so no frame is queued and the push id limit stays frozen for the rest of the connection. After: the remaining capacity is computed with a saturating subtraction in a small `push_capacity_low` helper, so the exhausted case reads as zero capacity and the update is sent. The tradeoff is one extra helper over an inline expression; keeping the bound in the controller means the `PushId` arithmetic operators stay unchanged for the other call sites that are already range-checked.

### DIFF
--- a/neqo-http3/src/push_controller.rs
+++ b/neqo-http3/src/push_controller.rs
@@ -404,14 +404,25 @@ impl PushController {
 
     pub fn maybe_send_max_push_id_frame(&mut self, base_handler: &mut Http3Connection) {
         let push_done = self.push_streams.number_done();
-        if self.max_concurent_push > 0
-            && (self.current_max_push_id - push_done) <= (self.max_concurent_push / 2).into()
-        {
+        if self.max_concurent_push > 0 && self.push_capacity_low(push_done) {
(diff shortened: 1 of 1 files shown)


martinthomson (changes requested):
This is not the right fix here.  This is a classic off-by-one error, not a case that should be hidden with a saturating subtraction.

I have one possible fix outstanding, but I'm considering whether changing m`current_max_push_id` from the maximum allowed value to the maximum allowed *count* would be a better way to address this particular problem.

larseggert:
Closing here in favor of #3707. @uwezkhan **thank you** for flagging this!

PR #3707, opened 2026-06-15
Fix off-by-one in max_push_id computation

martinthomson:
This is the code-minimal version of the change.

A more complete change would involve changing the meaning of `max_current_push_id` to `max_push_count` or similar, from which the value sent in the frame would be derived.

That's more intrusive, and we don't really care much about this code, so that's probably fine.

BTW, the concrete consequence of this particular underflow is that we don't send a frame when there are no active push streams, even if we probably should.  That will lock up server push completely, which would be pretty bad if we cared about it.  The catch, thanks to @uwezkhan, is a good one.

Closes #3706.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -5622,7 +5622,8 @@ mod tests {
         send_push_data_and_exchange_packets(&mut client, &mut server, PushId::new(4), true);
         assert_eq!(client.state(), Http3State::Connected);
 
-        // Read push stream with push_id 5 to make it change to closed state.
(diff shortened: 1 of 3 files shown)


martinthomson:
Oh, I should have added: a more complete fix would also demonstrate the effect of the bug by initiating a server push toward a client with a concurrent push limit of 1, have it complete, then fail to receive a MAX_PUSH_ID frame because of the underflow.  That would have been fixed by the change in #3706, but that would also have sent a MAX_PUSH_ID one concluded stream earlier, always.

larseggert:
@martinthomson test failures?

larseggert (approved):

PR #3708, opened 2026-06-15
Fix new clippy lints in Rust 1.98

martinthomson:
I had a few misfires on some other lints, but those stopped coming up. So this might not be the end of the changes for this time around.


larseggert (approved):

martinthomson:
I got those warnings a few times when testing locally, but they didn't reproduce.

PR #3709, opened 2026-06-15
feat: Use `FxHasher` more

larseggert:
There were a few locations where we didn't yet.


martinthomson (approved):

martinthomson:
The name "FxBuildHasher" is pretty unfortunate, but the results seem solid.

PR #3711, opened 2026-06-15
clamp webtransport datagram size to avoid prefix underflow

uwezkhan:
Repro: a peer advertises `max_datagram_frame_size=1` and a WebTransport session lands on a stream id >= 64 (two-byte varint), then the application queries the datagram size.

Cause: both the client (`webtransport_max_datagram_size`) and the server (`WebTransportRequest::max_datagram_size`) take `Connection::max_datagram_size()` and subtract the session-id prefix varint length. The minuend is capped by the peer's `max_datagram_frame_size`, so it can be smaller than the prefix. The old "max varint length is 8" note only bounded the subtrahend.

Fix: `saturating_sub` at both sites.

Before: debug builds panic with subtract-with-overflow; release builds wrap to a near-`u64::MAX` size, so the caller thinks a huge datagram fits and the later send is rejected as `TooMuchData`.
After: returns 0, matching how a session with no usable room is already reported.

Tradeoff: 0 here reads the same as "too small for any payload", but the disabled case still surfaces as the existing `NotAvailable` error, so the only new behavior is the clamp.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -874,8 +874,8 @@ impl Http3Client {
     ///
     /// The function returns `NotAvailable` if datagrams are not enabled.
     pub fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64> {
-        let prefix_len =
-            u64::try_from(Encoder::varint_len(session_id.as_u64())).map_err(|_| Error::Internal)?;
+        let qsid = session_id.as_u64() >> 2;
+        let prefix_len = u64::try_from(Encoder::varint_len(qsid)).map_err(|_| Error::Internal)?;
         Ok(self.conn.max_datagram_size()?.saturating_sub(prefix_len))
     }
 
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -873,14 +873,10 @@ impl Http3Client {
     /// # Errors
     ///
     /// The function returns `NotAvailable` if datagrams are not enabled.
-    ///
-    /// # Panics
-    ///
-    /// This cannot panic. The max varint length is 8.
     pub fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64> {
-        Ok(self.conn.max_datagram_size()?
-            - u64::try_from(Encoder::varint_len(session_id.as_u64()))
-                .map_err(|_| Error::Internal)?)
+        let prefix_len =
+            u64::try_from(Encoder::varint_len(session_id.as_u64())).map_err(|_| Error::Internal)?;
+        Ok(self.conn.max_datagram_size()?.saturating_sub(prefix_len))
     }
 
     /// Sets the `SendOrder` for a given stream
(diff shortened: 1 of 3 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -873,14 +873,10 @@ impl Http3Client {
...
-    ///
-    /// # Panics
-    ///
-    /// This cannot panic. The max varint length is 8.
     pub fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64> {
-        Ok(self.conn.max_datagram_size()?
-            - u64::try_from(Encoder::varint_len(session_id.as_u64()))
-                .map_err(|_| Error::Internal)?)
+        let prefix_len =
+            u64::try_from(Encoder::varint_len(session_id.as_u64())).map_err(|_| Error::Internal)?;
```
martinthomson: Not the fault of this patch, which is fine, but this code is wrong.  It seems like you have a talent for finding those.

The size is determined by the **Quarter** Stream ID (https://www.rfc-editor.org/info/rfc9297/#section-2.1)

```suggestion
        let qsid = session_id.as_u64() >> 2;
        let prefix_len = u64::try_from(Encoder::varint_len(qsid)).map_err(|_| Error::Internal)?;
```

Same for below.

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
```diff
@@ -155,3 +155,25 @@ fn datagrams_multiple_session() {
...
+#[test]
+fn max_datagram_size_smaller_than_session_prefix() {
+    let params = || {
+        wt_default_parameters()
+            .connection_parameters(ConnectionParameters::default().datagram_size(1))
+    };
+    let mut wt = WtTest::new_with_params(params(), params());
+
+    let mut wt_session = wt.create_wt_session();
+    while wt_session.stream_id().as_u64() < 64 {
```
martinthomson:
```suggestion
    while wt_session.stream_id().as_u64() < 256 {
```

See other comments.

neqo-http3/src/server_events.rs
```diff
@@ -371,17 +371,13 @@ impl WebTransportRequest {
...
-    /// This cannot panic. The max varint length is 8.
     pub fn max_datagram_size(&self) -> Res<u64> {
         let max_size = self.stream_handler.conn.borrow().max_datagram_size()?;
-        Ok(max_size
-            - u64::try_from(Encoder::varint_len(
-                self.stream_handler.stream_id().as_u64(),
-            ))
-            .map_err(|_| Error::Internal)?)
+        let prefix_len = u64::try_from(Encoder::varint_len(
+            self.stream_handler.stream_id().as_u64(),
```
martinthomson:
```suggestion
            self.stream_handler.stream_id().as_u64() >> 2,
```

martinthomson (approved):
The bug fix on its own is good.  However, the code itself is bad.  We're overestimating the size of the overheads.  Feel free not to fix that here and we can do a followup.

neqo-http3/src/connection_client.rs
uwezkhan: Good catch. Switched both sites to the quarter stream id (`>> 2`) so the prefix length comes from the value that's actually on the wire. Kept the saturating_sub since the qsid varint can still be 2 bytes while max_datagram_size is 1 (qsid >= 64 means stream id >= 256), so the clamp still does work. Bumped the test loop to 256 and assert on the qsid so it exercises the 2-byte prefix.

martinthomson (approved):

PR #3716, opened 2026-06-16
stats/cc: add new slow_start_exit stats

omansfeld:
This is a first batch of the simplest extra stats to help understand connection patterns for the upload path and make sense of slow start experiment results.

I plan to record in Glean:

- `bytes_in_flight/detection_cwnd` ratio as a histogram labeled on exit_reason (`heuristic` vs `congestion_event`) to get an idea of how full the wire is when we exit slow start
- `detection_cwnd` as a labeled memory distribution, labeled on exit_reason. This should help figure out if heuristics are able to signal at all, i.e. if we see much lower values for loss exits than for heuristic exits then there isn't much for the heuristics to work with in those cases
- `loss_amount` as a histogram can give us an indication if the "early" losses we see are because of congestion or random


neqo-transport/src/cc/classic_cc.rs
```diff
@@ -511,7 +517,14 @@ where
             return false;
         };
 
-        let congestion = self.on_congestion_event(last_lost_packet, Loss, now, cc_stats);
+        let congestion = self.on_congestion_event(
+            last_lost_packet,
+            Loss,
+            bif_before_loss,
+            Some(loss_amount),
```
martinthomson: Why does this need to be an Option?  Zero loss is fine for ECN.  If you wanted to distinguish them properly, you would have:

```rust
enum ExitReason {
  Loss(usize),
  EcnCeMarks(usize),
}
```

neqo-transport/src/stats.rs
```diff
@@ -175,9 +175,21 @@ pub struct CongestionControlStats {
     /// None if we haven't exited slow start or if we re-entered after spurious congestion.
     /// When exiting via congestion event, this is the cwnd AFTER the reduction.
     pub slow_start_exit_cwnd: Option<usize>,
+    /// The congestion window when slow start exit was detected.
+    /// None if we haven't exited slow start or if we re-entered after spurious congestion.
+    /// When exiting via congestion event, this is the cwnd BEFORE the reduction.
+    pub slow_start_exit_detection_cwnd: Option<usize>,
     /// The reason slow start was exited. None if we haven't exited slow start or if we re-entered
     /// after spurious congestion.
     pub slow_start_exit_reason: Option<SlowStartExitReason>,
```
martinthomson: See above comment.  I think that this can be tweaked to include the loss count, the number of CE markings, None, or a heuristic.  I'm almost inclined to suggest a new struct to wrap up all the slow start exit-related stuff here anyway and this takes in the loss amount.  The other fields might also get rolled into that as well.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -478,6 +483,7 @@ where
                 // were sent before the rebinding.
                 self.bytes_in_flight = self.bytes_in_flight.saturating_sub(pkt.len());
                 if !pkt.is_pmtud_probe() {
+                    loss_amount += 1;
```
larseggert: You could do "len(maybe_lost_packets_after) - len(maybe_lost_packets_before)" to get the length instead of +1ing a bunch of times

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -371,10 +371,12 @@ where
                 now,
             ) {
                 qdebug!("Exited slow start by algorithm");
-                self.current.congestion_window = exit_cwnd;
-                self.current.ssthresh = Some(exit_cwnd);
+                cc_stats.slow_start_exit_bytes_in_flight = Some(self.bytes_in_flight);
+                cc_stats.slow_start_exit_detection_cwnd = Some(self.current.congestion_window);
```
larseggert: So you want to log the old window and not exit_cwnd?

neqo-transport/src/cc/classic_cc.rs
omansfeld: We already have the `CongestionTrigger` enum, so I think we could just pass loss-size/ecn-ce-marks using that, in a similar fashion to what you proposed.

neqo-transport/src/stats.rs
omansfeld: I was also almost inclined to make it a struct, so I guess that's what I'll do now :) I think we have enough exit metrics to warrant it.

neqo-transport/src/cc/classic_cc.rs
omansfeld: We already record the `exit_cwnd` in `slow_start_exit_cwnd`. Now I want to additionally record the window size that we detected slow start exit at to have easier reasoning of "the connection reached cwnd xxx before detecting slow start exit" (currently for loss we'd need to revert the `x0.7` to get the real number).

I don't want to change the existing telemetry because I need it to compare to SEARCH's `bdp_estimate` metrics. And if we ever choose to use those then doing the math to get from one to the other gets much harder than a simple `x0.7`.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -371,10 +371,12 @@ where
...
                 cc_stats.slow_start_exit_cwnd = Some(exit_cwnd);
                 cc_stats.slow_start_exit_reason = Some(SlowStartExitReason::Heuristic);
```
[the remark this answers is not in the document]
omansfeld: The new approach with a struct takes care of this. The loss amount is carried by the `Loss(usize)` enum variant, which is part of the exit reason stat. That stat will be set to `Heuristic`, thus not carry old state in the case described above.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -371,10 +371,12 @@ where
...
+                self.current.congestion_window = exit_cwnd;
```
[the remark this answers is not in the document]
omansfeld: The new approach with a struct takes care of this. The loss amount is carried by the `Loss(usize)` enum variant, which is part of the exit reason stat. That stat will be set to `Heuristic`, thus not carry old state in the case described above.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -2028,9 +2057,14 @@ mod tests {
...
         let rtt_estimate = RttEstimate::new(RTT);
 
+        let initial_cwnd = cc.cwnd();
+
         assert!(cc.current.phase.in_slow_start());
         assert_eq!(cc_stats.slow_start_exit_cwnd, None);
         assert_eq!(cc_stats.slow_start_exit_reason, None);
+        assert_eq!(cc_stats.slow_start_exit_bytes_in_flight, None);
+        assert_eq!(cc_stats.slow_start_exit_detection_cwnd, None);
+        assert_eq!(cc_stats.slow_start_exit_loss_amount, None);
```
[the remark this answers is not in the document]
omansfeld: See above, the new struct takes care of the `loss_amount = None` edge case.

mxinden (approved):

PR #3717, opened 2026-06-17
Move WebTransport implementation out of core code

martinthomson:
This means that if you want webtransport, you need to import the trait.

This file includes both the client and server pieces because it isn't that much code.  The net effect is a reduction in code size for other files.

This makes webtransport a little more of a discrete add-on.  So that if you hare looking at the WebTransport API, you only see those functions and the amount that it interacts with the rest of the implementation is reduced (an enum arm for events, some switching based on type).

Note that this integrates the fix from #3711, but not the test that was added.  The fix needs to move, but I think we should accept #3711 first.  Taking the fix makes it easy to resolve conflicts.  And the code was untested (!) before.

### DIFF
--- a/neqo-http3/src/webtransport.rs
+++ b/neqo-http3/src/webtransport.rs
@@ -1,0 +1,437 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::{
+    cell::RefCell,
+    fmt::{self, Display, Formatter},
+    ops::Deref,
+    rc::Rc,
+    time::Instant,
+};
+
+use neqo_common::{Bytes, Encoder, Header, qdebug};
+use neqo_transport::{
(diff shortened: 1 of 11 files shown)


larseggert:
I like it, but how does this intersect with the stack @jesup is trying to land?

neqo-http3/src/webtransport.rs
```diff
@@ -0,0 +1,437 @@
...
+        out: &mut [u8],
+    ) -> Res<()> {
+        self.handler()
+            .validate_extended_connect_session(session_id)?;
+        self.connection()
+            .webtransport_export_keying_material(session_id, label, context, out)
+    }
+}
+
+pub(crate) trait WebTransportExportKeyingMaterial {
```
larseggert: If you named this `ExportKeyingMaterial`, users could refer to it as `webtransport::ExportKeyingMaterial`, which at least to me is more readable. (Also for other types/traits.)

larseggert (approved):

mxinden (commented):
In favor of the architectural change (extension trait pattern).

neqo-http3/src/webtransport.rs
martinthomson: I've gone a fair bit further than that and removed WebTransport from ALL the names, because this is a great idea.

martinthomson:
> In favor of the architectural change (extension trait pattern).

I think that I might do that for connect_udp as well, once this lands.

PR #3720, opened 2026-06-17
fix: reject invalid NEW_CONNECTION_ID retire_prior and cid length

uwezkhan:
Repro: send a NEW_CONNECTION_ID frame whose Retire Prior To exceeds its Sequence Number, or whose connection ID length is zero.

Cause: Frame::decode only rejected connection IDs longer than ConnectionId::MAX_LEN, so both forms decoded fine even though RFC 9000, Section 19.15 requires each to be a FRAME_ENCODING_ERROR.

Fix: reject retire_prior > sequence_number and empty connection IDs in the decoder.

Before, an oversized Retire Prior To reached Paths::retire_cids and could retire the primary path's connection ID without a replacement (tripping the debug_assert in path.rs); a zero-length CID was accepted as a usable remote CID. After, either form is dropped as a connection error during decode. Well-formed frames are unaffected.

### DIFF
--- a/neqo-transport/src/frame.rs
+++ b/neqo-transport/src/frame.rs
@@ -611,8 +611,11 @@ impl<'a> Frame<'a> {
             FrameType::NewConnectionId => {
                 let sequence_number = dv(dec)?;
                 let retire_prior = dv(dec)?;
+                if retire_prior > sequence_number {
+                    return Err(Error::FrameEncoding);
+                }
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
In retrospect, we should probably have defined NEW_CONNECTION_ID with Retire Prior To as a delta, relative to the current sequence number.  Though I guess that would not have eliminated bounds checking in parsing.

PR #3730, opened 2026-06-19
fix: reject max_ack_delay transport parameter of 2^14 or greater

uwezkhan:
RFC 9000, Section 18.2:

    max_ack_delay (0x0b): ... Values of 2^14 or greater are invalid.

`TransportParameter::decode` groups max_ack_delay with the unbounded integer parameters, so a peer can advertise any varint up to 2^62-1 and neqo accepts it. The neighboring arms already bound their values (ack_delay_exponent <= 20, min_ack_delay < 2^24).

Before: max_ack_delay >= 16384 is accepted and used as the peer ACK delay; qlog narrows it to u16, silently truncating.
After: it is rejected with a transport parameter error, matching the sibling checks.

Tradeoff: a non-conforming peer that previously completed the handshake now gets closed during parameter validation, which is the intended behavior per the RFC.


mxinden (approved):
Thank you!

PR #3731, opened 2026-06-20
fix: reject RETIRE_CONNECTION_ID for an unissued sequence number

uwezkhan:
`ConnectionIdManager::retire` acted on the sequence number in a RETIRE_CONNECTION_ID frame without comparing it against the connection IDs we have issued. A peer can put any value there, including one at or beyond `next_seqno`; today that frame is silently ignored. RFC 9000 Section 19.16 says a sequence number greater than any we have sent must be treated as a PROTOCOL_VIOLATION connection error.

Reject the frame in `retire` when the sequence number is not below `next_seqno` and propagate the error from the frame handler. The behavioral change is narrow: a sequence number we never issued now closes the connection instead of being dropped, while every in-range retirement behaves as before. Keeping the bound in the manager, next to the counter it checks, means the single caller can't skip it.

### DIFF
--- a/neqo-transport/src/cid.rs
+++ b/neqo-transport/src/cid.rs
@@ -513,7 +513,14 @@ impl ConnectionIdManager {
         self.connection_ids.contains(cid)
     }
 
-    pub fn retire(&mut self, seqno: u64) {
+    pub fn retire(&mut self, seqno: u64) -> Res<()> {
+        // RFC 9000, Section 19.16: receipt of a RETIRE_CONNECTION_ID frame whose
+        // sequence number is greater than any we have issued (i.e. not below the
+        // next sequence number to be used) is a connection error.
+        if seqno >= self.next_seqno {
+            return Err(Error::ProtocolViolation);
+        }
+
         // TODO(mt) - consider keeping connection IDs around for a short while.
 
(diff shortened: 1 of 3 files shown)


mxinden (commented):
I would lean towards logging (`debug!`) only here. This might be a misbehaving but benign server. Do others feel strongly about closing the connection?

uwezkhan:
I went with closing because 19.16 phrases it as a MUST: "Receipt of a RETIRE_CONNECTION_ID frame containing a sequence number greater than any previously sent to the peer MUST be treated as a connection error of type PROTOCOL_VIOLATION." A debug! and continue leaves us non-conforming on that MUST.

The practical difference is small either way: closing drops a peer that's already off-spec and probably about to misbehave, while logging-only tolerates it. I don't feel strongly about the benign case, so if you'd rather be lenient I'm happy to switch this to debug! only. Just flagging that closing is what the spec asks for.

mxinden:
Thanks. Let's wait for others to comment.

martinthomson (approved):
Yeah, we should do this.  There is a compatibility risk in shipping it, but I'd estimate it being low.  No server should do this and I doubt that any would have.

PR #3732, opened 2026-06-20
fix: reject duplicate transport parameters

uwezkhan:
`TransportParameters::decode` walks the peer's transport parameters and calls `set()` for each, which writes into a per-id slot. A peer that sends the same parameter twice is accepted silently, with the second value overwriting the first.

RFC 9000 Section 7.4 says an endpoint MUST NOT send a parameter more than once and SHOULD treat receipt of a duplicate as a TRANSPORT_PARAMETER_ERROR. The SHOULD is there to let endpoints skip tracking parameters they don't understand (RFC 9413), not for robustness, so a duplicate of a parameter we do parse is rejected.

Before, a repeated id was last-value-wins with no trace. After, the decode loop notices an id it has already parsed, logs a qwarn, and returns `Error::TransportParameter`, closing the handshake.

Tradeoff: a peer that repeats a parameter no longer gets through, which is the stricter reading but matches what Chrome does on duplicates. Added a regression test for the duplicate-rejected case and one confirming distinct parameters still decode.

### DIFF
--- a/neqo-transport/src/tparams.rs
+++ b/neqo-transport/src/tparams.rs
@@ -14,7 +14,7 @@ use std::{
 };
 
 use enum_map::{Enum, EnumMap};
-use neqo_common::{Buffer, Decoder, Encoder, Role, hex, qdebug, qinfo, qtrace};
+use neqo_common::{Buffer, Decoder, Encoder, Role, hex, qdebug, qinfo, qtrace, qwarn};
 use nss::{
     HandshakeMessage, ZeroRttCheckResult, ZeroRttChecker,
     constants::{TLS_HS_CLIENT_HELLO, TLS_HS_ENCRYPTED_EXTENSIONS},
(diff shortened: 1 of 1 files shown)


mxinden (commented):
Instead of an error, I suggest we stay defensive, only logging when we see a duplicate transport parameter, but not closing the connection attempt. In other words, stick with the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). Thoughts?

uwezkhan:
Makes sense, the RFC treatment is a SHOULD so staying lenient is fine. Switched it over: a duplicate now logs a qwarn and keeps the first value instead of returning an error, so the handshake stays up. Tradeoff is we no longer reject a peer that repeats a parameter, but that's the point of the robustness principle. Updated the test to assert the duplicate is ignored and the first value wins.

martinthomson (changes requested):
This one is tougher than #3731.  The SHOULD there exists only to allow implementations to not track transport parameters that they don't understand, more than it does for "robustness" reasons (see also RFC 9413, which addresses that question more fully).  I'd like to see duplicates raise `Error::TransportParameter` if we can.

To help assess that, I checked Chrome first.  And they reject on duplicated transport parameters.  So I think we should also reject here.

mxinden:
@uwezkhan please go with Martin's suggestion above. Sorry for leading you down the wrong path.

uwezkhan:
Sounds good, no worries. Switched it back to raising `Error::TransportParameter` on a duplicate, with a qwarn right before we close so there's a trace. That lines up with the Chrome behavior you found, and the comment now notes the SHOULD is about untracked parameters per RFC 9413 rather than robustness. Tests assert the duplicate is rejected and that distinct parameters still decode.

larseggert (approved):

neqo-transport/src/tparams.rs
```diff
@@ -374,6 +370,20 @@ impl TransportParameters {
...
+                    // > An endpoint MUST NOT send a parameter more than once in a given transport
+                    // > parameters extension. An endpoint SHOULD treat receipt of duplicate
+                    // > transport parameters as a connection error of type
+                    // > TRANSPORT_PARAMETER_ERROR.
+                    //
+                    // The SHOULD only exists to let endpoints skip tracking parameters they do
+                    // not understand (see RFC 9413), not for robustness, so reject a duplicate of
+                    // a parameter we have already parsed.
+                    if tps.params[tipe].is_some() {
+                        qwarn!("Duplicate transport parameter {tipe}");
```
martinthomson: I prefer to keep warning and error log messages for stuff that we think we can avoid.  I don't want to give an attacker the ability to access these.  I can go as high as INFO, but might even prefer a DEBUG level log for this, because it is directly available to peers.

```suggestion
                        qinfo!("Duplicate transport parameter {tipe}");
```

martinthomson (approved):

PR #3733, opened 2026-06-21
reject trailing data in CANCEL_PUSH, GOAWAY, and MAX_PUSH_ID frames

uwezkhan:
CANCEL_PUSH, GOAWAY, and MAX_PUSH_ID each carry a single varint, but `HFrame::decode` reads that varint and returns without looking at the rest of the payload. A peer that appends extra bytes inside the frame length has them silently discarded. RFC 9114 Section 7.1 says a payload with bytes after the identified fields must be a connection error of type H3_FRAME_ERROR.

Before, these three arms accepted the trailing bytes; after, they return `Error::HttpFrame` (0x106) once `decode_varint` leaves anything in the decoder. The check lives in the decoder because that is the only place the frame length and the consumed length are both known; the frames that legitimately have a variable tail (HEADERS, PUSH_PROMISE, the PRIORITY_UPDATE pair) read their remainder and are untouched. The tradeoff is strictness over leniency for a fixed-shape frame, which matches how SETTINGS already errors on a short trailing setting.

### DIFF
--- a/neqo-http3/src/frames/hframe.rs
+++ b/neqo-http3/src/frames/hframe.rs
@@ -174,9 +174,15 @@ impl FrameDecoder<Self> for HFrame {
                 HFrameType::HEADERS => Some(Self::Headers {
                     header_block: dec.decode_remainder().to_vec(),
                 }),
-                HFrameType::CANCEL_PUSH => Some(Self::CancelPush {
-                    push_id: dec.decode_varint().ok_or(Error::HttpFrame)?.into(),
-                }),
+                HFrameType::CANCEL_PUSH => {
+                    let push_id = dec.decode_varint().ok_or(Error::HttpFrame)?;
+                    if dec.remaining() != 0 {
+                        return Err(Error::HttpFrame);
+                    }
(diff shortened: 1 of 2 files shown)


mxinden (approved):
Should we yet again be liberal on what to accept here? On the one hand, ignoring these extra bytes does not directly bring us into an inconsistent state. On the other, the remote might expect us to interpret those bytes, ...

Closing seems fine to me here.

uwezkhan:
The robustness-principle angle crossed my mind too, but 7.1 makes this a MUST rather than a SHOULD: a payload with bytes after the identified fields "MUST be treated as a connection error of type H3_FRAME_ERROR." So it isn't really a be-liberal call here, and the remote can't legitimately expect us to interpret a tail on a fixed-shape frame. It also keeps these three arms consistent with SETTINGS, which already errors on a short trailing setting. The only thing we give up is leniency on frames that never carry a variable tail.

larseggert:
Closing here in favor of #3734. @uwezkhan thanks again for the submission!

PR #3738, opened 2026-06-22
chore: replace usize↔u64 casts with named helpers

### LINKED ISSUE #3737: Need a helper for u64/usize conversion
I'm starting to think that we need a helper for this conversion.  It can be `#[inline(always)]`, but I don't like seeing this same code repeated everywhere.

_Originally posted by @martinthomson in https://github.com/mozilla/neqo/pull/3690#discussion_r3439854197_

larseggert:
Add `to_u64(usize) -> u64` and `to_usize(u64) -> usize` to `neqo-common`, both `const fn`, backed by a compile-time `const_assert_eq!(usize::BITS, u64::BITS)` instead of runtime `try_from().expect()`.

Replace all `try_from().expect("usize fits in u64/usize")` call sites and qualifying `as u64`/`as usize` casts crate-wide.

Convert lossless widening `u8`/`u16`/`u32 → u64`/`usize` casts to `u64::from()`/`usize::from()` where not in const context.

Fixes #3737

### DIFF
--- a/neqo-common/src/lib.rs
+++ b/neqo-common/src/lib.rs
@@ -22,6 +22,7 @@ pub mod tos;
 use std::fmt::Write as _;
 
 use enum_map::Enum;
+use static_assertions::const_assert_eq;
 use strum::Display;
 
 #[cfg(feature = "build-fuzzing-corpus")]
@@ -82,6 +83,30 @@ pub const fn const_max(a: usize, b: usize) -> usize {
 #[must_use]
 pub const fn const_min(a: usize, b: usize) -> usize {
     [a, b][(a > b) as usize]
+}
+
+// Both conversions below are safe on all targets where usize and u64 are the
+// same width (i.e., 64-bit targets). The assertion enforces this at compile time.
+const_assert_eq!(usize::BITS, u64::BITS);
+
+/// Convert a `usize` to `u64`.
+#[expect(clippy::inline_always, reason = "trivial numeric conversion")]
+#[inline(always)]
+#[must_use]
+pub const fn to_u64(v: usize) -> u64 {
+    v as u64
+}
+
+/// Convert a `u64` to `usize`.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "const_assert_eq above ensures usize::BITS == u64::BITS"
+)]
+#[expect(clippy::inline_always, reason = "trivial numeric conversion")]
+#[inline(always)]
+#[must_use]
+pub const fn to_usize(v: u64) -> usize {
+    v as usize
 }
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Enum, Display)]
--- a/neqo-common/src/lib.rs
+++ b/neqo-common/src/lib.rs
@@ -89,11 +89,41 @@ pub const fn const_min(a: usize, b: usize) -> usize {
 // same width (i.e., 64-bit targets). The assertion enforces this at compile time.
 const_assert_eq!(usize::BITS, u64::BITS);
 
+/// A trait for values that represent a length or byte count, convertible
+/// between the wire domain (`u64`) and the in-memory domain (`usize`).
+pub trait Length: Copy {
+    fn as_u64(self) -> u64;
+    fn as_usize(self) -> usize;
+}
+
+impl Length for u64 {
+    fn as_u64(self) -> u64 {
+        self
+    }
+    fn as_usize(self) -> usize {
+        to_usize(self)
+    }
+}
+
+impl Length for usize {
+    fn as_u64(self) -> u64 {
+        to_u64(self)
+    }
+    fn as_usize(self) -> usize {
+        self
+    }
+}
+
 /// Convert a `usize` to `u64`.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "const_assert_eq above ensures usize::BITS == u64::BITS"
+)]
 #[expect(clippy::inline_always, reason = "trivial numeric conversion")]
 #[inline(always)]
 #[must_use]
 pub const fn to_u64(v: usize) -> u64 {
+    debug_assert!(v as u64 as usize == v);
     v as u64
 }
 
@@ -106,6 +136,7 @@ pub const fn to_u64(v: usize) -> u64 {
 #[inline(always)]
 #[must_use]
 pub const fn to_usize(v: u64) -> usize {
+    debug_assert!(v as usize as u64 == v);
     v as usize
 }
 
--- a/neqo-common/src/lib.rs
+++ b/neqo-common/src/lib.rs
@@ -90,18 +90,14 @@ pub const fn const_min(a: usize, b: usize) -> usize {
 const_assert_eq!(usize::BITS, u64::BITS);
 
 /// A trait for values that represent a length or byte count, convertible
-/// between the wire domain (`u64`) and the in-memory domain (`usize`).
+/// to the wire domain (`u64`).
 pub trait Length: Copy {
     fn as_u64(self) -> u64;
-    fn as_usize(self) -> usize;
 }
 
 impl Length for u64 {
     fn as_u64(self) -> u64 {
         self
-    }
-    fn as_usize(self) -> usize {
-        to_usize(self)
     }
 }
 
@@ -109,18 +105,15 @@ impl Length for usize {
     fn as_u64(self) -> u64 {
         to_u64(self)
     }
-    fn as_usize(self) -> usize {
-        self
-    }
 }
 
 /// Convert a `usize` to `u64`.
 #[expect(
     clippy::cast_possible_truncation,
-    reason = "const_assert_eq above ensures usize::BITS == u64::BITS"
+    reason = "debug_assert roundtrip `v as u64 as usize` contains a u64→usize cast; \
+              const_assert_eq above ensures it is lossless on all supported targets"
 )]
-#[expect(clippy::inline_always, reason = "trivial numeric conversion")]
-#[inline(always)]
+#[inline]
 #[must_use]
 pub const fn to_u64(v: usize) -> u64 {
     debug_assert!(v as u64 as usize == v);
@@ -132,8 +125,7 @@ pub const fn to_u64(v: usize) -> u64 {
     clippy::cast_possible_truncation,
     reason = "const_assert_eq above ensures usize::BITS == u64::BITS"
 )]
-#[expect(clippy::inline_always, reason = "trivial numeric conversion")]
-#[inline(always)]
+#[inline]
 #[must_use]
 pub const fn to_usize(v: u64) -> usize {
     debug_assert!(v as usize as u64 == v);
--- a/neqo-qpack/src/huffman_decode_helper.rs
+++ b/neqo-qpack/src/huffman_decode_helper.rs
@@ -6,6 +6,8 @@
 
 use std::sync::OnceLock;
 
+use neqo_common::to_usize;
+
 use crate::huffman_table::HUFFMAN_TABLE;
 
 // Since we're encoding the table length as a u16, we need to ensure that it fits.
@@ -35,7 +37,7 @@ fn make_huffman_tree(prefix: u32, len: u8) -> HuffmanDecoderNode {
         found = true;
         if iter.len == len + 1 {
             // This is a leaf
-            let bit = usize::try_from(iter.val & 1).expect("u32 fits in usize");
+            let bit = to_usize(u64::from(iter.val & 1));
             next[bit] = Some(Box::new(HuffmanDecoderNode {
                 next: [None, None],
                 #[expect(
--- a/neqo-transport/benches/range_tracker.rs
+++ b/neqo-transport/benches/range_tracker.rs
@@ -12,6 +12,7 @@
 use std::hint::black_box;
 
 use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use neqo_common::to_u64;
 use neqo_transport::send_stream::RangeTracker;
 
 const CHUNK: usize = 1000;
@@ -20,11 +21,11 @@ fn build_coalesce(count: usize) -> RangeTracker {
     let mut used = RangeTracker::default();
     used.mark_acked(0, CHUNK); // frontier at CHUNK
     // One big Sent range covering exactly the gap and all the acked chunks placed below.
-    used.mark_sent(CHUNK as u64, 2 * count * CHUNK);
+    used.mark_sent(to_u64(CHUNK), 2 * count * CHUNK);
     // ACK every *other* chunk so the acked ranges stay separate (a Sent chunk between
     // each prevents merging); the gap at [CHUNK, 2*CHUNK) keeps the frontier blocked.
     for i in 1..=count {
-        used.mark_acked((2 * i * CHUNK) as u64, CHUNK);
+        used.mark_acked(to_u64(2 * i * CHUNK), CHUNK);
     }
     used
 }
@@ -36,7 +37,7 @@ fn coalesce(c: &mut Criterion, count: usize) {
             // Fill the gap and jump the frontier past all `count` acked ranges in one
             // call; coalesce_acked then walks every entry below the new frontier.
             |used: &mut RangeTracker| {
-                used.mark_acked(CHUNK as u64, 2 * count * CHUNK);
+                used.mark_acked(to_u64(CHUNK), 2 * count * CHUNK);
                 black_box(used);
             },
             BatchSize::SmallInput,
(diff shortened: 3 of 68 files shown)


neqo-common/src/lib.rs
```diff
@@ -84,6 +85,30 @@ pub const fn const_min(a: usize, b: usize) -> usize {
     [a, b][(a > b) as usize]
 }
 
+// Both conversions below are safe on all targets where usize and u64 are the
+// same width (i.e., 64-bit targets). The assertion enforces this at compile time.
+const_assert_eq!(usize::BITS, u64::BITS);
```
martinthomson, line 90: "The size of this primitive is how many bytes it takes to reference any location in memory. "  I understand that `size_t` on Windows, even 64-bit Windows, is 32 bits.  I guess that Rust is less prone to that particular peculiarity.

neqo-common/src/lib.rs
```diff
@@ -84,6 +85,30 @@ pub const fn const_min(a: usize, b: usize) -> usize {
...
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "const_assert_eq above ensures usize::BITS == u64::BITS"
+)]
+#[expect(clippy::inline_always, reason = "trivial numeric conversion")]
+#[inline(always)]
+#[must_use]
+pub const fn to_usize(v: u64) -> usize {
+    v as usize
+}
```
martinthomson, line 133: I think that what we're really looking for is for functions that clearly indicate that the item in question carries a size or length, it just happens to be in a form that needs adapting.

To that end, I would suggest these be named "len_as_u64" and "len_as_usize" or something like that, to clearly signal intent.  A straight cast is not appropriate in many cases.

I would also appreciate a `debug_assert!()` on each so that we can catch problems with the cast.

neqo-common/src/lib.rs
```diff
@@ -22,6 +22,7 @@ pub mod tos;
 use std::fmt::Write as _;
 
 use enum_map::Enum;
+use static_assertions::const_assert_eq;
```
martinthomson, line 25: Do we need the new dependency when the `const _: () = assert!()` pattern works?

martinthomson (commented):
I wonder whether we can streamline this a little.

There are a number of places where common functions could take responsibility for doing the conversion.

Consider `Encoder`:

```rust
fn encode_len(&mut self, v: impl Length) {
   self.encode_varint(v.as_u64());
}
```

Which could use a trait like:

```rust
pub trait Length: Copy {
  fn as_u64(self) -> u64;
  fn as_usize(self) -> usize;
}

// Sadly, this can't be for T: Into<u64>; we'll need extra impls for each primitive type.
impl Length for u64  {  
  fn as_u64(self) -> u64 { self }
  fn as_usize(self) -> usize { self as usize }  
}
impl Length for usize {
  fn as_u64(self) -> u64 { self as u64 }
  fn as_usize(self) -> usize { self  }  
}
```

neqo-common/src/lib.rs
larseggert: `static_assertions` is a neqo/Gecko dep already. Are you saying we should drop it and go with the pure Rust parttern?

neqo-common/src/lib.rs
larseggert: But we're not converting lengths everywhere. It's a length conversion in maybe half the cases, so renaming to `len_as_<...>` is sensible there. For the other cases, we'd need to leave it as is?

(Or do you want me to the trait when it's a length and leave the other ones unchanged?)

martinthomson: I was trying to reduce the number of call sites that needed explicit conversion, not eliminate them entirely.

I do think that we should be using explicit conversions when things are not lengths, but the general exception to the need for a conversion should be that we already have to track a count in our code, ideally associated with us holding something in memory or something like that.

Of course, if sizeof(usize) == sizeof(u64), maybe this is far more academic.

neqo-common/src/lib.rs
martinthomson: If we are using it already, that's fine.  I thought that the Cargo.lock changes were the direct result of you adding, but I guess that was something else.

larseggert: Cargo.lock changed, because we now use the crate in neqo-common (it's in use elsewhere already).

neqo-transport/src/cc/search.rs
```diff
@@ -217,8 +217,10 @@ impl Search {
     /// Computes delivered bytes between two bin indices. Widens the result to `u64` to avoid
     /// saturation or overflow further down the line on 32-bit systems with large bandwidths.
     const fn compute_delv(&self, old: usize, new: usize) -> u64 {
-        self.acked_bins[new % Self::NUM_ACKED_BINS]
-            .saturating_sub(self.acked_bins[old % Self::NUM_ACKED_BINS]) as u64
+        to_u64(
+            self.acked_bins[new % Self::NUM_ACKED_BINS]
```
martinthomson: This looks like a case where we should be tracking `u64` rather than `usize`.

martinthomson (commented):
Looking through this, I think that virtually all of these are directly related to lengths or offets into memory that we hold.  So the length thing mostly holds.

There are few exceptions.  I see HPACK Huffman indexing that could use `From` directly; removed parentheses in `network.rs`; and the one comment in SEARCH that indicates that maybe that code should be using `u64`.

Many of these could avoid casting at callsites by using the `Length` trait I mentioned on a select few functions: the encoder, the `mark_as_*` functions, and maybe a couple of others.

neqo-common/src/lib.rs
```diff
@@ -84,6 +85,53 @@ pub const fn const_min(a: usize, b: usize) -> usize {
...
+/// Convert a `usize` to `u64`.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "debug_assert roundtrip `v as u64 as usize` contains a u64→usize cast; \
+              const_assert_eq above ensures it is lossless on all supported targets"
+)]
+#[inline]
+#[must_use]
+pub const fn to_u64(v: usize) -> u64 {
+    debug_assert!(v as u64 as usize == v);
```
martinthomson, line 119:
```suggestion
    debug_assert_eq!(v as u64 as usize, v);
```

?

neqo-qpack/src/huffman_decode_helper.rs
```diff
@@ -35,7 +37,7 @@ fn make_huffman_tree(prefix: u32, len: u8) -> HuffmanDecoderNode {
         found = true;
         if iter.len == len + 1 {
             // This is a leaf
-            let bit = usize::try_from(iter.val & 1).expect("u32 fits in usize");
+            let bit = to_usize(u64::from(iter.val & 1));
```
martinthomson, line 40: Just flagging that this is a rare instance of having to call `u64::from()` with `to_usize()`.  As such, that means that the following is not necessary:

```rust
pub fn to_usize<T>(v: T) -> usize where T: Copy, u64: From<T> {
    debug_assert_eq!(u64::from(v) as usize as u64, u64::from(v));
    u64::from(v) as usize
}
```

neqo-transport/benches/range_tracker.rs
```diff
@@ -20,11 +21,11 @@ fn build_coalesce(count: usize) -> RangeTracker {
     let mut used = RangeTracker::default();
     used.mark_acked(0, CHUNK); // frontier at CHUNK
     // One big Sent range covering exactly the gap and all the acked chunks placed below.
-    used.mark_sent(CHUNK as u64, 2 * count * CHUNK);
+    used.mark_sent(to_u64(CHUNK), 2 * count * CHUNK);
```
martinthomson, line 24: Does it make sense to use `Length` in the interface to `mark_sent()`/`mark_acked()`/etc...?

neqo-transport/src/cc/tests/hystart.rs
```diff
@@ -46,7 +46,7 @@ fn maybe_enter_css(
     cc_stats: &mut CongestionControlStats,
 ) {
     // First round with base RTT
-    let window_end = HyStart::N_RTT_SAMPLE as u64;
+    let window_end = to_u64(HyStart::N_RTT_SAMPLE);
```
martinthomson: I think that `N_RTT_SAMPLE` can be made a `u64`.  The same applies to `HyStart::rtt_sample_count`.  That would eliminate all conversions relating to these values.

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -277,7 +278,7 @@ fn congestion_avoidance_no_two_mss_cap() {
     let n = 3 * (cwnd0 / mtu) + 1;
     let mut pkts = Vec::with_capacity(n);
     for pn in 0..n {
```
martinthomson:
```suggestion
    for pn in 0..to_u64(n) {
```

neqo-transport/src/cc/tests/new_reno.rs
```diff
@@ -277,7 +278,7 @@ fn congestion_avoidance_no_two_mss_cap() {
...
-        let p = sent::make_packet(u64::try_from(pn).unwrap(), now, mtu);
+        let p = sent::make_packet(to_u64(pn), now, mtu);
```
martinthomson:
```suggestion
        let p = sent::make_packet(pn, now, mtu);
```

neqo-transport/src/cc/tests/hystart.rs
martinthomson: I think that `N_RTT_SAMPLE` can be made a `u64`.  The same applies to `HyStart::rtt_sample_count`.  That would eliminate all conversions relating to these values.

neqo-transport/src/cc/tests/search.rs
```diff
@@ -564,7 +565,7 @@ fn no_sent_bytes() {
     // Verify the stat is recorded through the full on_packets_acked path.
     let mut cc_stats = CongestionControlStats::default();
     now += bin_duration + Duration::from_nanos(1);
-    search.record_acked_bytes(0);
+    search.record_acked_bytes(0_usize);
```
martinthomson: This change doesn't seem necessary; but what do I know.

neqo-transport/src/cc/classic_cc.rs
```diff
@@ -1013,7 +1013,7 @@ mod tests {
     const SUB_PC: Duration = Duration::from_millis(100 * PERSISTENT_CONG_THRESH as u64);
     /// The minimum time between packets to cause persistent congestion.
     /// Uses an odd expression because `Duration` arithmetic isn't `const`.
-    const PC: Duration = Duration::from_nanos(100_000_000 * (PERSISTENT_CONG_THRESH as u64) + 1);
+    const PC: Duration = Duration::from_nanos(100_000_000 * PERSISTENT_CONG_THRESH as u64 + 1);
```
martinthomson: These can't use `to_u64()`?  It's `const`.

neqo-transport/src/cc/search.rs
```diff
@@ -214,30 +214,25 @@ impl Search {
...
-            .saturating_sub(self.sent_bins[old % Self::NUM_SENT_BINS]))
-            as u64;
-        let sent = low_idx * fraction + high_idx * (Self::SCALE as u64 - fraction);
-        sent / Self::SCALE as u64
+        let low_idx = self.sent_bins[(new - 1) % Self::NUM_SENT_BINS]
+            .saturating_sub(self.sent_bins[(old - 1) % Self::NUM_SENT_BINS]);
+        let high_idx = self.sent_bins[new % Self::NUM_SENT_BINS]
+            .saturating_sub(self.sent_bins[old % Self::NUM_SENT_BINS]);
+        let sent = low_idx * fraction + high_idx * (to_u64(Self::SCALE) - fraction);
+        sent / to_u64(Self::SCALE)
```
martinthomson: `Search::SCALE` is a mess.  It's a `usize` that is never used as one.  It is anything between `u32` and `u128`.  I think that it makes sense to make it `u32`, but it might need more surgery than that.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -531,7 +531,7 @@ fn too_many_datagram_events() {
 
     let mut client = new_client(
         ConnectionParameters::default()
-            .datagram_size(u64::try_from(DATA_SIZE).unwrap())
+            .datagram_size(to_u64(DATA_SIZE))
```
martinthomson, line 534: A case for `Length`?

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -1069,9 +1065,7 @@ fn session_flow_control_stop_sending_state_recv() {
     const SMALL_MAX_DATA: usize = 1024;
 
     let mut client = default_client();
-    let mut server = new_server(
-        ConnectionParameters::default().max_data(u64::try_from(SMALL_MAX_DATA).unwrap()),
-    );
+    let mut server = new_server(ConnectionParameters::default().max_data(to_u64(SMALL_MAX_DATA)));
```
martinthomson, line 1109: As above, maybe `ConnectionParameters` can use `Length` where it makes sense.

neqo-transport/src/addr_valid.rs
```diff
@@ -208,7 +208,7 @@ impl AddressValidation {
             .zip(TOKEN_IDENTIFIER_RETRY.iter())
             .map(|(a, b)| (a ^ b).count_ones())
             .sum();
-        usize::try_from(difference).expect("u32 fits in usize") < TOKEN_IDENTIFIER_RETRY.len()
+        (difference as usize) < TOKEN_IDENTIFIER_RETRY.len()
```
martinthomson:
```suggestion
        to_usize(u64::from(difference)) < TOKEN_IDENTIFIER_RETRY.len()
```
Missed one.

neqo-transport/src/fc.rs
```diff
@@ -116,14 +116,14 @@ where
 
     /// Consume flow control.
     pub fn consume(&mut self, count: usize) {
-        let amt = u64::try_from(count).expect("usize fits into u64");
+        let amt = to_u64(count);
```
martinthomson, line 119: Prime candidate for `Length`.

neqo-transport/src/recv_stream.rs
```diff
@@ -1286,9 +1280,9 @@ mod tests {
 
         // Add three chunks.
         s.inbound_frame(0, &[0; CHUNK_SIZE]);
-        let offset = u64::try_from(CHUNK_SIZE).unwrap();
+        let offset = to_u64(CHUNK_SIZE);
         s.inbound_frame(offset, &[0; EXTRA_SIZE]);
-        let offset = u64::try_from(CHUNK_SIZE + EXTRA_SIZE).unwrap();
+        let offset = to_u64(CHUNK_SIZE + EXTRA_SIZE);
         s.inbound_frame(offset, &[0; EXTRA_SIZE]);
```
martinthomson, line 1286: This looks like another case for `Length`.

neqo-transport/src/send_stream.rs
```diff
@@ -198,7 +198,7 @@ impl RangeTracker {
         reason = "OK here."
     )]
     pub fn mark_acked(&mut self, new_off: u64, new_len: usize) {
-        let end = new_off + u64::try_from(new_len).expect("usize fits in u64");
+        let end = new_off + to_u64(new_len);
```
martinthomson, line 201: `Length` again?

martinthomson (approved):
This is good.  I would take this without any of the suggestions, some of which can be handled in followups.

It strikes me though that we should avoid casting if the variable is only ever used as a `Length`.  Otherwise, there could be two conversions.

neqo-common/src/lib.rs
larseggert: `debug_assert_eq` doesn't work in `const` because it uses `Debug` (I had that first.)

neqo-transport/benches/range_tracker.rs
larseggert: I tried that, but it requires adding a bunch of `_u64` prefixes to test integer literals, because Rust can't auto-determine the type for an `impl` parameter and so defaults to `i32`.

I can make the change, but just be prepared it will be noisy.

neqo-common/src/lib.rs
martinthomson: Of course.  I should have known that.  It relies on `PartialEq + Debug` and both are traits (const traits are not stabilized).

neqo-transport/benches/range_tracker.rs
martinthomson: OK, I thought it might be easier, but if it isn't, we can think about doing it later.

neqo-transport/src/connection/tests/datagram.rs
larseggert: `datagram_size` is `const`; `impl Trait` in `const fn` requires nightly

neqo-qpack/src/huffman_decode_helper.rs
larseggert: We have some callers that require `to_usize` to be `const`.

neqo-transport/src/fc.rs
larseggert: Same issue wrt `i32` as above. If we made this change, we need to make changes such as

```rust
fc.consume(1_u64)
```

everywhere we call it with a literal.

PR #3743, opened 2026-06-23
fix: reject stream frames for the wrong unidirectional stream half

uwezkhan:
`Streams::input_frame` already closes the connection with `STREAM_STATE_ERROR` when a `STREAM_DATA_BLOCKED` frame arrives for a send-only stream (RFC 9000, 19.13), but the four sibling per-stream frames skip that check. Once the targeted half-stream exists, a `RESET_STREAM` or `STREAM` frame for a send-only stream, or a `STOP_SENDING` or `MAX_STREAM_DATA` frame for a receive-only stream, is accepted and silently dropped rather than rejected.

Before, only `STREAM_DATA_BLOCKED` carried the directional guard; after, the other arms apply the same `is_send_only`/`is_recv_only` test before touching the stream. The check belongs in `input_frame` because that is where a frame is dispatched against the connection role, and sections 19.4, 19.5, 19.8, and 19.10 each make the wrong-direction case a connection error rather than a per-stream concern. The tradeoff is that a peer which previously got these frames ignored now sees the connection closed, matching what neqo already does for `STREAM_DATA_BLOCKED`.

### DIFF
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -566,6 +566,36 @@ fn illegal_stream_related_frames() {
 }
 
 #[test]
+/// Server sends a stream-related frame for the wrong half of an existing
+/// unidirectional stream. This should cause the client to close the connection
+/// with `STREAM_STATE_ERROR`.
+fn wrong_directional_stream_frames() {
+    // Frames a sender may not receive. A send-only stream is client-initiated
+    // unidirectional (id 2); it has to exist so the directional check is what
+    // closes the connection rather than the not-yet-created check.
+    for frame_type in [FrameType::ResetStream, FrameType::Stream] {
+        let mut client = default_client();
+        let mut server = default_server();
+        connect(&mut client, &mut server);
+        assert_eq!(client.stream_create(StreamType::UniDi).unwrap(), 2);
+        let dgram = send_with_extra(&mut server, Writer(vec![frame_type.into(), 2, 0, 0]), now());
+        client.process_input(dgram, now());
+        assert!(client.state().closed());
+    }
+
+    // Frames a receiver may not receive. A receive-only stream is
+    // server-initiated unidirectional (id 3); obtain_stream creates it.
+    for frame_type in [FrameType::StopSending, FrameType::MaxStreamData] {
+        let mut client = default_client();
+        let mut server = default_server();
+        connect(&mut client, &mut server);
+        let dgram = send_with_extra(&mut server, Writer(vec![frame_type.into(), 3, 0, 0]), now());
+        client.process_input(dgram, now());
+        assert!(client.state().closed());
+    }
+}
+
+#[test]
 /// Regression <https://github.com/mozilla/neqo/pull/2358>.
 fn legal_out_of_order_frame_on_remote_initiated_closed_stream() {
     const REQUEST: &[u8] = b"ping";
--- a/neqo-transport/src/connection/tests/stream.rs
+++ b/neqo-transport/src/connection/tests/stream.rs
@@ -6,7 +6,7 @@
 
 use std::{cmp::max, collections::HashMap, fmt::Debug};
 
-use neqo_common::{event::Provider as _, qdebug};
+use neqo_common::{Role, event::Provider as _, qdebug};
 use test_fixture::now;
 
 use super::{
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/tests/stream.rs
```diff
@@ -565,6 +565,36 @@ fn illegal_stream_related_frames() {
...
+        assert!(client.state().closed());
+    }
+
+    // Frames a receiver may not receive. A receive-only stream is
+    // server-initiated unidirectional (id 3); obtain_stream creates it.
+    for frame_type in [FrameType::StopSending, FrameType::MaxStreamData] {
+        let mut client = default_client();
+        let mut server = default_server();
+        connect(&mut client, &mut server);
+        let dgram = send_with_extra(&mut server, Writer(vec![frame_type.into(), 3, 0, 0]), now());
```
martinthomson:
```suggestion
        assert_eq!(server.stream_create(StreamType::UniDi).unwrap(), 3);
        let dgram = send_with_extra(&mut server, Writer(vec![frame_type.into(), 3, 0, 0]), now());
```

The stream creation check isn't necessary, but if you are going to do it once...

Or, take the next suggestion.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -565,6 +565,36 @@ fn illegal_stream_related_frames() {
...
+        client.process_input(dgram, now());
+        assert!(client.state().closed());
```
martinthomson: It occurs to me that this is almost entirely duplicative of the code in the block above.

```rust
fn client_rejects(frame_type: FrameType, stream_creator: Role) {
    connect(&mut client, &mut server);
    let creator = match stream_creator { Role::Client => client, Role::Server => server };
    let stream_id = u8::try_from(creator.stream_create(StreamType::UniDi).unwrap()).unwrap();
    let dgram = send_with_extra(&mut server, Writer(vec![u8::from(frame_type), stream_id, 0, 0]), now());
    client.process_input(dgram, now());
    assert!(matches!(
        client.state(),
        State::Closed(CloseReason::Transport(Error::StreamState))
    ));
}
```

martinthomson (approved):
This is a pretty embarrassing bug.  I wonder whether we lost something at some point.

neqo-transport/src/connection/tests/stream.rs
```diff
@@ -565,6 +565,36 @@ fn illegal_stream_related_frames() {
...
 
+#[test]
+/// Server sends a stream-related frame for the wrong half of an existing
+/// unidirectional stream. This should cause the client to close the connection
+/// with `STREAM_STATE_ERROR`.
+fn wrong_directional_stream_frames() {
+    // Frames a sender may not receive. A send-only stream is client-initiated
+    // unidirectional (id 2); it has to exist so the directional check is what
+    // closes the connection rather than the not-yet-created check.
+    for frame_type in [FrameType::ResetStream, FrameType::Stream] {
```
martinthomson: One thing I forgot: it would be good to include STREAM_DATA_BLOCKED here, even though it is already handled.  If that means adding more 0s to the frame that the test writes, that's safe to do.

neqo-transport/src/connection/tests/stream.rs
uwezkhan: Done. Folded both directions into a client_rejects(frame_type, stream_creator) helper and dropped the duplicate block. One deviation from the sketch: right after the local close the state is Closing, not Closed, so matches!(.., State::Closed(..)) never holds. Used the existing assert_error(&client, &CloseReason::Transport(Error::StreamState)) instead, which still pins the exact reason.

neqo-transport/src/connection/tests/stream.rs
uwezkhan: Added STREAM_DATA_BLOCKED to the send-only group. The 4-byte frame already covers it (type, stream id, then 0s as PADDING), so no extra bytes were needed.

neqo-transport/src/connection/tests/stream.rs
uwezkhan: Took the helper route, so the creating role makes the stream in one place now and the explicit id check is gone.

martinthomson (approved):
Perfect!  Thanks again for the good work here.

PR #3744, opened 2026-06-24
Simplify use of datagram modifiers in test fixtures

martinthomson:
There is only one or two places where dropping is possible.  That allowed for most modifier functions to remove the `Some()` wrapper.

The option form is kept for the innermost handshake, where the `None` option is genuinely used (in two tests, which did make me wonder if the cost is worth it, though the cost is small, so I kept it).

The migration tests also depend on being able to drop on the new path, but the complexity can be contained to those tests.

### DIFF
--- a/neqo-transport/src/connection/tests/ecn.rs
+++ b/neqo-transport/src/connection/tests/ecn.rs
@@ -332,7 +332,7 @@ pub fn migration_with_modifiers(
         .unwrap();
 
     let mut migrated = false;
-    let probe = client.process_output(now).dgram().unwrap();
+    let probe = new_path_modifier(client.process_output(now).dgram().unwrap());
     if !drop_new_path {
         assert_v4_path(&probe, true); // Contains PATH_CHALLENGE.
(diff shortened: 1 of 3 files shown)


neqo-transport/src/connection/tests/ecn.rs
```diff
@@ -445,86 +434,98 @@ pub fn migration_with_modifiers(
...
-        for new_path_mod in [noop, bleach, remark, ce, drop] {
+    for orig_path_mod in [identity, bleach, remark, ce] {
+        for new_path_mod in [identity, bleach, remark, ce] {
             let (before, after, migrated) =
-                migration_with_modifiers(orig_path_mod, new_path_mod, 0);
+                migration_with_modifiers(orig_path_mod, new_path_mod, false, 0);
             // Too few packets sent before and after migration to conclude ECN validation.
             assert_ecn_enabled(before);
             assert_ecn_enabled(after);
             // Migration succeeds except if the new path drops ECN.
```
martinthomson:
```suggestion
```

larseggert (approved):

PR #3745, opened 2026-06-24
fix: reject DATAGRAM frame when max_datagram_frame_size is 0

uwezkhan:
When an endpoint advertises a `max_datagram_frame_size` of 0 it has no DATAGRAM support, but `QuicDatagrams::handle_datagram` only compares the received payload length against `local_datagram_size`. With that limit at 0 the comparison rejects any non-empty frame, so an empty DATAGRAM frame is the one case that slips through: it is delivered to the application and the connection stays open. RFC 9221 Section 3 makes receipt of a DATAGRAM frame after advertising a max_datagram_frame_size of 0 a connection error of type PROTOCOL_VIOLATION.

Reject the frame when datagrams were never negotiated. The check lives in `handle_datagram` because that is the only place that sees both the received frame and the limit we advertised, and it lines up with the oversized-frame rejection already there. Before, an empty datagram against a disabled endpoint was silently accepted; after, it closes the connection, which is stricter but matches the RFC.

### DIFF
--- a/neqo-transport/src/connection/tests/datagram.rs
+++ b/neqo-transport/src/connection/tests/datagram.rs
@@ -54,7 +54,7 @@ impl crate::connection::test_internal::FrameWriter for InsertDatagram<'_> {
     }
 }
 
-struct InsertEmptyLenDatagram;
+struct InsertEmptyDatagram;
 
 impl crate::connection::test_internal::FrameWriter for InsertEmptyDatagram {
     fn write_frames(&mut self, builder: &mut packet::Builder<&mut Vec<u8>>) {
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -54,6 +54,15 @@ impl crate::connection::test_internal::FrameWriter for InsertDatagram<'_> {
     }
 }
 
+struct InsertEmptyLenDatagram;
+
+impl crate::connection::test_internal::FrameWriter for InsertEmptyLenDatagram {
```
martinthomson:
```suggestion
impl crate::connection::test_internal::FrameWriter for InsertEmptyDatagram {
```

martinthomson (approved):
I'm thinking that we should probably not advertise the MAX_DATAGRAM_SIZE transport parameter if the size is configured at 0.  I'll pick that one up though.

PR #3749, opened 2026-06-24
fix: reject reserved bits set in QUIC packet headers

uwezkhan:
The two reserved bits in a packet's first byte (0x0c for long headers, 0x18 for short headers) are header protected, so their real values only appear after `Public::decrypt` removes both header and packet protection. `decrypt` reads the first byte to recover the packet number length and key phase but never looks at the reserved bits, so a peer that authenticates a packet with either bit set is accepted. RFC 9000 17.2 and 17.3.1 require that case to be a connection error of type PROTOCOL_VIOLATION.

Before: the reserved bits were ignored on every received packet. After: `decrypt` checks them once the packet is authenticated and returns PROTOCOL_VIOLATION, which the receive loop now treats as fatal. The check runs after AEAD on purpose, so an off-path attacker flipping the protected bits only causes a decryption failure that gets dropped, while a genuine peer violation closes the connection. The tradeoff is that we no longer silently tolerate a peer that sets bits the spec keeps reserved.

### DIFF
--- a/neqo-transport/src/packet/mod.rs
+++ b/neqo-transport/src/packet/mod.rs
@@ -1350,8 +1350,10 @@ mod tests {
     /// A packet that authenticates but has a reserved bit set in the first byte
     /// is a connection error of type `PROTOCOL_VIOLATION`. The bit is set before
     /// the packet is encrypted so that it is covered by the AEAD and survives
-    /// header protection removal at the receiver.
-    fn reject_reserved_bit(builder: Builder<Vec<u8>>) {
+    /// header protection removal at the receiver. Each reserved bit is checked
+    /// on its own because the long and short header masks differ.
+    fn reject_reserved_bit(mut builder: Builder<Vec<u8>>, bit: u8) {
+        builder.as_mut()[0] |= bit;
         let packet = builder
             .build(&mut CryptoDxState::test_default_write())
             .expect("build");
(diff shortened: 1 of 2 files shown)


larseggert:
IIRC we're deliberately not checking these bits.

neqo-transport/src/packet/mod.rs
```diff
@@ -1332,6 +1347,55 @@ mod tests {
...
+            Type::Initial,
+            Version::default(),
+            None::<&[u8]>,
+            Some(ConnectionId::from(SERVER_CID)),
+            packet::LIMIT,
+        );
+        builder.initial_token(&[]);
+        builder.pn(0, 1);
+        builder.encode(SAMPLE_INITIAL_PAYLOAD);
+        builder.as_mut()[0] |= 0x08;
```
martinthomson: You could move this line into the function, noting that both short and long trigger on this bit.

However, I wonder whether this should also check the other bits as well; after all, the code is different for those bits.

martinthomson (approved):
I think we should take this, though maybe the test coverage could be a tiny bit more complete.

neqo-transport/src/packet/mod.rs
uwezkhan: Done. Moved the bit-setting into the helper and made it take the bit, so each test now exercises both reserved bits for its header form (0x10 and 0x08 short, 0x08 and 0x04 long) rather than just the shared 0x08.

PR #3750, opened 2026-06-24
chore: Update qlog to 0.18.0

larseggert:
And deal with the fallout.

Also bump smallvec to what Gecko now uses while I'm here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,8 +331,18 @@ version = "0.20.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6f63b86c8a8826a49b8c21f08a2d07338eec8d900540f8630dc76284be802989"
 dependencies = [
- "darling_core",
- "darling_macro",
+ "darling_core 0.20.10",
+ "darling_macro 0.20.10",
+]
+
+[[package]]
+name = "darling"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
+dependencies = [
+ "darling_core 0.23.0",
+ "darling_macro 0.23.0",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


Cargo.lock
```diff
@@ -331,8 +331,18 @@ version = "0.20.10"
...
 dependencies = [
- "darling_core",
- "darling_macro",
+ "darling_core 0.20.10",
+ "darling_macro 0.20.10",
+]
+
+[[package]]
+name = "darling"
+version = "0.23.0"
```
martinthomson, line 340: Two versions at the same is going to work out great.  Are you planning to add this to gecko as well?

Cargo.lock
```diff
@@ -601,6 +634,12 @@ dependencies = [
  "itoa",
 ]
 
+[[package]]
+name = "humantime"
```
martinthomson, line 638: New dependency!

Cargo.lock
```diff
@@ -1104,22 +1143,21 @@ dependencies = [
 
 [[package]]
 name = "serde_with"
-version = "3.12.0"
+version = "3.21.0"
```
martinthomson, line 1146: Pretty significant bump here.

Cargo.lock
```diff
@@ -1161,12 +1199,9 @@ checksum = "f8fadd59c855ef2080decdef8ff161eb6661b86933c9d82e5ba29dc602a55aba"
 
 [[package]]
 name = "smallvec"
-version = "1.15.1"
+version = "1.15.2"
```
martinthomson, line 1202: Matches gecko ✅

test/qvis.py
```diff
@@ -149,7 +149,9 @@ def _append_stream_budget(sid: int, t: float) -> int:
     def _stream_frame(fr: dict[str, Any]) -> tuple[int, int, int]:
         sid = int(fr.get("stream_id", 0))
         off = int(fr.get("offset", 0))
-        return sid, off, off + int(fr.get("length") or 0)
+        raw = fr.get("raw") or {}
```
martinthomson, line 152: Not specific to this line in particular, but what is the testing plan for this visualizer?

neqo-qpack/src/qlog.rs
martinthomson, line 1: Can we delete this entire file now?  The function is *called*, but that call can be trimmed out too.

martinthomson (approved):
I'm not 100% on reviewing dependencies.  I'll trust that you know how to manage the interactions with Gecko.

test/qvis.py
larseggert: Let me move this out of the neqo tree.

Cargo.lock
larseggert: Yes, this will need Gecko updates. I did a bit of checking and this one isn't trivial due to `servo`.

larseggert: https://github.com/cloudflare/quiche/pull/2762

PR #3758, opened 2026-06-25
fix: Do not double-count PMTUD-related packet events in stats

larseggert:
And minor cleanups.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -2469,8 +2469,8 @@ impl Connection {
     }
 
     /// Write frames to the provided builder.  Returns a list of tokens used for
-    /// tracking loss or acknowledgment, whether any frame was ACK eliciting, and
-    /// whether the packet was padded.
+    /// tracking loss or acknowledgment, whether any frame was ACK eliciting,
+    /// whether the packet was padded, and whether this is a PMTUD probe.
     fn write_frames(
         &mut self,
         path: &PathRef,
@@ -2479,10 +2479,11 @@ impl Connection {
         builder: &mut packet::Builder<&mut Vec<u8>>,
         coalesced: bool, // Whether this packet is coalesced behind another one.
         now: Instant,
-    ) -> (recovery::Tokens, bool, bool) {
+    ) -> (recovery::Tokens, bool, bool, bool) {
         let mut tokens = recovery::Tokens::new();
         let primary = path.borrow().is_primary();
         let mut ack_eliciting = false;
+        let mut is_pmtud_probe = false;
 
         if primary {
             let stats = &mut self.stats.borrow_mut().frame_tx;
@@ -2515,7 +2516,7 @@ impl Connection {
 
         if profile.ack_only() {
             // If we are CC limited we can only send ACKs!
-            return (tokens, false, false);
+            return (tokens, false, false, false);
         }
 
         if primary {
@@ -2531,6 +2532,7 @@ impl Connection {
                         &mut self.stats.borrow_mut(),
                     );
                     ack_eliciting = true;
+                    is_pmtud_probe = true;
                 }
                 self.write_appdata_frames(builder, &mut tokens, now);
             } else {
@@ -2569,7 +2571,7 @@ impl Connection {
             false
         };
 
-        (tokens, ack_eliciting, padded)
+        (tokens, ack_eliciting, padded, is_pmtud_probe)
     }
 
     fn write_closing_frames<B: Buffer>(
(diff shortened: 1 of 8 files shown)


neqo-transport/src/recovery/mod.rs
```diff
@@ -510,6 +513,10 @@ impl Loss {
         self.qlog = qlog;
     }
 
+    fn record_packet_loss(&self, packets: &[sent::Packet]) {
+        self.stats.borrow_mut().lost += packets.iter().filter(|p| !p.is_pmtud_probe()).count();
+    }
```
mxinden: If a PMTUD probe contains either application data, or necessary QUIC related data (e.g. pings keeping the connection alive), I do think it should count as loss.

mxinden (changes requested):

neqo-transport/src/recovery/mod.rs
larseggert: Probes always carry a PING. So I'll always double-count them.

neqo-transport/src/connection/mod.rs
```diff
@@ -2810,7 +2810,9 @@ impl Connection {
                 now,
             );
 
-            self.stats.borrow_mut().packets_tx += 1;
+            if !tokens.iter().any(recovery::Token::is_pmtud_probe) {
```
martinthomson: I'd rather not have to grovel through all the tokens to find this out.  Is this something we can pass around?

I see two options, immediately:

1. Add a bool to `recovery::Tokens`
2. Add a bool to `packet::Builder`

It looks like you want the former based on how you need this for the ACK tracking.

neqo-transport/src/recovery/mod.rs
```diff
@@ -267,6 +267,9 @@ impl LossRecoverySpace {
         for p in &acked {
             self.remove_packet(p);
             eliciting |= p.ack_eliciting();
+            if p.is_pmtud_probe() {
```
martinthomson, line 273: Are you sure about the ordering here.  This makes it possible to return an empty list of acknowledged packets, but true for `eliciting`.  That's going to be super weird.

neqo-transport/src/stats.rs
martinthomson: I think that you can probably guess what I'm going to say about this chunk of stats...

neqo-transport/src/connection/mod.rs
```diff
@@ -2810,7 +2812,9 @@ impl Connection {
                 now,
             );
 
-            self.stats.borrow_mut().packets_tx += 1;
+            if !is_pmtud_probe {
+                self.stats.borrow_mut().packets_tx += 1;
+            }
```
mxinden, line 2817: I don't think we should exclude PMTUD probes from `packets_tx`. E.g. in `neqo_glue` we calculate loss ration as `stats.lost / stats.packets_tx`.

https://github.com/mozilla-firefox/firefox/blob/c686530f70e9d4b8b4285d931a5630504939fca8/netwerk/socket/neqo_glue/src/lib.rs#L740

If we count PMTUD probes into `stats.lost`, but not `stats.packets_tx`, our loss ratio metric is off.

neqo-transport/src/connection/tests/pmtud.rs
mxinden, line 1: Test additions look good to me.

neqo-transport/src/connection/mod.rs
```diff
@@ -2479,10 +2479,11 @@ impl Connection {
         builder: &mut packet::Builder<&mut Vec<u8>>,
         coalesced: bool, // Whether this packet is coalesced behind another one.
         now: Instant,
-    ) -> (recovery::Tokens, bool, bool) {
+    ) -> (recovery::Tokens, bool, bool, bool) {
```
mxinden, line 2482: I understand @martinthomson's [concern](https://github.com/mozilla/neqo/pull/3758/#discussion_r3478829609) to not iterate through `tokens` to determine `is_pmtud_probe` for the sake of performance. But introducing an additional (anonymous) `bool` comes with a readability cost, i.e. is not free either.


Before we introduce this additional noisy `is_pmtud_probe` `bool` performance optimization to `write_frames`, I would want to see a proof that iterating `tokens` on each outgoing QUIC packet has a performance impact.

neqo-transport/src/stats.rs
mxinden, line 1: I am in favor of grouping the PMTUD probes into a separate `struct`.

mxinden (changes requested):
I am in favor of the various clean-ups. I don't understand the actual core of the pull request, i.e. the change in stats counting.

I consider the high packet loss due to PMTUD on Firefox Nightly a blocker for PMTUD roll-out. I think we need to get to the root of the problem, i.e. why does PMTUD increase the 75th loss percentile, even though assumably the interface MTU should be correct in >75% of the cases.

This pull request does not fix the above issue, it just changes how we count loss. Thus with this pull request alone we can not re-enable PMTUD on Firefox Nightly. Thus I don't think this pull request should block https://github.com/mozilla/neqo/pull/3768.

neqo-transport/src/connection/mod.rs
larseggert: So **do** we want to double-count sent/lost PMTU probes in the overall packet stats **and** in the PMTUD-specific stats? (i.e., do we want to drop this PR?)

mxinden: Yes, I think we should continue as is, e.g. count lost PMTUD probes both as `stats.lost` and `stats.pmtud.lost`.

larseggert: OK. @omansfeld is that OK with you (since you were concerned about the H3 loss spike with PMTUD on?)

omansfeld: I think the problem is that user data has to be retried due to pmtud. That concern doesn’t vanish through filtering pmtud losses from statistics, it just makes it invisible from the metric that we usually watch for regressions. Long story short I agree with @mxinden that we shouldn’t filter them from `stats.lost`. I think we need to find out why we see 3 lost pmtud probes in p75 and somehow fix that.

PR #3761, opened 2026-06-25
fix(transport/quic-datagram): remove inbound queue limit

### LINKED ISSUE #1373: Make sure to buffer all Datagrams that are received in a single UDP datagram
Currently, we only buffer a certain fixed number of datagrams. If a UDP pack contains more HTTP datagrams than we would buffer, we will just drop them without informing the application. We should at least buffer all datagrams received in a single UDP packet and give the application chance to receive them.

mxinden:
The inbound queue head-dropped the oldest datagram once full. The drop itself hurts: the discarded datagram is gone for good, and on the MASQUE connect-udp path it looks like loss to the inner connection, shrinking its congestion window and throughput.

Removing the limit is safe as long as we cap how much we read from the socket per iteration and deplete neqo's event queue regularly, so the queue cannot grow without bound. Both are still to be done on the consumer side.

Also removes the now-unused plumbing:

- the `incoming_datagram_queue` connection parameter
- the `IncomingDatagramDropped` event
- the `incoming_datagram_dropped` stat
- the `check_datagram_queued` bookkeeping

Fixes #1373

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -954,7 +954,6 @@ impl Http3Client {
                 }
                 ConnectionEvent::SendStreamComplete { .. }
                 | ConnectionEvent::OutgoingDatagramOutcome { .. }
-                | ConnectionEvent::IncomingDatagramDropped
                 | ConnectionEvent::SconeUpdated(_)
                 | ConnectionEvent::PathMigrated { .. } => {}
             }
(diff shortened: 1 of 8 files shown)


larseggert (approved):
So we'll limit in neqo-glue?

mxinden:
@larseggert yes. In `neqo_glue` we would cap the receive loop so once `bytes_read` exceeds a high threshold (e.g. 10 MB), `break` like the `WouldBlock` branch and return `NS_OK`. No legitimate connection should ever read 10 MB in a single pass, so this only bounds a misbehaving/malicious peer that would otherwise make us buffer datagrams unboundedly. Even when the cap is hit, nothing is lost: `ProcessEvents` passes all buffered events up to the upper layer right after (e.g. the inner proxied MASQUE connection), and the level-triggered poll re-fires `RecvData` to continue reading the rest from the socket.

PR #3762, opened 2026-06-25
fix: reject zero Insert Count Increment on QPACK decoder stream

uwezkhan:
A zero-valued Insert Count Increment on the QPACK decoder stream is accepted today:

    decoder stream: 0x00  (Insert Count Increment, Increment = 0)
    Encoder::call_instruction -> increment_acked(0) -> Ok(())

RFC 9204 Section 4.4.3 lists that in the same sentence as an increment that raises the count past the inserts the encoder has sent, both a connection error of type QPACK_DECODER_STREAM_ERROR. `increment_acked` already rejects the latter; the zero case fell through.

Before: a zero increment is a silent no-op and the connection stays open.
After: it returns Error::DecoderStream (0x0202), the same as an over-large increment.

Tradeoff: less tolerant of a misbehaving peer decoder, but a zero increment conveys nothing and the RFC requires the error, so both halves of 4.4.3 now behave the same.

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -230,6 +230,12 @@ impl Encoder {
         qdebug!("[{self}] call instruction {instruction:?}");
         match instruction {
             DecoderInstruction::InsertCountIncrement { increment } => {
+                // RFC 9204, Section 4.4.3: an Increment of zero is a connection error
+                // of type QPACK_DECODER_STREAM_ERROR, the same as one that raises the
+                // count past the inserts sent (rejected in `increment_acked`).
+                if increment == 0 {
+                    return Err(Error::DecoderStream);
+                }
                 qlog::qpack_read_insert_count_increment_instruction(
                     qlog,
                     increment,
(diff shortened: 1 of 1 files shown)


neqo-qpack/src/encoder.rs
```diff
@@ -1008,6 +1014,29 @@ mod tests {
...
+    fn insert_count_increment_zero_is_rejected() {
+        let mut encoder = connect(false);
+
+        // 0x00 is an Insert Count Increment instruction carrying an Increment of 0.
+        encoder
+            .peer_conn
+            .stream_send(encoder.recv_stream_id, &[0x00])
+            .unwrap();
+        let out = encoder.peer_conn.process_output(now());
+        drop(encoder.conn.process(out.dgram(), now()));
```
martinthomson:
```suggestion
        encoder.conn.process_input(out.dgram().unwrap(), now());
```

martinthomson (approved):
We can avoid generating packets by using process_input, but this is a useful fix.

neqo-qpack/src/encoder.rs
uwezkhan: Good call, switched to process_input so the test doesn't bother building the response packet. Behavior is the same and the assert still gets DecoderStream.

PR #3763, opened 2026-06-26
Hex debugging improvements

martinthomson:
Yes, this is a lot of change, but it makes a big difference when debugging.

For example, running `RUST_LOG=neqo cargo test` on main, my machine took almost 28 seconds.  With these changes, the same took around 19 seconds.

There are some improvements to the debug implementations of certain common items that I see in logs as well.  Transport Parameters, Headers, and HTTP/3 frames.

@larseggert, WDYT?

### DIFF
--- a/neqo-common/src/hex.rs
+++ b/neqo-common/src/hex.rs
@@ -34,9 +34,6 @@ impl<T> $n<T> {
     #[doc = concat!(r#"assert_eq!(format!("{:?}", Example("v".to_string(), vec!"#, stringify!($in), ")), expected);")]
     )*
     /// ```
-    ///
-    /// # Errors
-    /// Propagates any errors from the formatter.
     pub const fn new(v: T) -> Self {
         Self(v)
     }
--- a/neqo-common/src/hex.rs
+++ b/neqo-common/src/hex.rs
@@ -145,34 +145,31 @@ mod tests {
 
     #[test]
     fn hex_output() {
-        assert_eq!(format!("{}", &Hex::new([])), "");
-        assert_eq!(format!("{}", &Hex::new([0xab, 0xcd])), "abcd");
+        assert_eq!(format!("{}", Hex::new([])), "");
+        assert_eq!(format!("{}", Hex::new([0xab, 0xcd])), "abcd");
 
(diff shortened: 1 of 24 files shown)


neqo-common/src/hex.rs
```diff
@@ -0,0 +1,181 @@
...
+impl<T: AsRef<[u8]>> fmt::Debug for $n<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        <Self as fmt::Display>::fmt(self, f)
+    }
+}
+    };
+}
+
+/// A fast hex implementation with fixed overhead.
+fn chunk_hex<T: AsRef<[u8]>>(f: &mut fmt::Formatter<'_>, v: T) -> fmt::Result {
```
martinthomson, line 79: This is the heart of the performance gains here.  It avoids a heap allocation for every call to `hex()` and friends (this is the `String` we allocate in the old implementation).  But it ended up being critical to only invoke the write functions on the `Formatter` occasionally.  My original implementation sent a byte at a time out and that was a lot slower than this.

neqo-common/src/hex.rs
```diff
@@ -0,0 +1,181 @@
...
+    ///     }
+    /// }
+    $(
+    #[doc = concat!(r##"let expected = concat!(r#"Example("v", "#, "##, stringify!($out), r#", ")");"#)]
+    #[doc = concat!(r#"assert_eq!(format!("{:?}", Example("v".to_string(), vec!"#, stringify!($in), ")), expected);")]
+    )*
+    /// ```
+    ///
+    /// # Errors
+    /// Propagates any errors from the formatter.
```
martinthomson:
```suggestion
```

larseggert (approved):
I like it. I wonder if we can use https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html instead of implementing the lengthy `Debug` traits manually? There are also some bot comments that look sensible.

martinthomson:
I looked at strum, but it doesn't have an obvious way to extend it with filters for specific fields.  Most of the stuff that needs to be hex is a `Vec<u8>` buried in a larger struct or enum and that's just hard to get right.

PR #3765, opened 2026-06-26
fix: reject qpack required insert count that reconstructs to zero

uwezkhan:
`calc_req_insert_cnt` reconstructs a header block's Required Insert Count with the RFC 9204 Section 4.5.1.1 algorithm, but stops one step short: a non-zero Encoded Insert Count that reconstructs to a Required Insert Count of 0 is supposed to be a decoding error. With fewer than MaxEntries insertions, an Encoded Insert Count of 1 wraps back to 0, so a peer can send a field section whose prefix claims a dynamic dependency the encoder never expressed.

Reject a reconstructed Required Insert Count of 0 with the same `Error::Decompression` (QPACK_DECOMPRESSION_FAILED) the neighboring checks use. The check sits next to the reconstruction because that is the only place the encoded value and the wrap correction are both in hand. Before, such a block decoded to an empty or static-only field section; after, it closes the connection. The tradeoff is strictness over leniency, but no conformant encoder produces this since a Required Insert Count of 0 is always encoded as 0.

### DIFF
--- a/neqo-qpack/src/header_block.rs
+++ b/neqo-qpack/src/header_block.rs
@@ -930,9 +930,9 @@ mod tests {
     }
 
     /// RFC 9204 Section 4.5.1.1: a non-zero Encoded Insert Count that
-    /// reconstructs to a Required Insert Count of 0 is a decoding error. With
-    /// fewer than `MaxEntries` insertions, an Encoded Insert Count of 1 wraps
-    /// back to 0 and must be rejected.
+    /// reconstructs to a Required Insert Count of 0 is a decoding error.
+    /// A value of 1 maps to a value of 0, which is invalid unless the
+    /// value has wrapped, which this test doesn't cover.
     #[test]
     fn req_insert_count_reconstructed_to_zero() {
         let table = HeaderTable::new(false);
--- a/neqo-qpack/src/header_block.rs
+++ b/neqo-qpack/src/header_block.rs
@@ -300,6 +300,9 @@ impl<'a> HeaderDecoder<'a> {
                 }
                 req_insert_cnt -= full_range;
             }
+            if req_insert_cnt == 0 {
+                return Err(Error::Decompression);
+            }
             Ok(req_insert_cnt)
         }
     }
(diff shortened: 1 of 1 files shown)


neqo-qpack/src/header_block.rs
martinthomson, line 295: This should be `div_floor()` now.  Or just `/`.  (Not in this fix.)

neqo-qpack/src/header_block.rs
```diff
@@ -926,6 +929,28 @@ mod tests {
         );
     }
 
+    /// RFC 9204 Section 4.5.1.1: a non-zero Encoded Insert Count that
+    /// reconstructs to a Required Insert Count of 0 is a decoding error. With
+    /// fewer than `MaxEntries` insertions, an Encoded Insert Count of 1 wraps
+    /// back to 0 and must be rejected.
```
martinthomson: The implication here is that the wrapping is important for this test case.  It is not.  I would rephrase this as:

```suggestion
    /// RFC 9204 Section 4.5.1.1: a non-zero Encoded Insert Count that
    /// reconstructs to a Required Insert Count of 0 is a decoding error.
    /// A value of 1 maps to a value of 0, which is invalid unless the
    /// value has wrapped, which this test doesn't cover.
```

It's OK not to test the wrapping case, because that means cycling through a two full tables worth of entries.  But the wrapping that exists only exists to protect against underflow on the subtraction.  That's because we want to get the value of `encoded - 1` modulo `full_range`.

The logic is the same basic theory as packet number reconstruction, FWIW.

uwezkhan: Applied your wording. The old phrasing did read like the wrap was the point of the test, which it isn't here.

neqo-qpack/src/header_block.rs
uwezkhan: Makes sense. Left it untouched here since it's a separate cleanup; happy to switch it to div_floor or plain / in a followup.

larseggert: https://github.com/mozilla/neqo/issues/3766

larseggert (approved):

PR #3776, opened 2026-07-01
fix(common): support 32-bit targets (partial revert of #3738)

mxinden:
#3738 added `const_assert_eq!(usize::BITS, u64::BITS)` and a `to_usize` helper. The assertion requires a 64-bit usize and breaks compilation on 32-bit targets. Firefox still ships 32-bit Tier-1 builds (Windows x86 and Android ARMv7), see
https://firefox-source-docs.mozilla.org/contributing/build/supported.html

This is a partial revert of #3738: it removes the u64 -> usize direction. The `to_usize` helper and the assertion are dropped, and the call sites are restored to their original fallible `usize::try_from(...)` form, which handles the 32-bit case correctly. A centralized infallible `to_usize` is not viable, since narrowing a u64 to usize is lossy on 32-bit, and as a const fn with a debug_assert it can panic during const evaluation, i.e. at compile time.

The usize -> u64 direction (`to_u64`) is always lossless and is kept.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -758,7 +758,6 @@ dependencies = [
  "log",
  "qlog",
  "sha1",
- "static_assertions",
  "strum",
  "test-fixture 0.30.0",
  "thiserror",
--- a/neqo-common/Cargo.toml
+++ b/neqo-common/Cargo.toml
@@ -23,7 +23,6 @@ hex = { workspace = true, optional = true }
 log = { workspace = true }
 qlog = { workspace = true }
 sha1 = { version = "0.10", default-features = false, optional = true }
-static_assertions = { workspace = true }
 strum = { workspace = true }
 thiserror = { workspace = true }
 
--- a/neqo-common/src/codec.rs
+++ b/neqo-common/src/codec.rs
@@ -9,7 +9,7 @@ use std::{
     io::{self, Cursor},
 };
 
-use crate::{Length, hex::HexWithLen, to_u64, to_usize};
+use crate::{Length, hex::HexWithLen, to_u64};
 
 pub const MAX_VARINT: u64 = (1 << 62) - 1;
 
@@ -52,7 +52,8 @@ impl<'a> Decoder<'a> {
     /// Only use this for tests because we panic rather than reporting a result.
     #[cfg(any(test, feature = "test-fixture"))]
     fn skip_inner(&mut self, n: Option<u64>) {
-        self.skip(to_usize(n.expect("invalid length")));
+        #[expect(clippy::unwrap_used, reason = "Only used in tests.")]
+        self.skip(usize::try_from(n.expect("invalid length")).unwrap());
     }
 
     /// Skip a vector.  Panics if there isn't enough space.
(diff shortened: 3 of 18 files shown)


martinthomson:
I'd prefer to keep the bidirectional translation, but instead keep a warning on the lossy direction (`u64` to `usize`) that this is intended for use in places where the underlying value has intrinsic constraints on it being a size or length.  Ideally, that is because the value is a measure of something we hold in memory or something like that.  The alternative is a lot of these `TryFrom` invocations with unwrapping.

(Separately, is there a way to stop clippy from propagating panic docs warnings from these functions?)

mxinden:
> I'd prefer to keep the bidirectional translation, but instead keep a warning on the lossy direction (`u64` to `usize`) that this is intended for use in places where the underlying value has intrinsic constraints on it being a size or length. Ideally, that is because the value is a measure of something we hold in memory or something like that. The alternative is a lot of these `TryFrom` invocations with unwrapping.
> 
> (Separately, is there a way to stop clippy from propagating panic docs warnings from these functions?)

@martinthomson can we delay this discussion to a future pull request? This pull request is a revert only, unblocking all other changes in the v0.30.0 release.

martinthomson:
> @martinthomson can we delay this discussion to a future pull request? This pull request is a revert only, unblocking all other changes in the v0.30.0 release.

I'd really prefer to avoid the regressions that come from this.

If you must act (and #3779 is understandably rough) then I'd prefer that you limit this change.  It should be possible to remove the assertion that `sizeof(usize) == sizeof(u64)` and keep everything else.

mxinden:
> > @martinthomson can we delay this discussion to a future pull request? This pull request is a revert only, unblocking all other changes in the v0.30.0 release.
> 
> I'd really prefer to avoid the regressions that come from this.
> 
> If you must act (and #3779 is understandably rough) then I'd prefer that you limit this change. It should be possible to remove the assertion that `sizeof(usize) == sizeof(u64)` and keep everything else.

Do you feel strongly about this @martinthomson? In my eyes this pull request does not introduce a regression but simply reverts the regression introduced in https://github.com/mozilla/neqo/pull/3738. https://github.com/mozilla/neqo/pull/3738 is syntactic sugar only, i.e. does not provide any functionality. Instead https://github.com/mozilla/neqo/pull/3738 now blocks all `v0.30.0` changes from landing in Firefox.

> It should be possible to remove the assertion that `sizeof(usize) == sizeof(u64)` and keep everything else.

I don't think a simple `to_usize` as introduced in https://github.com/mozilla/neqo/pull/3738 is viable. In my eyes it hides a footgun, namely the bit truncation on 32bit systems. I suspect debugging such accidental truncation to be difficult. Seen from a different perspective, I do think it is good that Rust `std` does not provide such simple conversion function. If we do introduce such conversion function, I would want the name to be explicit about the possible truncation.

martinthomson:
I do feel strongly about that.  The first step to dealing with the problem of truncation is to move it to a single point that you can control.  Because right now, we have code all over the codebase that is happily truncating and we have no real way to contain it.  So reverting the change is a setback in terms of us being able to start to control things.  That's why I opened #3779, because it is starting to put real controls in place.

martinthomson (approved):
Spoke with Max, we'll do something like to_usize, but we won't call it that because it should panic always, not just in debug mode.  And we found some real bugs to fix around integer overflow in our existing code, so we'll need to fix each of those.

mxinden:
Merging here. I will work both on (1) `assert_to_usize` and (2) the above mentioned integer overflow bugs. Once that is done, I will propose a new release to land in Firefox.

PR #3777, opened 2026-07-02
Reliable reset handling for HTTP/3 

martinthomson:
This follow-up to the RESET_STREAM_AT PR (#3754) makes changes to the
event handling in the HTTP/3 stack to make reliable delivery work
better.

This is entirely unnecessary, but I think that the change is worthwhile
doing on balance.  I wanted to improve test coverage and found that the
HTTP/3 event handling wasn't quite able to handle reliable resets
properly.  Making it proper took very little extra work (for the LLM),
so I think this is worth it.

The fixes here aren't likely to make a material difference in the real
world -- HTTP/3 requests that receive any type of reset probably need to
be dropped anyway -- but it makes our code more robust and better
tested.


neqo-http3/src/frames/reader.rs
```diff
@@ -184,16 +184,18 @@ impl FrameReader {
...
+            let (output, read, fin) =
+                match stream_reader.read_data(&mut self.buffer[..to_read], now) {
+                    Ok((0, f)) => (None, false, f),
+                    Ok((amount, f)) => {
+                        qtrace!("FrameReader::receive: reading {amount} byte, fin={f}");
+                        (self.consume::<T>(amount)?, true, f)
+                    }
+                    // A `RESET_STREAM` will cause the transport to report `NoMoreData`.
+                    // Don't treat that as an error here, let the event handling deal with it.
+                    Err(Error::Transport(TransportError::NoMoreData)) => break Ok((None, false)),
```
mxinden: With `NoMoreData` handled here, does this `debug_assert` need updating?

https://github.com/mozilla/neqo/blob/9b564e79cf0bf0a7bf1cf74170c2a3bbc89d9610/neqo-http3/src/lib.rs#L374-L379

mxinden (approved):

neqo-http3/src/frames/reader.rs
martinthomson: I'm looking at this and it seems like the error mapping is entirely unnecessary on the receive path.

@mxinden, can you sanity check me here?  I think that the upshot of this is that we can trim out the mapping of error codes on this path.  It loses the `TransportStreamDoesNotExist` option on this path, but that is an error condition we can explain (see the updated comments in the latest).

I'm also going to see if I can make the error mapping routines less weird as well.  That's not business that really belongs where it is right now.

martinthomson: I want to flag that further investigation revealed a bug.  The test case is a little convoluted, but it shows a connection error being generated in response to a perfectly valid sequence of events.  I'm working on fixing it.

neqo-http3/src/frames/reader.rs
```diff
@@ -178,25 +178,32 @@ impl FrameReader {
     ///
     /// # Errors
     ///
-    /// May return `HttpFrame` if a frame cannot be decoded.
-    /// and `TransportStreamDoesNotExist` if `stream_recv` fails.
+    /// May return [`Error::HttpFrame`] if a frame cannot be decoded.
+    /// Can return [`Error::Transport`] with [`TransportError::InvalidStreamId`]
+    /// if the stream was closed as a side-effect of previous reads,
+    /// without invoking `process()` or `process_output()` to synchronize state.
```
martinthomson:
```suggestion
    /// May return [`Error::HttpFrame`] if a frame cannot be decoded.
    /// Absorbs [`TransportError::NoMoreData`] and [`TransportError::InvalidStreamId`]
    /// into a zero-length read if the stream was reset and the stream closed.
    /// The reset is propagated through events.
```

neqo-http3/src/frames/reader.rs
larseggert: @martinthomson I assume you did the fix? If yes, please resolve the comment so the auto-merge fires.

PR #3778, opened 2026-07-02
fix: reject qpack header acknowledgment for a fully-acked stream

uwezkhan:
Cause: on the QPACK decoder stream, `Encoder::header_ack` looks the stream up in `unacked_header_blocks` and, when the entry is missing, does nothing. A peer can send a Header Acknowledgment for a stream that never sent a header block referencing the dynamic table (or whose blocks were all already acknowledged, so the entry was removed), and the encoder accepts it silently. RFC 9204 Section 4.4.1 says such an acknowledgment is a connection error of type QPACK_DECODER_STREAM_ERROR.

Fix: return `Error::DecoderStream` (0x0202) when the stream has no outstanding acknowledgment. The check lives on the encoder because that is the only place the set of unacknowledged header blocks is tracked.

Before: a spurious Header Acknowledgment (e.g. 0x81 for stream 1 with nothing outstanding) is a no-op and the connection stays open.
After: the same instruction closes the connection with QPACK_DECODER_STREAM_ERROR.

Tradeoff is strictness over tolerance toward a decoder that over-acknowledges, matching how the sibling Insert Count Increment instruction already errors on an out-of-range value. Added a regression test that drives the encoder with a Header Acknowledgment for a stream that sent nothing.

### DIFF
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -177,6 +177,8 @@ impl Encoder {
         // every header block with a non-zero Required Insert Count has already been
         // acknowledged is a connection error of type QPACK_DECODER_STREAM_ERROR. Such a
         // stream has no entry in `unacked_header_blocks` (entries are removed once empty).
+        // We only remove entries in response to messages from the peer, so this will only fail
+        // if the peer references a stream that never generated unacknowledged blocks.
         let hb_list = self
             .unacked_header_blocks
             .get_mut(&stream_id)
--- a/neqo-qpack/src/encoder.rs
+++ b/neqo-qpack/src/encoder.rs
@@ -171,32 +171,38 @@ impl Encoder {
         Ok(())
     }
 
-    fn header_ack(&mut self, stream_id: StreamId) {
+    fn header_ack(&mut self, stream_id: StreamId) -> Res<()> {
         self.stats.header_acks_recv += 1;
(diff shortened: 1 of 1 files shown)


neqo-qpack/src/encoder.rs
```diff
@@ -171,32 +171,38 @@ impl Encoder {
         Ok(())
     }
 
-    fn header_ack(&mut self, stream_id: StreamId) {
+    fn header_ack(&mut self, stream_id: StreamId) -> Res<()> {
         self.stats.header_acks_recv += 1;
+        // RFC 9204, Section 4.4.1: a Header Acknowledgment referring to a stream on which
+        // every header block with a non-zero Required Insert Count has already been
+        // acknowledged is a connection error of type QPACK_DECODER_STREAM_ERROR. Such a
+        // stream has no entry in `unacked_header_blocks` (entries are removed once empty).
```
martinthomson, line 179:
```suggestion
        // stream has no entry in `unacked_header_blocks` (entries are removed once empty).
        // We only remove entries in response to messages from the peer, so this will only fail
        // if the peer references a stream that never generated unacknowledged blocks.
```

martinthomson (approved):
The effects of this caused me to do some analysis, which only showed that this change was good.  (Maybe there are other problems to address, but I'll follow up separately.)

PR #3780, opened 2026-07-03
Cap qpack tracking

martinthomson:
This puts a hard cap on the number of streams that we'll track in the QPack encoder.  Past that limit, we'll avoid the dynamic table.

An attacker that never acknowledges qpack streams will cause our encoder to hold state for every stream that was ever opened.  That's essentially unbounded, even if the per-stream state is pretty small.  So stop doing that.

This turned out to be a little fiddly, but I think that the end result is pretty comprehensible.

There is an interesting question about the value of the new setting: in one version of this, the value is determined by the number of streams we are willing to track at the transport layer.  Coordinating the values to do that turns out to be super difficult to do.  In the end, I decided not to put the effort in for that.  This is on the basis that the tracking of streams is generally something that doesn't have the same lifetime as a stream.  

A long-lived stream will generally only ever use the encoder near the start of its life, then the tracking goes away.  So the limit we have is probably safe.  I doubt we'll see more than 1k concurrent HEADERS sent to a server.  We're unlikely to get that many streams from a server, let alone be able to make that many concurrent requests.  At which point, a few requests without compression is fine.

As for whether 1k is too much, the cost here is dependent on how many entries each stream references, but that many outstanding `HashSet`s is probably fine.

### DIFF
--- a/neqo-http3/src/conn_params.rs
+++ b/neqo-http3/src/conn_params.rs
@@ -13,6 +13,7 @@ use neqo_transport::ConnectionParameters;
 const QPACK_MAX_TABLE_SIZE_DEFAULT: u64 = 65536;
 const QPACK_TABLE_SIZE_LIMIT: u64 = (1 << 30) - 1;
 const QPACK_MAX_BLOCKED_STREAMS_DEFAULT: u16 = 20;
+const QPACK_MAX_TRACKED_STREAMS_DEFAULT: usize = 1 << 12;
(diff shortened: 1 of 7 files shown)


larseggert (approved):
Looks reasonable, but I I'm not very deep in QPACK. Should we expose limit overflow in stats?

martinthomson:
Yes, but I'll do that in a follow-up.

PR #3782, opened 2026-07-04
fix: reject server-only transport parameters sent by a client

uwezkhan:
Injecting a server-only transport parameter into a client and running the
handshake against a default server lands here:

    thread 'connection::tests::migration::server_only_tparams_from_client'
    panicked at neqo-transport/src/connection/tests/mod.rs:314:14:
    bad state Confirmed

RFC 9000, Section 18.2 marks `original_destination_connection_id`,
`retry_source_connection_id`, and `stateless_reset_token` as only sent by a
server (a client MUST NOT send the reset token). `process_tps` already rejects a
client's `preferred_address`; the other three fall through.

Before: a server accepts all three from a client. `validate_cids` only checks
the two connection-id parameters when the role is Client, and `process_tps`
reads `stateless_reset_token` for either role, so a client-supplied reset token
lands on the primary path.

After: a server rejects any of the three with a transport parameter error,
alongside the existing `preferred_address` check. The client path is untouched
and still validates these against the values it observed during the handshake.

Tradeoff is strict rejection of a known set over ignoring unexpected
parameters, matching how `preferred_address` is handled today.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -3036,6 +3036,23 @@ impl Connection {
             let tps = self.tps.borrow();
             let remote = tps.remote_handshake().ok_or(Error::TransportParameter)?;
 
+            // RFC 9000, Section 18.2: `original_destination_connection_id`,
+            // `retry_source_connection_id`, and `stateless_reset_token` are only sent
+            // by a server. A server that receives one from a client rejects it with a
+            // transport parameter error. `preferred_address`, the fourth server-only
+            // parameter, is handled by the check below.
+            if self.role == Role::Server
+                && [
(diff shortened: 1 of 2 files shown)


martinthomson:
Thanks for this, but I think that there is a cleaner solution: see #3783.

uwezkhan:
Agreed, decode-time rejection is cleaner. It covers every path that parses transport parameters instead of just process_tps, and the offending parameter never gets stored in the first place. If useful, the handshake-level test from this branch (client sends each of the three, server closes with TRANSPORT_PARAMETER_ERROR) should apply on top of #3783. Happy to close this in favor of it.

martinthomson:
Thanks, your test is great.  I've merged that over on #3782.

PR #3785, opened 2026-07-06
fix: discard retry with scid matching the initial dcid

uwezkhan:
handle_retry validates a received Retry against a duplicate Retry, an empty token, and a bad integrity tag, but never compares the Retry's Source Connection ID to the Destination Connection ID the client used in its Initial. RFC 9000 Section 17.2.5.2 requires a client to discard a Retry whose Source Connection ID matches that Destination Connection ID. The Retry integrity key is a fixed public value, so a valid tag does not by itself show the Retry came from the server; a party that observed the Initial can compute one, which is why this comparison exists.

Before, such a Retry was adopted: the client took its Source Connection ID as the new remote CID, saved the token, and re-keyed the initial secrets. After, the packet is dropped and the client stays in WaitInitial. The check sits beside the existing integrity check in handle_retry, the one place a client processes a Retry, so no caller needs a separate guard. The tradeoff is a stricter client that discards one more Retry shape, against the negligible case of a server that reuses the client's random Initial DCID as its own connection ID.

### DIFF
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -2037,6 +2037,8 @@ fn retry_scid_matching_initial_dcid() {
         let mut dec = Decoder::from(&initial[5..]);
         let dcid = dec.decode_vec(1).expect("client DCID").to_vec();
         let scid = dec.decode_vec(1).expect("client SCID").to_vec();
+        assert!(!dcid.is_empty(), "client DCID is non-empty");
+        assert!(!scid.is_empty(), "client SCID is non-empty");
         (dcid, scid)
     }
 
@@ -2051,10 +2053,10 @@ fn retry_scid_matching_initial_dcid() {
     let (dcid, scid) = initial_cids(&initial);
     let retry = crate::packet::Builder::retry(
         Version::default(),
-        &scid,   // Retry Destination CID: the client's Source CID
-        &dcid,   // Retry Source CID equal to the client's Initial Destination CID
+        &scid,   // Destination CID: copied from the client as required
+        &dcid,   // Source CID: copied from client (bad!)
         &[0x01], // non-empty token
-        &dcid,   // integrity tag is keyed on the client's Initial Destination CID
+        &dcid,   // as required: a seed for authenticating the Retry
     )
     .expect("build retry");
     drop(client.process(Some(datagram(retry)), now()));
--- a/neqo-transport/src/connection/tests/handshake.rs
+++ b/neqo-transport/src/connection/tests/handshake.rs
@@ -17,6 +17,8 @@ use std::{
     time::Duration,
 };
 
+#[cfg(not(feature = "disable-encryption"))]
+use neqo_common::Decoder;
 use neqo_common::{Datagram, event::Provider as _, qdebug, to_u64};
 use nss::{AuthenticationStatus, constants::TLS_CHACHA20_POLY1305_SHA256, generate_ech_keys};
 #[cfg(not(feature = "disable-encryption"))]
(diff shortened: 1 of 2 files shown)


larseggert:
@uwezkhan thanks for all the contributions, including this one. Can we send you some Mozilla swag as a little thank you?

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2022,3 +2024,56 @@ fn export_keying_material_after_closed() {
...
+#[cfg(not(feature = "disable-encryption"))]
+#[test]
+fn retry_scid_matching_initial_dcid() {
+    // Read the connection IDs carried in the client's Initial: skip the first byte
+    // and the 4-byte version, then two length-prefixed connection IDs (DCID, SCID).
+    fn initial_cids(initial: &[u8]) -> (Vec<u8>, Vec<u8>) {
+        let mut dec = Decoder::from(&initial[5..]);
+        let dcid = dec.decode_vec(1).expect("client DCID").to_vec();
+        let scid = dec.decode_vec(1).expect("client SCID").to_vec();
+        (dcid, scid)
```
martinthomson, line 2042: This should probably assert that both DCID and SCID are non-empty.  Otherwise the `[0]` below will fail.  (They should be, it's just reinforcement.)

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2022,3 +2024,56 @@ fn export_keying_material_after_closed() {
...
+    let mut client = default_client();
+    let initial = client
+        .process_output(now())
+        .dgram()
+        .expect("a datagram")
+        .to_vec();
+    let (dcid, scid) = initial_cids(&initial);
+    let retry = crate::packet::Builder::retry(
+        Version::default(),
+        &scid,   // Retry Destination CID: the client's Source CID
```
martinthomson:
```suggestion
        &scid,   // Destination CID: copied from the client as required
```

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2022,3 +2024,56 @@ fn export_keying_material_after_closed() {
...
+        &dcid,   // Retry Source CID equal to the client's Initial Destination CID
```
martinthomson:
```suggestion
        &dcid,   // Source CID: copied from client (bad!)
```

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2022,3 +2024,56 @@ fn export_keying_material_after_closed() {
...
+        &[0x01], // non-empty token
+        &dcid,   // integrity tag is keyed on the client's Initial Destination CID
```
martinthomson:
```suggestion
        &[0x01], // non-empty token
        &dcid,   // as required: a seed for authenticating the Retry
```

martinthomson (commented):
This is really good.  Thanks.

uwezkhan:
> @uwezkhan thanks for all the contributions, including this one. Can we send you some Mozilla swag as a little thank you?

Thanks Lars that's really kind I'd be happy to. What's the best way to get you my shipping details? Didn't want to drop my address in the thread here.

larseggert:
@uwezkhan please email me at leggert [at] mozilla.com

neqo-transport/src/connection/tests/handshake.rs
uwezkhan: Done, added asserts for both. Also took the three comment rewordings as suggested.

uwezkhan:
> please email me at leggert [at] mozilla.com

Will do.

martinthomson (approved):

PR #3787, opened 2026-07-06
chore: Reduce benchmark CI noise

larseggert:
Tighten CPU tuning, use consistent Criterion noise thresholds, and switch to scipy's Mann-Whitney U so outlier-heavy hyperfine runs don't skew significance. Also fix a perfcompare.py gate that never fired.


larseggert:
@codspeedbot explain why this is faster

PR #3797, opened 2026-07-07
fix: bound the RETIRE_CONNECTION_ID queue

larseggert:
Cap pending retirements at 2x active_connection_id_limit per RFC 9000 Section 5.1.2, closing with CONNECTION_ID_LIMIT_ERROR if exceeded.
https://bugzilla.mozilla.org/show_bug.cgi?id=2010266

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -3417,7 +3417,9 @@ impl Connection {
                     stateless_reset_token,
                 ))?;
                 self.paths.retire_cids(retire_prior, &mut self.cids);
-                if self.cids.len() >= ConnectionIdManager::ACTIVE_LIMIT {
+                if self.cids.len() >= ConnectionIdManager::ACTIVE_LIMIT
+                    || self.paths.retire_queue_len() > ConnectionIdManager::MAX_RETIRE_QUEUE
+                {
                     qinfo!("[{self}] received too many connection IDs");
                     return Err(Error::ConnectionIdLimitExceeded);
                 }
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -3417,10 +3417,18 @@ impl Connection {
                     stateless_reset_token,
                 ))?;
                 self.paths.retire_cids(retire_prior, &mut self.cids);
-                if self.cids.len() >= ConnectionIdManager::ACTIVE_LIMIT
-                    || self.paths.retire_queue_len() > ConnectionIdManager::MAX_RETIRE_QUEUE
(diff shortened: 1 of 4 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -3417,7 +3417,9 @@ impl Connection {
                     stateless_reset_token,
                 ))?;
                 self.paths.retire_cids(retire_prior, &mut self.cids);
-                if self.cids.len() >= ConnectionIdManager::ACTIVE_LIMIT {
+                if self.cids.len() >= ConnectionIdManager::ACTIVE_LIMIT
+                    || self.paths.retire_queue_len() > ConnectionIdManager::MAX_RETIRE_QUEUE
```
martinthomson: Your assumption here is that it is OK to temporarily exceed the queue length, because you will never get more than `ACTIVE_LIMIT` above that length.  That's probably worth noting too.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -1069,26 +1069,44 @@ fn migration_invalid_address() {
...
+            seqno,
+            count,
+            cid_gen,
+        }
+    }
+
+    /// A single `NEW_CONNECTION_ID` frame that retires every previously issued
+    /// connection ID.
+    fn retire_all(cid_gen: Rc<RefCell<dyn ConnectionIdGenerator>>) -> Self {
+        Self::new(100, 1, cid_gen)
```
martinthomson: The comment about the starting value of 100 has been lost, but it's worth keeping, I think.  Because it's not obvious where it comes from.

Also, you can probably hard code the initial value of `seqno` to 100 and not have it in the function signature at all.

neqo-transport/src/connection/tests/migration.rs
```diff
@@ -1128,6 +1145,48 @@ fn retire_all() {
...
+    // The client never runs `process_output` here, so it cannot send (or get
+    // acknowledgement for) any `RETIRE_CONNECTION_ID` frames: its pending-retirement
+    // queue only grows.  Because every frame retires all prior connection IDs, the
+    // store stays at one active connection ID and the existing
+    // active-connection-ID-limit check never fires; only a dedicated bound on the
+    // retirement queue can stop this.
+    let count = to_u64(ConnectionIdManager::MAX_RETIRE_QUEUE) + 1;
+    let ncids = send_with_extra(
+        &mut server,
+        NewConnectionIds::new(100, count, cid_gen),
```
martinthomson: Another magic number here!

martinthomson (approved):
LGTM

PR #3798, opened 2026-07-07
fix: limit discontiguous STREAM reassembly ranges

larseggert:
Don't let a peer consume excess memory via many tiny out-of-order STREAM frames.
https://bugzilla.mozilla.org/show_bug.cgi?id=2048871

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -353,9 +353,11 @@ impl RxStreamOrderer {
                 // Continue, since we may have more overlaps
             }
 
-            for start in to_remove {
-                self.data_ranges.remove(&start);
-            }
+            self.mutate_ranges(|data_ranges| {
+                for start in to_remove {
+                    data_ranges.remove(&start);
+                }
+            });
         }
 
         if !to_add.is_empty() {
@@ -441,20 +443,21 @@ impl RxStreamOrderer {
         reason = "OK here."
     )]
     pub fn discard_after(&mut self, offset: u64) {
-        self.data_ranges.split_off(&offset);
-        // Truncate a range that straddles `offset`.
-        if let Some(mut e) = self.data_ranges.last_entry() {
-            // Note: no underflow risk, all ranges that start at or after offset are gone.
-            let start = *e.key();
-            let keep = to_usize(offset - start);
-            let data = e.get_mut();
-            data.truncate(keep);
+        let retired = self.retired;
+        self.end = self.mutate_ranges(|data_ranges| {
+            data_ranges.split_off(&offset);
+            // Truncate a range that straddles `offset`.
+            data_ranges.last_entry().map_or(retired, |mut e| {
+                // Note: no underflow risk, all ranges that start at or after offset are gone.
+                let start = *e.key();
+                let keep = to_usize(offset - start);
+                let data = e.get_mut();
+                data.truncate(keep);
 
-            // No overflow risk: neither start nor offset can exceed 1<<62.
-            self.end = start + to_u64(data.len());
-        } else {
-            self.end = self.retired;
-        }
+                // No overflow risk: neither start nor offset can exceed 1<<62.
+                start + to_u64(data.len())
+            })
+        });
     }
 
     /// Count non-adjacent ("split") runs of buffered data, for test validation.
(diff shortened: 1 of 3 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -433,6 +436,27 @@ impl RxStreamOrderer {
...
+    /// Number of discontiguous runs of buffered data.
+    ///
+    /// `inbound_frame` may store perfectly contiguous data as several *adjacent*
+    /// `data_ranges` entries once an entry reaches `RANGE_TARGET` bytes (a
+    /// deliberate optimization that avoids repeatedly reallocating a growing
+    /// `Vec`). As a result `data_ranges.len()` overcounts: it cannot tell a
+    /// genuine gap apart from a split of otherwise-contiguous data. This returns
+    /// the number of maximal contiguous runs (entries separated by an actual
+    /// gap), which is the quantity the anti-fragmentation limit must bound.
+    fn discontiguous_ranges(&self) -> usize {
```
martinthomson: I tend to agree with the bot on this one.  How about you hand out extra credit whenever we deliberately choose to add a new item that is contiguous.  Subtract one, saturating, from that every time we completely release a chunk through a read.  Reads will always be contiguous, so we will always subtract more than we add, but that's fine.  This is an anti-abuse measure and we are OK being a little sloppy in the accounting as long as an abusing peer cannot exceed our limits.  If we only give out credit for large ranges, not when we in-fill gaps, the amount of extra credit could be low, but we'd never end up really badly short.

Then your test is `data_ranges.len() > MAX_RANGES + bonus_contiguous_ranges`.  O(1) always.

In practice, the number of ranges we track should be pretty low, but we can afford to track a few more than 2k ranges, which would make this even more generous for high bandwidth scenarios.

neqo-transport/src/recv_stream.rs
```diff
@@ -1056,6 +1089,12 @@ impl RecvStream {
             .is_some_and(RxStreamOrderer::data_ready)
     }
 
+    fn discontiguous_ranges(&self) -> usize {
```
martinthomson: Why is the rejection implemented on this struct and not `recv_buf`/`RxStreamOrderer`?  That has all the logic about range collapsing and whatnot.  It could do the check in `RxStreamOrderer::inbound_frame`.  Then you aren't exposing private logic outside the encapsulation of the type.

martinthomson (commented):
WDYT about the different logic I suggest below?  Having an O(n^2) exposure on every insert is not ideal.  The alternative I suggest is a little harder to write tests for, but I think it works.

neqo-transport/src/recv_stream.rs
larseggert: The best way I can see this work is by changing the `inbound_frame` API to make it return an `Error`. This would be somewhat noisy. I'm making the change, but just FYI.

neqo-transport/src/recv_stream.rs
```diff
@@ -198,6 +200,9 @@ impl RxStreamOrderer {
     /// extended chunk can end up slightly larger than `RANGE_TARGET` (by up to one frame's worth).
     const RANGE_TARGET: usize = 4096;
 
+    /// Maximum number of non-adjacent ("split") ranges allowed.
+    const MAX_SPLIT_RANGES: usize = 4096;
```
martinthomson, line 207: Are we comfortable with a 16MB cap on BDP?  Because that's what this means.

martinthomson (approved):

neqo-transport/src/recv_stream.rs
larseggert: Only if nothing reads? Oh, and also: this limit is only on **non-adjacent** entries - you'd expect most normal entries to be adjacent.

martinthomson: Yeah, of course.

martinthomson:
@larseggert, I think that we settled this one.  Any reason why you aren't merging it?

larseggert:
I'm still looking into the bench performance regression, which was stalled by needing to get base vs. PR profiles from CI.

larseggert:
39977acad87aa34fd30ccd7900405ec9875df59b adds a lot of noise to the diff due to introducing early returns to reduce indentation.

larseggert:
Closing in favor of #3896

PR #3799, opened 2026-07-07
fix: limit frame lengths before buffering

larseggert:
Reject lengths over the existing QPACK decoded-size limit up front.
https://bugzilla.mozilla.org/show_bug.cgi?id=2025290

### DIFF
--- a/neqo-http3/src/frames/capsule.rs
+++ b/neqo-http3/src/frames/capsule.rs
@@ -11,6 +11,10 @@ use crate::Res;
 
 pub const CAPSULE_TYPE_DATAGRAM: HFrameType = HFrameType(0x00);
 
+/// Limit on the declared length of a `DATAGRAM` capsule we'll buffer before decoding.
+pub const MAX_DATAGRAM_BYTES: usize =
+    neqo_common::to_usize(neqo_transport::MAX_DATAGRAM_FRAME_SIZE);
+
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum Capsule {
     Datagram { payload: Bytes },
@@ -48,6 +52,14 @@ impl FrameDecoder<Self> for Capsule {
 
     fn is_known_type(frame_type: HFrameType) -> bool {
         frame_type == CAPSULE_TYPE_DATAGRAM
+    }
+
+    fn max_frame_data(frame_type: HFrameType) -> usize {
+        if frame_type == CAPSULE_TYPE_DATAGRAM {
+            MAX_DATAGRAM_BYTES
+        } else {
+            usize::MAX
+        }
     }
 }
 
@@ -115,6 +127,11 @@ mod tests {
     fn decode_unknown_capsule_type() {
         let res = Capsule::decode(HFrameType(0x17), 4, Some(&[0xaa, 0xbb, 0xcc, 0xdd])).unwrap();
         assert_eq!(res, None);
+    }
+
+    #[test]
+    fn max_frame_data_unknown_type_is_unbounded() {
+        assert_eq!(Capsule::max_frame_data(HFrameType(0x17)), usize::MAX);
     }
 
     #[test]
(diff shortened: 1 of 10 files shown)


neqo-http3/src/frames/wtframe.rs
```diff
@@ -86,6 +97,14 @@ mod tests {
         )));
     }
 
+    #[test]
+    fn max_frame_data_unknown_type_is_unbounded() {
+        assert_eq!(
+            WebTransportFrame::max_frame_data(HFrameType(0x1234)),
```
martinthomson: Is this a reserved for greasing type?  If it ever becomes real, then that's a problem.  How about 0x1230?  That IS a greasing type that will never have another size.

Reference: https://datatracker.ietf.org/doc/html/rfc9297#section-5.4-3

neqo-http3/src/frames/tests/hframe.rs
```diff
@@ -11,10 +11,15 @@ use test_fixture::fixture_init;
...
 use crate::{
     Priority, PushId,
-    frames::HFrame,
+    frames::{HFrame, HFrameType, reader::FrameDecoder as _},
     settings::{HSetting, HSettingType, HSettings},
 };
 
+#[test]
+fn max_frame_data_unknown_type_is_unbounded() {
+    assert_eq!(HFrame::max_frame_data(HFrameType(0x1234)), usize::MAX);
```
martinthomson:
```suggestion
    assert_eq!(HFrame::max_frame_data(HFrameType(0x124b)), usize::MAX);
```

See below.

neqo-http3/src/frames/tests/reader.rs
```diff
@@ -129,6 +154,95 @@ fn frame_reading_with_stream_push_promise() {
     }
 }
 
+// Oversized HEADERS/PUSH_PROMISE lengths are rejected before any payload is buffered.
+#[test]
+fn headers_frame_length_exceeds_cap_rejected() {
```
martinthomson: You should DRY this up a little.  Many of these can be folded into a loop.  For example, one with:

```rust
const CHECKS: _ = &[(MAX_HEADER_BYTES, &[HFrameType::HEADERS, HFrameType::PUSH_PROMISE]), ... ];
```

neqo-http3/src/frames/capsule.rs
```diff
@@ -117,6 +129,11 @@ mod tests {
         assert_eq!(res, None);
     }
 
+    #[test]
+    fn max_frame_data_unknown_type_is_unbounded() {
+        assert_eq!(Capsule::max_frame_data(HFrameType(0x17)), usize::MAX);
```
martinthomson, line 134: Is it a problem that a Capsule is not a Frame?

neqo-http3/src/frames/hframe.rs
```diff
@@ -15,6 +15,21 @@ use crate::{Error, Priority, PushId, Res, frames::reader::FrameDecoder, settings
...
+/// decoded. Worst-case memory is bounded by the transport's stream concurrency limit, roughly
+/// `max_streams_bidi * MAX_HEADER_BYTES`.
+pub const MAX_HEADER_BYTES: usize = neqo_qpack::reader::LiteralReader::MAX_LEN;
+
+/// Limit for frame types that carry at most a single varint.
+pub const MAX_SINGLE_VARINT_FRAME_BYTES: usize = 8;
+
+/// Limit for other buffered frame types (`SETTINGS`, `PRIORITY_UPDATE_*`).
+/// Matches Google quiche's `kPayloadLengthLimit`.
+pub const MAX_BUFFERED_FRAME_BYTES: usize = 1024 * 1024;
```
martinthomson: That's awfully large.  I think that the frames you describe could be a lot smaller than this.  I'd make it a handful of kilobytes instead.  Like 4k.

larseggert: Can do. I didn't have a good feeling for what would be reasonable here, so I went with what Chrome does.

neqo-http3/src/frames/capsule.rs
larseggert: Yeah, this is an existing ugliness in the current code. I'm not sure _this_ is the PR to fix that though?

martinthomson (approved):
Very good.

PR #3801, opened 2026-07-07
chore: Drop msquic from perfcompare

larseggert:
It's too noisy.


martinthomson (changes requested):
Don't comment it out, cut it out.

PR #3802, opened 2026-07-08
reject PRIORITY_UPDATE for an unpromised push id

uwezkhan:
On the server, a PRIORITY_UPDATE for a push (type 0xf0701) arriving on the client control stream had its element_id coerced straight into a StreamId and delivered as a priority-update event, with no check on the id. The request variant beside it (0xf0700) already validates its element_id and returns H3_ID_ERROR for anything that is not a client-initiated bidi request stream, so a client could send a push-variant frame carrying any value and have it accepted as a stream priority signal.

RFC 9218 Section 7.2 makes a push-variant PRIORITY_UPDATE that names a Push ID above the maximum, or one never promised, an H3_ID_ERROR. This server never promises a push (it sends no PUSH_PROMISE and already rejects CANCEL_PUSH), so no Push ID is valid and the frame is now rejected with Error::HttpId, matching the request arm. Before, every push-variant frame was accepted and surfaced a bogus StreamId; after, it closes the connection. The tradeoff is that once server push lands this needs to compare against the real maximum Push ID instead of rejecting outright, but until then unconditional rejection is the only correct behavior.

### DIFF
--- a/neqo-http3/src/connection_server.rs
+++ b/neqo-http3/src/connection_server.rs
@@ -361,15 +361,13 @@ impl Http3ServerHandler {
                         HFrame::Goaway { .. } | HFrame::CancelPush { .. } => {
                             Err(Error::HttpFrameUnexpected)
                         }
-                        HFrame::PriorityUpdatePush {
-                            element_id,
-                            priority,
-                        } => {
-                            // TODO: check if the element_id references a promised push stream or
-                            // is greater than the maximum Push ID.
-                            self.events
-                                .priority_update(StreamId::from(element_id), priority);
-                            Ok(())
+                        HFrame::PriorityUpdatePush { .. } => {
+                            // RFC 9218, Section 7.2: the push variant of PRIORITY_UPDATE must
+                            // reference a promised push stream, and one referencing a Push ID
+                            // greater than the maximum (or never promised) is a connection error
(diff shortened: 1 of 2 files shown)


larseggert (commented):
LGTM.

@uwezkhan a general comment: Our server is mostly a toy, so when you pick which issue to work on, fixes to client-side issues are much more valuable to us.

larseggert (approved):

uwezkhan:
Understood, I'll focus on client-side issues going forward.

mxinden:
Note that I am suggesting to remove all of our Server Push logic in https://github.com/mozilla/neqo/pull/3816.

uwezkhan:
Makes sense. The difference is just how the frame dies: this PR keeps PriorityUpdatePush and closes the connection with H3_ID_ERROR, while #3816 removes the type entirely, so 0xf0701 gets ignored as an unknown frame per RFC 9114. Both are conformant since an endpoint that doesn't implement the extension ignores unknown frames. If #3816 lands first this one is superseded and can just be closed.

larseggert (approved):

larseggert:
#3816 has merged, I assume this is now moot?

martinthomson:
Yep.

PR #3803, opened 2026-07-08
feat: add neqo-server `--stats`, emit connection stats as JSON

larseggert:
Server now supports `--stats` (mirroring the client), printing a connection's stats once when it closes. Both client and server serialize stats via serde instead of a hand-maintained Debug formatter, which was silently missing several fields. Output is compact (short arrays/objects inlined, keys sorted, None fields omitted) and can optionally be appended to a file instead of logged.

The motivation here is that I wanted to see if the stats from CI run outliers gave an indication for why they were outliers and realized we (1) didn't save the stats and (2) even if we did I needed a custom parser for the data.


neqo-bin/src/lib.rs
```diff
@@ -291,6 +293,24 @@ fn now() -> Instant {
...
 
+/// Report `stats` as compact JSON: appended to `path` if given, or logged
+/// via `qinfo!` otherwise.
+///
+/// # Errors
+///
+/// If `path` is given and the file can't be opened or written to.
+pub fn report_stats(stats: &Stats, path: Option<&Path>) -> std::io::Result<()> {
+    let json = json::compact(stats);
+    let Some(path) = path else {
```
martinthomson, line 319: Given the line counts, this would look better as `if let Some(path) = path { writeln!(...) } else { qinfo(...); Ok((()) }`

neqo-bin/src/client/mod.rs
```diff
@@ -160,9 +160,11 @@ pub struct Args {
     #[arg(name = "upload-size", long, default_value = "100")]
     upload_size: usize,
 
-    /// Print connection stats after close.
+    /// Print connection stats after close. Optionally give a filename to
+    /// append the stats to (as JSON), instead of logging them.
     #[arg(name = "stats", long)]
-    stats: bool,
+    #[expect(clippy::option_option, reason = "clap shape for flag with opt value")]
+    stats: Option<Option<PathBuf>>,
```
martinthomson: Can't we have common options (and code) for both binaries?

neqo-common/src/json.rs
```diff
@@ -0,0 +1,164 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! JSON formatting shared by callers that print structured stats.
```
martinthomson: If this is only used by the neqo-bin binaries, maybe it could live there...

larseggert: Also used in neqo-transport tests and test-fixture.

larseggert:
Could I get a review here?

martinthomson (commented):
EDIFFTOOBIG, for me and this week at least.

mxinden (commented):
Generally in favor of printing json.

neqo-bin/src/server/mod.rs
```diff
@@ -211,6 +212,39 @@ impl Args {
...
+impl<T> Default for ReportedConnections<T> {
+    fn default() -> Self {
+        Self(Vec::new())
+    }
+}
+
+impl<T> ReportedConnections<T> {
+    /// Records `conn` as reported. Returns `true` the first time a given
+    /// connection is seen, `false` on every subsequent call for the same one.
+    pub(super) fn insert(&mut self, conn: &Rc<RefCell<T>>) -> bool {
```
martinthomson: Rather than have this as a simple insertion, I think that you should have it hold the `report_stats` function.  Then, that code can include a short circuit for when the connection has already been seen.

neqo-bin/src/server/mod.rs
```diff
@@ -211,6 +212,39 @@ impl Args {
...
+        self.0.retain(|w| w.strong_count() > 0);
+        if self.0.iter().any(|w| ptr::eq(w.as_ptr(), Rc::as_ptr(conn))) {
```
martinthomson: Rather than using `ptr_eq` for connection comparison, I think that you can use the initial connection ID, which can be obtained using `Connection::odcid()`.  That's an `Option`, so you can fail to report stats for a connection that didn't have a value.

neqo-bin/src/server/http09.rs
```diff
@@ -230,6 +237,14 @@ impl super::HttpServer for HttpServer {
                             .send_ticket(now, b"hi!")
                             .unwrap();
                     }
+                    ConnectionEvent::StateChange(state) if state.closed() => {
+                        if let Some(path) = &self.stats
+                            && self.reported.insert(&acr.connection())
+                            && let Err(e) = report_stats(&acr.borrow().stats(), path.as_deref())
```
martinthomson: I'm not a big fan of if statements with so much side effect.

It will be longer, but I think you want:


```suggestion
                        if let Some(path) = &self.stats {
                            self.reported.report_stats(acr, path.as_deref());
                        }
```

This is a lot shorter than what you have here, because I'm suggesting that you move the bulk of the logic to `ReportedConnections`.

neqo-bin/src/client/mod.rs
```diff
@@ -460,8 +455,10 @@ impl<'a, H: Handler> Runner<'a, H> {
             }
         }
 
-        if self.args.stats {
-            qinfo!("{:?}", self.client.stats());
+        if let Some(path) = &self.args.shared.stats
+            && let Err(e) = report_stats(&self.client.stats(), path.as_deref())
+        {
+            qerror!("Failed to report stats: {e}");
         }
```
martinthomson: Can't `report_stats()` generate the error message?  Then this would be:


```suggestion
        if let Some(path) = &self.args.shared.stats {
            report_stats(&self.client.stats(), path.as_deref());
        }
```

neqo-common/src/json.rs
```diff
@@ -0,0 +1,165 @@
...
+    let key = serde_json::to_string(key).expect("string always serializes");
+    write!(out, "{key}: ").expect("write! to a String cannot fail");
+}
+
+/// Render `value`, choosing between [`render_inline`] and [`render_block`]
+/// depending on whether the inline form fits within `opts.max_line_length`
+/// once `depth` levels of indentation are accounted for.
+fn render(value: &Value, depth: usize, opts: &Options, out: &mut String) {
+    let mut inline = String::new();
+    render_inline(value, &mut inline);
```
martinthomson: I know that this is not likely to be a problem here, but this is pretty inefficient.  Every time you hit something, you write it out twice.  Recursively.  I think that what you want is for `render_inline` function to return an `Option` so that it can itself detect when it exceeds the line length constraint and abort.  That is still trial and error, but not quite as bad as doing a full serialization.

neqo-common/src/json.rs
```diff
@@ -0,0 +1,165 @@
...
+        out.push(close);
+        return;
+    }
+    out.push(open);
+    out.push('\n');
+    let pad = " ".repeat((depth + 1) * opts.indent);
+    for (i, (key, val)) in entries.enumerate() {
+        out.push_str(&pad);
+        if let Some(k) = key {
+            write_key(out, k);
```
martinthomson: You aren't subtracting the length of the key from the offset that `render()` adds.  That extra is added to the line length invisibly by the block on line 56-61.

neqo-common/src/json.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+        render(val, depth + 1, opts, out);
+        if i + 1 < n {
+            out.push(',');
```
martinthomson: This isn't counted toward the line length either.

martinthomson (commented):
I like the general shape of this, but I have some structural suggestions.  The stats saving could be simpler.

The JSON serializer seems like it isn't as rigorous as it could be as well.

neqo-bin/src/server/mod.rs
larseggert: IIRC the issue with that was that there isn't really a connection closing trigger then and we couldn't remove from the `HashSet`. But I guess our server is just a toy and we don't care whether the `HashSet` never shrinks?

larseggert:
@martinthomson I removed the JSON condensing code, it became too noisy. And I don't think adding a new dep to neqo and Gecko just for that is worth it.

larseggert:
I would really like to merge this. There are a bunch of cases where debugging without access to server-side stats is a big limitation. @mxinden @martinthomson

test-fixture/src/lib.rs
```diff
@@ -301,6 +303,75 @@ pub fn connect() -> (Connection, Connection) {
     (client, server)
 }
 
+/// The part of a multi-connection server that [`handshake_with_server`] drives.
+///
+/// The method names differ from the inherent ones they forward to: sharing them
+/// would leave the bodies below resolving to the inherent methods only by
+/// preference, so renaming or extending either one would turn these into
+/// unbounded recursion rather than a compile error.
```
martinthomson:
```suggestion
```

LLMs have poor judgment about what to put in comments.  Of course that is how this works.

Also, this doesn't matter, because in places where this is `&dyn ProcessServer`, the concrete type methods are not available.  Because the implementations are direct passthroughs with no signature change, it is totally fine to reuse the names.

neqo-bin/src/lib.rs
```diff
@@ -291,26 +300,65 @@ fn now() -> Instant {
...
 
     impl TempDir {
         fn new() -> Self {
+            // Tests run concurrently, so the name has to be unique per instance:
+            // sharing one means the first `Drop` deletes the other's directory.
+            static SEQ: AtomicUsize = AtomicUsize::new(0);
             let dir = std::env::temp_dir().join(format!(
-                "neqo-bin-test-{}",
+                "neqo-bin-test-{}-{}",
                 SystemTime::now()
```
martinthomson: A little crazy (but not that surprising, really) that you had a ns-level collision and needed to add the counter.  In that case, I think you can probably drop this back to ms rather than ns.

neqo-bin/src/server/mod.rs
```diff
@@ -211,6 +213,47 @@ impl Args {
...
+    pub(super) fn report(&mut self, conn: &Rc<RefCell<Connection>>) {
+        let Some(path) = &self.path else {
+            return;
+        };
+        let conn_weak = Rc::downgrade(conn);
+        self.reported.retain(|w| w.strong_count() > 0);
+        if self.reported.iter().any(|w| Weak::ptr_eq(w, &conn_weak)) {
+            return;
+        }
+        self.reported.push(conn_weak);
```
martinthomson: This is almost too clever for me.  Damn those crafty LLMs.

neqo-bin/src/server/mod.rs
```diff
@@ -211,6 +213,47 @@ impl Args {
...
+pub(super) struct StatsReporter {
+    #[expect(clippy::option_option, reason = "clap flag-with-optional-value shape")]
+    path: Option<Option<PathBuf>>,
+    /// Connections already reported. `Weak`, so that entries for connections
+    /// the server has since dropped can be pruned.
+    reported: Vec<Weak<RefCell<Connection>>>,
+}
+
+impl StatsReporter {
+    pub(super) fn new(args: &SharedArgs) -> Self {
```
martinthomson: I think that you should pass the OptionOption in rather than take a dependency on the entire SharedArgs structure.  Also, I saw comments about two args.  I'd prefer that: `enable: bool, path: Option<PathBuf>`.

neqo-bin/src/server/http3.rs
```diff
@@ -217,3 +224,28 @@ impl super::HttpServer for HttpServer {
...
+    fn make_server(args: &Args) -> HttpServer {
+        fixture_init();
+        HttpServer::new(
+            args,
+            anti_replay(),
+            Rc::new(RefCell::new(RandomConnectionIdGenerator::new(10))),
+        )
+    }
+
+    stats_tests!(make_server);
```
martinthomson: NO.  This can't be.

neqo-bin/src/server/mod.rs
```diff
@@ -605,14 +648,84 @@ pub fn run(
...
+    macro_rules! stats_tests {
+        ($make:ident) => {
+            impl StatsServer for HttpServer {
+                fn transport(&mut self) -> &mut dyn ProcessServer {
+                    &mut self.server
+                }
+
+                fn stats(&self) -> &StatsReporter {
+                    &self.stats
+                }
```
martinthomson: A tiny bit of copy-pasta isn't so bad.  The trait decoration stuff would allow this to be largely shared implementation, as it is already.  The duplicated code would be a single line test function.

martinthomson (approved):
I don't like your macro.  It hardly seems worth the effort as well.

PR #3805, opened 2026-07-09
fix: reject server push when the maximum push id is unset

uwezkhan:
Default client (`max_concurrent_push_streams` = 0), server sends a PUSH_PROMISE with push id 0:

    check_push_id(0):  0 > current_max_push_id(0)  ->  false  ->  accepted

A client that does not enable push never sends a MAX_PUSH_ID frame, so its maximum push id is unset. RFC 9114 (7.2.3, 7.2.7) makes a PUSH_PROMISE or CANCEL_PUSH carrying a push id above what the client advertised an H3_ID_ERROR, and with nothing advertised every push id, including 0, is above the maximum. `current_max_push_id` is a plain `PushId` initialized to 0, so the `>` check cannot tell "unset" from "0 allowed" and admits push id 0. `check_push_id` is the single gate for the PUSH_PROMISE, CANCEL_PUSH, and push-stream paths.

Before: a push-disabled client accepts a PUSH_PROMISE or CANCEL_PUSH with push id 0, decodes the header block, and reports a PushPromise event to the application.

After: `check_push_id` rejects any push id while push is disabled and closes the connection with H3_ID_ERROR. `can_receive_push()` is false exactly when no MAX_PUSH_ID is ever sent, so the guard matches the unset-maximum state and leaves the enabled path (push id vs the advertised maximum) unchanged.

Tradeoff: strict rejection over silently dropping an unsolicited push. A push-enabled client is unaffected because `can_receive_push()` is true for it.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -5630,6 +5630,80 @@ mod tests {
         assert_eq!(client.state(), Http3State::Connected);
     }
 
+    // A client that does not enable server push never sends a MAX_PUSH_ID frame,
+    // so its maximum Push ID stays unset and every Push ID is above it. The
+    // default connect helpers assume a MAX_PUSH_ID frame on the control stream,
+    // so drive the setup here and drain the control stream without that check.
+    fn connect_push_disabled_and_send_request() -> (Http3Client, TestServer, StreamId) {
+        fixture_init();
+        let mut client = Http3Client::new(
+            DEFAULT_SERVER_NAME,
+            Rc::new(RefCell::new(CountingConnectionIdGenerator::default())),
+            DEFAULT_ADDR,
+            DEFAULT_ADDR,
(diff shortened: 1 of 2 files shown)


larseggert (approved):

uwezkhan:
gentle ping

martinthomson:
We're going to remove the affected code (#3816), I think.  No sense fixing it if that is the case.

uwezkhan:
Makes sense. I checked #3816: after the removal, PUSH_PROMISE on a request stream and CANCEL_PUSH on the control stream both fall through to HttpFrameUnexpected, so a push-disabled client closes the connection either way, just with H3_FRAME_UNEXPECTED instead of H3_ID_ERROR. That covers what this PR was after. I'll close this once #3816 lands.

larseggert:
#3816 has merged

PR #3806, opened 2026-07-11
reject connection-specific header fields in HTTP/3 messages

uwezkhan:
headers_valid() scans received field names and values for illegal bytes but never rejects the connection-specific field names themselves, so a server response carrying Connection, Transfer-Encoding, Upgrade, Keep-Alive, or Proxy-Connection is accepted. RFC 9114 Section 4.2 makes any message that contains one of these malformed. Today recv_message quietly strips the field with a retain() filter and delivers the rest, which is what malformed_response_excluded_header asserted even though its own comment claimed the stream was reset.

Before, the field was dropped and the response delivered; after, headers_valid rejects it so the stream resets with H3_MESSAGE_ERROR, matching the Section 4.3 CR/LF/NUL checks sitting next to it. The check lives in headers_valid because that is the one point both the client response path and the server request path pass through, and it keeps the send path honest through the existing debug assertion. The tradeoff is strictness over leniency: TE stays allowed, but only with the value trailers.

### DIFF
--- a/neqo-http3/src/headers_checks.rs
+++ b/neqo-http3/src/headers_checks.rs
@@ -124,6 +124,18 @@ pub fn headers_valid(headers: &[Header], message_type: MessageType) -> Res<()> {
         if header.value().iter().any(|b| matches!(b, 0 | 0x0a | 0x0d)) {
             return Err(Error::InvalidHeader); // illegal characters.
         }
+
+        // Connection-specific header fields make a message malformed (RFC 9114,
+        // Section 4.2). TE is the only one that may appear, and only with the
+        // value "trailers". Names are already lowercased above.
+        if !is_pseudo {
+            match header.name() {
+                "connection" | "keep-alive" | "proxy-connection" | "transfer-encoding"
+                | "upgrade" => return Err(Error::InvalidHeader),
+                "te" if header.value() != b"trailers" => return Err(Error::InvalidHeader),
+                _ => {}
+            }
+        }
     }
     // Clear the regular header bit, since we only check pseudo headers below.
     pseudo_state.remove(PseudoHeaderState::Regular);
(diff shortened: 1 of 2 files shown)


neqo-http3/src/headers_checks.rs
```diff
@@ -124,6 +124,18 @@ pub fn headers_valid(headers: &[Header], message_type: MessageType) -> Res<()> {
         if header.value().iter().any(|b| matches!(b, 0 | 0x0a | 0x0d)) {
             return Err(Error::InvalidHeader); // illegal characters.
         }
+
+        // Connection-specific header fields make a message malformed (RFC 9114,
+        // Section 4.2). TE is the only one that may appear, and only with the
+        // value "trailers". Names are already lowercased above.
+        if !is_pseudo {
```
martinthomson: This outer is_pseudo check isn't needed.  The question would then be whether this check avoids any performance cost from the extra comparisons.  I'd think not.

neqo-http3/src/headers_checks.rs
```diff
@@ -348,6 +360,47 @@ mod tests {
...
+    #[test]
+    fn reject_connection_specific_header() {
+        // A message carrying a connection-specific header field is malformed
+        // (RFC 9114, Section 4.2), in either direction.
+        for name in [
+            "connection",
+            "keep-alive",
+            "proxy-connection",
+            "transfer-encoding",
+            "upgrade",
```
martinthomson, line 371:
```suggestion
            "upgrade",
            "te",
```

neqo-http3/src/headers_checks.rs
```diff
@@ -348,6 +360,47 @@ mod tests {
...
+        ];
+        assert!(headers_valid(&ok, MessageType::Request).is_ok());
+
+        let bad = vec![
+            Header::new(":method", "GET"),
+            Header::new(":scheme", "https"),
+            Header::new(":path", "/"),
+            Header::new("te", "gzip"),
+        ];
+        assert!(headers_valid(&bad, MessageType::Request).is_err());
```
martinthomson: See above.
```suggestion
```

martinthomson (approved):

neqo-http3/src/headers_checks.rs
uwezkhan: Done, dropped it. Pseudo-header names start with a colon so the match can never hit them, and the extra comparisons are noise next to the byte scans above.

neqo-http3/src/headers_checks.rs
uwezkhan: Done, te is in the loop now.

neqo-http3/src/headers_checks.rs
uwezkhan: Dropped the separate gzip block; the loop covers the non-trailers value. The trailers case stays as the lone Ok check.

PR #3807, opened 2026-07-13
fix: reject qpack dynamic reference at or above required insert count

uwezkhan:
2 inserts (absolute index 0, 1); block declares RIC=1, Base=2, then 0x80
    -> reference resolves to absolute index 1 (>= RIC)
    -> decode_header_block returns Ok(Headers), not QPACK_DECOMPRESSION_FAILED

decode_header_block validates each dynamic reference only through get_dynamic, which confirms the entry is still in the live table but never checks it against the block's own Required Insert Count. RFC 9204 Section 2.2.3 requires a reference whose absolute index is at or above the Required Insert Count to be a decoding error, because the encoder committed to exactly that many entries; the entry can still be present when later insertions arrive on the encoder stream.

Before: a client decoding a response or push header block accepts such a reference and returns headers whose value depends on insertions the block never declared.

After: the reference is rejected with Decompression (QPACK_DECOMPRESSION_FAILED), the same as every other malformed reference in the block. References within the count decode unchanged.

The check sits at the reference resolution in the decoder, wrapping the four dynamic reads, since that is the only layer holding both the resolved absolute index and req_insert_cnt. It also covers Section 2.2.1, since a Required Insert Count smaller than the block needs surfaces as a reference at or above it. The tradeoff is one more decode that fails closed against trusting an over-optimistic Required Insert Count.

### DIFF
--- a/neqo-qpack/src/header_block.rs
+++ b/neqo-qpack/src/header_block.rs
@@ -23,7 +23,7 @@ use crate::{
     },
     qpack_send_buf::Encoder as _,
     reader::{LiteralReader, ReceiverBufferWrapper, parse_utf8},
-    table::{ADDITIONAL_TABLE_ENTRY_SIZE, HeaderTable},
+    table::{ADDITIONAL_TABLE_ENTRY_SIZE, DynamicTableEntry, HeaderTable},
 };
 
 #[derive(Default, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


neqo-qpack/src/header_block.rs
```diff
@@ -316,12 +316,30 @@ impl<'a> HeaderDecoder<'a> {
         Ok(Header::new(parse_utf8(entry.name())?, entry.value()))
     }
 
+    // Resolve a dynamic-table reference from a field line representation. A reference
+    // whose absolute index is at or above the Required Insert Count is a decoding error
+    // per RFC 9204, Section 2.2.3, even when the entry is still present in the table
+    // because later insertions arrived. This also covers Section 2.2.1: such a reference
+    // means the declared Required Insert Count was smaller than the block actually needs.
```
martinthomson:
```suggestion
    /// Resolve a dynamic-table reference from a field line representation. A reference
    /// whose absolute index is at or above the Required Insert Count is a decoding error
    /// per RFC 9204, Section 2.2.3, even when the entry is still present in the table
    /// because later insertions arrived. This also covers Section 2.2.1: such a reference
    /// means the declared Required Insert Count was smaller than the block actually needs.
```

Might as well use doc comments.

martinthomson (approved):

martinthomson:
Let's hold this one.  We might just rip server push out entirely.

uwezkhan:
Sure. One note for when you come back to it: the check isn't push-specific. It sits in decode_header_block, which the client runs on every header block it decodes, so plain response headers on request streams hit the same path. Push promises were just the second caller, so this stays relevant if push goes away with #3816.

martinthomson:
Oops, I meant to add that comment to a different PR.  Ignore that.  I'll merge this when CI finishes.

PR #3812, opened 2026-07-15
usize conversions

martinthomson:
Audit these.

No doubt I missed something here, but I think that this is an improvement over what we had.

### DIFF
--- a/neqo-http3/src/push_controller.rs
+++ b/neqo-http3/src/push_controller.rs
@@ -13,7 +13,7 @@ use std::{
     slice::SliceIndex,
 };
 
-use neqo_common::{Header, qerror, qinfo, qtrace, to_u64, expect_usize};
+use neqo_common::{Header, expect_usize, qerror, qinfo, qtrace, to_u64};
 use neqo_transport::{Connection, StreamId};
 
 use crate::{
--- a/neqo-http3/src/push_controller.rs
+++ b/neqo-http3/src/push_controller.rs
@@ -165,6 +165,7 @@ impl Display for PushController {
 
 impl PushController {
     pub const fn new(max_concurent_push: u64, conn_events: Http3ClientEvents) -> Self {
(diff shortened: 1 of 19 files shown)


larseggert (approved):

neqo-http3/src/push_controller.rs
```diff
@@ -79,7 +79,7 @@ impl ActivePushStreams {
             return None;
         }
 
-        let inx = usize::try_from(u64::from(push_id - self.first_push_id)).ok()?;
+        let inx = expect_usize(u64::from(push_id - self.first_push_id));
```
mxinden, line 82: Should `PushId` be a `u32` instead? If yes, doesn't need to happen in this pull request.

neqo-transport/src/fc.rs
```diff
@@ -115,15 +115,16 @@ where
...
         self.used += amt;
     }
 
     /// Get available flow control.
-    pub fn available(&self) -> usize {
-        usize::try_from(self.limit - self.used).unwrap_or(usize::MAX)
+    #[expect(clippy::cast_possible_truncation, reason = "value is capped")]
+    pub const fn available(&self) -> usize {
+        const_min_u64(self.limit - self.used, to_u64(usize::MAX)) as usize
     }
```
mxinden, line 128: Nit, I don't think `available` being `const` is important. With that in mind, I believe this is simpler:


```suggestion
    pub const fn available(&self) -> usize {
        usize::try_from(self.limit - self.used).unwrap_or(usize::MAX)
    }
```

mxinden (approved):
Thank you, especially for the `static_assertions`.

neqo-http3/src/push_controller.rs
mxinden: Let's remove it entirely: https://github.com/mozilla/neqo/pull/3816

PR #3814, opened 2026-07-16
fix: discard server Initial packets that carry a token

uwezkhan:
preprocess_packet validates a received Initial's Token Length only in the server arm, through is_valid_initial. A client-received Initial matches none of those arms, so the token that decode_long already parsed is ignored and the packet goes on to be decrypted and processed. RFC 9000, Section 17.2.2 says a server MUST set the Token Length of an Initial to 0, and a client that receives an Initial with a non-zero Token Length MUST discard it or close with PROTOCOL_VIOLATION. A server, or an off-path injector that derived the public Initial keys, can hand the client an Initial that carries a token and neqo accepts it.

Before: the client silently accepted the illegal token. After: preprocess_packet drops the packet as soon as a client sees an Initial with a non-empty token, beside the existing client Initial/Handshake path check so coalesced and replayed packets are covered too. Discarding is the safer of the two behaviors the RFC permits here, because the token length sits outside packet protection and closing on it would let an off-path packet end the connection; the tradeoff is that a genuinely misbehaving server gets no error signal back.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1543,6 +1543,21 @@ impl Connection {
             return Ok(PreprocessResult::Next);
         }
 
+        // RFC 9000, Section 17.2.2: a server MUST set the Token Length field of an
+        // Initial to 0, and a client that receives an Initial with a non-zero Token
+        // Length MUST discard the packet or close with PROTOCOL_VIOLATION. Discarding
+        // is preferred here: the token length sits in the unprotected header, so an
+        // off-path injection must not be able to tear down the connection.
+        if packet.packet_type() == packet::Type::Initial
+            && self.role == Role::Client
+            && !packet.token().is_empty()
+        {
+            self.stats
+                .borrow_mut()
+                .pkt_dropped("Client received an Initial with a token");
+            return Ok(PreprocessResult::Next);
+        }
+
         match (packet.packet_type(), &self.state, &self.role) {
             (packet::Type::Initial, State::Init, Role::Server) => {
                 let version = packet.version().ok_or(Error::ProtocolViolation)?;
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1543,22 +1543,18 @@ impl Connection {
             return Ok(PreprocessResult::Next);
         }
 
-        // RFC 9000, Section 17.2.2: a server MUST set the Token Length field of an
-        // Initial to 0, and a client that receives an Initial with a non-zero Token
-        // Length MUST discard the packet or close with PROTOCOL_VIOLATION. Discarding
-        // is preferred here: the token length sits in the unprotected header, so an
-        // off-path injection must not be able to tear down the connection.
-        if packet.packet_type() == packet::Type::Initial
-            && self.role == Role::Client
-            && !packet.token().is_empty()
-        {
-            self.stats
-                .borrow_mut()
-                .pkt_dropped("Client received an Initial with a token");
-            return Ok(PreprocessResult::Next);
-        }
-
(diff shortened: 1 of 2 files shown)


neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2079,3 +2079,66 @@ fn retry_scid_matching_initial_dcid() {
...
+    // A token-free server Initial decrypts and is processed (also confirming the
+    // forged packet above is well-formed).
+    let mut client = default_client();
+    let ci = client
+        .process_output(now())
+        .dgram()
+        .expect("a datagram")
+        .to_vec();
+    let dropped = client.stats().dropped_rx;
+    drop(client.process(Some(datagram(server_initial(&ci, &[]))), now()));
```
martinthomson:
```suggestion
    client.process_input(datagram(server_initial(&ci, &[])), now());
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1543,6 +1543,21 @@ impl Connection {
...
+        // off-path injection must not be able to tear down the connection.
+        if packet.packet_type() == packet::Type::Initial
+            && self.role == Role::Client
+            && !packet.token().is_empty()
+        {
+            self.stats
+                .borrow_mut()
+                .pkt_dropped("Client received an Initial with a token");
+            return Ok(PreprocessResult::Next);
+        }
```
martinthomson: This can be moved into the match below.  The pattern ignores the state, but that's not so bad.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2079,3 +2079,66 @@ fn retry_scid_matching_initial_dcid() {
...
+
+    // The same Initial carrying a non-empty token is discarded.
+    let mut client = default_client();
+    let ci = client
+        .process_output(now())
+        .dgram()
+        .expect("a datagram")
+        .to_vec();
+    let dropped = client.stats().dropped_rx;
+    drop(client.process(Some(datagram(server_initial(&ci, &[0x01]))), now()));
```
martinthomson:
```suggestion
    client.process(datagram(server_initial(&ci, &[0x01])), now());
```

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2079,3 +2079,66 @@ fn retry_scid_matching_initial_dcid() {
...
+        let version = Version::default();
+        let mut write =
+            CryptoDxState::new_initial(version, CryptoDxDirection::Write, "server in", &seed, 0)
+                .expect("initial keys");
+        let mut builder = crate::packet::Builder::long(
+            Encoder::default(),
+            crate::packet::Type::Initial,
+            version,
+            Some(dcid.as_slice()),
+            Some(&[0x33; 8][..]),
```
martinthomson: I think that you can do something simpler.
```suggestion
            Some(b"fakescid"),
```

neqo-transport/src/connection/mod.rs
uwezkhan: Done, it's an arm in the match now, ignoring the state.

neqo-transport/src/connection/tests/handshake.rs
uwezkhan: Done.

neqo-transport/src/connection/tests/handshake.rs
uwezkhan: process takes an Option, so I used process_input here as well.

neqo-transport/src/connection/tests/handshake.rs
uwezkhan: Done, that's simpler.

mxinden (approved):
Thank you for the help!

PR #3816, opened 2026-07-17
fix(http3): remove server push

mxinden:
[Server push](https://www.rfc-editor.org/rfc/rfc9114.html#section-4.6) is disabled in Firefox: [`neqo_glue`](https://searchfox.org/mozilla-central/source/netwerk/socket/neqo_glue/src/lib.rs) sets `max_concurrent_push_streams` to 0, so neqo never raises [`MAX_PUSH_ID`](https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.7) and no push can ever happen. Firefox has since dropped server push entirely ([Bug 1955565](https://bugzilla.mozilla.org/show_bug.cgi?id=1955565)). The feature is dead in practice, so remove it wholesale: `PushId`, `PushController`, the client push API (`cancel_push`, `push_read_data` and the `Http3ClientEvent` push variants), the `Http3Parameters::max_concurrent_push_streams` builder, and the push stream type.

Also drop [`PRIORITY_UPDATE_PUSH`](https://www.rfc-editor.org/rfc/rfc9218.html#section-7.2), which only prioritizes pushed responses and is unreachable without push.

The `CANCEL_PUSH`, `PUSH_PROMISE` and `MAX_PUSH_ID` frame types and the push stream type (`0x1`) are no longer handled: neqo logs them at debug level and ignores them. Per [RFC 9114](https://www.rfc-editor.org/rfc/rfc9114.html) an endpoint ignores unknown frames and aborts unknown unidirectional streams with `STOP_SENDING`, so a peer that still sends push traffic is handled gracefully rather than crashing.


neqo-http3/src/frames/tests/reader.rs
```diff
@@ -246,16 +208,39 @@ fn unknown_frame() {
...
 
+// Server push frames are no longer supported: they are ignored (not errored), and the reader keeps
+// working afterwards.
+#[test]
+fn server_push_frames_are_ignored() {
+    let push_frames: &[&[u8]] = &[
+        &[0x03, 0x01, 0x05],                         // CANCEL_PUSH push_id=5
+        &[0x05, 0x05, 0x04, 0x61, 0x62, 0x63, 0x64], // PUSH_PROMISE push_id=4
+        &[0x0d, 0x01, 0x05],                         // MAX_PUSH_ID push_id=5
+        &[0x80, 0x0f, 0x07, 0x01, 0x01, 0x0a],       // PRIORITY_UPDATE_PUSH
```
martinthomson: One failing of this approach is that we are not going to catch badly-formatted frames.  If the CANCEL_PUSH frame is 1k in length, we'll ignore it.

That suggests that we might want something like what the servers need: some metric for "useless frames", which tracks when the peer wastes effort and cuts connections that aren't productive.

martinthomson (commented):
It's time.  I skimmed this, which seems about right.  Lots of red.

neqo-http3/src/frames/tests/reader.rs
mxinden: Do you consider this a serious attack? Unless we see this in the wild, I suggest simply ignoring them as we do here for now.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -49,9 +49,12 @@ impl NewStreamType {
...
             (QPACK_UNI_STREAM_TYPE_ENCODER, StreamType::UniDi, _) => Ok(Some(Self::Decoder)),
             (QPACK_UNI_STREAM_TYPE_DECODER, StreamType::UniDi, _) => Ok(Some(Self::Encoder)),
-            (HTTP3_UNI_STREAM_TYPE_PUSH, StreamType::UniDi, Role::Client)
-            | (WEBTRANSPORT_UNI_STREAM, StreamType::UniDi, _)
+            (WEBTRANSPORT_UNI_STREAM, StreamType::UniDi, _)
             | (WEBTRANSPORT_STREAM, StreamType::BiDi, _) => Ok(None),
+            (HTTP3_UNI_STREAM_TYPE_PUSH, StreamType::UniDi, _) => {
+                qdebug!("Ignoring HTTP/3 server push stream (server push is not supported)");
+                Ok(Some(Self::Unknown))
+            }
```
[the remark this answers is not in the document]
mxinden: Test server only. Fine to ignore in my eyes.

mxinden:
@larseggert @martinthomson either of you have time for another review?

neqo-http3/src/frames/hframe.rs
```diff
@@ -178,30 +141,37 @@ impl FrameDecoder<Self> for HFrame {
...
         }
     }
 
     fn decode(frame_type: HFrameType, frame_len: u64, data: Option<&[u8]>) -> Res<Option<Self>> {
         if frame_type == HFrameType::DATA {
             Ok(Some(Self::Data { len: frame_len }))
+        } else if matches!(
+            frame_type,
+            HFrameType::CANCEL_PUSH
+                | HFrameType::PUSH_PROMISE
```
martinthomson: I think that this is the only one that I would consider changing.  If we get a PUSH_PROMISE, we're talking about a server that is ignoring the fact that we haven't enabled server push.  We can treat that as a connection-level error.

neqo-http3/src/stream_type_reader.rs
```diff
@@ -49,9 +49,12 @@ impl NewStreamType {
             (HTTP3_UNI_STREAM_TYPE_CONTROL, StreamType::UniDi, _) => Ok(Some(Self::Control)),
             (QPACK_UNI_STREAM_TYPE_ENCODER, StreamType::UniDi, _) => Ok(Some(Self::Decoder)),
             (QPACK_UNI_STREAM_TYPE_DECODER, StreamType::UniDi, _) => Ok(Some(Self::Encoder)),
-            (HTTP3_UNI_STREAM_TYPE_PUSH, StreamType::UniDi, Role::Client)
-            | (WEBTRANSPORT_UNI_STREAM, StreamType::UniDi, _)
+            (WEBTRANSPORT_UNI_STREAM, StreamType::UniDi, _)
             | (WEBTRANSPORT_STREAM, StreamType::BiDi, _) => Ok(None),
+            (HTTP3_UNI_STREAM_TYPE_PUSH, StreamType::UniDi, _) => {
+                qdebug!("Ignoring HTTP/3 server push stream (server push is not supported)");
+                Ok(Some(Self::Unknown))
```
martinthomson: Is this really what we want?  Because any server push stream we receive is clearly in violation of our request not to use server push.

martinthomson (commented):
Two questions only: do we want to just ignore attempts to use server push, or is that an error condition.  I am leaning towards treating that as a connection error.

larseggert:
Connection error seems appropriate. We should monitor telemetry after landing to see if we get unexpectedly many errors.

larseggert (approved):

neqo-http3/src/connection.rs
```diff
@@ -717,9 +712,6 @@ impl Http3Connection {
                     .insert(stream_id, Box::new(ControlStreamRemote::new(stream_id)));
             }
 
-            NewStreamType::Push(push_id) => {
-                qinfo!("[{self}] A new push stream {stream_id} push_id:{push_id}");
-            }
```
martinthomson, line 725: Do we need to reject a new push stream as well?  I think we probably do.

neqo-http3/src/frames/hframe.rs
```diff
@@ -178,30 +141,39 @@ impl FrameDecoder<Self> for HFrame {
...
     }
 
     fn decode(frame_type: HFrameType, frame_len: u64, data: Option<&[u8]>) -> Res<Option<Self>> {
         if frame_type == HFrameType::DATA {
             Ok(Some(Self::Data { len: frame_len }))
+        } else if frame_type == HFrameType::MAX_PUSH_ID {
+            // Server push is not supported. As we never send MAX_PUSH_ID, a client that receives
+            // one treats it as a connection error of type H3_FRAME_UNEXPECTED.
+            // See <https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.7>.
+            Err(Error::HttpFrameUnexpected)
```
martinthomson: Is this error code correct for our server implementation?

It's OK if we abort in that case, but the correct behavior for a server is to ignore this.

neqo-http3/src/frames/hframe.rs
```diff
@@ -178,30 +141,39 @@ impl FrameDecoder<Self> for HFrame {
...
+            frame_type,
+            HFrameType::CANCEL_PUSH | HFrameType::PUSH_PROMISE | HFrameType::PRIORITY_UPDATE_PUSH
+        ) {
+            // Server push is not supported. These frames carry a push ID, but we never raise the
+            // maximum push ID above zero (we do not send MAX_PUSH_ID), so any push ID exceeds it
+            // and this is a connection error of type H3_ID_ERROR. See PUSH_PROMISE
+            // <https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.5>, CANCEL_PUSH
+            // <https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.3>, and PRIORITY_UPDATE
+            // <https://www.rfc-editor.org/rfc/rfc9218.html#section-7.2>.
+            Err(Error::HttpId)
```
martinthomson: As above, a server that receives CANCEL_PUSH without enabling server push would not treat that as an ID error, it would regard that as an unexpected frame.  Right?

martinthomson (commented):
This thread just keeps on giving.  These are low priority correctness things, with little impact on the final decision, so treat accordingly.

neqo-http3/src/connection.rs
mxinden: Yes. A push stream is rejected in `NewStreamType::final_stream_type` (`stream_type_reader.rs`) as soon as the stream type is read, so it never reaches `handle_new_stream` here. A client rejects it with `H3_ID_ERROR` (it never sent `MAX_PUSH_ID`) and a server with `H3_STREAM_CREATION_ERROR` (RFC 9114, Section 6.2.2).

larseggert:
@mxinden this needs a rebase for the CI/NSS fixes, but somehow I don't have that button in GitHub.

PR #3819, opened 2026-07-17
fix: check push id before decoding a blocked push promise

uwezkhan:
handle_push_promise checks the Push ID only through new_push_promise, which runs after the QPACK header block decodes. Before this change, a header block that refers to a dynamic table entry the server has not delivered yet blocks the decoder, so the frame lands on blocked_push_promise with no check, and a server that keeps the decoder blocked can push any Push ID unvalidated. RFC 9114 Section 7.2.5 wants a Push ID above the maximum treated as H3_ID_ERROR.

After, check_push_id runs at the top of handle_push_promise, before any decode or buffering, so the error is raised as soon as the frame arrives whatever the decoder state. The check stays in the push controller since that is where the maximum lives; the tradeoff is that a valid Push ID is now checked twice, once here and again in new_push_promise after unblocking. In-range Push IDs behave as before.

### DIFF
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -5596,6 +5596,27 @@ mod tests {
         assert_closed(&client, &Error::HttpId);
     }
 
+    // A PUSH_PROMISE whose Push ID exceeds the maximum is an H3_ID_ERROR even when
+    // its header block blocks the QPACK decoder. max_push_id is 4, so push_id 5 is
+    // out of range; the header block refers to the dynamic table, so the client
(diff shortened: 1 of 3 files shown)


larseggert (approved):

mxinden:
Note that server push will be removed in https://github.com/mozilla/neqo/pull/3816. Let's hold here for now.

uwezkhan:
Makes sense, happy to hold. If #3816 lands and the push logic goes away, this fix is moot and I'll close it then.

larseggert:
#3816 has merged, I assume this is now moot?

martinthomson:
Indeed.

PR #3837, opened 2026-08-01
Support handling queuing and highwatermarks/max-age for outgoing datagrams per the WebTransport spec

jesup:
Support handling queuing and highwatermarks/max-age datagrams per the WebTransport spec
Was https://github.com/mozilla/neqo/pull/3469

### DIFF
--- a/neqo-common/src/bytes.rs
+++ b/neqo-common/src/bytes.rs
@@ -39,6 +39,15 @@ impl Bytes {
         self.len() == 0
     }
 
+    #[must_use]
+    pub fn into_vec(self) -> Vec<u8> {
+        if self.offset == 0 {
+            self.data
+        } else {
+            self.data[self.offset..].to_vec()
+        }
+    }
+
     /// Skips the first `n` bytes, consuming and returning `self`.
     ///
     /// # Panics
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1657,10 +1657,82 @@ impl Http3Connection {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<(
+        bool,
+        Option<extended_connect::datagram_queue::DatagramOutcome>,
+    )> {
         self.validate_extended_connect_session(session_id)?
             .borrow_mut()
             .send_datagram(conn, buf, id, now)
+    }
+
+    /// # Errors
+    /// Returns `InvalidStreamId` if the session does not exist or is not a `WebTransport`
+    /// session.
+    pub fn webtransport_set_datagram_high_water_mark(
+        &self,
+        session_id: StreamId,
+        mark: f64,
+    ) -> Res<()> {
+        self.webtransport_session(session_id)?
+            .borrow_mut()
+            .set_datagram_high_water_mark(mark);
+        Ok(())
+    }
+
+    /// # Errors
+    /// Returns `InvalidStreamId` if the session does not exist or is not a `WebTransport`
+    /// session.
+    pub fn webtransport_set_datagram_max_age(
+        &self,
+        session_id: StreamId,
+        age_ms: f64,
+        now: Instant,
+    ) -> Res<Vec<extended_connect::datagram_queue::DatagramOutcome>> {
+        Ok(self
+            .webtransport_session(session_id)?
+            .borrow_mut()
+            .set_datagram_max_age(age_ms, now))
+    }
+
+    /// Drain one session's datagram queue into the QUIC layer.
+    ///
+    /// Applies to every extended-CONNECT protocol, not just `WebTransport`: connect-udp
+    /// sessions queue their datagrams the same way.
+    ///
+    /// # Errors
+    /// Returns `InvalidStreamId` if the session does not exist.
+    fn process_datagram_queue(
+        &self,
+        session_id: StreamId,
+        conn: &mut Connection,
+        now: Instant,
+    ) -> Res<Vec<extended_connect::datagram_queue::DatagramOutcome>> {
+        Ok(self
+            .get_extended_connect_session(session_id)?
+            .borrow_mut()
+            .process_datagram_queue(conn, now))
+    }
+
+    pub(crate) fn process_all_datagram_queues(
+        &self,
+        conn: &mut Connection,
+        now: Instant,
+        mut on_outcome: impl FnMut(StreamId, extended_connect::datagram_queue::DatagramOutcome),
+    ) {
+        let session_ids: Vec<StreamId> = self
+            .recv_streams
+            .iter()
+            .filter_map(|(id, stream)| stream.extended_connect_session().is_some().then_some(*id))
+            .collect();
+
+        for session_id in session_ids {
+            if let Ok(outcomes) = self.process_datagram_queue(session_id, conn, now) {
+                for outcome in outcomes {
+                    on_outcome(session_id, outcome);
+                }
+            }
+        }
     }
 
     /// If the control stream has received frames `MaxPushId`, `Goaway`, `PriorityUpdateRequest` or
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1670,11 +1670,11 @@ impl Http3Connection {
         session_id: StreamId,
         max_age: Duration,
         now: Instant,
-    ) -> Res<Vec<extended_connect::DatagramOutcome>> {
-        Ok(self
-            .webtransport_session(session_id)?
+    ) -> Res<()> {
+        self.webtransport_session(session_id)?
             .borrow_mut()
-            .set_datagram_max_age(max_age, now))
+            .set_datagram_max_age(max_age, now);
+        Ok(())
     }
 
     /// Drain every session's datagram queue into the QUIC layer.
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1687,6 +1687,11 @@ impl Http3Connection {
     /// Also refreshes the cache [`Self::next_datagram_expiry`] reads, using
     /// the same `recv_streams` walk this already has to do rather than
     /// paying for a second one.
+    ///
+    /// Invariant callers must preserve: the cache is only as fresh as the
+    /// last call to this function, so a caller that clamps a callback delay
+    /// against [`Self::next_datagram_expiry`] must call this *first*, in the
+    /// same processing round, or it clamps against last round's deadline.
     pub(crate) fn process_all_datagram_queues(&mut self, conn: &mut Connection, now: Instant) {
         self.datagram_next_expiry = self
             .recv_streams
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1693,6 +1693,10 @@ impl Http3Connection {
     /// against [`Self::next_datagram_expiry`] must call this *first*, in the
     /// same processing round, or it clamps against last round's deadline.
     pub(crate) fn process_all_datagram_queues(&mut self, conn: &mut Connection, now: Instant) {
+        if !self.webtransport.enabled() && !self.connect_udp.enabled() {
+            self.datagram_next_expiry = None;
+            return;
+        }
         self.datagram_next_expiry = self
             .recv_streams
             .values()
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -17,7 +17,8 @@ use neqo_common::{
 };
 use neqo_qpack as qpack;
 use neqo_transport::{
-    AppError, CloseReason, Connection, DatagramTracking, State, StreamId, StreamType, ZeroRttState,
+    AppError, CloseReason, Connection, DatagramTracking, OutputBatch, State, StreamId, StreamType,
+    ZeroRttState,
     streams::{SendGroupId, SendOrder},
 };
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
@@ -2032,6 +2033,24 @@ impl Http3Connection {
     #[must_use]
     pub fn recv_streams_mut(&mut self) -> &mut HashMap<StreamId, Box<dyn RecvStream>> {
         &mut self.recv_streams
+    }
+}
+
+/// Clamp an [`OutputBatch::Callback`]'s delay to `expiry`, if that's sooner:
+/// the transport layer's callback delay has no notion of an http3-layer
+/// datagram queue's max-age deadline, so without this an otherwise-idle
+/// connection with a queued datagram would wait on whatever unrelated timer
+/// happens to fire next instead of waking up in time to expire it. A no-op
+/// on every other [`OutputBatch`] variant, and on an `expiry` that's already
+/// passed or absent.
+pub fn clamp_to_expiry(out: OutputBatch, expiry: Option<Instant>, now: Instant) -> OutputBatch {
+    match out {
+        OutputBatch::Callback(delay) => OutputBatch::Callback(
+            expiry
+                .filter(|expiry| *expiry > now)
+                .map_or(delay, |expiry| delay.min(expiry - now)),
+        ),
+        out => out,
     }
 }
 
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -304,6 +304,11 @@ pub struct Http3Connection {
     /// read by [`Self::next_datagram_expiry`] instead of re-walking
     /// `recv_streams`.
     datagram_next_expiry: Option<Instant>,
+    /// The stream id of every extended-CONNECT session's control stream,
+    /// kept in sync with `recv_streams` on session create/remove so
+    /// [`Self::process_all_datagram_queues`] doesn't need to walk every
+    /// stream on the connection to find the handful that are sessions.
+    extended_connect_session_ids: HashSet<StreamId>,
 }
 
 impl Display for Http3Connection {
@@ -343,6 +348,7 @@ impl Http3Connection {
             send_group_generator: SendGroupGenerator::default(),
             role,
             datagram_next_expiry: None,
+            extended_connect_session_ids: HashSet::default(),
         }
     }
 
@@ -662,6 +668,7 @@ impl Http3Connection {
             // TODO: investigate whether this code can automatically retry failed transactions.
             self.send_streams.clear();
             self.recv_streams.clear();
+            self.extended_connect_session_ids.clear();
             Ok(())
         } else {
             debug_assert!(false, "Zero rtt rejected in the wrong state");
@@ -787,6 +794,7 @@ impl Http3Connection {
         }
         self.send_streams.clear();
         self.recv_streams.clear();
+        self.extended_connect_session_ids.clear();
     }
 
     /// This function will not handle the output of the function completely, but only
@@ -1225,6 +1233,7 @@ impl Http3Connection {
             Box::new(Rc::clone(&extended_conn)),
             Box::new(Rc::clone(&extended_conn)),
         );
+        self.extended_connect_session_ids.insert(id);
 
         let final_headers = Self::create_request_headers(&RequestDescription {
             method: "CONNECT",
@@ -1335,6 +1344,7 @@ impl Http3Connection {
                 Box::new(Rc::clone(&extended_conn)),
                 Box::new(extended_conn),
             );
+            self.extended_connect_session_ids.insert(stream_id);
             self.streams_with_pending_data.insert(stream_id);
         } else {
             self.cancel_fetch(stream_id, Error::HttpRequestRejected.code(), conn)?;
@@ -1685,8 +1695,10 @@ impl Http3Connection {
     /// callback here for a caller to mis-tag by protocol.
     ///
     /// Also refreshes the cache [`Self::next_datagram_expiry`] reads, using
-    /// the same `recv_streams` walk this already has to do rather than
-    /// paying for a second one.
+    /// [`Self::extended_connect_session_ids`] instead of walking all of
+    /// `recv_streams`: most streams on a connection aren't extended-CONNECT
+    /// sessions at all, and the per-feature check below only helps when the
+    /// feature is off entirely.
     ///
     /// Invariant callers must preserve: the cache is only as fresh as the
     /// last call to this function, so a caller that clamps a callback delay
@@ -1698,8 +1710,9 @@ impl Http3Connection {
             return;
         }
         self.datagram_next_expiry = self
-            .recv_streams
-            .values()
+            .extended_connect_session_ids
+            .iter()
+            .filter_map(|id| self.recv_streams.get(id))
             .filter_map(|s| s.extended_connect_session())
             .filter_map(|session| {
                 session.borrow_mut().process_datagram_queue(conn, now);
@@ -1939,6 +1952,7 @@ impl Http3Connection {
         if let Some(s) = &stream
             && s.stream_type() == Http3StreamType::ExtendedConnect
         {
+            self.extended_connect_session_ids.remove(&stream_id);
             self.send_streams.remove(&stream_id)?;
             if let Some(wt) = s.extended_connect_session() {
                 self.remove_extended_connect(&wt, conn);
@@ -1955,12 +1969,15 @@ impl Http3Connection {
         let stream = self.send_streams.remove(&stream_id);
         if let Some(s) = &stream
             && s.stream_type() == Http3StreamType::ExtendedConnect
-            && let Some(wt) = self
+        {
+            self.extended_connect_session_ids.remove(&stream_id);
+            if let Some(wt) = self
                 .recv_streams
                 .remove(&stream_id)?
                 .extended_connect_session()
-        {
-            self.remove_extended_connect(&wt, conn);
+            {
+                self.remove_extended_connect(&wt, conn);
+            }
         }
         stream
     }
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -9,7 +9,7 @@ use std::{
     fmt::{self, Debug, Display, Formatter},
     mem,
     rc::Rc,
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use neqo_common::{
@@ -300,6 +300,10 @@ pub struct Http3Connection {
     webtransport: ExtendedConnectFeature,
     connect_udp: ExtendedConnectFeature,
     send_group_generator: SendGroupGenerator,
+    /// Cached result of the last [`Self::process_all_datagram_queues`] call,
+    /// read by [`Self::next_datagram_expiry`] instead of re-walking
+    /// `recv_streams`.
+    datagram_next_expiry: Option<Instant>,
 }
 
 impl Display for Http3Connection {
@@ -338,6 +342,7 @@ impl Http3Connection {
             recv_streams: HashMap::default(),
             send_group_generator: SendGroupGenerator::default(),
             role,
+            datagram_next_expiry: None,
         }
     }
 
@@ -1637,10 +1642,73 @@ impl Http3Connection {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<extended_connect::DatagramQueueOutcome> {
         self.validate_extended_connect_session(session_id)?
             .borrow_mut()
             .send_datagram(conn, buf, id, now)
+    }
+
+    /// # Errors
+    /// Returns `InvalidStreamId` if the session does not exist or is not a `WebTransport`
+    /// session.
+    pub fn webtransport_set_datagram_high_water_mark(
+        &self,
+        session_id: StreamId,
+        mark: Option<usize>,
+    ) -> Res<()> {
+        self.webtransport_session(session_id)?
+            .borrow_mut()
+            .set_datagram_high_water_mark(mark);
+        Ok(())
+    }
+
+    /// # Errors
+    /// Returns `InvalidStreamId` if the session does not exist or is not a `WebTransport`
+    /// session.
+    pub fn webtransport_set_datagram_max_age(
+        &self,
+        session_id: StreamId,
+        max_age: Duration,
+        now: Instant,
+    ) -> Res<Vec<extended_connect::DatagramOutcome>> {
+        Ok(self
+            .webtransport_session(session_id)?
+            .borrow_mut()
+            .set_datagram_max_age(max_age, now))
+    }
+
+    /// Drain every session's datagram queue into the QUIC layer.
+    ///
+    /// Applies to every extended-CONNECT protocol, not just `WebTransport`: connect-udp
+    /// sessions queue their datagrams the same way. Each session reports its own
+    /// outcomes through its own `events`/`connect_type`, so there is no per-outcome
+    /// callback here for a caller to mis-tag by protocol.
+    ///
+    /// Also refreshes the cache [`Self::next_datagram_expiry`] reads, using
+    /// the same `recv_streams` walk this already has to do rather than
+    /// paying for a second one.
+    pub(crate) fn process_all_datagram_queues(&mut self, conn: &mut Connection, now: Instant) {
+        self.datagram_next_expiry = self
+            .recv_streams
+            .values()
+            .filter_map(|s| s.extended_connect_session())
+            .filter_map(|session| {
+                session.borrow_mut().process_datagram_queue(conn, now);
+                session.borrow().next_datagram_expiry()
+            })
+            .min();
+    }
+
+    /// The earliest instant at which any session's outgoing datagram queue
+    /// needs [`Self::process_all_datagram_queues`] called again to expire a
+    /// stale datagram, even if nothing else - no packets to send or receive,
+    /// no other timer - would otherwise trigger it. `None` if no session has
+    /// a datagram queued with an expiry pending.
+    ///
+    /// Reads the cache [`Self::process_all_datagram_queues`] refreshes,
+    /// instead of walking `recv_streams` a second time.
+    pub(crate) const fn next_datagram_expiry(&self) -> Option<Instant> {
+        self.datagram_next_expiry
     }
 
     /// Frames whose handling is specific to the client and server (`Goaway`,
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -32,7 +32,7 @@ use nss::{AuthenticationStatus, ResumptionToken, SecretAgentInfo, agent::Certifi
 use crate::{
     Error, Http3Parameters, NewStreamType, Priority, ReceiveOutput, Res, SendGroupId,
     client_events::{Http3ClientEvent, Http3ClientEvents, WebTransportEvent},
-    connection::{Http3Connection, Http3State, RequestDescription},
+    connection::{Http3Connection, Http3State, RequestDescription, clamp_to_expiry},
     features::ConnectType,
     frames::HFrame,
     request_target::RequestTarget,
@@ -826,20 +826,7 @@ impl Http3Client {
         // Update H3 for any transport state changes and events
         self.process_http3(now);
 
-        // `self.conn`'s callback delay only accounts for transport-layer
-        // timers; it has no notion of an http3-layer datagram queue's
-        // max-age deadline. Clamp it so an otherwise-idle connection still
-        // wakes up in time to expire a stale queued datagram, instead of
-        // waiting on whatever unrelated timer happens to fire next.
-        match out {
-            OutputBatch::Callback(delay) => OutputBatch::Callback(
-                self.base_handler
-                    .next_datagram_expiry()
-                    .filter(|expiry| *expiry > now)
-                    .map_or(delay, |expiry| delay.min(expiry - now)),
-            ),
-            out => out,
-        }
+        clamp_to_expiry(out, self.base_handler.next_datagram_expiry(), now)
     }
 
     /// This function takes the provided result and check for an error.
--- a/neqo-http3/src/connection_client.rs
+++ b/neqo-http3/src/connection_client.rs
@@ -752,6 +752,14 @@ impl Http3Client {
                 }
                 let res = self.base_handler.process_sending(&mut self.conn, now);
                 self.check_result(now, &res);
+
+                // Drain per-session datagram queues into the QUIC layer.
+                // Datagrams are enqueued by webtransport_send_datagram() and only
+                // moved to the QUIC send queue here, during process_http3(). This
+                // is called from process_output()/process(), so datagrams are sent
+                // on the next outgoing packet after the caller triggers processing.
+                self.base_handler
+                    .process_all_datagram_queues(&mut self.conn, now);
             }
             Http3State::Closed { .. } => {}
             _ => {
@@ -794,6 +802,12 @@ impl Http3Client {
     ///    state-change event.
     ///  - Therefore the HTTP/3 layer processing (`process_http3`) is called again.
     ///
+    /// The returned [`OutputBatch::Callback`] duration is also clamped to any
+    /// pending WebTransport datagram-queue max-age deadline
+    /// (`Http3Connection::next_datagram_expiry`), so a session with a queued
+    /// datagram and no other timer pending still gets `process_output`
+    /// called again in time to expire it.
+    ///
     /// [1]: ../neqo_transport/enum.Output.html
     /// [2]: ../neqo_transport/struct.ConnectionEvents.html
     /// [3]: ../neqo_transport/struct.Connection.html#method.process_output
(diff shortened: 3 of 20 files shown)


jesup:
This PR is part of a stack of 2 bookmarks:

1. `trunk()`
1. [users/jesup/stats](https://github.com/mozilla/neqo/pull/3835)
1. **users/jesup/highwatermarks ← this PR**

---

neqo-http3/src/features/extended_connect/webtransport_session.rs
```diff
@@ -302,4 +338,32 @@ impl Protocol for Session {
...
         Ok(())
     }
+
+    fn set_datagram_high_water_mark(&mut self, mark: f64) {
+        self.set_datagram_high_water_mark(mark);
+    }
+
+    fn set_datagram_max_age(&mut self, age_ms: f64, now: Instant) -> Vec<DatagramOutcome> {
+        self.set_datagram_max_age(age_ms, now)
+    }
```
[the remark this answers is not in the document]
jesup: Verified this isn't actually recursive: `self.set_datagram_high_water_mark(mark)` inside the `impl Protocol for Session` block resolves to the inherent method on `Session`, not back to `Protocol::set_datagram_high_water_mark` — Rust always prefers an inherent method over a trait method for a concrete receiver type. Confirmed with a standalone repro calling through `&mut dyn Protocol`; no stack overflow.

Switched all four methods here (`set_datagram_high_water_mark`, `set_datagram_max_age`, `enqueue_datagram`, `drain_datagram_queue`) to the explicit `Self::method(self, ...)` form already used by `register_send_group`/`validate_send_group` in this same file, since the implicit `self.` form is genuinely ambiguous-looking even though it compiles correctly.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    pub fn enqueue(
+        &mut self,
+        data: Bytes,
+        id: Option<u64>,
+        payload_len: usize,
+        now: Instant,
+    ) -> (bool, Option<DatagramOutcome>) {
+        let dropped = if self.queue.len() >= self.hard_limit {
```
mxinden: I believe within Firefox itself, we should never drop a datagram. Dropping datagrams within Firefox is avoidable. Previously dropping datagrams had significant performance impact. See https://phabricator.services.mozilla.com/D308822 and https://phabricator.services.mozilla.com/D239162.

Do I understand correctly, that on the JS side we do have a way to enforce backpressure, i.e. that the datagram writable is able to slow down the producer? If  this is true, I believe we should propagate backpressure all the way from Neqo to the JS environment, thus not having to drop here.

I have to give this more thought, but a rough sketch:

1. When filling the last queue slot, or slightly going above the queue limit, Neqo returns a signal that the queue is full.
2. With the event, the writable in JS would block.
3. When Neqo sends a datagram on the wire, bringing the queue below its limit, emit an event, unblocking the writable.

I believe going slightly over the limit is fine, i.e. it is arbitrary anyways. This would also simplify the multi-session per connection case, where we have a race condition of two sessions trying to fill the last remaining slot in the datagram queue.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,345 @@
...
+// except according to those terms.
+
+use std::{
+    collections::VecDeque,
+    time::{Duration, Instant},
+};
+
+use neqo_common::{Bytes, qdebug, qtrace};
+
+pub const DEFAULT_HARD_LIMIT: usize = 1000;
```
mxinden, line 14: 1000 datagrams is a large standing buffer, and the cost is latency (bufferbloat). At ~1200 B that's ~1.2 MB of pending, potentially outdated data. With such a buffer we could saturate a BDP of a 100 Mbps / ~96 ms path, assuming the JS side **only** enqueues once per RTT. I assume the JS side would want to / be able to enqueue multiple times per RTT, thus a much lower buffer should suffice.

Per the spec the outgoing side should backpressure at `outgoingMaxBufferedDatagrams` rather than grow deep and head-drop, so this should really just be a memory backstop. Suggest 64: still arbitrary, but rooted in the GSO segment cap (one `sendmmsg` batch), i.e. the amount we actually need buffered to amortize syscall overhead.

mxinden (commented):
Two early comments. I will need to give this more thought. I am currently working on https://github.com/mozilla/neqo/issues/2852 which directly overlaps with this work.

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: The concern would be if a video codec dumps an iframe on us.  An H.264 (or really any codec) 4k iframe at 40Mbps average (which one can see; I see that today in GeForceNow.com streaming games in WebRTC even at 1440p) could be 5x large than the average (or even larger, up to 8x).  That could hit 400-1100 datagrams just for the iframe (and other packets may be queued at the same time - audio, data, etc).  And that assumes a single video stream.

I'm going to deep-drive on how we'd handle streams like this and come back with some proposed changes; there are disconnects between the internal 10-datagram buffer and the WebTransport buffers/backpressure

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: Discussion on slack... Senders may be fire-and-forget senders, doing adaptation via remote loss results (similar to RTCP in webrtc), so backpressure's impact may be limited.

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: See slack conversation; updates to this are in process in gecko/neqo

neqo-common/src/bytes.rs
```diff
@@ -39,6 +39,15 @@ impl Bytes {
         self.len() == 0
     }
 
+    #[must_use]
+    pub fn into_vec(self) -> Vec<u8> {
+        if self.offset == 0 {
+            self.data
+        } else {
+            self.data[self.offset..].to_vec()
```
martinthomson: I *think* that this is better:

```suggestion
            drop(self.data.drain(..self.offset));
            self.data
```

because it doesn't necessarily perform a fresh allocation.

neqo-transport/src/stats.rs
martinthomson, line 193: These changes don't relate at all to the subject of the PR.

neqo-http3/src/webtransport.rs
```diff
@@ -215,6 +244,37 @@ impl ClientSession for Http3Client {
...
 
+    fn webtransport_set_datagram_high_water_mark(
+        &mut self,
+        session_id: StreamId,
+        high_water_mark: f64,
+    ) -> Res<()> {
+        self.connection_and_handler()
+            .1
+            .webtransport_set_datagram_high_water_mark(session_id, high_water_mark)
+    }
```
martinthomson:
```suggestion
        let (_conn, handler) = self.connection_and_handler();
        handler.webtransport_set_datagram_high_water_mark(session_id, high_water_mark)
    }
```

Not only is this clearer, it's shorter.

neqo-http3/src/webtransport.rs
```diff
@@ -215,6 +244,37 @@ impl ClientSession for Http3Client {
...
+
+    fn webtransport_set_datagram_max_age(
+        &mut self,
+        session_id: StreamId,
+        max_age: f64,
+        now: Instant,
+    ) -> Res<()> {
+        let expired = self
+            .connection_and_handler()
+            .1
```
martinthomson: See above.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,345 @@
...
+        old.into_iter().map(|d| d.id).collect()
+    }
+
+    pub fn enqueue(
+        &mut self,
+        data: Bytes,
+        id: Option<u64>,
+        payload_len: usize,
+        now: Instant,
+    ) -> (bool, Option<DatagramOutcome>) {
```
martinthomson: This type could be simplified greatly.

At this stage, there are three states that we're interested in:

1. The queue has space below the watermark.
2. The queue has space, but it is above the watermark.
3. The queue is full, so something needs to be dropped.

That suggests that `DatagramOutcome` is not at all the right type to use for this function.  A new type:

```rust
enum DatagramQueueOutcome {
    /// The queue was below the watermark.
    Ok,
    /// The queue had space, but it was above the watermark.
    AboveWatermark,
    /// The queue was full and the indicated ID was dropped.
    Overflowed {
        dropped: DatagramId,
    }
}
/// Because named types help understand what something is.
type DatagramId = u64;
```

martinthomson (changes requested):
I can't review this.  It's at least two changes.  The high water mark stuff would be big enough, but this includes stats changes.

neqo-http3/src/connection.rs
```diff
@@ -1612,12 +1657,84 @@ impl Http3Connection {
...
+            .collect();
+
+        for session_id in session_ids {
+            if let Ok(outcomes) = self.process_datagram_queue(session_id, conn, now) {
+                for outcome in outcomes {
+                    on_outcome(session_id, outcome);
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jesup: Applied, with one change: a plain for (id, stream) in &self.recv_streams trips clippy::iter_over_hash_type, so it uses the iter().filter_map(...).for_each(...) shape already used by drain_webtransport_sessions.  Same effect: no Vec, no second hash lookup.  The Http3Connection::process_datagram_queue is gone.

neqo-transport/src/stats.rs
```diff
@@ -187,6 +187,13 @@ pub struct SearchResetStats {
 /// Congestion Control stats
 #[derive(Default, Clone, PartialEq)]
 pub struct CongestionControlStats {
+    /// Congestion window size, in bytes, on the primary path. Not updated continuously: this
+    /// is a snapshot taken only when stats are pulled from the connection (see
+    /// [`crate::Connection::stats`]).
+    pub cwnd: usize,
+    /// Bytes in flight (sent but not yet acked or declared lost) on the primary path. Like
+    /// `cwnd`, this is a snapshot taken only when stats are pulled from the connection.
+    pub bytes_in_flight: usize,
```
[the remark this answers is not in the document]
jesup: This was a rebase artifact

neqo-transport/src/stats.rs
jesup: This was a rebase artifact, resolved

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -439,6 +565,10 @@ impl Session {
         // dgram_context_id returns the payload after stripping any context ID
         match self.protocol.dgram_context_id(datagram) {
             Ok(slice) => {
+                if let Some(stats) = self.protocol.stats_mut() {
+                    stats.datagrams_received += 1;
+                    stats.datagram_bytes_received += slice.len() as u64;
```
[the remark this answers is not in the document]
jesup: No longer part of this patch (rebase issue)

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: Done, with two adjustments:
  - dropped is Option<DatagramId>: datagrams can be sent untracked (DatagramTracking::None), and evicting one of those must still report overflow so the app applies backpressure, just without an ID.
  - Overflowed now takes priority over the watermark comparison. A full queue always means backpressure, so the outcome no longer depends on whether the app happened to set outgoingHighWaterMark above the hard limit.

DatagramOutcome (the per-datagram fate reported via events) is unchanged and still uses Sent/Expired/Overflowed/Dropped; DatagramId is used in both.

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: Fixed with a combination of neqo and gecko changes; see conversation over remaining details in quicdev

jesup:
> I can't review this. It's at least two changes. The high water mark stuff would be big enough, but this includes stats changes.

Broken up

neqo-http3/src/connection_client.rs
```diff
@@ -812,7 +836,21 @@ impl Http3Client {
...
+        // wakes up in time to expire a stale queued datagram, instead of
+        // waiting on whatever unrelated timer happens to fire next.
+        match out {
+            OutputBatch::Callback(delay) => OutputBatch::Callback(
+                self.base_handler
+                    .next_datagram_expiry()
+                    .map_or(delay, |expiry| {
+                        delay.min(expiry.saturating_duration_since(now))
+                    }),
+            ),
```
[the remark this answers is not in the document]
jesup: Fixed as suggested: filter out expiries that aren't in the future before clamping. Also covers the happy-path case: expire_old_datagrams's strict age > max_age leaves a datagram queued at exactly its expiry instant, so next_expiry() == now hit the same Callback(0) loop even without a stuck teardown.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -390,43 +399,150 @@ impl Session {
...
+            .collect()
+    }
+
+    /// Drain the per-session datagram queue and hand datagrams to the QUIC layer.
+    ///
+    /// Only as many datagrams as the QUIC layer can currently accept are handed
+    /// over; the rest stay queued for a later call, so that a burst is paced
+    /// against what the connection can send instead of overflowing the QUIC
+    /// layer's fixed-size queue. See [`WebTransportDatagramQueue::drain`].
+    pub(crate) fn process_datagram_queue(
```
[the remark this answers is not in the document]
jesup: Fixed both directions:
- process_datagram_queue now no-ops once the session leaves State::Active.
- close()/close_session() drain the queue and report Dropped for every tracked datagram still queued, via a new datagram_outcome() event on ExtendedConnectEvents and take_all() on the queue.
Added tests for both.

neqo-http3/src/connection.rs
```diff
@@ -1637,12 +1637,77 @@ impl Http3Connection {
...
+    ///
+    /// Applies to every extended-CONNECT protocol, not just `WebTransport`,
+    /// same as [`Self::process_all_datagram_queues`].
+    pub(crate) fn next_datagram_expiry(&self) -> Option<Instant> {
+        self.recv_streams
+            .values()
+            .filter_map(|s| s.extended_connect_session())
+            .filter_map(|session| session.borrow().next_datagram_expiry())
+            .min()
+    }
```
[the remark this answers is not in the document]
jesup: Cached it rather than adding the separate id-set: process_all_datagram_queues already walks every session, so it now writes the min into a new datagram_next_expiry: Option<Instant> field (made the method &mut self rather than using a Cell), and next_datagram_expiry() just reads that field.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,536 @@
...
+    /// The queue had space, but it was at or above the high water mark.
+    AboveWatermark,
+    /// The queue was full, so the oldest datagram was dropped to make room.
+    /// `dropped` is its tracking ID, or `None` if it was sent untracked.
+    Overflowed { dropped: Option<DatagramId> },
+}
+
+#[derive(Debug)]
+pub struct QueuedDatagram {
+    pub data: Bytes,
```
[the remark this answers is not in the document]
jesup: Fixed: QueuedDatagram::data is now Vec<u8>, session.rs enqueues the Vec directly and connection.rs's process_datagram_queue passes it straight to conn.send_datagram() without into_vec(). Dropped Bytes::into_vec and its test from neqo-common.

neqo-http3/src/connection_server.rs
```diff
@@ -199,6 +199,10 @@ impl Http3ServerHandler {
         if !self.check_result(conn, now, &res) && self.base_handler.state().active() {
             let res = self.base_handler.process_sending(conn, now);
             self.check_result(conn, now, &res);
+
+            // Process datagram queues to send any queued datagrams
+            self.base_handler
```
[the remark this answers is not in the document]
jesup: Added webtransport_set_datagram_high_water_mark/webtransport_set_datagram_max_age to ServerHandler and ServerSession (mirroring the client side, using the same implementation-defined default max-age), changed webtransport_send_datagram to return DatagramQueueOutcome instead of discarding it, and plumbed outcomes out through Http3ServerConnEvents/Http3ServerEvent (and the equivalent connect-udp path).

neqo-http3/src/connection.rs
```diff
@@ -1637,12 +1637,78 @@ impl Http3Connection {
...
```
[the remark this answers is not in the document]
jesup: Sorry, forgot to move the bookmark to include the patch

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: Still ongoing separately: buffer size (`DEFAULT_HARD_LIMIT`) and the default outgoing max-age are both still under discussion (see the quicdev thread referenced above); will follow up here once that settles.

jesup:
Still ongoing separately: buffer size (`DEFAULT_HARD_LIMIT`) and the default outgoing max-age are both still under discussion (see the quicdev thread  referenced above); will follow up here once that settles.

neqo-http3/src/connection_client.rs
```diff
@@ -752,6 +752,24 @@ impl Http3Client {
...
+                    now,
+                    |session_id, outcome| {
+                        self.events.insert(Http3ClientEvent::WebTransport(
+                            WebTransportEvent::DatagramOutcome {
+                                session_id,
+                                outcome,
+                            },
+                        ));
+                    },
+                );
```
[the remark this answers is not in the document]
jesup: Fixed the connect-udp-reported-as-WebTransport bug on both client and server by having process_datagram_queue report through its own events/connect_type instead of a callback that hardcoded WebTransport

neqo-http3/src/connection_client.rs
```diff
@@ -812,7 +836,21 @@ impl Http3Client {
...
+        match out {
+            OutputBatch::Callback(delay) => OutputBatch::Callback(
+                self.base_handler
+                    .next_datagram_expiry()
+                    .map_or(delay, |expiry| {
+                        delay.min(expiry.saturating_duration_since(now))
+                    }),
+            ),
+            out => out,
+        }
```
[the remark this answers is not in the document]
jesup: Already had a fix for this; wasn't in the bookmark

neqo-http3/src/connection_server.rs
```diff
@@ -199,6 +199,13 @@ impl Http3ServerHandler {
         if !self.check_result(conn, now, &res) && self.base_handler.state().active() {
             let res = self.base_handler.process_sending(conn, now);
             self.check_result(conn, now, &res);
+
+            // Process datagram queues to send any queued datagrams
+            let events = self.events.clone();
+            self.base_handler
+                .process_all_datagram_queues(conn, now, move |session_id, outcome| {
+                    events.webtransport_datagram_outcome(session_id, outcome);
+                });
```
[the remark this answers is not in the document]
jesup: Gave the server the same max-age wakeup guarantee the client already has

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    fn expire_old_datagrams(&mut self, now: Instant) -> Vec<Option<DatagramId>> {
+        let split = self.queue.partition_point(|d| d.age(now) > self.max_age);
+        if split == 0 {
+            return Vec::new();
+        }
+        let kept = self.queue.split_off(split);
+        let old = std::mem::replace(&mut self.queue, kept);
+        old.into_iter().map(|d| d.id).collect()
+    }
```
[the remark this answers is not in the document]
jesup: expire_old_datagrams now pops from the front instead of assuming partition_point's sortedness invariant

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
```diff
@@ -75,3 +83,103 @@ fn max_datagram_size_smaller_than_session_prefix() {
...
+        matches!(
+            e,
+            Http3ClientEvent::WebTransport(WebTransportEvent::DatagramOutcome {
+                session_id: sid,
+                outcome: DatagramOutcome::Expired(7),
+            }) if sid == session_id
+        )
+    };
+    assert!(wt.client.events().any(wt_expired_event));
+}
```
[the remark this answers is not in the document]
jesup: Tests for most of these written.  Two items remain, and  both are much harder than the rest: 

client-side callback-clamp - verified impossible via the public API on a healthy connection, since the client drains its datagram queue eagerly on every single process_output call, so by the time you'd observe an idle Callback, the queue's already empty (I tried it, disabled the clamp, and it made no observable difference - there's nothing to falsify).

QUIC-refusal Dropped - I found the exact mechanism (a 0-RTT connection whose remembered max_datagram_frame_size is larger than the real one confirmed at 1-RTT), but reproducing it requires either ~100+ lines of raw frame-level Connection plumbing bypassing Http3Server/WtTest entirely (the existing 0-RTT test infra doesn't support WebTransport sessions), or an uncertain attempt at resuming a WT session over 0-RTT that no existing test does.  So I plan to leave these.

larseggert:
@jesup I don't think these performance regressions are just due to runner variance. Suggest to point Claude at the main and PR profiles to diagnose them.

Also, could you mark those review comments in the history that you think are addressed as "resolved"? Timeline is very noisy otherwise.

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
```diff
@@ -75,3 +83,243 @@ fn max_datagram_size_smaller_than_session_prefix() {
     assert_eq!(wt_session.max_datagram_size(), Ok(0));
     assert_eq!(wt.max_datagram_size(wt_session.stream_id()), Ok(0));
 }
+
+#[test]
+fn datagram_high_water_mark_reported_via_send_datagram() {
```
larseggert: There is a **lot** of repeated code in these tests. Factor into some helpers to reduce LoC?

neqo-http3/src/webtransport.rs
```diff
@@ -62,6 +62,31 @@ pub trait ClientSession {
...
+    /// # Errors
+    ///
+    /// Returns error if the session ID is invalid or is not a WebTransport session.
+    fn webtransport_set_datagram_high_water_mark(
+        &mut self,
+        session_id: StreamId,
+        high_water_mark: Option<usize>,
+    ) -> Res<()>;
+
+    /// [`Duration::MAX`] means no limit, as does the default.
```
larseggert: `Option<Duration>` would avoid overloading `Duration::MAX`.

neqo-http3/src/webtransport.rs
```diff
@@ -62,6 +62,31 @@ pub trait ClientSession {
     /// This cannot panic. The max varint length is 8.
     fn webtransport_max_datagram_size(&self, session_id: StreamId) -> Res<u64>;
 
+    /// `None` means no limit. The caller is responsible for mapping the
+    /// `unrestricted double` `outgoingHighWaterMark` onto this, including NaN
+    /// and the infinities.
+    ///
```
larseggert: What `unrestricted double`?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -61,6 +65,9 @@ pub(crate) struct Session {
     /// CONNECT request.
     protocol: Box<dyn Protocol>,
     draining: bool,
+    /// Outgoing datagrams awaiting handover to the QUIC layer. Shared by every
+    /// extended-CONNECT protocol; the queue itself is protocol-agnostic.
+    datagram_queue: WebTransportDatagramQueue,
```
mxinden: Firefox only ever has one WebTransport session per QUIC connection, so this queue has no other session to be fair against.

neqo-http3/src/features/extended_connect/datagram_queue.rs
mxinden: I would start without this queue and add the queue and corresponding depth only once a benchmark shows we need it.

neqo-http3/src/features/extended_connect/datagram_queue.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// The instant at which the oldest queued datagram crosses `max_age`, if
+    /// any datagram is queued. The caller uses this to schedule a wakeup so
+    /// that expiry is not left to whichever unrelated timer happens to fire
+    /// next (see `Http3Connection::next_datagram_expiry`).
+    ///
+    /// Datagrams are served/expired oldest-first, so the front of the queue
+    /// is always the next one to expire. `None` if the queue is empty, or if
+    /// `max_age` is `DEFAULT_MAX_AGE` (no expiry has ever been requested).
+    #[must_use]
+    pub fn next_expiry(&self) -> Option<Instant> {
```
mxinden, line 254: Not a big problem, but this is a lot of machinery for throwing a datagram away. I would drop expired datagrams lazily, when `process_output` next looks at the queue, instead of waking up the socket thread for each drop.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -390,43 +401,187 @@ impl Session {
...
+        }
+
+        for dgram in to_send {
+            let tracking = dgram
+                .id
+                .map_or(DatagramTracking::None, DatagramTracking::Id);
+            // A datagram the QUIC layer refuses is gone: `drain` already removed it
+            // from the queue. Report it so the application isn't left waiting, and
+            // keep going, since a later datagram may still be small enough to fit.
+            let outcome = match conn.send_datagram(dgram.data, tracking) {
```
mxinden, line 569: Once a datagram reaches the transport queue there is no age policy on it any more, so the delay bound stops at the handover. With a single queue in `neqo-transport` that problem would be gone.

neqo-http3/src/features/extended_connect/datagram_queue.rs
mxinden: In addition, this queue does not only apply to WebTransport, but to MASQUE connect-udp as well. I don't think MASQUE connect-udp (nor WebTransport) should have an additional queue.

mxinden (commented):
I would like to argue against the 3rd queue introduced in this pull request.

**The content process has one send queue (the `WritableStream` behind `datagrams.writable`). The parent process has one, `QuicDatagrams`. This adds a third queue, in the parent process, directly on top of the first.** The spec describes `outgoingMaxBufferedDatagrams` as "the length of the underlying sink's outgoing queue in number of datagrams, beyond which backpressure is exerted", and `QuicDatagrams` is that sink.

I am wary of adding queues generally. A queue adds delay to everything behind it, it lets the application spend CPU producing data we might later throw away, and it postpones the backpressure signal that would have stopped that work in the first place. It also adds lifecycle state that has to be right on close, on expiry, and on reordering. For a best-effort, latency-first path those costs apply on every datagram, while the benefit only appears under bursts.

Martin raised the same concern on [#3469](https://github.com/mozilla/neqo/pull/3469) : *"In my mind, any queues for sending stuff in WebTransport would be outside of the stack. So I found this a little surprising to see."*

**A per-session queue is only needed when several sessions share one connection and something has to arbitrate between them.** We do not have that. Firefox uses one QUIC connection per WebTransport session: a unique id per channel ([`nsHttpChannel.cpp:8189`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpChannel.cpp#8189)) folded into the connection-info hash key ([`nsHttpConnectionInfo.cpp:285`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpConnectionInfo.cpp#285)), coalescing refused in `nsHttpConnectionMgr`, and `CanReuse()==false` for the session's lifetime ([`ConnectionEntry.cpp:511`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/ConnectionEntry.cpp#511)). We may want several sessions per connection one day. I would rather not support that today than pay for it now.

**The deadline is the one thing genuinely missing from `QuicDatagrams`, and it is cheap to add there.** It brings no WebTransport concepts with it.

Thus I suggest simply adding the max age mechanism to the `neqo-transport` `QuicDatagram` queue:

```diff
 pub struct QuicDatagram {
     data: Vec<u8>,
     tracking: DatagramTracking,
+    /// Drop instead of sending once this instant passes.
+    expires: Option<Instant>,
 }

     pub fn add_datagram(
         &mut self,
         data: Vec<u8>,
         tracking: DatagramTracking,
+        expires: Option<Instant>,
         stats: &mut Stats,
     ) -> Res<()> {
-        self.datagrams.push_back(QuicDatagram { data, tracking });
+        self.datagrams.push_back(QuicDatagram { data, tracking, expires });
```

plus, in `write_frames` where the queue is already walked:

```diff
+            if dgram.expires.is_some_and(|t| now >= t) {
+                self.conn_events
+                    .datagram_outcome(dgram.tracking(), OutgoingDatagramOutcome::Expired);
+                continue;
+            }
```

No timer required.

**Backpressure, not expiry, should be the mechanism.** #3859 makes the transport  signal when space frees. Expiry after acceptance means the application already encoded and IPC'd data we knew we could not send. Jan-Ivar's framing is that `outgoingMaxAge` "encourages push-based apps to create lots of work that will ultimately be tossed". I read it as a last resort for data that went stale after we accepted it, not as the main mechanism.

**On the 1000 slots:** [Gall's law](<https://en.wikipedia.org/wiki/John_Gall_(author)#Gall's_law>), "a complex system that works is invariably found to have evolved from a simple system that worked". I would rather ship the simple version, measure, and add a queue when a benchmark shows we need one. Adding a queue later is easy. The hard part is debugging it: a queue adds complexity, and performance problems in particular are hard to diagnose.

Concretely: land #3859, put the deadline in `QuicDatagrams`, and drop this queue entirely. I think this can end up much simpler than it is now.

jesup:
> @jesup I don't think these performance regressions are just due to runner variance. Suggest to point Claude at the main and PR profiles to diagnose them.

Sure.  Previous runs didn't show these, so I'll check.

> Also, could you mark those review comments in the history that you think are addressed as "resolved"? Timeline is very noisy otherwise.

I have been clearing all of them everytime I make code updates.  Every time I push, the automated AI review finds ~4-5 "new" issues which often were pre-existing (are we asking it to surface that many each pass?), though from previous experience (stats), eventually it stops finding more.  There were 2 sets of issues waiting currently because I pulled and rebased and re-pushed, so it re-reviewed.

neqo-http3/src/server.rs
```diff
@@ -141,14 +141,29 @@ impl Http3Server {
...
+                self.http3_handlers
+                    .values()
+                    .filter_map(|h| h.borrow().next_datagram_expiry())
+                    .filter(|expiry| *expiry > now)
+                    .map(|expiry| expiry - now)
+                    .min()
+                    .map_or(delay, |max_age_delay| delay.min(max_age_delay)),
+            ),
+            out => out,
         }
```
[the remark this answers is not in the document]
jesup: Was already done

neqo-http3/src/features/extended_connect/datagram_queue.rs
jesup: I updated it to be proactive (in response to AI review) because lazy expiry-on-next-process_output means an idle connection where the app hasn't sent anything since queuing never expires; max_age degrades from a deadline to "whenever something else happens to run."  The wakeup makes it an actual deadline to match the spec.
Your point is reasonable, however.  The question is mostly one of "how closely do we want to follow the spec?"  I would be ok with dropping this bit; your call.

mxinden: I am in favor of dropping it. You can document the shortcoming if you like.

neqo-http3/src/features/extended_connect/session.rs
jesup: If we teach transport about age, we can cover the age part of the spec better.  But moving the entire queue down into transport would mean teaching it about sessions (I have a later patch to round-robin between sessions for fairness), and also we'd need to move all the sendorder/sendgroup and max-age work all down into transport; it would have to gain a lot of details and machinery about how to handle these webtransport-specific things

neqo-http3/src/features/extended_connect/session.rs
jesup: shared connections is perennially on the list-of-things-to-do - but we can defer the fairness until we actually do it, for sure.

neqo-http3/src/features/extended_connect/session.rs
mxinden: See above. I don't think we will get to run multiple sessions on the same QUIC connection any time soon.

> **A per-session queue is only needed when several sessions share one connection and something has to arbitrate between them.** We do not have that. Firefox uses one QUIC connection per WebTransport session: a unique id per channel ([`nsHttpChannel.cpp:8189`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpChannel.cpp#8189)) folded into the connection-info hash key ([`nsHttpConnectionInfo.cpp:285`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpConnectionInfo.cpp#285)), coalescing refused in `nsHttpConnectionMgr`, and `CanReuse()==false` for the session's lifetime ([`ConnectionEntry.cpp:511`](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/ConnectionEntry.cpp#511)). We may want several sessions per connection one day. I would rather not support that today than pay for it now.

> we'd need to move all the sendorder/sendgroup

We already supports this in `neqo-transport` for streams, right? Thus doing the same for datagrams seems fine.

> and max-age work all down into transport;

I think that is minimal, see my diff in https://github.com/mozilla/neqo/pull/3837#pullrequestreview-5089187036.

neqo-http3/src/features/extended_connect/session.rs
mxinden: Yes, very much in favor of deferring. Let's work on the minimal working version first.

neqo-http3/src/features/extended_connect/session.rs
jesup: Note that the fairness bit can wait, so it wouldn't need to know about sessions. But it would need sendgroup/sendorder knowledge for every datagram, and would need to implement the webtransport algorithms for choosing what to send.

mxinden: Not great, agreed, but I think fine, especially if the alternative is a third queue as introduced in this pull request.

Follow-up question: do we need to implement sendgroup and sendorder for datagrams today, or can this be done in the future?

jesup:
@mxinden Given some of the issues I highlight above with moving the h3 datagram queue down into transport, let's consider an alternate:

Remove the queue from the transport.

Transport doesn't pre-queue stream data centrally either:  Connection::write_frames walks priority order and pulls from each SendStream's own buffer as space allows. Eliminating QuicDatagrams's internal VecDeque and replacing it with a pull hook (a small trait transport which http3's Session/DatagramQueue implements: "here's up to N bytes of room in this packet, got something?").  This just extends that existing pattern across the transport boundary instead of introducing a new one.

This would solve the "transport's queue doesn't handle max-age" as well.

This would change the external API for neqo-transport, of course.  Of course we could do both, if there is anything that would care (and not get rewritten to match) - keep the queue for anything that doesn't supply a pull hook.  But I suspect we would just modify any other sender (i.e. masque).

mxinden:
I think that is worth prototyping, yes. I would suspect that it is not worth its complexity.

I would suggest the following:

1. Implement max age in `neqo-transport` today, thus unblocking our ongoing WebTransport work.
2. In the future, introduce WebTransport datagram sendorder and send groups, either in `neqo-transport`, or first refactor `neqo-transport` according to your suggestion, then implement sendorder and send groups in `neqo-http3`'s WebTransport code.

jesup:
> I think that is worth prototyping, yes. I would suspect that it is not worth its complexity.
> 
> I would suggest the following:
> 
>     1. Implement max age in `neqo-transport` today, thus unblocking our ongoing WebTransport work.
> 
>     2. In the future, introduce WebTransport datagram sendorder and send groups, either in `neqo-transport`, or first refactor `neqo-transport` according to your suggestion, then implement sendorder and send groups in `neqo-http3`'s WebTransport code.

Since the very next set of patches to review is all the sendgroup/sendorder code, I think we should consider doing so now.  We'll need sendgroup/sendorder to pass Interop2026, and that's the largest and most important remaining PR after this one

jesup:
> Since the very next set of patches to review is all the sendgroup/sendorder code, I think we should consider doing so now. We'll need sendgroup/sendorder to pass Interop2026, and that's the largest and most important remaining PR after this one

The other option is to continue implementing with queues in both places, as today, and land the removal of the queue after the interop work is done (since this won't affect our interop score or usability in general; at worst it will mean max-age handling might be slightly non-spec-compliant in a way very hard to see.

This would actually be my preference; it can be reviewed and landed in a more deliberate manner.  I can write the patches for it now though

neqo-http3/src/features/extended_connect/session.rs
jesup: Yes, they're part of Interop2026.  Google is already using them in production code

jesup:
Bulk response to the AI reviews:

- Server callback-clamp ordering bug: fixed.  process_http3 now recomputes next_datagram_expiry after processing active connections instead of before, so it's never stale.  Added the invariant note on expire_datagram_queues's docs.
- Idle-connection test coverage: looked into driving via the returned Output::Callback instead of a hand-picked instant, but with this test's small workload the datagram gets sent via normal pacing before its 30ms max-age; so that assertion isn't testing what it looks like it's testing. Left the existing tests as-is; happy to discuss a pacing-independent way to cover the clamp specifically if someone thinks it's worth the extra complexity.
- connect_udp.rs hardcoded 1000: already superseded by the byte-budget work later in the stack.
- Capsule fallback bypassing the queue: confirmed, documented on DatagramOutcome::Sent, DatagramQueueOutcome::Ok, Session::send_datagram, and  ClientSession::connect_udp_send_datagram.
- Missing test for remaining_datagram_queue_capacity: added, pinning the full/decrement/no-underflow boundary.

jesup:
- send_datagram now expires stale datagrams before enqueuing, so a full queue of already-stale ones reports Expired instead of Overflowed/Dropped. Added a regression test.
- should_be_processed's expiry check is now gated on active(): it was spinning forever on a connection that went inactive, since nothing clears its cached expiry once that happens.
- WebTransportEvent, ConnectUdpEvent (client + server), and both ServerEvents (webtransport + connect-udp) are now #[non_exhaustive].
- process_all_datagram_queues now walks a HashSet<StreamId> of session ids instead of every stream on the connection.
- The server's process_http3 now tracks which handler owns the cached datagram-expiry minimum and only does a full rescan when that handler was reprocessed or removed, instead of rescanning every handler every call.
- Reverted the should_be_processed loop back to extend()
- Extracted a clamp_to_expiry helper shared by Http3Server::process_multiple and Http3Client::process_multiple_output, replacing the two slightly-different copies of the clamp logic.
- Moved the misplaced process_multiple comment, added a module-doc note that outgoingMaxAge stops being enforced at handover to the QUIC layer, noted DatagramOutcome::id() is FFI-only, DEFAULT_HARD_LIMIT is now re-exported and used in the backpressure test instead of a hardcoded 1000.

Note: per discussion, some of this may be mooted or heavily modified if we move the datagram queue logic all down into transport (likely); I'll deal with that in the move though.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -390,43 +402,203 @@ impl Session {
...
+                    buf,
+                    now,
+                )?;
+                if let DatagramTracking::Id(id) = id.into() {
+                    self.report_datagram_outcome(DatagramOutcome::Sent(id));
+                }
+                return Ok(DatagramQueueOutcome::Ok);
+            }
+            qdebug!("[{self}] peer does not support QUIC DATAGRAM");
+            return Err(Error::Transport(neqo_transport::Error::NotAvailable));
```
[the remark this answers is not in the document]
jesup: Agreed on flagging the behavior change in the PR description (previously fell through to conn.send_datagram()'s TooMuchData, now returns NotAvailable directly).  On the suggested error value: I'd rather keep Error::Transport(NotAvailable) than fold this into Error::Unavailable.  Unavailable already means "session not in Active state" a few lines above. "Peer will never support QUIC DATAGRAM" is a different, permanent condition, and Gecko's error mapping may want to distinguish "retry later" from "don't bother." We can revisit if there's a reason to collapse them that I'm missing.

neqo-http3/src/connection_client.rs
```diff
@@ -812,7 +826,20 @@ impl Http3Client {
...
+        // waiting on whatever unrelated timer happens to fire next.
+        match out {
+            OutputBatch::Callback(delay) => OutputBatch::Callback(
+                self.base_handler
+                    .next_datagram_expiry()
+                    .filter(|expiry| *expiry > now)
+                    .map_or(delay, |expiry| delay.min(expiry - now)),
+            ),
+            out => out,
+        }
```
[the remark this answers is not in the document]
jesup: Tried this, dropping it for now.  The server-side test gets a datagram stuck behind others by exploiting the server's anti-amplification limit right after the handshake; a client has no such limit, and these are tiny (2-byte) datagrams that coalesce into very few packets regardless of how many filler ones I queued, so I couldn't get it to reliably stall the way the server does without sinking more time into the harness than seemed worth it this round. clamp_to_expiry is now shared code (see the overview reply), so the logic isn't entirely uncovered; just not exercised client-side specifically.  Open to another pass at this if it matters enough to prioritize.

PR #3848, opened 2026-08-02
feat: set max_pto connection param to 7

mxinden:
Set the `ConnectionParameters` default for `max_pto` to 7, turning on the black-hole detector added in #3662 by default instead of leaving it opt-in.

A connection that sees 7 consecutive unacknowledged PTOs is almost certainly black-holed. Closing it then (`Error::TooManyPtos`, sent on the wire as `NO_VIABLE_PATH`) lets the application recover on a fresh path or another protocol rather than hanging until the idle timeout. At a 20ms RTT the 7th PTO fires around 2.5 to 5.7s, well inside the 30s idle timeout.

I plan to enable this on the Firefox side soon, so it is worth running neqo's own unit tests with the detector on by default already.


neqo-http3/tests/httpconn.rs
```diff
@@ -472,45 +433,62 @@ fn zerortt() {
...
+            }
+            break;
         }
-
-        if let Output::Callback(t) = hconn_c.process_output(now) {
-            now += t;
-        }
+        assert_eq!(hconn_c.state(), Http3State::Connected);
+
+        // Advance to the next keepalive either peer schedules.
```
martinthomson:
```suggestion
        // Advance to the next keepalive each peer schedules.
```

neqo-http3/tests/httpconn.rs
```diff
@@ -472,45 +433,62 @@ fn zerortt() {
...
+    // keepalive and its acknowledgement. No packet is ever lost, so the
+    // consecutive-PTO count stays at zero and max_pto never trips.
+    let mut now = now();
+    let end = now + ConnectionParameters::DEFAULT_IDLE_TIMEOUT * 4;
+    while now < end {
+        // Exchange all keepalives and acknowledgements pending at this instant.
+        loop {
+            if let Some(d) = hconn_c.process_output(now).dgram() {
+                if let Some(ack) = hconn_s.process(Some(d), now).dgram() {
+                    drop(hconn_c.process(Some(ack), now));
```
martinthomson: Dropping packets like this is going to make this test pretty unreliable.  Why not process_input() ?

neqo-http3/tests/httpconn.rs
```diff
@@ -472,45 +433,62 @@ fn zerortt() {
...
-
-        if let Output::Callback(t) = hconn_c.process_output(now) {
-            now += t;
-        }
+        assert_eq!(hconn_c.state(), Http3State::Connected);
+
+        // Advance to the next keepalive either peer schedules.
+        let c_cb = match hconn_c.process_output(now) {
+            Output::Callback(t) => t,
+            _ => Duration::from_secs(1),
```
martinthomson: This is going to hide problems because it drops packets, just like the above.

If this is functioning correctly, there should be no packets to send on the end of each loop, just a timeout.  After all, each endpoint is idle and you haven't advanced time.

neqo-http3/tests/httpconn.rs
```diff
@@ -472,45 +433,62 @@ fn zerortt() {
...
+    while now < end {
+        // Exchange all keepalives and acknowledgements pending at this instant.
+        loop {
+            if let Some(d) = hconn_c.process_output(now).dgram() {
+                if let Some(ack) = hconn_s.process(Some(d), now).dgram() {
+                    drop(hconn_c.process(Some(ack), now));
+                }
+                continue;
             }
+            if let Some(d) = hconn_s.process_output(now).dgram() {
```
martinthomson: The client is the only one that should initiate the keepalive here.  So this arm isn't really needed, nor do I expect it to do anything.

So you could drive the loop far more simply: have the client send a ping, get it acknowledged, consume the acknowledgment, then slide forward by half an idle timeout.  Throughout, unless the server is generating that acknowledgment, it should return the idle timeout every time.

larseggert (approved):
Bot comments seem to make sense

neqo-transport/src/connection/params.rs
larseggert, line 140:
```suggestion
    /// locally. `None` disables the check, leaving the idle timeout
```

larseggert:
@mxinden a rebase should make clippy happy. (Please tech your bot to let maintainers rebase for it?)

PR #3852, opened 2026-08-03
fix: reject 0-RTT downgrade of h3 datagram and webtransport settings

uwezkhan:
Without the change, on a resumed connection:

    ---- zero_rtt_new_server_setting_h3_datagram_smaller stdout ----
    assertion `left == right` failed
      left: Connected
     right: Closing(Application(265))

`handle_settings` already rejects a server that lowers a setting it sent before
0-RTT, but the list it walks holds only the three QPACK and header-size
settings. `SETTINGS_H3_DATAGRAM` and `SETTINGS_ENABLE_WEBTRANSPORT` also go into
the resumption token, so a client can send HTTP Datagrams and WebTransport
CONNECTs in 0-RTT against a remembered value the server is then free to
withdraw.

Before: the server omits `SETTINGS_H3_DATAGRAM` on the resumed connection,
`NegotiationState` flips to `Failed`, and the connection stays up.

After: the connection closes with H3_SETTINGS_ERROR, which RFC 9297, Section
2.1.1 asks of a client that stored the setting with its 0-RTT state.
draft-ietf-webtrans-http3, Section 3.2 says the same for the WebTransport
session limit.

Raising a setting from 0 to 1 stays allowed, so only the downgrade direction
changes. The tradeoff is that a server which quietly dropped either feature on
resumption now loses the connection. `SETTINGS_ENABLE_CONNECT_PROTOCOL` is left
out, since RFC 9220 has no comparable 0-RTT rule.

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1713,6 +1713,8 @@ impl Http3Connection {
                     HSettingType::MaxHeaderListSize,
                     HSettingType::MaxTableCapacity,
                     HSettingType::BlockedStreams,
+                    HSettingType::EnableWebTransport,
+                    HSettingType::EnableH3Datagram,
                 ] {
                     let zero_rtt_value = settings.get(*st);
                     let new_value = new_settings.get(*st);
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1713,6 +1713,12 @@ impl Http3Connection {
                     HSettingType::MaxHeaderListSize,
                     HSettingType::MaxTableCapacity,
                     HSettingType::BlockedStreams,
+                    // [RFC 9297, Section 2.1.1](https://www.rfc-editor.org/rfc/rfc9297.html#section-2.1.1)
+                    // requires a client that stored SETTINGS_H3_DATAGRAM with its
+                    // 0-RTT state to terminate the connection with H3_SETTINGS_ERROR
+                    // if the server's new value is smaller than the stored one.
+                    // [draft-ietf-webtrans-http3, Section 3.2](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3#section-3.2)
+                    // states the same rule for the WebTransport setting.
                     HSettingType::EnableWebTransport,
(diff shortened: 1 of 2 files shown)


neqo-http3/src/connection.rs
```diff
@@ -1670,6 +1670,8 @@ impl Http3Connection {
                     HSettingType::MaxHeaderListSize,
                     HSettingType::MaxTableCapacity,
                     HSettingType::BlockedStreams,
+                    HSettingType::EnableWebTransport,
+                    HSettingType::EnableH3Datagram,
```
mxinden, line 1723: Worth referencing the RFC?

[RFC 9297, Section 2.1.1](https://www.rfc-editor.org/rfc/rfc9297.html#section-2.1.1):

> "If a client stores the value of the SETTINGS_H3_DATAGRAM setting with their 0-RTT state, they MUST validate that the new value of the SETTINGS_H3_DATAGRAM setting sent by the server in the handshake is greater than or equal to the stored value; if not, the client MUST terminate the connection with error H3_SETTINGS_ERROR."

mxinden (approved):
As always, thank you for your help!

larseggert:
Might also want to rebase to pull in CI fixes I made yesterday, to get a clean run.

neqo-http3/src/connection.rs
uwezkhan: Good idea. Added a comment on those two entries citing RFC 9297, Section 2.1.1 and draft-ietf-webtrans-http3, Section 3.2.

uwezkhan:
Rebased onto main and pushed. The zero_rtt tests still pass locally.

larseggert:
Thank you!

uwezkhan:
@larseggert @mxinden  Is there any other area in the repo you’d recommend I look into next? Or is there any other Mozilla repo you’d suggest I contribute to? I’d be happy to take a look. Thanks!

larseggert:
Here are some suggestions that are actively being worked on:
- https://github.com/mozilla/nss-rs
- https://github.com/mozilla/happy-eyeballs

Issue #2927, opened 2025-01-30
MTU: Support ios, tvos and visionos


mxinden:
iOS does not ship with `<net/route.h>`, see https://github.com/mozilla/mtu/issues/82.

https://github.com/mozilla/mtu/pull/83 dropped support for the Apple OS variants, see e.g. code below:

https://github.com/mozilla/mtu/blob/8329753853c5cc0b695a8b72f953f8a1b9828c24/build.rs#L47-L52

https://github.com/mozilla/mtu/blob/8329753853c5cc0b695a8b72f953f8a1b9828c24/src/lib.rs#L101-L107

We should add support for `target_os` `ios`, `tvos` and `visionos` in the future.

PR #3858, opened 2026-08-06
fix(mtu): gate unix helpers on supported platforms, not just non-windows

shoemoney:
`mtu` doesn't pass `clippy -D warnings` on the Apple mobile targets. On `aarch64-apple-ios`,
`aarch64-apple-tvos` and `aarch64-apple-visionos` it fails with 12 errors each; `aarch64-apple-darwin`
is clean.

**To be clear about scope: no CI job builds these targets today, so nothing of yours is red.**
`clippy.yml` runs host-only with no `--target`, `check-android` runs `cargo ndk test --no-run` rather
than clippy, and `check-mtu.yml` covers Android/Linux/macOS/BSD/Solaris. This is lint hygiene on
platforms the crate already claims to compile for, not a broken gate.

## Cause

`asserted_const_with_type!`, the `routesocket` module, `unlikely_err` and `aligned_by` are gated
`#[cfg(not(target_os = "windows"))]`, but their only consumers are `bsd.rs` and `linux.rs`. On
ios/tvos/visionos neither consumer compiles, so everything behind that gate is dead code and the
`-D warnings` gate turns each one into an error.

Fixing the unsupported-platform stub alone would leave 10 of the 12 — the gate itself is the issue.

## The 12, identical on all three targets

```
error: unused macro definition: `asserted_const_with_type`
error: unused variable: `remote`
error: function `unlikely_err` is never used
error: function `aligned_by` is never used
error: type alias `AtomicRouteSocketSeq` is never used
error: type alias `RouteSocketSeq` is never used
error: static `SEQ` is never used
error: struct `RouteSocket` is never constructed
error: associated functions `new` and `new_seq` are never used
error: function `check_result` is never used
error: docs for function returning `Result` missing `# Errors` section
error: unneeded `return` statement
error: could not compile `mtu` (lib) due to 12 previous errors
```

## The change

A `supported_unix` cfg alias in `mtu/build.rs`, mirroring the existing `bsd` alias idiom:

```rust
supported_unix: {
    any(target_os = "macos", bsd, target_os = "linux", target_os = "android")
}
```

Six `#[cfg(not(target_os = "windows"))]` become `#[cfg(supported_unix)]` at the helper and test
sites, and the unsupported-platform `interface_and_mtu_impl` becomes non-`pub` like the other
platform impls (which is what drops `missing_errors_doc`), with the unused parameter and needless
return quieted.

**Left alone deliberately:** the `not(target_os = "windows")` on the `no_ipv6` binding in `inet_v6`.
It's the non-windows half of a pair — narrowing it would leave the binding undefined on iOS.

## Verification

| target | before | after |
|---|---|---|
| `aarch64-apple-ios` | 12 errors, exit 101 | **exit 0** |
| `aarch64-apple-tvos` | 12 errors, exit 101 | **exit 0** |
| `aarch64-apple-visionos` | 12 errors, exit 101 | **exit 0** |
| `aarch64-apple-darwin` | exit 0 | **exit 0** |

All three targets install prebuilt std on **stable** — no nightly, no `-Z build-std`.

Regression checks, all exit 0: `x86_64-unknown-linux-gnu`, `aarch64-linux-android`,
`x86_64-unknown-freebsd`, `x86_64-unknown-netbsd`, `x86_64-pc-windows-gnu`. **The BSD ones are the
load-bearing check** — they confirm `cfg_aliases` expands the nested `bsd` alias inside
`supported_unix`; had it not, BSD would have silently lost `routesocket`.

Also: `cargo test -p mtu` 9/9 pass on darwin, `cargo +nightly fmt --all -- --check` clean, and
`cargo hack clippy -p mtu --target aarch64-apple-ios --feature-powerset --no-dev-deps
--exclude-features gecko -- -D warnings` exits 0.

## Related, not closed

`#2927` ("MTU: Support ios, tvos and visionos", `help wanted`) asks for real MTU *lookup* on these
platforms — iOS has no `<net/route.h>`. This PR does not implement that; it only makes the existing
stub lint-clean. Worth linking, not closing.

## Known gap, deliberately not in this PR

`--all-targets` on iOS still hits one `error[E0080]`: `test::LOOPBACK` at `mtu/src/lib.rs:172` has no
iOS arm and falls through to `unreachable!()`, which const-eval executes. **Pre-existing** — on
pristine `main` iOS `--all-targets` has 12 lib errors plus 3 lib-test errors including this one;
with this PR it's the only one left. Picking loopback name/MTU values for iOS is a value decision
rather than a lint fix, so I've left it to you. Happy to follow up if you name the values.

### DIFF
--- a/mtu/build.rs
+++ b/mtu/build.rs
@@ -13,6 +13,14 @@ fn main() {
                 target_os = "netbsd",
                 target_os = "solaris"
             )
+        },
+        supported_unix: {
+            any(
+                target_os = "macos",
+                bsd,
+                target_os = "linux",
+                target_os = "android"
+            )
         }
     }
 }
--- a/mtu/src/lib.rs
+++ b/mtu/src/lib.rs
@@ -52,7 +52,7 @@ use std::{
     net::IpAddr,
 };
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(supported_unix)]
 macro_rules! asserted_const_with_type {
     ($name:ident, $t1:ty, $e:expr, $t2:ty) => {
         #[allow(
@@ -75,7 +75,7 @@ mod linux;
 #[cfg(target_os = "windows")]
 mod windows;
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(supported_unix)]
 mod routesocket;
 
 #[cfg(any(target_os = "macos", bsd))]
@@ -91,14 +91,14 @@ fn default_err() -> Error {
 }
 
 /// Prepare an error for cases that "should never happen".
-#[cfg(not(target_os = "windows"))]
+#[cfg(supported_unix)]
 fn unlikely_err(msg: String) -> Error {
     debug_assert!(false, "{msg}");
     Error::other(msg)
 }
 
 /// Align `size` to the next multiple of `align` (which needs to be a power of two).
-#[cfg(not(target_os = "windows"))]
+#[cfg(supported_unix)]
 const fn aligned_by(size: usize, align: usize) -> usize {
     if size == 0 {
         align
@@ -116,8 +116,8 @@ const fn aligned_by(size: usize, align: usize) -> usize {
     target_os = "visionos",
     target_os = "redox"
 ))]
-pub fn interface_and_mtu_impl(remote: IpAddr) -> Result<(String, usize)> {
-    return Err(default_err());
+fn interface_and_mtu_impl(_remote: IpAddr) -> Result<(String, usize)> {
+    Err(default_err())
 }
 
 /// Return the name and maximum transmission unit (MTU) of the outgoing network interface towards a
@@ -203,7 +203,7 @@ mod test {
     }
 
     #[test]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(supported_unix)]
     fn aligned_by() {
         for (size, align, expected) in [
             (0, 8, 8),
@@ -253,7 +253,7 @@ mod test {
 
     #[test]
     #[should_panic(expected = "test error")]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(supported_unix)]
     #[cfg(debug_assertions)]
     fn unlikely_error_panics() {
         crate::unlikely_err("test error".to_string());


larseggert:
@shoemoney are these APIs supported on iOS/tvOS/visionOS? If so, I would prefer to actually fix the code so the crate works there.

shoemoney:
Good question — I went and checked against the actual SDKs rather than guessing. The answer splits in two, and it's the reason I gated rather than fixed.

**The interface/MTU half is fully supported on all three.** `getifaddrs`/`freeifaddrs`, `if_indextoname`, `struct if_data` (so `ifi_mtu`), `struct sockaddr_dl` and `AF_LINK` are all declared in the iPhoneOS, AppleTVOS and XROS SDKs, and they are ABI-identical to macOS. I checked the ABI rather than assuming it — compiling this against each SDK succeeds:

```c
#include <stddef.h>
#include <sys/socket.h>
#include <net/if.h>
#include <net/if_dl.h>
#include <ifaddrs.h>
#define A(x, v) _Static_assert((x) == (v), #x " mismatch")
A(sizeof(struct if_data), 96);
A(offsetof(struct if_data, ifi_mtu), 8);
A(sizeof(struct sockaddr_dl), 20);
A(AF_LINK, 18);
A(PF_ROUTE, 17);
static int (*const g)(struct ifaddrs **) = getifaddrs;
static char *(*const i)(unsigned int, char *) = if_indextoname;
```

**The route-socket half is not.** `net/route.h` ships only in the macOS SDK. It is absent from iPhoneOS, AppleTVOS and XROS, and so are all the symbols we need out of it:

```
$ ls $(xcrun --sdk macosx --show-sdk-path)/usr/include/net/
bpf.h dlil.h ethernet.h if_arp.h if_dl.h if_llc.h if_media.h if_mib.h
if_types.h if_utun.h if_var_status.h if_var.h if.h kext_net.h ndrv.h
net_kev.h pfkeyv2.h route.h

$ ls $(xcrun --sdk iphoneos --show-sdk-path)/usr/include/net/
ethernet.h if_dl.h if_types.h if_var_status.h if_var.h if.h net_kev.h pfkeyv2.h
```

`RTM_GET`, `RTM_VERSION`, `RTA_DST`, `RTAX_MAX` and `struct rt_msghdr` are absent from those SDKs entirely, not just relocated — `grep -rw` over the whole include tree finds them only under macOS. (`PF_ROUTE` and `NET_RT_DUMP` *are* still in `sys/socket.h` on iOS, so you can open the socket and name the sysctl; you just have no declared way to build or parse a message over it. The sysctl `NET_RT_DUMP` route-table read has the same problem — it needs the same `rt_msghdr` layout.)

Tested with Xcode 26.3 (17C519): MacOSX26.2, iPhoneOS26.2, AppleTVOS26.2, XROS26.2.

The catch is that our bindings are checked in as pre-generated bindgen output, so `mtu` *does* compile for `aarch64-apple-{ios,tvos,visionos}` if you just point `bindings/macos.rs` at those targets — I tried it, and it builds and passes clippy on all three. But that means shipping a hand-carried copy of a kernel interface Apple deliberately does not vend on those platforms, with no header to keep us honest if the layout ever diverges. I did not want to do that behind your back, which is why the PR gates instead.

So: I don't think the routing-socket implementation can be made correct there. What I think *can* work, entirely on declared API, is to skip the route socket on the mobile Apple targets and get the outgoing interface a different way — `connect()` a `SOCK_DGRAM` socket to the remote (no packets sent, but the kernel does the route lookup), `getsockname()` for the source address, match that against the `getifaddrs` list to get the interface, then read `ifi_mtu` off the `AF_LINK` entry for that name. That reuses `if_name_mtu` and the `IfAddrs` walk in `bsd.rs` as-is; only `if_index_mtu` gets a mobile-Apple sibling.

The honest caveat is that I can only verify that on a real device, and I can't tell you today whether the sandbox lets a non-entitled app see anything useful from `getifaddrs` on iOS — Apple has been tightening link-layer introspection for years (the MAC-address zeroing in iOS 11 being the well-known one), and I have not found a definitive answer for `ifi_mtu` specifically. It may return a usable MTU, or it may return something the crate should treat as unknown.

Happy to go either way: land this PR as the gate and do the `connect`/`getsockname` implementation as a follow-up, or fold it into this PR if you'd rather have it in one go. If you have someone who can run a test binary on an actual iOS device, that would settle the sandbox question before either of us writes the code.

shoemoney:
Good question — I went and checked against the actual SDKs rather than guessing. The answer splits in two, and it's the reason I gated rather than fixed.

**The interface/MTU half is fully supported on all three.** `getifaddrs`/`freeifaddrs`, `if_indextoname`, `struct if_data` (so `ifi_mtu`), `struct sockaddr_dl` and `AF_LINK` are all declared in the iPhoneOS, AppleTVOS and XROS SDKs, and they are ABI-identical to macOS. I checked the ABI rather than assuming it — compiling this against each SDK succeeds:

```c
#include <stddef.h>
#include <sys/socket.h>
#include <net/if.h>
#include <net/if_dl.h>
#include <ifaddrs.h>
#define A(x, v) _Static_assert((x) == (v), #x " mismatch")
A(sizeof(struct if_data), 96);
A(offsetof(struct if_data, ifi_mtu), 8);
A(sizeof(struct sockaddr_dl), 20);
A(AF_LINK, 18);
A(PF_ROUTE, 17);
static int (*const g)(struct ifaddrs **) = getifaddrs;
static char *(*const i)(unsigned int, char *) = if_indextoname;
```

**The route-socket half is not.** `net/route.h` ships only in the macOS SDK. It is absent from iPhoneOS, AppleTVOS and XROS, and so are all the symbols we need out of it:

```
$ ls $(xcrun --sdk macosx --show-sdk-path)/usr/include/net/
bpf.h dlil.h ethernet.h if_arp.h if_dl.h if_llc.h if_media.h if_mib.h
if_types.h if_utun.h if_var_status.h if_var.h if.h kext_net.h ndrv.h
net_kev.h pfkeyv2.h route.h

$ ls $(xcrun --sdk iphoneos --show-sdk-path)/usr/include/net/
ethernet.h if_dl.h if_types.h if_var_status.h if_var.h if.h net_kev.h pfkeyv2.h
```

`RTM_GET`, `RTM_VERSION`, `RTA_DST`, `RTAX_MAX` and `struct rt_msghdr` are absent from those SDKs entirely, not just relocated — `grep -rw` over the whole include tree finds them only under macOS. (`PF_ROUTE` and `NET_RT_DUMP` *are* still in `sys/socket.h` on iOS, so you can open the socket and name the sysctl; you just have no declared way to build or parse a message over it. The sysctl `NET_RT_DUMP` route-table read has the same problem — it needs the same `rt_msghdr` layout.)

Tested with Xcode 26.3 (17C519): MacOSX26.2, iPhoneOS26.2, AppleTVOS26.2, XROS26.2.

The catch is that our bindings are checked in as pre-generated bindgen output, so `mtu` *does* compile for `aarch64-apple-{ios,tvos,visionos}` if you just point `bindings/macos.rs` at those targets — I tried it, and it builds and passes clippy on all three. But that means shipping a hand-carried copy of a kernel interface Apple deliberately does not vend on those platforms, with no header to keep us honest if the layout ever diverges. I did not want to do that behind your back, which is why the PR gates instead.

So: I don't think the routing-socket implementation can be made correct there. What I think *can* work, entirely on declared API, is to skip the route socket on the mobile Apple targets and get the outgoing interface a different way — `connect()` a `SOCK_DGRAM` socket to the remote (no packets sent, but the kernel does the route lookup), `getsockname()` for the source address, match that against the `getifaddrs` list to get the interface, then read `ifi_mtu` off the `AF_LINK` entry for that name. That reuses `if_name_mtu` and the `IfAddrs` walk in `bsd.rs` as-is; only `if_index_mtu` gets a mobile-Apple sibling.

The honest caveat is that I can only verify that on a real device, and I can't tell you today whether the sandbox lets a non-entitled app see anything useful from `getifaddrs` on iOS — Apple has been tightening link-layer introspection for years (the MAC-address zeroing in iOS 11 being the well-known one), and I have not found a definitive answer for `ifi_mtu` specifically. It may return a usable MTU, or it may return something the crate should treat as unknown.

Happy to go either way: land this PR as the gate and do the `connect`/`getsockname` implementation as a follow-up, or fold it into this PR if you'd rather have it in one go. If you have someone who can run a test binary on an actual iOS device, that would settle the sandbox question before either of us writes the code.

larseggert:
I'm inclined to leave this as-is until someone with actual hardware wants to use this on one of these platforms.

shoemoney:
closing...

PR #3859, opened 2026-08-06
fix(transport): back-pressure the outbound QUIC datagram queue

### LINKED ISSUE #2852: QUIC datagram outbound backpressure

mxinden:
When the outgoing datagram queue was full, `add_datagram` head-dropped the oldest queued datagram. On the MASQUE connect-udp path that silently discards an inner-connection packet the application already handed us, which looks like loss to the inner connection and hurts upload throughput.

Back-pressure instead of dropping, mirroring stream sends. `send_datagram` now always queues the datagram and returns whether the queue still has room:

- `Ok(true)`: queued, and space remains for more.
- `Ok(false)`: queued, but the queue is now full. The caller should stop producing datagrams until it receives `OutgoingDatagramSpaceAvailable`. Nothing already queued is dropped.

The datagram is accepted even when the queue is already at capacity. The application has already produced it, so neqo holds it here, ready to send, rather than leaving it in a queue upstream that neqo cannot reach once space frees. `Ok(false)` is a high-watermark signal to stop, not a rejection.

Once a slot frees while blocked, emit `OutgoingDatagramSpaceAvailable` (surfaced to HTTP/3 clients as `Http3ClientEvent::OutgoingDatagramSpaceAvailable`, analogous to `SendStreamWritable`) so the caller resumes.

This covers the QUIC datagram queue only. The connect-udp HTTP DATAGRAM Capsule fallback (used when the peer offers no QUIC datagram support) still drops on a full control stream and reports success; giving it backpressure is left for a follow-up (see the note on `write_datagram_capsule`).

Removes the now-unreachable `OutgoingDatagramOutcome::DroppedQueueFull` variant and the `dropped_queue_full` stat.

Fixes #2852.

Conflicts with https://github.com/mozilla/neqo/pull/3837.

### DIFF
--- a/neqo-http3/src/client_events.rs
+++ b/neqo-http3/src/client_events.rs
@@ -115,6 +115,8 @@ pub enum Http3ClientEvent {
     WebTransport(WebTransportEvent),
     /// `ConnectUdp` events
     ConnectUdp(ConnectUdpEvent),
+    /// The outgoing QUIC datagram queue has space again after having been full.
+    OutgoingDatagramSpaceAvailable,
 }
 
 #[derive(Debug, Default, Clone)]
@@ -290,6 +292,10 @@ impl ExtendedConnectEvents for Http3ClientEvents {
         };
         self.events.push(event);
     }
+
+    fn capsule_space_available(&self) {
+        self.datagram_space_available();
+    }
 }
 
 impl Http3ClientEvents {
@@ -298,6 +304,15 @@ impl Http3ClientEvents {
         if stream_type == StreamType::BiDi {
             self.events.push(Http3ClientEvent::RequestsCreatable);
         }
+    }
+
+    /// Emit the outgoing QUIC datagram queue's resume event, forwarding
+    /// `neqo-transport`'s [`ConnectionEvent::OutgoingDatagramSpaceAvailable`].
+    ///
+    /// [`ConnectionEvent::OutgoingDatagramSpaceAvailable`]: neqo_transport::ConnectionEvent::OutgoingDatagramSpaceAvailable
+    pub(crate) fn datagram_space_available(&self) {
+        self.events
+            .push_unique(Http3ClientEvent::OutgoingDatagramSpaceAvailable);
     }
 
     /// Add a new `AuthenticationNeeded` event
--- a/neqo-http3/src/connect_udp.rs
+++ b/neqo-http3/src/connect_udp.rs
@@ -62,18 +62,25 @@ pub trait ClientSession {
 
     /// Send a connect-udp datagram.
     ///
+    /// # Returns
+    ///
+    /// `Ok(false)` when the outgoing QUIC datagram queue is full; the sender
+    /// should then wait for an [`OutgoingDatagramSpaceAvailable`] event.
+    ///
     /// # Errors
     ///
     /// It may return [`Error::InvalidStreamId`] if a stream does not exist anymore.
     /// The function returns `TooMuchData` if the supply buffer is bigger than
     /// the allowed remote datagram size.
+    ///
+    /// [`OutgoingDatagramSpaceAvailable`]: crate::Http3ClientEvent::OutgoingDatagramSpaceAvailable
     fn connect_udp_send_datagram<I: Into<DatagramTracking>>(
         &mut self,
         session_id: StreamId,
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()>;
+    ) -> Res<bool>;
 }
 
 impl ClientSession for Http3Client {
@@ -118,7 +125,7 @@ impl ClientSession for Http3Client {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<bool> {
         qtrace!("connect_udp_send_datagram session:{session_id:?}");
         let (conn, handler) = self.connection_and_handler();
         handler.connect_udp_send_datagram(conn, session_id, buf, id, now)
@@ -153,6 +160,7 @@ trait Handler {
         now: Instant,
     ) -> Res<()>;
 
+    /// Returns `Ok(false)` when the outgoing QUIC datagram queue is full.
     fn connect_udp_send_datagram<I: Into<DatagramTracking>>(
         &self,
         conn: &mut Connection,
@@ -160,7 +168,7 @@ trait Handler {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()>;
+    ) -> Res<bool>;
 }
 
 impl Handler for Http3Connection {
@@ -232,7 +240,7 @@ impl Handler for Http3Connection {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<bool> {
         self.extended_connect_send_datagram(session_id, conn, buf, id, now)
     }
 }
@@ -256,6 +264,7 @@ pub(crate) trait ServerHandler {
         now: Instant,
     ) -> Res<()>;
 
+    /// Returns `Ok(false)` when the outgoing QUIC datagram queue is full.
     fn connect_udp_send_datagram<I: Into<DatagramTracking>>(
         &mut self,
         conn: &mut Connection,
@@ -263,7 +272,7 @@ pub(crate) trait ServerHandler {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()>;
+    ) -> Res<bool>;
 }
 
 impl ServerHandler for Http3ServerHandler {
@@ -300,7 +309,7 @@ impl ServerHandler for Http3ServerHandler {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<bool> {
         self.mark_needs_processing();
         self.base_handler_mut()
             .connect_udp_send_datagram(conn, session_id, buf, id, now)
@@ -391,7 +400,7 @@ impl ServerSession {
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<bool> {
         let session_id = self.stream_handler.stream_id();
         self.stream_handler
             .handler
--- a/neqo-http3/src/features/extended_connect/connect_udp_session.rs
+++ b/neqo-http3/src/features/extended_connect/connect_udp_session.rs
@@ -6,6 +6,7 @@
 
 use std::{
     fmt::{self, Display, Formatter},
+    num::NonZeroUsize,
     time::Instant,
 };
 
@@ -19,6 +20,7 @@ use crate::{
         session::{DgramContextIdError, State},
     },
     frames::{FrameReader, StreamReaderRecvStreamWrapper, capsule::Capsule},
+    send_message::SendMessage,
 };
 
 #[derive(Debug)]
@@ -132,19 +134,31 @@ impl Protocol for Session {
         let mut dgram_data = Encoder::default();
         self.write_datagram_prefix(&mut dgram_data);
         dgram_data.encode(buf);
-
-        if conn.stream_avail_send_space(self.session_id)? < dgram_data.len() {
-            qdebug!("Not enough space to send datagram capsule, dropping it.");
-            return Ok(());
-        }
         // TODO: Make Capsule abstract over either an owned (Bytes) or borrowed (&[u8]) type
         // to avoid this allocation.
         let capsule = Capsule::Datagram {
             payload: Bytes::from(Vec::from(dgram_data)),
         };
+
+        // The capsule goes on the control stream wrapped in an HTTP/3 DATA frame, so the
+        // flow-control window has to hold both wrappers, not just the payload.
+        let needed = SendMessage::data_frame_len(capsule.encoded_len());
+        if conn.stream_avail_send_space(self.session_id)? < needed {
+            qdebug!("[{self}] datagram capsule exceeds control-stream flow-control space");
+            // Ask to be told when the stream can hold a capsule this size again.
+            // Repeated refusals overwrite this, so the event tracks the most recent
+            // size rather than the largest; the sender retries and re-arms.
+            if let Some(watermark) = NonZeroUsize::new(needed) {
+                conn.stream_set_writable_event_low_watermark(self.session_id, watermark)?;
+            }
+            return Err(Error::FlowControlLimit);
+        }
         let mut enc = Encoder::default();
         capsule.encode(&mut enc);
         control_stream_send.send_data_atomic(conn, enc.as_ref(), now)?;
+        // Drop the watermark a previous refusal raised, so neqo-transport stops
+        // suppressing writable events for other users of the control stream.
+        conn.stream_set_writable_event_low_watermark(self.session_id, NonZeroUsize::MIN)?;
         qtrace!("[{self}] sent datagram via HTTP DATAGRAM Capsule");
         Ok(())
     }
--- a/neqo-http3/src/features/extended_connect/session.rs
+++ b/neqo-http3/src/features/extended_connect/session.rs
@@ -61,6 +61,9 @@ pub(crate) struct Session {
     /// CONNECT request.
     protocol: Box<dyn Protocol>,
     draining: bool,
+    /// Set when a datagram capsule was refused with `FlowControlLimit`, so a
+    /// resume event fires once the control stream is writable again.
+    datagram_capsule_blocked: bool,
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
@@ -120,6 +123,7 @@ impl Session {
             events,
             protocol,
             draining: false,
+            datagram_capsule_blocked: false,
         }
     }
 
@@ -150,6 +154,7 @@ impl Session {
             events,
             protocol,
             draining: false,
+            datagram_capsule_blocked: false,
         })
     }
 
@@ -226,6 +231,15 @@ impl Session {
             self.state = State::Done;
         }
         Ok(())
+    }
+
+    fn stream_writable(&mut self) {
+        // The control stream has flow-control space again; if a datagram capsule
+        // was refused, tell the sender it can resume.
+        if self.datagram_capsule_blocked {
+            self.datagram_capsule_blocked = false;
+            self.events.capsule_space_available();
+        }
     }
 
     fn close(&mut self, close_type: CloseType) {
@@ -390,18 +404,21 @@ impl Session {
         self.control_stream_send.send_data(conn, buf, now)
     }
 
+    /// Send a datagram, as a QUIC datagram or, when the peer offers no QUIC
+    /// datagram support, an HTTP DATAGRAM Capsule. Returns `Ok(false)` when the
+    /// outgoing QUIC datagram queue is full.
+    ///
     /// # Errors
     ///
-    /// Returns an error if:
-    /// - The session is not in Active state (`Error::Unavailable`).
-    /// - QUIC datagram or HTTP DATAGRAM Capsule sending fails.
+    /// `Error::Unavailable` if the session is not Active; other errors if
+    /// sending the datagram or capsule fails.
     pub(crate) fn send_datagram<I: Into<DatagramTracking>>(
         &mut self,
         conn: &mut Connection,
         buf: &[u8],
         id: I,
         now: Instant,
-    ) -> Res<()> {
+    ) -> Res<bool> {
         qtrace!("[{self}] send_datagram state={:?}", self.state);
         if self.state != State::Active {
             qdebug!("[{self}]: cannot send datagram in {:?} state.", self.state);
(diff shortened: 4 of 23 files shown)


neqo-transport/src/quic_datagrams.rs
```diff
@@ -127,6 +134,12 @@ impl QuicDatagrams {
                 debug_assert!(builder.len() <= builder.limit());
                 stats.frame_tx.datagram += 1;
                 tokens.push(recovery::Token::Datagram(*dgram.tracking()));
+                if self.blocked {
+                    // A slot just freed up; let the application resume sending
+                    // QUIC datagrams.
+                    self.blocked = false;
+                    self.conn_events.datagram_space_available();
+                }
```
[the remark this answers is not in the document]
martinthomson: I would move this check outside the loop.  Not just because of the MTU-based dropping, but because you only need to check once anyway.

I'm also thinking about the way in which we might want to use this to tune the queue depth, rather than having it be a fixed or configured value.

What we want is for the depth to be enough that this code never runs out of datagrams to send, but also that it is not so deep that the number left each time is ever so high that they spend a long time in the buffer.  Think about having two traps: one for when this function drains the buffer (to increase the buffer size) and one for when this function fails to drop below a lower watermark (to reduce the buffer size).  In a well-behaved application, we should not need a deep buffer at all, but it will be hard to keep the buffer from draining as CWND ramps up.

martinthomson (changes requested):
Looks like the LLM review has picked up a bunch of nasties here.

neqo-transport/src/events.rs
```diff
@@ -185,6 +187,12 @@ impl ConnectionEvents {
             .push_back(ConnectionEvent::Datagram(data.to_vec()));
     }
 
+    pub fn datagram_space_available(&self) {
+        // Only one such signal needs to be outstanding at a time.
+        self.remove(|evt| matches!(evt, ConnectionEvent::OutgoingDatagramSpaceAvailable));
+        self.insert(ConnectionEvent::OutgoingDatagramSpaceAvailable);
+    }
```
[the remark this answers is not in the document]
mxinden: Addressed.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -127,6 +134,12 @@ impl QuicDatagrams {
...
             } else if tokens.is_empty() {
                 // If the packet is empty, except packet headers, and the
                 // datagram cannot fit, drop it.
```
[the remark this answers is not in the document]
mxinden: Done.

neqo-transport/src/quic_datagrams.rs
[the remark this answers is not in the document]
mxinden: > I would move this check outside the loop. Not just because of the MTU-based dropping, but because you only need to check once anyway.

Done.

> I'm also thinking about the way in which we might want to use this to tune the queue depth, rather than having it be a fixed or configured value.

Discussed out-of-band. Deferred for now. If we identify this code path as a bottleneck, we can consider these optimizations.

neqo-transport/src/events.rs
```diff
@@ -185,6 +187,12 @@ impl ConnectionEvents {
             .push_back(ConnectionEvent::Datagram(data.to_vec()));
     }
 
+    pub fn datagram_space_available(&self) {
+        // Only one such signal needs to be outstanding at a time. `insert`
+        // already skips a duplicate, so there is no need to remove and re-add.
```
mxinden: Will be simplified with https://github.com/mozilla/neqo/pull/3905.

mxinden:
@martinthomson friendly ping. Could you give this another review?

neqo-transport/src/quic_datagrams.rs
```diff
@@ -127,37 +134,68 @@ impl QuicDatagrams {
...
+        // Sending or dropping datagrams above may have freed slots. Resume the
+        // application if it was blocked and the queue is now below capacity.
+        // Covering the drop path is what stops a queue emptied entirely by
+        // `DroppedTooBig` from stalling. The length guard matters because the
+        // queue can sit one over capacity (a full queue still accepts one more,
+        // returning `false`), so one freed slot may not suffice.
+        if self.blocked && self.datagrams.len() < self.max_queued_outgoing_datagrams {
+            self.blocked = false;
+            self.conn_events.datagram_space_available();
+        }
```
mxinden, line 169: Following up on your question in https://github.com/mxinden-bot/neqo/pull/9#discussion_r3869619323, an space-available event is emitted as soon as any space becomes available, not just when the queue is entirely drained.

neqo-http3/src/connect_udp.rs
```diff
@@ -62,18 +62,26 @@ pub trait ClientSession {
 
     /// Send a connect-udp datagram.
     ///
+    /// # Returns
+    ///
+    /// `Ok(true)` if the datagram was queued with room to spare, or `Ok(false)`
+    /// if it was queued but the queue is now full and the sender should stop
+    /// until an [`OutgoingDatagramSpaceAvailable`] event (backpressure).
```
mxinden, line 69: > This same comment is repeated in many places. I understand we're threading a result value through the stack, but this is a lot of repetition.

@larseggert do you feel strongly about this? I think these comments are helpful to understand the return value not just at the API surface, but as well on these internal functions.

mxinden (commented):
@larseggert while waiting on Martin, could you help out with a review?

Following up on a couple of comments from https://github.com/mxinden-bot/neqo/pull/9:

> A lot of the comments talk about the queue being "full" when elsewhere they more correctly say that the queue is above its high water mark. It would be good to be consistently use the water mark terminology, since "full" strongly implies nothing more can be queued, which is false.

On `Ok(false)` the producer SHOULD stop sending, this is not a MAY, thus I think the "full" terminology is right.

> Is there not a better way to make sure there is exactly one than a remove followed by an insert? (Also elsewhere.)

Event de-duplication is now improved with https://github.com/mozilla/neqo/pull/3905. Happy to do more follow-ups, but preference for doing so in a separate pull request.

neqo-http3/src/connect_udp.rs
larseggert: Nope. I just find them a bit repetitive and long, but 🤷

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
```diff
@@ -75,3 +76,70 @@ fn max_datagram_size_smaller_than_session_prefix() {
...
+    );
+
+    wt.exchange_packets();
+    assert!(
+        wt.server
+            .events()
+            .any(|e| matches!(e, Http3ServerEvent::OutgoingDatagramSpaceAvailable { .. })),
+        "OutgoingDatagramSpaceAvailable was not forwarded to the HTTP/3 server"
+    );
+}
```
larseggert, line 145: Nit: Could these two tests be consolidated?

neqo-http3/src/client_events.rs
```diff
@@ -115,6 +115,9 @@ pub enum Http3ClientEvent {
     WebTransport(WebTransportEvent),
     /// `ConnectUdp` events
     ConnectUdp(ConnectUdpEvent),
+    /// The outgoing QUIC datagram queue has space again after having been full.
+    /// The application can resume sending HTTP/3 datagrams.
```
larseggert, line 119:
```suggestion
```

Nit: What the app does is really up to it.

neqo-http3/src/server_connection_events.rs
```diff
@@ -44,6 +44,9 @@ pub enum Http3ServerConnEvent {
     },
     /// Connection state change.
     StateChange(Http3State),
+    /// The outgoing QUIC datagram queue has space again after having been full,
+    /// so a sender that received backpressure can resume.
```
larseggert, line 48:
```suggestion
    /// The outgoing QUIC datagram queue has space again after having been full.
```
Nit: As above.

neqo-http3/src/server_events.rs
```diff
@@ -277,6 +277,12 @@ pub enum Http3ServerEvent {
         stream_id: StreamId,
         priority: Priority,
     },
+    /// The outgoing QUIC datagram queue has space again after having been full,
+    /// so a session on this connection that received backpressure (a `false`
+    /// return from `send_datagram`) can resume sending.
```
larseggert, line 282:
```suggestion
    /// The outgoing QUIC datagram queue has space again after having been full.
```
(Nit)

neqo-transport/src/events.rs
```diff
@@ -78,6 +77,9 @@ pub enum ConnectionEvent {
         id: u64,
         outcome: OutgoingDatagramOutcome,
     },
+    /// Space has become available in the outgoing QUIC datagram queue after it
+    /// was full. The application can resume sending QUIC datagrams.
```
larseggert, line 81:
```suggestion
    /// Space has become available in the outgoing QUIC datagram queue after it
    /// was full.
```

larseggert (approved):
I wonder if a module-level comment block somewhere would be a better place to describe the logic this implements in one place and completely. A lot of the comments are pretty repetitive and wordy, and they could then be much shorter (= point to the mod-level doc).

Not a blocker.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -411,22 +422,22 @@ impl Session {
...
         if conn.remote_datagram_size() == 0 && self.protocol.datagram_capsule_support() {
             qtrace!("[{self}] remote_datagram_size is 0, trying HTTP DATAGRAM Capsule");
-            return self.protocol.write_datagram_capsule(
-                &mut self.control_stream_send,
-                conn,
-                buf,
-                now,
-            );
+            // The Capsule path does not report backpressure; see
+            // `Protocol::write_datagram_capsule`.
```
martinthomson: Do we really want this?

I'm looking at `<connect_udp_session::Session as Protocol>::write_datagram_capsule()` and it *drops* capsules that it cannot write.

I think that this would be a good time to reconsider this choice.  We basically never fail to send a datagram currently -- only reporting an error when the datagram is too large, which covers the case where the peer doesn't support datagrams.  Reporting dropped datagram capsules with `Err(whatever)` rather than a silent drop on capsules is probably better than this approach.

The fact that this always returns `true` when the datagram is written is fine.  Capsules are buffered in the stream, which provides a hard upper limit, rather than the soft limit we have for other variants.

The other thing that concerns me is the lack of events for datagrams that are sent this way.  That's more workable if hitting a flow control limit causes an error to be returned, less so otherwise.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -652,6 +663,11 @@ pub(crate) trait Protocol: Debug + Display {
     fn datagram_capsule_support(&self) -> bool;
 
     /// Write a datagram as an HTTP DATAGRAM Capsule to the control stream.
+    ///
+    /// NOTE: On a full control stream this drops the datagram and still reports
+    /// success, so the Capsule path applies no backpressure. Unlike the QUIC
+    /// datagram queue, it neither reports "no space" nor emits a resume event.
+    /// To be fixed in the future.
```
martinthomson: See above.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -447,51 +448,319 @@ fn outgoing_datagram_queue_full() {
...
+#[test]
+fn backpressure_respecting_sender_never_stalls() {
+    const TOTAL: u64 = 50;
+
+    let (mut client, mut server) = connect_datagram();
+    let mut next_id: u64 = 1;
+    let mut received: u64 = 0;
+    let mut blocked = false;
+
+    for _ in 0..10_000 {
```
martinthomson: This doesn't iterate this many times.  I think that you could make this 0..TOTAL safely.  You should get at least one datagram through each loop iteration.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -447,51 +448,319 @@ fn outgoing_datagram_queue_full() {
...
+        // Send while the queue accepts and we have not been told to stop.
+        while !blocked && next_id <= TOTAL {
+            let payload = vec![u8::try_from(next_id % 256).unwrap()];
+            match client.send_datagram(payload, Some(next_id)) {
+                // Queued with space remaining.
+                Ok(true) => next_id += 1,
+                // Queued, but the queue is now full: stop until resumed.
+                Ok(false) => {
+                    next_id += 1;
```
martinthomson: You could move the increment out of the match.  The error case panics.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -447,51 +448,319 @@ fn outgoing_datagram_queue_full() {
...
+            }
+        }
+
+        // Move one datagram-bearing packet to the server, if any.
+        if let Some(out) = client.process_output(now()).dgram() {
+            server.process_input(out, now());
+        }
+
+        // Count everything the server received.
+        while let Some(event) = server.next_event() {
```
martinthomson:
```suggestion
        for event in server.events() {
```

neqo-transport/src/quic_datagrams.rs
```diff
@@ -127,37 +134,68 @@ impl QuicDatagrams {
...
     }
 
     /// Add a datagram to the send queue.
     ///
+    /// Unless it returns an error (see below), the QUIC datagram is queued. The
+    /// returned bool reports whether the outgoing QUIC datagram queue still had
+    /// room afterwards:
+    ///
+    /// - `Ok(true)`: queued, and space remains for more.
+    /// - `Ok(false)`: queued, but the queue is now full. The application should stop producing QUIC
```
martinthomson:
```suggestion
    /// - `Ok(false)`: queued, but the queue is full. The application should stop producing QUIC
```

neqo-http3/src/features/extended_connect/tests/webtransport/datagrams.rs
mxinden: 👍 Done

mxinden:
@martinthomson following up on your comment regarding missing datagram capsule backpressure in https://github.com/mozilla/neqo/pull/3859#discussion_r3946177997. Good challenge, thanks. I believe I found an elegant solution. The capsule path now returns `FlowControlLimit` instead of dropping the datagram, and arms [`stream_set_writable_event_low_watermark`](https://github.com/mozilla/neqo/blob/2d9598a4e47c48c417415092e178f249fd7f72a0/neqo-transport/src/connection/mod.rs#L3946) so an `OutgoingDatagramSpaceAvailable` event fires once the control stream can hold a capsule again.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -132,16 +134,23 @@ impl Protocol for Session {
...
+        if conn.stream_avail_send_space(self.session_id)? < needed {
+            qdebug!("[{self}] datagram capsule exceeds control-stream flow-control space");
+            // Ask to be told when the stream can hold a capsule this size again.
+            if let Some(watermark) = NonZeroUsize::new(needed) {
+                conn.stream_set_writable_event_low_watermark(self.session_id, watermark)?;
+            }
+            return Err(Error::FlowControlLimit);
+        }
         let mut enc = Encoder::default();
         capsule.encode(&mut enc);
```
martinthomson, line 157: At some point we'll not encode this way.  The temporary buffer is wasteful.  Plumbing through to `BufferedStream::send_atomic_with` would be an improvement if this is hot at any point.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
```diff
@@ -132,16 +134,23 @@ impl Protocol for Session {
...
         };
+
+        // The capsule goes on the control stream wrapped in an HTTP/3 DATA frame, so the
+        // flow-control window has to hold both wrappers, not just the payload.
+        let needed = SendMessage::data_frame_len(capsule.encoded_len());
+        if conn.stream_avail_send_space(self.session_id)? < needed {
+            qdebug!("[{self}] datagram capsule exceeds control-stream flow-control space");
+            // Ask to be told when the stream can hold a capsule this size again.
+            if let Some(watermark) = NonZeroUsize::new(needed) {
+                conn.stream_set_writable_event_low_watermark(self.session_id, watermark)?;
```
martinthomson, line 152: If this function is called multiple times, the size of the last write failure determines when the event is generated.  Is that OK?  I think that it's worth noting that if you fail with 1000 bytes then attempt 100 bytes, the notice will come when you can send the latter, not necessarily the former.  This differs from the regular datagram sending arrangements, where it's purely about the count.

The other thing to consider is whether there might be cause to set a low watermark on the stream for other reasons.  I don't think that there is in this case, but this is something that could trap other users of the same stream.

neqo-http3/tests/connect_udp.rs
```diff
@@ -716,3 +671,233 @@ fn connect_udp_session_rejected_by_webtransport_create_stream() {
...
+    assert!(
+        !client
+            .events()
+            .any(|e| matches!(e, Http3ClientEvent::OutgoingDatagramSpaceAvailable)),
+        "resume event fired before the window reopened"
+    );
+
+    // The proxy reads the buffered capsules and grants more flow-control credit,
+    // reopening the control stream, which must surface the resume event.
+    exchange_packets(&mut client, &mut proxy, false, None);
```
martinthomson, line 836: You should get the watermark event.  Why is that not being checked?

neqo-http3/tests/connect_udp.rs
```diff
@@ -716,3 +671,233 @@ fn connect_udp_session_rejected_by_webtransport_create_stream() {
...
+/// framing does not must be refused, not accepted and then truncated and lost.
+///
+/// After the CONNECT request and one `BIG` datagram the window holds `SMALL`'s
+/// payload but not its framing; `SMALL` sits mid-gap so the boundary tolerates
+/// small changes in what the request consumed.
+#[test]
+fn datagram_capsule_at_flow_control_boundary_is_refused_not_lost() {
+    const WINDOW: u64 = 150;
+    const BIG: usize = 113;
+    const SMALL: usize = 4;
```
martinthomson: I'm not following the formula here.  From what I can see, each of these capsules should have a three and two byte overhead respectively.  That's nowhere near 150.  So where is the remainder going?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -228,6 +233,15 @@ impl Session {
         Ok(())
     }
 
+    fn stream_writable(&mut self) {
+        // The control stream has flow-control space again; if a datagram capsule
+        // was refused, tell the sender it can resume.
+        if self.datagram_capsule_blocked {
+            self.datagram_capsule_blocked = false;
+            self.events.capsule_space_available();
+        }
```
martinthomson, line 242: Should this clear the low watermark set when the capsule write failed?

martinthomson (approved):

neqo-http3/tests/connect_udp.rs
mxinden: I am not sure I understand. The `OutgoingDatagramSpaceAvailable` event is checked right below.

neqo-http3/src/features/extended_connect/connect_udp_session.rs
mxinden: > If this function is called multiple times, the size of the last write failure determines when the event is generated. Is that OK? I think that it's worth noting that if you fail with 1000 bytes then attempt 100 bytes, the notice will come when you can send the latter, not necessarily the former. This differs from the regular datagram sending arrangements, where it's purely about the count.

I will add a note. I believe this is fine. No matter whether the higher or lower watermark wins, I assume we will get enough stream credit eventually, thus the event will trigger, thus I believe this is fine.

> The other thing to consider is whether there might be cause to set a low watermark on the stream for other reasons. I don't think that there is in this case, but this is something that could trap other users of the same stream.

I believe there is none today. That said, if one would be set in the future, it might cause some datagrams to be lost, depending on the users retransmit strategy, but it would not lead to a dead-lock, given that eventually the server should grant us enough stream credit.

neqo-http3/tests/connect_udp.rs
mxinden: The remainder is the preceding HTTP CONNECT request. I refactored the test to remove the implicit dependency.

neqo-http3/src/features/extended_connect/session.rs
mxinden: Done with the latest push.

PR #3861, opened 2026-08-11
chore: Allow `assert_is_empty`

larseggert:


martinthomson (commented):
Hard to see how this is a net benefit.  Code we write will call `is_empty()`, so why should the tests differ?

larseggert:
@martinthomson would you prefer to suppress `assert_is_empty` instead?

Agree that the rationale is pretty weak:
>A boolean assertion only reports that the emptiness check failed. It does not show what the asserted value contained.

martinthomson:
Yeah, I think that's a misfiring lint.  Anywhere we have `!x.is_empty()` and this sort of thing makes sense, but a direct `is_empty()` call is annoying.

martinthomson (approved):

PR #3870, opened 2026-08-11
chore: minor AI review follow-ups to https://github.com/mozilla/neqo/pull/3835

mxinden:
Minor AI review follow-ups to https://github.com/mozilla/neqo/pull/3835

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1445,10 +1445,11 @@ impl Http3Connection {
         &self,
         session_id: StreamId,
     ) -> Res<extended_connect::stats::SessionStats> {
-        self.webtransport_session(session_id)?
-            .borrow()
-            .stats()
-            .ok_or(Error::InvalidStreamId)
+        let Some(stats) = self.webtransport_session(session_id)?.borrow().stats() else {
+            debug_assert!(false, "a WebTransport session should always have stats");
+            return Err(Error::InvalidStreamId);
+        };
+        Ok(stats)
     }
 
     pub(crate) fn extended_connect_close_session(
--- a/neqo-transport/src/recovery/mod.rs
+++ b/neqo-transport/src/recovery/mod.rs
@@ -268,12 +268,20 @@ impl LossRecoverySpace {
             self.remove_packet(p);
             eliciting |= p.ack_eliciting();
             if p.lost() {
-                // A packet declared lost that is subsequently acknowledged was never
-                // really lost, so take it back out of the loss counters. Every packet
-                // marked lost was counted by `count_lost`, so this cannot underflow.
                 stats.late_ack += 1;
-                stats.lost -= 1;
-                stats.bytes_lost -= p.len();
+                if let Some(reduced) = stats.lost.checked_sub(1) {
+                    stats.lost = reduced;
+                } else {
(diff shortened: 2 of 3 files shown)


larseggert (approved):

neqo-http3/src/connection.rs
```diff
@@ -1445,10 +1445,14 @@ impl Http3Connection {
...
-            .stats()
-            .ok_or(Error::InvalidStreamId)
+        let stats = self.webtransport_session(session_id)?.borrow().stats();
+        // `webtransport_session` guarantees a WebTransport session, whose `stats()`
+        // is always populated; the error path is unreachable in practice.
+        debug_assert!(
+            stats.is_some(),
+            "a WebTransport session should always have stats"
+        );
+        stats.ok_or(Error::InvalidStreamId)
```
```suggestion
        if let Some(stats) = self.webtransport_session(session_id)?.borrow().stats() {
            stats
        } else {
             debug_assert!(false, "a WebTransport session should always have stats");
             Err(Error::InvalidStreamId)
         }
```
The text of the message captures the same information as the comment.  Keep a leash on that LLM of yours :)

neqo-transport/src/recovery/mod.rs
```diff
@@ -270,10 +270,13 @@ impl LossRecoverySpace {
...
                 // really lost, so take it back out of the loss counters. Every packet
-                // marked lost was counted by `count_lost`, so this cannot underflow.
+                // marked lost was counted by `count_lost`, so the counters stay
+                // balanced; saturate defensively so a future accounting change cannot
+                // wrap these `usize` counters in release builds.
                 stats.late_ack += 1;
-                stats.lost -= 1;
-                stats.bytes_lost -= p.len();
+                debug_assert!(stats.lost >= 1 && stats.bytes_lost >= p.len());
+                stats.lost = stats.lost.saturating_sub(1);
```
martinthomson: The right way to do this is:

```rust
if Some(reduced) = stats.lost.checked_sub(1) {
    stats.lost = reduced;
} else {
    debug_assert!(false, "spurious losses should have been lost already");
}
if Some(reduced) = stats.bytes_lost.checked_sub(p.len()) {
  stats.bytes_lost = reduced;
} else {
  debug_assert!(false, "spurious lost bytes should have been counted already");
}
```
No need for a saturation check; instead, avoid the decrement entirely if the accounting is off.

Also, no need for comments if you have the nice assertion text.

neqo-transport/src/recovery/mod.rs
```diff
@@ -270,10 +270,13 @@ impl LossRecoverySpace {
             if p.lost() {
                 // A packet declared lost that is subsequently acknowledged was never
                 // really lost, so take it back out of the loss counters. Every packet
-                // marked lost was counted by `count_lost`, so this cannot underflow.
+                // marked lost was counted by `count_lost`, so the counters stay
+                // balanced; saturate defensively so a future accounting change cannot
+                // wrap these `usize` counters in release builds.
```
martinthomson: End the comment after the first period if you are adding the checks.  (GitHub won't let me suggest this because it's shit.)

martinthomson (approved):
Annoying opinionated review!

larseggert:
@mxinden can we finalize this?

PR #3871, opened 2026-08-12
chore: Reformat the profile PR comment

larseggert:
And make it show all data.

### DIFF
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -336,15 +336,39 @@ jobs:
           REPOSITORY: ${{ github.repository }}
           RUN_ID: ${{ github.run_id }}
         run: |
-          ARTIFACTS=$(gh api "/repos/$REPOSITORY/actions/runs/$RUN_ID/artifacts" \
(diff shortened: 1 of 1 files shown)


.github/workflows/profile.yml
```diff
@@ -336,15 +336,39 @@ jobs:
...
-            echo "### Performance Profiles"
+            echo "### Performance profiles for [profiler.firefox.com](https://profiler.firefox.com/)"
             echo
-            if [ -n "$ARTIFACTS" ]; then
-              echo "Performance profiles for [profiler.firefox.com](https://profiler.firefox.com/):"
-              echo
-              echo "$ARTIFACTS"
+            if [ -n "$BENCHES$COMPARISONS" ]; then
+              section "Benchmarks" "$BENCHES"
+              section "Comparisons" "$COMPARISONS"
```
martinthomson, line 371: Rather than put the (potentially large) output of `links()` in a variable, could you instead put the call to links inline?  A simpler jq test can be used to determine whether to include the sections at all.  I think that it would be OK to test for the "pr" side having some results to report.

larseggert: https://github.com/mozilla/neqo/pull/3875

PR #3873, opened 2026-08-12
fix: `server_rejects_server_only_tparams` needs to init NSS

larseggert:
CI didn't catch that, because all tests run within a single process, and so if another test inits NSS, we don't see a failure.


.github/workflows/check.yml
```diff
@@ -121,6 +121,32 @@ jobs:
           # shellcheck disable=SC2086
           cargo hack test $BUILD_TYPE --locked --feature-powerset --all-targets --exclude-features gecko,ci,bench,build-fuzzing-corpus,test-fixture --mutually-exclusive-features blapi,disable-encryption,default
 
+      - name: Check test isolation
+        # `cargo test` runs a whole test binary in one process, so a test that
+        # depends on global state another test set up (e.g. NSS initialization)
+        # can pass there while failing when run alone. Catch that by running
+        # every test in its own process.
+        run: |
```
martinthomson: Having this run on every CI run is probably a waste.  If this only ran on a single configuration, that would be fine, wouldn't it?

martinthomson (commented):
I'm not sure that the full one-test-per-process thing is really necessary.  At some level, this is just a matter of ensuring that we initialize fixtures when they are needed.

PR #3885, opened 2026-08-17
feat: Show more QUIC info in test.sh

larseggert:


test/test.sh
martinthomson, line 20: Would it be possible to test for the tools that are involved in running this?  `hash tshark || echo "Need tshark"`, and tmux, and tsomethingelse?

PR #3886, opened 2026-08-17
chore: Refactor SNI code

larseggert:
Make it IMO more readable and is a fallout of implementing PPDPLPMTUD.

### DIFF
--- a/neqo-transport/src/crypto.rs
+++ b/neqo-transport/src/crypto.rs
@@ -26,7 +26,7 @@ use std::{
 
 use enum_map::EnumMap;
 use neqo_common::{
-    Buffer, Encoder, Role, expect_usize,
+    Buffer, Encoder, Role,
     hex::{Hex, HexSnipMiddle},
     qdebug, qinfo, qtrace, to_u64,
 };
@@ -1461,9 +1461,6 @@ impl Display for CryptoStates {
     }
 }
 
-/// What one CRYPTO frame covers: an offset into the stream and a length.
-type Chunk = (u64, usize);
-
 #[derive(Debug, Default)]
 pub struct CryptoStream {
     tx: TxBuffer,
@@ -1629,7 +1626,7 @@ impl CryptoStreams {
             cs: &mut CryptoStream,
             space: PacketNumberSpace,
             tokens: &mut recovery::Tokens,
-            (offset, len): Chunk,
+            (offset, len): (u64, usize),
             stats: &mut FrameStats,
         ) {
             cs.tx.mark_as_sent(offset, len);
@@ -1650,17 +1647,16 @@ impl CryptoStreams {
             if offset == 0 {
                 neqo_common::write_item_to_fuzzing_corpus("find_sni", data);
             }
-            let sliced = (sni_slicing && offset == 0)
-                .then(|| slice_sni(data, offset, builder.limit()))
-                .flatten();
-            let written = if let Some((first, second)) = sliced {
-                (
-                    write_at(data, offset, first, builder),
-                    write_at(data, offset, second, builder),
-                )
-            } else {
+            let written = match (sni_slicing && offset == 0)
+                .then(|| slice_sni(data, builder.limit()))
+                .flatten()
+            {
+                Some((first, second)) => (
+                    write_chunk(offset, data, first, builder),
+                    write_chunk(offset, data, second, builder),
+                ),
                 // No SNI found, slicing disabled, or data not at offset 0: write the entire data.
-                (write_chunk(offset, data, builder), None)
+                None => (write_chunk(offset, data, 0..data.len(), builder), None),
             };
 
             match written {
@@ -1671,6 +1667,7 @@ impl CryptoStreams {
                     mark_as_sent(cs, space, tokens, second, stats);
                     // We only end up in this arm if we successfully sliced above. In that case,
                     // don't try and fit more crypto data into this packet.
+                    // So an empty chunk must stay `Some`, or the SNI's second half lands here too.
                     break;
                 }
             }
@@ -1678,13 +1675,17 @@ impl CryptoStreams {
     }
 }
 
-/// Write a CRYPTO frame with as much of `data` as fits, or `None` if not even one byte does.
+/// Write a CRYPTO frame with as much of `data[range]` as fits, or `None` if not even one byte
+/// does. `data` itself starts at `offset` in the stream.
 fn write_chunk<B: Buffer>(
     offset: u64,
     data: &[u8],
+    range: Range<usize>,
     builder: &mut packet::Builder<B>,
-) -> Option<Chunk> {
-    let mut header_len = 1 + Encoder::varint_len(offset) + 1;
+) -> Option<(u64, usize)> {
+    let at = offset + to_u64(range.start);
+    let data = &data[range];
+    let mut header_len = 1 + Encoder::varint_len(at) + 1;
 
     if builder.remaining() < header_len + 1 {
         return None;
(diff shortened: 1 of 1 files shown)


neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+}
+
+/// Truncate both halves so that they fill packets of `limit` bytes roughly evenly.
+const fn limit_chunks(
+    left: Range<usize>,
+    right: Range<usize>,
+    limit: usize,
+) -> (Range<usize>, Range<usize>) {
+    // `Range::len` is not `const`, so subtract instead.
+    let (left_len, right_len) = (left.end - left.start, right.end - right.start);
```
martinthomson: This doesn't need to be const, so this sort of haxx isn't a pro move, it's more of an obfuscation trick.

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+    let data = &data[range];
+    let mut header_len = 1 + Encoder::varint_len(at) + 1;
+
+    if builder.remaining() < header_len + 1 {
+        return None;
+    }
+    // Calculate len based on minimum of the data and the remaining space (less the header).
+    let length = min(data.len(), builder.remaining() - header_len);
+    // The header counted only one byte for the length varint; correct that and redo.
+    header_len += Encoder::varint_len(to_u64(length)) - 1;
```
martinthomson: You can compute the length field length from `range.len()`.  It saves doing it twice.  If you need to truncate down, you might truncate two bytes (64 becomes 63 and thereby reduces the header length by an extra one), but it's still single-shot.

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
-            } else {
-                // SNI slicing disabled or data not at offset 0, write the entire data.
-                (write_chunk(offset, data, builder), None)
+            let written = match (sni_slicing && offset == 0)
+                .then(|| slice_sni(data, builder.limit()))
+                .flatten()
+            {
+                Some((first, second)) => (
+                    write_chunk(offset, data, first, builder),
+                    write_chunk(offset, data, second, builder),
```
martinthomson:
```suggestion
                    write_chunk(offset + first.start, &data[first], builder),
                    write_chunk(offset + second.start, &data[second], builder),
```

Simplifies the `write_chunk()` interface greatly, which currently gets a lot of redundant information.

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+                ),
+                // No SNI found, slicing disabled, or data not at offset 0: write the entire data.
+                None => (write_chunk(offset, data, 0..data.len(), builder), None),
```
martinthomson:
```suggestion
                None => (write_chunk(0, data, builder), None),
```

neqo-transport/src/crypto.rs
```diff
@@ -1622,38 +1622,11 @@ impl CryptoStreams {
...
-            Some((offset, length))
-        }
-
         fn mark_as_sent(
             cs: &mut CryptoStream,
             space: PacketNumberSpace,
             tokens: &mut recovery::Tokens,
-            offset: u64,
-            len: usize,
+            (offset, len): (u64, usize),
```
martinthomson: Not sure that I like this change.

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
-                    )
-                } else {
-                    // No SNI found, write the entire data.
-                    (write_chunk(offset, data, builder), None)
-                }
-            } else {
-                // SNI slicing disabled or data not at offset 0, write the entire data.
-                (write_chunk(offset, data, builder), None)
+            let written = match (sni_slicing && offset == 0)
+                .then(|| slice_sni(data, builder.limit()))
```
martinthomson: If you want to split the CRYPTO frame, such that you have two CRYPTO frames in each packet, you should be dividing by `builder.limit() - 6`.  That's to account for the headers on each frame, at their minimum size.  Otherwise, this will sometimes overflow weirdly.  As it is, you probably want -8 instead, because it would be unusual for there to be less than 64 bytes in each chunk.

...

```suggestion
                .then(|| slice_sni(data, limit))
```

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+    // `Range::len` is not `const`, so subtract instead.
+    let (left_len, right_len) = (left.end - left.start, right.end - right.start);
+    if left_len + right_len <= limit {
+        // Nothing to do. Both chunks will fit into one packet, meaning the SNI isn't spread
+        // over multiple packets. But at least it's in two unordered CRYPTO frames.
+        (left, right)
+    } else if left_len <= limit {
+        // `left` is short enough to fit into this packet. So send from the *end*
+        // of `right`, so that the second half of the SNI is in another packet.
+        // When `left_len == limit` this empties `right`; see `write_frame` for why that is OK.
```
martinthomson: These are overly commented.  The condition makes it clear what is going on.

As for the comment this references in `write_frame`, that's not right.  The reason this is fine is that we only mark the parts that were sent as sent.  The rest will be sent later.

```suggestion
        // The SNI is at the start of `right`, so send the end of it.
```
Plus whatever wrapping.

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+        let skip = right_len + left_len - limit;
+        (left, right.start + skip..right.end)
+    } else if right_len <= limit {
+        // `right` is short enough to fit into this packet. So only send a part of `left`.
+        // The SNI begins at the end of `left`, so send the beginning of it in this packet.
```
martinthomson:
```suggestion
        // The SNI begins at the end of `left`, so send the beginning of it in this packet.
```

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
+const fn limit_chunks(
+    left: Range<usize>,
+    right: Range<usize>,
+    limit: usize,
+) -> (Range<usize>, Range<usize>) {
+    // `Range::len` is not `const`, so subtract instead.
+    let (left_len, right_len) = (left.end - left.start, right.end - right.start);
+    if left_len + right_len <= limit {
+        // Nothing to do. Both chunks will fit into one packet, meaning the SNI isn't spread
+        // over multiple packets. But at least it's in two unordered CRYPTO frames.
```
martinthomson:
```suggestion
```

neqo-transport/src/crypto.rs
```diff
@@ -1703,47 +1647,105 @@ impl CryptoStreams {
...
-                        write_chunk(left_offset, left, builder),
-                    )
-                } else {
-                    // No SNI found, write the entire data.
-                    (write_chunk(offset, data, builder), None)
-                }
-            } else {
-                // SNI slicing disabled or data not at offset 0, write the entire data.
-                (write_chunk(offset, data, builder), None)
+            let written = match (sni_slicing && offset == 0)
```
martinthomson:
```suggestion
            let limit = builder.limit().saturating_sub(8);  // With space for two CRYPTO frame headers.
            let written = match (sni_slicing && offset == 0 && limit > 0)
```

PR #3888, opened 2026-08-17
fix: Tighter implementation of Section 6.2.3 of RFC 9002

larseggert:


neqo-transport/src/crypto.rs
```diff
@@ -76,6 +76,8 @@ pub struct Crypto {
     tls: Agent,
     streams: CryptoStreams,
     states: CryptoStates,
+    /// Whether [`Self::resend_unacked_early`] has already resent anything.
+    resent_early: bool,
```
martinthomson: I just want to flag the meaning of "resent".  It's very much not the same as "resend".

neqo-transport/src/crypto.rs
```diff
@@ -1561,13 +1581,10 @@ impl CryptoStreams {
...
-    pub fn resend_unacked(&mut self, space: PacketNumberSpace) {
-        if space != PacketNumberSpace::ApplicationData
-            && let Some(cs) = self.get_mut(space)
-        {
-            cs.tx.unmark_sent();
-        }
+    /// This can help speed up handshake times. Returns whether anything was queued to resend.
+    pub fn resend_unacked(&mut self, space: PacketNumberSpace) -> bool {
+        space != PacketNumberSpace::ApplicationData
+            && self.get_mut(space).is_some_and(|cs| cs.tx.unmark_sent())
```
martinthomson, line 1630: Given that we never call this with ApplicationData, would this be better?


```suggestion
        debug_assert_ne!(space, PacketNumberSpace::ApplicationData);
        self.get_mut(space).is_some_and(|cs| cs.tx.unmark_sent())
```

neqo-transport/src/recv_stream.rs
```diff
@@ -203,6 +203,21 @@ impl RxStreamOrderer {
         Self::default()
     }
 
+    /// Whether every byte of `[offset, offset + len)` has already been received or read.
+    /// Call this before [`Self::inbound_frame`], which wipes the ability to determine this.
+    #[must_use]
+    pub fn covered(&self, offset: u64, len: usize) -> bool {
+        let end = offset + to_u64(len);
```
martinthomson, line 242: Any sense in a shortcut here?


```suggestion
        let end = offset + to_u64(len);
        if end > self.end {
            return false;
        }
```

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2142,3 +2142,124 @@ fn client_initial_with_token() {
...
+/// up the one chance an endpoint gets to resend ahead of the PTO.
+#[test]
+fn crypto_resend_chance_survives_useless_duplicate() {
+    fixture_init();
+    let mut client = default_client();
+    let mut server = default_server();
+    let mut now = now();
+
+    // Give the server only the first part of the ClientHello. It cannot answer yet, so it has no
+    // CRYPTO of its own outstanding, and SNI slicing puts offset 0 in this first datagram.
```
martinthomson: Is this comment right?  The slicing puts offset 0 in the *second* datagram.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2142,3 +2142,124 @@ fn client_initial_with_token() {
...
+    let ch1 = client.process_output(now).dgram().expect("a datagram");
+    let ch2 = client.process_output(now).dgram().expect("a datagram");
+    server.process_input(ch1, now);
+
+    // A duplicate here finds nothing to resend.
+    let dup = duplicate_crypto(&mut client, &mut now);
+    assert_eq!(crypto_sent_for(&mut server, dup, &mut now), 0);
```
martinthomson, line 2265: This is misleading.  The duplicate CRYPTO frame will be received by the server and buffered because it is data it hasn't seen before.

The result should be that the server takes that byte (which is a bad byte, so the handshake would fail).  Hence what you are seeing here.  This is fine, but this is a busted test case.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2142,3 +2142,124 @@ fn client_initial_with_token() {
...
+    let (_initial, rest) = split_datagram(&dup);
+    let (handshake, _short) = split_datagram(&rest.expect("a coalesced Handshake packet"));
+    assert_handshake(&handshake);
+    assert_eq!(crypto_sent_for(&mut client, handshake, &mut now), 0);
+}
+
+/// A duplicate that arrives while nothing is outstanding has nothing to resend, so it must not use
+/// up the one chance an endpoint gets to resend ahead of the PTO.
+#[test]
+fn crypto_resend_chance_survives_useless_duplicate() {
```
martinthomson, line 2252: This test case is a bit broken.  It's worth having something in this shape, so the code might not need to change, but the commentary needs a lot of improvement.

neqo-transport/src/connection/tests/handshake.rs
```diff
@@ -2142,3 +2142,124 @@ fn client_initial_with_token() {
     client.process_input(datagram(server_initial(&ci, &[0x01])), now());
     assert_eq!(client.stats().dropped_rx, dropped + 1);
 }
+
+/// Produce a datagram carrying CRYPTO the peer already has, in whatever spaces are available.
+/// The frame needs a packet to ride along in, so wait out any pacing delay first.
+fn duplicate_crypto(c: &mut Connection, now: &mut Instant) -> Datagram {
```
martinthomson: This might be better named "make_dummy_crypto" or something like that.  Because it's not always a duplicate in the tests that rely on it.

neqo-transport/src/crypto.rs
larseggert: I think we do, at least for ping?

neqo-transport/src/connection/tests/handshake.rs
larseggert: No, the first. We sent Y-end, 0-X in the first and X-Y in the second. IIRC.

PR #3890, opened 2026-08-18
fix: `pacer_spend_pacing_limited` was defunct

larseggert:
Because it never advanced `now`.

### DIFF
--- a/neqo-transport/benches/pacer.rs
+++ b/neqo-transport/benches/pacer.rs
@@ -17,29 +17,33 @@ use neqo_transport::Pacer;
 const RTT: Duration = Duration::from_millis(50);
 const MTU: usize = 1_350;
 const CWND: usize = MTU * 100;
+const CALLS: usize = 1_000;
 
-/// PRIMARY: `spend()` when pacing is enabled and each call is credit-limited.
-/// This is the dominant case during a paced bulk transfer: the pacer has just
-/// enough credit for each packet and the call does the full division.
-///
-/// For pacing to actually limit, the inter-packet interval
-/// `RTT * MTU / (CWND * 2)` must exceed GRANULARITY (1 ms).
-/// With RTT=50ms, MTU=1350: CWND must be ≤ 33 750 bytes.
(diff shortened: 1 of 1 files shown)


neqo-transport/benches/pacer.rs
```diff
@@ -13,33 +13,44 @@ use std::{hint::black_box, time::Duration};
...
+/// PRIMARY: `spend()` with pacing enabled, in the steady state of a paced bulk
+/// transfer: the sender waits out the pacing interval, then spends the credit
+/// that accrued over it, so `self.c` returns to zero on every call.
 ///
-/// For pacing to actually limit, the inter-packet interval
-/// `RTT * MTU / (CWND * 2)` must exceed GRANULARITY (1 ms).
-/// With RTT=50ms, MTU=1350: CWND must be ≤ 33 750 bytes.
-/// `CWND_LIMITED`=MTU*10=13 500 gives a 2.5ms inter-packet interval.
+/// The interval also has to be one a real sender would wait out, which is a
+/// property of `next()`, not of `spend()`.
```
martinthomson: I'm not sure that the new comment is really reflective of what the test is attempting to achieve.  It seems to be relative to a previous state, which isn't what comments are for.

neqo-transport/benches/pacer.rs
```diff
@@ -75,10 +87,12 @@ fn pacer_spend_disabled(c: &mut Criterion) {
         b.iter_batched(
             || Pacer::new(false, now, CWND, MTU),
             |mut p| {
-                for _ in 0..1_000 {
+                // `spend()` only stores `now` here, so the pacer has to escape
+                // each iteration or the stores are all eliminated.
+                for _ in 0..CALLS {
                     p.spend(now, RTT, CWND, MTU);
+                    black_box(&p);
```
martinthomson, line 84: I realize that I don't understand how `black_box()` operates after reading this change.

Why is this one inconsistent with the one on line 74?  They would seem to both achieve the same thing.

And then there is the `black_box()` call that wraps the same constants on line 49.

😵‍💫

PR #3891, opened 2026-08-18
fix: Report packet and datagram sizes accurately in qlog

larseggert:
Report each packet at its own length and payload length on the wire, including the padding header protection adds.

Add datagrams_sent, datagrams_received, datagram_dropped and packet_buffered, with a datagram_id on packet events, so a trace accounts for every byte.

Hold event timestamps to the last one written, which qlog requires.

Log local transport parameters, and only count as dropped what preprocessing actually discards.

### DIFF
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -129,15 +129,16 @@ impl Qlog {
         F: FnOnce() -> Option<qlog::events::EventData>,
     {
         self.add_event_with_stream(now, |s, now| {
-            f().map_or(Ok(false), |ev_data| {
-                s.add_event_data_with_instant(ev_data, now).map(|()| true)
-            })
+            let Some(ev_data) = f() else {
+                return Ok(false);
+            };
+            s.add_event_data_with_instant(ev_data, now)?;
+            Ok(true)
         });
     }
 
-    /// If logging enabled, closure is given the Qlog stream to write events and
-    /// frames to, along with the time to write them at. Reports whether it
-    /// wrote an event.
+    /// If logging enabled, closure is given the Qlog stream to write timestamped events and
+    /// frames to. Reports whether it wrote an event.
     pub fn add_event_with_stream<F>(&mut self, now: Instant, f: F)
     where
         F: FnOnce(&mut QlogStreamer, Instant) -> Result<bool, qlog::Error>,
@@ -295,22 +296,14 @@ mod test {
         // Back at the earlier time, and so must be logged at it.
         log.add_event_at(|| Some(EV_DATA), now);
 
-        // Absolute times depend on the trace's reference time, so compare the two.
+        // Events carry the same data, so if written at the same time they produce identical lines.
         let output = contents.to_string();
-        let times = output
+        let events = output
             .lines()
             .filter(|l| l.contains("spin_bit_updated"))
-            .filter_map(|l| {
-                l.split("\"time\":")
-                    .nth(1)?
-                    .split(',')
-                    .next()?
-                    .parse::<f64>()
-                    .ok()
-            })
             .collect::<Vec<_>>();
-        assert_eq!(times.len(), 2, "expected both events in {output}");
-        assert_eq!(times[0], times[1], "clamped to a non-event in {output}");
+        assert_eq!(events.len(), 2);
+        assert_eq!(events[0], events[1]);
     }
 
     #[test]
--- a/neqo-common/src/qlog.rs
+++ b/neqo-common/src/qlog.rs
@@ -6,6 +6,7 @@
 
 use std::{
     cell::RefCell,
+    cmp::max,
     fmt::{self, Display},
     fs::OpenOptions,
     io::BufWriter,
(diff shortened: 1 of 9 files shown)


neqo-common/src/qlog.rs
```diff
@@ -102,24 +109,38 @@ impl Qlog {
...
         F: FnOnce() -> Option<qlog::events::EventData>,
     {
-        self.add_event_with_stream(|s| {
-            if let Some(ev_data) = f() {
-                s.add_event_data_with_instant(ev_data, now)?;
-            }
-            Ok(())
+        self.add_event_with_stream(now, |s, now| {
+            f().map_or(Ok(false), |ev_data| {
+                s.add_event_data_with_instant(ev_data, now).map(|()| true)
```
martinthomson:
```suggestion
                s.add_event_data_with_instant(ev_data, now)?;
                Ok(true)
```

I know that the map is technically correct, but it's just weird.

neqo-transport/src/connection/tests/mod.rs
```diff
@@ -142,8 +143,12 @@ fn zero_len_cid_client(local_addr: SocketAddr, remote_addr: SocketAddr) -> Conne
 }
 
 pub fn new_server(params: ConnectionParameters) -> Connection {
+    new_server_with_qlog(params).0
+}
+
+fn new_server_with_qlog(params: ConnectionParameters) -> (Connection, test_fixture::SharedVec) {
     fixture_init();
-    let (log, _contents) = new_neqo_qlog();
+    let (log, contents) = new_neqo_qlog();
```
martinthomson, line 151: Is this really necessary to have on every test?  I know that it was there previously, but what benefit do we get, as opposed to having a wrapper in the qlog module?  Or a helper that calls `set_qlog()` after creating the server.

neqo-transport/src/connection/tests/qlog.rs
```diff
@@ -0,0 +1,571 @@
...
+fn packets_name_their_datagram() {
+    let (mut client, contents) = new_client_with_qlog(ConnectionParameters::default());
+    let mut server = default_server();
+    connect(&mut client, &mut server);
+    drop(client);
+
+    let trace = contents.to_string();
+    let datagrams = named(&trace, "transport:datagrams_sent")
+        .filter_map(|l| number(l, "\"datagram_ids\":["))
+        .collect::<Vec<_>>();
```
martinthomson: Couldn't you collect into the `HashMap` and save a lot of code?

neqo-transport/src/connection/tests/qlog.rs
```diff
@@ -0,0 +1,571 @@
...
+                if let (Some(id), Some(len)) = (number(e, "\"datagram_id\":"), raw_length(e)) {
+                    per.entry(id).or_default().1 += len;
+                }
+            }
+        }
+        let off = per
+            .iter()
+            .filter(|(_, (a, b))| a != b)
+            .map(|(id, (a, b))| (*id, *a, *b))
+            .collect::<Vec<_>>();
```
martinthomson: You don't need to collect here.

neqo-transport/src/qlog.rs
```diff
@@ -56,37 +57,61 @@ fn to_hex<T: AsRef<[u8]>>(v: T) -> String {
     Hex::new(v).to_string()
 }
 
-pub fn connection_tparams_set(qlog: &mut Qlog, tph: &TransportParametersHandler, now: Instant) {
+/// Some qlog fields are `u32` and not optional, while the varints they carry
+/// are wider. Saturate rather than wrap.
+fn to_u32(v: u64) -> u32 {
```
martinthomson:
```suggestion
fn force_u32(v: u64) -> u32 {
```

Let's make it clear that this is not something that is gentle or lossless.

neqo-transport/src/connection/mod.rs
```diff
@@ -1819,23 +1899,44 @@ impl Connection {
...
-                        neqo_common::write_item_to_fuzzing_corpus("packet", packet.data());
-                        (packet, remainder)
-                    }
-                    Err(e) => {
-                        qinfo!("[{self}] Garbage packet: {e}");
-                        self.stats.borrow_mut().pkt_dropped("Garbage packet");
-                        break;
-                    }
-                };
+            // Bound here so the `cid_manager` borrow ends before `self` is needed below.
```
```suggestion
```

neqo-transport/src/connection/mod.rs
```diff
@@ -1819,23 +1899,44 @@ impl Connection {
...
+                Err(e) => {
+                    qinfo!("[{self}] Garbage packet: {e}");
+                    self.stats.borrow_mut().pkt_dropped("Garbage packet");
+                    // The header did not parse, so the packet type is unknown.
+                    self.drop_rest_of_datagram(
+                        None,
+                        slc_len,
+                        dgram_len,
+                        datagram_id,
+                        PacketDroppedTrigger::Invalid,
```
martinthomson: If this trigger captures the reasons we pass as strings to `pkt_dropped` effectively, then you could move the `pkt_dropped` call into `drop_rest_of_datagram`.  Is it enough for that?

martinthomson (commented):
Overall, this seems good.  It's a pretty big change, even if it is all justifiable.  I'm starting to think that Claude isn't good for keeping a tight leash on the size of changes.  Though maybe that's OK, it's not easy to review.

larseggert:
Fair comment about the size of the diff. Some of it comes from me wanting to not introduce more clippy suppressions for long methods, which led to refactoring changes being part of this PR. Ditto for changing some symbol names to what I thought were more suitable terms.

PR #3893, opened 2026-08-18
fix(transport): respond to each key update only once

mxinden:
A key update is detected from the first packet that decrypts with the new key phase, which the decrypt path recognizes from the key state having recorded a single packet number. A network duplicate of that packet decrypts with the same keys and leaves its recorded range unchanged, so it is detected as a key update again: `key_update_received` runs a second time and trips the `read_update_time.is_none()` assertion in `maybe_update_write`. With assertions compiled out it instead drives an unsolicited key update once the read keys have rotated, ratcheting the key phase ahead of the peer.

Record the read epoch of the update we last responded to and ignore any detection at or below it. A key phase is only updated once, so a duplicate can no longer re-enter the key update path, before or after the read keys rotate.

Reported at [Bugzilla bug 2063380](https://bugzilla.mozilla.org/show_bug.cgi?id=2063380).

### DIFF
--- a/neqo-transport/src/connection/tests/keys.rs
+++ b/neqo-transport/src/connection/tests/keys.rs
@@ -222,6 +222,64 @@ fn key_update_consecutive() {
     check_discarded(&mut client, &dgram, false, 1, 0);
 }
 
+// Receiving the same key update packet twice must not be treated as two key
+// updates. The packet carrying a new key phase is decrypted with a key state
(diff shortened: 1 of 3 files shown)


martinthomson (approved):

mxinden:
Merging here. I will follow-up on some of code review comments in a new pull request, as they don't need to be part of one atomic change.

PR #3896, opened 2026-08-19
fix: limit discontiguous STREAM reassembly ranges (alternative)

larseggert:
Simpler alternative to #3798

### DIFF
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -199,7 +199,7 @@ impl RxStreamOrderer {
     const RANGE_TARGET: usize = 4096;
 
     /// Maximum number of gaps allowed in the data a stream holds.
-    const MAX_GAPS: usize = 4096;
+    pub(crate) const MAX_GAPS: usize = 4096;
 
     /// Limit the entries a stream can be made to hold, which flow control does not.
     ///
--- a/neqo-transport/src/recv_stream.rs
+++ b/neqo-transport/src/recv_stream.rs
@@ -17,7 +17,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use neqo_common::{Buffer, Role, expect_usize, qtrace, to_u64};
+use neqo_common::{Buffer, Role, expect_usize, qtrace, qwarn, to_u64};
 use smallvec::SmallVec;
 use strum::Display;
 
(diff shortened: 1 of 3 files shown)


neqo-transport/src/recv_stream.rs
```diff
@@ -258,7 +284,7 @@ impl RxStreamOrderer {
             }
             // new_end > new_start >= end, so direct assignment is correct.
             self.end = new_end;
-            return;
+            return self.check_gap_limit();
```
martinthomson: This can be moved to the else branch on L283.  You don't need to check unless new_start > self.end.

neqo-transport/src/recv_stream.rs
```diff
@@ -362,6 +399,21 @@ impl RxStreamOrderer {
             // exists, so self.end is already correct — the max() is a no-op in that case.
             self.end = max(self.end, new_end);
         }
+
+        self.check_gap_limit()
```
martinthomson: Here too, I think that this can be run only adjacent to L394, where a new range is inserted.

neqo-transport/src/recv_stream.rs
```diff
@@ -362,6 +399,21 @@ impl RxStreamOrderer {
...
+    }
+
+    #[cfg(test)]
+    fn runs(&self) -> usize {
```
martinthomson:
```suggestion
    fn count_runs(&self) -> usize {
```

neqo-transport/src/recv_stream.rs
```diff
@@ -350,9 +376,20 @@ impl RxStreamOrderer {
 
         if !to_add.is_empty() {
             self.received += to_u64(to_add.len());
-            if extend {
-                if let Some((_, buf)) = self.data_ranges.range_mut(..=new_start).next_back() {
+            if let Some(prev_start) = extend {
+                // Absorb the range that follows, now that nothing separates them.
+                let add_end = new_start + to_u64(to_add.len());
+                let next = self
```
martinthomson: Oh, this is subtle.  This is the case where the new data closes a gap, so you will always have both a prev range and a next range.  You have to pull the next range off first, because you can't hold two mutable references to the ranges.  Then you can then combine the three into one.  

The two `Some()` checks below will always pass, except for the case where next becomes `None`, which happens when the two outer ranges are too long already.  In that case you add the new stuff to prev, but leave next alone.

Worth a bit of commentary, I think.  It took me a while to work this part out.

martinthomson (approved):
I absolutely do not care about the RxStreamOrderer public API.  If we could only make it public conditionally, that would be fine with me (sadly, this is not easy to achieve with cfg_attr and feature flags, because public-ness is not an attribute, so it's not really worth any effort, but nor is it worth keeping the API stable).

PR #3898, opened 2026-08-21
fix: Increase the socket TX buffer to hold a GSO batch

larseggert:
Because at least on macOS, it's 9KB and so doesn't by default.

Performance impact:

- no Apple fast, default TX and RX buffers:
  ``` thrpt:  [986.32 MiB/s 992.87 MiB/s 999.06 MiB/s] ```

- Apple fast, default TX and RX buffers:
  ``` thrpt:  [1.3364 GiB/s 1.3432 GiB/s 1.3495 GiB/s] ```

- Apple fast, default TX and 1MB RX buffers **(current neqo default)**:
  ``` thrpt:  [1.3294 GiB/s 1.3345 GiB/s 1.3396 GiB/s] ```

- Apple fast, 64KB TX and 1MB RX buffers:
  ``` thrpt:  [1.7714 GiB/s 1.7769 GiB/s 1.7823 GiB/s] ```

(Found while investigating Patricia's bug, which is in part also due to the too-small macOS default TX socket buffers.)


larseggert:
"GSO batch" = max. `sendmsg_x` payload with Internet MTU

neqo-udp/src/lib.rs
```diff
@@ -317,6 +366,21 @@ mod tests {
...
+            .state
+            .send_buffer_size((&socket.inner).into())
+            .expect("send buffer size to be readable");
+        let min = min_send_buf_size(&socket.state);
+        assert!(size >= min, "send buffer is {size}, want {min}");
+    }
+
+    #[test]
+    fn send_buffer_covers_one_batch() -> Result<(), io::Error> {
+        assert_send_buffer_covers_one_batch(&socket()?);
```
martinthomson: Is this test OS/platform/hardware dependent?

larseggert: It should be. We should size the TX buffer on all platforms so GSO bursts can fit.

neqo-udp/src/lib.rs
```diff
@@ -237,6 +240,53 @@ impl<'a> Iterator for DatagramIter<'a> {
     }
 }
 
+const fn min_send_buf_size(max_gso_segments: usize) -> usize {
+    max_gso_segments * 2048
```
mxinden, line 244: Why `2048`? I believe the effective max GSO size on Linux is 64 KB. Is that different on MacOS?

mxinden (commented):
> * no Apple fast, default TX and RX buffers:
>       `thrpt:  [986.32 MiB/s 992.87 MiB/s 999.06 MiB/s]`

I would guess 99.99% of all Firefox users do not have a 1GiB/s internet connection, nor do I think any of the remaining Firefox stack can saturate such a throughput. Thus I don't think this patch will have an impact.

In case you agree, are these performance optimizations worth the complexity they introduce?

neqo-udp/src/lib.rs
larseggert: 2048 is a standin for 1500.

The macOS issue is not whether we increase to 32KB, 64Kb, or 128KB, it;s that the default 9KB is smaller than what we send in 0-RTT sometimes with sendmsg_x, we hence drop part of it and never RTX due to the missing timer. The larger default reduces the likelihood, and setting the timer in neqo-glue recovers from when it still happens.

larseggert:
> > * no Apple fast, default TX and RX buffers:
> >   `thrpt:  [986.32 MiB/s 992.87 MiB/s 999.06 MiB/s]`
> 
> I would guess 99.99% of all Firefox users do not have a 1GiB/s internet connection, nor do I think any of the remaining Firefox stack can saturate such a throughput. Thus I don't think this patch will have an impact.

This is the download bench over loopback. You can also read the performance improvement as a CPU cycle reduction, which I think we **do** care about.

neqo-udp/src/lib.rs
mxinden: In regards to the ongoing issue on Patricia's machine, I don't think the above theory is correct.

Say Firefox were to send a batch of datagrams. Say that MacOS `sendmsg_x` would silently drop some of them, we would then already schedule a timer:

https://github.com/mozilla-firefox/firefox/blob/7b61047039c98a25b9ef5c6824089ab6c60bf54c/netwerk/socket/neqo_glue/src/lib.rs#L1464

I don't think scheduling more timers is the solution here. I believe we need to figure out why MacOS does not give Firefox a wake-up signal after having returned a `WouldBlock` on a send. See also https://github.com/mozilla/neqo/pull/3899#issuecomment-5367774363.

larseggert: See my comment here: https://github.com/mozilla/neqo/pull/3899#issuecomment-5368257969

I think there is a case where we *don't* set this timer. Or I am misunderstanding the code.

PR #3899, opened 2026-08-21
feat: Add `next_timeout` to the public API

larseggert:
So neqo users can query the next timeout without producing output.

### DIFF
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1163,7 +1163,7 @@ impl Connection {
         self.streams.cleanup_closed_streams();
     }
 
-    /// Get the time that we next need to be called back, relative to `now`.
+    /// Time until we next need a callback, `None` if never, zero if already due.
     fn next_delay(&mut self, now: Instant, paced: bool) -> Option<Duration> {
         qtrace!("[{self}] Get callback delay {now:?}");
 
@@ -1226,9 +1226,7 @@ impl Connection {
         // drive it.
 
         let earliest = delays.into_iter().min()?;
-        // TODO(agrover, mt) - need to analyze and fix #47
-        // rather than just clamping to zero here.
-        debug_assert!(earliest > now);
+        // Zero if already due, which only a caller that skipped `process_timer` sees.
         let delay = earliest.saturating_duration_since(now);
         qdebug!("[{self}] delay duration {delay:?}");
         self.hrtime.update(delay / 4);
@@ -1246,9 +1244,16 @@ impl Connection {
     }
 
     /// The minimum delay after which [`Self::process_multiple_output`] needs to be called again.
+    ///
+    /// Sending or receiving before then can shorten it.
+    ///
+    /// `None` means no callback needed, except for a not-yet-started client, which needs one now.
+    ///
+    /// Pacing is excluded. While the sender is pacer-blocked, [`Self::process_multiple_output`]
+    /// returns a shorter delay than this.
     #[must_use]
     pub fn next_timeout(&mut self, now: Instant) -> Option<Duration> {
-        // Not paced: waking early is harmless.
+        // `Pacer::next` can be in the past when it is not blocking, so leave it out.
         self.next_delay(now, false)
     }
 
@@ -1279,9 +1284,12 @@ impl Connection {
 
         match self.output(now, max_datagrams) {
             SendOptionBatch::Yes(dgram) => OutputBatch::DatagramBatch(dgram),
-            SendOptionBatch::No(paced) => self
-                .next_delay(now, paced)
-                .map_or(OutputBatch::None, OutputBatch::Callback),
+            SendOptionBatch::No(paced) => {
+                self.next_delay(now, paced).map_or(OutputBatch::None, |d| {
+                    debug_assert!(!d.is_zero());
+                    OutputBatch::Callback(d)
+                })
+            }
         }
     }
 
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1245,7 +1245,7 @@ impl Connection {
             .expect("max_datagrams is 1")
     }
 
-    /// The delay after which [`Self::process_multiple_output`] needs to be called again.
+    /// The minimum delay after which [`Self::process_multiple_output`] needs to be called again.
     #[must_use]
     pub fn next_timeout(&mut self, now: Instant) -> Option<Duration> {
         // Not paced: waking early is harmless.
--- a/neqo-transport/src/connection/mod.rs
+++ b/neqo-transport/src/connection/mod.rs
@@ -1164,13 +1164,17 @@ impl Connection {
     }
 
     /// Get the time that we next need to be called back, relative to `now`.
-    fn next_delay(&mut self, now: Instant, paced: bool) -> Duration {
+    fn next_delay(&mut self, now: Instant, paced: bool) -> Option<Duration> {
         qtrace!("[{self}] Get callback delay {now:?}");
+
+        if matches!(self.state, State::Init | State::Closed(_)) {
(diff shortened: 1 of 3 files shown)


neqo-transport/src/connection/mod.rs
```diff
@@ -1241,6 +1241,16 @@ impl Connection {
             .expect("max_datagrams is 1")
     }
 
+    /// The delay after which [`Self::process_multiple_output`] needs to be called again.
```
```suggestion
    /// The minimum delay after which [`Self::process_multiple_output`] needs to be called again.
```

If `process_input()` or any of the other API points that affect the connection is called, that time might move up, so this is a minimum.  Maybe you could even expand this to be clearer about what is going on.

```suggestion
    /// The delay after which [`Self::process_output`] or [`Self::process_multiple_output`]
    /// needs to be called, even if nothing else changes.
    ///
    /// This covers packet pacing, retransmission, and other maintenance, like idle timeouts.
    /// If data is sent (e.g., with [`Self::stream_send`] or [`Self::send_datagram`]) before then,
    /// this time will almost certainly reduce.
    /// If data is received (e.g., with [`Self::process_input`]) before then,
    /// this time might reduce.
```

martinthomson (approved):

martinthomson:
Your automated code reviews are getting confused about the point of recomputation of the delay.  Hopefully my comment tweak helps there, but I would disregard their comments on caching the output of `process_output()`.

I have no opinion on whether the returned value should be relative or absolute, but I think we know that relative is what is going to be used, so I'm inclined to disregard that comment too.

neqo-transport/src/connection/mod.rs
```diff
@@ -1221,14 +1225,12 @@ impl Connection {
         // timeout for it  It is expected that other activities will
         // drive it.
 
-        let earliest = delays.into_iter().min().expect("at least one delay");
-        // TODO(agrover, mt) - need to analyze and fix #47
-        // rather than just clamping to zero here.
-        debug_assert!(earliest > now);
```
mxinden, line 1227: This `debug_assert` has caught multiple regressions in the past. Why remove it?

neqo-transport/src/connection/mod.rs
```diff
@@ -1268,13 +1284,12 @@ impl Connection {
...
-                    OutputBatch::Callback(timeout.duration_since(now))
-                }
-                _ => OutputBatch::Callback(self.next_delay(now, paced)),
-            },
+            SendOptionBatch::No(paced) => {
+                self.next_delay(now, paced).map_or(OutputBatch::None, |d| {
+                    debug_assert!(!d.is_zero());
+                    OutputBatch::Callback(d)
+                })
+            }
```
mxinden, line 1292: Why this change in behavior on `State::Init`, `Closed` and `Closing`?

mxinden (commented):
What are you planning to do with the timer on the Firefox side?

neqo-transport/src/connection/mod.rs
larseggert: It moved to `debug_assert!(!d.is_zero());`

larseggert:
> What are you planning to do with the timer on the Firefox side?

I think Patricia's issue is coming from us not setting a timer when we buffer data in the `WouldBlock` case in neqo-glue. The change is roughly:
```diff
diff --git a/netwerk/socket/neqo_glue/src/lib.rs b/netwerk/socket/neqo_glue/src/lib.rs
index 9727eec11403..b5702009e60b 100644
--- a/netwerk/socket/neqo_glue/src/lib.rs
+++ b/netwerk/socket/neqo_glue/src/lib.rs
@@ -1400,6 +1417,11 @@ pub extern "C" fn neqo_http3conn_process_output_and_send(
                         if static_prefs::pref!("network.http.http3.pr_poll_write") {
                             qdebug!("Buffer outbound datagram to be sent once UDP socket has write-availability.");
                             conn.buffered_outbound_datagram = Some(dg);
+                            if let Some(ms) =
+                                conn.conn.next_timeout(Instant::now()).and_then(timer_ms)
+                            {
+                                set_timer_func(context, ms);
+                            }
                             return ProcessOutputAndSendResult {
                                 // Propagate WouldBlock error, thus indicating that
                                 // the UDP socket should be polled for
@@ -1407,10 +1429,9 @@ pub extern "C" fn neqo_http3conn_process_output_and_send(
                                 result: NS_BASE_STREAM_WOULD_BLOCK,
                                 bytes_written: bytes_written.try_into().unwrap_or(u32::MAX),
                             };
-                        } else {
-                            qwarn!("dropping datagram as socket would block");
-                            break;
                         }
+                        qwarn!("dropping datagram as socket would block");
+                        continue;
                     }
                     Err(e) if e.raw_os_error() == Some(libc::EIO) && dg.num_datagrams() > 1 => {
                         // GSO is unsupported on this send path. quinn-udp has now
@@ -1462,18 +1483,13 @@ pub extern "C" fn neqo_http3conn_process_output_and_send(
                 );
             }
             OutputBatch::Callback(to) => {
-                let timeout = if to.is_zero() {
-                    Duration::from_millis(1)
-                } else {
-                    to
-                };
-                let Ok(timeout) = u64::try_from(timeout.as_millis()) else {
+                let Some(ms) = timer_ms(to) else {
                     return ProcessOutputAndSendResult {
                         result: NS_ERROR_UNEXPECTED,
                         bytes_written: 0,
                     };
                 };
-                set_timer_func(context, timeout);
+                set_timer_func(context, ms);
                 break;
             }
             OutputBatch::None => {
```

neqo-transport/src/connection/mod.rs
larseggert: I don't think there is a functional change here, i.e., the results are the same as before.

mxinden:
Not setting a timer on `WouldBlock` is by design. Neqo timers operate on the order milliseconds. I expect `WouldBlock` wake-ups to operate on the order of microseconds. Thus, on `WouldBlock`, instead of scheduling a timer, I believe we should prioritize sending out the buffered data, i.e. `poll` the socket for write availability.

In the case of Patricia's bug, I don't think scheduling a timer is the right fix. We need to figure out why Firefox does not get a wake-up on the socket after it returned a `WouldBlock` on send.

neqo-transport/src/connection/mod.rs
mxinden: @larseggert `is_zero` is not the same guarantee as `earliest > now`, i.e. we should never return a timer in the past.

neqo-transport/src/connection/mod.rs
mxinden: You are right. Thanks. For the record, duplicated logic in `next_delay`:

https://github.com/mozilla/neqo/blob/40854d06fef597785417adc090c494cb518d2499/neqo-transport/src/connection/mod.rs#L1170-L1178

larseggert:
If I understand things correctly, while `buffered_outbound_datagram` is non-empty, the glue short-circuits `process_multiple_output`:

```rust
let output = conn.buffered_outbound_datagram.take()
    .map(OutputBatch::DatagramBatch)
    .unwrap_or_else(|| conn.conn.process_multiple_output(now, max_gso_segments));
```

So it isn't just that no timer is armed at the moment of `WouldBlock` — none can ever be armed afterwards, because the path that yields `Callback` is bypassed.

Or am I missing something?

mxinden:
We only set `buffered_outbound_datagram` on `WouldBlock`:

https://github.com/mozilla-firefox/firefox/blob/2916f047242de012d1aa0c572be9e5239644cc48/netwerk/socket/neqo_glue/src/lib.rs#L1398-L1402

Once there is write availability, our polling logic wakes up, `neqo_http3conn_process_output_and_send` is called the `buffered_outbound_datagram` is send. Once sent, we call `process_output` on Neqo again, either resulting in another datagram, or in a timer to be scheduled.

PR #3902, opened 2026-08-24
chore: Use `derive_more` to reduce boilerplate (again)

larseggert:
Respin of #3290. Uses `displaydoc` to derive `Display` rather than `derive_more::Display`, to avoid the Gecko-blocklisted dep issye that caused the revert in #3338.

Gecko needs a bump to `displaydoc@0.2.6`, which lands cleanly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,6 +361,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive_more"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "093242cf7570c207c83073cf82f79706fe7b8317e98620a47d5be7c3d8497678"
+dependencies = [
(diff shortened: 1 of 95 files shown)


Cargo.lock
```diff
@@ -370,6 +390,17 @@ dependencies = [
  "crypto-common",
 ]
 
+[[package]]
+name = "displaydoc"
+version = "0.2.6"
```
mxinden, line 395: Firefox uses `0.2.4`:

https://github.com/mozilla-firefox/firefox/blob/dddd3ebf8c01bcbf38c5ba24812e9df1ea6fff8a/Cargo.lock#L1891-L1893

larseggert: Yes, but this needs 0.2.6. I checked that Fx can upgrade to 0.2.6 cleanly. (Earlier versions trip clippy.)

PR #3905, opened 2026-08-27
refactor: Extract shared event queue for transport and http3

mxinden:
The event queues in `neqo-transport` and `neqo-http3` each wrapped their own `Rc<RefCell<VecDeque<_>>>` and both named the append method `insert`, though transport's dropped duplicates and http3's did not.

Move the wrapper into `neqo-common` as `event::Queue<T>`, offering `push` to append and `push_unique` to append only if no equal event is queued. No behavior change.

Noticed while working on https://github.com/mozilla/neqo/pull/3859.

### DIFF
--- a/neqo-common/src/event.rs
+++ b/neqo-common/src/event.rs
@@ -71,10 +71,10 @@ impl<T> Queue<T> {
         self.events.borrow_mut().clear();
     }
 
-    /// Determine whether there are pending events.
+    /// Determine whether the queue is empty.
     #[must_use]
-    pub fn has_events(&self) -> bool {
-        !self.events.borrow().is_empty()
+    pub fn is_empty(&self) -> bool {
+        self.events.borrow().is_empty()
     }
 
     /// Take the first event.
@@ -185,13 +185,13 @@ mod tests {
         q.push(2);
         q.push(3);
         q.remove_matching(|e| *e == 2);
-        assert!(q.has_events());
+        assert!(!q.is_empty());
         assert_eq!(q.next_event(), Some(1));
         assert_eq!(q.next_event(), Some(3));
         assert_eq!(q.next_event(), None);
         q.push(4);
         q.clear();
-        assert!(!q.has_events());
+        assert!(q.is_empty());
     }
 
     #[test]
--- a/neqo-common/src/event.rs
+++ b/neqo-common/src/event.rs
@@ -4,6 +4,92 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use std::{cell::RefCell, collections::VecDeque, rc::Rc};
+
+/// A shared FIFO of events. Clones share the same underlying queue.
+#[derive(Debug)]
+pub struct Queue<T> {
+    events: Rc<RefCell<VecDeque<T>>>,
+}
+
+impl<T> Default for Queue<T> {
+    fn default() -> Self {
+        Self {
+            events: Rc::default(),
+        }
+    }
+}
+
+impl<T> Clone for Queue<T> {
+    fn clone(&self) -> Self {
+        Self {
+            events: Rc::clone(&self.events),
+        }
+    }
+}
+
+impl<T> Queue<T> {
+    /// Append `event`, regardless of what is already queued.
+    pub fn push(&self, event: T) {
+        self.events.borrow_mut().push_back(event);
+    }
+
+    /// Append `event` unless an equal event is already queued.
+    /// An existing equal event keeps its place in the queue.
+    pub fn push_unique(&self, event: T)
+    where
+        T: PartialEq,
+    {
+        let mut q = self.events.borrow_mut();
+        if !q.contains(&event) {
+            q.push_back(event);
+        }
+    }
+
+    /// Append `event` unless a queued event matches `is_duplicate`.
+    /// An existing match keeps its place in the queue.
+    pub fn push_unique_by<F>(&self, event: T, is_duplicate: F)
+    where
+        F: Fn(&T) -> bool,
+    {
+        let mut q = self.events.borrow_mut();
+        if !q.iter().any(is_duplicate) {
+            q.push_back(event);
+        }
+    }
+
+    /// Remove all queued events matching `f`.
+    pub fn remove_matching<F>(&self, f: F)
+    where
+        F: Fn(&T) -> bool,
+    {
+        self.events.borrow_mut().retain(|event| !f(event));
+    }
+
+    /// Remove all queued events.
+    pub fn clear(&self) {
+        self.events.borrow_mut().clear();
+    }
+
+    /// Determine whether there are pending events.
+    #[must_use]
+    pub fn has_events(&self) -> bool {
+        !self.events.borrow().is_empty()
+    }
+
+    /// Take the first event.
+    #[must_use]
+    pub fn next_event(&self) -> Option<T> {
+        self.events.borrow_mut().pop_front()
+    }
+
+    /// Take all events.
+    #[must_use]
+    pub fn take_all(&self) -> VecDeque<T> {
+        self.events.take()
+    }
+}
+
 /// An event provider is able to generate a stream of events.
 pub trait Provider {
     type Event;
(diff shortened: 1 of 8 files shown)


larseggert:
Is there maybe even an external crate we could use?

mxinden:
> Is there maybe even an external crate we could use?

What would be the benefit of an external crate @larseggert?

larseggert:
> What would be the benefit of an external crate @larseggert?

Core reduction. But I made this comment before looking at the actual PR, which is not that large :-)

neqo-common/src/event.rs
```diff
@@ -4,6 +4,92 @@
...
+    /// Remove all queued events.
+    pub fn clear(&self) {
+        self.events.borrow_mut().clear();
+    }
+
+    /// Determine whether there are pending events.
+    #[must_use]
+    pub fn has_events(&self) -> bool {
+        !self.events.borrow().is_empty()
+    }
```
larseggert, line 78: Would it be slightly more idiomatic to name this `is_empty`? (Caller need to change then though, and if we rename those too, maybe too much churn.)

larseggert (approved):

neqo-common/src/event.rs
martinthomson: Add an alias instead?  Then deprecate this variant later?

neqo-common/src/event.rs
```diff
@@ -4,6 +4,92 @@
...
+
+    /// Take all events.
+    #[must_use]
+    pub fn take_all(&self) -> VecDeque<T> {
+        self.events.take()
+    }
+}
+
 /// An event provider is able to generate a stream of events.
 pub trait Provider {
```
martinthomson, line 94: I suspect that you could simplify this trait after this change.  This design allows for a lot more flexibility, but if you are doing to use `Queue<T>` always, you might imagine that this turns into an implementation of `AsMut<Queue<T>>` or something like that (sadly, it's not a straight replacement).

At a minimum, you could add default implementations if you added `fn queue(&self) -> Queue<Self::Event>`.

neqo-http3/src/client_events.rs
```diff
@@ -4,9 +4,11 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::{cell::RefCell, collections::VecDeque, rc::Rc};
-
-use neqo_common::{Bytes, Header, event::Provider as EventProvider, qtrace};
+use neqo_common::{
+    Bytes, Header,
+    event::{Provider as EventProvider, Queue},
```
martinthomson:
```suggestion
    event::{Provider as EventProvider, Queue as EventQueue},
```

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
+            .push_unique(ConnectionEvent::ResumptionToken(token));
     }
 
     pub fn client_0rtt_rejected(&self) {
         // If 0rtt rejected, must start over and existing events are no longer
         // relevant.
-        self.events.borrow_mut().clear();
-        self.insert(ConnectionEvent::ZeroRttRejected);
+        self.events.clear();
+        self.events.push_unique(ConnectionEvent::ZeroRttRejected);
```
martinthomson: This can be:

```suggestion
        self.events.push(ConnectionEvent::ZeroRttRejected);
```

Because an empty queue isn't going to need a check.

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
             _ => (),
         }
-        self.insert(ConnectionEvent::StateChange(state));
+        self.events.push_unique(ConnectionEvent::StateChange(state));
     }
 
     pub fn client_resumption_token(&self, token: ResumptionToken) {
-        self.insert(ConnectionEvent::ResumptionToken(token));
+        self.events
+            .push_unique(ConnectionEvent::ResumptionToken(token));
```
martinthomson: I think that you want `push()` here instead.  There can be multiple tokens.  Is this potentially a case where we are dropping tokens?

neqo-common/src/event.rs
mxinden: > Add an alias instead?

That would effectively be dead-code. This is only used in github.com/mozilla/neqo.

> Then deprecate this variant later?

Given that this is only used in github.com/mozilla/neqo, why the two step removal?

neqo-transport/src/events.rs
```diff
@@ -192,53 +214,25 @@ impl ConnectionEvents {
...
-        }
-    }
-
-    fn remove<F>(&self, f: F)
-    where
-        F: Fn(&ConnectionEvent) -> bool,
-    {
-        self.events.borrow_mut().retain(|evt| !f(evt));
+        self.events
+            .push_unique(ConnectionEvent::PathMigrated { local, remote });
```
martinthomson: This can probably use `push()` as well.

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
+        self.events.remove_matching(|evt| {
             matches!(evt,
                 ConnectionEvent::SendStreamWritable { stream_id: x } |
                 ConnectionEvent::SendStreamStopSending { stream_id: x, .. }
                 if *x == stream_id)
         });
 
-        self.insert(ConnectionEvent::SendStreamComplete { stream_id });
+        self.events
+            .push_unique(ConnectionEvent::SendStreamComplete { stream_id });
```
martinthomson: If you maybe added the ...Complete event to the remove_matching call above, you could just push without doing a second full scan.

neqo-common/src/event.rs
mxinden: Closing here with https://github.com/mozilla/neqo/pull/3905/commits/571bb02822a13a453b5afbc5ad3b3f738febff88. If you feel strongly, I am happy to do a follow-up.

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
     pub fn ech_fallback_authentication_needed(&self, public_name: String) {
-        self.insert(ConnectionEvent::EchFallbackAuthenticationNeeded { public_name });
+        self.events
+            .push_unique(ConnectionEvent::EchFallbackAuthenticationNeeded { public_name });
     }
 
     pub fn new_stream(&self, stream_id: StreamId) {
-        self.insert(ConnectionEvent::NewStream { stream_id });
+        self.events
+            .push_unique(ConnectionEvent::NewStream { stream_id });
```
martinthomson: This is never going to need a full scan, is it?


```suggestion
            .push(ConnectionEvent::NewStream { stream_id });
```

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
     pub fn authentication_needed(&self) {
-        self.insert(ConnectionEvent::AuthenticationNeeded);
+        self.events
+            .push_unique(ConnectionEvent::AuthenticationNeeded);
     }
 
     pub fn ech_fallback_authentication_needed(&self, public_name: String) {
-        self.insert(ConnectionEvent::EchFallbackAuthenticationNeeded { public_name });
+        self.events
+            .push_unique(ConnectionEvent::EchFallbackAuthenticationNeeded { public_name });
```
martinthomson, line 103: A total one-off:

```suggestion
            .push(ConnectionEvent::EchFallbackAuthenticationNeeded { public_name });
```

neqo-transport/src/events.rs
```diff
@@ -90,99 +90,121 @@ pub enum ConnectionEvent {
...
 pub struct ConnectionEvents {
-    events: Rc<RefCell<VecDeque<ConnectionEvent>>>,
+    events: Queue<ConnectionEvent>,
 }
 
 impl ConnectionEvents {
     pub fn authentication_needed(&self) {
-        self.insert(ConnectionEvent::AuthenticationNeeded);
+        self.events
+            .push_unique(ConnectionEvent::AuthenticationNeeded);
```
martinthomson:
```suggestion
            .push(ConnectionEvent::AuthenticationNeeded);
```

martinthomson (approved):
The design looks sound, but here are a few things to chew on.

neqo-common/src/event.rs
mxinden: I am not sure how this is a straight forward improvement, in other words whether it is worth the complexity.

Either way, I don't think we should do this as part of this pull request.

neqo-transport/src/events.rs
mxinden: I left this one as push_unique, because I do not think it is a one-off today.

The event does not mean "ECH failed", which would indeed happen once. It means "authentication of the public name is pending", and that state is recomputed on every `handshake` call.

Will proceed here without. Happy to do a follow-up pull request.

Issue #2781, opened 2025-07-05
Review uses of `SmallVec`


larseggert:
Based on the experience with #2779, we should review the existing uses of `SmallVec` to confirm they actually improve performance. I think most or all of them predate us doing proper performance testing.

tiya-9975:
Thanks @larseggert for opening this issue.

I’m new to neqo and would like to help with this review.  
To make sure I approach this correctly, is the expected first step to
identify and document the current `SmallVec` usage sites and the rationale
behind them, before attempting any benchmarking or refactoring?

Please let me know if there are specific modules or prior discussions I
should start with.

larseggert:
> To make sure I approach this correctly, is the expected first step to
> identify and document the current `SmallVec` usage sites and the rationale
> behind them, before attempting any benchmarking or refactoring?

Thanks for your interest in helping out! Yes, that would be a good start.

TarunVishwakarma1:
Is this still open?

I like to pick this.

larseggert:
Yes.

TarunVishwakarma1:
Started working through these. Opened #3906 for tracking.rs:454 — the collected SmallVec is into_boxed_slice()d two lines later, so the inline capacity doesn't avoid an allocation.

Also looked at the three uses in send_stream.rs (221, 318, 398). Those are local scratch in RangeTracker and TxBuffer, neither of which derives Clone, so they don't have the shape that caused the #2779 regression. I'd leave those as-is.

Will work through the remaining sites one at a time.

TarunVishwakarma1:
#3906 merged. ```tracking.rs:454``` turned out performance-neutral, so no win there, but one fewer ```SmallVec```that wasn't earning its inline capacity. Working through the remaining sites.

TarunVishwakarma1:
Opened #3911 for ```addr_valid.rs``` — ```NewTokenState``` was 312 bytes vs 48 with plain ```Vec```, and it's a ```Connection``` field rather than scratch, so the ```#[expect(clippy::large_enum_variant)]``` can come out too.

Also checked ```send_stream.rs``` (221, 318, 398) — local scratch in non-```Clone``` types, I'd leave those. Still to look at: ```recv_stream.rs```, ```connection/mod.rs```, ```qlog.rs```, ```cid.rs```

PR #3906, opened 2026-08-30
chore: use Vec instead of SmallVec in RecvdPackets::write_frame

TarunVishwakarma1:
Per #2781.

The `SmallVec` here is collected and then `into_boxed_slice()`d two lines
later, so the inline capacity never actually avoids an allocation — it just
adds a copy pass, plus ~776 bytes of stack frame on every ACK write
(`MAX_TRACKED_RANGES` is 32, `PacketRange` is 24 bytes).

I couldn't resolve this locally. Running `transfer_simulated` A/B/A/B, two
runs of the *identical* build gave a ±20% confidence interval (p = 0.57 and
0.81), so the local noise floor is far above any effect this change could
plausibly have.

Opening as a draft to get CI's instruction-count numbers. Happy to close it
if they show a regression.

Also checked the three `SmallVec` uses in `send_stream.rs` (lines 221, 318,
398) while looking at this issue. Those are local scratch in `RangeTracker`
and `TxBuffer`, neither of which derives `Clone`, so they don't have the
shape that caused the #2779 regression. I'd leave those alone.

### DIFF
--- a/neqo-transport/src/tracking.rs
+++ b/neqo-transport/src/tracking.rs
@@ -18,7 +18,6 @@ use enumset::{EnumSet, EnumSetType};
 use log::{Level, log_enabled};
 use neqo_common::{Buffer, Ecn, MAX_VARINT, qdebug, qtrace, qwarn, to_u64};
 use nss::Epoch;
-use smallvec::SmallVec;
 use strum::{Display, EnumIter};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


TarunVishwakarma1:
Working through these one at a time. tracking.rs:454 in #3906, send_stream.rs looks fine as-is. Will pick up the others once CI numbers are in.

TarunVishwakarma1:
Hi Maintainers, 

Please approve the workflow as this needs workflow approval to get CI benchmark numbers, which is the point of the PR.

larseggert (approved):
Performance neutral, but still a nice little cleanup. Thanks!

TarunVishwakarma1:
CodSpeed confirms performance-neutral. Thanks for the quick review!

PR #3908, opened 2026-08-31
fix: Narrow `AckRate` and `RecoveryToken`

larseggert:
`AckRate` 24 -> 8 bytes
`RecoveryToken` 32 -> 24 bytes

Prep work for reducing `Token`.

### DIFF
--- a/neqo-transport/src/ackrate.rs
+++ b/neqo-transport/src/ackrate.rs
@@ -8,7 +8,7 @@
 
 use std::{cmp::max, time::Duration};
 
-use neqo_common::{Buffer, qtrace, to_u64};
+use neqo_common::{Buffer, qtrace};
 
 use crate::{
     connection::params::ConnectionParameters,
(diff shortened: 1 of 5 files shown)


neqo-transport/src/ackrate.rs
```diff
@@ -19,15 +19,19 @@ use crate::{
...
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub struct AckRate {
     /// The maximum number of packets that can be received without sending an ACK.
-    packets: usize,
-    /// The maximum delay before sending an ACK.
-    delay: Duration,
+    packets: u16,
+    /// The maximum delay before sending an ACK, in microseconds.
+    delay: u32,
```
martinthomson, line 27: 2^32 microseconds is more than 4000 seconds.  Do we really need a u32 for this, or could we just say that 64 milliseconds (u16 in us) is the largest delay we tolerate?

neqo-transport/src/send_stream.rs
```diff
@@ -2416,7 +2416,7 @@ impl<'a> IntoIterator for &'a mut SendStreams {
 pub struct RecoveryToken {
     id: StreamId,
     offset: u64,
-    length: usize,
+    length: u32, // `u32`, because a STREAM frame cannot be larger than an IP datagram
```
martinthomson, line 2419: Can't this be u16 instead?

martinthomson (approved):
You could go even smaller, I think.

neqo-transport/src/send_stream.rs
larseggert: It can, but Rust seems to align fields to eight bytes, so it doesn't matter.

martinthomson: If you don't need it to be larger, then why not?  Also, doesn't that make it possible to use `usize::from()` rather than our weird cast method?  That's a win.

larseggert: Actually, it can't be narrower, since it needs to store the peer's max_ack_delay.

martinthomson: That's not THIS, but the other one, right?

Besides, for max_ack_delay:

> Values of 2^14 or greater are invalid.

-- https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.28.1

martinthomson: We could save a value in microseconds, but it seems reasonable to have this be in milliseconds.  Or are we concerned about https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency-14#section-4-4.6.1 ?

PR #3910, opened 2026-08-31
fix: Box `ConnectionIdEntry`

larseggert:
To avoid large enum variant.

### DIFF
--- a/neqo-transport/src/cid.rs
+++ b/neqo-transport/src/cid.rs
@@ -589,7 +589,7 @@ impl ConnectionIdManager {
 
         while let Some(entry) = self.lost_new_connection_id.pop() {
             if entry.write(builder, stats) {
-                tokens.push(recovery::Token::NewConnectionId(entry));
+                tokens.push(recovery::Token::NewConnectionId(Box::new(entry)));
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
Where does this leave us?  Are we below 16 bytes yet?

mxinden (commented):
Before merging here can you please:

1. Include a reasoning why dynamic dispatch is conceptually more performant here than the current stack allocation? Are the other enum variants much smaller than this one? Does it exceed the standard 64 byte cache line size? Are the concerns in https://github.com/mozilla/neqo/pull/3910/changes#r3893742569 relevant?
2. Include some proof that this is effectively an optimization, either through a Firefox profile, or a benchmark here.

larseggert:
> Where does this leave us? Are we below 16 bytes yet?

24 bytes. Rust seems to want to pad to 8 bytes and there are some variants that need us to store the variant separately (no niche). Wonder if there is a way to force Rust to binpack better.

larseggert:
@mxinden I had hoped the benches would provide the justification.

There should also be further synergies with #3908 and #3907.

PR #3916, opened 2026-09-01
fix(http3): allow WebTransport when the peer lacks reliable reset 

mxinden:
Since #3756, neqo makes the peer's reset_stream_at transport parameter (reliable reset) a requirement for WebTransport, following the WebTransport draft (draft-ietf-webtrans-http3, Section 4.4), which resets a WebTransport data stream with RESET_STREAM_AT.

reset_stream_at is a very new QUIC extension that few deployed servers implement yet, so requiring it breaks WebTransport with most current servers, including the WebTransport web platform test server. It failed every WebTransport web platform test in Firefox and backed out the neqo v0.31.0 uplift ([Bug 2065405](https://bugzilla.mozilla.org/show_bug.cgi?id=2065405)). The stream layer already handles peers without reliable reset: WebTransportSendStream falls back to a plain RESET_STREAM.

Stop requiring reliable reset from the peer. neqo still advertises reset_stream_at and uses RESET_STREAM_AT when the peer supports it.

Note this targets the `v0.31` branch and bumps the patch version to `v0.31.1`. Once merged, I will bring the fix into `main`.

### DIFF
--- a/neqo-http3/src/features/extended_connect/mod.rs
+++ b/neqo-http3/src/features/extended_connect/mod.rs
@@ -17,7 +17,7 @@ mod tests;
 
 use std::{cell::RefCell, fmt::Debug, mem, rc::Rc};
 
-use neqo_common::{Bytes, Header, Role};
+use neqo_common::{Bytes, Header, Role, qdebug};
 use neqo_transport::StreamId;
 
 use crate::{
@@ -97,10 +97,6 @@ impl TransportPrerequisites {
             reliable_reset,
         }
     }
-
-    const fn all(&self) -> bool {
-        self.datagrams && self.reliable_reset
-    }
 }
 
 #[derive(Debug)]
@@ -130,12 +126,11 @@ impl ExtendedConnectFeature {
         settings: &HSettings,
         transport_prereqs: &TransportPrerequisites,
     ) {
-        // A WebTransport client must confirm the server supports everything WebTransport needs:
-        // extended CONNECT, HTTP/3 datagrams (SETTINGS), and the datagram/reliable-reset
-        // transport parameters.
+        // reset_stream_at is also required (draft Section 4.4), but too few servers support it
+        // yet, so we don't gate on it for now (falls back to RESET_STREAM). See #3917.
         let conditions_met = match self.connect_type {
             ExtendedConnectType::WebTransport => {
-                transport_prereqs.all()
+                transport_prereqs.datagrams
                     && settings.get(HSettingType::EnableH3Datagram) == 1
                     && (self.role == Role::Server
                         || (settings.get(HSettingType::EnableConnect) == 1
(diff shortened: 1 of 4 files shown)


neqo-http3/src/features/mod.rs
```diff
@@ -216,6 +216,27 @@ mod tests {
         ));
     }
 
+    /// WebTransport is negotiable even when the peer does not advertise the `reset_stream_at`
+    /// transport parameter. The WebTransport draft (Section 4.4) requires reliable reset, but
```
larseggert: Should we add a tests that confirms WT also enables WITH reliable resets?

larseggert (approved):
I guess the clippy failure is due to the 0.31 branch target.

neqo-http3/src/features/mod.rs
mxinden: :+1: done

neqo-http3/src/features/mod.rs
```diff
@@ -216,6 +216,43 @@ mod tests {
...
+        ]);
+        let mut f =
+            ExtendedConnectFeature::new(ExtendedConnectType::WebTransport, Role::Client, true);
+        // The peer advertises HTTP/3 datagrams but not reliable reset (reset_stream_at).
+        f.handle_settings(&full, &TransportPrerequisites::new(true, false));
+        assert!(
+            f.enabled(),
+            "WebTransport must not require the peer to support reliable reset"
+        );
+    }
```
larseggert: Could be one test with a loop over `[false, true]`.

neqo-http3/src/features/extended_connect/mod.rs
```diff
@@ -130,12 +126,14 @@ impl ExtendedConnectFeature {
...
         transport_prereqs: &TransportPrerequisites,
     ) {
-        // A WebTransport client must confirm the server supports everything WebTransport needs:
-        // extended CONNECT, HTTP/3 datagrams (SETTINGS), and the datagram/reliable-reset
-        // transport parameters.
+        // A WebTransport client must confirm the server supports extended CONNECT, HTTP/3
+        // datagrams (SETTINGS), and the QUIC datagram transport parameter. Reliable stream
+        // reset (reset_stream_at, required by the draft's Section 4.4) is not gated here: the
+        // stream layer falls back to a plain RESET_STREAM for peers without it, and requiring
+        // it would break WebTransport with most current servers.
```
martinthomson: I think that you want a different comment here.  This is a temporary condition only and we need an issue to track the change we'll eventually need to make in order to have this fixed.

Also, we should probably put work in to ensure that aioquic supports this in its webtransport stack, so that we don't get too badly stuck.

mxinden: Rephrased and tracked in https://github.com/mozilla/neqo/issues/3917.

PR #3927, opened 2026-09-02
chore: Disable Rust "security" updates in dependabot

larseggert:
We're stuck with the Gecko versions anyway.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,20 +3,17 @@
 
 version: 2
 updates:
-  # We don't really use Dependabot for Rust dependencies, since we are tracking Gecko,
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,20 +3,17 @@
 
 version: 2
 updates:
-  # We don't really use Dependabot for Rust dependencies, since we are tracking Gecko,
-  # but we do want to be notified of security vulnerabilities.
```
mxinden, line 7: Do we have someone on the Mozilla side that monitors these instead @larseggert?

larseggert: `cargo deny` in Gecko CI, and then usually someone picks it up.

PR #3940, opened 2026-09-04
Reject 0-RTT downgrade of the extended CONNECT setting

srkyn:
On a resumed connection, Neqo currently accepts a new SETTINGS frame that omits `SETTINGS_ENABLE_CONNECT_PROTOCOL` even when the stored 0-RTT value was 1. The feature negotiation then disables extended CONNECT, but the connection remains open.

RFC 9114 Section 7.2.4.2 requires `H3_SETTINGS_ERROR` when a server accepts 0-RTT and then omits an understood setting that previously had a non-default value. The setting's default is 0, so omission after a stored value of 1 is covered by that rule.

This adds `EnableConnect` to the existing 0-RTT settings comparison and includes a regression test for the omitted-setting case.

Validation:

- `cargo test --locked --release -p neqo-http3`
- `cargo check --locked -p neqo-http3`
- `cargo fmt --all -- --check`
- `cargo clippy --locked -p neqo-http3 --all-targets -- -D warnings`

### DIFF
--- a/neqo-http3/src/connection.rs
+++ b/neqo-http3/src/connection.rs
@@ -1699,6 +1699,8 @@ impl Http3Connection {
                     HSettingType::MaxHeaderListSize,
                     HSettingType::MaxTableCapacity,
                     HSettingType::BlockedStreams,
+                    // RFC 9114, Section 7.2.4.2 covers omitted non-default settings.
+                    HSettingType::EnableConnect,
                     // [RFC 9297, Section 2.1.1](https://www.rfc-editor.org/rfc/rfc9297.html#section-2.1.1)
                     // requires a client that stored SETTINGS_H3_DATAGRAM with its
                     // 0-RTT state to terminate the connection with H3_SETTINGS_ERROR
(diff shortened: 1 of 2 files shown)


neqo-http3/src/connection_client.rs
```diff
@@ -4276,6 +4276,26 @@ mod tests {
         );
     }
 
+    #[test]
+    fn zero_rtt_new_server_setting_connect_smaller() {
```
martinthomson:
```suggestion
    fn zero_rtt_extended_connect_disabled() {
```

neqo-http3/src/connection.rs
```diff
@@ -1699,6 +1699,8 @@ impl Http3Connection {
                     HSettingType::MaxHeaderListSize,
                     HSettingType::MaxTableCapacity,
                     HSettingType::BlockedStreams,
+                    // RFC 9114, Section 7.2.4.2 covers omitted non-default settings.
```
```suggestion
```

LLMs have this habit of fixating on the reason for their change, but this comment applies to the entire list.

martinthomson (approved):

neqo-http3/src/connection_client.rs
srkyn: Renamed to `zero_rtt_extended_connect_disabled` in 18a4bc8.

neqo-http3/src/connection.rs
srkyn: Removed the item-specific comment in 18a4bc8.

larseggert:
Thank you!

PR #3948, opened 2026-09-07
perf: don't allocate in UDP send path

larseggert:
Fill a caller-owned buffer, reused across calls. The batch borrows that buffer, so neither it nor the send path is generic over the buffer type.

vs #2747: An owning batch pulls two buffer types into the output path, and each hand-down adds a `&mut`, so three functions become six copies — the code bloat that regressed `streams*`. Borrowing drops it to one, non-generic, and removes `available()`, `copy_into` and `copy_output_into`. Net −91 lines. Batch size is now bounded by `max_datagrams`, not buffer capacity.


martinthomson:
> transfer/1-conn/1-100mb-req (aka. Upload)/mtu-1504: 💔 Performance has regressed by +67.852%.

😱

larseggert:
Yeah, the mix of regressions and improvements is odd

PR #3955, opened 2026-09-08
chore: Use `EnumMap` more

larseggert:
Makes the code here a bit cleaner.

### DIFF
--- a/neqo-transport/src/connection/params.rs
+++ b/neqo-transport/src/connection/params.rs
@@ -6,6 +6,7 @@
 
 use std::{cmp::max, num::NonZeroUsize, time::Duration};
 
+use enum_map::{Enum, EnumMap};
 use neqo_common::to_u64;
 
 pub use crate::recovery::FAST_PTO_SCALE;
@@ -15,7 +16,7 @@ use crate::{
     rtt::GRANULARITY,
     stream_id::StreamType,
     tparams::{
-        PreferredAddress, TransportParameter,
+        PreferredAddress, TransportParameter, TransportParameterId,
         TransportParameterId::{
             ActiveConnectionIdLimit, DisableMigration, GreaseQuicBit, IdleTimeout, InitialMaxData,
             InitialMaxStreamDataBidiLocal, InitialMaxStreamDataBidiRemote, InitialMaxStreamDataUni,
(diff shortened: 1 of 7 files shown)


neqo-transport/src/connection/params.rs
```diff
@@ -101,6 +102,28 @@ pub enum PreferredAddressConfig {
...
 
+/// A stream class with an independently configurable initial receive flow-control limit.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Enum)]
+pub enum StreamDataLimit {
+    /// Receiving data on bidirectional streams that this endpoint creates.
+    BiDiLocal,
+    /// Receiving data on bidirectional streams that the peer creates.
+    BiDiRemote,
+    /// Receiving data on unidirectional streams that the peer creates.
+    UniDi,
```
martinthomson, line 113:
```suggestion
    BidiLocal,
    /// Receiving data on bidirectional streams that the peer creates.
    BidiRemote,
    /// Receiving data on unidirectional streams that the peer creates.
    Unidi,
```

Isn't this the correct CamelCasing?

neqo-transport/src/connection/params.rs
```diff
@@ -113,19 +136,10 @@ pub struct ConnectionParameters {
...
-    max_stream_data_bidi_local: u64,
-    /// Initial flow control limit for receiving data on unidirectional streams that the peer
-    /// creates.
-    max_stream_data_uni: u64,
-    /// Initial limit on bidirectional streams that the peer creates.
-    max_streams_bidi: u64,
-    /// Initial limit on unidirectional streams that this endpoint creates.
-    max_streams_uni: u64,
+    /// Initial flow control limits for receiving stream data.
+    max_stream_data: EnumMap<StreamDataLimit, u64>,
```
martinthomson, line 140: 🧑‍🍳🤌

neqo-transport/src/connection/params.rs
```diff
@@ -176,11 +190,9 @@ impl Default for ConnectionParameters {
...
             hystart_css_baseline: HyStartCssBaseline::CurrentRoundMinRtt,
             max_data: INITIAL_LOCAL_MAX_DATA,
-            max_stream_data_bidi_remote: to_u64(INITIAL_LOCAL_MAX_STREAM_DATA),
-            max_stream_data_bidi_local: to_u64(INITIAL_LOCAL_MAX_STREAM_DATA),
-            max_stream_data_uni: to_u64(INITIAL_LOCAL_MAX_STREAM_DATA),
-            max_streams_bidi: LOCAL_STREAM_LIMIT_BIDI,
-            max_streams_uni: LOCAL_STREAM_LIMIT_UNI,
+            max_stream_data: EnumMap::from_array([to_u64(INITIAL_LOCAL_MAX_STREAM_DATA); 3]),
+            // `StreamType` variant order: `BiDi`, `UniDi`.
+            max_streams: EnumMap::from_array([LOCAL_STREAM_LIMIT_BIDI, LOCAL_STREAM_LIMIT_UNI]),
```
martinthomson, line 195: Seems brittle.  Why not use `enum_map! { StreamType::Bidi => LOCAL_STREAM_LIMIT_BIDI, StreamType::Unidi => LOCAL_STREAM_LIMIT_UNI }`?  It's more typing, but clearer (and no need for a comment).

martinthomson (approved):
Much nicer.

neqo-transport/src/connection/params.rs
larseggert: Yes, but we already use `StreamType::BiDi/UniDi` elsewhere.

neqo-transport/src/connection/params.rs
larseggert: `enum_map!` macro's expansion contains an unconditional `mem::forget()` call on a `Drop`-bearing guard type, which trips `clippy::mem_forget`. I didn't want to add a suppression. `enum-map@3.0.0` fixes this, but Gecko pins the older version.

larseggert: Looks like we can bump it in neqo and nss-rs...

larseggert: Not easily. It would add `syn@3` as a duplicate version in Gecko (or we need to manually derive the `Enum` trait everywhere.

PR #3958, opened 2026-09-09
fix: Be a bit smarter with invalidating `first_unmarked`

larseggert:


neqo-transport/src/send_stream.rs
```diff
@@ -137,25 +137,44 @@ impl RangeTracker {
...
+                return (prev_end, Some(cur_off - prev_end));
             }
         }
-        self.first_unmarked = Some((prev_end, None));
         (prev_end, None)
     }
 
+    /// Drop the `first_unmarked` cache if it satisfies `pred`.
+    fn invalidate_cache_if(&mut self, pred: impl FnOnce((u64, Option<u64>)) -> bool) {
+        if self.first_unmarked.is_some_and(pred) {
```
martinthomson: I think that you want `Option::take_if` here.

neqo-transport/src/send_stream.rs
```diff
@@ -137,25 +137,44 @@ impl RangeTracker {
...
+        self.first_unmarked = Some(res);
+        res
+    }
 
+    /// Scan `used` for the first unmarked range, ignoring the cache.
+    fn scan_first_unmarked_range(&self) -> (u64, Option<u64>) {
+        let mut prev_end = self.acked;
         for (&cur_off, &(cur_len, _)) in &self.used {
             if prev_end == cur_off {
                 prev_end = cur_off + cur_len;
```
```suggestion
                prev_end += cur_len;
```

neqo-transport/src/send_stream.rs
```diff
@@ -137,25 +137,44 @@ impl RangeTracker {
...
+        let res = self.scan_first_unmarked_range();
+        self.first_unmarked = Some(res);
+        res
+    }
 
+    /// Scan `used` for the first unmarked range, ignoring the cache.
+    fn scan_first_unmarked_range(&self) -> (u64, Option<u64>) {
+        let mut prev_end = self.acked;
         for (&cur_off, &(cur_len, _)) in &self.used {
             if prev_end == cur_off {
```
martinthomson, line 165: There's an invariant at play here that might be worth a `debug_assert!()`.  That invariant is that `self.used.first().is_none_or(|(&start, _)| self.acked <= start)`.  Or, `debug_assert!(cur_off >= prev_end)`.  We sort of get that with the underflow detection in `cur_off - prev_end` below, but it's not that obvious.

neqo-transport/src/send_stream.rs
```diff
@@ -391,7 +423,8 @@ impl RangeTracker {
             return;
         }
 
-        self.first_unmarked = None;
+        // A gap widens whenever the unmark starts at or before its end.
+        self.invalidate_cache_if(|(c_off, c_len)| off <= c_off + c_len.unwrap_or(0));
```
martinthomson: Is this a case where you can instead update the cached value?

If there is no cache, then you leave it alone.

If there is a cache, and
* this range ends before it starts, swap to `(off, Some(len))`,
* this range starts before it and ends at after the exist, swap to `(off, Some(len))`,
* this range overlaps with the start of it, swap to `(off, first_unmarked.1.map(|flen| flen + (first_unmarked.0 - off)))`,
* this range overlaps with the end of it, swap to `(fu.0, Some(len + (off - fu.0)))`,
* otherwise, leave it be.

This is one case where you know exactly what to do.  The cases where you are removing an unmarked range need a scan to rebuild it, but this one never does.

martinthomson (commented):
Looks solid, but I think that you can be a bit more aggressive about updating the cache.

neqo-transport/src/send_stream.rs
larseggert: Oh, yes!

neqo-transport/src/send_stream.rs
larseggert: Thanks for this - this led to diagnosing a preexisting issue (#3962).

neqo-transport/src/send_stream.rs
```diff
@@ -477,6 +488,20 @@ impl RangeTracker {
         if let Some((new_cur_off, new_cur_len, cur_state)) = to_add {
             self.used.insert(new_cur_off, (new_cur_len, cur_state));
         }
+
+        // Unmarking only grows the unmarked set, so the first unmarked range can
+        // only move down, to `unmarked`. Everything below that is untouched and
+        // so still contiguous, which makes the next range above the new start the
+        // exact bound.
```
martinthomson: Not sure that this comment is helping.  I think that the key is that the new unmarked range might include more than what was just unmarked.

However, I think that the logic is simpler than this.  If the cache exists, then the start offset will be `min(max(off, self.acked), cached_off)`.  The length is more complicated.

If the new range ends before the old range starts, then the length is the passed length.

If the new range ends at `cached_off` or later, then the length is extended to the old range (`cached_len.map(|len| len + cached_off - max(off, self.acked))`)

If the new range ends after the end of the cached range, there are two cases:
 * The end of the unmarked range cut a marked range.  In this case you can remember that (much like you remember `unmarked` currently, but you don't need to remember the offset).  In this case, the length is the passed-in value: `Some(len)`.
 * If the last marked range was completely removed, that is the only case where you need to scan.  That's the calculation you have, though you can scan `(start + len ..)` (adding 1 if you have the above).

martinthomson (commented):
There *might* be some more tests implied by the longer comment I added below...

martinthomson:
@larseggert the bot suggestion might be worth it.  I like that the code is a lot simpler, for one.  It's far easier to guarantee correctness of a minimum offset than trying to keep the tail synchronized.

larseggert:
This needs more work. I finally have a local stable benchmarking setup (Orange Pi Zero 3), so I can have the bot iterate in the design. It looks like spending effort to precisely set the cache vs. just invalidating it can cause regressions, if the setting runs more often and costs more than recovering from an unset cache rarely.

martinthomson:
Sounds good.  I marked this as draft then.

larseggert:
I don't think there is a win here.

PR #3962, opened 2026-09-10
fix: Keep the tail of a range that spans an unmarked range

larseggert:
Unmarking a range inside a larger sent range truncated that range at the start of the unmarked one and discarded everything past its end, so a single lost frame in the middle of a coalesced range caused everything after it to be retransmitted.

### DIFF
--- a/neqo-transport/src/send_stream.rs
+++ b/neqo-transport/src/send_stream.rs
@@ -399,42 +399,34 @@ impl RangeTracker {
         let mut to_add = None;
 
         // Walk backwards through possibly affected existing ranges
-        for (cur_off, (cur_len, cur_state)) in self.used.range_mut(..off + len).rev() {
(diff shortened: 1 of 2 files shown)


neqo-transport/src/send_stream.rs
```diff
@@ -410,7 +410,13 @@ impl RangeTracker {
                             off + len
                         );
                     } else {
+                        let cur_end_off = *cur_off + *cur_len;
                         *cur_len = off - cur_off;
+                        if cur_end_off > end_off {
+                            // This range spans the unmarked one, so keep its tail.
+                            assert_eq!(to_add, None);
```
```suggestion
```

martinthomson (approved):
I'll let you work out if you want to dive deep in the tests the way the LLM is suggesting.  It doesn't strike me as super important.

PR #3964, opened 2026-09-10
fix: Pad to the PMTUD probe size only when sending a probe

larseggert:
A pending PMTID probe that cannot be sent (because it exceeds cwnd) must not cause the size of other packets sent to increase.


martinthomson (commented):
The bots seem to have picked up on some genuine issues here.

PR #3965, opened 2026-09-11
refactor: derive transport packet/crypto parameters instead of passing them

larseggert:


neqo-transport/src/connection/mod.rs
```diff
@@ -1727,7 +1725,7 @@ impl Connection {
             last_ecn_mark != ecn_mark && stats.ecn_rx_transition[last_ecn_mark][ecn_mark].is_none()
         }) {
             stats.ecn_rx_transition[last_ecn_mark][ecn_mark] =
-                Some((packet.packet_type(), packet_number));
+                Some((packet.packet_type(), packet.pn()));
```
martinthomson, line 1728: Noice.  Unrelated, of course.

martinthomson (approved):
I'll take this for the unrelated fix in postprocess_packet alone, but the rest is just pure cleanup.

martinthomson:
The bot comments I didn't dismiss look like something you should consider.

mxinden (approved):
nice cleanup

PR #3966, opened 2026-09-11
refactor: store qpack peer instruction stream id instead of passing it per call

larseggert:


neqo-qpack/src/decoder.rs
```diff
@@ -256,6 +258,17 @@ impl Decoder {
         self.local_stream_id
     }
 
+    /// # Panics
+    ///
+    /// When a stream has already been added.
+    pub fn add_recv_stream(&mut self, stream_id: StreamId) {
+        assert!(
```
martinthomson: Shouldn't we be returning a Result for this sort of thing?

neqo-qpack/src/encoder.rs
```diff
@@ -82,9 +81,23 @@ impl Encoder {
...
         }
     }
 
+    /// Decoder stream has been created. Add the stream id.
+    ///
+    /// # Panics
+    ///
+    /// If a stream has already been added.
+    pub fn add_recv_stream(&mut self, stream_id: StreamId) {
+        assert!(
```
martinthomson: As above.

neqo-qpack/src/encoder.rs
```diff
@@ -421,8 +423,11 @@ impl Encoder {
         // by the main loop.
         let mut encoder_blocked = self.send_encoder_updates(conn).is_err();
 
-        let mut encoded_h =
-            HeaderEncoder::new(self.table.base(), self.use_huffman, self.max_entries);
+        let mut encoded_h = HeaderEncoder::new(
+            self.table.base(),
+            self.use_huffman,
+            self.table.capacity() / 32,
```
martinthomson: This needs a constant!

martinthomson (approved):

PR #3967, opened 2026-09-11
refactor: remove duplicated stream/session id storage in neqo-http3

larseggert:


martinthomson (approved):
This one seems to be a wash in terms of red/green ratio, but I think it's every so slightly cleaner out the other end.  Not sure.

PR #3970, opened 2026-09-11
refactor: remove redundant stream_id param on SendData::send

larseggert:


test-fixture/src/sim/connection.rs
```diff
@@ -243,8 +243,9 @@ impl SendData {
         }
     }
 
-    fn send(&mut self, c: &mut Connection, stream_id: StreamId) -> GoalStatus {
+    fn send(&mut self, c: &mut Connection) -> GoalStatus {
         const DATA: &[u8] = &[0; 4096];
+        let stream_id = self.stream_id.expect("stream created");
```
martinthomson: Is this better as:

```suggestion
        self.make_stream(c);
        let Some(stream_id) = self.stream_id else { return GoalStatus::Waiting; };
```

There's the possibility that the stream can't be made, but this would be a more robust.

test-fixture/src/sim/connection.rs
```diff
@@ -269,8 +270,7 @@ impl Goal for SendData {
     }
 
     fn process(&mut self, c: &mut Connection, _now: Instant) -> GoalStatus {
-        self.stream_id
-            .map_or(GoalStatus::Waiting, |stream_id| self.send(c, stream_id))
+        self.stream_id.map_or(GoalStatus::Waiting, |_| self.send(c))
```
martinthomson: If you take what I have above, this becomes:


```suggestion
        self.send(c))
```

martinthomson (approved):
Good change, but could it be better?

PR #3974, opened 2026-09-11
fix: drop HTTP/3 handlers whose connection is gone

larseggert:
A handler is only removed on observing `Http3State::Closed`, which does not arrive when the transport retires the connection first.

This means a longer-running server accumulated entries, causing slowdowns.

### DIFF
--- a/neqo-http3/src/server.rs
+++ b/neqo-http3/src/server.rs
@@ -15,7 +15,7 @@ use std::{
 
 use neqo_common::{Datagram, qtrace};
 use neqo_transport::{
-    ConnectionIdGenerator, Output, OutputBatch,
+    ConnectionIdGenerator, Output, OutputBatch, State,
     server::{ConnectionRef, Server, ValidateAddress},
 };
 use nss::{AntiReplay, Cipher, PrivateKey, PublicKey, ZeroRttChecker};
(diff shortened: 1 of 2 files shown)


neqo-http3/src/server.rs
```diff
@@ -140,15 +140,25 @@ impl Http3Server {
...
+                .process_multiple(Option::<Datagram>::None, now, max_datagrams);
+            if !matches!(out, OutputBatch::DatagramBatch(_)) {
+                self.http3_handlers.retain(|c, _| {
+                    let State::Closed(error) = c.borrow().state().clone() else {
+                        return true;
+                    };
+                    self.events
+                        .connection_state_change(c.clone(), Http3State::Closed(error));
+                    false
+                });
```
martinthomson:
```suggestion
                    if let State::Closed(error) = c.borrow().state().clone() {
                        self.events
                            .connection_state_change(c.clone(), Http3State::Closed(error));
                        false
                    } else {
                        true
                    }
                });
```

neqo-http3/src/server.rs
```diff
@@ -604,6 +615,69 @@ mod tests {
...
+                {
+                    closed_count += 1;
+                }
+            }
+            _ = server.process_output(t + Duration::from_millis(i + 1));
+        }
+        assert_eq!(closed_count, 1, "Closed must be emitted exactly once");
+    }
+
+    /// A handler must not outlive its connection.
```
martinthomson:
```suggestion
    /// A handler must not be retained once its connection is closed.
```

neqo-http3/src/server.rs
```diff
@@ -604,6 +615,69 @@ mod tests {
...
+        let t = now() + Duration::from_secs(30);
+        _ = server.process(bout.dgram(), t);
+
+        let mut closed_count = 0;
+        for i in 0..5 {
+            while let Some(e) = server.next_event() {
+                if let Http3ServerEvent::StateChange {
+                    state: Http3State::Closed(_),
+                    ..
+                } = e
```
martinthomson:
```suggestion
                if matches!(e, Http3ServerEvent::StateChange {
                    state: Http3State::Closed(_),
                    ..
                }) {
```

(Not sure about fmt implications, but I prefer `if matches!(...)` to `if let ...` for readability.

neqo-http3/src/server.rs
```diff
@@ -604,6 +615,69 @@ mod tests {
...
+        _ = b.stream_send(sid, b"ping");
+        let bout = b.process_output(now());
+
+        // Past A's drain period; B's ACK makes the output step return early, skipping retain.
+        let t = now() + Duration::from_secs(30);
+        _ = server.process(bout.dgram(), t);
+
+        let mut closed_count = 0;
+        for i in 0..5 {
+            while let Some(e) = server.next_event() {
```
martinthomson: You could also do `closed_count += server.events().filter(|e| matches!(e, ...)).count();`

neqo-http3/src/server.rs
```diff
@@ -604,6 +615,69 @@ mod tests {
...
+            "{} handlers left for {CONNECTIONS} closed connections",
+            server.http3_handlers.len()
+        );
+        assert!(server.events().any(|e| matches!(
+            e,
+            Http3ServerEvent::StateChange {
+                state: Http3State::Closed(_),
+                ..
+            }
+        )));
```
martinthomson: This should count the number of closes as well.

martinthomson (approved):

PR #3975, opened 2026-09-11
fix: exit the client once the connection close is sent

larseggert:
The closing period is 3xPTO of waiting with no work left. Since the benchmarks include this when timing, it can cause variations that are not due to stack behavior.

Gecko does not block on it either: `Http3Session::CloseInternal` hands neqo the close and releases its transactions immediately, so this gets us closer to Gecko's behavior.

### DIFF
--- a/neqo-bin/src/client/mod.rs
+++ b/neqo-bin/src/client/mod.rs
@@ -443,10 +443,12 @@ impl<'a, H: Handler> Runner<'a, H> {
             match (handler_done, self.client.is_closed()?) {
                 // more work; or no more work, already closing connection
                 (true, CloseState::Closing) | (false, _) => {}
-                // no more work, closing connection
+                // no more work, close and stop
                 (true, CloseState::NotClosing) => {
                     self.client.close(now(), 0, "kthxbye!");
-                    continue;
+                    // Skip the closing period (RFC 9000 Section 10.2) after flushing.
+                    self.process_output().await?;
+                    break;
                 }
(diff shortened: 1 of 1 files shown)


neqo-bin/src/client/mod.rs
```diff
@@ -443,10 +443,12 @@ impl<'a, H: Handler> Runner<'a, H> {
...
                 // more work; or no more work, already closing connection
                 (true, CloseState::Closing) | (false, _) => {}
-                // no more work, closing connection
+                // no more work, close and stop
                 (true, CloseState::NotClosing) => {
                     self.client.close(now(), 0, "kthxbye!");
-                    continue;
+                    // Skip the closing period (RFC 9000 Section 10.2) after flushing.
+                    self.process_output().await?;
+                    break;
```
martinthomson, line 451: Are there any cases where we need to hold?  I'm thinking about those cases where session tickets are needed.  I guess that your plan here is to let that use the fallback.

martinthomson (approved):

neqo-bin/src/client/mod.rs
larseggert: I don't think there is a case where we need to stick around. Once we are in State::Closing, incoming packets are not processed anymore. And resumption tokens are being handled before we go into `closed()`.

larseggert:
Flagging this for @mxinden to look at, since the client code was mostly his.

mxinden (approved):

PR #3980, opened 2026-09-13
fix: use peer capacity for QPACK insert count encoding

### LINKED ISSUE #3973: QPACK encoder computes `MaxEntries` from locally-clamped capacity
RFC 9204 §4.5.1.1 defines `MaxEntries = floor(MaxTableCapacity / 32)`, where `MaxTableCapacity` is "the maximum capacity of the dynamic table **as specified by the decoder**" (§3.2.3: the value the decoder announces via `SETTINGS_QPACK_MAX_TABLE_CAPACITY`). Both sides must agree on `MaxEntries`, since it's the modulus used to encode/decode the Required Insert Count on every header block.

`Encoder::set_max_capacity` (`neqo-qpack/src/encoder.rs`) receives exactly that peer-announced value as `cap` (see `neqo-http3/src/connection.rs`'s call with `settings.get(HSettingType::MaxTableCapacity)`), but clamps it to our own local budget before storing it:

```rust
let new_cap = min(self.max_table_size, cap); // self.max_table_size = max_table_size_encoder, a local setting
self.change_capacity(new_cap);
```

`max_entries` is derived from this clamped `new_cap` (via `self.table.capacity()`), not from the unclamped `cap`. Whenever `max_table_size_encoder` is configured below the peer's announced capacity, our `MaxEntries` ends up smaller than the value the peer's decoder uses to reconstruct `ReqInsertCount` from `EncInsertCount`. The two sides then disagree on the `2 * MaxEntries` modulus, and decoding can go wrong once insert counts wrap.

By contrast, `Decoder::new` computes `max_entries` directly from its own `max_table_size_decoder` setting — exactly the value it announces — so the decoder side is self-consistent.

This predates #3966 (that PR just replaced the stored `max_entries` field with an equivalent inline computation; the underlying clamped-capacity issue was already present).

**Fix:** `set_max_capacity` should retain the unclamped announced value separately and use that (divided by 32) for `MaxEntries`, rather than deriving it from the locally-clamped table capacity.

Hanabi9248:
Fixes #3973.

Keep the peer-advertised maximum separately from the locally capped table capacity, and use it for the Required Insert Count modulus required by RFC 9204 §4.5.1.1. The local memory limit still controls the actual table size.

The regression uses a 64-byte local limit and a 128-byte peer limit. After four acknowledged inserts, the prefix must encode 5; the old calculation produces 1. It also covers matching capacities.

Validation on Linux: 1,693 workspace tests passed (12 ignored), 45 clippy feature combinations passed, and documentation and dependency checks passed. The regression fails with the original modulus restored. Final nightly formatting passes.

Validation runs: [regression and negative control](https://github.com/Hanabi9248/neqo/actions/runs/34733094915), [workspace checks](https://github.com/Hanabi9248/neqo/actions/runs/34734105494), [final formatting](https://github.com/Hanabi9248/neqo/actions/runs/34735172630). The workspace run predates the final whitespace-only formatting change.


Hanabi9248:
I checked the reported upload regression with a separate comparison of d65f313 and 5953290 on a GitHub-hosted Ubuntu 24.04 runner (loopback MTU 1504, 100 samples per measurement). Three interleaved rounds of the 100 MiB upload benchmark gave these central time estimates:

| Round | Base (ms) | PR (ms) | Difference |
| --- | ---: | ---: | ---: |
| 1 | 153.69 | 156.63 | +1.91% |
| 2 | 150.51 | 151.48 | +0.64% |
| 3 | 154.96 | 151.73 | -2.08% |

[Workflow and logs](https://github.com/Hanabi9248/neqo/actions/runs/34763415407). The execution order was base/PR, PR/base, base/PR. I did not reproduce the 18.8% slowdown on this host, but it differs from the upstream on-prem runner, so this does not rule out that result.

PR #3982, opened 2026-09-14
Add transport outgoing-datagram queue engine

jesup:


jesup:

1. `trunk()`
1. **users/jesup/datagram_engine ← this PR**
1. [users/jesup/add_queue_to_connection](https://github.com/mozilla/neqo/pull/3983)
1. [users/jesup/hook_up_queue](https://github.com/mozilla/neqo/pull/3985)

---

larseggert:
@jesup naive question: why is this still 1.6K LoC? I thought the queue moved down into the QUIC layer now, should the adaptions you need me much lighter-weight?

jesup:
This patch is >50% tests.   Of the remainder, almost half are comments; there's ~400LoC of actual non-test code here, and this is the basis for all the other patches.  I could pull a few bits out of the initial patch, but it wouldn't dramatically reduce the size, and it would add yet more patches to the stack.

I will ask claude to re-evaluate all the patches and push on DRY to see if they can be made smaller.

jesup:
Also: the code saved was mostly just some api-layer-crossing code, which was a small part of this patch (it eliminated adding some API changes to transport, but this (in other patches) has a different set of API changes to receive data on max-age, send back expiry reasons, etc).  I moved it down here at Max's request to avoid two levels of queues and to centralize the handling.   Because I needed to move stuff down and redo interfaces, I took the opportunity to clean the stack to remove internal commits and provide a single clean commit, and also to minimize later PRs that modified the initial patch's design

jesup:
For reference, asking claude to apply (more) DRY to all the patches (including the ones I haven't uploaded yet) produced some small reductions, generally one function pulled out each in ~3 or 4 of the ~14 commits (in 10 or 11 PRs) that are currently uploaded -- one in this patch, but it didn't actually change the line count, or at most by 1 line.  So no major reductions.  Looking at several of the DRY patches, while they centralized code (good thing, reduces risk of changes not hitting every usage for example), none of them actually reduced LoC by more than a line or two, at most.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1528 @@
...
+//! Per-session outgoing-datagram scheduling: byte budget, high-water-mark
+//! backpressure, send-group/send-order priority, and max-age expiry.
+//!
+//! [`DatagramQueue`] holds one session's outgoing datagrams, ordered by
+//! send-group/send-order priority and bounded by both a byte budget and
+//! `outgoingMaxAge`, so a burst is delivered in priority order and anything
+//! left stale is shed instead of sent late. [`crate::quic_datagrams::QuicDatagrams`] holds
+//! one `DatagramQueue` per session (keyed by an opaque handle) and
+//! round-robins between them at packet-build time, so a datagram is never
+//! buffered anywhere without this age policy applying to it.
```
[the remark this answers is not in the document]
jesup: Partly fixed; target constructs `DatagramQueue`, so `clippy -D warnings` fails - is fixed byexporting the type. The docs are left describing the stack's endstate after the next PR or two; folding #3983 into this PR would make them true on their own

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1528 @@
...
+///   Equal-order datagrams are served FIFO.
+///
+/// [`crate::quic_datagrams::QuicDatagrams`] round-robins *between* sessions at packet-build time
+/// the same way this type round-robins between groups *within* a session; see its own doc comment.
+///
+/// This queue is scheduled independently of [`crate::send_stream::SendStreams`], the stream
+/// scheduler. A [`crate::streams::SendGroupId`] shared between a `WebTransportSendStream` and a
+/// datagram writable does not get cross-type starvation avoidance, as the WebTransport spec's
+/// send-order rules for a shared `SendGroupId` require: each scheduler is fair only among its own
+/// kind.
```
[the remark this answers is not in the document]
jesup: Already documented in the module, and still true: closing it means a single scheduler over `SendStreams` and this queue, which is a bigger change than this stack.  `peek_next_len` exposing no group/order metadata is a consequence of that, not an independent decision

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1528 @@
...
+                    *gid,
+                )
+            })
+            .min()
+    }
+
+    /// Evict the oldest datagram from the globally lowest-priority bucket.
+    ///
+    /// "Lowest priority" means the lowest `send_order` across all groups. Ties
+    /// are broken by `group_id` (lowest first) for determinism.
```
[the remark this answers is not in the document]
jesup: deliberate, and now documented. Under the byte budget something has to be dropped, and the application's own ordering is a better signal than group identity; picking the victim group first by age or occupancy trades that for group isolation.  Happy to switch if you prefer the latter

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1672 @@
...
+    /// [`DatagramQueueOutcome::Ok`] drains back below the high water mark,
+    /// for the caller to fire a resume signal. `false` every other time,
+    /// including every call while the queue was never blocked.
+    ///
+    /// Max-age expiry is the expected way a queue sheds load here, so a
+    /// caller that only checks this after a send stalls a blocked
+    /// application whenever the queue empties by expiring rather than by
+    /// sending.
+    pub fn resume_if_unblocked(&mut self) -> bool {
+        if self.blocked && self.below_watermark() {
```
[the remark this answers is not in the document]
jesup: FIxed and added a regression test

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1672 @@
...
+        // datagrams would pile up while the caller is not ready to take any.
+        let expired = self.expire(now, default_max_age);
+        let count = budget.min(self.total_count);
+        let mut to_send = Vec::with_capacity(count);
+        while to_send.len() < count {
+            let dgram = self
+                .take_next()
+                .expect("total_count bounds the loop, so a datagram is always available");
+            to_send.push(dgram);
+        }
```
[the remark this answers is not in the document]
jesup: simplified to iter::from_fn, dropping a redundant bound + expect

jesup:
I'll push an update for the nits that the AI review found

jesup:
> * `max_queued_bytes` is still not configurable outside the crate

Added the setter, symmetric with `set_high_water_mark`, and wired it up through `QuicDatagrams` and `Connection` the same way `set_datagram_high_water_mark` is.

The O(groups) cost is real but only pays once the O(1) gating check (`total_bytes + new_charge > max_queued_bytes`, which *does* run every `enqueue`) actually trips - i.e. only under sustained byte-budget pressure, expected to be rare. A bench of the rare path doesn't tell us anything about packet-build latency

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+        if group_empty {
+            self.groups.remove(&group_id);
+        }
+        Some(dgram)
+    }
+
+    /// Returns the outcome for the caller to apply backpressure with. On
+    /// [`DatagramQueueOutcome::Overflowed`] the caller reports a
+    /// [`DatagramOutcome::Dropped`] event for each tracked ID in `dropped`;
+    /// on [`DatagramQueueOutcome::Rejected`], for `id` itself.
```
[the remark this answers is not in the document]
jesup: right, and worse than an estimate issue: a full queue of already-stale traffic can reject or evict a fresh write in its favor.  Documented the same precondition `peek_next_len` already states, rather than threading `default_max_age` into `enqueue`'s signature; consistent with leaving the four-method `default_max_age` duplication from the last pass as a documented tradeoff

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+    pub fn enqueue(
+        &mut self,
+        data: Vec<u8>,
+        id: Option<DatagramId>,
+        now: Instant,
+        send_group_id: SendGroupId,
+        send_order: SendOrder,
+    ) -> DatagramQueueOutcome {
+        let new_charge = charge(data.len());
```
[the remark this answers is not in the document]
jesup: `enqueue` now shrinks the incoming buffer before charging it: cheapest of the three options.  Regression test included, confirmed to fail without the fix.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+            .map_or(Duration::MAX, |d| d / DEFAULT_MAX_AGE_RTT_MULTIPLIER_DEN),
+        DEFAULT_MAX_AGE_FLOOR,
+    )
+}
+
+/// Floor applied to an explicit `outgoingMaxAge`. Below this, the value stops
+/// bounding anything meaningful relative to our own timer granularity. There
+/// is deliberately no ceiling: an application that asks for a longer buffer
+/// than [`default_max_age`] gets it, bounded only by the byte budget.
+const EXPLICIT_MAX_AGE_FLOOR: Duration = Duration::from_millis(1);
```
martinthomson, line 114:
```suggestion
const EXPLICIT_MAX_AGE_FLOOR: Duration = GRANULARITY;
```

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+        qtrace!("Setting max age to {max_age:?} (clamped: {clamped:?})");
+        self.max_age = clamped;
+        self.expire(now, default_max_age)
+    }
+
+    /// Remove and return every datagram older than `max_age`, oldest first
+    /// within each send-group. Returns one entry per expired datagram,
+    /// `Some(id)` for tracked ones.
+    #[must_use]
+    pub fn expire(&mut self, now: Instant, default_max_age: Duration) -> Vec<Option<DatagramId>> {
```
martinthomson, line 430: The algorithm used for removing from a group in `expire_old` could be more efficient.  As you note elsewhere, the datagrams are stored in order, so any that could expire will be at the head.  You can use `VecDeque::partition_point` to do a binary search for that point and then either use `split_off` to cut the store in two OR use `drain(...pp)` to return a value.

The main wrinkle here is that `account_removed` wants its own mutable borrow.  That might push toward using `split_off()` and its owned copy.  But it might be better to inline the accounting instead.  You could do the accounting in the retain call.

Note that `expire_old` can return `impl Iterator<Item = QueuedDatagram>` to save an allocation there if you were to use `drain()`.  This function needs an owned `Vec` to hold multiple return values, so you don't have to worry about not visiting every value that was returned (a potential failure mode that comes from using `drain()`).

The result looks like this:

```rust
self.groups.retain(|_, group| {
    expired.extend(group.expire_old(now, max_age).filter_map(|dgram| {
        self.total_count -= 1;
        self.total_bytes -= charge(dgram);
        dgram.id
    });
    !group.is_empty()
});
expired
```

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+    /// account for their bytes.
+    ///
+    /// The boundary is inclusive, to agree with the deadline
+    /// [`DatagramQueue::next_expiry`] hands the timer: a caller that wakes at
+    /// exactly `timestamp + max_age` has to find something to shed, or it
+    /// reschedules the same instant forever.
+    fn expire_old(&mut self, now: Instant, max_age: Duration) -> Vec<QueuedDatagram> {
+        let mut expired = Vec::new();
+        self.by_order.retain(|_, queue| {
+            while queue.front().is_some_and(|d| d.age(now) >= max_age) {
```
martinthomson, line 266: See below regarding the use of `partition_point` and a more efficient drop algorithm.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+            {
+                outranked = true;
+                break;
+            }
+            let Some(dgram) = self.evict_lowest_priority() else {
+                break;
+            };
+            dropped.push(dgram.id);
+        }
+        if outranked && dropped.is_empty() {
```
martinthomson, line 533: This doesn't look right.  The loop above iterates until it frees enough space or it finds that the next thing to drop is higher priority than this new datagram.  That means that there is three possible outcomes:

1. Enough space is cleared by dropping lower priority data.  The return should be `Overflowed(Vec<...>)`.
2. Nothing could be dropped (the condition this tests).  The return value should be `Rejected`.
3. Something could be dropped, but not enough to make space for this new datagram.  That falls through, causing the data to be added to the queue above the maximum.

This third option is handled below as an overflow, which is basically OK.  However, the effect of that logic is that you might get some proportion of a single datagram over the maximum, which is surprising.

Could this be simplified a little so that rejection could be the same as overflow, except that the ID for this datagram is added to the overflow set?

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+                .lowest_priority_key()
+                .is_some_and(|victim| send_order < victim.0)
+            {
+                outranked = true;
+                break;
+            }
+            let Some(dgram) = self.evict_lowest_priority() else {
+                break;
+            };
+            dropped.push(dgram.id);
```
martinthomson, line 531: It's strange to be reporting a bunch of `None` values in `Overflowed`.  Why not:


```suggestion
            dropped.extend(dgram.id);
```

By this point, the datagram is owned, so this should take the `Option` as an `IntoIterator` and only add any ID that is `Some`.  The cost of that is that untracked datagrams don't trigger the `is_empty()` check below.  You'd need a boolean to track the need to return `Overflowed` with an empty list of IDs.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+            if self
+                .lowest_priority_key()
+                .is_some_and(|victim| send_order < victim.0)
+            {
+                outranked = true;
+                break;
+            }
+            let Some(dgram) = self.evict_lowest_priority() else {
+                break;
+            };
```
martinthomson, line 530: This performs a lookup for the lowest priority datagram twice.  Worse, if there is a need to drop multiple datagrams, this will loop and do the same work again.  In many cases, that will hit the same queue, because that's how apps work.

If you were to pass the amount of space needed to the function, you could have a group drop datagrams of the same order until ether space is available or you need to look for another group to trim (at which point the full scan is justified).

If you were to pass `send_order` in, that can be a cap on the groups you consider, which avoids the double check.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+    /// Remove and return the next datagram in scheduling order, advancing the
+    /// round-robin cursor. Nothing may mutate `self` between a
+    /// [`Self::peek_next_len`] and the `take_next` acting on it, or the two
+    /// can disagree about which datagram is next.
+    pub fn take_next(&mut self) -> Option<QueuedDatagram> {
+        let group_id = self.next_group_id()?;
+        let group = self.groups.get_mut(&group_id)?;
+        let order = group.highest_order()?;
+        let dgram = group.pop_front(order)?;
+        let drained = group.is_empty();
```
martinthomson, line 623: Inline this.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+                self.max_queued_bytes
+            );
+            // A full queue is backpressure whatever the high water mark says,
+            // and the caller is told to wait for a resume signal on every
+            // outcome but `Ok`: arm one, or it waits forever. Gated on
+            // `new_charge`, the exact charge this write needs, so a retry of
+            // this same write isn't told to resume before there is actually
+            // room for it.
+            self.blocked = true;
+            self.blocked_charge = new_charge;
```
martinthomson, line 550: I clearly don't understand the purpose of this variable.

In the event that a datagram is rejected, there will be some amount of data that would have overflowed the cap.  But it's not necessarily the size of that datagram.

And, as noted, there can be cases where the count can exceed the cap by some amount.

The bit that is a bit hard to follow is the bit where the blocked_charge is assigned to 1.  That doesn't match this line.

These calls also overwrite the value, as though this were a target.  But that doesn't track with how this is being used.

I suspect that the answer is that this variable isn't necessary.  It's sufficient to track that something blocked and to wait for the total number of bytes to drop below the limit.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+        outcome
+    }
+
+    /// The scheduling order's next group, starting at `rr_next` and wrapping.
+    /// `None` if the queue is empty.
+    fn next_group_id(&self) -> Option<SendGroupId> {
+        self.groups
+            .range(self.rr_next..)
+            .chain(self.groups.iter())
+            .map(|(id, _)| *id)
```
martinthomson, line 594:
```suggestion
            .map(|(id, _)| *id)
            .chain(self.groups.keys().copied())
```

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+
+    /// Remove and return the next datagram in scheduling order, advancing the
+    /// round-robin cursor. Nothing may mutate `self` between a
+    /// [`Self::peek_next_len`] and the `take_next` acting on it, or the two
+    /// can disagree about which datagram is next.
+    pub fn take_next(&mut self) -> Option<QueuedDatagram> {
+        let group_id = self.next_group_id()?;
+        let group = self.groups.get_mut(&group_id)?;
+        let order = group.highest_order()?;
+        let dgram = group.pop_front(order)?;
```
martinthomson, line 622: This looks like another double-dip that could be addressed by having the group pop the highest order item.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+            .take(budget)
+            .collect();
+        (expired, to_send)
+    }
+
+    /// Every remaining queued datagram, regardless of age. Used when the
+    /// session is closing and nothing will call [`Self::take_next`] or
+    /// [`Self::expire`] again.
+    #[must_use]
+    pub fn take_all(&mut self) -> Vec<QueuedDatagram> {
```
martinthomson, line 712: What needs this?  Because it's pretty inefficient and it would need a good reason.  If it is just to report on dropped IDs, that could be done differently.

Also, this should probably use `impl Iterator<Item = QueuedDatagram>` as a return value.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+    pub const fn len(&self) -> usize {
+        self.total_count
+    }
+}
+
+impl Default for DatagramQueue {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
martinthomson, line 768: Maybe pick one.  It seems like you could use a `Default` impl for this.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1906 @@
...
+    Rejected,
+    /// The incoming datagram was accepted, but the byte budget was
+    /// exceeded, so one or more of the lowest-priority datagrams already
+    /// queued were evicted to make room - a single incoming datagram can be
+    /// larger than what any one eviction frees. Always non-empty. Each
+    /// entry is an evicted datagram's tracking ID, or `None` if it was
+    /// enqueued untracked; the caller reports a [`DatagramOutcome::Dropped`] event
+    /// for each tracked one separately, to avoid reporting the same
+    /// eviction twice.
+    Overflowed { dropped: Vec<Option<DatagramId>> },
```
martinthomson, line 162: The `Option` here seems unnecessary to me.  It's not like the caller can do anything about a `None`, other than (maybe) counting them.

neqo-transport/src/streams.rs
```diff
@@ -39,7 +39,10 @@ pub type SendOrder = i64;
 ///
 /// A newtype around `u64` rather than a bare alias, so a raw integer (or a `SendOrder`)
 /// cannot be passed where a send-group id is expected.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+///
+/// `Ord`/`PartialOrd` are derived (numeric on the wrapped `u64`) so this can key a
+/// `BTreeMap`, used by the outgoing-datagram queue for deterministic round-robin.
```
mxinden:
```suggestion
```

Nit, I don't think this is necessary.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+/// Per-send-group priority queue.
+///
+/// Datagrams are organized by `send_order`; higher order = higher priority.
+/// Within the same `send_order`, datagrams are served FIFO (insertion order).
+#[derive(Debug, Default)]
+struct GroupQueue {
+    /// Keyed by `send_order` (ascending). Higher key = higher priority.
+    by_order: BTreeMap<SendOrder, VecDeque<QueuedDatagram>>,
+    /// Total datagram count in this group.
+    count: usize,
```
larseggert: Is this not always `by_order.len()`?

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+    /// removed as soon as it becomes empty, so every entry here is non-empty.
+    groups: BTreeMap<SendGroupId, GroupQueue>,
+    /// Group ID at which the next round-robin round starts. Persisted across
+    /// calls so a drain that stops on its budget resumes at the group after
+    /// the last one served; restarting at the lowest group ID every time
+    /// would starve higher-numbered groups whenever the budget is smaller
+    /// than the number of groups.
+    rr_next: SendGroupId,
+    /// Total datagram count across all groups.
+    total_count: usize,
```
larseggert, line 369: Do you need this as a field rather then having a `total_count()` fn that computes based on `groups`?

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+            .is_none_or(|mark| self.total_count < mark)
+    }
+
+    /// The `max_age` actually in force: the application's explicit
+    /// `outgoingMaxAge` if it set one, else `default_max_age`.
+    fn effective_max_age(&self, default_max_age: Duration) -> Duration {
+        self.max_age.unwrap_or(default_max_age)
+    }
+
+    pub fn set_high_water_mark(&mut self, mark: Option<usize>) {
```
mxinden:
```suggestion
    pub fn set_high_water_mark(&mut self, mark: Option<NonZeroUsize>) {
```

Does a high water mark of `0` make sense, given that a datagram is always queued and never directly send?

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+
+        outcome
+    }
+
+    /// The scheduling order's next group, starting at `rr_next` and wrapping.
+    /// `None` if the queue is empty.
+    fn next_group_id(&self) -> Option<SendGroupId> {
+        self.groups
+            .range(self.rr_next..)
+            .chain(self.groups.iter())
```
larseggert:
```suggestion
        .next()
        .or_else(|| self.groups.iter().next())
```

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+
+    /// Expire every datagram that has reached `max_age`. Returns the expired
+    /// datagrams themselves (not just their IDs), so the caller can also
+    /// account for their bytes.
+    ///
+    /// The boundary is inclusive, to agree with the deadline
+    /// [`DatagramQueue::next_expiry`] hands the timer: a caller that wakes at
+    /// exactly `timestamp + max_age` has to find something to shed, or it
+    /// reschedules the same instant forever.
+    fn expire_old(&mut self, now: Instant, max_age: Duration) -> Vec<QueuedDatagram> {
```
larseggert: GroupQueue::expire_old allocates a fresh Vec<QueuedDatagram> per group (line 264), returns it, and then DatagramQueue::expire extends all of them into a second all_expired: Vec<QueuedDatagram>, then maps that into a third Vec<Option<DatagramId>> via collect() (line 454)

The intermediate Vec<QueuedDatagram> exists only to hold datagrams long enough to read data.len() for accounting. If expire_old instead accepted a &mut Vec<Option<DatagramId>> output parameter and returned (count_expired, bytes_freed), the whole chain collapses to one allocation — the result Vec — and datagram data is freed inline rather than kept alive until the accounting loop.

Then expire collects the accounting totals in two local usizes inside retain and applies them after — one Vec, no intermediate copies.

larseggert (commented):
The comments are still VERY wordy.

jesup:
> The comments are still VERY wordy.

True.  I prefer wordy comments in many cases, because it provides context as to the 'why' and other constraints when I (or someone) comes back to this later.  I also think (but haven't tested) that it improves the results when using AI tools on the code in the future (for similar reasons).  That said, I can do a pass to reduce wordiness

neqo-transport/src/datagram_queue.rs
jesup: If we return Overflowed{} for case 2 instead of Rejected, then for an untracked write (id: None, which is a real, supported case: enqueue's id parameter is Option<DatagramId>):
   Case 2: rejected: Overflowed { dropped: [] } (nothing to add -- it's untracked). 
   Case 3: Admitted-with-untracked-eviction: also Overflowed { dropped: [] }. 
This is an identical return value for two opposite outcomes: did the datagram make it into the queue or not?
For a tracked write this is ok (you can check whether your own ID shows up in dropped), which is presumably what you had in mind.

Even if we accept that merger, it only changes the enum; it doesn't remove any code.  `blocked = Some(charge)` still has to exist and still has to track the case a rejected retry needs.  Whether that path is called Rejected or Overflowed{}, the queue still has to remember "the next retry needs N bytes free before I tell it to resume."  So changing Rejected to Overflowed{} doesn't simplify the actual resume logic.

neqo-transport/src/datagram_queue.rs
jesup: Thanks - fixed.   The eviction loop now derives the victim (order, group) once per bucket rather than once per evicted datagram, draining a whole same-priority burst (like a video frame's datagrams) via a new evict_at(group_id, order) before re-checking anything.

This saves when a bucket holds more than one datagram; i.e., when an app enqueues several datagrams at the identical  (group, send_order).  However, this may be a common case for things like e.g. video frames in a webrtc-like application.  It adds some code and complexity, but this is a hot path.

neqo-transport/src/datagram_queue.rs
jesup: I rewrote a number of things that eliminated this as part of that work.

neqo-transport/src/datagram_queue.rs
jesup: `new()` is `const fn`, `Default::default()` can't be.  Since new isn't needed elsewhere in the stack, just going with Default.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,1960 @@
...
+/// what is actually queued either way, and is a backstop only:
+/// `outgoingMaxAge` is what actually bounds delay in the common case, by
+/// shedding stale datagrams before this budget is ever reached.
+///
+/// 256KB is one [`DEFAULT_MAX_AGE_FLOOR`] (20ms) window at 100 Mbps, so
+/// anything deeper would expire before it could plausibly be sent; it is also
+/// ~207 datagrams at a 1200B MTU-sized payload once [`PER_DATAGRAM_OVERHEAD`]
+/// is charged, comfortably above a 64-datagram GSO-batch floor a single
+/// `sendmmsg` call can amortize.
+const DEFAULT_MAX_QUEUED_BYTES: usize = 256 * 1024;
```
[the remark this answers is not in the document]
jesup: Will address in #3983

neqo-transport/src/datagram_queue.rs
jesup: Not always `by_order.len()` - it's the *datagram* count, and one bucket can hold several (the same-priority-burst case from the eviction-loop discussion).  But looking into this showed the field itself was unnecessary: it was only ever compared to zero, and `by_order` never holds an empty bucket, so `by_order.is_empty()` already answers that for free.  Removed the field.

neqo-transport/src/datagram_queue.rs
jesup: Kept as a field, unlike its `GroupQueue` sibling above - `below_watermark()` needs the actual count, not just emptiness, and runs on every `enqueue`, so computing it on demand would mean an O(groups) sum on a hot path.  Added a doc note explaining the asymmetry.

neqo-transport/src/datagram_queue.rs
jesup: Good question, and you're right that nothing validates `Some(0)` on this side - the doc already says that mapping (including NaN/infinities) is the WebIDL caller's job.  Making it unrepresentable here would need the type to change through a bunch of code already stacked on top of this commit, to the WebIDL boundary.

neqo-transport/src/datagram_queue.rs
jesup: Fixed as suggested - `expire_old` now writes ids directly into the caller's output vec and returns bytes freed instead of returning owned `QueuedDatagram`s for `expire` to re-walk.  One allocation instead of three, and datagram payloads don't outlive the drain that reads them.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+    /// larger than what any one eviction frees. `evicted` is the total
+    /// number evicted, tracked or not, for aggregate stats; `dropped` holds
+    /// only the tracking ID of each evicted datagram that had one (an
+    /// untracked eviction needs no further caller action, so it contributes
+    /// nothing to `dropped` - `dropped` can be shorter than `evicted`, or
+    /// empty, even though eviction happened), for the caller to report a
+    /// [`DatagramOutcome::Dropped`] event for each one separately.
+    Overflowed {
+        dropped: Vec<DatagramId>,
+        evicted: usize,
```
mxinden:
```suggestion
        dropped_ids: Vec<DatagramId>,
        dropped: usize,
```

I am a bit confused by the naming. If I understand correctly all _dropped_ datagrams are as well evicted and all _evicted_ are as well _dropped_. How about the suggestion above?

neqo-transport/src/datagram_queue.rs
mxinden: > Good question, and you're right that nothing validates `Some(0)` on this side - the doc already says that mapping (including NaN/infinities) is the WebIDL caller's job. Making it unrepresentable here would need the type to change through a bunch of code already stacked on top of this commit, to the WebIDL boundary.

If it is checked in WebIDL, let's make it unrepresentable here, i.e. use a `NonZeroUsize` throughout the stack please.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+            .collect()
+    }
+
+    #[test]
+    fn datagram_outcome_id_returns_the_tracking_id() {
+        assert_eq!(DatagramOutcome::Expired(7).id(), 7);
+        assert_eq!(DatagramOutcome::Dropped(9).id(), 9);
+    }
+
+    // ── Basic behaviour ────────────────────────────────────────────────────────
```
mxinden: We don't use these test headings elesewhere, please either remove or use Rust modules.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+        drop(q.enqueue(vec![0, 1], Some(1), t, g(0), 1));
+        drop(q.enqueue(vec![0, 2], Some(2), t, g(0), 3));
+        drop(q.enqueue(vec![0, 3], Some(3), t, g(0), 1)); // same as id=1
+        drop(q.enqueue(vec![0, 4], Some(4), t, g(0), 3)); // same as id=2
+
+        // Expected: id=2 then id=4 (order 3), then id=1 then id=3 (order 1)
+        assert_eq!(drain_ids(&mut q), vec![2, 4, 1, 3]);
+    }
+
+    // ── Round-robin between groups ─────────────────────────────────────────────
```
mxinden: Same. Please either remove or use modules.

larseggert:
@jesup clippy is unhappy.

neqo-transport/src/datagram_queue.rs
jesup: I did ripple through a change to the gecko side; I can look at an explicit unrepresentable value now

neqo-transport/src/datagram_queue.rs
mxinden: As far as I am aware, nothing in Firefox uses `DatagramId`. Thus I would as well be fine removing `dropped: Vec<DatagramId>` entirely.

neqo-transport/src/datagram_queue.rs
jesup: Done

neqo-transport/src/datagram_queue.rs
jesup: connect-udp sends them down (via neqo_http3conn_connect_udp_send_datagram, from OnProcessDatagram: "mSession->SendHTTPDatagram(mStreamId, data, mTrackingId++);"), but I believe it doesn't look at drops.  WebTransport doesn't send them down.  So we can remove them -- but a bunch of tests rely on them to assert *which* datagram got evicted; this would have to change.  I'll look at that

jesup: checked against the my Gecko patch queue - confirmed WebTransport always sends untracked and connect-udp discards both the synchronous return value and the async outcome event today, so nothing currently reads it.  Removed the field; `Overflowed` now reports only the aggregate `dropped` count.  Reworked the handful of tests that depended on the ids for eviction-identity checks to derive the same invariant from real behavior instead (draining and checking what's left, or - for the one live-connection integration test - what the peer actually received).

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+                evicted: evicted_count,
+            }
+        } else if self.below_watermark() {
+            DatagramQueueOutcome::Ok
+        } else {
+            // The datagram was admitted, so any freed byte is progress
+            // for the next write, not a specific charge to wait for.
+            self.blocked = Some(1);
+            DatagramQueueOutcome::AboveWatermark
+        };
```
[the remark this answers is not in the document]
jesup: Rejected now does Some(self.blocked.map_or(new_charge, |c| c.max(new_charge))); the other two paths do Some(self.blocked.unwrap_or(1))

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+
+    fn push(&mut self, send_order: SendOrder, dgram: QueuedDatagram) {
+        // Sized for one entry: with distinct `send_order` values per datagram
+        // the default capacity-4 growth would cost ~4x `PER_DATAGRAM_OVERHEAD`
+        // per bucket, which the byte budget does not account for.
+        self.by_order
+            .entry(send_order)
+            .or_insert_with(|| VecDeque::with_capacity(1))
+            .push_back(dgram);
+    }
```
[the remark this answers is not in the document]
jesup: added a debug_assert! in GroupQueue::push checking the new datagram's timestamp against the bucket's last entry

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+        // Backwards-compatibility: with one group and equal priorities, behaves like before.
+        let mut q = DatagramQueue::default();
+        let t = now();
+        q.set_max_queued_bytes(3 * charge(1)); // room for exactly three 1-byte datagrams
+        drop(q.enqueue(vec![1], Some(1), t, g(0), 0));
+        drop(q.enqueue(vec![2], Some(2), t, g(0), 0));
+        drop(q.enqueue(vec![3], Some(3), t, g(0), 0));
+        drop(q.enqueue(vec![4], Some(4), t, g(0), 0));
+        assert_eq!(q.len(), 3);
+    }
```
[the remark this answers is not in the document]
jesup: renamed to byte_budget_evicts_fifo_within_one_order_bucket, asserts the exact outcome and drain order

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2058 @@
...
+        data.shrink_to_fit();
+        let new_charge = charge(data.len());
+        // A single incoming datagram can be larger than what any one eviction
+        // frees, so evict until there is room rather than at most once. If
+        // eviction empties the queue and the datagram alone still exceeds the
+        // budget, let it in anyway: refusing it would need a new error path,
+        // and the next enqueue evicts it immediately in turn.
+        let mut dropped = Vec::new();
+        let mut evicted_count: usize = 0;
+        let mut evicted_any = false;
```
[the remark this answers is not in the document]
jesup: removed, check evicted_count directly

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2009 @@
...
+            self.groups.remove(&group_id);
+        }
+        self.account_removed(&dgram);
+        dgram
+    }
+
+    /// Call after anything that can unblock the queue: a removal
+    /// ([`Self::take_next`], [`Self::expire`], [`Self::set_max_age`]) or a
+    /// relaxed bound ([`Self::set_high_water_mark`],
+    /// [`Self::set_max_queued_bytes`]). `true` once, the moment a queue that
```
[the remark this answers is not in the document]
jesup: added both - raising the high water mark and raising the byte budget, each asserting `resume_if_unblocked()` flips from `false` to `true` with nothing removed.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2057 @@
...
+            max_queued_bytes: DEFAULT_MAX_QUEUED_BYTES,
+            high_water_mark: None,
+            blocked: None,
+            max_age: None,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
larseggert: Yes, please add `#[cfg_attr(coverage_nightly, coverage(off))]` for all test modules you add.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2117 @@
...
+                .is_none_or(|last| dgram.timestamp >= last.timestamp),
+            "expire_old's partition_point needs non-decreasing timestamps per bucket"
+        );
+        // Sized for one entry: with distinct `send_order` values per datagram
+        // the default capacity-4 growth would cost ~4x `PER_DATAGRAM_OVERHEAD`
+        // per bucket, which the byte budget does not account for.
+        self.by_order
+            .entry(send_order)
+            .or_insert_with(|| VecDeque::with_capacity(1))
+            .push_back(dgram);
```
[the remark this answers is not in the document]
jesup: Reverted to `.or_default()` and dropped the stale justification comment.

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2117 @@
...
+        SendGroupId::new(id)
+    }
+
+    fn drain_ids(q: &mut DatagramQueue) -> Vec<u64> {
+        let (_, to_send) = q.drain(now(), usize::MAX, NO_DEFAULT);
+        to_send
+            .into_iter()
+            .map(|d| d.id.expect("test datagrams are tracked"))
+            .collect()
+    }
```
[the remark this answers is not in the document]
jesup: Actually this was ~100 instances

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2139 @@
...
+        // The queue should send highest send_order first.
+        let mut q = DatagramQueue::default();
+        let t = now();
+        enq(&mut q, vec![0, 1], 1, t, 0, 10); // order 10
+        enq(&mut q, vec![0, 2], 2, t, 0, 30); // order 30 (highest)
+        enq(&mut q, vec![0, 3], 3, t, 0, 20); // order 20
+
+        let sent = drain_ids(&mut q);
+        assert_eq!(sent, vec![2, 3, 1], "highest order first");
+    }
```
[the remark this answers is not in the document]
mxinden: @jesup worth adding a test for negative `SendOrder`?

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2139 @@
...
+    /// rejected - or evict a live datagram - to make room for entries that
+    /// are already past `max_age`.
+    pub fn enqueue(
+        &mut self,
+        data: Vec<u8>,
+        id: Option<DatagramId>,
+        now: Instant,
+        send_group_id: SendGroupId,
+        send_order: SendOrder,
+    ) -> DatagramQueueOutcome {
```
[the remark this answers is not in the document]
mxinden: I am fine ignoring this for now.

mxinden (approved):
No more comments from my end. Code looks clean, especially the `enqueue` function. Well done.

Preference to merge here after https://github.com/mozilla/neqo/pull/4006.

neqo-transport/src/datagram_queue.rs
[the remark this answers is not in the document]
jesup: Added

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2160 @@
...
+            .entry(send_order)
+            .or_default()
+            .push_back(dgram);
+    }
+
+    fn pop_front(&mut self, order: SendOrder) -> Option<QueuedDatagram> {
+        let queue = self.by_order.get_mut(&order)?;
+        let dgram = queue.pop_front()?;
+        if queue.is_empty() {
+            self.by_order.remove(&order);
```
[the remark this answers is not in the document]
jesup: Agreed it's worth benchmarking; per discussion with lars and mxinden this will be done after initial landings

neqo-transport/src/datagram_queue.rs
```diff
@@ -0,0 +1,2160 @@
...
+        enq(&mut q, vec![3], 3, t, 0, 0);
+        enq(&mut q, vec![4], 4, t, 0, 0);
+        assert_eq!(q.len(), 4);
+
+        // A 67-byte datagram (charge 131) needs to evict three of the
+        // existing 1-byte ones (each freeing only 65) before it fits: after
+        // evicting two, 130 bytes are free, and 130 + 131 still exceeds the
+        // 260-byte budget.
+        assert_eq!(
+            q.enqueue(vec![9; 67], Some(5), t, g(0), 0),
```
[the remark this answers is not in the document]
jesup: Done, all sites derive from PER_DATAGRAM_OVERHEAD now

PR #3983, opened 2026-09-14
Add outgoing-datagram queues to QuicDatagrams/Connection

jesup:


jesup:

1. `trunk()`
1. [users/jesup/datagram_engine](https://github.com/mozilla/neqo/pull/3982)
1. **users/jesup/add_queue_to_connection ← this PR**
1. [users/jesup/hook_up_queue](https://github.com/mozilla/neqo/pull/3985)

---

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +219,145 @@ impl QuicDatagrams {
...
+        &mut self,
+        session: StreamId,
+        data: Vec<u8>,
+        id: Option<DatagramId>,
+        now: Instant,
+        send_group_id: SendGroupId,
+        send_order: SendOrder,
+    ) -> DatagramQueueOutcome {
+        self.queue_mut(session)
+            .enqueue(data, id, now, send_group_id, send_order)
```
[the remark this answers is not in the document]
jesup: Correct, nothing enforces it on this path: `write_frames` only checks the path MTU.  The HTTP/3 layer does the check before enqueueing (#3989 in this stack), and I have corrected the doc here, which wrongly implied the peer's limit was applied at packet-build time; it now says the caller must not enqueue past `remote_datagram_size`, or anything at all while it is 0, and points at `send_datagram` for the checked path. I would rather not add a second error path to `enqueue_datagram` for a case its only caller already rejects.  I could move #3989 down to before this one if needed.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +219,145 @@ impl QuicDatagrams {
...
+    /// caller that wants to report per-datagram outcomes can use them.
+    pub fn expire_datagrams(
+        &mut self,
+        now: Instant,
+        default_max_age: Duration,
+    ) -> Vec<Option<DatagramId>> {
+        self.queues
+            .values_mut()
+            .flat_map(|queue| queue.expire(now, default_max_age))
+            .collect()
```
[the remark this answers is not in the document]
jesup: the IDs are returned; it is `process_timer` that ignores them, because at that layer there is nobody to report to. The HTTP/3 sweep added in #3986 pulls them per session and turns them into `DatagramOutcome::Expired` events (#3991), and `next_delay` guarantees a tick by the time anything is due, so the timer's own sweep only ever finds leftovers

neqo-transport/src/quic_datagrams.rs
```diff
@@ -81,8 +90,26 @@ pub struct QuicDatagrams {
...
+    /// but `QuicDatagrams` treats it as an opaque tag — this does not give
+    /// transport a "session" concept, any more than [`crate::send_stream::
+    /// SendStreams`] keying per-stream priority state by `StreamId` does.
+    ///
+    /// An entry is created lazily on first use and only removed by
+    /// [`Self::drop_session_datagrams`] (session teardown) — *not* whenever
+    /// it happens to drain empty, since a temporarily empty queue can still
+    /// carry an application-set high-water-mark/max-age that must survive
+    /// until the session closes.
+    queues: BTreeMap<StreamId, DatagramQueue>,
```
mxinden: Firefox does not support session pooling today, i.e. a QUIC connection is exclusively used by a single WebTransport session only. For the sake of reducing complexity, I don't think we should introduce the concept of sessions in `quic_datagrams.rs` already as part of Interop 2026. E.g. the above `BTreeMap` is unused, as we would only be tracking at most a single session only. I suggest removing the above session management till we tackle Bug 1815735.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1815735 for details.

neqo-transport/src/datagram_queue.rs
```diff
@@ -40,9 +40,8 @@ use crate::{
 ///
 /// 256KB is one [`DEFAULT_MAX_AGE_FLOOR`] (20ms) window at 100 Mbps, so
 /// anything deeper would expire before it could plausibly be sent; it is also
-/// ~207 datagrams at a 1200B MTU-sized payload once [`PER_DATAGRAM_OVERHEAD`]
-/// is charged, comfortably above a 64-datagram GSO-batch floor a single
-/// `sendmmsg` call can amortize.
+/// ~213 datagrams at a 1200B MTU-sized payload, comfortably above a
+/// 64-datagram GSO-batch floor a single `sendmmsg` call can amortize.
```
mxinden:
```suggestion
/// 64-datagram GSO-batch floor a single `sendmsg` call can amortize.
```

Nit: we don't use Linux's sendmmsg. We use the more modern GSO sendmsg.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
 
+    /// [`crate::datagram_queue::default_max_age`] evaluated against this
+    /// connection's current RTT estimate. Internal to `Connection` so
+    /// per-session datagram-queue callers below do not each need their own
+    /// route to the RTT estimate.
+    fn datagram_default_max_age(&self) -> Duration {
+        let min_rtt = self
+            .paths
+            .primary()
+            .map_or(Duration::ZERO, |p| p.borrow().rtt().minimum());
```
mxinden:
```suggestion
            .map_or(DEFAULT_INITIAL_RTT, |p| p.borrow().rtt().minimum());
```

Simplifies `datagram_queue.rs` `default_max_age`? Semantically I believe `DEFAULT_INITIAL_RTT` makes more sense as well.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -81,8 +90,26 @@ pub struct QuicDatagrams {
...
+    /// Datagram queued for sending via [`Self::add_datagram`]: a plain,
+    /// count-bounded FIFO with no per-session tracking, priority, or age
+    /// policy, for callers that don't need any of that.
     datagrams: VecDeque<QuicDatagram>,
+    /// Per-session outgoing-datagram queues (byte budget, high-water-mark,
+    /// send-group/send-order priority, max-age), keyed by an opaque `u64`
+    /// handle. In practice this is the session's control-stream `StreamId`,
+    /// but `QuicDatagrams` treats it as an opaque tag — this does not give
+    /// transport a "session" concept, any more than [`crate::send_stream::
+    /// SendStreams`] keying per-stream priority state by `StreamId` does.
```
mxinden:
```suggestion
    /// send-group/send-order priority, max-age), keyed by the session's control stream ID.
```

Agreed that it is fine for this concept to leak into `neqo-transport`. No need to detail the reason why it is fine.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +221,170 @@ impl QuicDatagrams {
...
+        send_order: SendOrder,
+    ) -> DatagramQueueOutcome {
+        self.queue_mut(session)
+            .enqueue(data, id, now, send_group_id, send_order)
+    }
+
+    /// See [`DatagramQueue::set_high_water_mark`].
+    pub fn set_datagram_high_water_mark(&mut self, session: StreamId, mark: Option<usize>) {
+        self.queue_mut(session).set_high_water_mark(mark);
+    }
```
mxinden, line 308:
```suggestion
    /// See [`DatagramQueue::set_high_water_mark`].
    pub fn set_datagram_high_water_mark(&mut self, session: StreamId, mark: Option<usize>) {
        self.queue_mut(session).set_high_water_mark(mark);
        if queue.resume_if_unblocked() {
            self.conn_events.datagram_space_available();
        }
    }
```

Would we want to wake the application here, if the water mark increased?

neqo-transport/src/streams.rs
mxinden:
```suggestion
```

I believe this is intuitive by the code alone.

neqo-transport/src/connection/mod.rs
jesup: It doesn't change the behavior (`default_max_age` already treats a zero RTT as "no sample yet" and substitutes `DEFAULT_INITIAL_RTT`) but this is clearer.

neqo-transport/src/quic_datagrams.rs
jesup: You're right - fixed.  Mirrors `set_datagram_max_age`'s existing resume check.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -914,3 +916,183 @@ fn datagram_fill_gap4() {
     datagram_overfill(&mut client, &mut server, 4);
     assert!(*called.borrow());
 }
+
+// ── Per-session queues (`Connection::enqueue_datagram` et al.) ─────────────
```
mxinden, line 920: We don't use test headings elsewhere. Preference for using a Rust module.

neqo-transport/src/connection/tests/datagram.rs
```diff
@@ -914,3 +916,183 @@ fn datagram_fill_gap4() {
...
+            ConnectionEvent::Datagram(data) => Some(data),
+            _ => None,
+        })
+        .collect();
+    assert_eq!(
+        payloads,
+        vec![vec![b'A'; 4], vec![b'B'; 4], vec![b'A'; 4], vec![b'B'; 4]],
+        "sessions must be served round-robin, not one drained before the other starts"
+    );
+}
```
mxinden, line 984: I believe this can be a lot shorter:


```suggestion
#[test]
fn per_session_queues_round_robin_across_sessions() {
    let (mut client, mut server) = connect_datagram();
    let now = now();

    for (session, tag) in [(0, b'A'), (4, b'B'), (0, b'A'), (4, b'B')] {
        _ = client.enqueue_datagram(
            StreamId::new(session),
            vec![tag],
            None,
            now,
            SendGroupId::new(0),
            0,
        );
    }

    let out = client
        .process_output(now)
        .dgram()
        .expect("four 1-byte datagrams fit in one packet");
    server.process_input(out, now);

    let order: Vec<u8> = server
        .events()
        .filter_map(|e| match e {
            ConnectionEvent::Datagram(d) => d.first().copied(),
            _ => None,
        })
        .collect();
    assert_eq!(
        order, b"ABAB",
        "sessions must be served round-robin, not one drained before the other starts"
    );
}
```

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+    /// connection's current RTT estimate. Internal to `Connection` so
+    /// per-session datagram-queue callers below do not each need their own
+    /// route to the RTT estimate.
+    fn datagram_default_max_age(&self) -> Duration {
+        let min_rtt = self
+            .paths
+            .primary()
+            .map_or(DEFAULT_INITIAL_RTT, |p| p.borrow().rtt().minimum());
+        crate::datagram_queue::default_max_age(min_rtt)
+    }
```
mxinden, line 4145: The current control flow seems overly complicated. `set_datagram_max_queued_bytes` calls `datagram_default_max_age` which calls down into `crate::datagram_queue` to return a `Duration` which is then again passed down into `crate::datagram_queue`. Why not simply pass the (default) RTT down into `crate::datagram_queue` on `set_datagram_max_queued_bytes` directly?

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+
+    /// Enqueue a datagram on `session`'s outgoing queue: byte-budgeted,
+    /// high-water-mark-tracked, and scheduled by `send_group_id`/
+    /// `send_order` and age. `session` is opaque to `Connection` — nothing
+    /// here requires it to be a real stream, only a stable per-caller tag
+    /// (in practice, a WebTransport session's control-stream `StreamId`).
```
mxinden, line 4151:
```suggestion
    /// (in practice, an Extended CONNECT session's control-stream `StreamId`).
```

MASQUE connect-udp will use the same code path.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+    /// `send_group_id` of `SendGroupId::new(0)` means ungrouped.
+    ///
+    /// Unlike [`Self::send_datagram`], this never fails: the *path* MTU is
+    /// applied later, at packet-build time, exactly like queued stream data.
```
mxinden, line 4155: Given that we plan to remove `send_datagram`, please make sure to remove this comment as well.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+    ///
+    /// The peer's limit is not, though: the caller must not enqueue anything
+    /// longer than [`Self::remote_datagram_size`], or anything at all while
+    /// that is `0` (the peer sent no `max_datagram_frame_size` and so does
+    /// not support DATAGRAM frames at all). Neither is checked here or at
+    /// packet-build time, and sending either violates the peer's transport
+    /// parameters. Callers that want the check made for them should use
+    /// [`Self::send_datagram`], which has an error path for it.
```
mxinden, line 4163: I suggest removing the whole paragraph. Also `send_datagram` is going away.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+        id: Option<DatagramId>,
+        now: Instant,
+        send_group_id: SendGroupId,
+        send_order: SendOrder,
+    ) -> DatagramQueueOutcome {
+        self.quic_datagrams
+            .enqueue_datagram(session, data, id, now, send_group_id, send_order)
+    }
+
+    /// See `DatagramQueue::set_high_water_mark`.
```
mxinden, line 4177: Could be a doc link.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+    pub fn set_datagram_high_water_mark(&mut self, session: StreamId, mark: Option<NonZeroUsize>) {
+        self.quic_datagrams
+            .set_datagram_high_water_mark(session, mark);
+    }
+
+    /// See `DatagramQueue::set_max_queued_bytes`.
```
mxinden, line 4183: Could be a doc link.

neqo-transport/src/connection/mod.rs
```diff
@@ -4116,6 +4132,102 @@ impl Connection {
...
+    pub fn set_datagram_max_queued_bytes(&mut self, session: StreamId, bytes: usize) {
+        self.quic_datagrams
+            .set_datagram_max_queued_bytes(session, bytes);
+    }
+
+    /// See `DatagramQueue::set_max_age`. Unlike the inner engine's own
```
mxinden, line 4189: Could be a doc link.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +218,174 @@ impl QuicDatagrams {
...
+
+    /// Expire stale datagrams on every session's queue, returning one entry
+    /// per expired datagram across all sessions (`Some(id)` for tracked
+    /// ones). Not gated on there being anything else to do: expiry is not a
+    /// send, so it must happen even when nothing else is scheduled (see
+    /// [`DatagramQueue::expire`]'s doc comment).
+    ///
+    /// Nothing currently reads the returned IDs outside tests, but a
+    /// caller that wants to report per-datagram outcomes can use them.
```
mxinden, line 368: I would prefer not introducing dead code. Either we use it, or we don't expose the `DatagramId`s.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +218,174 @@ impl QuicDatagrams {
...
+    ///
+    /// Expiry is the *expected* way one of these queues sheds load, so this
+    /// is also where a blocked sender is usually resumed: a queue whose
+    /// datagrams all expire while the application waits is never revisited
+    /// by a send, and the wait would never end.
```
mxinden, line 373: I find this comment confusing. With this function it is revisited, no? I suggest removing the whole comment.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +218,174 @@ impl QuicDatagrams {
...
+                return;
+            }
+        }
+    }
+
+    /// Starting at `start` and wrapping around, the first session with a
+    /// non-empty queue. Unlike the inner per-session `DatagramQueue`'s own
+    /// group map, `queues` keeps empty entries around (see its doc comment),
+    /// so this filters emptiness explicitly rather than relying on every
+    /// entry present being non-empty.
```
mxinden, line 253: I don't think this comment is needed.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +218,174 @@ impl QuicDatagrams {
...
+
+    /// Take the next datagram off `session`'s queue, resuming a blocked
+    /// sender and advancing the round-robin cursor past it. Must only be
+    /// called immediately after a [`DatagramQueue::peek_next_len`] on the
+    /// same session that returned `Some`, with no other mutation in between.
+    fn take_from_session_queue(&mut self, session: StreamId) -> QueuedDatagram {
+        let queue = self
+            .queues
+            .get_mut(&session)
+            .expect("next_active_session_from only returns known sessions");
```
mxinden, line 270: Nit: I would prefer these two invariants to be close to the call of `peek_next_len`, i.e. return `Option<QueuedDatagram>` from `take_from_session_queue` and `expect` in `write_frames`. Upholding invariants in the future across functions is difficult.

neqo-transport/src/quic_datagrams.rs
```diff
@@ -172,6 +218,174 @@ impl QuicDatagrams {
...
+        &mut self,
+        session: StreamId,
+        max_age: Option<Duration>,
+        now: Instant,
+        default_max_age: Duration,
+    ) -> Vec<Option<DatagramId>> {
+        let queue = self.queue_mut(session);
+        let expired = queue.set_max_age(max_age, now, default_max_age);
+        // Shrinking the limit can expire enough to put a blocked queue back
+        // under its high water mark, and nothing else will revisit it.
```
mxinden, line 326:
```suggestion
```

Nit: I don't think this is needed.

neqo-transport/src/datagram_queue.rs
mxinden, line 1: Is this pull request still correctly stacked on top of https://github.com/mozilla/neqo/pull/3982/? Because the changes below are already introduced in https://github.com/mozilla/neqo/pull/3982/.

neqo-transport/src/connection/tests/datagram.rs
jesup: Fixed, but kept `Some(1..4)` tracking ids rather than your `None`, so the test still exercises `DatagramTracking` in a round-robin scenario

neqo-transport/src/connection/mod.rs
jesup: Left as-is for now: this helper is shared by 4 call sites in this file; and pushing the conversion down into `quic_datagrams`/`datagram_queue` would relocate it, not remove a layer, and touches #3982's module boundary rather than this PR's.  If you still want this, let me know and I'll do it.

neqo-transport/src/connection/mod.rs
jesup: You're right. Trimmed both paragraphs (this and the following comment); kept the two facts that are still true (never fails on MTU; peer's limit isn't checked here) without naming the function that's going away.

neqo-transport/src/connection/mod.rs
jesup: (for this and the following doc-link comments):

I tried converting these to real intra-doc links, but `datagram_queue` is a private module, so that trips `rustdoc::broken_intra_doc_links` under CI's exact `cargo doc` invocation (fails even with `--document-private-items`, which only masks it locally).  Left as plain code-span text instead

neqo-transport/src/quic_datagrams.rs
jesup: These aren't actually dead — #3991, later in this stack, reads each returned ID to fire a per-datagram `Expired` outcome event.  Reworded the doc to point at that instead of saying nothing reads them

neqo-transport/src/quic_datagrams.rs
jesup: Left this one; it's explaining a non-obvious invariant (this map keeps empty entries, unlike the inner queue's own group map) rather than restating the code. Happy to reduce it further if you still think it's noise

neqo-transport/src/quic_datagrams.rs
jesup: Agreed, moved both `.expect()`s out to `write_frames`

neqo-transport/src/datagram_queue.rs
jesup: Not a real stacking problem; this PR's diff had a stray 1-line doc fix (`sendmmsg`→`sendmsg`) that landed here instead of in #3982 when an earlier review round's fixup got squashed in.  Will relocate it into #3982 so this PR's diff is clean of #3982 files

neqo-transport/src/quic_datagrams.rs
mxinden: > reads each returned ID to fire a per-datagram Expired outcome event. 

And is that per-datagram Expired outcome event then read?

neqo-transport/src/connection/mod.rs
mxinden: I am not opposed to a helper function, I am opposed to the indirection. You can keep a helper, e.g. `fn min_rtt` in `connection/mod.rs`.

jesup: Done.  min_rtt() is the only remaining helper now

neqo-transport/src/quic_datagrams.rs
jesup: I've dropped the per-datagram expiry results

PR #3984, opened 2026-09-14
Expose Connection::expire_datagrams for callers that need the expired IDs

jesup:

PR #3985, opened 2026-09-14
WebTransport: rewire the new datagram queue onto Connection's engine

jesup:


jesup:

1. `trunk()`
1. [users/jesup/datagram_engine](https://github.com/mozilla/neqo/pull/3982)
1. [users/jesup/add_queue_to_connection](https://github.com/mozilla/neqo/pull/3983)
1. **users/jesup/hook_up_queue ← this PR**

---

neqo-http3/src/connect_udp.rs
```diff
@@ -241,7 +244,10 @@ impl Handler for Http3Connection {
         id: I,
         now: Instant,
     ) -> Res<bool> {
-        self.extended_connect_send_datagram(session_id, conn, buf, id, now)
+        // connect-udp has no sendGroup/sendOrder concept of its own: always
+        // ungrouped, default priority.
+        self.extended_connect_send_datagram(session_id, conn, buf, id, now, SendGroupId::new(0), 0)
+            .map(|outcome| matches!(outcome, DatagramQueueOutcome::Ok))
```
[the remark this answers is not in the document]
jesup: Correct.  The test really does fail on this commit; `outgoing_datagram_queue(1)` sizes the legacy connection-wide FIFO, which neither WebTransport nor connect-udp sends through any more; this was resolved in #3987. Both test updates now happen here. The `Rejected`/`Overflowed` half is fixed at the source: the queue now arms its blocked flag on every outcome but `Ok`, so the resume event does arrive

neqo-http3/src/connection.rs
```diff
@@ -1623,18 +1628,85 @@ impl Http3Connection {
...
+    /// Expire stale outgoing datagrams on every active extended-CONNECT
+    /// session's queue, reporting how many were shed per session. Called
+    /// once per `process_http3` tick; `Connection::process_timer` (driven by
+    /// `next_delay`) already guarantees a tick happens by the time any
+    /// datagram's max-age is up, so this poll-based sweep never needs to
+    /// race an async transport event to catch an expiry promptly.
+    ///
+    /// Returns the total number of datagrams expired, for the caller to fold
+    /// into a stats counter.
+    pub(crate) fn expire_datagram_queues(&self, conn: &mut Connection, now: Instant) -> usize {
```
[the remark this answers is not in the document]
jesup: Right; clippy fails here because of it. The two `process_http3` call sites moved into this commit from #3986, so the sweep, `Session::expire_datagrams` and the stats hook all have callers now.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -390,6 +397,7 @@ impl Session {
     ) -> Res<()> {
         qdebug!("[{self}]: close_session");
         self.state = State::Done;
+        conn.drop_session_datagrams(self.id);
```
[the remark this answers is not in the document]
jesup: True, but there's nothing to report them through yet at this point in the stack: `DatagramOutcome` and the session's outcome path are in #3991, which routes exactly these IDs, and #3992 extends it to the peer-reset teardown. (No `unused_must_use` here: neither the method nor `Vec` is `#[must_use]`.)

neqo-http3/src/connect_udp.rs
```diff
@@ -241,7 +244,10 @@ impl Handler for Http3Connection {
         id: I,
         now: Instant,
     ) -> Res<bool> {
-        self.extended_connect_send_datagram(session_id, conn, buf, id, now)
+        // connect-udp has no sendGroup/sendOrder concept of its own: always
+        // ungrouped, default priority.
+        self.extended_connect_send_datagram(session_id, conn, buf, id, now, SendGroupId::new(0), 0)
```
mxinden, line 249:
```suggestion
        self.extended_connect_send_datagram(session_id, conn, buf, id, now, SendGroupId::default(), SendOrder::default())
```

Would be more descriptive, no?

jesup: `SendGroupId::new(0)` is the existing convention for ungrouped across this stack (see `send_stream.rs`'s `NULL_GROUP_ID`, and `enqueue_datagram`'s doc), including every other call site of this constructor already in the tree.  Adding `Default` just for this one call would make it the odd one out rather than the other way around.  If you want that,  we can do follow-up that adds `Default` and switches every `SendGroupId::new(0)` site over.

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -410,24 +418,31 @@ impl Session {
...
+    /// datagram support, an HTTP DATAGRAM Capsule.
     ///
     /// # Errors
     ///
-    /// `Error::Unavailable` if the session is not Active; other errors if
-    /// sending the datagram or capsule fails.
+    /// Returns an error if:
+    /// - The session is not in Active state (`Error::Unavailable`).
+    /// - `send_group_id` is neither `SendGroupId::new(0)` (ungrouped) nor a group already
+    ///   registered for this session's streams (`Error::InvalidInput`).
```
mxinden, line 428: Why "streamS" i.e. plural?

neqo-http3/src/features/extended_connect/session.rs
```diff
@@ -442,18 +457,83 @@ impl Session {
...
-            // Fullness is reported by the error above, never as `Ok(false)`.
-            return res.map(|()| true);
+            res?;
+            // This path never touches the queue, so it carries no
+            // backpressure signal.
+            return Ok(DatagramQueueOutcome::Ok);
+        }
+
+        if send_group_id != SendGroupId::new(0) && !self.validate_send_group(send_group_id) {
+            qdebug!("[{self}]: rejecting datagram with unregistered send group {send_group_id:?}");
```
mxinden, line 467: Instead of requiring an up-front registration, couldn't we as well lazily register new send groups? Not critical here.

neqo-http3/tests/connect_udp.rs
```diff
@@ -665,45 +665,24 @@ fn connect_udp_session_rejected_by_webtransport_create_stream() {
...
-        "resume event fired before the queue drained"
-    );
-
-    exchange_packets(&mut client, &mut proxy, false, None);
-    assert!(
-        client
-            .events()
-            .any(|e| matches!(e, Http3ClientEvent::OutgoingDatagramSpaceAvailable)),
-        "OutgoingDatagramSpaceAvailable was not forwarded through connect-udp"
+        Ok(true)
```
mxinden, line 685: This test is testing connect-udp backpressure, an essential property for a performant connect-udp stack.

Breaking backpressure on connect-udp is not an option. This was the whole puprose of https://github.com/mozilla/neqo/pull/3859. The solution is not to rewrite the test into a no-op, but to fix connect-udp backpressure, e.g. by always setting a high watermark, e.g. the previous `10`.

mxinden (changes requested):

neqo-http3/src/features/extended_connect/session.rs
jesup: Registration is per-session (`register_send_group` acts on the session, not individual streams); the streams just make use of whichever groups their session has registered.  Reworded to 'registered for this session' to avoid implying per-stream registration.

neqo-http3/src/features/extended_connect/session.rs
jesup: Leaving this as a followup

neqo-http3/tests/connect_udp.rs
jesup: I had a patch for this, it may have accidentally been part of the following PR because I didn't squash it into this one.  In any case, fixed.

mxinden: @jesup can you point me to the fix? The test is still broken in this pull request.

jesup: The watermark fix is `CONNECT_UDP_DATAGRAM_HIGH_WATER_MARK` in `neqo-http3/src/connection.rs` (built into every connect-udp session, since connect-udp has no API to set one); it had been squashed into the wrong PR and is in this one now. 

And you're right that the test was still a no-op: it's back as `outgoing_datagram_space_available_forwarded`,

PR #3994, opened 2026-09-14
feat: read server body data into a reusable buffer

larseggert:
The server allocated a zeroed MAX_EVENT_DATA_SIZE buffer per Data event and truncated it, so a 100 MiB body cost ~102k oversized allocations. Read into a buffer owned by Http3Server instead, and give each event an exactly-sized copy. Raise MAX_EVENT_DATA_SIZE to 32 KiB to match the stream I/O buffer size.


neqo-http3/src/server.rs
```diff
@@ -34,13 +34,16 @@ use crate::{
 
 type HandlerRef = Rc<RefCell<Http3ServerHandler>>;
 
-const MAX_EVENT_DATA_SIZE: usize = 1024;
+/// Maximum stream data per [`Http3ServerEvent::Data`], and the read buffer's size.
+const MAX_EVENT_DATA_SIZE: usize = 32 * 1024;
```
martinthomson, line 40: BTW, the mutant here is hilarious.

You mention in the PR comment that this is aligned with some other value.  Maybe reference that value.

martinthomson (approved):

PR #3995, opened 2026-09-14
feat: reuse reassembly buffers via a free list

larseggert:
Keep buffers freed from stream reassembly on a list shared by a connection's streams, so reassembly reuses them rather than allocating and growing one per RANGE_TARGET bytes.


neqo-transport/src/recv_stream.rs
```diff
@@ -1202,6 +1328,12 @@ impl RecvStream {
     /// `false` if the stream transitions to `AbortReading` or was already
     /// in a terminal or aborting state.
```
martinthomson: I don't know why LLMs do this, but you have displaced the doc comment on stop_sending.

neqo-transport/src/recv_stream.rs
```diff
@@ -190,16 +191,116 @@ impl RecvStreams {
...
+            let buffer = inner.buffers.pop().unwrap_or_default();
+            inner.retained -= buffer.capacity();
+            buffer
+        };
+        buffer.extend_from_slice(data);
+        self.wrap(buffer)
+    }
+
+    /// A buffer sized to `data`, for a range that will not grow.
+    fn exact(&self, data: &[u8]) -> RangeBuf {
```
martinthomson: Weird name.

Isn't it possible to have a single implementation?

```rust
fn wrap<T>(&self, data: T) -> RangeBuf where Vec<u8>: From<T> {
  RangeBuf { data: Vec::from(data), free_list: self.clone() }
}
```

martinthomson (commented):
This seems pretty tolerable, but the performance numbers don't really seem to justify the added complexity.  Aren't modern allocators sufficiently good at this stuff that we don't need to second-guess them?

larseggert:
If they were, I don't think we'd measure the improvement. But fair: @mxinden keeps telling me I micro-optimize too much :-)

larseggert:
The main benefit here is that we don't need to call into the allocator on the hot path. Besides the few percents of speed, the allocator seems to be the largest source of variations for the transfer benchmarks.

(Also, this speeds up the receive path, which is what is mostly used by Gecko.)

PR #4001, opened 2026-09-16
ci: Use CodSpeed-hosted runners for perf CI

larseggert:
We now have access to their [`amd-ryzen-9950x-x86_64`](https://codspeed.io/docs/features/macro-runners#amd-ryzen-9950x-x86_64) models.


larseggert:
The new CodSpeed runners aren't faster and don't have tighter variances than our own.

larseggert:
But there are more of them than our three. Let's give this another go.

PR #4002, opened 2026-09-16
fix(http3): map real send/close errors instead of HttpInternal

larseggert:
`SendMessage::send` mapped any error from `send_buffer`/ `stream_close_send` to `Error::HttpInternal`, which `map_error` debug-asserts can never happen — crashing debug builds whenever the transport stream errored on a real (documented) condition.

Use `map_stream_send_errors`, already used by `send_data`, so the real error surfaces instead. Widened its signature to `impl Into<Self>` so all callers drop their `.into()`/closure boilerplate.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2071513


neqo-http3/src/lib.rs
```diff
@@ -334,8 +334,8 @@ impl Error {
     ///
     /// On unexpected errors, in debug mode.
     #[must_use]
-    pub fn map_stream_send_errors(err: &Self) -> Self {
-        match err {
+    pub fn map_stream_send_errors<E: Into<Self>>(err: E) -> Self {
```
martinthomson: I wonder whether we just need a better implementation of `impl From<TransportError> for Error` instead.  If the implementation that `thiserror` gives us isn't right, we can write our own.  That would eliminate any need for mapping at all.

martinthomson (changes requested):
We can eliminate the mapping altogether, can't we?

The same goes for `map_stream_create_errors()` I think.

neqo-http3/src/lib.rs
larseggert: Awesome suggestion, thanks! That removes a whole bunch of code.

neqo-http3/src/send_message.rs
```diff
@@ -347,3 +336,52 @@ impl Display for SendMessage {
...
+            &mut client,
+        )
+        .unwrap();
+
+        client.stream_reset_send(stream_id, 0).unwrap();
+
+        assert!(matches!(
+            msg.send(&mut client, now()),
+            Err(Error::TransportStreamDoesNotExist)
+        ));
```
martinthomson, line 385: Should this also check that the connection doesn't break by asserting that the connection remains open?

neqo-http3/src/send_message.rs
```diff
@@ -248,22 +242,17 @@ impl SendStream for SendMessage {
...
+        let sent = self.stream.send_buffer(conn, now)?;
 
         qtrace!("[{self}] {sent} bytes sent");
         if !self.has_data_to_send() {
             if self.state.done() {
-                Error::map_error(
-                    conn.stream_close_send(self.stream_id()),
-                    Error::HttpInternal(6),
-                )?;
+                conn.stream_close_send(self.stream_id())?;
```
martinthomson, line 255: This is not going to map to an internal error, but I think that's OK.  Any error closing the stream would be amazing, so it's not going to cause things to break.

martinthomson (approved):

neqo-http3/src/webtransport.rs
```diff
@@ -385,11 +385,7 @@ impl ExportKeyingMaterial for Connection {
         wt_context.encode_vec(1, label);
         wt_context.encode_vec(1, context);
 
-        self.export_keying_material("EXPORTER-WebTransport", wt_context.as_ref(), out)
-            .map_err(|e| match e {
-                TransportError::InvalidInput => Error::InvalidInput,
-                other => Error::Transport(other),
-            })
+        Ok(self.export_keying_material("EXPORTER-WebTransport", wt_context.as_ref(), out)?)
```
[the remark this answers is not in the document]
martinthomson: This is a bad suggestion.  `Ok(x?)` is perfect.  Readable and concise.

neqo-http3/src/lib.rs
```diff
@@ -387,6 +351,20 @@ impl From<QpackError> for Error {
     }
 }
 
+impl From<TransportError> for Error {
```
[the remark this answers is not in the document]
martinthomson: It might be worth adding the `#[source]` annotation in this case, in place of the `#[from]`.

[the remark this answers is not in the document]
larseggert: Sorry, I don't follow. We don't always have a source for these?

PR #4004, opened 2026-09-17
perf: use a contiguous UDP receive buffer

larseggert:
Splits the receive-buffer allocation change out of #3948 so that PR remains focused on the UDP send path.

This stores receive slots in one boxed allocation and uses a 65,536-byte stride so contiguous slots retain the same page offset.


larseggert:
Not worth it.

